PR #12, opened 2015-09-11
Add Serbian translation

pejakm:


theScrabi:
Nice thanks.

theScrabi:
Since I have added a new preference option, i had to add another tag into the strings.xml. If you want to translate that take please do it until Monday. That's when i want to release NewPipe 0.3.5.

PR #20, opened 2015-09-16
Add Dutch translation

TheLastProject:
I have added a line saying that the Dutch XML file is UTF-8, because of the ü character, as I am not sure how Android deals with this. Please ensure the Dutch version does not crash.

Also, would you consider adding NewPipe to Weblate? It is Free of charge for Free Software projects, it is fully Free itself and F-Droid uses it as well: https://weblate.org/en/hosting/.


theScrabi:
Ok I'll see what i can do :)

theScrabi:
I reviewed the app on Weblate.

PR #30, opened 2015-09-19
Update video search list

darkon5:
I adapted the search list so it would look nicer. I can only test on my screen, but I think it's an improvement. On my screen I get to see one more video on list.

(I only changed video_item.xml and dimentxt.xml, the rest was automated by Android Studio)

Before
Before

After
After

PD: I'm kind of newbie in this Git and Android Studio stuff. I'll try to keep the weirdness to a minimun.

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <option name="DEFAULT_COMPILER" value="Javac" />
     <resourceExtensions />
     <wildcardResourcePatterns>
       <entry name="!?*.java" />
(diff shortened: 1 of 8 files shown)


theScrabi:
Hm ok I'll merge it later. I was also thinking about another change. I want to put the duration on top of the video like the original app does, and then put the creation date where the duration was.

darkon5:
You mean something like this? (the duration needs proper positioning)

screenshot_2015-09-20-13-48-30

theScrabi:
Hm i had put it in the bottom right corner and mad it transparent like the original. But thats ok to :)

theScrabi:
Actually i like that :) lets try it out.

darkon5:
Well, I was changing it to the bottom left of the thumbnail too xD 

screenshot_2015-09-20-14-50-07

Colors can be changed on durationback.xml
(Original white colors -> Stroke: #4fa5d5, BG: #ffffff)

PS: Sorry if I'm slowing you down in the development. I'm learning ASAP about Android dev.

theScrabi:
You are speeding it up !!! Now i don't need to do it my self. Thanks.

PR #34, opened 2015-09-20
Duration now shown on top of the thumbnail.

darkon5:
TODO: Replace "Upload date here" with actual upload time, change colors to taste.

(I leave these to you as I think you will make better changes than I could)

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <option name="DEFAULT_COMPILER" value="Javac" />
     <resourceExtensions />
     <wildcardResourcePatterns>
       <entry name="!?*.java" />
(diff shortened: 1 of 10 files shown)


darkon5:
+ Spanish translation :)

darkon5:
I made another version where the title is more thin and the upload date and author have different lines.
This is good for the videos with long author names, this way text won't glitch that often.

-> THIS IS NOT INCLUDED IN THIS PULL REQUEST <-

Should I add these changes too?

screenshot_2015-09-20-18-17-01

theScrabi:
Ah we have a problem. Sometimes the user name is so long, that the layout crashes. For now i put the itemUploadDateView benath the username, but that doesn't look pretty good.
screenshot_2015-09-20-20-58-26

darkon5:
In my smaller version it does not happen:

screenshot_2015-09-20-21-34-10

I'm going to make the final touches and send it for testing. It shouldn't crash with anything.

theScrabi:
k thanks

PR #36, opened 2015-09-20
3 Line Search List

darkon5:
Latest version of the list design, as the text is in different lines has more space and shouldn't get broken.


theScrabi:
k thanks.

PR #43, opened 2015-09-24
Set the title for external video players

epitron:
This makes the external player experience nicer.

I had to set the title twice on the intent, since "Intent.EXTRA_TITLE" didn't work on MX Player. (MX Player's supported intent data is here: https://sites.google.com/site/mxvpen/api).

I think I did that correctly. Anyhow, it works. :)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/ActionBarHandler.java
+++ b/app/src/main/java/org/schabi/newpipe/ActionBarHandler.java
@@ -201,11 +201,17 @@ public class ActionBarHandler {
         if(!videoTitle.isEmpty()) {
             if (PreferenceManager.getDefaultSharedPreferences(context)
                     .getBoolean(context.getString(R.string.useExternalPlayer), false)) {
+
+                // External Player
                 Intent intent = new Intent();
                 try {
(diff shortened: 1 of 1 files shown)


theScrabi:
Ok I see if i can also apply that to audio streaming. :) Thanks for the help.

epitron:
Np!

theScrabi:
Well i could set the title for music streams like you set it for video streams. But all it does is showing the title in the video player activity of VLC. In the audio only activity it displays the url again.

epitron:
I guess each player has specific quirks.

This looks like the point where VLC is getting the title from the intent: https://github.com/mstorsjo/vlc-android/blob/master/vlc-android/src/org/videolan/vlc/gui/video/VideoPlayerActivity.java#L1898

It seems to only check for the intent extras when `getAction().equals(PLAY_FROM_VIDEOGRID)`. I'm not sure what most of this function is doing, but maybe audio is a different `action`?

theScrabi:
That might be it.
Well I don't like VLC for streaming anyway. It has bad audio quality. So I somehow one day want to make a player for NP itself.

PR #64, opened 2015-10-30
add Persian translation

mamins1376:


theScrabi:
If you like you can use weblate for the future :)

PR #67, opened 2015-11-02
Remove Deprecated methods

shekhar12020:
+ Remove deprecated method from app/src/main/java/org/schabi/newpipe/VideoItemListFragment.java

onAttach(Activity activity) is replaced by onAttach(Context context)

+ Remove deprecated method from pp/src/main/java/org/schabi/newpipe/VideoListAdapter.java
context.getResource().getColor(resId) is replaced by ContextCompat.getColor(context,resId)


theScrabi:
All right thanks :)

PR #71, opened 2015-11-03
Add russian translation of strings.xml

chschtsch:


theScrabi:
Just a reminder, you can use weblate if you like.

PR #72, opened 2015-11-03
Code cleanup & minor additions

medavox:
* Added an entry in manifest to make PlayVideoActivity's parent activity VideoItemDetailActivity in a (partially successful) attempt to make the Back soft button take us back. (It requires 2 presses of the back button for some reason).

* Added an UnkownHostException handler to print an error when a user tries to search, and there's no internet connection

* Changed SteamingService interface methods so they return new instances of each class (Factory design pattern), rather than Class objects for each, which loses type safety. Also changed the calls to these methods.

* De-duplicated some error message codes about an unknown format in VideoInfo

* removed empty, unused method displayList from VideoItemListFragment

* General typo corrections for comments, when encountered

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.gitignore
 .gradle
 /local.properties
 /.idea/workspace.xml
@@ -5,3 +6,5 @@
 .DS_Store
 /build
 /captures
+#.idea/gradle.xml
+#.idea/misc.xml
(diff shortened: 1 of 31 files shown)


medavox:
This app has some pretty awesome scraping functionality in it!
It does a difficult job well, though I'd be happy to clean the code up further to improve readability, type safety etc?

theScrabi:
Permission is hereby granted ;D

medavox:
Awesome, thanks :)

PR #89, opened 2015-11-13
remove quotation marks added by weblate

chschtsch:


theScrabi:
Lol how comes that these are so much commits?

PR #100, opened 2015-11-17
Refactor

medavox:
So many changes, this needs discussing before pushing to master.

A summary:

+ Implemented timestamps
* renamed `VideoInfoItem` to `VideoPreviewInfo`
* Moved streaming service-related classes into their own, new package: "services"
+ Added javadoc to some classes and methods (where functionality is known well enough to explain)
- De-duplicated common fields between `VideoInfo` and `VideoPreviewInfo` by moving them into a common superclass: `AbstractVideoInfo`
- Removed 2 methods in `PlayVideoActivity` which only call `super()`, and therefore are unnecessary: `onResume()` and `onPostCreate(Bundle)`
+ Added `VideoInfo(AbstractVideoInfo)` constructor
    - to support converting `VideoPreviewInfo`s into `VideoInfo`s, to reuse scraped info (yet to be implemented)
* Made the Extractor class behave as a per-video object;
    - most method return values are video-specific, so it makes sense (to me) to have Extractor be stateful. 
    - The only stateless methods are getVideoUrl(), getVideoId() and loadDecryptionCode(String)
* Implemented a constructor for YoutubeExtractor, which performs all initialisation work, such as fetching `Jsoup.Document`, and `playerArgs:JSONObject`

I think that's everything; if I missed anything, this pull request diff will show it (the 'files changed' tab)

Feel free to ask questions and inspect my code :)


medavox:
I'll merge this some time on Saturday, if there are no objections.

chschtsch:
I tried to fix a new bug in this branch whith unformatted view count in next video and similar videos, but I feel like the code needs more refactoring. Something like moving all data formatting based on locale to service-specific Extractor or creating a new class for that. So you can access both raw and formatted data where- and whenever you need.

I'm not so good yet at understanding Java and Android, but if there is a method that is used more than once (like formatting data), it should be separated and be able to import from different place of the code. Maybe separating and ordering everything would make understanding of code easier not only for me. What do you think?

medavox:
I completely agree; in fact I did plan on doing that in this branch, but thought I should push the changes I *had* made first, so they could be seen by another pair of eyes before they diverged too far from master.

I'd like to replace the `Extractor.extractVideoPreviewInfo()` method with a series of field-extracting methods, like with `VideoInfo` (to which locale-specific formatting can be applied), but this would probably require the `VideoPreviewInfo` extraction functionality to pulled out into its own class, which seems a bit verbose.

Somewhere down the line, alongside the existing `VideoInfo` and `VideoPreviewInfo` data objects, we have to add objects which store information about Channels' pages, and Playlists.

And we have to generalise these data objects to all video streaming sites, or subclass them for each site (which would be harder to maintain).

But right now, I definitely think a centralised Locale-handling class would be a good idea. I'm thinking, store the raw values in `VideoInfo`/`VideoPreviewInfo`, then format them when they are needed by calling the locale-formatting methods from `VideoItemDetailFragment`/`VideoItemDetailActivity`.

I can't get much done today, and this weekend is family time, but I'll check up here when I can.

theScrabi:
@medavox I'm still not very confident with field-extracting methodes. If you have one methode doing everything, you could split it yourself, and parallelize it if it was necessary. I'm not a very good programmer, but i don't see how we could do that with field-extracting methods.

medavox:
Parallelising the field extraction methods could be done by overriding the Extractor constructor (which is where all the setup not specific to YouTube now takes place), and providing a more specific setup routine. There, worker threads could be instantiated, which each call a method.

Currently, Extractor is instantiated in a separate worker thread, as is the call to getVideoInfo(), so the heavy work still does not take place on the main thread.

I'm not 100% certain about this method-per-field design (as you may have noticed from me changing my mind ;) ), so I'd be happy if you wanted to change it back to one method for setting up the VideoInfo. After implementing background media play, I'd like to look at how to incorporate Playlists and Channels support into the existing group of classes implementing video metadata handling.

theScrabi:
Hm before adding new features, and new services, we should really talk about how the final architecture of our extractor should look like. I need to admit I'm not quiet confident with the method-per-field design :/, although is see the advantages of it. Maybe there is some middle way.

medavox:
Regarding method-per-field, that's completely fine!
And you are right about needing to discuss architecture.

The way I see it so far (only based on my dealing with YoutubeExtractor) we have 2 video metadata objects, with corresponding extraction routines:

Object                     | Info of |Extraction Routine
-------------------------|----------|--------------
`VideoInfo`              | the video being played | used to be `Extractor.getVideoInfo()`, now lots of methods
`VideoPreviewInfo` | videos previewed on that page/results of a search | currently `Extractor.extractVideoPreviewInfo()`

But in future we'll need to scrape info from playlist and channel pages, (each needing an Extractor function on its page), and also recognise them in preview lists (ChannelPreviewInfo, PlayPreviewInfo)?

I'm guessing there could be some common functionality, in the Java, and possibly in the HTML/JSON scraping methods.

So what do you think about these additions:

Object                     | Info of                            | Extraction Routine
-------------------------|-------------------------------|--------------
`ChannelInfo`?            | Stuff on a channel page(s):name, description, playlists/uploads | `Extractor.getChannelInfo()`?
`ChannelPreviewInfo`? | thumbnail of channel in search results | `Extractor.getChannelPreviewInfo()`?
`PlaylistInfo`? | looks like the search results page to users; basically the list of videos on a playlist page | `Extractor.getPlaylistInfo()`?
`PlaylistPreviewInfo`? | Playlist preview info, eg number of videos, thumbnail | `Extractor.getPlaylistPreviewInfo()`?

The pattern, obviously, is that each `PreviewInfo` is like a baby version of the full `Info` object, and contains a subset of the full `Info`'s fields. These common fields can be reused when creating a full `Info` from a `PreviewInfo` (say when a user clicks on a `VideoPreviewInfo`), reducing the amount of re-parsing needing doing.

What do you think? based on this, I'd say we could reuse a fair bit of the search result display functionality to implement playlist display. Although I haven't looked at the relevant HTML.

theScrabi:
I wouldn't do it through Extractor. If i where to design this, I would add an extra class for channels and videos. This is why SearchEngine and Extractor are both part of the service. Extractor it self is meant to only parse info for the video itself.
I designed it like this: Every youtube page gets its own class in NewPipe !!!

Also please consider a design that works well with unit tests (for CI). We should keep that in mind.

theScrabi:
Well I spoke with a college, and he said  method-per-field is the better design. Although I still think we should not let Extractor do everything. Also I'd like to rename Extractor into VideoExtractor, so its more clear what it's made for.

medavox:
OK then, thanks for getting another opinion. I wasn't really sure either, so it's good to hear what someone else thinks.

Right now I'm only working with the new `BackgroundPlayer` class (with a few changes in `ActionBarHandler`, to reference it), so my changes won't be disruptive to other parts of the code.

So if you have any free time at all, would you be able to write up some method names/class names and plain-language descriptions (no code - just method signatures and descriptions of what they should do), and we can fill in the method bodies later?

That way, we have a basic design to work to, and it also means you get to specify the design as you imagine it :) It'd be kind of like interfaces, except we don't need the interfaces after the classes have been finished.

We could even use UML or something, if you're happy with it. But I think this would be fine.

Because they won't be functional classes in their first form, we can put this in another branch maybe?

theScrabi:
Uff I never used UML before. So if we went this way, I might need time to learn it first.
At the moment I'm about to set NewPipe up for CI that way we can automatically get notified when youtube has changed it's website, and of course the build will be checked. And after that id like to care about Downloading. Since that is still our greatest mess. At the moment I think we don't need to hurry with the Extractor and Service part.

medavox:
OK, that sounds good to me.
As I said above, the BackgroundPlayer shouldn't interfere much (if at all) with other parts of the project; it's basically just an audio stream player, VLC, but internal.

We don't have to use UML. CI sounds like something that would definitely be useful, though.

PR #102, opened 2015-11-19
 fix bug with upload date + better way to parse it

chschtsch:
The method to parse a date with spliting a string, that was used before, was actually buggy and showed one month ahead of actual upload date in the layout, e.g. Sep 18, 2015 instead of Aug 18, 2015
SimpleDateFormat was used instead.


medavox:
This is much better. Thanks!

PR #106, opened 2015-11-27
Make some strings non-translatable

Xenotium:
Apart from app name, which at the moment is the same no matter what language, there are URLs here. They shouldn't reside here (URLs) but they (including app name) should be made at least "translatable=false".


chschtsch:
Nice.
I thought about that, but didn't know about this method.

PR #107, opened 2015-11-30
Slightly update readme file + add screenshots

chschtsch:


theScrabi:
Hm :/ make the screenshots smaller and please add the icon again.

theScrabi:
Also don't make screenshots of a version with layout problems xDDD.

theScrabi:
Here take this as an example:
https://github.com/grote/Transportr

chschtsch:
OK. The icon was kinda weird and small, so I decided to remove it for now. Also, it's better to put link to http://dasochan.nl/newpipe/ on top of GitHub page, where it belongs:

screen shot 2015-12-01 at 8 23 12 pm

Or also, we can make gh-pages for this site, what do you think?

Readme in Transportr is looking pretty, I like it.
Maybe we can make icon slightly bigger, like theirs. And yes, small screenshots look good too (useful when multiple pages).

However, layout is an issue. I tried to convert to CardView, but it's challenging for me. Have you worked on it yet? We can discuss it on Slack, because I can do it, if somebody will explain why it's not as simple as change xml tags names and adding design libraries.

theScrabi:
I already fixed stuff (as you se i used 3 commits to do so not very good but well).

chschtsch:
Ok, sorry, I have outdated info.

PR #115, opened 2015-12-17
Feature search suggestion

shekhar12020:
SearchBar suggestion feature implemented 
Created 2 java files for list adapter & list item view creator
+ SuggestionItemViewCreator.java
+ SuggestionListAdapter.java

Added xml file for list row
+ suggestion_item.xml

+Added Method searchSuggestion(String query) for call suggestion API added callback to handle suggestion list item.

+Call method inside SearchView:
        @Override
        public boolean onQueryTextChange(String newText) {
            listFragment.searchSuggestion(newText);
            return true;
        }


theScrabi:
Ok I'll try to test, and merge it the next few days :)

Issue #120, opened 2015-12-22
Thumbnails stretched on landscape view


iqqr:
Been using NewPipe since earlier versions on my tablet, lastest ones now get thumbnails stretched and the vídeo title cropped by the table row. It's a pain in the ass having to use portrait view instead of landscape because I need to keep it horizontal due batery problems and charger cord short stuff.

![newpipe](https://cloud.githubusercontent.com/assets/16397441/11950044/e34c60ec-a852-11e5-8a1b-52a53e371ef4.jpg)

Two versions back was fine now I don't know what changed in the current one. Kinda lacks its responsiveness.

![newpipe_old](https://cloud.githubusercontent.com/assets/16397441/11950195/3ab3df08-a854-11e5-8fc0-1ffb08cd15cf.jpg)

Note: Reinstalled, nothing fixed.

chschtsch:
@iqqr, thanks for your notice! We definitely need to fix this. The thing is that I don't have a tablet for testing, so I literally made these changes blindly.

I will make it the same size as normal portrait view for phones probably by tomorrow.

chschtsch:
You know what, actually, I'm already ready to push the changes:

![device-2015-12-22-160712](https://cloud.githubusercontent.com/assets/6196182/11954928/cf689e0e-a8bd-11e5-844a-1ed1d3115146.png)

Although, when I open the video, I get memory error in my emulator, so I would need a real device to make sure everything works fine. Maybe you could build a version from my fork to test it out?

iqqr:
Unfortunately I'm away from PC and wouldn't be able to build it on my own. Did you manage to compile some apk I could run? I wish I could be more of help.
BTW, I'm currently browsing on a 7" tablet with 1024x600px res, 169dpi screen density. Just in case. Thanks for taking care of it

chschtsch:
Ok, try this build:
[NewPipe-7.0.zip](https://github.com/theScrabi/NewPipe/files/70536/NewPipe-7.0.zip)
SHA-1 checksum: cf0780f9124ed8cc06038d68a863c0ad9cc98a83

iqqr:
Greg, you're a life saver. It fixed the whole thing without memory problem, everything works fine. Thank you mate

![working_newpipe](https://cloud.githubusercontent.com/assets/16397441/11974882/5a405dd2-a936-11e5-8781-5ca5b43ecc67.png)

chschtsch:
Hey, you're welcome!
[...]

PR #121, opened 2015-12-22
fix sw600dp-land layout problem (#120)

chschtsch:
Before:

After:

Before push, it would be nice to test it out first. (My computer doesn't have enough memory for tablet emulator)


theScrabi:
hm I'll try to test it.

PR #126, opened 2015-12-25
Fix view count parsing for large numbers

Adam77Root:
Just noticed this today when testing the experimental audio player with Gangnam Style.
The patch plays well with view_count field of AbstractVideoInfo class as it is already of type long.


theScrabi:
Ok thanks :)

Issue #101, opened 2015-11-19
Share video from Youtube app to Newpipe


nicozimmermann94:
Is there any way to view a video in newpipe coming from the stock youtube app? I want to watch videos from my subscription box. Maybe add an way to let the video "share" to newpipe?

medavox:
Off the top of my head, you could probably use the Youtube app to fire intents to watch videos, which NewPipe then handles. You'd have to keep from registering a default Youtube video handling app, though.

You can probably do this now. Just browse videos with the Youtube app, and when you tap on one to play it (as long as there isn't a default app set), you can select NewPipe from the Choose An App menu that pops up.

theScrabi:
Well i think the "share" idea is not that bad. People could watch videos via stock youtube app, and listen, or download them via NP.

razzintown:
So is there no way to get NewPipe into the "Share menu"?

theScrabi:
There is xD it just needs to be implemented :)

theScrabi:
https://github.com/theScrabi/NewPipe/issues/84

eighthave:
Its easy to get NewPipe in the *Share* menu (aka `ACTION_SEND`) the question is whether that is a good user experience.  The *Share* menu is supposed to be for sharing content as in messaging.  It is not supposed to be a way to hack around the normal ways of viewing.

theScrabi:
Hm I think if some request this feature and some reject it, we should vote for this.

nicozimmermann94:
@eighthave What about Dropbox? Most of the times it is in the share menu and I think it is not necessary messaging content.

eighthave:
Dropbox makes sense for Share/ACTION_SEND since you're sending content
to it, which it then stores.  NewPipe does not store anything or send
messages.  It shows videos.  And for that there is the ACTION_VIEW Intent.

ildar:
Very easy: Easy Open Link (Open URLs via select and share) - https://f-droid.org/app/de.audioattack.openlink

theScrabi:
Problem solved, or considered as workaround?

eighthave:
I consider this issue solved. Granted, the way Android handles the difference between Share/`ACTION_SEND` and view/`ACTION_VIEW` is not always great, but I don't think that's a problem that NewPipe can solve.

Issue #84, opened 2015-11-12
sharing from Firefox does not work


rgh36167:
When sharing youtube links from Firefox share menu NewPipe isn't offered as an option, apparently the intent pattern does not match.

Examined by IntentIntercept the intent looks like this:

```
intent:#Intent;action=android.intent.action.SEND;type=text/plain;launchFlags=0x10080001;B.org.mozilla.firefox.intent.extra.DEVIC

intent:#Intent;action=android.intent.action.SEND;type=text/plain;launchFlags=0x10080001;B.org.mozilla.firefox.intent.extra.DEVICES_ONLY=true;S.android.intent.extra.SUBJECT=ZDF%20-%20Islam%20-
%20Effects%20on%20Germany%20-%20YouTube;S.android.intent.extra.TEXT=https%3A%2F%2Fm.youtube.com%2F%23%2Fwatch%3Fv%3DKVWAIKoatWM;S.android.intent.extra.TITLE=ZDF%20-%20Islam%20-%20Effects%20on
%20Germany%20-%20YouTube;end

ACTION: android.intent.action.SEND
DATA: null
TYPE: text/plain

FLAGS:
FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET
FLAG_GRANT_READ_URI_PERMISSION
FLAG_RECEIVER_FOREGROUND

EXTRAS:
EXTRA 1:
Class: java.lang.Boolean
Key: org.mozilla.firefox.intent.extra.DEVICES_ONLY
Value: true
EXTRA 2:
Class: java.lang.String
Key: android.intent.extra.SUBJECT
Value: ZDF - Islam - Effects on Germany - YouTube
EXTRA 3:
Class: java.lang.String
Key: android.intent.extra.TEXT
Value: https://m.youtube.com/#/watch?v=KVWAIKoatWM
EXTRA 4:
Class: java.lang.String
Key: android.intent.extra.TITLE
Value: ZDF - Islam - Effects on Germany - YouTube


14 ACTIVITIES MATCH THIS INTENT:
Messaging (com.android.mms - com.android.mms.ui.ComposeMessageActivity)
Bluetooth (com.android.bluetooth - com.android.bluetooth.opp.BluetoothOppLauncherActivity)
Hangouts (com.google.android.talk - com.google.android.apps.hangouts.phone.ShareIntentActivity)
Translate (com.google.android.apps.translate - com.google.android.apps.translate.TranslateActivity)
[...]

PR #134, opened 2016-01-01
Intent filters and other fixes

eighthave:
This flushes out the Intent Filters to cleanly match all the youtube URL patterns I could find.  It also includes some minor fixes related to translations and changes the default download folder to use the Android "Movies" folder.  Lots more comments in each commit.


theScrabi:
Please don't switch the default Download folder to "Movies", NewPipe also downloads audio files, witch would mess up the idea of a "Movie" folder. One of the thinks I want to do next is to write a prober downloading part for NewPipe, therefore such changes are rather redundant.

eighthave:
How about making so video files go to Movies and audio files go to Music?  For me the central issue is to use the standard locations.

eighthave:
The other option is to put everything into *Download*

theScrabi:
I don't like download either. I think I'll be OK with your first suggestion.

eighthave:
also, I think this helps with #101 and #84 but not using the *Share* menu

theScrabi:
I'll test and merge your code in the next few days.

eighthave:
I think it probably makes sense to change the wording in `DownloadDialog` to match, so instead of *video* and *audio*, have it be *Movie* and *Music*.  That then provides a nice hint as to where the files will be downloaded.

medavox:
What about audio books? And how to videos?

theScrabi:
As I said id like to completely redactor downloading in the next few weeks. So all changes you do now might be obsolete.

theScrabi:
@eighthave I saw you forgot to mark your code with Copyright, and license. Id like to add gpl to your code.

eighthave:
of course, GPL is great!  Thanks for the merge, Tor one coming up next :)

medavox:
I just mean that labeling all audio as "music", and all video as "movies" isn't strictly accurate. No code needs writing to support audiobooks uploaded to YouTube. But labeling them as music may be misleading.

eighthave:
Think about the UX here.  I think if someone is downloading an audio
book, chances are they are going to play it using a music app.  And that
music app should use the default location for sound files:
/sdcard/Music.  I think the names of the folders is less important than
the user experience associated with them.

This is not a new or unique idea.  For example, OSX, Windows, and
GNU/Linux all have some standard version of these folders:

* ~/Documents
* ~/Downloads
* ~/Music
* ~/Pictures

Are there any dedicated audio book player apps?  Do they use the Music
folder?  Something like Audible is a closed system as far as I
understand it, it only plays files from its Amazon-based system. So it
is not a good example.

This seems to be one, but I just found it, I know nothing about it:
https://play.google.com/store/apps/details?id=ak.alizandro.smartaudiobookplayer&hl=en

chschtsch:
Look at iTunes' behavior.
I don't use iTunes anymore, but as far as I know, it keeps all the data in `/Music/iTunes/iTunes media/` folder, including audiobooks.
So, I think it's OK to just stick to movies and music folders. Think of them like movies is for any video files and music is for any audio.

theScrabi:
Also think about the technical difficulties, how should the app know what is music, and what is podcast. But still my opinion on this is, that we should generally divide NP content from the "default" content, by storing it inside its own folder.

theScrabi:
@eighthave who should be the copyright holder? Me or the guardien project?

medavox:
Everyone replying to my last comment above is correct; I just wanted to make sure that this was a conscious decision, rather than an oversight.

eighthave:
About the copyright, it doesn't matter to me.  I generally like lots of
copyright holders because it makes it harder to convince all of the
contributors to change to a non-free license.

PR #136, opened 2016-01-02
Tor support for all except streaming

eighthave:
I started with @gjedeer 's work and then worked on making Tor support more tightly integrated.  Since `DownloadManager` does not let you set the proxy settings, I wrote a basic, custom download manager for downloading over Tor.  This does not touch the video streaming at all, so with "Use Tor" enabled, video streaming will not go over Tor.  That'll require ExoPlayer, I think.

Ideally, NewPipe would represent the state of Tor in the UI somehow.  Orbot will broadcast out the status, and its easy to receive that with a `BroadcastReceiver`.  The open question is how to represent the state of Tor in the UX.  The states are `OFF, STARTING, ON, STOPPING`.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -17,7 +17,7 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
-    
+
     lintOptions {
         checkReleaseBuilds false
         // Or, if you prefer, you can continue to check for errors in release builds,
@@ -35,4 +35,5 @@ dependencies {
     compile 'com.android.support:recyclerview-v7:23.1.1'
     compile 'org.jsoup:jsoup:1.8.3'
     compile 'org.mozilla:rhino:1.7.7'
+    compile 'info.guardianproject.netcipher:netcipher:1.2'
 }
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -7,6 +7,7 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
     <application
+        android:name=".App"
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
         android:logo="@mipmap/ic_launcher"
--- a/app/src/main/java/org/schabi/newpipe/App.java
+++ b/app/src/main/java/org/schabi/newpipe/App.java
@@ -1,0 +1,32 @@
+package org.schabi.newpipe;
+
+import android.app.Application;
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
+
+import info.guardianproject.netcipher.NetCipher;
+import info.guardianproject.netcipher.proxy.OrbotHelper;
+
+public class App extends Application {
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        // if Orbot is installed, then default to using Tor, the user can still override
+        if (OrbotHelper.requestStartTor(this)) {
+            SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(this);
+            configureTor(prefs.getBoolean(getString(R.string.useTor), true));
+        }
+    }
+
+    /**
+     * Set the proxy settings based on whether Tor should be enabled or not.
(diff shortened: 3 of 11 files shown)


chschtsch:
Do you think, that Tor's icon in status bar won't be enough?

chschtsch:
We could make a subheader under Tor's checkbox in settings, that will represent Tor's state. This way user could check the status in settings.
Also, thanks for your work, I'm working on themes right now, and I also needed to create App class.

eighthave:
Thanks for the merge!  FYI, I've been using NewPipe forced over Tor, including streaming, for a bit now. It works well over Tor.  I'm using Orwall to do the forcing using `iptables` rules.

As for my last comment, I think I shouldn't have said UI, but only UX. Like @chschtsch said, the Orbot notification is enough visual representation. But one thing I haven't fully worked out is how apps should behave when Orbot is in the `OFF, STARTING` or `STOPPING` states.  I think the `STARTING` case is probably the easiest: the app should just do whatever it normally does to represent ongoing activity, like spinning circle or whatever.

What if the user turns off Orbot?  Then the NewPipe would get the `STOPPING` then `OFF` state.  Should NewPipe just try to start Tor again the next time the user does something that needs network access?  Or should it show some status that represents that it cannot reach the internet since Tor is off.  In any case, that's all just something to think about while working on the app :-)

theScrabi:
I think the last option should be the right one. Because I think you should not bypass the opinion of the user. It's his decision whether he starts tor again or not.

theScrabi:
@eighthave I've imported your branch, but your work is not very stable. If I have Orbot installed, but don't want to use it NewPipe fails, and deactivating tor is not working either, because then it just prints "Network Error".

I'd like to release, but until this is fixed I won't do it.

theScrabi:
So the next thing is to add exoplayer. I suggest we should add an ExoPlayVideoActivity, and not delete or override PlayerVideoActivity, in order to guaranty compatibility to older devices.

eighthave:
I totally agree that we should not bypass the opinion of the user, that's what makes this a tricky question.  There are lots of decisions of the user at play here, the user has decided to:

* install Orbot
* use NewPipe over Orbot (i.e. not disable "Use Tor")
* click on a new video or search, which requires internet

We used to have more decision points, like requiring users to also turn on "Use Tor" after installing Orbot.  But now, we've decided that most users who install Orbot want as much protection as possible, so the decision to install Orbot should be enough to also automatically set up apps to use Orbot.  Then users can always opt-out with the "Use Tor" preference for a customized setup.

So for this case, if we say that disabling Orbot should stop apps from having internet access, then disabling Orbot should really be represented as "turn off internet".  Android already provides standard ways to disable WiFi, mobile data, and the cellular/GSM radio, so that seems redundant.

In any case, it would probably be easiest to implement the last option, so perhaps that the best place to start.  NewPipe could just show a "SnackBar" like in this example that lets the user either open Orbot, or just try to start it: https://www.youtube.com/watch?v=28sltNHqdAs

n8fr8:
Great progress! I think there are three possible states:

- "Require Tor" always, whether Orbot is installed and running, and prompt user to install and/or autostart Tor if it is not running

- "Optional / If Available": Use Tor opportunistically if Orbot is installed and running, but don't do anything it if is not installed or not running

- "Never Use Tor": Don't use Tor, since I am using Orbot for something else entirely

I like the idea of using the Snackbar as a subtle confirmation prompt for the user. You can display a short message, and a confirmation action:

"You must install Orbot to use Tor. INSTALL?"

"You must start Orbot to use Tor. START?"

eighthave:
The "Optional / If Available" state worries me.  I think that makes it too easy to mess up and leak information.  People are using Orbot because they want to protect their privacy.

n8fr8:
Okay, what about then an OTR style solution

1) Require and autostart (Always)
2) Require but prompt to start or ask to disable/not use (Manual)
3) Never

eighthave:
I think *auto/never/prompt* makes sense for auto-starting Tor, but that really should be implemented in Orbot, not the apps. So the apps would always send the same start Intent, then if Orbot was set to *auto*, it would just start.  If set to *never*, it would ignore, then perhaps respond with a status Intent saying `STATUS_OFF`. And if set to *prompt*, it would show the main Orbot screen when it received the start Intent (i.e. the old interaction model for starting Orbot).

PR #140, opened 2016-01-04
Scrabi

chschtsch:


theScrabi:
hm what was that?
A description would be nice.

chschtsch:
Oh, I'm sorry. I just tried to merge the forks, but put it the wrong way.
Still ain't used to GitHub UI :)
Later I probably will open an actual pull request with implemented app theming.

theScrabi:
Aw ok :P

PR #162, opened 2016-01-29
to solve issue 38

iit2014086:
volume controls are applied on music and not phone volume


theScrabi:
I'm sorry I can't merge this code. It's not even compiling.

PR #176, opened 2016-02-14
Preference keys should not be based on localized names

pylerSM:


theScrabi:
Ok I agree, although I won't merge this.
All settings keys are listed in the settings_keys.xml file. This file is excepted from translating.
Also you just removed the @string/ prefix, but the values behind are not necessarily equivalent to the values they're referencing at. Therefore users would loose there settings after the next update.

theScrabi:
If you want to be precise you could at a translatable="false" attribute to the <resources> tag in settings_keys.xml .

PR #186, opened 2016-02-16
Added ExoPlayer support

faridk:


medavox:
Could you not have included ExoPlayer as a library, rather than bringing external source code into the tree?

faridk:
I already included ExoPlayer as a library in build.gradle file. The external source code is just a demo implementation of ExoPlayer which I took from here https://github.com/google/ExoPlayer/tree/master/demo

theScrabi:
Ok thanks faridk. I'll take a look at it. But is it possible you extend your code for some features?

theScrabi:
AAAA no.

theScrabi:
Thanks for the code, but there is still to much work to do. Also your version is not working anymore since youtube removed dashmpd. But I'll try and take your code as base.

faridk:
Are you sure that they did? Because I'm still able to play videos using ExoPlayer.

theScrabi:
Hm when used it the last time it didn't work. However even if it works again. I have to change a huge amount. If you like you kann take a look at the [addExoplayer](https://github.com/theScrabi/NewPipe/tree/addExoplayer) branch to see what I've changed so far.

PR #227, opened 2016-03-07
Restoring search query after orientation change

kamadi:
Restoring query text in SearchView after orientation change.


theScrabi:
All right nice work thanks :)

PR #238, opened 2016-03-12
Move YoutubeStreamPreviewInfoExtractor to separate class

rrooij:
The class should be separate because channel support in the future would require video preview extraction too, which doesn't differ much from the extraction done from the search results.


theScrabi:
Hm, ok :)

PR #248, opened 2016-03-23
Start-screen-hint

Ashanmaril:
Added a start screen hint telling the user to search. Excuse my noob-ness as I'm pretty new to Android development.


Ashanmaril:
I'm not sure if this is an okay way of doing this, but I think the app would benefit from some kind of start message, since every time I open it I think it had an issue booting up cause I'm just met with a blank screen. A few other people I've talked to had the same experience so I think this would be a nice addition.

theScrabi:
Al right, I see :) Thanks.

PR #249, opened 2016-03-25
Added new Snackbar for suggestion after searching.

CHRINIMUE:
I've added a new Snackbar for a suggestion after searching. So you can choose these suggestion after searching. But at the moment there is one problem with the updating form the search bar. So there is shown the old request.
I don't know how to fix it :-/ I hope someone can help me :D 

screenshot_2016-03-25-15-27-00
screenshot_2016-03-25-15-26-47

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/VideoItemListFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/VideoItemListFragment.java
@@ -6,6 +6,7 @@ import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.os.Handler;
 import android.preference.PreferenceManager;
+import android.support.design.widget.Snackbar;
 import android.support.v4.app.ListFragment;
 import android.util.Log;
 import android.view.LayoutInflater;
@@ -195,16 +196,29 @@ public class VideoItemListFragment extends ListFragment {
         this.streamingService = streamingService;
     }
 
-    private void updateListOnResult(SearchResult result, int requestId) {
-        if(requestId == currentRequestId) {
+    public void updateListOnResult(final SearchResult result, int requestId) {
+        if (requestId == currentRequestId) {
             setListShown(true);
             updateList(result.resultList);
-            if(!result.suggestion.isEmpty()) {
-                Toast.makeText(getActivity(),
-                        String.format(getString(R.string.did_you_mean), result.suggestion),
-                        Toast.LENGTH_LONG).show();
-            }
+            if (!result.suggestion.isEmpty())
+                Snackbar.make(getView(), String.format(getString(R.string.did_you_mean, result.suggestion)), Snackbar.LENGTH_LONG)
+                        .setAction(R.string.yes, new View.OnClickListener() {
+                            @Override
+                            public void onClick(View v) {
+                                query = result.suggestion;
+                                updatesearchQuery(query);
+                                Toast.makeText(getActivity(), R.string.search + ": " + String.format(query), Toast.LENGTH_LONG).show();
+                            }
+                        })
+                        .show();
         }
+    }
+
+    public void updatesearchQuery(String query) {
+        search(query);
+        //VideoItemListActivity VIL = new VideoItemListActivity();
+        //this.searchView = searchView;
+        // searchView.setQuery(query,true)                          <-- This don't work
     }
 
     private void updateList(List<StreamPreviewInfo> list) {
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -23,6 +23,7 @@
     <string name="settings_activity_title">Settings</string>
     <string name="use_external_video_player_title">Use external video player</string>
     <string name="use_external_audio_player_title">Use external audio player</string>
+    <string name="yes">Yes</string>
 
     <string name="download_path_title">Download path video</string>
     <string name="download_path_summary">Path to store downloaded videos in.</string>


theScrabi:
ah nice

app/src/main/java/org/schabi/newpipe/VideoItemListFragment.java
```diff
@@ -195,18 +196,31 @@ public void setStreamingService(StreamingService streamingService) {
...
+                                query = result.suggestion;
+                                updatesearchQuery(query);
+                                Toast.makeText(getActivity(), R.string.search + ": " + String.format(query), Toast.LENGTH_LONG).show();
+                            }
+                        })
+                        .show();
         }
     }
 
+    public void updatesearchQuery(String query) {
```
theScrabi: Please write proper camel case, meaning write `updateSearchQuery()` instead of `updatesearchQuery`.

theScrabi:
I see one of your problems, you are working inside the fragment. But the logic for the tabled search view, and the action bar search view are both covered in the code of the activity. You will find it [here](https://github.com/theScrabi/NewPipe/blob/master/app/src/main/java/org/schabi/newpipe/VideoItemListActivity.java#L78).

Also please change the color of the YES text. Like this [example here](https://github.com/theScrabi/NewPipe/blob/master/app/src/main/java/org/schabi/newpipe/ErrorActivity.java#L110)

CHRINIMUE:
OK. Thank you for the tips. I will try it again.

CHRINIMUE:
Sorry, I don't know what I have to do. 
I changed this:
VideoItemListActivity:
```
public void updateSearchQuery(String query){
        //SearchVideoQueryListener SVQL;
        //SVQL = new SVQL(query);
        //SVQL.onQueryTextChange(query);
        //SuggestionSearchRunnable SuggestionSearchRunnable;
        //new SuggestionSearchRunnable(1, query);

        //SearchSuggestionListener SSL;
        //SSL = new SSL(1);
        //SSL.onSuggestionClick(1);
        
        //SearchView searchView;
        //searchView view = new searchView
        //searchView = searchView;
        //new searchView(SearchView searchView);
        //searchView.setOnQueryTextListener(new SearchVideoQueryListener());

        //searchQuery = query;
        //searchView.setQuery(query,true);
        //listFragment.search(query);
        //listFragment.startSearch(query, 1);
    }

```
I tied everything in the comments.
And the VideoItemListFragment:
```
public void updateListOnResult(final SearchResult result, int requestId) {
        if (requestId == currentRequestId) {
            setListShown(true);
            updateList(result.resultList);
            if (!result.suggestion.isEmpty()) {
            VideoItemListActivity VILA = new VideoItemListActivity();
                Snackbar.make(getView(),String.format(getString(R.string.did_you_mean, result.suggestion)), Snackbar.LENGTH_LONG)
                        .setActionTextColor(Color.YELLOW)
                        .setAction(R.string.yes, new View.OnClickListener() {
                            @Override
                            public void onClick(View v) {
                                query = result.suggestion;
                                search(query);
                                VideoItemListActivity VILA = new VideoItemListActivity();
                                VILA.updateSearchQuery(query);
                            }
                        })
                        .show();
            }
        }
    }
```

CHRINIMUE:
OK, I tried it again. But which Method is used to update the text in the search bar. I tryed the SearchVideoQueryListener and SearchView updateQuery. But last don't work. I realy don't know what I have to do. :-/

(Sorry for my bad Java skills, but I am trying to learn and improve my skills :D )

theScrabi:
> (Sorry for my bad Java skills, but I am trying to learn and improve my skills :D )
Well than you should be double that careful what code you are contributing!!!

theScrabi:
Ok now, I gave you the wrong answer.
You are right, the response is proceed inside the fragment, but all the search views are handled from the activity. Now we have a problem, because the fragment has to inform the activity that there is a suggestion.  So we need a callback.

In the fragment I suggest to write something like this:

    public OnResponseSuggestionListene onResponseSuggestionListener= null; 
    //implementation of the interface

    public interface OnResponseSuggestionListener {  //interface for the callback
        void onSuggestionReceived(String suggestion);
    }

    public void setOnResponseSuggestionListener(OnResponseSuggestionListener listener) {
       //through this methode the activity is able to implement the listener.
       onResponseSuggestionListener = listener;
    }

Now replace that [Toast](https://github.com/theScrabi/NewPipe/blob/master/app/src/main/java/org/schabi/newpipe/VideoItemListFragment.java#L203) with your snack bar, and when someone hits yes you call `onResponseSuggestionListener.onSuggestionReceived(result.suggestion);`.

On the Activity side in the onCreat() method somewhere after the fragment was created you write:

    listFragment.setOnResponseSuggestionListener(new OnResponseSuggestionListener {
        void onSuggestionReceived(String suggestion) {
            searchView.setQuery(suggestion,true);  //this will put the suggestion into the search bar and start a new search request
        }
    });

theScrabi:
also the crawler might have to be fixed, since it not always outputs a correct suggestion response.

theScrabi:
@CHRINIMUE still working on it?

CHRINIMUE:
Sorry, at the moment I will improve my knowledge about Java and Android Application programing, because I don't want to add critical code.
And I don't want to go on your nerves. 
But in a few weeks I will try it again ...
So I will close this pull request ...

theScrabi:
You are not nerving me, im happy about every contribution. But trying to get better bevore contributing is also a novel step. God luck :)

PR #261, opened 2016-04-06
Let's tweak the UI strings a bit

zhaofengli:
The following 4 commits fix common mistakes in the UI strings:

### `98afe79` strings.xml: Remove unnecessary punctuations
Periods, colons and other unnecessary punctuations should not be used in labels.
See https://www.google.com/design/spec/style/writing.html#writing-capitalization-punctuation

### `2a6e7f3` search_language_title: Change "Preferable" to "Preferred"
Why? See http://english.stackexchange.com/questions/128996/how-would-one-know-when-to-choose-preferred-or-preferable

### `8060c6a` strings.xml: Fix grammar
"Download path video" -> "Video download path"
"Download path audio" -> "Audio download path"

### `a5fc6db` default_resolution_title: Fix capitalization
"Default Resolution" -> "Default resolution"

See https://www.google.com/design/spec/style/writing.html#writing-capitalization-punctuation

#### OCD...
It has been fixed!

### DIFF
--- a/app/src/main/res/values-ar/strings.xml
+++ b/app/src/main/res/values-ar/strings.xml
@@ -25,7 +25,7 @@
     <string name="background_player_name">مشغل NewPipe في الخلفية</string>
     <string name="background_player_playing_toast">جاري التشغيل في الخلفية</string>
     <string name="cancel">إلغاء</string>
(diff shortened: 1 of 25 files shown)


theScrabi:
Ok sounds legit.

PR #314, opened 2016-06-15
File picker for download preferences.

satiricon:
I still have to make the latest filepicker version work but meanwhile it's working pretty well with an older version.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,5 +29,5 @@ This document contains guidelines on making contributions to NewPipe.
 
 ## Communication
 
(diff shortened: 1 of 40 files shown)


satiricon:
I don't know why the test would not pass. I tested on my machine and they passed flawlessly.

theScrabi:
Yea it's just that view count thing. This never worked correctly. The rest is fine though.

... need to fix that test one day.

theScrabi:
File picker does not seem to change download directory.

theScrabi:
I noticed, if you call filePickerActivity, it will return immediately giving a Nullpointer as result.

satiricon:
@theScrabi I'm checking this out.

PR #327, opened 2016-06-30
fix build. remove wrong xliff namespaces from spanish translation.

adelolmo:
This pull request should get the build back to normal.


theScrabi:
Thanks :)

adelolmo:
About the test failing (YoutubeSearchEngineTest.testIfNoErrorOccur): to fix it you have to change the search term. If you use "cats" instead of "star wars" the test will pass.

adelolmo:
I finally use the term "this is something boring" for the search in YoutubeSearchEngineTest.
The key is to avoid search which have any ads or channels. Plus the results must have more than 1000 views in the first result page (not 100% sure here).

theScrabi:
Yea I already thought about lowering the minimum views, since that sometimes fails. I just want to test if the numbers are (at least a bit) correct.

May have to change that.

PR #316, opened 2016-06-16
Added rewind capability to notification control

jaytj95:
added a button to notification panes (regular & expanded) that allows the user to start the audio over (`seekTo(0)`)

PR #337, opened 2016-07-30
Add fastforward button to notification

Gracecr:
Expanding off of #316, This adds a fast forward button to the notification pane that skips ahead 15 seconds. This is the same behavior as in the video player.


timofonic:
Why 15 seconds instead more granularity?

Gracecr:
I did it this way because this matches the behavior of the built in player.

Gracecr:
Merged conflicts. Sorry about the two reverted commits.

theScrabi:
We are already working on something similar: https://github.com/TeamNewPipe/NewPipe/pull/377

PR #339, opened 2016-08-02
Set IMMERSIVE_STICKY for Exoplayer

### LINKED ISSUE #253: Exoplayer : the top bar doesn't disappear
Exoplayer work great but something that can be improved is the top bar that doesn't disappear whereas it should.

Gracecr:
Fixes #253. Mostly taken from [the docs](https://developer.android.com/training/system-ui/immersive.html#sticky).

Interestingly, the docs recommend the use of IMMERSIVE for video playback while the actual YouTube app uses IMMERSIVE_STICKY. Personally, I think that IMMERSIVE_STICKY is more functional and is what people will be used to, so I went with that.


theScrabi:
Sorry but explayer is getting a new ui anyway.

PR #350, opened 2016-08-21
Added Splash Screen to bypass blank screen on start up.

KeizerDev:
This will just bypass the blank screen and "physically" removes startup time.


KeizerDev:
Ps. I would like to join the slack channel of this project :smiley:

PR #359, opened 2016-09-28
Launch video player in single task mode

Nutomic:
This makes sure VideoItemDetailActivity is always started as a seperate app, and is displayed in recent apps seperately.

Changes in VideoItemDetailActivity are necessary to replace the existing video/fragment with the new one.

The change in ActionBarHandler was required to avoid a crash when opening a new video from a different app.


theScrabi:
Merged. Will come with my next push.

Nutomic:
Thanks :)

PR #364, opened 2016-10-03
Added feature to open NewPipefrom the share menu of YT.

SUguy:
NewPipe can now open videos from the share menu of YT. This is the most convenient and fastest way. Shouldn't this be included in the main app?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'com.android.application'
 
 android {
     compileSdkVersion 24
-    buildToolsVersion '23.0.3'
+    buildToolsVersion '24.0.3'
 
     defaultConfig {
         applicationId "org.schabi.newpipe"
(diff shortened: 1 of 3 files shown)


timofonic:
I'm just an user. What do you mean? YT app or site? I fail to understand it.

theScrabi:
He means to put the feature of this [app](https://github.com/TeamNewPipe/ShareWithNewPipe) into main NewPipe app.

SUguy:
Yeap... Shouldn't it be included in the main app....?

teod0r:
I think that it should be included in the app.
For those who want to just download, there should be an option in settings with 3 options:
* always download
* always play video
* always ask if  download or play

SUguy:
@theScrabi I back @teod0r. Can you please conduct a **poll** to collect user opinion.

theScrabi:
All right now. You want it you get it, but I did it my self though :)

PR #376, opened 2016-11-06
auto detect downloaded files as new media files

fr3ts0n:
add media scanner notification after successful download
* added DownloadListener
* on finished download send broadcast intent to media scanner


theScrabi:
I like that commit :) Although I might have to find out why CI is failing again.

theScrabi:
you forgot to add a copyright notice to the DownloadNotice

PR #377, opened 2016-11-10
Add basic support of playlist

BlenderViking:
Hi,

I made the playlist support for your greatest app 👍 

You can made a local playlist and add item for the search :

[](https://cloud.githubusercontent.com/assets/5410525/20182780/413286c2-a763-11e6-9bc0-06e90dd424f3.png)

If you tap on the more button these contextual menu is show :

[](https://cloud.githubusercontent.com/assets/5410525/20182977/efa3bdde-a763-11e6-87d4-88325b40af5f.png) [](https://cloud.githubusercontent.com/assets/5410525/20182980/f2188cc0-a763-11e6-801c-5d29be15aca4.png)
playlist 55
[](https://cloud.githubusercontent.com/assets/5410525/20183107/89834e42-a764-11e6-8eb2-7454f9cba101.png)

You can access to playlist screen with the contextual menu : 

[](https://cloud.githubusercontent.com/assets/5410525/20183227/ec252016-a764-11e6-9ab8-1967bfc5c708.png) [](https://cloud.githubusercontent.com/assets/5410525/20183256/02fd5380-a765-11e6-9b1e-7c99a29ce23a.png)
[](https://cloud.githubusercontent.com/assets/5410525/20183446/a0e7073a-a765-11e6-9ece-75bd3ccb7ccf.png)

An auto-play option mode for local playlist is now available : 
[](https://cloud.githubusercontent.com/assets/5410525/20183300/1fbcb51a-a765-11e6-93c9-75ef1c8dd3b2.png) [](https://cloud.githubusercontent.com/assets/5410525/20183381/62be9bd0-a765-11e6-8a9e-39dfe32ce500.png)

Retrieve a playlist directly from youtube is now supported : 

[](https://cloud.githubusercontent.com/assets/5410525/20183409/80781890-a765-11e6-9bb5-8b435943f6f5.png) [](https://cloud.githubusercontent.com/assets/5410525/20220009/29d78a1c-a82c-11e6-926c-cfb5619eaa6b.png)

It's possible to convert to local playlist a youtube playlist, but the content is limited to local loaded items

[](https://cloud.githubusercontent.com/assets/5410525/20220018/31080db6-a82c-11e6-9212-74fec7d4de2a.png)

The local playlist is store on sqlite database : 

I have added [stetho](https://facebook.github.io/stetho/) on project for help to debug sqlite database

With chrome a select my device :

sqlite

And i can query directly my device database :

table_schema

It's missing :

- Allow youtube playlist to autoplay audio or video

PR #382, opened 2016-11-25
Add basic support of playlist #377 (clone)

BlenderViking:
StreamInfoItem:

- [x] Fix StreamInfoItem Layout (margin the three dots so they are in the right location)
- [x] make the StreamInfoItem button cover the whole Item not just the top (like it was before)
- [x] Remove the share and play option form the StreamInfoItem context menu
- [x] Make the StreamInfoItem context menu a local drop down menu (like the once of the download items)

stream-preview-info 2

stream-preview-info 3
stream-preview-info 1

Extractor:

- [ ] Make the extractor part of the NewPipe extractor
- [ ] Make the extractor fit the NewPipe architecture
- [ ] Write unit tests for the extractor (can only be done before the first two parts are done)
ChannelActivity:

 Create a new layout for play list and queuing
- [x] Use the original layout for ChannelActivity again

Queuing- PlayListActivity:

- [x]  Allow drag and drop for manage item position (persist on sqlite database)
- [x] Allow Swipe for remove item on playlist
swipe
- [x] Stick to android style guidelines (only one floatingbutton each activity)
Solved by using a contextual menu for provide user action choice (like menu provide by hangout) using this lib : https://github.com/futuresimple/android-floating-action-button recommended by these stackoverflow thread: https://stackoverflow.com/questions/30699302/android-design-support-library-fab-menu
menu-local-playlist
external-playlist

Notification:

- [x] Make the rewind button, and the "last video button" be the same
(if the music runs for more than 2 second rewind, if it runs longer go back to the last video)
- [x] Make, that the notification wont crash if there is no more video in the queue and someone hits the next/last button

I didn't find how to replace my previous [pull request](https://github.com/TeamNewPipe/NewPipe/pull/377), so  I created a new request :-/


theScrabi:
Yep we are getting closer :) NNNNNNNNICE work

BlenderViking:
:tada:

Issue #386, opened 2016-12-04
Network Error


be-neth:
Hello,

I cannot use Newpipe when using my VPN on Android 7.
From logcat when searching "test" video I have the following exception :

~~~~
12-04 11:07:43.719 20419 23161 W System.err: java.io.IOException: java.io.FileNotFoundException: https://www.youtube.com/results?search_query=test&page=0&filters=video
12-04 11:07:43.720 20419 23161 W System.err:    at org.schabi.newpipe.Downloader.dl(Downloader.java:88)
12-04 11:07:43.721 20419 23161 W System.err:    at org.schabi.newpipe.Downloader.download(Downloader.java:65)
12-04 11:07:43.721 20419 23161 W System.err:    at org.schabi.newpipe.Downloader.download(Downloader.java:47)
12-04 11:07:43.721 20419 23161 W System.err:    at org.schabi.newpipe.extractor.services.youtube.YoutubeSearchEngine.search(YoutubeSearchEngine.java:88)
12-04 11:07:43.721 20419 23161 W System.err:    at org.schabi.newpipe.extractor.SearchResult.getSearchResult(SearchResult.java:32)
12-04 11:07:43.721 20419 23161 W System.err:    at org.schabi.newpipe.search_fragment.SearchWorker$SearchRunnable.run(SearchWorker.java:103)
12-04 11:07:43.721 20419 23161 W System.err:    at java.lang.Thread.run(Thread.java:761)
12-04 11:07:43.722 20419 23161 W System.err: Caused by: java.io.FileNotFoundException: https://www.youtube.com/results?search_query=test&page=0&filters=video
12-04 11:07:43.722 20419 23161 W System.err:    at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getInputStream(HttpURLConnectionImpl.java:250)
12-04 11:07:43.722 20419 23161 W System.err:    at com.android.okhttp.internal.huc.DelegatingHttpsURLConnection.getInputStream(DelegatingHttpsURLConnection.java:210)
12-04 11:07:43.722 20419 23161 W System.err:    at com.android.okhttp.internal.huc.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java)
12-04 11:07:43.722 20419 23161 W System.err:    at org.schabi.newpipe.Downloader.dl(Downloader.java:78)
12-04 11:07:43.722 20419 23161 W System.err:    ... 6 more
~~~~

[...]

PR #387, opened 2016-12-07
reCaptcha

be-neth:
The following patches try to implement reCaptcha from Youtube in a simple manner : 

If HTTP 429 is returned by Downloader : reCaptchaException is thrown and reCatpchaAcitivity is started.
The Activity is a simple WebView (with Javascript enabled) displaying reCaptcha.

Ref #386 .

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -153,6 +153,9 @@
             android:label="@string/title_activity_channel"
             android:theme="@style/AppTheme.NoActionBar" />
 
+        <activity
+            android:name=".ReCaptchaActivity"
+            android:label="@string/reCaptchaActivity" />
     </application>
 
 </manifest>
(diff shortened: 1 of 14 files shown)


theScrabi:
I was not able to test it correctly since I never had the case where I had to enter a captcha. Can you enter some screenshots, to see how it would look like :)

be-neth:
Hi,

Sure :

1. After entering something in the search Bar
screenshot_20161225-095036

2. Captcha display in webview
screenshot_20161225-095044

3. Resolving Captcha
screenshot_20161225-095110

4. After Submit
screenshot_20161225-095201

But there is still two issues with this pull request that i need to fix :
1. After exiting the reCaptcha activity, I need to re-click on the search field to refresh the query.
2. If I click from outside Newpipe Application on a Youtube link the reCaptcha exception is thrown but the activity is not started.

Do you plan to work on NewPipe at the 33c3 ? I will be there too :)

theScrabi:
Well why not :) Meet me at the NewPipe talk, so we can talk about that :)

Issue #388, opened 2016-12-13
[BUG] Incorrect list of videos


teod0r:
First 20 results are displayed twise, they are dubblicated

BlenderViking:
Hi,

I don't have this problem. 

Can you provide some screenshot, device and network capacities ?

Regards,

teod0r:
Unfortunatly I can't provide screenshots because my devise does not provide the ability to take screenshots. I also tryed apps from markets and they didn't work too. I don't know if it is aproblem of ROM or device.

What do you mean by device and network capacities?
I use samsung galaxy tab 4 10.1 with cyanogenmod 11 (android 4.4.4)

theScrabi:
I can confirm that bug, I noticed that to.

coffeemakr:
This bug should be resolved [Version 0.8.6](https://github.com/TeamNewPipe/NewPipe/releases/tag/v0.8.6)

PR #390, opened 2016-12-19
Fix dublicated results

coffeemakr:
Fix for  #388:
Google begins with page 1 so we can add 1 to the page before creating the URL.
I've tested it on my Phone and it worked.

Examples:
https://www.youtube.com/results?search_query=test&page=0&filters=video
is the same as 
https://www.youtube.com/results?search_query=test&page=1&filters=video
but
https://www.youtube.com/results?search_query=test&page=2&filters=video
is different

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ android:
     - build-tools-23.0.3
 
     # The SDK version used to compile NewPipe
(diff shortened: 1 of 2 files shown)


theScrabi:
Nice nice thanks :)

PR #396, opened 2016-12-27
Feature: Background progress bar

coffeemakr:
Hi
I've added a progress bar to the custom expanded notification view. And change the following things:

- Add progress bar to the expanded layout
- Create a subclass of NotificationBuilder to update the progress, title, artist and playing state more easily 
- Add the Parcelable "PlaybackState" which contains information about the current playback state 
- Add action "PLAYBACK_STATE" which is broadcasted by PlayerThread
- Uncomment the while loop at the end of PlayerThread.run() to update the playback state 
- Update notification on a received broadcast message in the broadcast receiver
- Fix some typos 

I think the broadcast message could also be used by activities to show controls when a stream is played in the background.

I hope you find this useful
Regards

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /.idea
 /*.iml
 gradle.properties
+*~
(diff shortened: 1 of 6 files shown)


theScrabi:
All right. The seekbar goes to the edge of the notification.
Can you do a margin on the right site, so the margin on the right and left site are the same.

coffeemakr:
Done so. The preview image has a margin of 8dp on the right side but the close button has 5dp around.
I've chosen the 8dp.

PR #399, opened 2016-12-31
Added dark theme

ksyko:
Some icons not visible (thumbs and RSS)
App needs to be restarted manually.


theScrabi:
Why did you close it again?

PR #404, opened 2017-01-03
Improve reCaptcha integration

be-neth:
This pull request fixes the following issues from first reCaptcha integration ( #387 ).

> After exiting the reCaptcha activity, I need to re-click on the search field to refresh the query.

Fixed.

> If I click from outside Newpipe Application on a Youtube link the reCaptcha exception is thrown but the activity is not started.

Fixed.

I change the way of starting reCaptchaActivity, using startActivityForResult() and onActivityResult().


theScrabi:
all right thanks

PR #415, opened 2017-01-16
Keep search query

coffeemakr:
Hi I've noticed that the search results (and the query) disappear when the screen is rotated so I made a fix.

 * Keep search query when fragment is restored
 * Simplify SuggestionListAdapter
    * Use ResourceCursorAdapter for view creation
    * Remove ViewHolder class
 * Fix deprecation warning
 * Some clean code

Regards


theScrabi:
Nice work :) ... as always ;D

coffeemakr:
I'm glad you like it!

Commit  8e451b2 doesn't really belong to search query but i figured I could append it here. Otherwise I can open a new pull request.

PR #416, opened 2017-01-16
Fix for  video not playable

coffeemakr:
Seems to me like a quick and dirty fix but it's working for me.


theScrabi:
O oh boy relay. I should program more often :) thanks.

coffeemakr:
Really? strage.

coffeemakr:
Ah I saw the error when clicking on videos not on the channel first. It may seem silly but you could try onResume which worked for me.

Issue #407, opened 2017-01-07
Can't play videos since last update


OffifialMITX:
After updating to the latest version (0.8.7) I can't watch videos if I:

1. Do any search
2. Click on any link
3. Go to the profile
4. Click on any link
5. Tap on the big video thumbmail

If I do it nothing happens after 5.

theScrabi:
Noticed that, It's a nasty bug :/

yenon:
It does work after forcing screen rotation, maybe a listener is not being set? I'm on Android 7.1.1, if that helps.

swarnk:
Same thing. I can't watch next or suggested videos. Big play button do nothing...

theScrabi:
Tasty bug :/

theScrabi:
Ok I'm debugging it.

OffifialMITX:
I've installed 0.8.7, no problems at all.

theScrabi:
I see. I tried to diff master with 0.8.6 but I can't find anything. Even debugging does not help.

OffifialMITX:
I found something that could be useful, it looks like an invisible element appears in front of the button. Also for some reason the standard resolution is being ignored.
<table><tr><td><img alt="Screenshot 1" width="300" src="https://cloud.githubusercontent.com/assets/22666602/21981619/9d2d1f5a-dbe8-11e6-8946-96b3e56542ef.png"></td><td><img alt="Screenshot 2" width="300" src="https://cloud.githubusercontent.com/assets/22666602/21981697/efeddd1a-dbe8-11e6-963d-ec9d3e17d442.png"></td></tr></table>

OffifialMITX:
I also found out that if you tap on the top part of the thumbmail (above the play button) it works.

swarnk:
Confirm this. Tap above the play button is works for me too.

coffeemakr:
For me it works when renaming onActivityCreated to onStart. The activity already exists so this method is not called.

OffifialMITX:
This invisible "barrier" disappears on orientation change.

theScrabi:
aha thats it O.o
Now that's ugly. Ok now basicly there is a transparent button laying on top of the thumbnail, and in the image on the right side that button was resized, on the left site it was not resized properly.

theScrabi:
Should work now.

PR #426, opened 2017-01-25
Remove thumbnail button

coffeemakr:
This is related to  #407. As you know the button seems not to adapt to the size of the thumbnail image view.

I thought it would be easier to just remove the "thumbnail background button"  and I added the onClickListener to the thumbnail. It seems to work fine and I have no problem playing the video.

I have to admit I'm not exactly sure what the OnLayoutChangeListener does and why it's only for versions above 18 (newer devices). If you like an can leave this part as it was.

I've tested it on Api Version 25. Let me know if this works for you.


theScrabi:
Hm we can't remove the < sdk18 button since oder devices would not be supported.

PR #467, opened 2017-02-27
Black theme and code improvement

ksyko:
Removed ThemeableActivity to set theme. Instead using static method to set theme (util/ThemeHelper).


KeizerDev:
Please pull from `TeamNewPipe:master`

theScrabi:
Seems like we've worked on channel activity at the same time :P
merged anyway.

PR #474, opened 2017-03-09
Add basic popup mode

mauriciocolli:
Hey,

I implemented a popup mode, and this is how it works:

- Add ExoMedia library
    - Has similar API of VideoView
    - Is based on ExoPlayer
    - Handle almost everything automatically
        - It evens plays the `.m3u8` (HLS), that are parsed by the youtube-dl, that means we can implement livestream!

- Had to re-implement the ExoPlayerActivity, as I add the ExoMedia library
exoplayer

- Add a button in the menu, that opens the popup with the selected resolution
openinmenu

- Notification control with thumbnail, here the user closes the popup
notif

- Add an option to go directly to popup, this opens with the 
default resolution selected on the `Settings`
openwith

- When opening a video and already have one open, the player just sets the video and start to play (without closing)
alreadyopen

- By  single-taping the player  -> pause/resume
    - By double-tapping the left side (highlighted), the player do a rewind,
        if it's in the right, a fast forward
player

- Demonstration on external apps
externalapps

- Demonstration with NewPipe visible
device

If anyone want to test, here's the apk:
[app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/830281/app-debug.apk.zip)

This implementation is basic yet, it doesn't have resize for example, but i'm look for implement it in the future.
Any suggestion is very welcome.


ksyko:
Another great addition.

theScrabi:
Yea I don't know why github does not understand that I merged it by hand, but trust me your code is part of 0.8.12 now.

mauriciocolli:
What do you guys think of the idea of adding a `Fullscreen` and a `Quality Selector` button?

ideacrop

Is it too clumsy? Too much things? Position?
Tell me your opinion

alpha-k:
@mauriciocolli that design looks really good, no more accidentally going to end of video trying to fullscreen the video, and the resolution option is a nice touch! And on 1080p screens It will not look as cluttered, I love it! 

Also, thank you so much for taking time to implement this feature, along with downloading this makes NewPipe a must-have app!

mauriciocolli:
@alpha-k I guess adding a option in the `Settings` to disable the top controls, for those don't want to have it, till I make the popup resizable, is the solution.

alpha-k:
Sounds good, allows for people who want it to use it and others can turn it off.

Regarding the resizing, just a quick suggestion, I used to use an ios tweak called VideoPane, which was really good in snapping to the edges of the screen or to the center of the screen, and resized with pinch to zoom. Would that maybe be a good way to implement resizing for NewPipe? Here's a demo of the tweak https://www.youtube.com/watch?v=2SY-3k42lxM&t=32

theScrabi:
@mauriciocolli 
- Resizing the popup window would be a nice thing
- If its resizable than I think the fullscreen button was not to clumpsi
- Please :'| can you make the UI of the pouplayer the UI of the fullscreen player
- also if you are interested, we still need exoplayer as our default player

mauriciocolli:
Regarding the ExoPlayer, the ExoMedia already uses it (check [this](https://github.com/brianwernick/ExoMedia#exomedia)), it just handles for us, with it we can play almost anything like: 
- m3u8
- mpd
- Normal videos and Live streams

Check more in [this](https://github.com/brianwernick/ExoMedia/blob/21c025ecfc2e6d4256e9e23647527225f6d61b41/library/src/main/java/com/devbrackets/android/exomedia/type/MediaSourceType.java) link, and let's ditch the `videoView` implementation, just too slow and don't work with these formats

And about the FullScreen player, because some problems, I think it's (much) better to open the actual player (in fullscreen), when entering this mode, it will be something **like** this:

fullscreen2

And yes, I'll try to change the fullscreen layout to match the popup's one.
The position of the popup I want to remember too, but I'll implement it in the future.
What do you guys @alpha-k, @theScrabi think? Is it seamless enough?

alpha-k:
Seems like there will be a slight lag in switching, as the video needs to pause and resume again in the full player.. Is there a way to seamlessly continue the playback of the video without pausing? Even though it's only 2 seconds, that little bit of lag can be distracting.. 

Alternatively if the resizing is implemented, maybe you have a auto-resize function that gets called when pressing full screen, that rotates the video panel by 90 degrees and panel height is set to display width, filling it in landscape mode.. But the down side is it won't go to the main player.. although i'm guessing you tried this and ran into some issues as you mentioned..

PR #488, opened 2017-03-27
Implement screen orientation toggle

mauriciocolli:
- Add `Screen Orientation Toggle` in fullscreen player
o


theScrabi:
Ok nice :)

By the way I encountered a small bug. The screen does not stay active while playing. It turns of after some time.

mauriciocolli:
Fixed! :D

PR #492, opened 2017-04-01
Fix next video and refactor

mauriciocolli:
- Fix next video
- Fix margins and dimensions
- Remove redundant styles
- Nicer animation/transitions

As the `Universal Image Loader` is discontinued a [long](https://github.com/nostra13/Android-Universal-Image-Loader#project-news) time ago, I'm planning to switching it to more updated libraries, like _Fresco, Picasso, Glide._

PS: I changed the `VideoItemDetailFragment` to `activity` for now,  **_maybe_** I'll try to implement it as a fragment in the `MainActivity` (as the `SearchFragment` is already).

PS 2: @theScrabi remember to remove [this](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/stream_info/StreamInfo.java#L243) `if` on the NewPipeExtractor


theScrabi:
So you pulled out all the code from the fragment and put it oin the activity?

mauriciocolli:
As I said in my first post, I plan to integrate this to the `MainActivity`, and this it's _**a temporary fix**_ that was **simpler** to implement (the fragment would need to be refactored anyway).

But of course, it'll take time, so it's up to you decide if we have the bug until then (or fix yourself), or use this as a "**workaround**" (remember that is not just the fragment, it have other things (styles, dimensions, animations, etc.),  that'll make the switch  later on easier).

theScrabi:
Hm ... Yes. However, the thing I don't understand is why should we need to get rid of the fragment now, when we need to have it again later?

PR #499, opened 2017-04-09
Migrate to fragments and improvements

### LINKED ISSUE #487: Missing first "related video" in chronological series

I feel like there might be an off by one in the extractor: when watching a video that is part of a series with numbered episodes, Youtube always makes sure that the next episode chronologically is the first element in the suggested videos.

However in NewPipe, the next episode not only isn't in the first position but isn't in the suggested list at all.

This might be an off by one, but it could also that the extractor doesn't parse the "Up next" box that is before the list of suggested videos on each Youtube page.

mauriciocolli:
- Migrate to fragments
- Fix #487
- Don't show `"Open in popup mode"` to channel links
- New backstack of videos
- Change the subscribers count to format using `NumberFormat`, for example some locales use `.`  others `,` this handles it automatically (the old method had a bug for leading zero, e.g. 4.82.125 instead of 4.082.125)
- Add string `'subscribers'` for channels with more than 1 subscriber (plural)
- Popup player chooses the default format and resolution based on the new preference (format)
- Fix `taskaffinity` of the router activities
- Show title before loading, as it is available from the items already loaded


mauriciocolli:
@theScrabi Hey, I see that you forgot to remove [this](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/stream_info/StreamInfo.java#L243) `If`  in the NewPipeExtractor

theScrabi:
Holy crap I have the suspicion that after you are done I'm not familiar with my software anymore xD.
However I like that you clean up my crappy work :+1:

PR #500, opened 2017-04-12
Add resolution support up to 4k and 60 fps

### LINKED ISSUE #291: Add support for 1080p
As the title says, support for Full-HD and possibly more would be really nice. Thanks

mauriciocolli:
- Up to 4k with 60 fps
    - Not every device can play in that resolution and bitrate
    - Add option to hide the high resolutions greater than 1080p (2k, 4k) for not clutter the menus, for those who don't need it
- Add a preference to set the default resolution of the popup, will be used when opening in popup mode from another app
- Resolves #291 

Related pull request (for the core part) > https://github.com/TeamNewPipe/NewPipeExtractor/pull/3

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/channel/ChannelFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/channel/ChannelFragment.java
@@ -413,4 +413,9 @@ public class ChannelFragment extends Fragment implements ChannelExtractorWorker.
     public void onError(int messageId) {
         Toast.makeText(activity, messageId, Toast.LENGTH_LONG).show();
     }
+
+    @Override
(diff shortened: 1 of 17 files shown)


theScrabi:
Nice work, the only thing you might have to do is disable the support for downloading videos higher than 1080p, because they would be downloaded without sound.

doodhout:
I suppose this PR is included in the 0.9.3 release from a few days ago.

Some issues: I use VLC for Android and MX Player as an external player when watching videos through NewPipe. When watching 720p60, 1080p or 1080p60 (and higher of course), then there is no sound to be heard.

Can it be that the sound is only successfully muxed when using the (new) integrated player?

PR #516, opened 2017-04-26
Implement state saving and improvements

### LINKED ISSUE #420: List of search results disappears on device rotation
When I perform this sequence of events, the list of results disappears

1. Click on Search Icon
2. Enter Text
3. Perform Search
4. Orientation Change

### LINKED ISSUE #449: [BUG] NewPipe reloads content of search results when switching between apps
NewPipe reloads search results after switching between apps. Even if search results are scrolled down it reloads them after switching back to NewPipe window.

### LINKED ISSUE #466: Switching to other program or program overview ("alt-tab") stops/resets playback
If I even so much as go to the alt-tab-ish overview of programs and switch right back to newipe, the video that was playing is reset to a fresh state as if I only opened it upon switching back. Playback continues from 0 and the video has to be downloaded again. Was like this half a year ago when I first tried newpipe too.

mauriciocolli:
- Settings refactored
    - Now when the user change the theme, the change is immediate, no more restarting
- Search, Videos and Channel
    - Don't reload on device rotation
    - Don't reload the items again, e.g. you go from the search to a video, if you go back to the search it won't reload everything again, same for videos and channel, that provides a quick navigation
    - VideoDetail UX improved
- Implement Toolbar app-wide
    - Ditch the SearchView, use custom implemented search with AutoCompleteTextView
    - Replace old action bar
- Remove Toast errors
    - Add error messages with a retry button in the layout itself
- Improve workers
    - New workers for search and suggestions
- Add cache functionality
    - Now the app will cache the StreamInfo's that are already loaded, so if the user decides to come back to that video, it'll open instantaneously

The list of related videos had an issue, we were trying to inflate 20+ views, that's just too much (at least in mid-low devices, like mine) and it causes to skip frames, and as result, a jittering and laggy animation.
So I limited the related streams to a reasonable amount and added a expand button to load more.

Closes #420, closes #449 and closes #466

@theScrabi Hey, remember to merge the https://github.com/TeamNewPipe/NewPipeExtractor/pull/5


mauriciocolli:
This last commit change the `build-tools` of travis from `25.0.0` to `25.0.2`, matching the version on the `build.gradle`, but it didn't passed because it need the changes from the NewPipeExtractor to work.

theScrabi:
Oh I thought I had already updated the build tools :P Might have been another app. Thanks.

PR #538, opened 2017-05-08
Add Bangla translation (for Bangladesh)

cwgt:
added translation for [Bangla](https://en.wikipedia.org/wiki/Bengali_language) for Bangladesh.


cwgt:
@theScrabi please publish 0.9.5 to f-droid.

PR #545, opened 2017-05-09
Fix image loading bug

mauriciocolli:
After closing a player (popup in this case), and the image loader was still loading, `onLoadingComplete` is still being called and the notification control was being created after the popup was destroyed (getting stuck, user cannot close it) as the `loadImage` task is not cancelled, and we cannot cancel these with the current imageloader.

And that's one more reason that we **still** need to change it, the one that we are using (UIL) is discontinued since 2015.


theScrabi:
Well you are right. Than we should get rid of our current image loader.

PR #546, opened 2017-05-09
Change order of buttons in the title bar

### LINKED ISSUE #471: Move kodi button to better position

Hi,

I really like the "play on kodi" button, however it is in the dropdown menu and not accessable as quick as it could be. I have appended an image with the kodi button on a better position. In this case it might be better to diable the feature by default, because people could get distraced by a second play button without purpose for them.

newpipe-kodi

mueller-ma:
Closes #471 
The new order is: Share, Kodi, Download, Browser

### DIFF
--- a/app/src/main/res/menu/video_detail_menu.xml
+++ b/app/src/main/res/menu/video_detail_menu.xml
@@ -3,9 +3,9 @@
       xmlns:app="http://schemas.android.com/apk/res-auto">
 
     <item
-        android:id="@+id/menu_item_share"
-        android:icon="?attr/share"
-        android:title="@string/share"
+        android:id="@+id/menu_item_download"
+        android:icon="?attr/download"
+        android:title="@string/download"
         app:showAsAction="always"/>
 
     <item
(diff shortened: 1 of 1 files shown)


app/src/main/res/menu/video_detail_menu.xml
```diff
@@ -3,16 +3,10 @@
...
-        android:id="@+id/menu_item_download"
-        android:icon="?attr/download"
-        android:title="@string/download"
-        app:showAsAction="always"/>
-
-    <item
         android:id="@+id/menu_item_share"
         android:icon="?attr/share"
         android:title="@string/share"
-        app:showAsAction="ifRoom"/>
```
theScrabi: As @mauriciocolli [noted](https://github.com/TeamNewPipe/NewPipe/issues/471#issuecomment-300211270) it, the download button should always stay visible, it should not stay on `ifRoom`.

mueller-ma: @theScrabi I changed it. Should I squash the commits?

PR #550, opened 2017-05-11
Hide the 'Threads' slider from Download dialog.

SkyzohKey:
This pull request hides (`visibility: gone`) the 'Threads' slider in the Download dialog as it makes the UI weird and not user-friendly, also changed the radio buttons order to make it more straightforward as I guess many users download songs with NewPipe, removing a little user friction against the UI.

### DIFF
--- a/app/src/main/res/layout/dialog_url.xml
+++ b/app/src/main/res/layout/dialog_url.xml
@@ -43,17 +43,18 @@
         tools:ignore="RtlHardcoded">
 
         <RadioButton
-            android:id="@+id/video_button"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:checked="true"
-            android:text="@string/video"/>
-
-        <RadioButton
             android:id="@+id/audio_button"
             android:layout_width="wrap_content"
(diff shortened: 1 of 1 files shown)


mauriciocolli:
I'm currently refactoring the `Download*` classes, and I thought in getting rid of this, the normal user don't have any idea of what the hell is this.

So I'll default the threads to 8 (which is a good number) and maybe include an option in the settings for the more advanced users. What do you think?

PS: I guess getting rid of this GigaGet is the best option, it's an old project

theScrabi:
8 threads per download is much to high as a default. Think of the overheat.
For a default 2-3 ar quiet enough.

As you said an advaced option for altering the threads would be nice.

mauriciocolli:
Yes, there are more important things, just told it for the sake of record

martian742:
Opinion:  hide the threads from the dialogue OR don't hide it, I personally don't care.

But please **let me set my own default in the settings**, I prefer to set 1 thread most of the time so I end up touching the slider with every single download. :)

Thanks for the awesome app though! I use it every day.

Issue #528, opened 2017-05-03
play without full screen


ghost:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

-------------

it will nice to play video without full screen.

theScrabi:
Please give a small explenation.

cwgt:
when I click on play button it looks like this 

![screenshot_2017-05-03-21-45-27](https://cloud.githubusercontent.com/assets/17215321/25669037/3b254164-304a-11e7-8718-71f903825173.png)

but I like play this way (I know popup will be resizable.)

![screenshot_2017-05-03-21-46-25](https://cloud.githubusercontent.com/assets/17215321/25669094/62cfc072-304a-11e7-8724-6afdf04cdec6.png)

theScrabi:
Maybe it's because of the content of the video.

cwgt:
I played some videos but all of them works this way. (I am using android 4.2.2)

cwgt:
@theScrabi

I was trying to build it myself (not a Android dev) with AIDE.

but it can not find 'APPLICATION_ID' and 'VERSION_NAME' (not a member of BuildConfig). how to generate them?  (tried to add `buildConfigField` with no luck)

theScrabi:
Hm strange, version code and version name can be found [here](https://github.com/TeamNewPipe/NewPipe/blob/master/app/build.gradle#L11)

Did you try to build the app without the IDEA Ide? You could try to build it via commandline:
`[dir/to/NewPipe/root: ]$ ./gradlew assemble`

cwgt:
I was trying to build it on android (AIDE is a android app).

it seems, it was a problem of AIDE
 (https://plus.google.com/112950748876729349038/posts/8wKAAqaqVEm).

theScrabi:
Please use your computer to build it.

SkyzohKey:
@theScrabi I guess he's wanting to say that the player should be able to play the video without going fullscreen as it currently do. Something similar to how YouTube official app handle that. :)

cwgt:
@SkyzohKey exactly. I can get expected result by using resizable popup.

SkyzohKey:
I'm working on this right now, you can expect a PR at the end of the day I guess.

cwgt:
@SkyzohKey :smile:

[...]

PR #551, opened 2017-05-11
Add in details view video playing.

SkyzohKey:
Ok well, this is not perfect but it should give a base for further modifications.
I searched the net to embed an Activity inside a Fragment, or a Fragment into
another Fragment but didn't found how. So i just made some copy/paste + code
modifications to enable in details view Video playing, avoiding to open another
activity. This can be enabled via the settings (disabled by default) and is an
experimental stuff, probably not meant to be merged, but at least, it works!

Hope something more competent than me in Android development (cc @theScrabi) will
take a look at this and found a way to make the code smarter.

This PR tries to address the issue #528 :p

There are still issues:
- [ ] I'm not able to theme the `MediaController` without changing completly the view Theme.
- [ ] The video player is missing buttons for fullscreen, changing orientation and playing in loop.
- [ ] Code is an evil shit and copies a lot of code (don't repeat yourself, xD) from `PlayerVideoActivity.java`

My best hope is to convert the PlayerVideoActivity as a reusable View that we could just load with some arguments instead of all that garbage code. c:

Enjoy!

[Debug APK](https://a.doko.moe/lbzrcw.apk) !

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -4,9 +4,12 @@ import android.app.Activity;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.media.AudioManager;
+import android.media.MediaPlayer;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
+import android.os.Handler;
 import android.preference.PreferenceManager;
 import android.support.annotation.FloatRange;
 import android.support.annotation.NonNull;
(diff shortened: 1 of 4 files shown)


TobiGr:
While testing the apk you provided I found some issues which might be solved before a merge can be done. But I didn't dig into the code. That's something @theScrabi or @mauriciocolli can do. So maybe you should wait for their response before fixing the issues.

- On my tablet (800x1280) the video view is not centred when the resolution of the video is lower than the screen resolution. Sometimes it is a little centred but more on the left side than in the middle, sometimes it is completley on the left side of the screen. 
newpipe_details_view_video_playing_1
If the video's resolution is higher than the screen's, the app gets slow and stops responding and after a while Android asks to close it.
Maybe the video view should be as width as the screen.

- An orientation change results in a stopped video. Furthermore the "normal" details view with the video's thumbnail is displayed.

- Two orientation changes right after another cause a crash of the UI:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'void android.widget.MediaController.hide()' on a null object reference 
  at org.schabi.newpipe.fragments.detail.VideoDetailFragment.hideUi(VideoDetailFragment.java:1068) 
  at org.schabi.newpipe.fragments.detail.VideoDetailFragment.access$2500(VideoDetailFragment.java:81) 
  at org.schabi.newpipe.fragments.detail.VideoDetailFragment$10.run(VideoDetailFragment.java:1055) 
  at android.os.Handler.handleCallback(Handler.java:815) 
  at android.os.Handler.dispatchMessage(Handler.java:104) 
  at android.os.Looper.loop(Looper.java:194) 
  at android.app.ActivityThread.main(ActivityThread.java:5576) 
  at java.lang.reflect.Method.invoke(Native Method) at java.lang.reflect.Method.invoke(Method.java:372) 
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:956) 
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:751)
```

At the moment I cannot test it on an other smaller device.

mauriciocolli:
He used the slow android's `VideoView`, and I already told him in the issue (#528) to use the exoplayer.
Also orientation changes have to be considered.

TobiGr:
Sorry. I missed that.

PR #559, opened 2017-05-14
Change popup default resolution

### LINKED ISSUE #558: Default Resolution

Why is the default resolution of the pop-up player (480p) higher than the default resolution of the normal player (360p)? That seems weird to me, because the pop-up player is usually smaller than the normal player.

mauriciocolli:
Set the popup default resolution as the normal player default.

- Fix #558


wb9688:
Well, that was fast :)

PR #583, opened 2017-06-08
BackgroundPlayer: Add timestamp in expanded notification

SpajicM:


mauriciocolli:
I'll test this possibility a little more, but here some problems:

- You remove [this](https://github.com/TeamNewPipe/NewPipe/pull/583/commits/c89567ce7ad3fc5193bc34e8c32290bf6da485b1#diff-0b87dfa694cf6a7720aee912b16ab32dL291) `else if`, so if the duration is less than 15s the first `if` will be true, but the `else` in the second check will be true too, so the first check become useless
- The `PROGRESS_LOOP_INTERVAL` in the `BasePlayer` is just 100ms  (for normal video players, it's a good time), but you removed the corrected interval from the checks, because this is **way** too short for notification updates, that's 10 per second, it just lags all the phone.

And thinking about it, even 1s loop is kind of bad, and that's the reason why I didn't included the time originally, so I don't know if it's worth it (sacrifice performance over displaying the time).

Also, I think we are (wondering) in creating a specific UI just for the background player (open something like a dialog with more "advanced" options), so idk if this will be necessary, but let's see.

SpajicM:
You are right, underped if's now (force-pushed).

- Will leave it 1sec for now if you decide to merge.
- Separate UI for background player would be nice, since I had a few more ideas such as long-press for longer step in seeking back and forward, also putting SeekBar instead of ProgressBar could be useful, but RemoteViews do not support those.

mauriciocolli:
With the `1s` interval, my phone is freezing/hanging shortly every time the notification is updated (it's not a powerful device, but should be usable as NewPipe aims to be lightweight).

And that's the problem, I don't know if updating the notification that often is very efficient, I never saw one music app that display the current time in the notification (maybe because they have a UI).

When we implement the playlist, I think we will replace the fast-seek button with next-previous, so the user will not be able to control directly the time anyway (it only have rudimentary controls right now).

So, considering the previous reasons, I think it's better to remove the progress bar and make a proper UI for the background player. What do you guys think?

theScrabi:
Yes a propper ui would be something I realy want to have.

SpajicM:
@mauriciocolli yes I see the problem, on high-end devices it doesn't lag, but it seems to drain battery.

What kind of UI do you have in mind? Like a full music player, or something minimal, like pop-up Google Play Music has when you pick a file from downloads (see attached photo)? I'd prefer a minimal one.

mauriciocolli:
@SpajicM I prefer a minimal design too.

This is what I come up with (it's already functional but the design it's just a rough overview):

controls

theScrabi:
Hi, I imagined something like this:
[]()

where the controls are sticked to the bottom, and the rest of the UI keeps controllable.
As an extra we could make it so when we pull this control up, a more powerful control would show up

[]()

Issue #580, opened 2017-06-02
spaces converted to %20 in existing download directories


turboteckel:
If I have a folder with a space in its name and point NewPipe to download to it, it will create a new folder with the same name but with the spaces converted to %20. Haven't tested it with other characters.

NewPipe 0.9.7
Android 4.3

coffeemakr:
Good catch and thanks for reporting.
The issue should be resolved in the next release.

coffeemakr:
@theScrabi you can close this issue

PR #586, opened 2017-06-13
Fix download path handling (#580)

coffeemakr:
This fixes the URL encoding artifacts encountered when selecting folders with non-url characters.

Instead of stripping the trailing 'file:/' just call the getPath method.


theScrabi:
All right thanks, I'll merge and test it next weekend :)

PR #588, opened 2017-06-16
Speed up detail page loading ⏱

coffeemakr:
* Reorder layout to make it flat and remove replace some relative Layouts. Especially the following changes made a difference:
    * Replace top-level RelativeLayout in the main activity.
    * Flatten the stream item (because it's created often). Also I combined two TextViews because they caused a longer measurement time. 
    * Other layout changes:
      * Remove buttons where the layout can be clicked

* Move some costy calculation from `handleStreamInfo` to a background thread. Especially `Html.fromHtml` used much CPU time. 
 * ~Don't call animation listener immediately in the `animateView` method if the view is already shown/hidden. Instead add it to the UI handlers queue.~ 
 * The same for the `handleStreamInfo` call in `VideoDetaiLFragment.onActivityBackPressed`.

Other changes not related to performance:
 * Replace self-written logic to format duration strings with String.format calls. I removed the leading zeros for the first number. (e.g 01:20 ➡️  1:20)

If I missed another layout change please tell me so I can fix it!


app/src/main/res/layout/fragment_video_detail.xml
```diff
@@ -365,49 +329,43 @@
...
-                            android:layout_alignParentLeft="true"
                             android:layout_marginLeft="12dp"
                             android:text="@string/next_video_title"
                             android:textAllCaps="true"
                             android:textAppearance="?android:attr/textAppearanceMedium"
                             android:textSize="@dimen/video_item_detail_next_text_size"
                             tools:ignore="RtlHardcoded"/>
 
-                        <LinearLayout
+                        <android.support.v7.widget.RecyclerView
```
theScrabi: Role this one back please.
Since we are not using a coordinator layout you can't nest scroll views.
This means scrolling does not work properly if you use this.

mauriciocolli:
There're some problems in the video detail:
- The title, views and like/dislike are misaligned
- There's no more ripple in some buttons
- The related videos list is broken, now is just behaving like a nested scrolling
    - Try to expand the list, scroll all the way down, collapse and try to scroll up again

The most time-taking operation is creating the items of the related videos list (inflating every single item is taking some time), now is working faster because it is showing less items, but we have to focus on trying to optimize it.

coffeemakr:
I'll take a look at it and revert some changes!

coffeemakr:
> There's no more ripple in some buttons
I don't see any ripples in the current version. Could you tell me where you see them?

app/src/main/res/layout/fragment_video_detail.xml
coffeemakr: Oh! I didn't have a nested scroll view (maybe my phone's to old).
Maybe it could have been disabled with [setNestedScrollingEnabled(false)](https://developer.android.com/reference/android/support/v4/view/NestedScrollingChild.html#setNestedScrollingEnabled(boolean)).

coffeemakr:
The RecyclerView is reverted and the layout corrected to look the same as before (except the triangle after the title is a bit higher which can be corrected if wanted).

[Screenshot](https://user-images.githubusercontent.com/5737978/27252960-a547db72-536a-11e7-82ab-80952494b57a.png)

If you want I can also revert all or some more layout changes.

mauriciocolli:
There's some very small adjustments:
- Restore the ripple of the play button, you can do that by setting the foreground to this:
    - `android:foreground="?attr/selectableItemBackground"`
- Replace the [margin of the upload buttons](https://github.com/TeamNewPipe/NewPipe/pull/588/files#diff-eee8ebe2adaa79758d846ae84c730bc8R238) by padding, this makes it look better when pressed
- Fix indentation [here](https://github.com/TeamNewPipe/NewPipe/pull/588/files#diff-eee8ebe2adaa79758d846ae84c730bc8R98)
    - I recommend you to run a reformat code (`Ctrl + Shift + Alt + L`) with the two checkboxes checked (Optimize imports and rearrange code), this will fix the indentation and reorder the code according to importance, leading to a nicely formatted file

Nice work!

coffeemakr:
* Formatted code
* Restored the ripple
* Converted **all** margins of the uploader to paddings.

mauriciocolli:
Some things left to fix:
- Uploader text `stream_item`, is overlapping the title when in landscape (font is bigger) because it has some margin in the bottom
- Ripple from clear button is strange (size) in `toolbar_search_layout` due to removal of the `Framelayout`
- The scroll position is not saved when rotating the screen in `VideoDetailFragment`
    - After a long debug I found that `View.post` the `execOnEnd` in `AnimationUtil` is the culprit, I guess the view see something in the queue and don't save properly its state
    - Another issue is that the related videos are not there when the restore save instance happens (because you are posting to another handler), hence the view cannot restore its state properly (in this case, the scroll position)
    - So the `initRelatedVideos` has to happen before the views restore its state (synchronously), and it made me ask, why do we need to use `uiHandler` for that?

coffeemakr:
I'll take a look at it and I'm sorry that there are so many bug introduced by my changes. I'll be more careful next time. Thank you for your detailed review!

coffeemakr:
> So the `initRelatedVideos` has to happen before the views restore its state (synchronously), and it made me ask, why do we need to use uiHandler for that?

I've used the UI-Handler with the `UICallback` to handle message from the background handler and populate the changes to the UI (see [Communicating with the UI Thread](https://developer.android.com/training/multiple-threads/communicate-ui.html)). `uiHandler.post(..)` is pretty much the same as `getActivity().runOnUIThread(..)`.

> Just found out that there's nothing to do with the view, the culprit is this line VideoDetailFragment.java#870, as you changed to View.post, the call to handleStreamInfo was being delayed, thus causing the problem that I talked about.

I thought delaying the `initRelatedVideos` a bit could improve the responsibility of the app because the main thread has time to handle event between the "main-UI" and the related videos.
In order to make the remembered position work I've synchronized the execution in `AnimationUtils` and in `handleStreamInfo`.

mauriciocolli:
Now it just have one side effect:

###### Meh, it's kinda of nice.
test

But yes, (beside that) it's all good to merge.

coffeemakr:
I've fixed this issue which also existed in the channel fragment and the search fragment. I took the liberty to  create a new layout containing both the progress bar and the error & retry components.

But it still does look differently:
 * [Channel Fragment](https://user-images.githubusercontent.com/5737978/27609195-5f71e6c4-5b8a-11e7-9c4f-34647afd7e6f.png)
 * [Stream Fragment with loading icon](https://user-images.githubusercontent.com/5737978/27609329-c14174be-5b8a-11e7-9eab-a38b2b766935.png)
 * [Stream fragment without loading icon](https://user-images.githubusercontent.com/5737978/27609330-c1424f56-5b8a-11e7-8b54-fe7a6c055080.png)
 * [Search fragment](https://user-images.githubusercontent.com/5737978/27609331-c147fde8-5b8a-11e7-8867-b89e0836e2dd.png)

While currently the retry controls are shown at the top:
 [Current search fragment](https://user-images.githubusercontent.com/5737978/27609421-00f111b4-5b8b-11e7-91cb-0e20fe72e109.png)

mauriciocolli:
That's good, I think it's ready to merge!

mauriciocolli:
@coffeemakr now we have a new issue, if the thread is running and the user exits of the fragment and the result comes in this moment, it faces a NullPointerException because the fragment is already destroyed (the referenced view is null):
```
 java.lang.NullPointerException: Attempt to invoke virtual method 'void android.widget.TextView.setText(java.lang.CharSequence)' on a null object reference
   at org.schabi.newpipe.fragments.detail.VideoDetailFragment$UICallback.handleMessage(VideoDetailFragment.java:1228)
   at android.os.Handler.dispatchMessage(Handler.java:101)
   at android.os.Looper.loop(Looper.java:164)
   at android.app.ActivityThread.main(ActivityThread.java:6540)
   ...
```

coffeemakr:
Alright I'll add a check.

PR #596, opened 2017-06-26
Adds support for adjustable playback speed

### LINKED ISSUE #153: Add adjustable playback speeds
Could you add adjustable playback speeds to the bundled media player? For example, -1.5x, 1x, 1.5x, etc.

marcobiscaro2112:
ExoPlayer was updated to 2.4.2, which supports playback speed change.

A speed selector was also added in the MainPlayer and PlayerPopup.

Fixes #153.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -50,5 +50,5 @@ dependencies {
     compile 'de.hdodenhof:circleimageview:2.0.0'
     compile 'com.github.nirhart:parallaxscroll:1.0'
     compile 'com.nononsenseapps:filepicker:3.0.0'
-    compile 'com.google.android.exoplayer:exoplayer:r2.3.1'
+    compile 'com.google.android.exoplayer:exoplayer:r2.4.2'
 }
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -22,6 +22,7 @@ import com.google.android.exoplayer2.DefaultLoadControl;
 import com.google.android.exoplayer2.ExoPlaybackException;
 import com.google.android.exoplayer2.ExoPlayer;
 import com.google.android.exoplayer2.ExoPlayerFactory;
+import com.google.android.exoplayer2.PlaybackParameters;
 import com.google.android.exoplayer2.SimpleExoPlayer;
 import com.google.android.exoplayer2.Timeline;
 import com.google.android.exoplayer2.extractor.DefaultExtractorsFactory;
@@ -47,6 +48,8 @@ import com.nostra13.universalimageloader.core.ImageLoader;
 import com.nostra13.universalimageloader.core.listener.SimpleImageLoadingListener;
 
 import java.io.File;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
 import java.util.Formatter;
 import java.util.Locale;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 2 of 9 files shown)


mauriciocolli:
Hey, thanks for the work.

We just have some things to fix:
- The position of playback speed selector in the main player, is it the best spot? 
    -  I think it's better to put with the others and give some style (like bold text) and as the player is usually used in the landscape orientation, there's a lot of space for it, I think of something like [this](https://gist.github.com/mauriciocolli/3cb2fdc330e1a5a31ac0ae7df1081cf7)
- Regarding the popup player, is hard to find a position because is small
    - For now the position is not very good, when the popup is very small it just overlaps the others buttons and when it does have space, it just not fit into the design
    - I'll think more about it, so if you have an idea just tell me
- Remember the playback speed when switching to popup/fullscreen

marcobiscaro2112:
> The position of playback speed selector in the main player, is it the best spot?

I don't know. I didn't put the selector in the suggested spot at first because there is almost no space left for the title in portrait mode: 

image

Anyway, I've changed it to the suggested spot now.

> Regarding the popup player, is hard to find a position because is small 

Yes. Actually I've tried some things and the place where I put it was the least worse I found.

> when the popup is very small it just overlaps the others buttons and when it does have space, it just not fit into the design

I couldn't reproduce this. When I tested the popup at the smallest size possible, it didn't overlap:

screenshot_20170626-191224

> Remember the playback speed when switching to popup/fullscreen

Done.

mauriciocolli:
> I couldn't reproduce this. When I tested the popup at the smallest size possible, it didn't overlap:

I mean when the resolution text is big:

screenshot_1498517799

But anyway, I think the minimum size of the popup it's just too small (for example, it's just impossible to use the seekbar (it's already difficult in its normal size)), and now that we'll have a speed selector we can increase a little.

---

Now another problem, if a user use the speed selector in a high resolution, probably the device won't be capable of handling it, it's _already_ an issue to high resolutions alone imagine with the speed being 2x.

I think that's the motive that YouTube itself didn't implement it in their app, the crash rate would increase significantly, how many clueless users would be like wtf...

But for now I think we can just ignore this issue, our userbase is not as massive and I guess our average user is more capable of figuring it out what happened (I hope).

theScrabi:
@mauriciocolli all right, if you are god with that would you please test and merge it.

mauriciocolli:
There's one more problem that I found:

If you change the playback speed while the player is paused, then when you switch to fullscreen/popup you don't get the playback speed that you set, the `getPlaybackSpeed()` return the old value because the `ExoPlayer.getPlaybackParameters()` return the old too, [from the javadoc (setPlaybackParameters())](https://google.github.io/ExoPlayer/doc/reference/com/google/android/exoplayer2/ExoPlayer.html#setPlaybackParameters-com.google.android.exoplayer2.PlaybackParameters-):

> ExoPlayer.EventListener.onPlaybackParametersChanged(PlaybackParameters) **_will be called whenever_** the currently active playback parameters change.

As you can see from the docs (I think) that's the correct behavior, only when the listener is called is when the `playbackParameters` of the `ExoPlayer` is set, so we'll have to keep a variable to keep track of the changes, another quote from javadoc:

> When that listener is called, the parameters passed to it **_may not match_** playbackParameters. For example, the chosen speed or pitch **_may be out of range_**, in which case they are constrained to a set of permitted values. If it is not possible to change the playback parameters, the listener will not be invoked.

And I guess that's the reason that it don't change immediately, the player have to do some check if it's possible to change the parameters (I did some overview through the ExoPlayer source code, and that seems to be the case).

theScrabi:
@marcobiscaro2112 @mauriciocolli what is the status on this?

mauriciocolli:
Thanks for your work!

marcobiscaro2112:
@mauriciocolli Sorry for not responding before, it was a very very busy week.

About the problem you found, I'll open a new PR to fix this little bug, as the code was already merged.

PR #599, opened 2017-06-28
Code Improvements

coffeemakr:
Hello

I thought it would be nice to clean up the code a bit and I hope I'm not spamming :)

There are only a few functional changes:
 * A common class for the `RecyclerView.OnScrollListener` loading additional items: ecc2aae
 * An enum called UserAction for error reporting (instead of ints and strings): b963cc5
    * This also changes the way the `ErrorInfo` is written to a parcel. Therefore i created a instrumented test for this class to test the process of writing to and reading from a parcel: 1c29780

Removed code:
 * From Utilities:
   * [findViewById](a790a52119db67e5eb25fda6da66ec16a2eadc3e): because it's a unchecked cast to save a few characters.
   * [isMusicFile](79c4d5beffe29145c28a3799b58cff574c50155a): never used
   * [isVideoFile](79c4d5beffe29145c28a3799b58cff574c50155a): never used
   * [isDirectoryAvailable](https://github.com/TeamNewPipe/NewPipe/commit/de69d88f4ed4fbdccc310f4b6501ec4744350a82#diff-d791e158184c692a0e6cba479ef84397L184)
   * [isDownloadDirectoryAvailable](https://github.com/TeamNewPipe/NewPipe/commit/de69d88f4ed4fbdccc310f4b6501ec4744350a82#diff-d791e158184c692a0e6cba479ef84397L189)
   * [showDirectoryChooser](https://github.com/TeamNewPipe/NewPipe/commit/de69d88f4ed4fbdccc310f4b6501ec4744350a82#diff-d791e158184c692a0e6cba479ef84397L193): Is implemented in `SettingsFragment.onPreferenceTreeClick`

Non-Functional changes:
  * Add annotations
    * for resource, string, color and drawable id's
    * in overridden methods where the arguments are annotated (7bd42d5)
    * Some Nullable and NonNull annotations
  * Format code (especially in GigaGet code: 2756ac8). Unfortunately the diff is therefore a bit unclear.
  * Fix typos
  * Replace deprecated method calls with the corresponding compatibility method call. dd08787

I tried to make small commits so the changes can be reverted but it'll make sense to squash them when merging.
Regards

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -5,7 +5,7 @@ READ THIS GUIDELINES CAREFULLY BEFORE CONTRIBUTING.
 
 ## Crash reporting
 
-Do not report crashes in the GitHub issue tracker. NewPipe has an automated crash report system that will ask you to send a report if a crash occures.
+Do not report crashes in the GitHub issue tracker. NewPipe has an automated crash report system that will ask you to send a report if a crash occurs.
 
 ## Issue reporting/feature request
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,6 +10,8 @@ android {
         targetSdkVersion 25
         versionCode 35
         versionName "0.9.8"
+
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
     }
     buildTypes {
         release {
@@ -31,6 +33,11 @@ android {
 }
 
 dependencies {
+    androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2') {
+        exclude module: 'support-annotations'
+    }
+
+
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:1.10.19'
     testCompile 'org.json:json:20160810'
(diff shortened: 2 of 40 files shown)


theScrabi:
Please squash your commits into one

mauriciocolli:
Hey wait a second, is easier to review multiple commits and—as he already said, to revert changes too.

> I tried to make small commits so the changes can be reverted but it'll make sense to squash them when merging.

theScrabi:
All right :P

mauriciocolli:
IMHO, I think we have to get rid of giga, it's just too old and [deprecated](https://github.com/PaperAirplane-Dev-Team/GigaGet) and I'm already looking for a good replacement for now ([android-arsenal](https://android-arsenal.com/tag/179?sort=updated) seems to have some good libraries), if you have suggestions just tell me.

Regarding this pull request, it's already good to merge, just waiting for you to squash the commits.

coffeemakr:
> IMHO, I think we have to get rid of giga, it's just too old and deprecated and I'm already looking for a good replacement for now (android-arsenal seems to have some good libraries), if you have suggestions just tell me.

It seems pretty buggy to me too (e.g. raise conditions to download a block twice). I've made some experiments with the [builtin download manager](https://developer.android.com/reference/android/app/DownloadManager.html) and it seem to work pretty well. Maybe this is also an option.

> Regarding this pull request, it's already good to merge, just waiting for you to squash the commits.

Squashed!

mauriciocolli:
I'd prefer some download manager that has a way to download in chunks, but I'm yet thinking if this is that important, the built-in download manager is indeed nice.

Thanks for your work.

PR #603, opened 2017-07-01
Fix margin at the bottom

### LINKED ISSUE #601: Space in DetailedView where there should be no space

Reproduce:
Go to `DetaildView` and scroll to the bottom.

bildschirmfoto vom 2017-06-30 21-03-33

coffeemakr:
I've fixed #601 which was caused by a padding in the layout for the loading and error views.

The following steps lead to this solution:
 * It was hard to determine how much space the progress bar would take so it can be hidden without changing the layout because it uses a style `@style/Widget.AppCompat.ProgressBar`.
 * I've decided to set the progress bar to `INVISIBLE` instead of `GONE` so there's no need for a margin "hack"
 * To achieve this I didn't want to change the hiding process in each fragment (video, channel and search) so I made the views private and added methods to the BaseFragment to cotroll this process.
 * To remove some complexity from `BaseFragment` I created the `LoadingIndicator` custom view and moved all the logic to this class.

Don't hesitate to request any changes.


mauriciocolli:
Now there's a wasted space that will confuse users (if they are on landscape):
[](https://user-images.githubusercontent.com/13389907/27765114-d6e92f26-5e7f-11e7-918c-f5c3daa107fe.png)

In the current implementation just removing the padding from the root layout and putting somewhere else is enough to fix this issue specifically.

I was wondering, is coupling all layouts/elements together good? I'd prefer to keep the ProgressBar and the error layout separated, it'd easier to place wherever we want and give custom properties that way.

And I think that doesn't matter that they overlap (ProgressBar and the error layout), because they aren't supposed to be visible together, so just placing them in the center shouldn't be a problem (in the case of the video detail, below the title).

coffeemakr:
I too was wondering if it is a good idea and thought it easy to separate again but I'll put the views back into BaseFragment.

> they aren't supposed to be visible together

They are shown together when an error is shown and the retry button is pressed. The progress bar just doesn't show up for a long time. You can press the button repeatedly to see it.

Should I make a minimal fix instead of this?

mauriciocolli:
I took that in consideration when I said that, in a real case scenario I think this won't be a problem, because as soon as the user hit the retry button, the error message disappear and the loading bar appear.

We are seeing that just because the error is being thrown very quickly (the system knows that there're no internet) and the retry button is more useful when occurs a timeout, for example.

> Should I make a minimal fix instead of this?

Do you mean just remove the padding? I was thinking in reverting some changes, something like [this](https://gist.github.com/mauriciocolli/88deba4a39fc3267a348abeac8b95207).

coffeemakr:
> retry button, the error message disappear and the loading bar appear.

Ah I see. In my implementation the error message is kept until loaded successfully. I thought that's they it should be because the animation shows them together. (here from v9.8.0 in the channel fragment)
Channel loading progress

Maybe the progress bar could be hidden until the error message is hidden. 

> Do you mean just remove the padding?

Yes I thought of this but your solution is better. I'll close this request and create a new one.

PR #612, opened 2017-07-05
Add about activity

coffeemakr:
I have created a about activity which includes the following information:
 * NewPipe's License
   Maybe this should include something like `(c) 2017 by ...` or similar. 
 * Note on how to contribute and a link to GitHub
 * Copyright of the software components used including full license texts

**Additional ideas**
 * Donate button
 * Tab containing the "privacy policy":
   * What data is sent to YouTube
   * What data is included in the crash reports

**Implementation Notes**
The software components list is created by inflating the layout and adding it to the `LinearLayout` because I couldn't get it working otherwise. `ListView` doesn't support the `NestedScrollingView` scroll behavior and I didn't get the `RecyclerView` to handle the context menu correctly (+ the scrolling needs to be disabled).   

**Screenshots**
 * [About](https://user-images.githubusercontent.com/5737978/27877057-8be983d0-61b9-11e7-94cb-52f5e21a5703.png)
 * [Licenses](https://user-images.githubusercontent.com/5737978/27877055-8be8ae2e-61b9-11e7-89fd-a7385c210f0d.png)
* [License context menu](https://user-images.githubusercontent.com/5737978/27877056-8be92cf0-61b9-11e7-81d2-c3447a294990.png)

Please let me know if you like it, have additions or want anything changed.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -212,11 +212,7 @@
         <activity
             android:name=".about.AboutActivity"
             android:label="@string/title_activity_about"
-            android:parentActivityName=".MainActivity"
             android:theme="@style/AppTheme">
-            <meta-data
-                android:name="android.support.PARENT_ACTIVITY"
-                android:value="org.schabi.newpipe.MainActivity" />
         </activity>
     </application>
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -209,7 +209,15 @@
                 <data android:mimeType="text/plain"/>
             </intent-filter>
         </activity>
-
+        <activity
+            android:name=".about.AboutActivity"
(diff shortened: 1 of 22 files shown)


mauriciocolli:
Very small issues:
- The up button recreate the whole MainActivity
    - We can just finish the current activity when `android.R.id.home` (up button) is pressed, because the only way to open it, is through some other activity (always will have another one in the stack)
    - Then we can remove the parent activity of the manifest
- Some dimens introduced are not used

If you want I can fix those issues after the merge.
Also, what do you think is better: about in the menu, in the settings or both?

Thanks for this!

coffeemakr:
I think I fixed the two issues. I always allow edits from maintainers in my pull request so feel free to commit any changes to these branches.

In my opinion the about page is not related to the settings so I wouldn't include it in the settings.

theScrabi:
@coffeemakr cool thing :+1:

Issue #563, opened 2017-05-16
Requirements for Database


theScrabi:
So I got asked how to build the database.

My Idea for the design was this:
I want a central sql file which creates the database, and which will tell us the architecture of the database. Now this script should be put into android resources. There should be a class, some sort of `SQLLoadExecuter` that will load and run this script when the database does not yet exist.

Now every table in our database must contain a `serviceId` element.

Imaginable tables are:
- Search history
- Bookmarks/favourites
- subscribed channels
- history

Everything that is related to the database should be put into the package `org.schabi.newpipe.database?`

What do you think?

mauriciocolli:
Yeah that was basically what I was thinking.

Regarding the sql file, I don't have much experience from db in android, but do we need it?
From my understanding we create a subclass of `SQLiteOpenHelper` and that deals with when it is created (like @BlenderViking did [here][1]).

---
I'm considering using some DB library, that would not only be easier to maintain, but faster to develop too (I'm testing some of them yet). 

So if we find one great library, should we use it (by the previous reasons) or should we develop our own "custom" database?


[1]: https://github.com/BlenderViking/NewPipe/tree/playlist_sqlite/app/src/main/java/org/schabi/newpipe/playlist

theScrabi:
On android you basicly use SQLite, and thats build in. So before including some more libs I think we should use the standart. I worked with that before and I think it's not that hard to use.

Importend to me is having an sql file that will create the tables. Thats for us to see better how the database is build up.

theScrabi:
@TheAssassin had the Idea of using KeyVelue store instead of an SQL database, in order to ensure our data keeps consisted even if we update it a lot. What do you thing?

mauriciocolli:
But then how would that work? Serialize the object? Json? Use the "prefix" approach (e.g. `video<Id>_title`)?

TheAssassin:
It doesn't necessarily have to be key-value only, something hierarchic (JSON-like) would be ideal.

The problems that I mentioned in the discussion with @theScrabi is that keeping the SQL schema in sync would require some database migration framework to be in place. I am not sure if there's some "standard" framework for this purpose for Android, but yeah, the way NewPipe development works might make it hard to keep the database in a consistent state.

[...]

PR #620, opened 2017-07-09
Added basic channel subscription and feed pages

karyogamy:
Hi,

I have added channel subscription and feed support as pages on the main UI.

Specifically, this change introduces Room persistence and RxJava2 into the app. 

Since Room is now part of the Android architecture component, it is a good replacement 
for direct SQLite access and it saves on boilerplate code. 

As for RxJava2, the reason for using it is twofold:
 - It is directly supported by Room persistence and provides instant update (reactive) 
 when the table is changed.
 - It allows easy caching and sharing for web requests, where multiple fragments can
 obtain synchronized results reactively without needing to send multiple requests.
 This is used in subscription and feed fragments where the needed data are extracted
 from the same request.

Here are some screenshots:

The feed is the latest stream item of the channel, if it has one.
Please note, the "1 video" means that channel has no new videos/streams available.
This requires an UI change that may need to add an additional enum into info item which is
inside the extractor.

In addition, a reactive subscription button is added to the channel fragment:
This button has a debounce(delay) time of 0.5 second to prevent rapid clicking/database change.

However, when subscription list becomes too large, slowdown can occur as each channel sends
a web request to the source (in this case, Youtube) concurrently. Of course, the thread count
can be reduced to a constant if needed.

There are also problems caused by https://github.com/TeamNewPipe/NewPipeExtractor/issues/6,
which manifests as multiple subscriptions are the same channel. This is rectified by when
a mismatch with the database entries occurs and the subscription list is re-updated.

Features missing: 
- Rigorous exception handling (activity simply quits when anything other than IOException occurs)


theScrabi:
@karyogamy I looked at the app, and the codebase, and I like this. There are a few thinks that can be changed, but in my opinion we could do this after a merge.

@mauriciocolli can we have this pls? What is your opinion?

karyogamy:
@theScrabi Thanks, and to address your points:
1.
The reason the description is empty is that channel extractor (and ChannelInfo) currently does not have access to the description element for user/channel (at least on Youtube). This information is only available during search and inside the "About" page in the channel pages.

2 & 3.
I disabled swiping across the view pager because, on subscriptions fragment, you can swipe a channel to unsubscribe from it, which makes it very easy to unsubscribe accidentally when you just want to swipe the pages. Since we have no pager swipe, I've opted to move the tabs to the bottom, so I can access them more easily with my hand.

However, if the swipe to unsubscribe is unnecessary, I can remove it and re-enable page swiping and move the tab to the top.

4.
Good idea. Will do.

mauriciocolli:
Some problems that I saw after an overview:
- Handle configuration changes properly (https://github.com/TeamNewPipe/NewPipe/pull/620/commits/8121a53561f36296be8614fd6355cbc745fa2839#diff-c8cc2dd42271f2bf39c6aa81eb6a5529R23)
    - Please don't do that, these are not the only configuration changes, and what if we have a landscape layout? And what if the application went to background and was killed?
- Are we requesting a `ChannelInfo` for every subscribed channel and picking the first video?
    - As you said, this will be a problem if the user have a lot of subscriptions, I think it will hit the Google's limit and he'll have to resolve captchas more often
- Stetho being included in every build, even in release ones

But anyway, I think we can create another branch and merge it there, @theScrabi what do you think?

karyogamy:
@mauriciocolli 
- I looked into this, and you are right. However, saved instance states are not retained

karyogamy:
@mauriciocolli 
- I looked into this, and you are right. However, under main fragment and its children fragments,
the saved instance states are not retained across activity destory and re-create because of the 
following: 
```
public static void openMainFragment(FragmentManager fragmentManager) {
    ImageLoader.getInstance().clearMemoryCache();
    fragmentManager.beginTransaction()
            .setCustomAnimations(R.anim.custom_fade_in, R.anim.custom_fade_out, R.anim.custom_fade_in, R.anim.custom_fade_out)
            .replace(R.id.fragment_holder, new MainFragment())
            .addToBackStack(null) // This line does not exist currently
            .commit();
}
```
- Having no back stack on the main fragment drops the instance states, while adding the fragment to the backstack causes a additional blank page with toolbar when back is pressed on the main fragment. I'll probably take some more time to look into this issue later.

- As for the feed streams, you are correct. I'm getting it from the first item of channel info.
For the concurrency issue, we can simply change rxjava scheduler to use a pool of limited number
of threads to avoid going over the limit.

- I'll probably going to do something like this: http://littlerobots.nl/blog/stetho-for-android-debug-builds-only/

mauriciocolli:
I'm currently refactoring some parts of the app and that was an issue too, I fixed it like this:
- To remove the need of this fragment being recreated every time (as it's accessible from all parts of the app that have up button), I'm by adding to the backstack using a TAG
- I created a function that is called `goToMainFragment`, all it does is:
    - Call the `fragmentManager.popBackStackImmediate` with the TAG, if it returns _true_, it means that the fragment was already created and the pop was successful (all top fragments are popped)
    - If returns _false_, pop all the fragments and then just creates the fragment as normally would (adding to the backstack using the TAG), .
- Regarding the empty app after all fragments being popped, just adding a check if the backstack count is 0 already fix it.

karyogamy:
@theScrabi @mauriciocolli 
I think I have covered about everything except for merging tabs with toolbar and swiping.

Tell me if you guys still want these missing features or if there's anything else that might be a good addition.

theScrabi:
All right thanks.
But I think swapping would still be something useful, would be nice if you could try to add this tho.
However I would consider to merge anyway.

mauriciocolli:
Well, it seems like a perfect opportunity to learn more about Rx, and regarding the db, I think that we can close #563 then, Room it is.

Also, I don't think this is ready to general release, I've just tested 120+ subscriptions and it lagged the app (I think the cause is the big extractor creation (or just too much in parallel as you warned), as it have to hold some relatively heavy things,  GC was going like crazy) cpu staying in ~70-80% and ram usage in 45-75MB (my phone is low end), only in the initial fetch.

Another issue is that you have to load everything again in every start, leading to waste of data/bandwidth.
Look at it (keep in mind this is a rough estimation):
- Each Youtube channel page (the `/videos` one) has ± 160 KB
120*160 = 19200KB = 19.2MB
- I think at least 50 channels is a reasonable amount of channel to the average user:
50*160 = 8000KB = 8MB

And yet is a relatively big download to just initialize the app every time.
###### Of course, this amount of subscriptions is usually in Youtube itself, that I'm sure a large amount of users want to be able to import them.

This makes me question if having a feed is reasonable...

karyogamy:
@theScrabi Alright, I'll add back the swiping. However, I don't think merging the tabs with the toolbar is a trivial change since they are in different layouts. 

@mauriciocolli Yeah, that was my observation as well. Eventually we will need to do the same with playlists and this might just be too much internetz for my meager data plan.

How about this?
- We reduce the subscription service to only observe local db, make subscription completely offline (we will keep minimum identifiers for the users, such as channel name and thumbnail). 
- Subscription entries in the database can be updated when user opens up the respective channel fragment. Subscription fragment loses the ability to tell if a new video is available.
- As for the feed fragment, we keep doing the same thing as now but pull a fixed number of items initially and pull some more when user scroll to the bottom of the list.

theScrabi:
By the way there is one thing I just remebered. Do you store the serviceId in your db entries? If not please add this. (I can't look at the xode right now)

karyogamy:
@theScrabi Yes, serviceId and page url together are used as unique index.

karyogamy:
@theScrabi @mauriciocolli 
I reworked the subscription system and now it should be lightweight enough to support a large subscription pool. 

This is a proof of concept, so it might contain some bugs. 
Also, you will need to wipe your database since I changed the schema a bit.

theScrabi:
Hm. Could you try to use the already exiating channeItems for the list, and than also save things like description subscriber vount etc.

karyogamy:
I shrunk the ui for subscription items primarily because I don't understand why we need to display descriptions and subscriber counts (looking at Youtube app's design for subscription list). 

Also, getting description will require another call to the user channel (not implemented in extractor):
channel video page: https://www.youtube.com/user/xyz/videos
channel description page: https://www.youtube.com/user/xyz/about

mauriciocolli:
Just to inform you (not opining if we need to show it or not), the description is already included from the page we are extracting, so it's very easy to implement it:

```html
<html>
...
<head>
       ...
      <meta property="og:description" content="Welcome to Trap City, the place where you can find the hottest trap music. Trap City is founded in 12 Oct. 2012 by Gabriel Isik &amp; Idmon Yildiz">
...
```

In fact, it have quite a lot of information in it.

karyogamy:
Cool, I suppose using channel info for subscription fragment is workable then.

However, I still think showing channel stats (in a list of channels) offers little help to the user who already has subscribed to that channel. 

The only for reason doing this, from how I see it, is being able to reuse the existing channel item.

theScrabi:
`However, I still think showing channel stats (in a list of channels) offers little help to the user who already has subscribed to that channel.`
Well I'm concerned about UI consistency. Basically if we already have a channel item we may not need a second one.

app/build.gradle
```diff
@@ -1,5 +1,12 @@
 apply plugin: 'com.android.application'
 
+allprojects {
+    repositories {
+        jcenter()
+        maven { url 'https://maven.google.com' }
```
theScrabi: Does maven.google.com host OpenSource software? We need to enjure this is supported by f-droid.

theScrabi: Acording to [that this](https://forum.f-droid.org/t/maven-google-com/519) maven.google.com seems not to be supported by fdroid.

theScrabi:
Which of the newly introduced dependencies are part of maven.google.com, and are also necessary?

app/build.gradle
karyogamy: Unfortunately, Room persistence (and all of android architecture component) is only available from that repo at the moment. 

As you have noticed [here](https://gitlab.com/fdroid/fdroidserver/merge_requests/275), changes required to use maven.google.com on f-droid is already merged, and waiting to become available on the server.  

I say we wait for a bit to see what happens.

karyogamy:
@theScrabi 

The latest release (0.8) of F-Droid server has included maven.google.com as a permitted repo, see [here](https://gitlab.com/fdroid/fdroidserver/blob/0.8/fdroidserver/scanner.py#L90).
However, I'm not entirely sure if their production server is running 0.8 right now as I have only checked the tag versions for release.

Also, since Room Persistence now depends on Android support library version 26, I've taken the liberty to bump the target version to 26 as well.

theScrabi:
@karyogamy yea that's a good idea to do so :)

mauriciocolli:
Some things I think we could improve:
- Save the feed items into the database (tell me your opinion on this)
   - Check/fetch them periodically or update it whenever the user access that channel
   - I guess we'd do that to playlists anyway, and the feed is very similar to a playlist (not to say equal)
- Show more items (5~10) at the same time (from the same channel)
   - Fetching a entire info for just one stream seems a bit unnecessary
   - I think the problem with that would be keeping track for the load more feature (I think that's what blocked you of doing that in the first place?)
- Improve the empty states

---

Other problem is that we are saving the info list into the bundle (risk of `TransactionTooLargeException`, there are already reports in Sentry), that's happening to others fragments too, and in my refactor I already fixed, just waiting to this pull request and the other to be merged (so no need to fix in this pr).

Regarding the download size, I just found out that the [feed xml (rss)](https://www.youtube.com/feeds/videos.xml?channel_id=UC65afEgL62PGFWXY7n6CUbA) is much lighter than fetching the whole channel (25KB vs 160KB), but the downside is that it would only work in YouTube and that makes not workable...

karyogamy:
- I've thought about storing feed items in database as well though it just seems to defeat the purpose of a feed:
  * If we update periodically, the feed will often be too outdated if the update window is too long (say, longer than the lifecycle of a single app use). Similar problem if we update it only when the user accesses the channel page.
  * If we update to quickly, then we would have to download too much since now user loses control of how much feed they want to load. Similar problem if we use an expiry for each channel.
  * Maybe we can cache channels that are requested by the user, so the update window is just the app lifecycle and there is no need to store it to the database. How do you think about this?

- Right now the strategy for feed item is just popping the top of the channel info's stream list. 
  * We can easily change that number so we get a random list of the channel's newest streams. Though, these streams would be grouped together on the feed list.
  * With the cache change I mentioned above, this will likely no longer be an issue since we can just pull streams out of the cached ChannelInfo.

- Agreed. It might be a good idea to make a generic empty list image, since #626 is already using such feature.

------------------

Uh oh, I didn't realize this might happen and just looked at the method counts. 
Looks like RxJava is responsile for adding about 10K methods. Guess it was a bad idea after all :P

mauriciocolli:
In my refactor, I've already updated the workers (replace them) with RxJava and I've included a cache for all `Info` types (with the refactor of the extractor, that became possible/easier), so in the application's lifecycle, we already have a cache.

Regarding the load more feature, I misunderstood a little bit, I thought that more items of the existing channels were being loaded (right now the load more are used to get channels that didn't had some video in the list, right?).  And a issue with it, as you already talked about, is that the list would be pretty much random (if we sort it, then the videos would jump around and that would be confusing to the user). 

A solution would be fetch all the channels (showing to the user something like "Fetching channel 1/50"), sort them (by upload date, for example), and then show to the user. But that's the complicate part, we'd have to do this in every start if we don't save it somewhere (and that have some downsides as you mentioned, but that's how rss feeds sort of work).

So doing a feed (at least a sorted one) is pretty complicated without api/server control, and currently, we are making a sort of random feed, so, my question is (just to make it clear):

The main's goal of NewPipe's feed would be just to get random (but recent, except for channels that don't posted videos recently) videos from their subscriptions?

---

A problem to keep in mind is that the `onScrollToBottom` is fired multiple times, thus calling more than once `requestFeed(count)` ...

karyogamy:
With how feed works currently, it is probably more fit to call it "What's new" than "Feed". I think it is really a matter of which one is more valuable to the user: a random or periodically updated recent list of videos.

If we need to implement a true feed, then it would be best to keep track of an update expiry and have a setting to let the user decide how long the update window is. Then have a service that that pulls the new streams once the expiry is reached. With this we can simply pull the entire feed when there is an update since they are all stored in db.

Though, if we want to do a merge soon, I say we can just change the tab name to something else and work on a true feed after the merge (with less dependencies change).

What do you think?

------------------
Thanks for pointing out the on scroll problem. Will fix.

mauriciocolli:
Yeah, I would like a feed that is stored on the device (similar to rss) and in addition to those ways to update it, we can implement a way to force refresh if the users want to (like pull to refresh). 

Though, I guess the playlists have more priority than implementing that (is a really wanted feature, even for me).

If we change it to "What's new" I think it'd be more appropriate to change the place of it to something like this:



That way we wouldn't force-download something relatively big in every start (~1MB with the current initial items, but if we have a tablet, we'd need more for sure) and I know we could make the feed load only when the tab is clicked/selected, but I don't know if a "what's new" fits in the design of the start screen (and with the upcoming playlists, we will add a new tab, so...).

Regarding the items, I don't think we need to show the description (just the title and subscribers count) and they could be a little smaller (like in the screenshot).

karyogamy:
Feed fragment is now merged into the Subscription tab as a subitem and the initial load now fetches for 16 items. 
However, I've opted to stay with the old item layout for the channels since unnecessarily complicates the info item building as seen [here](https://github.com/TeamNewPipe/NewPipe/pull/620/commits/7d001a1ffe26fb8de6016383fac2fde3bc32c5b4#diff-9dac7eb546b86799d6b7be14cde0ce52R116). Maybe later we can create a layer over the existing info item types so we decouple UI item types from extractor item types.

Also, I looked a bit into replacing the current feed with a real feed. It seems that the upload date fetched by Youtube extractor is in the form of rounded periods (e.g. "2 years ago", "10 minutes ago"). This can cause problems when trying to sort by date since other services may not use the same format (e.g. Soundcloud). To solve this, we will need to parse the the date inside the extractor and store an approximate timestamp for use later. What do you think about this?

I agree with the playlist feature should be the top priority right now. Let me know if there is an active development I can help with, or if I should start from scratch.

wb9688:
@karyogamy: In my SoundCloud service I'm actually converting the real date as returned by SoundCloud to what you call a rounded period to be consistent with the YouTube service.

mauriciocolli:
There's no need to convert it in the extractor, actually, is a bad idea as we can't change according with the language of the user (that translation/conversion has to happen in NewPipe), that's a thing we have to fix in Youtube (and I don't know if we can, as it is YouTube itself that shows it in that format, but I'm trying to find a way).

wb9688:
@mauriciocolli: But we'll need a single format for time, right?

mauriciocolli:
Yes, and the [standard](https://xkcd.com/1179/) format would be the [ISO_8601](https://en.wikipedia.org/wiki/ISO_8601) (that SoundCloud uses).

karyogamy:
@mauriciocolli

To deal with those search item timestamps, how about introducing Joda Time and use it to parse dates from all services to the standard format?
For Youtube videos, I suppose the best we can do is to get the timestamp when the video is pulled, parse the upload date into a Period which we can convert to milliseconds and use them to approximate the initial upload date (until the user actually opens the video and gets the actual upload date).

@wb9688

Good call, thanks. Only the debug build needs to be fixed since it is not using proguard.

mauriciocolli:
That's an option, I'll think about it.

Regarding the proguard, isn't it disabled even on release build? (at least minify is disabled)

But enabling it have some good results (very few configurations, but of course, I think we'll have to keep some libraries (like rhino that uses some reflection) and it will rise up a little these numbers):

mauriciocolli:
I think we can merge it now.

And with the introduction of a database, we want to test the app a little before we release, so @theScrabi, **don't release** just yet (another reason is the other pull request too #626).

wb9688:
@mauriciocolli: And what about SoundCloud support? Btw when will you merge
your NewPipe refactor (which uses the latest NewPipeExtractor)?

mauriciocolli:
@wb9688 I will merge the open pull requests first.

And SoundCloud is already compatible (at least in theory) as we abstract the implementation details from the resulting data objects, the specifics (like search) we have to decide how we are going to implement it yet (we are discussing in the other issue).

wb9688:
@mauriciocolli: I meant to ask if we'll also wait until we've added support
for using other services in NewPipe for the next release.

mauriciocolli:
@wb9688 If we decide how to implement it in a reasonable period of time, then it'll get included (I hope it does).

wb9688:
@mauriciocolli: I hope it too. Design 2 isn't that hard to implement I
think, but I'll have to wait for your PR first. In the mean time I'll just
write the tests for the SoundCloud service in NewPipeExtractor.

(Is there a reason YouTube is called Youtube in NewPipe? I just followed
that example, so SoundCloud is called Soundcloud, but I don't know if that
was intentional.)

theScrabi:
@mauriciocolli you think searching multiple services at once is something that can work?
I thought such a search might take to much resources, and take to long to be usable.

If we don't yet know I'd not yet implement multiple search right from the start.

mauriciocolli:
We could fire the requests simultaneously, I'll do some tests to see if it's workable (we'd have to limit how much services to search at the same time if we reach a big number of services).

One thing though, how would we show the results from these searches...

theScrabi:
Well that's another problem. Just throwing the data at the user might also not help much. We might need to figure out how to display it, but we can see about that if we know that multisearch actually works.

Issue #570, opened 2017-05-19
Background audio playback resumes by itself


chrisvella:
- [X] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [X] I checked if the issue/feature exists in the latest version.

First of all thank you for the great program. I use it every day.

I have an issue with NewPipe resuming background audio playback on various events i.e. getting a message. This can be pretty embarrassing haha

I am using version 0.9.6 from fdroid.

mauriciocolli:
Yeah I already fixed it, just wait for the next version.

Also this a duplicate of #519.

Issue #519, opened 2017-04-27
paused background audio/pop-up video re-enables itself after another audio process is terminated


oncle-anatole:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

Hello,

Since 0.9.3 it seems that paused background audio or paused poped-up video re-enables themselves after another process involving the audio stack is terminated.
In concrete terms: 
- I start playing a video on background audio with newpipe, I pause it, then I play a song on my stock music player, I eventually close my music player (terminating the process, not letting it run on background) > NewPipe's background audio immediatly re-enables itself
- another example : with a paused background audio, I end a call > ditto, NewPipe re-enables itself
- With my phone's notification profile set to ring on notification reception, paused NewPipe background audio re-enables itself just after my phone receives an audible notification

This is happening on a 6.0.1 based OEM rom (ZTE's MiFavor).

theScrabi:
Confirmed

PR #624, opened 2017-07-11
Added option to ignore permanent audio focus loss

karyogamy:
Ignoring non-transient audio focus loss, so multiple streams can play at the same time.

Also added the option to resume background audio on audio focus regain (say, after a phone call interruption). However, this option only works on background audio and popup/main video player will not resume.

Perhaps I should categorize these settings under experimental rather than video and audio.

Let me know what you guys think.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -47,6 +47,8 @@ import com.google.android.exoplayer2.util.Util;
 import com.nostra13.universalimageloader.core.ImageLoader;
 import com.nostra13.universalimageloader.core.listener.SimpleImageLoadingListener;
 
+import org.schabi.newpipe.R;
+
 import java.io.File;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
@@ -307,6 +309,7 @@ public abstract class BasePlayer implements ExoPlayer.EventListener, AudioManage
     public void onAudioFocusChange(int focusChange) {
         if (DEBUG) Log.d(TAG, "onAudioFocusChange() called with: focusChange = [" + focusChange + "]");
         if (simpleExoPlayer == null) return;
+
         switch (focusChange) {
             case AudioManager.AUDIOFOCUS_GAIN:
                 onAudioFocusGain();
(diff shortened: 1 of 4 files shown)


karyogamy:
@theScrabi Can you please enable retrying on Travis? Seems like it glitched. I ran the tests on local and the one failed on CI is passing.

theScrabi:
It's not a glitch we don't moc our data, and therfore it fails from time to time.

mauriciocolli:
Restarted the build and it was successful.

The option to resume the playing in focus gain was the default, but due to some issues (reports #519, #570) it was removed, but as you added options to enable/disable this behavior, I think it's good (the default should be disable, as you did).

> this option only works on background audio and popup/main video player will not resume

It affects all the players as you modified the base player.

theScrabi:
@karyogamy 
> It affects all the players as you modified the base player.
Is this acceptable behavior?

karyogamy:
@theScrabi 

That was my intent originally. However, it does seem somewhat unnecessary.

What do you think about this? Should we only ignore focus loss on the background player?

theScrabi:
That's a good question @mauriciocolli what would you do?

mauriciocolli:
Thinking more about this, is it really worth the trouble?

For example, would anyone want the music blasting as soon as a call is finished? (now it does even if it was paused when focus was lost) There are even users that created issues about this. (I posted previously)

And are there use cases for multiple players playing together?

karyogamy:
@mauriciocolli 

Yeah, you are right. Now that I thought about it again, forcing permanent audio focus probably work best only on apps that controls a single player, whereas NewPipe has three. 

Though, with permanent audio focus, we can do things such as overlaying silent videos with external audio, or having multiple audio playing at the same time.

I will probably remove the audio focus feature and keep the resuming after interruption checkbox, since it is quite useful for headphone users. What do you think?

mauriciocolli:
Yes, I agree with the resume option.

> For e... **(now it does even if it was paused when focus was lost)** There a...

I knew this was strange (I said it previously), and after some digging I found a bug (fixed in https://github.com/TeamNewPipe/NewPipe/commit/f7b322da49759d96d54ebfd38bd1b7a1672c6f78) and now it's working correctly.

PR #626, opened 2017-07-12
Add search and watch history

coffeemakr:
Added a search and watch history. The UI is not that beautiful and I'm open to ideas.

I had to make two changes in order to allow starting a search from another activity:
 * Make MainActivity a single task so the `onNewIntent` method is called.
 * Set the `wasLoading` in `SearchFragment` (and keep it set until `onResume`). 

The following things may need a fix:

- [ ] The history disabled screen is not that good ¯\\\_(ツ)\_/¯
- [ ] Maybe the history could be linked in the settings instead of the main menu.
- [ ] Currently only the date is shown for each history entry
- [x] The main activity returns to the "blank screen" when back is pressed after clicking on a history entry.
- [ ] A switch to enable/disable history may be nice in the history view
- [x] The history database could be integrated into the AppDatabase proposed by #620 
- [ ] Make search suggestions based on the history
- [x] Enable history by default

Regards


b1k3rdude:
Where is the above history supposed to appear, on the main/first page when you open the app..?

coffeemakr:
It's currently implemented as a new activity which can be reached by clicking the entry "History" in the "main menu" (where the settings can be opened)

karyogamy:
Looked at the app, works pretty well.

Couple of things I saw:
- Searched items cannot be deleted individually (Maybe delete on swipe?)
- Watched history does not include background audio plays

coffeemakr:
@karyogamy Thanks for pointing that out. I've implemented the two functionalities.

The drawback of swiping to remove the entries is that it conflicts with the swiping of the tabs. But I like it that way.

theScrabi:
Why did you make history opt out? Can't you make it opt in?

coffeemakr:
@theScrabi As far as I see it the history is opt-in:
 *  [`android:defaultValue="false"` in the Preferences](https://github.com/TeamNewPipe/NewPipe/pull/626/files#diff-e4598b574427faa1bad2a70176c8e512R165)
* [`sharedPreferences.getBoolean(..., false)`](https://github.com/TeamNewPipe/NewPipe/pull/626/files#diff-c8979777d606ccc9f805b5d2357f35c1R237)

Did you experience something else?

I'm looking forward to merge the database once it exists.

It is possible to use stream_items. Where would you put the watched date? And would you store the stream information in the database or reload it in the search fragment?

theScrabi:
Sorry i ment it the other way around. It's inactive when you use it the first time.

mauriciocolli:
As the other pull request is merged, now we can merge the database.

mauriciocolli:
Some things to fix/improve:
- Use RxJava for queries/insert the database 
    - Now we are getting exceptions because it don't allow doing queries in the main thread by default
- Don't repeat entries (if they are the last one, of course)
    - If you click watch in background, popup and the normal player, all of them will produce a entry in the history (which will appear that they are duplicated)
- Sort the history by most recent to the least recent
    - Lucky for us, it's just a parameter in the RecyclerView's layout manager
- Make the entities extend the BasicDAO, introduced by the other pull request
    - It includes some useful methods, like `listByService`, which will be very useful when we create a option to only show entries of a given service, for example.
- The user can't go back to the history after clicking in an item
    - This happens because we are using a separate activity, so I don't know if this is fixable (without getting too much complex)
    - We can ignore this issue for now

mauriciocolli:
Ok, it's good as it is right now.

The others issues/improvements have less priority and can wait, this pull request was delayed for too long...

theScrabi:
@coffeemakr I tested it for quite a while now. I like it, tho I found several things that should be changed/fixed:
1. When scrolling from left to right, it is unclear to the UI weather you want to delete an entry, or you want to just scroll to the other screen.
2. There is a padding in the scrolling area, please remove this padding for the video history screen. For the search history screen it would be nice if the padding was inside the items not inside the scroll area as well.
3. __(Optional)__ If you delete something a snackbar should pop up, so you can revert your delete change.

mauriciocolli:
@theScrabi coincidentally, I already fixed the points 2 and 3 (the undo option and the padding), the other point we have to discuss (like long click and swipe? I don't know if this is very user discoverable, but who knows).

theScrabi:
All right :) we don't actually have to use a long press in order to make it sweepable, but we have to change something, since I've accidentally deleted entries when I just wanted to switch the page.

mauriciocolli:
Yeah, that's the main reason why I implemented the undo.

karyogamy:
Once playlist feature is ready, we can treat watch history as just another playlist (like Youtube), which is shown under the playlist tab. This way, history activity can just be a single page that shows all the search history.

_Sent from my Sony F8331 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

mauriciocolli:
Where we put the history was already being discussed, and I was thinking the same too.

One thing though, would "Playlists" be an intuitive name to the history be in? For example, Youtube calls it "library", we could think of a more general name too, or even just use an icon.

karyogamy:
How about "Collections"?

I'm also fine with just copying Youtube and calling it "library".

PR #633, opened 2017-07-19
Prevent "time links" from beeing clicked

### LINKED ISSUE #632: Application version 0.9.10 (and previous too) crashes when choosing certain position from description
as in title — application can't handle for example track listing with time referrers from description section — it crashes instantly when tapping it.

coffeemakr:
Added a line to only link web URLs. Before this the links for time were linked and created an Intent with the data "#" which caused an `ActivityNotFoundException` (Fixes #632). 

I think it would be nice to implement support for times but with the current links it wont work.


theScrabi:
Hm I see, we had to parse the html/JS in order to allow clicking the timestamps.

Issue #542, opened 2017-05-08
Option to save file names with special characters


lightonflux:
NewPipe replaces less used unicode characters like "►" with a "_" in the file names of downloaded files. It would be nice if there was an option to save the files with their full original title.

I would prefer proper names, i know that most of my apps can handle them, so it would be nice to have.

theScrabi:
Hm ok :)

PR #636, opened 2017-07-24
Add renaming filename preferences

coffeemakr:
Added the option to specify the replacement character and the allowed file characters in the preferences.

Another option would be to test the filename (like the old method's for-loop) but with actual checks in between the replacements.


mauriciocolli:
I take the liberty to fix some issues (default value and some string escape) and now it's ready to merge.
Sorry for the delay.

Closes #542

PR #665, opened 2017-08-30
E-mail, M4A, History, YouTube, GitHub, SHA-1

comradekingu:
As per https://en.wikipedia.org/wiki/SHA-1 and https://en.wikipedia.org/wiki/MPEG-4_Part_14
It is/was (?) called "History" on YouTube https://www.youtube.com/watch?v=CKwMeqYia3U

Some of these are intended to make for fewer errors in translation.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -235,7 +235,7 @@
     <string name="tab_contributors">Contributors</string>
     <string name="tab_licenses">Licenses</string>
     <string name="github_url" translatable="false">https://github.com/TeamNewPipe/NewPipe</string>
-    <string name="app_description">A free lightweight YouTube frontend for Android.</string>
+    <string name="app_description">A libre lightweight YouTube frontend for Android.</string>
     <string name="app_license" translatable="false">NewPipe is Free Software: You can use, study share and improve it at your will. Specifically you can redistribute and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.</string>
     <string name="view_on_github">View on GitHub</string>
     <string name="app_license_title">NewPipe\'s License</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -52,15 +52,15 @@
     <string name="show_higher_resolutions_title">Show higher resolutions</string>
     <string name="show_higher_resolutions_summary">Only some devices support playing 2K/4K videos</string>
     <string name="play_with_kodi_title">Play with Kodi</string>
-    <string name="kore_not_found">Kore app not found. Install Kore?</string>
+    <string name="kore_not_found">Kore app not found. Install it?</string>
     <string name="fdroid_kore_url" translatable="false">https://f-droid.org/repository/browse/?fdfilter=Kore&amp;fdid=org.xbmc.kore</string>
     <string name="show_play_with_kodi_title">Show \"Play with Kodi\" option</string>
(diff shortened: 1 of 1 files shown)


mauriciocolli:
Regarding the history, I think we should stick with the specific naming to not have confusion. You can check the naming of the new layout of YouTube (although we don't make the same choices for everything and we don't intend to):

<image src="https://user-images.githubusercontent.com/13389907/29897197-2582a60c-8db6-11e7-9063-f5905afe2b10.png" width="500"/>

I don't know about the "libre", for now, we could let it as is.

The others (including the libre in WebM), please, submit them in our [Weblate service](https://hosted.weblate.org/projects/newpipe/strings/en), you can easily log in with your GitHub account and make the changes.

comradekingu:
@mauriciocolli I changed to "History" anyway, seeing as it is used elsewhere in the app.
Can't change SHA-1 since it is untranslatable.

mauriciocolli:
I think we'll have to change the naming when we add another service in NewPipe anyway (like SoundCloud that only have audio) so it's okay, I'll add a more descriptive summary to help the user in the mean time.

comradekingu:
@mauriciocolli "content" is used elsewhere now

mauriciocolli:
Can you explain what do you mean by that? Which content?

app/src/main/res/values/strings.xml
```diff
@@ -101,7 +101,7 @@
     <string name="play_btn_text">Play</string>
     <string name="content">Content</string>
     <string name="show_age_restricted_content_title">Show age restricted content</string>
```
comradekingu: Here. Media content to be precise.

app/src/main/res/values/strings.xml
```diff
@@ -101,7 +101,7 @@
...
-    <string name="video_is_age_restricted">Video is Age Restricted. Enable age restricted videos in the settings first.</string>
+    <string name="video_is_age_restricted">Age Restricted Video. Allowing such material is possible from Settings.</string>
```
comradekingu: This says video.

app/src/main/res/values/strings.xml
```diff
@@ -52,15 +52,15 @@
     <string name="show_higher_resolutions_title">Show higher resolutions</string>
     <string name="show_higher_resolutions_summary">Only some devices support playing 2K/4K videos</string>
     <string name="play_with_kodi_title">Play with Kodi</string>
-    <string name="kore_not_found">Kore app not found. Install Kore?</string>
+    <string name="kore_not_found">Kore app not found. Install it?</string>
     <string name="fdroid_kore_url" translatable="false">https://f-droid.org/repository/browse/?fdfilter=Kore&amp;fdid=org.xbmc.kore</string>
     <string name="show_play_with_kodi_title">Show \"Play with Kodi\" option</string>
     <string name="show_play_with_kodi_summary">Display an option to play a video via Kodi media center</string>
```
comradekingu: This also says "video"

mauriciocolli:
Oh, are you talking about the inconsistencies between video and audio, right?
> we'll have to change the naming _**`when`**_ we add another service

But I still don't get the "content" one, what's wrong with it?

comradekingu:
@mauriciocolli Content is good. As the saying gooes, "Content is king"

Seeing as "content" is used now, along with "video", for purposes where they are currently both video, and "content" is future proof, going with "video" in the interim serves no other purpose other than to be inconsistent.

mauriciocolli:
@comradekingu I don't know if we'll be using custom exception strings for each service (like some strings have "video" in it and are from YouTube specifically) or we'll be using general ones, so I think this can wait util we add the other service.

PR #669, opened 2017-09-03
Update extractor and refactored NewPipe

mauriciocolli:
- Very basic playlist support
    - It only supports list the videos from the playlists at this moment
- Fix bug on `InfoItemBuilder`
    - View was being set to invisible/gone (missing information) but never to visible again and when the view is recycled, it was invisible due to the previous call
- Improve `InfoListAdapter`
    - Nicer animation for inserting data
- Remove unused NetCipher library (tor support?)
    - I let the strings because maybe in the future we implement it
- Remove `ImageErrorLoadingListener`
    - There's no need to do that, for example, when the internet of a user went down in a middle of a video detail loading (or even before), it was throwing a exception and asking him to report (of course, it's just a useless report for us)
    - Instead of doing this, configure the image loader to show a image when the request fail (showImageOnFail(...))
- Refactor fragments
    - No need to nullify every view/variable in `onDestroyView`
    - Create new Base classes to remove some duplicate code
- Refactor settings
    - Split settings into different screens, making it easier to use
- Improved resolution and format selector helper (default resolution and format)
    - Write more tests to make it more robust
- Introduced RxJava (much easier and less error prone than managing the threads by ourselves)
- Fix `TransactionTooLargeException`
    - When a configuration change was occurring (e.g. rotating the phone) and the stack of history of videos or some list of video items was too big, we were receiving `TransactionTooLargeException` when trying to save that to `savedInstance` (Bundle)
    - Implemented a state saver that save it to a file in the cache folder or in-memory if it's just a configuration change (can save anything to it, but it's good to keep it to a bare minimum) 
- Improve layouts
    - Reduce the size of the uploader avatar on the video detail
    - Fix disabled likes/dislikes in video detail
    - Animations in Subscribe button (appearing and color when clicked)
- Fix memory leak in players
- History fragments
    - Undo button in history fragments
        - I find myself deleting items often enough to be annoying, as it's more natural to slide in a view pager, @coffeemakr do you think we have to change that? 
        (or even block some direction to delete items, like to go from the first fragment to the second you do a left swipe, so only allow to delete with a right one)
    - Fix bug in image loading in watched history
    - Fix view pager bug in rotation
         - When the view was rotated the sparse array was empty because the fragments were retained directly into the fragment manager. Extending `FragmentStatePagerAdapter` did the trick (have a "reliable" method to get the current fragment).
    - Save state on rotation/configuration change
- Subscription fragments
    - Feed refactored, now it loads until it covers the screen and automatically load more when the user scroll to the bottom
    - @karyogamy what do you think about updating the subscription every time a channel is loaded in the feed (performance wise)? I didn't included yet in this pr.
- Fix recreate call in the main activity that is needed for theme change
    - https://stackoverflow.com/questions/10844112/runtimeexception-performing-pause-of-activity-that-is-not-resumed
- Various minor fixes...

Still isn't ready for production, we have some bugs in the extractor and the translations are not yet complete (some are very incomplete), so the release will have to wait a little (also I want to test a bit more these changes).

Some `TODO`s (that I'll eventually fix, but are here just for the record):

- [ ] No way to tell if there's more videos in the search
- [ ] Fix playlist url id handlers (some playlists' id have different length of what's being accepted right now)
- [ ] Fix empty playlists
- [ ] Improve player error handling (playing a 403 stream for example)
- [ ] Mixed content and UI languages

*PS:* A trick part of keeping track of which channel we are subscribed to is that there are more than one url pointing to it, for example, Youtube have /user and /channel urls, this could be a problem...


wb9688:
Next time you make such big changes, could you please split it into
multiple commits?

About that PS: is there a way to let `cleanUrl()` always return the
/channel one?

coffeemakr:
> I find myself deleting items often enough to be annoying, as it's more natural to slide in a view pager, @coffeemakr do you think we have to change that?
(or even block some direction to delete items, like to go from the first fragment to the second you do a left swipe, so only allow to delete with a right one)

I agree that it's not intuitive and I don't think it will be better if swiping left on the left screen deletes the entry.

From [material design guide](https://material.io/guidelines/patterns/gestures.html#gestures-drag-swipe-or-fling-details):
>  Don’t use paging swipes when individual elements are swipeable.

So I would either disabling either the paging swipe or the item swipe. A long-press action is may be used for item deletion.

mauriciocolli:
@wb9688 it is possible when you are extracting a channel or video, in fact, it's already returning only /channel urls, the problem comes when there are lists involved (like search), then sometimes only the /user is available for each item (and it is not practical to fetch each page and get the clean url).

@coffeemakr so I think it's better to disabled the individual item swipe.

karyogamy:
Went through to changes, it's looking good. I like the refactoring on list fragments, this will make multi-service UI much less tedious to implement.

The reason I didn't update channel entities on feed fragment is because the subscription monitor. If we update the database when a change is spotted, RxRoom will observe it and send another event downstream. At which point, the fragment will try pulling channel data from Youtube again. If you happen to subscribe to a popular channel, then the subsciber count will always be changing and you can potentially get into an infinite loop.

This might be solved when the cache though, but it still waste some cycles due to the RxRoom subscription. To solve this, you can just convert the subscription monitor to a Single in the feed fragment, now that I don't see why we would need to constantly monitor the database in the feed.

As for youtube channels, I think /user identifiers are better for persistence as they are not UUIDs (like /channel) and are more likely used by the public (thus less likely to change). But since Youtube's sidebar subscription uses /channel, it might be a good idea to store it as a extra backup url.

mauriciocolli:
@karyogamy But I think we we are good sticking with the `/channel` one, as all channels have it and only some have the `/user`. And like your suggestion, we could add a array called `aliases` that contain one or more alternative urls.

wb9688:
Finally - thanks!

@theScrabi: Have we already decided on the design for selecting a service?
If so, I'll start implementing that.

theScrabi:
I thought so.

I would have wanted the [second design](https://user-images.githubusercontent.com/17365767/28756763-072e3a80-7575-11e7-809c-e655a608861c.jpg) created by @TobiGr.

mauriciocolli:
@theScrabi please, don't copy this verbatim and put into the changelog, there's too much info and it is redundant with what is already there.

theScrabi:
@mauriciocolli Great work. I like all of it :)

Although I need to agree to @wb9688. In the future we should try to apply smaller changes again.
I made the experience that constantly releasing versions with smaller changes more often, is better than releasing versions with big changes, but with a huge timespan between each release.

Also I'm sorry I almost broke your release :sob:

theScrabi:
Tho I have to admit, there are some annoying bugs I noticed so far:

 - [ ] Videos that reqire encrypting the url don't work anymore. Examples [1](https://www.youtube.com/watch?v=3tmd-ClpJxA), [2](https://www.youtube.com/watch?v=iw4oRym4HSM)
 - [ ] Age restricted content does not get blocked when it's disabled. [Example](https://www.youtube.com/watch?v=yy-iB9XPeeg)

mauriciocolli:
>  Videos that require decrypting the url don't work anymore

You could open an issue in the extractor reporting this bug.

> Age restricted content does not get blocked when it's disabled

This doesn't work for older versions of NewPipe too (not the result of this pull request), fortunately, it can be solved by simply checking the `Info`'s age limit.

PR #692, opened 2017-09-13
Open youtube links directly from search (fixes #35)

### LINKED ISSUE #35: Support direct youtube links in the search
It'd be nice if pasting a YouTube URL (especially the m.YouTube.com-style URLs would bring you to that specific video in the app.

Nutomic:
Pretty straightforward, the regex should match all URLs like `www.youtube.com`, `m.youtube.com` and `youtu.be`. It might also open some things like playlists or channels that aren't supported, not sure about that. Also I'm not sure what do when `NavigationHelper.openByLink()` throws an exception.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -43,9 +43,8 @@ public class RouterActivity extends AppCompatActivity {
     }
 
     protected void handleUrl(String url) {
-        try {
-            NavigationHelper.openByLink(this, url);
-        } catch (Exception e) {
+        boolean success = NavigationHelper.openByLink(this, url);
+        if (!success) {
             Toast.makeText(this, R.string.url_not_supported_toast, Toast.LENGTH_LONG).show();
         }
 
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
@@ -48,6 +48,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 
 import icepick.State;
 import io.reactivex.Notification;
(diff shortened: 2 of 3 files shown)


mauriciocolli:
I think we should try a different approach, here some key things that we already have:
- We have a method called [`acceptUrl`](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/src/main/java/org/schabi/newpipe/extractor/UrlIdHandler.java#L36) that services implement it (example [YouTube](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/src/main/java/org/schabi/newpipe/extractor/services/youtube/YoutubeChannelUrlIdHandler.java#L53) and [SoundCloud](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/src/main/java/org/schabi/newpipe/extractor/services/soundcloud/SoundcloudPlaylistUrlIdHandler.java#L57))
    - The services implement this for channel, playlist, videos...
- A method that goes through all services and try to accept the url [(getServiceByUrl)](https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/src/main/java/org/schabi/newpipe/extractor/NewPipe.java#L74)
    - If it's not accepted by any service, it throws an exception

Nutomic:
So basically, I just need to call `NavigationHelper.openByLink()`, and if it throws an exception, run the search? Makes sense. The only problem is that it would throw an exception whenever the query is not a URL, which is not a good idea.

Nutomic:
Okay I implemented it like that. Btw catching `Exception` is a bad idea, as it will hide things like `NullPointerException`. I would suggest replacing that with specific exception handlers.

mauriciocolli:
I think we should, instead of using `Exception`, throw an `ExtractionException` and just catch for that, the rest should bubble up and ask the user to report.

And something to note, the method `cleanUrl` of some services clean their "dirtyUrl" by fetching some resource from the network, and calling the method directly will result in a `NetworkOnMainThreadException`. 

**PS**: Though I'm trying to improve that use of network in `cleanUrl`, it's a problem that's inherited from the old architecture (a very serious one).

**PPS**: And for now, we should let only the YouTube open the url directly, as the support of multiple services are being worked on yet.

Nutomic:
Agreed about the exceptions. 

But I'm not sure what you mean with the rest of your comment. Do you want me to change something in my PR? And if yes what?

mauriciocolli:
I just note an issue that `cleanUrl` has, but you don't need to change anything about that.

And some things to fix that I noticed:
- You cannot go back because of those intent flags
    - And even if they are removed, the search fragment have to be popped
- Opening services that are not YouTube (see previous post)

Nutomic:
So should I start it as a normal intent without those flags? And I don't see any way in NavigationHelper to open Youtube links only.

mauriciocolli:
You can check in the `getIntentLink`:
```java
if (service != ServiceList.YouTube.getService()) {
    throw new ExtractionException("Service not supported at the moment");
}
```
You can either return null or throw an exception, IMO throwing a exception is cleaner.

Nutomic:
Okay added that.

PR #695, opened 2017-09-15
WIP: show search history items in suggestions

Nutomic:
This PR shows previous search queries in the suggestions dropdown (see screenshot).

screenshot_20170915-154251

Unfortunately I'm getting a compile error in the database code, can someone help me figure out what's wrong there?
```
:app:compileDebugJavaWithJavac/home/felix/workspace/newpipe/app/src/main/java/org/schabi/newpipe/database/BasicDAO.java:34: error: A DAO method can be annotated with only one of the following:Insert,Delete,Query,Update
    Flowable<List<Entity>> getItemsForQuery(String query, int limit);
                           ^
/home/felix/workspace/newpipe/app/src/main/java/org/schabi/newpipe/database/subscription/SubscriptionDAO.java:17: error: Abstract method in DAO must be annotated with interface android.arch.persistence.room.Query AND interface android.arch.persistence.room.Insert
public interface SubscriptionDAO extends BasicDAO<SubscriptionEntity> {
       ^
/home/felix/workspace/newpipe/app/src/main/java/org/schabi/newpipe/database/history/dao/WatchHistoryDAO.java:18: error: Abstract method in DAO must be annotated with interface android.arch.persistence.room.Query AND interface android.arch.persistence.room.Insert
public interface WatchHistoryDAO extends HistoryDAO<WatchHistoryEntry> {
       ^
3 errors
 FAILED
```

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
+++ b/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
@@ -36,7 +36,6 @@ public interface SearchHistoryDAO extends HistoryDAO<SearchHistoryEntry> {
     @Override
     Flowable<List<SearchHistoryEntry>> listByService(int serviceId);
 
-    // TODO: should be LIKE :query% but that throws more errors
     @Query("SELECT * FROM " + TABLE_NAME + " WHERE " + SEARCH + " LIKE :query || '%'" + ORDER_BY_CREATION_DATE + " LIMIT :limit")
     Flowable<List<SearchHistoryEntry>> getItemsForQuery(String query, int limit);
 }
--- a/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
+++ b/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
@@ -37,6 +37,6 @@ public interface SearchHistoryDAO extends HistoryDAO<SearchHistoryEntry> {
     Flowable<List<SearchHistoryEntry>> listByService(int serviceId);
 
     // TODO: should be LIKE :query% but that throws more errors
-    @Query("SELECT * FROM " + TABLE_NAME + " WHERE " + SEARCH + " LIKE :query" + ORDER_BY_CREATION_DATE + " LIMIT :limit")
+    @Query("SELECT * FROM " + TABLE_NAME + " WHERE " + SEARCH + " LIKE :query || '%'" + ORDER_BY_CREATION_DATE + " LIMIT :limit")
     Flowable<List<SearchHistoryEntry>> getItemsForQuery(String query, int limit);
 }
--- a/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
+++ b/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
@@ -11,6 +11,7 @@ import io.reactivex.Flowable;
 
 import static org.schabi.newpipe.database.history.model.SearchHistoryEntry.CREATION_DATE;
 import static org.schabi.newpipe.database.history.model.SearchHistoryEntry.ID;
+import static org.schabi.newpipe.database.history.model.SearchHistoryEntry.SEARCH;
 import static org.schabi.newpipe.database.history.model.SearchHistoryEntry.SERVICE_ID;
 import static org.schabi.newpipe.database.history.model.SearchHistoryEntry.TABLE_NAME;
 
@@ -34,4 +35,8 @@ public interface SearchHistoryDAO extends HistoryDAO<SearchHistoryEntry> {
     @Query("SELECT * FROM " + TABLE_NAME + " WHERE " + SERVICE_ID + " = :serviceId" + ORDER_BY_CREATION_DATE)
     @Override
     Flowable<List<SearchHistoryEntry>> listByService(int serviceId);
+
+    // TODO: should be LIKE :query% but that throws more errors
+    @Query("SELECT * FROM " + TABLE_NAME + " WHERE " + SEARCH + " LIKE :query" + ORDER_BY_CREATION_DATE + " LIMIT :limit")
+    Flowable<List<SearchHistoryEntry>> getItemsForQuery(String query, int limit);
 }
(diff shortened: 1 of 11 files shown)


coffeemakr:
Hi

I've checked the code and the problem is the following:

You've added the method `getItemsForQuery` to the `BasicDAO` with is the base interface for all DAOs. Since this method is only overriden (and annotated) in the `SearchHistoryDAO`, the "Room annotation processor" want to implement this method in all DAOs but doesn't know how.

Removing the method from `BasicDAO` solves this problem:
 
```diff
diff --git a/app/src/main/java/org/schabi/newpipe/database/BasicDAO.java b/app/src/main/java/org/schabi/newpipe/database/BasicDAO.java
index 8d708f46..03a94508 100644
--- a/app/src/main/java/org/schabi/newpipe/database/BasicDAO.java
+++ b/app/src/main/java/org/schabi/newpipe/database/BasicDAO.java
@@ -31,8 +31,6 @@ public interface BasicDAO<Entity> {
 
     Flowable<List<Entity>> listByService(int serviceId);
 
-    Flowable<List<Entity>> getItemsForQuery(String query, int limit);
-
     /* Deletes */
     @Delete
     int delete(final Entity entity);
diff --git a/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java b/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
index 6692f920..a5e6d24b 100644
--- a/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
+++ b/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
@@ -36,9 +36,7 @@ public interface SearchHistoryDAO extends HistoryDAO<SearchHistoryEntry> {
     @Override
     Flowable<List<SearchHistoryEntry>> listByService(int serviceId);
 
-    // TODO: why does this not compile?
     // TODO: should be LIKE :query% but that throws more errors
     @Query("SELECT * FROM " + TABLE_NAME + " WHERE " + SEARCH + " LIKE :query" + ORDER_BY_CREATION_DATE + " LIMIT :limit")
-    @Override
     Flowable<List<SearchHistoryEntry>> getItemsForQuery(String query, int limit);
 }
```

Nutomic:
Ah thanks! I was trying to copy the other methods that were already there. But the error message was really confusing.

theScrabi:
I thought it was better if we used [this](https://storage.googleapis.com/material-icons/external-assets/v4/icons/zip/ic_history_black_24dp.zip) icon in front of the text, instead of changing the text color.

Nutomic:
Right that makes sense.

app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
```diff
@@ -34,4 +35,8 @@
     @Query("SELECT * FROM " + TABLE_NAME + " WHERE " + SERVICE_ID + " = :serviceId" + ORDER_BY_CREATION_DATE)
     @Override
     Flowable<List<SearchHistoryEntry>> listByService(int serviceId);
+
+    // TODO: should be LIKE :query% but that throws more errors
```
mauriciocolli: https://stackoverflow.com/a/44448566/7762724

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -469,13 +476,19 @@ public boolean test(@io.reactivex.annotations.NonNull String s) throws Exception
...
+                    suggestionListAdapter.clearAdapter();
                 }
                 // Only pass through if suggestions is enabled and the query length is equal or greater than THRESHOLD_SUGGESTION
                 return showSuggestions && lengthCheck;
             }
         };
 
+        // TODO: this isn't working correctly
+        if (searchQuery.length() < THRESHOLD_SUGGESTION) {
+            List<SearchHistoryEntry> historyItems = searchHistoryDAO.getItemsForQuery(searchQuery + "%", 10).blockingFirst();
```
mauriciocolli: Don't use `blockingFirst`, instead try to integrate this with the RxJava statements below.

app/src/main/java/org/schabi/newpipe/fragments/list/search/SuggestionListAdapter.java
```diff
@@ -47,25 +51,38 @@ public SuggestionListAdapter(Context context) {
     public void bindView(View view, Context context, Cursor cursor) {
         ViewHolder viewHolder = new ViewHolder(view);
         viewHolder.suggestionTitle.setText(cursor.getString(INDEX_TITLE));
+        boolean isHistoryItem = Boolean.valueOf(cursor.getString(INDEX_IS_HISTORY_ITEM));
+        int color = isHistoryItem ? android.R.color.holo_blue_dark : android.R.color.secondary_text_light;
+        viewHolder.suggestionTitle.setTextColor(ContextCompat.getColor(context, color));
+    }
+
+    public void clearAdapter() {
+        //changeCursor(new MatrixCursor(columns, 0));
```
mauriciocolli: It's not doing anything.

mauriciocolli:
Just to inform, we already have this icon and it is already themed, usage is: `?attr/history`.

@Nutomic can you change the target branch of this pull request to the `dev` one? (the option is in the "Edit" button next to the title)

theScrabi:
@mauriciolli Would it make sense to lock down master branch, so only pull request ares are accepted, no direct commits?

wb9688:
@theScrabi: Yeah, that's what I suggested. Btw it'd also be useful if you
state somewhere that PRs should be made to the dev branch. Can we also
define a code style? Now it's pretty messy IMHO.

Nutomic:
Okay suggestions/history should work correctly now. I'm not familiar with RxJava, so there might be a better way to implement this.

I originally thought that history items should also be shown if the query is empty, but that isn't directly supported. What do you think about that? ([stackoverflow link](https://stackoverflow.com/q/2126717/1837158)) Related to that, the suggestions dropdown opens when you click the search icon and closes immediately. So we need to decide how to fix this (either hide suggestions until at least one character is entered, or make sure suggestions don't go away).

mauriciocolli:
In the meantime since I last talk, I was testing some things on this, and to overcome some issues that you mentioned, I've replaced the suggestion dialog with a proper RecyclerView and an easier-to-use adapter that comes with it.



It turns out that the test worked fine, so what do you think? Is it better? And, if you agree, what should we do this with this pull request?

Nutomic:
@mauriciocolli Can you push that code somewhere so I can check it out?

mauriciocolli:
@Nutomic I've opened a pull request #748

Issue #646, opened 2017-08-02
Design and theme issues


TobiGr:
I have found some cases where the current theme is not applied to all elements and only the light theme is displayed.

- [X] The item / folder picker in the settings has only a light theme (this dialog is called when the user wants to change the video / audio download path).
- [X] The menu in the tool bar in the about activity has only a light theme too.
- [ ] The license dialog's content (=> the license's file) is a HTML file. Can we apply our themes to it, too? If yes I'd suggest to color the links red and maybe to not underline them.

I also found some inconsistent styles and views:
- [ ] While the link to the about activity exists in the main activity there is no link to it in the tool bar of the downloads activity.
<table width="100%">

<tr width="100%">
<td width="50%"><img src="https://user-images.githubusercontent.com/17365767/28891075-88dc0994-77c9-11e7-845f-2a2aad786cd4.png" /></td>
<td width="50%"><img src="https://user-images.githubusercontent.com/17365767/28891070-84c692ca-77c9-11e7-8d67-ee371fe29ff5.png" /></td>
</tr>

</table>

- [X] While the tool bar's text color is dark in the main, downloads and settings activities, it is white in the about activity when the light theme is applied. I think white looks better. What do you think?<br>Maybe we also should change the color of the border at the bottom of the fragment's titles (ABOUT/LICENSES) to a ligth gray wich is used for the main text in the light theme.

<table width="100%">
<tr width="100%">
<td><img src="https://user-images.githubusercontent.com/17365767/28890466-7a55b502-77c7-11e7-96a0-3eef38d775a5.png" /></td>
<td><img src="https://user-images.githubusercontent.com/17365767/28890461-76b0257c-77c7-11e7-8de6-9b12880032e5.png" /></td>
</tr>
</table>

- [ ] The license dialogs have no separators which separate the dialog's header and footer from the main content as it's done in the settings (e.g. at preferred content language setting).

mauriciocolli:
I fixed some of the issues (and you fixed the file picker with the pull request), so can you mark them as fixed (strike-through or some todo list)?

TobiGr:
@mauriciocolli I have looked a little in the license dialog thing and I can't find a good solution to add separators to the alert dialo because we  can't style the bottom section of it. Styling the license HTML files should be quite easy by reading the file and adding the stylesheet before passing it to the WebView. So a possible solution to add separators is either to set a costum title for the dialog (wich only adds one separator) or adding the separators to the HTML file.

PR #699, opened 2017-09-18
Support dark theme in file picker

TobiGr:
The title says, what this PR does. This was the fastest way of doing this. If you have a better/cleaner solution, I can implement this too.

Addresses #646

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/FilePickerActivityHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/FilePickerActivityHelper.java
@@ -1,0 +1,18 @@
+package org.schabi.newpipe.util;
+
+import android.os.Bundle;
+import org.schabi.newpipe.R;
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/schabi/newpipe/util/FilePickerActivityHelper.java
```diff
@@ -0,0 +1,18 @@
...
+
+import android.os.Bundle;
+import org.schabi.newpipe.R;
+
+public class FilePickerActivityHelper extends com.nononsenseapps.filepicker.FilePickerActivity {
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        String currentTheme = ThemeHelper.getSelectedTheme(this);
+        if(currentTheme.equals("light_theme")) {
```
mauriciocolli: You can use `ThemeHelper#isLightThemeSelected` instead of hard coding it.

TobiGr: Good catch. I fixed it.

PR #703, opened 2017-09-19
Spelling of file formats

comradekingu:
As per
https://en.wikipedia.org/wiki/MPEG-4_Part_14#.MP4_versus_.M4A
https://en.wikipedia.org/wiki/WebM
https://en.wikipedia.org/wiki/3GP_and_3G2

### DIFF
--- a/app/src/main/res/values/settings_keys.xml
+++ b/app/src/main/res/values/settings_keys.xml
@@ -51,7 +51,7 @@
     <string-array name="video_format_description_list" translatable="false">
         <item>MPEG-4</item>
         <item>WebM</item>
-        <item>3GPP</item>
+        <item>3GP</item>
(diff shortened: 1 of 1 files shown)


mauriciocolli:
Only the "3GPP → 3GP" change is valid, the others are keys for the settings and are not visible to the user (they were left lower case in purpose) and if we change that now, users will lose its configurations.

comradekingu:
@mauriciocolli Thought they might be. However there are quite a few "m4a" strings and such. Should be M4A or .m4a

mauriciocolli:
The commits are now messed up, it seems like @theScrabi squashed some commits.

Can you rebase it?

theScrabi:
Thanks. Sorry for that debase mess.

PR #711, opened 2017-09-19
Use google() in buildscript

Jawnnypoo:
Hey there, 

Using `google()` in the buildscript block is going to become needed for Android Studio 3.0, so might as well add it now!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ android:
     # The SDK version used to compile NewPipe
     - android-26
 
-    # Additional components
-    - extra-android-m2repository
-
 script: ./gradlew -Dorg.gradle.jvmargs=-Xmx1536m assembleDebug lintDebug testDebugUnitTest
 
 licenses:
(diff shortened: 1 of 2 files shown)


theScrabi:
Why that?

karyogamy:
Unfortunately, google() points to [dl.google.com/dl/android/maven2/](dl.google.com/dl/android/maven2/), which might not be [accepted](https://gitlab.com/fdroid/fdroidserver/blob/master/fdroidserver/scanner.py#L79) on F-droid.

mauriciocolli:
@karyogamy that's what I thought too, but it seems to be accepted according to this issue: 
https://gitlab.com/fdroid/fdroidserver/issues/358

Jawnnypoo:
classpath 'com.android.tools.build:gradle:3.0.0' and onward will be accessible through Google's maven server instead of the SDK manager. 

F-Droid supports it, we use it in LabCoat

PR #725, opened 2017-09-22
Add themes to license dialog

TobiGr:
This PR enables themes for the license dialogs.

Addresses #646

### DIFF
--- a/app/src/main/assets/gpl_2.html
+++ b/app/src/main/assets/gpl_2.html
@@ -15,9 +15,9 @@ Version 2, June 1991
 </p>
 
 <pre>
-Copyright (C) 1989, 1991 Free Software Foundation, Inc.  
-51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
-
+Copyright (C) 1989, 1991 Free Software Foundation, Inc.<br/>
+51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA<br/>
+<br/>
 Everyone is permitted to copy and distribute verbatim copies
 of this license document, but changing it is not allowed.
 </pre>
--- a/app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.java
@@ -1,0 +1,111 @@
+package org.schabi.newpipe.about;
+
+import android.content.Context;
+import android.content.DialogInterface;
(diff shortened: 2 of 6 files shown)


app/src/main/assets/gpl_2.html
```diff
@@ -15,9 +15,9 @@
 </p>
 
 <pre>
-Copyright (C) 1989, 1991 Free Software Foundation, Inc.  
-51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
-
+Copyright (C) 1989, 1991 Free Software Foundation, Inc.<br/>
```
coffeemakr: Are those line breaks necessary? IMO there is no need to add them since the text is inside a `<pre>`.

mauriciocolli:
Some things I noticed:
- UTF-8 text is not displaying correctly in the webview, and a quick test shows that the file is fetched correctly (check by logging right after it's read). 
    - Test on MPL-2 to see this character `“`.
- I/O in the main thread

app/src/main/assets/gpl_2.html
TobiGr: You are right, but I use `pre { white-space: pre-wrap; }` to keep the width of the body at 100% of the view. That's why I added the line breaks manually.

TobiGr:
> UTF-8 text is not displaying correctly in the webview, and a quick test shows that the file is fetched correctly (check by logging right after it's read).
 
@mauriciocolli Thanks for the stackoverflow link, that works.
I don't have enough time to fix the I/O thing this week, so I am going to do this next week as soon as I can.

theScrabi:
No problem

app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.java
```diff
@@ -0,0 +1,111 @@
...
+import android.os.AsyncTask;
+import android.support.v7.app.AlertDialog;
+import android.webkit.WebView;
+import org.schabi.newpipe.R;
+import org.schabi.newpipe.util.ThemeHelper;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+
+public class LicenseFragmentHelper extends AsyncTask<Object, Void, Integer> {
```
TobiGr: @mauriciocolli Should I make this class an inner class of `LicenseFragment`?

theScrabi:
this does it :)

PR #732, opened 2017-09-23
Add kiosk and main page

theScrabi:
# Prototype do not merge!!!

This will replage the Blank screen with a Kiosk page. At the moment it's Youtube "Trending".

Things that need to be done:
- [x] Make Items touchable (results in stack trace right now)
- [x] Remove back arrow
- [x] Add option to settings to bring "Black page back"
- [x] add a proper handling for current service in [main_fragment](https://github.com/TeamNewPipe/NewPipe/blob/trending/app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java#L30)
- [x] add proper error handling when creating [instance](https://github.com/TeamNewPipe/NewPipe/blob/trending/app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java#L130) of KioskFragment
- [x] Make type/name/id translatable
- [x] s/kisok/kiosk/g

This is what it looks like atm:
screenshot_1506181400


app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -18,11 +18,16 @@
...
+import org.schabi.newpipe.fragments.list.kisok.KioskFragment;
 import org.schabi.newpipe.fragments.subscription.SubscriptionFragment;
 import org.schabi.newpipe.util.NavigationHelper;
 
 public class MainFragment extends BaseFragment implements TabLayout.OnTabSelectedListener {
     private ViewPager viewPager;
+    private boolean showBlankTab = false;
+
+    //todo: FIX THIS URGENTLY
+    public int currentServiceId = 0; //for youtube
```
wb9688: If #693 is merged, you could do `Integer.parseInt(PreferenceManager.getDefaultSharedPreferences(getActivity()).getString("service", "0"))`.

app/src/main/java/org/schabi/newpipe/fragments/list/kisok/KioskFragment.java
```diff
@@ -0,0 +1,148 @@
...
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * NewPipe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with OpenHitboxStreams.  If not, see <http://www.gnu.org/licenses/>.
```
wb9688: I think you meant NewPipe instead of OpenHitboxStreams. ;)

theScrabi: Yea that's copypasta, I took the code from one of my other apps.

app/src/main/java/org/schabi/newpipe/fragments/list/kisok/KioskFragment.java
```diff
@@ -0,0 +1,148 @@
...
+        animateView(headerRootLayout, false, 200);
+        animateView(itemsList, false, 100);
+    }
+
+    @Override
+    public void handleResult(@NonNull final KioskInfo result) {
+        super.handleResult(result);
+
+        animateView(headerRootLayout, true, 100);
+        headerTitleView.setText(result.type);
```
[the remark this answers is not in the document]
theScrabi: You are right. Id/type/name was not meant to be displayed directly.
I'll add a todo.

app/src/main/java/org/schabi/newpipe/util/ExtractorHelper.java
```diff
@@ -116,11 +116,29 @@ public PlaylistInfo call() throws Exception {
...
+    public static Single<KioskInfo> getKioskInfo(final int serviceId, final String url, final String contentCountry, boolean forceLoad) {
+        return checkCache(forceLoad, serviceId, url, Single.fromCallable(new Callable<KioskInfo>() {
+            @Override
+            public KioskInfo call() throws Exception {
+                return KioskInfo.getInfo(NewPipe.getService(serviceId), url, contentCountry);
+            }
+        }));
+    }
+
+    public static Single<NextItemsResult> getMoreKisokItems(final int serviceId, final String url, final String nextStreamsUrl) {
```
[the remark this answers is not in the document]
theScrabi: thanks :)

app/src/main/java/org/schabi/newpipe/fragments/list/kisok/KioskFragment.java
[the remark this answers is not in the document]
wb9688: Yeah, I was thinking about it, but I couldn't find a better way than bunch of `if` statements, so do you have any better ideas?

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
wb9688: Yeah, but #693 isn't merged yet, so you'd have to wait for that.

wb9688:
Btw can you fix the merge conflicts?

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
theScrabi: hm, don't you think it was possible to introduc it anyway, I mean its no code that will colide i guess.

theScrabi:
Resolved conflict.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
wb9688: You mean merging #693? Well, that's possible, but then I'd at least add the correct intents for SoundCloud (can you do that? I know nothing about Android's intents xD). After that I can just make another PR for the new service layout.

app/src/main/java/org/schabi/newpipe/fragments/list/kisok/KioskFragment.java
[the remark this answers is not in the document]
theScrabi: I see. Isn't it possible to get some stringresource like R.string.<some_string> programmatically? Something like `Resource.getString("some_string");` If not than we might have to find another solution.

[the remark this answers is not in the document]
mauriciocolli: I thought in doing a "generify" and use the most common term as the id/type, because it is more convenient to use a generic term to cover a lot of cases. 

_Example:_ SoundCloud uses "New & Hot", YouTube's equivalent is "Trending".
In this case, "Trending" is the selected one (as it is more known).

_Example 2:_ A lot of services use "Top ##", this can easily be "generified" because only the number changes.

###### Though there's a downside to it, we lost the "uniqueness" of each service.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
theScrabi: No don't merge #693 sorry. No I ment adding that line `Integer.parseInt(PreferenceManager.getDefaultSharedPreferences(getActivity()).getString("service", "0")).` wouldn't make any code collide. So we could put it into fragment_main anyway. ... What do you think?

wb9688: Oh, if that doesn't throw an error, you should add it.

mauriciocolli: It don't, but we are hardcoding keys (see settings_key.xml). Also, why are we parsing a `String` when we could save the `int` directly?

app/src/main/java/org/schabi/newpipe/util/ExtractorHelper.java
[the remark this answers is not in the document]
wb9688: You could apply [this](https://gist.github.com/wb9688/dc1ac3f7154a3f14e970380d429b24e4) with `git am --signoff < kisok.patch` I guess.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
wb9688: @mauriciocolli: I tried doing it with `int`s, but I got some weird error. If you could get it working, please commit that to #693.

app/src/main/java/org/schabi/newpipe/fragments/list/kisok/KioskFragment.java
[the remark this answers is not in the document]
theScrabi: I don't know I don't like generalizing either.
But why don't we just crawl the Name.

app/src/main/java/org/schabi/newpipe/fragments/list/kisok/KioskFragment.java
mauriciocolli: Also there's a dangling javadoc comment there (use /* instead of /** for copyright notice, since you're not documenting).

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
theScrabi: Well we we could put it into settings_keys.xml than its just one line that collides.

app/src/main/java/org/schabi/newpipe/fragments/list/kisok/KioskFragment.java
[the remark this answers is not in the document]
mauriciocolli: I thought in generify the extractor and then _**in the NewPipe itself**_, make a unique translation for each service, much better than hardcode-checking in this repository (and translating in the extractor repository would force us to use some way to get the translations).

E.g.:
```java
switch (kiosk.getType()) {
    case Top:
        getString(soundcloud_top); // Top
    case Trending:
        getString(soundcloud_trending); // New & Hot
}
```

>But why don't we just crawl the Name

From where? SoundCloud for example, doesn't have all languages that NewPipe support.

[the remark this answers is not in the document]
wb9688: Yeah, I don't like 'generifying' it. The problem with SoundCloud is also that they also have separate Top 50's and New & Hot's for different genres, which we don't have implemented yet though.

wb9688:
@theScrabi: Btw you haven't implemented the UI for eg. switching between Top 50 and New & Hot in SoundCloud's case, right?

app/src/main/java/org/schabi/newpipe/fragments/list/kisok/KioskFragment.java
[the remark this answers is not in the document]
theScrabi: Well I don't like it ether. Because many services might make thins different form other services, even tho they might almost do "similar" things. Also I'm 100% sure people who will impelment their own NewPipe fronted will get that wrong. In my opinion It's the isiest solution to just crawl the translation. If it's not available in the language you selected then the content of the service will bin in the fall-back language anyway, so taking the effort ant translating only the headline would not be worth the work.

theScrabi:
Not yet, atm its only trending.
Im willing to add filter, and selection at another time, and then not only for kiosk but also for channel and search, so you can also select latest first and such fuu.

theScrabi:
Not yet, atm its only trending.
Im willing to add filter, and selection at another time, and then not only for kiosk but also for channel and search, so you can also select latest first and such fuu.

theScrabi:
Looks like this now, and works ... somehow:
screenshot_1506337606

wb9688:
@theScrabi: I really don't like the idea of crawling the name
(@mauriciocolli: What do you think about that?). Besides that, is there a
reason why you're using that weird header instead of just putting the name
in the app bar as a page filter (so you could also choose for the other
kiosk types)? (And why those stars though?)

2017-09-25 13:07 GMT+02:00 Christian Schabesberger <notifications@github.com

karyogamy:
Since trending and other kiosk listings are network calls, wouldn't it be better to keep them as separate fragments? This way we can have a list of kiosk items on the main page (similar to "what's new" header on the subscription pager).

theScrabi:
@wb9688 the stars are just decoration. I thought it would look better. Tho i don't have an Idea how to style up the header.

However I thought about letting the user decide what he wants yo see on his startpage, so instead of just letting him see a kiosk we could add something to the settings so he can set a channel/the channel view/feed/blank page/and kiosk.

Now i would like to list the available kiosks in the dropdownmenue. (Not as separate fragments).

Also as headline for the startpage id like to put the service. Which was youtube in this case.

So in order to know what time of kiosk you are seeing id like to leave a header. Although its true the current header is ugly.

wb9688:
>Also as headline for the startpage id like to put the service. Which was
youtube in this case.

@theScrabi: Why not your color idea instead?

mauriciocolli:
> Off-topic: is there a way to test this without an actual device and without Android SDK's emulator that never works for me?

Did you tried GenyMotion? (although atm, the official emulator is superior IMO)

> I really don't like the idea of crawling the name

I also disagree with this, and as I gave an example before, SoundCloud doesn't have all the languages that we support.

>  If it's not available in the language you selected then the content of the service will bin in the fall-back language anyway, so taking the effort ant translating only the headline would not be worth the work.

How does it not worth the effort? Everyone understand what "Trending/New & Hot" means?

--- 

In general, I think this PR is just too much early...

theScrabi:
> In general, I think this PR is just too much early...

I opened the PR also for discussion even tho we don't merge now.

> @theScrabi: Why not your color idea instead?

Both


So I thought it was better for the frontage to also be able to select other fragments as well. So I created a fragment selector in the settings:

screenshot_1506442342 screenshot_1506442349 screenshot_1506442331 screenshot_1506442477

Also by far not finished.

timofonic:
@wb9688  @mauriciocolli About the Android emulator: Are you using Linux? If yes... What about using [Anbox](/anbox/anbox)?

wb9688:
@timofonic: I'm using Debian Unstable, but Anbox tells me it doesn't
support Debian.

theScrabi:
I took care about some critics.
First of all I've added the [KioskTranslator](https://github.com/TeamNewPipe/NewPipe/blob/f7534b3a0f72ca86f195de2a17e1bbab4669a74b/app/src/main/java/org/schabi/newpipe/util/KioskTranslator.java) utill with which we can use resource to translate Kiosk name. Crawling the translation will still remain for compatibility reasons, tho it will not be in use in the android front end.

Secondary I've added a sub menu into the main menu, from which one can select the kiosks of the current services.

I am done with my work, so what do you guys think?

screenshot_20171009-143234
My latest changes add a menu to the settings with which one can select Kiosks of multiple services. (It is disabled for now)

app/src/main/java/org/schabi/newpipe/fragments/list/kisok/KioskFragment.java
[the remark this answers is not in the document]
theScrabi: I've add [KioskTranslator](https://github.com/TeamNewPipe/NewPipe/blob/f7534b3a0f72ca86f195de2a17e1bbab4669a74b/app/src/main/java/org/schabi/newpipe/util/KioskTranslator.java) with which local translation is possible.

app/src/main/java/org/schabi/newpipe/util/ExtractorHelper.java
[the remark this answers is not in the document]
theScrabi: done

theScrabi:
I need your opinion.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -8,21 +9,43 @@
...
+import java.util.concurrent.ExecutionException;
+
 public class MainFragment extends BaseFragment implements TabLayout.OnTabSelectedListener {
     private ViewPager viewPager;
+    private boolean showBlankTab = false;
+
+    public int currentServiceId = -1;
+
+    /*//////////////////////////////////////////////////////////////////////////
+    // Konst
```
KeizerDev: It should be const. Please no german here :stuck_out_tongue:

TobiGr:
I have very little time , so I can't say anything about the code at the moment. But after a short test, I can say: it works great! 👍 
I only found these small things:
1) The title of a kiosk is always shown twice - in my opinion once too much. I'd keep the one in the toolbar.
2) If you change the content of the main page in the settings, the change will only be applied after a restart of the app.
3) The margin when selecting the channel is to big and the padding is a little small, which makes the whole dialogue look squashed. Same things happen with the "Select a Kiosk" dialog.

<table><tr><td></td>
<td>
</td></tr></table>
With tablet size, the channel selection does not look optimal either.

screenshot_2017-10-11-16-36-28

PS: I am really looking forward to see this merged because I like the new features a lot!

theScrabi:
I removed the header, tho I have no idea how to make the UI update immediately.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -112,6 +148,14 @@ public void onTabReselected(TabLayout.Tab tab) {
         @Override
         public Fragment getItem(int position) {
             switch (position) {
+                case 0:
+                    if(PreferenceManager.getDefaultSharedPreferences(getActivity())
```
karyogamy: Why not just remove subscription page as a kiosk candidate? This kind of check might be problematic when there are more than 2 pagers on the main fragment.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -126,7 +170,89 @@ public CharSequence getPageTitle(int position) {
...
+
+    /*//////////////////////////////////////////////////////////////////////////
+    // Main page content
+    //////////////////////////////////////////////////////////////////////////*/
+
+    private Fragment getMainPageFramgent() {
+        try {
+            SharedPreferences preferences =
+                    PreferenceManager.getDefaultSharedPreferences(getActivity());
+            final String set_main_page = preferences.getString(getString(R.string.main_page_content_key),
```
karyogamy: `set_main_page` should be in camel case.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -126,7 +170,89 @@ public CharSequence getPageTitle(int position) {
...
+                return 2;
+            }
+        }
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+    // Main page content
+    //////////////////////////////////////////////////////////////////////////*/
+
+    private Fragment getMainPageFramgent() {
```
karyogamy: Must we hard-code the strings in this method? If so, we should extract them.

app/src/main/java/org/schabi/newpipe/util/ServiceIconMapper.java
```diff
@@ -0,0 +1,35 @@
...
+ * NewPipe is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with NewPipe.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+public class ServiceIconMapper {
```
karyogamy: Are we allowed to use these icons? I thought they are subject to copyright.

app/src/main/res/values/strings.xml
```diff
@@ -266,4 +266,28 @@
...
+        <item>@string/feed_page_summary</item>
+        <item>@string/subscription_page_summary</item>
+        <item>@string/channel_page_summary</item>
+    </string-array>
+    <string name="select_a_channel">Select a channel</string>
+    <string name="no_channel_subscribed_yet">No channel subscribed yet</string>
+    <string name="select_a_kiosk">Select a kiosk</string>
+
+    <!-- Kiosk Names -->
+    <string name="kiosk">Kisok</string>
```
karyogamy: Spelling.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -126,7 +170,89 @@ public CharSequence getPageTitle(int position) {
...
+    /*//////////////////////////////////////////////////////////////////////////
+    // Select Kiosk
+    //////////////////////////////////////////////////////////////////////////*/
+
+    private void createKioskMenu(Menu menu, MenuInflater menuInflater)
+            throws Exception {
+        StreamingService service = NewPipe.getService(currentServiceId);
+        KioskList kl = service.getKioskList();
+        int i = 0;
+        for(final String ks : kl.getAvailableKisoks()) {
```
coffeemakr: Typo in extractor

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -1,12 +1,168 @@
...
+                        if(newValue.equals(getString(R.string.kiosk_page_key))) {
+                            SelectKioskFragment selectKioskFragment = new SelectKioskFragment();
+                            selectKioskFragment.setOnSelectedLisener(new SelectKioskFragment.OnSelectedLisener() {
+                                @Override
+                                public void onKioskSelected(String kioskId, int service_id) {
+                                    defaultPreferences.edit()
+                                            .putInt(getString(R.string.main_page_selected_service), service_id).apply();
+                                    defaultPreferences.edit()
+                                            .putString(getString(R.string.main_page_selectd_kiosk_id), kioskId).apply();
+                                    String summary = "";
```
coffeemakr: Hard-coded string format. A better option would be to use string resource with the value 
`%1$s/%2$s`

app/src/main/res/values/strings.xml
theScrabi: fixed

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
theScrabi: fixed

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
theScrabi: done

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
theScrabi: Well fix that once more pages are there. Also some people requested the Subscriptionpage to be shown directly when NewPipe is opened.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
theScrabi: fixed

app/src/main/java/org/schabi/newpipe/util/ServiceIconMapper.java
theScrabi: Acording to @TheAssassin no, so I deaktivate it for now.
We later have to discuse that.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
theScrabi: No, we discused about that. We came to conclution to not do that.

karyogamy: I meant extracting the hardcoded strings and urls to a static variable or maybe put them in the Constants file.

theScrabi: Ow I see.

theScrabi:
I acidently merged "dev" into this branch, now I don't know how to squash the commits again :/

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
theScrabi: I created some static variables in the header instead

theScrabi: fixed

PR #750, opened 2017-09-29
Fix top border of launcher icon

TobiGr:
There was a small typo which caused this tiny border at the top of the launcher icon:
launcher_icon_before
This blemish is usually not noticeable, but since we also want to provide our icons in the press kit, they should be perfect.

### DIFF
--- a/assets/new_pipe_icon_5.svg
+++ b/assets/new_pipe_icon_5.svg
@@ -505,7 +505,7 @@
      cx="96"
      cy="96"
      rx="88"
-     ry="87" />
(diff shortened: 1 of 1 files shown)


theScrabi:
I left that on purpouse. I remember according to stile guide you should leave a border. I might be wrong tho, and leaving it also ok if its better :)

TobiGr:
Oh ok. I didn't know this. But then I can close this pull request. It just looked weird to have this icon in a very large size with two different borders at the bottom and top.

Issue #737, opened 2017-09-25
Video crashes when returning to player


theScrabi:
Steps to reproduce:
1. Start a video in ExoPlayer
2. Switch to another app (or launcher)
3. Return to NewPipe
4. Tab on play again.

Stacktrace:
```
org.schabi.newpipe.report.ErrorActivity: java.lang.NullPointerException: Attempt to invoke virtual method 'int android.media.AudioManager.requestAudioFocus(android.media.AudioManager$OnAudioFocusChangeListener, int, int)' on a null object reference
                                                                                                           at org.schabi.newpipe.player.BasePlayer.onVideoPlayPause(BasePlayer.java:563)
                                                                                                           at org.schabi.newpipe.player.MainVideoPlayer$VideoPlayerImpl.onClick(MainVideoPlayer.java:280)
                                                                                                           at android.view.View.performClick(View.java:6256)
                                                                                                           at android.view.View$PerformClick.run(View.java:24697)
                                                                                                           at android.os.Handler.handleCallback(Handler.java:789)
                                                                                                           at android.os.Handler.dispatchMessage(Handler.java:98)
                                                                                                           at android.os.Looper.loop(Looper.java:164)
                                                                                                           at android.app.ActivityThread.main(ActivityThread.java:6541)
                                                                                                           at java.lang.reflect.Method.invoke(Native Method)
                                                                                                           at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
                                                                                                           at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
```

xtj-9182:
this happened to me twice already, my stacktraces are slightly different, maybe it helps:

[newpipev010exception1.txt](https://github.com/TeamNewPipe/NewPipe/files/1336974/newpipev010exception1.txt)

[newpipev010exception2.txt](https://github.com/TeamNewPipe/NewPipe/files/1336975/newpipev010exception2.txt)

rosariotech:
I got these problem too. Running Lineage 14.1 and the NewPipe 0.10.0. I got these erro whenever I go to another app or lock screen. When I play again, the app crashes.

rosariotech:
I just did the test enabling the option to use the old intern player (MediaFramework), and the crash doesn't happen.

theScrabi:
I think we can close this. @mauriciocolli provided a fix for this.

PR #756, opened 2017-10-08
 Fix crash when returning to player

tryton-vanmeer:

When switching apps or locking the phone, destroyPlayer is called which sets audioManager to null. So upon returning to the player and pressing play, the app crashes.
So now initPlayer checks if audioManager is null and sets it if needed.

Closes issue #737


theScrabi:
I cherry-picked your [fix](https://github.com/TeamNewPipe/NewPipe/commit/b70fd826e721e5bcf0e9341f7efa508ea687c53c).

Issue #749, opened 2017-09-29
Video fails loading (Guru Meditation...) if changing device rotation while loading.


krompus:
Please send a bug report.
Or paste a stack trace, and please describe how to reproduce that error :)

Phoenix616:
Can confirm that this is happening. Already reported but here is my crashreport:

```
{
   "user_action": "requested stream",
   "request": "none",
   "content_language": "en",
   "service": "<unknown>",
   "package": "org.schabi.newpipe",
   "version": "0.10.0",
   "os": "Linux Android 7.1.2 - 25",
   "time": "2017-10-07 01:17",
   "ip_range": "146.52.0.0",
   "exceptions": [
[...]

PR #757, opened 2017-10-08
Fix service id not initialized

### LINKED ISSUE #744: Search crash

Unfortunately I can't reproduce this crash, as far as I remember I had searched for something and was going through the list and watched videos and went back to the search result and at some point I think I double clicked back and it crashed but I can't reproduce it. I thought it was worth posting the stacktrace just in case:

[newpipev010exception3.txt](https://github.com/TeamNewPipe/NewPipe/files/1336998/newpipev010exception3.txt)

coffeemakr:

Code improvements and proguard rules for icepick (copy pasted from the [README](https://github.com/frankiesardo/icepick)).

This should fix all serviceId == -1 issues and maybe some others. The problem is that this only occurs when using the release configuration because Proguard probably removed some icepick annotation.  

This closes #744  and #749


app/src/main/java/org/schabi/newpipe/util/ExtractorHelper.java
```diff
@@ -50,7 +50,14 @@ private ExtractorHelper() {
         //no instance
     }
 
+    private static void checkServiceId(int serviceId) {
```
mauriciocolli: IMO we don't need to check that. All the functions uses `NewPipe.getService`, which throws a self-explanatory error in the case of an non-existent service: "There's no service with the id = xx".

coffeemakr: Yes the functions use this method but deferred. The exception may thrown at any time in any thread and the calling method is not shown. 
Have a look at https://sentry.schabi.org/sentry/newpipe/issues/382/. There is no simple way to see from which fragment the serviceId is provided.

coffeemakr: So it this a blocking issue?

coffeemakr:
I don't mean to be rude but this bug should be fixed ASAP. Generate a bugfix release or whatever but this issue has 10 reports a day.

theScrabi:
@coffeemakr Yea kind of true. So the way I see this is that we don't know where this issue happens, right?

@mauriciocolli If it helps we might need to introduce that checkServiceId.
In my opinion it would not hurt to have this function.

coffeemakr:
* `checkServiceId` is to fail faster.
 * It's clear that **all saved states are not restored** because IcePick is used and Proguard removes it.

PR #782, opened 2017-10-25
Better looking Tabs

theScrabi:
### =>

screenshot_20171026-003642

Tho when you swtich to the Subscription page there is no more title in the toolbar, I didn't yet figure out how to fix this.


wb9688:
Nice! Btw is there a reason why we aren't using vector drawables?

PR #783, opened 2017-10-26
Popup player fling to close, multitouch resize and fill screen

karyogamy:
As you can see, I failed my flinging horribly the first time because my fling is too weak. So please check it out and see if you think the fling velocity needed to close popup player should be lower.


theScrabi:
Works fits, nice :)

PR #784, opened 2017-10-26
Fix kiosk typo 

TobiGr:


theScrabi:
Amazing how much strugle one word can produce :/

Issue #704, opened 2017-09-19
Retain timestamp position while switching between background/normal/popup mode.


afya:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

Currently if I press background playback in the middle of the video, the background player starts from the beginning. Is it possible to have an option for it to retain the time position where I switched mode? And possibly the reverse?

Or simply allow the video to continue playing while screen is off.

Stypox:
This seems to work correctly now.
![ezgif-4-05a387fd4d1b](https://user-images.githubusercontent.com/36421898/59975929-70c82300-95be-11e9-9eff-640062692d96.gif)

PR #794, opened 2017-10-29
Add dropdown and ability to switch from normal player to background

SpajicM:

Being able to switch from video play to background seems like a very useful option to me, especially for music playback. Also, this feature was requested in #704 .

While it works as intended, I would appreciate some tips design wise:
- [ ] Existing audio icon seems slightly too big.
- [ ] We are already cutting the video title alot, is there any other placement for video options, maybe under the title or some drop down menu?

22894952_10210752372430789_239597543_n


theScrabi:
Yes thank you, I was waiting for that :)

theScrabi:
Tho we should consider to add a dropdown menu which will show up when to much icons are in the toolbar. It gets a little crowted up there. :P

karyogamy:
I agree with having a dropdown there as well. However, I'd say we wait until #680 is merged so there will be a UI to provide buttons for transitioning from any mode to another mode.

galaxyfreak:
How is play queue feature related to this? Am I missing something?

SpajicM:
How is play queue feature related to this? Am I missing something?

karyogamy:
Play Queue adds a new UI to the background player so users can do different operations to the playing queue. This adds space for a UI button which lets users to transition from background player mode to a popup or main video player mode.

SpajicM:
Tried out play queue commit. It doesn't really interfere with this.
This adds only main or popup -> background, not the other way around. Minus the fact with play queue icon up there, there is even less space for this icon.

theScrabi:
All right so, Playlist is merged. But now we should consider to emigrate the playlist over to the regular player and vice versa when someone switches between background and foreground.

SpajicM:
Still can't think of a way how to make nice.
Currently running with dropdown like this
screenshot_20171104-154203
screenshot_20171104-154208

theScrabi:
Leave important option visible as they are right now, and put less important options into the three dots menu.

SpajicM:
Alright, but which ones do you consider "important"?

theScrabi:
I'd say the switch rotation, and the switch to background are most important.

SpajicM:
I think it would make more sense to have both "modes", background and pop-up at same place, leaving and rotation, queue as icons.

I think this is pretty good, or atleast much better than before.
23365160_10210808205906591_462187900_n

karyogamy:
Good job, that's what I had in my mind as well. Although, it will look better if the popup and background item has an icon next to them. And if possible, please make quality and playback speed display their currently selected value as well.

theScrabi:
> it will look better if the popup and background item has an icon next to them

Thats what I had in mind as well.
If its not possible to put playback speed and quality into the menue leave them out.
In fact, if there is a little more space in the toolbar id like the quality menu to remain there. :)

SpajicM:
How about now? And how should we sort out inconsistant icon sizes?

23365098_10210809205651584_723376291_n

karyogamy:
Instead of having the rotate and dropdown buttons as drawables, you might want to add them as icons, which can have variable sizes and themed colors. Take a look at attrs.xml and [material icons](https://material.io/icons/). If the sizes are off after replacing them as icons, then you might have to manually adjust the width and height attribute of those image buttons.

theScrabi:
Hm if we put the toggle orientation button into the menu we should remember the rotation of the last played video. I could Imagine it was kind of annoying if I had to open that menu every time I want to play that video.

@SpajicM we are getting closer :)

SpajicM:
What do you suggest then?

Rotation is already kind of tricky to put in a menu because it's in MainPlayerActivity while pop-up and background are in VideoPlayer.

karyogamy:
To be honest, why not just leave the rotation out as well and do what was originally intended? If it requires too much work to save the space of a single button, we should just keep the same UI (with mode change dropdown, of course). Also, having the rotation button in the dropdown might make misclicking more costly anyways. If there is a need to add buttons in the future, we can discuss how to remap the UI then.

theScrabi:
@karyogamy I think the menu isn't a bad idea. We will need it anyway. If someone adds more features more buttons will come up so better now than then. 
@SpajicM with remember the orientation I meen NewPipe has to save the orientation of the last watched video. So I like your latest design. I just added a thought :)

karyogamy:
@theScrabi I'm not saying we need to get rid of the dropdown menu though. If @SpajicM thinks adding the rotation button into the queue along with persistence adds too much complexity, it might be a good idea to do another PR.

theScrabi:
Ah thats true :)

SpajicM:
Pushed commit with rotation in the menu. If someone wants to add rotation persistence, feel free to do it.

Also, I've added 3 dots icon in attrs.xml as @karyogamy suggested but that didn't change anything. it just looks slightly "above" the queue icon. And if I change ImageButton size, it's smaller, but still not aligned.

wb9688:
>I wouldn't put the channel title there, since channels can have a white banners and the text will disappear.

Can't you use a shadow then?

SpajicM:
@wb9688 wrong PR?

karyogamy:
Add a padding of `5dp` to the button and resize it to `wrap_content`:

`                <ImageButton
                    android:id="@+id/moreOptionsButton"
                    android:layout_width="wrap_content"
                    android:layout_alignParentRight="true"
                    android:layout_marginLeft="2dp"
                    android:background="#00ffffff"
                    android:clickable="true"
                    android:focusable="true"
                    android:padding="5dp"
                    android:scaleType="fitXY"
                    android:src="?attr/options"
                    tools:ignore="ContentDescription,RtlHardcoded"/>`



You might want to reduce the margin between this button and the edge of the screen. Also, we should dismiss the dropdown when configuration change is detected.

theScrabi:
Since the latest merges there are to much conflicts. I've tried to merge it, I could not :/.

SpajicM:
> Add a padding of 5dp to the button and resize it to wrap_content

Done, now it's fine
> Also, we should dismiss the dropdown when configuration change is detected.

Done, added it to playback & speed dropdowns too.

theScrabi:
Ok merged. But there are still two things that should be done until the next release, and they are:
- [ ] Add button to change from background player to video player
- [ ] Remember rotation of the last player video

PR #796, opened 2017-10-30
Fix history showing even when disabled

SpajicM:

23113347_10210756523254557_84089642_n


theScrabi:
Yep thanks. I put some more fixes on top :P

PR #802, opened 2017-10-31
Implement channel menu (closes #759)

### LINKED ISSUE #759: [bug] "Open in browser" does not work
1.open a video
2. go to channel
3. click three dots
4. select "open in browser"
nothing happens

coffeemakr:
- [X] I carefully reed the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

The share and open in browser menu items are not linked in channel view.

To do:
 - [x] Test it
 - [x] Check if the channel info is loaded.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
@@ -36,6 +36,7 @@ import org.schabi.newpipe.util.AnimationUtils;
 import org.schabi.newpipe.util.ExtractorHelper;
 import org.schabi.newpipe.util.KioskTranslator;
 import org.schabi.newpipe.util.Localization;
+import org.schabi.newpipe.util.NavigationHelper;
 
 import java.util.List;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


theScrabi:
I've tried it, but the open in browser function only opens NewPipe it self again.

coffeemakr:
Hm strange. It works for me. Maybe you have configured NewPipe to always open these links? I'm not sure if there is a possibility to say the link should be opened with another app.

theScrabi:
I did that I't didn't work. But it's not a suprise. I had such trouble with some other of my apps on Android 7. Are you running NP on Android 7?

wb9688:
It works fine for me too, I'm running Android 7. In settings > apps >
newpipe > open by default > open supported links is set to always ask and
it says no standard values setted up (if not press the button above it)?

coffeemakr:
@theScrabi Yes Android 7.0.

Unrelated question: Why do we use the field DEBUG in MainActivity instead of the BuildConfig.DEBUG?

coffeemakr:
> I've just realized it needs some checks while loading so no NullpointerException occurs.

Fixed

theScrabi:
Well that problem with my phone is not gone yet tho :/

PR #807, opened 2017-11-01
fix dex build error

bobpoekert:
fix 'Cannot merge new index 65536 into a non-jumbo instruction' build error

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -3,6 +3,10 @@ apply plugin: 'com.android.application'
 android {
     compileSdkVersion 26
     buildToolsVersion '26.0.1'
+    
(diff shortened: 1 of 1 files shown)


wb9688:
For the beta one? If so, I'd prefer just adding the same stuff
(minifyEnabled, shrinkResources and proguardFiles) to the beta as to the
release. @theScrabi: What do you think?

theScrabi:
@wb9688 yea of course, since problems with minify couldn't be found when disabled for beta.

theScrabi:
@bobpoekert I would like to know what you mean by dex build error?
Can't you build NewPipe?

PR #809, opened 2017-11-03
Remove unused "service_kosk_string" strings

TobiGr:
In #784 I corrected a typo: 
`<string name="service_kosk_string">%1$s/%2$s</string>` => `<string name="service_kiosk_string" translatable="false">%1$s/%2$s</string>`
While I did this some more users translated the false string on wablate. They got merged later and now cause a warning: `Warning:string 'service_kosk_string' has no default translation.` This PR removes them.

### DIFF
--- a/app/src/main/res/values-es/strings.xml
+++ b/app/src/main/res/values-es/strings.xml
@@ -273,5 +273,4 @@ abrir en modo popup</string>
     <string name="kiosk">Kiosco</string>
     <string name="trending">Tendencias</string>
     <string name="top_50">Top 50</string>
-    <string name="service_kosk_string">%1$s/%2$s</string>
 </resources>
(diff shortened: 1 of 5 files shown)


theScrabi:
That can almost not be. The string is in use. See here:
https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java#L125
and here:
https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java#L53

TobiGr:
Yes, but not ""service_**kosk**_string". I corrected the typo in #784.

theScrabi:
memetastic_meme

Amazing how such a small word can cause me such tremendous trouble :/

PR #810, opened 2017-11-04
Some Playback Enhancements

karyogamy:
- ~~Added ability to start playing directly from info item list view through action dropdown: now you can binge through your favourite channel.~~ Postponed until a better UI comes up.
- Added play buttons to channel fragment.
- Added popup player tossing when flung with weaker strength (useful for getting the popup out of keyboard's way).
- Modified background notification to display seeking buttons when play queue has only 1 item, per @SpajicM's request.
- Modified service player activity scrolling to be instant when the distance between the currently viewing and playing is too far apart.
- Fixed track progress reset when shuffle is clicked in quick succession. 
- Fixed NPE when service player activity UI is touched during service shutdown.


theScrabi:
May I be honest, I don't like the Item menu.
This is functionality you can easily get in the DetailScreen. Offering a second way of getting to this functionality will render things overcomplicated, and I want clean UI design.

TobiGr:
I like the idea of direct access to the videos. But I agree with @theScrabi. We could open a dialog after a longpress instead of having tons of menu's.

karyogamy:
I thought about this for a bit, and decided to postpone the "quick access" dropdown feature until a better UI design comes up in the future. New UI for this feature can be easily added since the infrastructure is already there. 
In turn, I've added play buttons to channel fragments similar to that of the playlist fragment.

@TobiGr The problem with long clicks is that the user wouldn't know it exists and there is no place for showing a tooltip on a listview. @theScrabi Suppose we add this feature in a future PR, what do you think about this approach?

TobiGr:
@karyogamy Yes you're right. We definitely need a What's new dialog and a beginner tour or something like that to show the users all features.

theScrabi:
screenshot_20171105-182311

All these buttons make it look a bit bladed.

karyogamy:
@theScrabi What do you recommend we change those buttons to? Maybe we can do something like Youtube, where we just use bar of icons instead of buttons?

theScrabi:
My concern is when someone sets a channel as front-page the buttons should not be displayed. You could test isFrontPage() and then hide the buttons if true.

theScrabi:
Maybe you can also add a triangle button like the on in the detail page. When someone click is these buttons pop up beneath.

theScrabi:
Maybe something like this:
b1
b2

karyogamy:
@theScrabi I wouldn't hide the subscribe button for channels, since it is quite important feature for channels. However, I think it's a good idea to hide the rest under the triangle drawer. We may need to change those buttons to icons eventually though, since there will likely be more functionalities in the future (e.g. save as playlist). 

My concern with hiding buttons is that it takes the user a lot of effort (compared to other apps) to reach those functionalities. This means it might be inevitable we will add secondary pathways to those functions for convenience reasons, we just need to be more selective about what functions and the type of UI we expose. What do you think?

theScrabi:
I have something elde in mind already... May feed better, and no hiding

drgsh:
Can it be done like this??
Button text can be displayed through toast on button long press, if it is must.
image

karyogamy:
@drgsh I had something like that in mind as well, though the buttons would be clickable icons and the icons will be at the same line as the buttons right now. I wouldn't put the channel title there, since channels can have a white banners and the text will disappear.

karyogamy:
@wb9688 Yeah, that would work. Though in the future, there might still not be enough space between the thumbnail and subscribe button, 4 icons should be enough for features in current plan.

theScrabi:
How about this:
32548123-e18ede4c-c484-11e7-9337-5e54ed6f089d

karyogamy:
I tried to do this at first actually, but ended up switching to buttons because it was not as easy to implement (difficult to align the icons) =P I'll look into this again. The icon idea by @drgsh might be more future-proof though, since we will end up change this design when new buttons won't fit.

theScrabi:
> future-proof though, since we will end up changing this design when new buttons won't fit.

I don't know about that. @drgsh's Idea is certainly not bad, but it does't look very nice.

karyogamy:
There you go. I've also added a custom title banner dialog on info item long click as @TobiGr suggested. However, only streams have this dialog since the secondary pathways for other info item types are indeed redundant. For typical streams, the dialog only shows options to enqueue. For streams in playlist and channels, the dialog allows you to start the queue in the middle of the list.

drgsh:
Can we use a FAB menu instead?? Something like this 



which will show popup, play all and bg play options on clicking fab.

I was looking at youtube mobile site which has simple fab, but since here we also have popup and bg play we can use fab menu.

This will work for both channel and playlist fragment and can leave the space around channel header for any other options/buttons.
Also it make sense (according to material guidelines) using FAB since playing a channel videos/playlist is a primary action one want to perform on that screen.

karyogamy:
@drgsh Thanks for the suggestion, but using FABs here would pose the same problems as using the triangle to hide the buttons. Also, I would avoid putting such a complex UI element inside nested fragments on the main activity. The FAB on Youtube app is good practice because it is single-action only and probably on a different activity than the main, just like the "clear" FAB on our history activity.

theScrabi:
@karyogamy very nice 👌

theScrabi:
But the play all looks a little of. Do you think the playlist play icon might fix it?

drgsh:
Sorry if i didnt made it clear before, but fab menu will look something like this initially



Clicking on the fab will show the additional playing options. Thus, the functionality isnt hidden from user like in case of triangle button since the "play" fab indicates one can play the channel videos/playlist.

@karyogamy in what sense you are saying it is being complex UI element(as in implemetation or making it hard for user to find)?

Sorry if i am being redundant.

karyogamy:
@drgsh No worries, it's good to have a different view on UI design. And yes, that bottom fab is what I was referring to. What I meant was that multi-action fabs requires too much actions from the user, just to access those features, similar to the triangle. Fabs are good for top-level activity UI, but it is not nice to have fab suddenly pop up in a nested fragment. 

I'm probably going to stick with the current design, since it is just a minor part of the PR. When time comes where there is not enough space for additional buttons, we will make room by eliminating some of the texts.

karyogamy:
@theScrabi I'll take a look into that, might just be the gravity being off by a little.

theScrabi:
> @theScrabi I'll take a look into that, might just be the gravity being off by a little.

I see. Well we can leave it. I like it that way :)

drgsh:
@karyogamy Yeah, you are right that it takes too many actions for user to get there but I thought it will be a better trade-off since there are more buttons to come. But as you were saying FABs are good for top level activity maybe it's better to stick with your current design.
Thanks for letting me try :)

theScrabi:
So I fixed the layout and merged.

TobiGr:
@karyogamy The longpress improvement is really useful. Thanks! But what's the function of this `STREAM`? I can see it everywhere but what important information does it give me?

theScrabi:
I didn't notice that. What does that stream mean?

karyogamy:
The dialog title banner was designed to work with all InfoItems, the string at the place of "STREAM" shows the type of the InfoItem. This was done when I intended to have playlists and channels have this dialog option. However, now that we don't have any other types of info item using dialog options, please feel free to change the visibility of that element to `gone`.

theScrabi:
Could you remove it please?

karyogamy:
Sure. Are we going to release anytime soon? If not, I'll just include this removal on my next feature PR.

theScrabi:
Hm. Ok.

Issue #816, opened 2017-11-07
Can translate NewPipe to Hindi, Tamil and Telugu


Sgt-Miller:
Hello, I'm here from the about page of NewPipe. Me and my friends can translate newpipe into Hindi, Tamil and Telugu. Is there a Transifex page for NewPipe? 
Thank-you!

wb9688:
We've Weblate at https://hosted.weblate.org/engage/NewPipe/

TobiGr:
I think the problem is, that @Sgt-Miller can't create translations for not existing languages. 
@theScrabi Someone with access to Weblate has to enable these languages or someone needs to create an empty strings.xml in this repository and then wait for a merge by Weblate.

theScrabi:
True

theScrabi:
[done](https://github.com/TeamNewPipe/NewPipe/commit/b242c86869cec3321acb1368201ddc3ef0f4d8cb). Should be available in Weblate soon.

Sgt-Miller:
Thanks

Sgt-Miller:
Thanks

Sgt-Miller:
Looks like Hindi and Telugu have been added in  
https://hosted.weblate.org/projects/newpipe/strings/        
What about tamil?       
Also, why are some source strings in arabic?        Should these be translated too?

theScrabi:
I see. What is the internetional short code for tamil? I must have misspelled it.

TobiGr:
[Wikipedia](https://en.wikipedia.org/wiki/Tamil_language) says `ta` or `tam`. So I guess `ta` is accepted by Weblate as it also offers `ta_LK`.

theScrabi:
Ok

PR #830, opened 2017-11-10
Fix Tamil language code

TobiGr:
Fix the small typo in b242c86869cec3321acb1368201ddc3ef0f4d8cb to enable Tamil language support. See #816. 
@theScrabi I hope Weblate won't crash.


theScrabi:
Hm it would have been better if you added the new language instead of renaming the folder, but well see.

Issue #840, opened 2017-11-12
On long press suggestion insert


ghost:
It'd be nice if I long press on a search suggestion, it'd insert it in the search bar, so I could add some text after it before searching it.

lawonga:
I'm not sure if I can agree with this, it doesn't feel... intuitive. Is there a better way we can approach this?

Issue #836, opened 2017-11-11
Background playing notification text color


Gotolei:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

If a theme elsewhere in the phone (in this instance, one set with Substratum) sets a foreground color on the notification, NewPipe picks it up instead of using its own color. At the same time, NewPipe sets its own background color. This can result in dark text on a dark background. [A couple](https://i.imgur.com/g4s21gw.jpg) [screenshots](https://i.imgur.com/VMFko5G.jpg)
As visible in the screenshots, this happens on both light and dark themes.

Could this be worked around by deciding whether to always use system colors or always use custom colors? The buttons and time bar don't have this issue. Perhaps the notification could use the same set of colors as the rest of the app, or have its own theme setting.

Issue #760, opened 2017-10-10
Feature request: Resume from original time after failure to play


sxmichaels:
[![bounty](https://api.bountysource.com/badge/issue?issue_id=50090160)](https://www.bountysource.com/issues/50090160-feature-request-resume-from-original-time-after-failure-to-play)

- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

I tend to listen to very long (several-hour) lectures on my phone via NewPipe's popup mode. One thing I've noticed happing frequently is that if I pause a video for quite a while and later resume it, it typically stops playing after a few seconds with a message to the effect of "failed to play video." (Presumably it has lost its streaming connection to Youtube.) This completely loses my place in the video I was watching, and particularly in cases where I am passively listening, I do not always know where to return to.

Is it possible to implement **a feature that saves the current place in a video when it fails to play, and resumes from that place when it is re-opened**? I don't think it would be necessary to save this data for long periods of time or per-video; I suspect that if it was just enough to resume a video immediately after it failed, that would be sufficient in 95% of cases.

theScrabi:
I noticed that to. It might be a considerable issue :)

wazateer1:
I came here to comment this exact thing. If not saving the timestamp, could you at least display the time on the crash screen (and maybe a youtu.be/videoid?t=timeofcrash link to restart it)?

rockybulwinkle:
Same, I'd love to see this feature. I've come to rely on it from Google's YouTube app

theScrabi:
@karyogamy I think you have to claim that bounty:
https://www.bountysource.com/issues/50090160-feature-request-resume-from-original-time-after-failure-to-play

PR #848, opened 2017-11-14
Added Expiration to InfoCache and other improvements.

karyogamy:
- Fixes problem described in #760 (I'm a simple man, I see bounty, I try to fix =P). This only fixes the problem caused by expiring Youtube links though, and does not save timestamp on source and renderer errors. Unexpected errors will have timestamp saved temporarily but you will be out of luck if your connection is bad or your device can't support certain formats, where the timestamp will reset. If we want to store the timestamp, we will need to put it in the database. Might be coming on the next PR on local playlists which will have a stream info table for this feature.
- Adds feature requested in #840, screenshot below.
- Fixes text color inconsistency issue in #836 by using fixed text coloring on notifications.
- Modifies the info item dialog banner to show uploader name (if one exists) instead of type name, screenshot below.
- Fixes popup player not running in foreground when launched externally, causing it to get gc'ed when video is still playing.
- Fixes exception when launching popup player externally without launching the player internally first after fresh installation.

Screenshots:


theScrabi:
> I'm a simple man, I see bounty, I try to fix =P

Will that mean that you are the first who gets paid for working on NewPipe? Congrats :)

theScrabi:
@karyogamy if you join us on Bountysource, I can pay you out your bounty.

karyogamy:
@theScrabi I was only joking, isn't the bounty was aimed at the persistent timestamp feature request rather than a just a bugfix? But anyways, keep the bounty. You and the existing team before me deserve it more. Or better yet, reinvest it into some other high priority issue =D

theScrabi:
¿You fixed that bug dind't you? Am I wrong? So whats the deal? If you fixed it you earned yourself 15$. People would not pay money if they didn't want others to get it for a certain work they did. Also I want to see how this works with boutysource (only if you like of cource).

karyogamy:
Alright, I've added myself on bountysource =D

karyogamy:
Please close #840 and #836.

PR #871, opened 2017-11-26
Some Player Quick Fixes

karyogamy:
- Fixes popup player fullscreen button not working anymore.
- Fixes main player UI not fading out after orientation toggle.

There is still problem where the UI does not fade out after pause and play after an orientation change. However, I can't find a root cause at the moment. Will come back to this after mid-December.


avently:
i already fixed it...

Issue #894, opened 2017-12-05
Remember orientation of last played video


theScrabi:
I would like NewPipe to remember the screen orientation of the last played video. This is because the orientation button has moved into a sub menu. Therefore you need two tabs to switch to landscape if you start a video.

I would like to add this before releasing the next version.

theScrabi:
Fixed this here: https://github.com/TeamNewPipe/NewPipe/pull/895

PR #895, opened 2017-12-05
remember last screen orientation

theScrabi:

This makes NewPipePlayer remember its last screen orientation when locking the screen orientation globally, like I suggested here:https://github.com/TeamNewPipe/NewPipe/issues/894


avently:
@theScrabi im trying to place this code in my PR. In what class orientation should be applied? In mainactivity or videoplayerfragment?

PR #904, opened 2017-12-07
Pull up support lib version so all always match

Jawnnypoo:
- [ X ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Pretty simple thing, hope it will make updating this less of a headache later

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -47,6 +47,9 @@ android {
     }
 }
 
+ext {
+    supportLibVersion = '27.0.2'
+}
 dependencies {
     androidTestImplementation('com.android.support.test.espresso:espresso-core:2.2.2') {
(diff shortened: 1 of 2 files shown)


theScrabi:
Have you checked if this will compile with the current fdroid build environment?

Jawnnypoo:
Hrm, wow that is frustrating. I thought I remembered F-Droid trusting by maven url and not by individual artifacts. Sounds good though 👍

avently:
Nice trick

theScrabi:
According to fdroid its supported now :)

Jawnnypoo:
Nice! 👍

PR #910, opened 2017-12-08
Fix failing test

coffeemakr:

The update of the extractor changed the signature of the `VideoStream` and `VideoStream`. I've changed the test accordingly.

**Update**: Also fixes installation of Android components in Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ android:
     # The SDK version used to compile NewPipe
     - android-27
 
+before_install: 
+  - yes | sdkmanager "platforms;android-27"
 script: ./gradlew -Dorg.gradle.jvmargs=-Xmx1536m assembleDebug lintDebug testDebugUnitTest
 
 licenses:
--- a/app/src/test/java/org/schabi/newpipe/util/ListHelperTest.java
+++ b/app/src/test/java/org/schabi/newpipe/util/ListHelperTest.java
@@ -14,35 +14,35 @@ import static org.junit.Assert.assertEquals;
 public class ListHelperTest {
     private static final String BEST_RESOLUTION_KEY = "best_resolution";
     private static final List<AudioStream> audioStreamsTestList = Arrays.asList(
-            new AudioStream("", MediaFormat.M4A.id,   /**/ 128),
-            new AudioStream("", MediaFormat.WEBMA.id, /**/ 192),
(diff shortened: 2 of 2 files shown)


coffeemakr:
I'm not sure why travis is complaining about SDK licenses. Seems like travis changed something on their end. I'll try to fix the configuration file.

wb9688:
It does if there's version mismatch between .travis.yml and build.gradle.

coffeemakr:
gradle and travis didn't change 😕 
From the logs:
>  Downloading SDK Platform Android 8.1.0, API 27, revision 1
> Download finished with wrong size. Expected 65606517 bytes, got 65738431 bytes.
>  Done. Nothing was installed.

theScrabi:
I see it fails because of the license not accepted. Can we fix this?

coffeemakr:
I think the main problem is that build tools 27.0.1 are not installed correctly. 

I suspect that Travis CI caches the build tools by using a proxy and didn't expect google to change the SDK.  (See https://issuetracker.google.com/70292819 for more information.)

The byte sizes are exactly the once show here: https://verification.f-droid.org/build-metadata/platform-27_r01.html where they compare the two different versions.

coffeemakr:
It may be possible to manually create the lincense file for android with the correct hash in it. I will have a look at it later.

wb9688:
Yeah, but I'd consider that a hack, we should just contact Travis.

wb9688:
This is Travis' response:

>Thank you for reporting this issue and I'm sorry for the inconveniences this is currently causing you.
>
>That's interesting and it might be caused because we are using the old `android` CLI to install the Android SDKs under the hood. I'm not sure how fast we'll be able to fix that but in the meantime, I was wondering if you could try to install it yourself using the `sdkmanager` CLI instead? Here's what you would need to add to your .travis.yml file:
>
>
>`before_install: - yes | sdkmanager "platforms;android-27"`
>
>Please let us know if this helps or not and we will follow-up upon your reply.
>
>Thanks in advance and have a great Friday!

coffeemakr:
I didn't expect the support for unpaid project to be this good and responsive.

coffeemakr:
The fix worked! 
@wb9688  Thank you for getting in contact with Travis. If you hear anything new I'd be pleased to hear more about this issue and what caused it.

Issue #998, opened 2018-01-16
add options of "add to the queue" in the history video list


speednroll:
you should add the same options that appear in the search results when you keep pressing a video in the history of videos already seen
![](https://framapic.org/YJTumHnVm1Ob/O7tjfgbyiCXr)

theScrabi:
Yea sounds usefull.

Issue #870, opened 2017-11-23
Add local playlist feature


caogia:
[![](https://api.bountysource.com/badge/issue?issue_id=51898501)](https://www.bountysource.com/issues/51898501-feature-request)

Hi, not sure how to request new feature on github. Please make a favorite/watch later/playlist/ or something that i could play it some moment later.
Does this available ? Thank you

theScrabi:
I think what you request is some sort of local Playlist management. We where discussing about this, and its on our agenda.

However please formulate you request a little more detailed, and please use a more meaningful tittle.

theScrabi:
This could all be implemented allowing local playlists.

theScrabi:
@karyogamy 
Apparently you earned yourself 15$
Please claim them [here](https://www.bountysource.com/issues/51898501-add-local-playlist-feature)

karyogamy:
@theScrabi Thanks.

Issue #537, opened 2017-05-08
FR: Resume position of a video in backstack


gerroon:
Hi

It would be great if NP can remember play positiitions of past played videos. For example I play a video then change to another in the middle of it. I would like NP to remember the time of where I left of when I get back to the first video.

thanks

theScrabi:
Hm cool idea actually.

mauriciocolli:
I think it'd great to integrated this feature with the history (when we implement it)

Loedje:
this feature to resume videos where you left off is a really handy feature on the youtube website when you're logged in and I would love to have it in the newpipe app.

ildar:
Very demanded feature.
Though the title of this issue needs to be changed as it's currently obscure.

rollac01:
It would be amazing to have this feature added if possible :-)

theScrabi:
Yea lets see, we have some stuff to do first, but maybe some day :)

PR #834, opened 2017-11-11
New embeded in fragment video player

avently:
What was done in this PR:
- all players implemented in one service.
- video player located at the same window as video title, description, related videos, etc. You don't have to switch to another window. Everything in one place.
- popup player supports playlists. Yes, next and previous videos can be selected via popup window.
- easy switching from one player to another. Playing position will be the same. Even current video from playlist will be the same in other player after switching. 
- swipe controls improved. Now you can rewind/forward a video via swipe to the flexible number of seconds/minutes. Also previous brightness value will be saved and used after restart.
- fullscreen and half screen support. If you want to read description while watching a video it is possible now. 
- option in settings which will allow to play an audio while you are hiding the app and using an other app. Video will be disabled and your internet traffic can relax with only audio. Then after returning back to the app video will start working again. 
- autoplaying all videos option. You don't have to click on a play button if you don't want it. This can be done automatically. 
- autoplaying while your phone uses landscape orientation.
- you will see next/previous video buttons only when its needed. 
- if you want to contribute to NewPipe you'll notice how much code was removed and simplified because now all players located in one file. That's insane,  yes.

 

- [ + ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #1004, opened 2018-01-17
Local Playlist and Database Normalization

karyogamy:
IMPORTANT: Please test this on an emulator since this PR includes a database migration script (that normalizes the watch history table) which is currently irreversible.

 - This PR aims at implementing #870 and is a work in progress.
 - Along with local playlists, this PR also tries to normalize the database schema so stream metadata can be reused. 
 - This allows us to create smart playlists, such as latest unique watch history (#998 is also implemented as part of this) and most watched streams. In the future, we can use this data to experiment with video recommendation using related videos and tags.

Here are some screenshots, which is probably better at explaining the new features. These UI are not final, please feel free provide ideas:

Here's an overview of the new database schemas:

- Streams: contains extracted stream metadata
- Stream History: [FK on Streams] contains timestamps of every replay
- Stream State: [FK on Streams] contains user states of a stream, such as saved progress of streams
- Playlist: contains name of user created playlist
- Playlist Stream Join: [FK on Streams and Playlist] contains what streams are in each playlist

Todos (not in order):
- [x] Bulk add playlist from queues (saving ~2000 videos takes about just under a second on emulator)
- [x] Integrate normalized database to display on history activity
- [x] Simpify history activity to use dialog instead of gesture for deleting entries
- [x] Allow modifiable playlist (i.e. reordering and deleting streams, rename playlist and change thumbnail)
- [x] Add query for wiping stream entries which have no joins with other tables
- [x] Refactor/Improve info InfoItem holders to show additional items
- [x] UI improvements
- [x] Add timestamp recording when streams changes midway (#537), no resuming in this PR 

Future PR:
 - Searchable playlist dialog.
 - Add UI for resuming past videos
 - Add query for list by services, currently throws unsupported exceptions
 - Video recommendations using history statistics and related video

Please feel free to give recommendations and report bugs, though please don't test this on a real device yet.


avently:
Nice work! Awesome PR. Waiting until it gets merged with your future improvements. Are the timestamps in git log real? I mean, did you really write all this code in 3-4 days?

theScrabi:
Awsome PR !!!
Is it possible to put videos from different services into one playlist?

karyogamy:
Thanks guys.

@avently It actually took me week to get the first commit in since I needed to think through the design. But the rest are just CRUD work and UI, which I mostly just piggybacked on the existing InfoItem UIs. Funny though, I thought about the same when you were able to merge and refactor the popup player so quickly =D

> possibility to edit playlist name.

Yep, updated the todos.

> check before adding something to playlist because now possible to add the same video twice.
> icon in a video that already added to playlist should be different. Now it's "plus" but if video already in playlist it should be "minus" (or something else if you want to support one video -> multiple playlists).

These are probably stretch goals though, since I want to minimize the impact on VideoDetailFragment before your PR is merged. 

> possibility to add to playlist from suggested videos.

I've thought about adding an option to add to playlist on every stream info item as well, but in #810 there was an agreement that the dialog controls do not include functionality that is easily accessible after the fragment is opened. @theScrabi what do you think about this?

Exception should be fixed, thanks again.

@theScrabi 
> Is it possible to put videos from different services into one playlist?

Yep, currently each playlist just keep track of the items on the streams table, which contains of the service id. In the future, we can even do queries to filter playlist by service ids, if someone fancies that.

avently:
@karyogamy
>here was an agreement that the dialog controls do not include functionality that is easily accessible after the fragment is opened.

In some places of the world an internet connection is slow or costs much money. And if user searches, for example, he maybe want to watch videos later from playlist/bookmarks. That's a use case. There is an option in settings which named autoplay. So every load of fragment will trigger video downloading. Also landscape orientation will trigger  to do it too.
We can remove two menu items because there will be two much buttons. This items are: Play in popup and Play in background. In popup player I already support next/previous buttons so this two menu items unneeded

karyogamy:
> In some places of the world an internet connection is slow or costs much money. And if user searches, for example, he maybe want to watch videos later from playlist/bookmarks. That's a use case. There is an option in settings which named autoplay. So every load of fragment will trigger video downloading. Also landscape orientation will trigger to do it too.

Not arguing with you here, I want to be able to do that as well, especially this feature is available on the official Youtube app as well. I'll will add this at the last stages when the PR is feature complete and we are just working on the UI. The code to do this is already there, we just need to add an item to the dialog.

> We can remove three menu items because there will be two much buttons. This items are: Play in popup and Play in background, and Play. In popup player I already support next/previous buttons so this three menu items unneeded. "Play" button is reeeealy useless since play action will be called after user selects a video with tap

I think this would better be done under a separate PR after yours is merged though, since this is a low hanging fruit on UI upgrade and is not really related to the core functions of both your PR and mine.

TobiGr:
I'm also about to write my exams, so I just took a quick look and it looks great!
I'd suggest to replace `BOOKMARKS` on the main page by the bookmark icon.
Maybe we should discuss what to do with the video information and quick actions on smaller phones as they are quite crowed:

screenshot_20180119-200817

karyogamy:
Haha, yeah =P The UI are a mess right now as I'm just aiming for feature completion at the moment. The icon for add to doesn't even line up with the icons to its right. 

Please feel free to suggest a new UI design and bookmark icon though, I'll implement it towards the end of the PR.

theScrabi:
I tought about this problem. Maybe we could add a dropdown menu when there is not enogh space, but Im not yet convinced of that idea.

karyogamy:
@theScrabi What about swapping the line for uploader with the addTo+background+popup buttons. This way we can pull another button from the top and make a row of 4 buttons similar to the Youtube app.

theScrabi:
Yea I think it was good.

wb9688:
In Android Studio you could just add one of the Material Design Icons and
I'd suggest the 'bookmark' icon or maybe the 'star' icon.

theScrabi:
Yea I'd prefer the bookmark icon.

avently:
@karyogamy as a suggestion for the future. If user opens a playlist (not just single video) from a channel or a search the "plus" button in the fragment should save the whole playlist as a local playlist. Maybe with a switch in the UI. Something like that:
Save all videos from playlist [×]

karyogamy:
I was thinking about that too, though I'm still trying to choose between saving the immediate partial playlist or downloading the full playlist (which might be a problem when the whole thing has few thousand videos) before saving.

karyogamy:
Local playlist is now feature complete and below are some updated screenshots:



Some things to note:
- The "add to playlist" option is added only to base list stream items (search, trending fragments, etc.). Tell me if it's better to have this option across other stream items (e.g. on channel and playlist fragments).
- I've opted to only put 3 icons in the control row on the new detail fragment UI. This is mostly because pulling down any icons from the top breaks consistency with other fragments.

Some cleaning up and bug hunting have already been done but there are likely some areas I have missed. So please feel free to take a look at the code, provide suggestions or ask questions.

justanidea:
Thank you so much for this amazing work!!

speednroll:
I am very anxious to see this work in newpipe! many thanks!!!

later next could add a fourth button, along with add to the playlist, background and popup, the button to see the comments
but that would be another topic thread

TobiGr:
Great work!
I just noticed when changing the player while playing a video the video is counted again and the number of views in the watch history increases. 
Maybe we should also re-think the watch and search history since you added the most watch section. I was a little confused that I am not able to delete entries in your new watch history but only in the old one. IMO we should combine both and offer two modes: a normal mode which opens the "Play & Enqueue" dialog and another which allows the user to delete entries.

karyogamy:
I agree, the old history panel should really be part of the fragments rather than a new management activity. The problem is that the new history uses aggregate entries and the old history shows entries like a broswer does (finer control over which entries to delete), and a switch would show two different list on the same fragment. IMHO, this should probably be done in another PR to simplify history, since it can easily introduce bugs and other unexpected behaviors.

However, deleting aggregate entries straight from the new history fragment can be easily done, as the same db queries are already used by the old history management. This is what I intend how the aggregate history to be used, but such feature should be easy to add in this PR. What do you think?

mauriciocolli:
@karyogamy what about an option which the user can "save" an online playlist (just url, name and thumbnail), similar to what YouTube does? I think this option would be better suited in the place of the current "Add to playlist".

It would be extremely useful for me, as I have some musics playlist which would be saved and it would be in-sync when I want to play it (I imagine this is a very common use case).

So the current alternatives to the this option ("Add to playlist"):
- We create an option to "Add all to local playlist" (which @avently suggested)
- Let the user select what they want to add (mass-adding) with an option to select all
    - I found this useful too, but I think we could let it for another PR

###### Also, I'll try to see this PR more in-depth soon.

karyogamy:
The first approach might be more feasible. We can add another table which only keeps remote playlists. The only problem I see is that the video count or thumbnails will be outdated until every time the user accesses the playlist. On such remote playlist, the user would open playlist fragment where everything is unmodifiable, including the name, thumbnail and video list. 
Also, I wonder if we should create a subscription page for remote playlist rather than lumping them together with the local playlist.

My problem with the second approach (adding all to playlist) is how we deal with the UI. Currently, you can manually load the playlist to its bottom and save it as a local playlist. Since the video list can be very long, meaning that when saving the playlist, we need to send out a bunch of request in sequence before we can save it. During this time, we either need a blocking UI, which is discouraged. Or we can either keep the request going in the background, which may lag the app and potentially the database. Or we kill the request when user switches out, which would also be undesirable. 

Let me know what you think about the options. Right now I'm trying to get a proof of concept out for subtitles and aspect ratio support on the video player, and will get back to this soon.

oceanwaves90:
Hi. Thank you for your work!

By the way It throws NPE on the lastest commit(actually on some recent commits), only when I try to open the video [https://www.youtube.com/watch?v=-Vpwatutnko](url). Other videos seem working right except for this video. I hope you fix this issue! thank you!

`4380/org.schabi.newpipe.debug E/class org.schabi.newpipe.report.ErrorActivity: java.lang.NullPointerException: Attempt to invoke virtual method 'int com.grack.nanojson.JsonArray.size()' on a null object reference
                                                                                                             at org.schabi.newpipe.extractor.services.youtube.YoutubeStreamExtractor.getAvailableSubtitlesInfo(YoutubeStreamExtractor.java:743)
                                                                                                             at org.schabi.newpipe.extractor.services.youtube.YoutubeStreamExtractor.onFetchPage(YoutubeStreamExtractor.java:558)
                                                                                                             at org.schabi.newpipe.extractor.Extractor.fetchPage(Extractor.java:61)
                                                                                                             at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:248)
                                                                                                             at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:240)
                                                                                                             at org.schabi.newpipe.util.ExtractorHelper.lambda$getStreamInfo$3$ExtractorHelper(ExtractorHelper.java:109)
                                                                                                             at org.schabi.newpipe.util.ExtractorHelper$$Lambda$3.call(Unknown Source)
                                                                                                             at io.reactivex.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:45)
                                                                                                             at io.reactivex.Single.subscribe(Single.java:2801)
                                                                                                             at io.reactivex.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
                                                                                                             at io.reactivex.Single.subscribe(Single.java:2801)
                                                                                                             at io.reactivex.internal.operators.maybe.MaybeFromSingle.subscribeActual(MaybeFromSingle.java:41)
                                                                                                             at io.reactivex.Maybe.subscribe(Maybe.java:3749)
                                                                                                             at io.reactivex.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.drain(MaybeConcatArray.java:153)
                                                                                                             at io.reactivex.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.request(MaybeConcatArray.java:78)
                                                                                                             at io.reactivex.internal.operators.flowable.FlowableElementAtMaybe$ElementAtSubscriber.onSubscribe(FlowableElementAtMaybe.java:66)
                                                                                                             at io.reactivex.internal.operators.maybe.MaybeConcatArray.subscribeActual(MaybeConcatArray.java:42)
                                                                                                             at io.reactivex.Flowable.subscribe(Flowable.java:13068)
                                                                                                             at io.reactivex.internal.operators.flowable.FlowableElementAtMaybe.subscribeActual(FlowableElementAtMaybe.java:36)
                                                                                                             at io.reactivex.Maybe.subscribe(Maybe.java:3749)
                                                                                                             at io.reactivex.internal.operators.maybe.MaybeToSingle.subscribeActual(MaybeToSingle.java:46)
                                                                                                             at io.reactivex.Single.subscribe(Single.java:2801)
                                                                                                             at io.reactivex.internal.operators.single.SingleSubscribeOn$SubscribeOnObserver.run(SingleSubscribeOn.java:89)
                                                                                                             at io.reactivex.Scheduler$DisposeTask.run(Scheduler.java:452)
                                                                                                             at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:66)
                                                                                                             at io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
                                                                                                             at java.util.concurrent.FutureTask.run(FutureTask.java:237)
                                                                                                             at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:154)
                                                                                                             at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:269)
                                                                                                             at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
                                                                                                             at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
                                                                                                             at java.lang.Thread.run(Thread.java:818)`

justanidea:
This video works with my phone

karyogamy:
@oceanwaves90 You are right, this is a problem with the subtitles extraction and has nothing to do with this PR. Still my fault though, wasn't careful enough null checking json docs. A [fix PR](https://github.com/TeamNewPipe/NewPipeExtractor/pull/66) has been created.

@theScrabi @mauriciocolli Remote playlist has now been added, please take a look and see if there's anything missing. Also, I've taken the liberty to add LeakCanary to the debug build, as @avently have mentioned in #834. This, however, creates a separate an additional app for reporting leaks which may not be desirable. What do you think about this?

avently:
This is so cool! Thank you, @karyogamy.
Hope this will be merged as soon as possible because playlist feature is a must-have thing. Even with remote playlists support, wow.

karyogamy:
@avently Hey, thanks for tell me about LeakCanary. Now I know what to do next, with the play queue activity leaking like crazy =P

theScrabi:
Nice work. Works good :)

theScrabi:
Allthough one small thing.
You may need to fix this:

karyogamy:
Yikes, didn't even notice this one. Fixed, thanks.

karyogamy:
I've decided to drop the dependency on LeakCanary in this PR, since it is hogging the app way too much. I'll create another PR to solely introduce LeakCanary as well as adding leak detection to fragments to it.

mauriciocolli:
@karyogamy I don't think having two menu items is a good idea, because is inefficient and as you may have noticed, there's a blinking when the buttons are repositioning (appearing/disappearing).

Instead, use the `setIcon `and `setTitle` for updating the menu buttons, something like this snippet:
```java
private void updateBookmarkButtonsVisibility() {
    if (playlistBookmarkButton == null) return;

    int iconAttr = playlistEntity == null ? R.attr.ic_playlist_add : R.attr.ic_playlist_check;
    int titleRes = playlistEntity == null ? R.string.bookmark_playlist : R.string.unbookmark_playlist;

    playlistBookmarkButton.setIcon(ThemeHelper.resolveResourceIdFromAttr(activity, iconAttr));
    playlistBookmarkButton.setTitle(titleRes);
}
```

app/src/main/java/org/schabi/newpipe/fragments/local/BaseLocalListFragment.java
```diff
@@ -86,19 +86,30 @@ protected void initListeners() {
     // Lifecycle - Menu
     //////////////////////////////////////////////////////////////////////////*/
 
+    /** Determines if the fragment is part of the main fragment view pager.
+     * If so, then this method must be overriden to return true
+     * in order to show the hamburger menu. */
+    protected boolean isPartOfFrontPager() {
```
mauriciocolli: I think it's better to just drop this function and [this call to setDisplayHomeAsUpEnabled](https://github.com/karyogamy/NewPipe/blob/0630423c8e41042521702a2b087391a1a1265bdc/app/src/main/java/org/schabi/newpipe/fragments/local/BaseLocalListFragment.java#L107..L113), and just let the already [implemented function in the activity](https://github.com/karyogamy/NewPipe/blob/0630423c8e41042521702a2b087391a1a1265bdc/app/src/main/java/org/schabi/newpipe/MainActivity.java#L293..L304) handle it.

This check should be used for other things such as: don't inflate menus or change the toolbar too much.

karyogamy:
> there's a blinking when the buttons are repositioning ... use the `setIcon` and `setTitle` for updating the menu buttons

Totally forgot about the new `ThemeHelper`, thanks for pointing that out. Fixed.
The visibility change/flicker is actually used prevent users from clicking to soon when the query is still being run. I changed this to use an atomic boolean for signalling if the button is ready, so the flickering should be gone now.

> I think it's better to just drop this function and this call to setDisplayHomeAsUpEnabled, and just let the already implemented function in the activity handle it.

Ahh, I see what you saying. Fixed/removed.

karyogamy:
Found a quirk in the Soundcloud playlist extractor. Where the url for playlist (info item) in search result uses `https`, the returned playlist info uses the permalink for its url, which is in `http`. 

This discrepency initially causes the bookmark button to stop working properly on Soundcloud when the playlist fragment is opened from a search page, but is now fixed. 

However, this makes me think that we should store links in the database using URN (excluding the http and https protocols) rather than URL. What do you guys think about this, please let me know.

theScrabi:
I just saw, in the new menu where we put popup playlist, etc we could also put the download symbol into.

theScrabi:
I just saw, in the new menu where we put popup playlist, etc we could also put the download icon into. Like the real yt app.

theScrabi:
I just saw, in the new menu where we put popup playlist, etc we could also put the download icon into. Like the real yt app.

karyogamy:
Can you clarify which popup playlist you are referring to? I don't have Youtube Red so no download button anywhere here.

theScrabi:
Basicly I mean this:
screenshot_20180211-023953

theScrabi:
> However, this makes me think that we should store links in the database using URN (excluding the http and https protocols) rather than URL

Yea maybe we could embed a method somewhere that uprise the protocol from http to https.

karyogamy:
> download icon ... Like the real yt app.

Done.

> > However, this makes me think that we should store links in the database using URN (excluding the http and https protocols) rather than URL

> Yea why not. Maybe we could embed a method somewhere that uprise the protocol from http to https.

Sorry, I didn't think this through yesterday and have just realized that it is possible for future streaming services to be `http` only. So for now, we should still let the extractor manage the url but we need to keep the protocol consistent. In later times, we can create an api in the extractor to provide all the possible protocols and create a migration script to get rid of the prefix in the urls in the database.

karyogamy:
Yikes, few seconds too late =P

theScrabi:
> > download icon ... Like the real yt app.
> Done.

I did do it as well now xD

avently:
@theScrabi what this line do?
https://github.com/TeamNewPipe/NewPipe/blame/dev/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L550

With it openPopupPlayer and openBackgroundPlayer called 4 times. Without it everything is fine

avently:
@karyogamy it is possible to add the same video into one playlist multiple times.

P.S. I merged your changes. Let me know if I broke something. And thanks for this PR

theScrabi:
@avently you are right, i fixed the line.

karyogamy:
@avently 
> it is possible to add the same video into one playlist multiple times.

Yes, though I haven't added the ability to see duplicate entries when the add dialog is open (which can get annoying), this functionality should be fairly easy to implement now if anyone wants to give it a shot =D

Issue #996, opened 2018-01-15
the three-point menu does not appear when you watch a video with the light theme on


speednroll:
I do not know if this has already been requested, but there is a minor error when you see a video in newpipe with the light theme

![](https://framapic.org/2CUlqaO1rmhH/hQzk0Mrupzqn)

simply the three-point menu that is in a corner is black and can not be seen. you have to change it by white

theScrabi:
Confirmed. The menu icons also don't fit.

theScrabi:
fixed

PR #1005, opened 2018-01-18
resizeMode zoom to support 18:9 screens ratio.

SkyzohKey:

**NOTE:** This applies to ALL devices, I don't know if it's going to have side-effect, this is not tested. 
@see Possible update by using TextureView => https://stackoverflow.com/a/35705070/3855007

### DIFF
--- a/app/src/main/res/layout/activity_main_player.xml
+++ b/app/src/main/res/layout/activity_main_player.xml
@@ -10,6 +10,7 @@
 
     <com.google.android.exoplayer2.ui.AspectRatioFrameLayout
         android:id="@+id/aspectRatioLayout"
+		 app:resize_mode="zoom"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:layout_centerInParent="true"


theScrabi:
Awsome. I'll take a look as soon as I find time to :)

theScrabi:
By the way when you are already working on the player UI, could you please also take a look at [this issue here](https://github.com/TeamNewPipe/NewPipe/issues/996).

wb9688:
I wouldn't do it like this, since then users don't have an option to watch it in the original aspect ratio.

TobiGr:
@SkyzohKey Did you try to build and run your changes? I can't do it and [travis either](https://travis-ci.org/TeamNewPipe/NewPipe/builds/330531932#L1574). 

> I wouldn't do it like this, since then users don't have an option to watch it in the original aspect ratio.

@wb9688 Maybe we have to implement a logic which decides whether we can zoom in / skew the video depending on the ratio difference and enable/disable it via a new video setting. And we should take care of other typical ratios (1:1 or phone videos which are handled completely different).

wb9688:
Yeah, and on for example Chrome OS, where you could resize the window of an
Android app iirc. So I'd go for a setting that's disabled by default.

theScrabi:
I'll close it for now.

PR #1018, opened 2018-01-21
Add preferred player

### LINKED ISSUE #643: [Request] Option to autoplay in background when started from another app
While it's already possible to auto-play the video when started from another app or to share to the pop-out mode, a feature I'm missing is the ability to have the link auto-play in the background (audio only) immediately after sharing the link.

mauriciocolli:
- Add an option in the share menu to open with the preferred player (replacing the old "open in popup") defined by the user in the settings
- When the user select a stream (video or audio), the options "Background" and "Popup" will enqueue the requested content and then select it, making it possible to the user go back to the previous or access the list in the player UI

##### If "Always ask" is selected (it is the default), the "Preferred player" in the share menu will open a dialog:

---

Closes #643


wb9688:
I'd make 'always ask' a checkbox. Or like Android's 'open with' menu, where
you can select an option and then have 2 buttons 'once' and 'always', which
you can press, or click your option again for 'once'.

mauriciocolli:
I reworked the UI with the tips:

| Before  | Radio Button | Without Radio <br>(background selection) |
| ------------- | ------------- | ------------- |
|   |  |  |

Which one do you guys prefer? (if you have more ideas, just tell me)

karyogamy:
Good work and welcome back =D 

As for the radio buttons, it depends on what the behavior is if the user clicks on the same option twice. If clicking on the same option twice is the same as clicking on the option once then "just once", then we should leave out the radio button. Otherwise, it is better to have it there.

Also, it would look nicer if the dialog is on a BottomSheetDialogs, though this is probably unneeded and may be inconsistent with other dialogs in the app. This feature is good as is.

theScrabi:
I'd like the number three :)

speednroll:
the third option looks comfortable

theScrabi:
Yes defiantly [number three](https://www.youtube.com/watch?v=6zQ55S-DJsM)

mauriciocolli:
Then it's decided.

@karyogamy will keep the idea in mind, and like you said, the current UI is good for now.

avently:
@mauriciocolli i merged your pull request and find an issue. Now you have added three choices - video player, background player and popup player. Also user can select just NewPipe (it will open video player). So why do we have three choices + video player (from NewPipe option)? I think we need two options + NewPipe.

theScrabi:
I think @avently is right, we could make this menu open when someome selects NewPipe. So we will save one MenuPoint.

PR #1022, opened 2018-01-21
Support playing videos on DIAL-Server

detached:
I would like to contribute the feature to play videos on smart TVs.

Netflix and Youtube created the DIAL protocol specification that describes the communication with apps installed on a smart TV (http://www.dial-multiscreen.org/).

On the video detail screen I added an entry to the menu header. In English it is "Play on TV", in german "Auf TV wiedergeben".

This item opens an activity showing the discovered DIAL-Servers:

Clicking on one of the discovered servers sends a play request to the Youtube app on the TV.

For implementing this feature I added a dependency to my DIAL client library JDial: https://github.com/detached/jdial
The library has no dependency to other libraries and is licensed under GPL v3.

The discovery of DIAL-Servers in the local network requires enabling multicast support in the wifi stack of the android device. For that I added the permission `CHANGE_WIFI_MULTICAST_STATE`.

I didn't found a refresh icon in your code base and I was not sure about your style guide and policies for adding content to the code. So I just used the stock android sync icon for this purpose.

I am not a experienced android developer, so please have a critical look at this code.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -89,4 +89,6 @@ dependencies {
 
     implementation 'frankiesardo:icepick:3.2.0'
     annotationProcessor 'frankiesardo:icepick-processor:3.2.0'
+
+    implementation 'de.w3is:jdial:1.2'
 }
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -9,6 +9,7 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
+    <uses-permission android:name="android.permission.CHANGE_WIFI_MULTICAST_STATE" />
 
     <application
         android:name=".App"
@@ -58,6 +59,10 @@
             android:label="@string/app_name"
             android:launchMode="singleTask"
             android:theme="@style/PlayerTheme"/>
+
+        <activity android:name=".player.PlayOnTvActivity"
+            android:launchMode="singleTask"
+            android:label="@string/play_on_tv"/>
 
         <activity
             android:name=".settings.SettingsActivity"
@@ -235,5 +240,8 @@
                 <data android:mimeType="text/plain"/>
             </intent-filter>
         </activity>
+        <service
+            android:name=".dial.DialService"
+            android:exported="false"/>
     </application>
 </manifest>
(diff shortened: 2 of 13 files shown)


theScrabi:
@wb9688 is working on [Pipecast](https://github.com/teamnewpipe/pipecast) right now, which is a library that should support DLNA, Chromecast and airplay. @wb9688 what do you think could this be interacted? Maybe you could use his UI.

wb9688:
1. DIAL is Cast v1, but we wanna use Cast v2 (or even better: Cast v3, but
that's like impossible iirc).
2. The 'TV' select UI is inconsistent with the rest of the UI (black
instead of dark, non-Material refresh button, not a pop-up).
3. I think you're doing it in a way that'd only support YouTube casting.

So I'm sorry your work probably was for nothing.

detached:
So you are not interested in supporting the DIAL protocol for video id handover at all but rather try to reverse engineer the proprietary google cast and airplay to stream videos over wifi?
I am asking because I could help with the implementation of DIAL.

theScrabi:
I would not say it was all for nothing :D. Yes @wb9688 is about UI consistency, but that can be fixet.
PipeCast needs a UI anyway meybe this can be made up for the pipecast UI. And also why cant we have DIAL as fallback?

wb9688:
@theScrabi: In this PR it looks like it'd only support YouTube URLs. If
DIAL can just use video URLs, there would be no problem with adding DIAL.

theScrabi:
I see.

detached:
DIAL is just a protocol for remote launching applications but does not specify the types of applications or what parameters they have to support. You can't just send a url to your TV and the TV tries to guess if it is media it can play. You have to specify exactly which application to launch and provide the parameters in the format of the application vendor. The official Youtube and Netflix apps are controlling their TV apps with this.

I just implemented the Youtube case here because you present your app as "A free lightweight YouTube frontend for Android.". But beside of youtube there are many more apps registered that could be controlled by this protocol (http://www.dial-multiscreen.org/dial-registry/namespace-database#TOC-Registered-Names).

I got that you are basically interested into implementing a way to handover a Youtube URL to a smart TV but not with my proposed UI. So how can we get this done? Should I open an issue on this project?

theScrabi:
I think maybe we have to overthing all of this.
This is not a whole solution, but at least its more that pipecast.

@detached I'm going to close this for now, but keep it up.
Maybe you could help on support propper cast. For example we still need a UI for pipecast.

matejdro:
Just wondering about this issue: Can Cast v2 and v3 procools fallback to DIAL protocol or are they chained to the Chromecast?

The thing is, you can cast to many Smart TVs directly with DIAL, without the need for Chromecast device. If you decide to only implement Chromecast-specific protocol, that functionality would be lost.

theScrabi:
Hm. Sounds usefull.

wb9688:
There are lots of TVs with built-in Cast now. I've never heard of a TV with
built-in DIAL though.

matejdro:
Pretty much every recent Smart TV has DIAL support (for example most Samsung and LG models). You can recognize it by the fact that "Casting" only works in Youtube and Netflix app, but not in every other Chromecast-supported app.

detached:
@wb9688 you should try to portscan your devise

I found my LG 42LB652V supporting this and the official youtube app implemented the discovery. That was why I created the java dial lib.

logos88:
Since this is from 2018, there are any progress to support DIAL or cast in NewPipe?

opusforlife2:
@wb9688

PR #1023, opened 2018-01-22
Update screenshots

TobiGr:

This updates the screenshots in the README and also on F-Droid to a newer version including trending and playlists.


theScrabi:
should we delete the old screenshots from the history again, in order to not bload up the reop?
Maybe we should host the screenshots somwhere else.

TobiGr:
> should we delete the old screenshots from the history again, in order to not bload up the reop?

We can do this although I am not  fan of deleting commits which are on several repositories.

> Maybe we should host the screenshots somwhere else.

This are the same screenshots as I used [here](https://newpipe.schabi.org/press/screenshots/) so we could link the other repository. But I don't know how Fastlane and F-Droid handle links.

TobiGr:
And these images need only 1.0MB. So I'd rather keep the files in this repository.

theScrabi:
Ow right, because of fastlane we need to keep them where they are.

PR #1041, opened 2018-01-28
Added function to import/export the database

theScrabi:
Added function to import / export the database. However the database might not be consistent because I just export the file.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/App.java
+++ b/app/src/main/java/org/schabi/newpipe/App.java
@@ -1,9 +1,12 @@
 package org.schabi.newpipe;
 
+import android.app.AlarmManager;
 import android.app.Application;
 import android.app.NotificationChannel;
 import android.app.NotificationManager;
+import android.app.PendingIntent;
 import android.content.Context;
+import android.content.Intent;
 import android.os.Build;
 import android.util.Log;
 
@@ -116,7 +119,6 @@ public class App extends Application {
         });
     }
 
-
     private void initACRA() {
         try {
             final ACRAConfiguration acraConfig = new ConfigurationBuilder(this)
@@ -149,4 +151,5 @@ public class App extends Application {
         NotificationManager mNotificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
         mNotificationManager.createNotificationChannel(mChannel);
     }
+
 }
--- a/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
@@ -1,6 +1,8 @@
 package org.schabi.newpipe.settings;
 
 import android.app.Activity;
+import android.app.AlertDialog;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.v7.preference.ListPreference;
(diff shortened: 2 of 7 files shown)


TobiGr:
Works good!
If you have time, an option to select the data you want to export, would be nice. E.g. if I want to export my subscriptions to another device, I don't want to override my history. 
And IMHO the possibility of overwriting or manipulating someone's history seems a bit suspect to me.

theScrabi:
> If you have time, an option to select the data you want to export, would be nice. E.g. if I want to export my subscriptions to another device, I don't want to override my history.

Yea some sort of sync with merge :)

karyogamy:
Works well, nice job. This should prove useful when we upgrade database versions in case something goes wrong with migration, though I think the import/export options should be placed in another settings fragment.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -1,102 +1,213 @@
...
+                && ZipHelper.extractFileFromZip(zipIn, newpipe_db_journal.getPath(), "newpipe.db-journal"))) {
+               Toast.makeText(getContext(), R.string.could_not_import_all_files, Toast.LENGTH_LONG)
+                       .show();
+            }
+
+            zipIn.close();
+
+            // restart app to properly load db
+            //App.restart(getContext());
+            System.exit(0);
```
karyogamy: This seems to work well, but is this a good practice to circumvent android lifecycle though? I'm not even sure why this works here, since `System.exit(0)` is suppose to pop off the top level settings activity after restarting.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -1,102 +1,213 @@
...
+        super.onActivityResult(requestCode, resultCode, data);
+        if (DEBUG) {
+            Log.d(TAG, "onActivityResult() called with: requestCode = [" + requestCode + "], resultCode = [" + resultCode + "], data = [" + data + "]");
+        }
+
+        if ((requestCode == REQUEST_IMPORT_PATH || requestCode == REQUEST_EXPORT_PATH)
+                && resultCode == Activity.RESULT_OK) {
+            String path = data.getData().getPath();
+            if(requestCode == REQUEST_EXPORT_PATH) {
+                exportDatabase(path + "/NewPipeData.zip");
```
TobiGr: Maybe we should ask the user whether he wants us to replace an older version of his db.

theScrabi: I thught about buttint a timestampt in the filename.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
theScrabi: Well I'm not sure. The way it should be done according to Stackoverflow is to make the app start using a pending intend with a timer, and after setting this up makeing the application quit.
However if I do it with pending intent the app will restart twice.
I'll check weather this is ok.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
theScrabi: I thught about buttint a timestampt in the filename.

TobiGr: Sounds good.

app/src/main/java/org/schabi/newpipe/App.java
```diff
@@ -149,4 +151,18 @@ public void initNotificationChannel() {
         NotificationManager mNotificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
         mNotificationManager.createNotificationChannel(mChannel);
     }
+
+    public static void restart(Context context) {
```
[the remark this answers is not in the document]
theScrabi: removed code.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
theScrabi: According to [stackoverflow](https://stackoverflow.com/questions/18292016/difference-between-finish-and-system-exit0?answertab=votes#tab-top) `System.exit()` will close the runtime, but emediatly restart the previous activity.

PR #1043, opened 2018-01-28
Make preferred string not translatable

TobiGr:
Thanks to @comradekingu who saw this on [Weblate](https://hosted.weblate.org/translate/newpipe/strings/nb_NO/?type=all&offset=259#comments).
@theScrabi I hope Weblate won't cause any trouble 😅


theScrabi:
Na and if it does that we can fix it.

Issue #993, opened 2018-01-14
[Feature request] RSS feed option to select as watched and load more videos


moshpirit:
I'm subscribed  to several channels (locally, I don't have a Google account) and in my feed there's some videos I already saw and there's only 16, so I would like to see more videos and hide those which I already saw.

salevdns:
An overlay over the thumbnail of an already watched video that indicates that it has been watched before would be great too.

Stypox:
I think this pr #2288 suits your needs, since it displays playback positions under videos.
And for the hiding-already-watched-videos-from-feed part, I think #2309 will solve your problem by showing videos in order.

Stypox:
Oh... closing since #2288 has just been merged :-D

PR #1044, opened 2018-01-28
[ON ICE] Sort the items in the "What's New" feed by the parsed upload date.

### LINKED ISSUE #822: [enhancement] "What's New" sorting improvement
So now the "What's New" page shows the latest video from each channel you subscribed to, but that's not how YouTube subscriptions work, youtubes' subscription list sorts videos by the publish date.

Now in NewPipe each channel have only one video in the "What's New" feed even if they published 10 videos recently "say 1 day ago" at the same time only one will show up in "What's New", and some other channels that you may subscribed to, haven't published in months and you have watched their videos it still shows them every time you go to "What's New".

So it will be nice to sort the "What's New" videos by the publish date.

wojcik-online:
In the "What's New" feed show all videos from the subscribed channels that are not older than a month and sort them by the upload date. Fixes #822.

### How the feed works
0. Show the last downloaded feed from cache.
1. Download all videos from the subscribed channels. (This can take a lot of time.)
2. Localized phrases form `strings.xml` are used to parse the relative upload dates.
3. Restore previous more accurate upload dates from cache.
    When a video is first encountered, it was uploaded 3 minutes ago. Days later when the upload date says '6 days ago' we still know the time of day because it is kept in the cache.
4. Show the videos in groups of 8 (or more) as the user scrolls down.
5. The list of available new videos is kept up to date.
6. The user can reload the the feed any time.

Screenshot of the feed

Depends on: [Parse the upload dates of StreamInfoItems.](https://github.com/TeamNewPipe/NewPipeExtractor/pull/61)

### Still to be done (optionally)

 - [x] **1. caching**
    The initial loading time can be very long. The app should store the previous result in the cache and display it while the new content is loading.

 - [x] **2. consistent relative dates**
    @mauriciocolli Unfortunately the functions from [`DateUtils`](https://developer.android.com/reference/android/text/format/DateUtils.html#getRelativeTimeSpanString(long,%20long,%20long)) do not work quite like expected. For example: when YouTube says '*1 day ago*', it means at least 24 hours, but less than 48 hours.  Android would display '*yesterday*' which is not necessarily correct.
   I cannot get the functions to match the format used by YouTube. They are not flexible enough. If we want to use consistent relative dates across all services, we will have to use an external library. Perhaps `PrettyTime` is OK.

 - [ ] **3.** A RSS reader look'n'feel as requested in #993.

 - [ ] **4. downloading service**
    A downloading service that runs in the background and updated the feed once a day / every few hours. This could be configured in the Settings menu. Additionally if there are new items, a notification could be displayed.


theScrabi:
I'd say point one is not optional, but a must be. We should take the database in account for that. However the database is currently refactored.

wojcik-online:
In the next few day, I'll implement caching the feed in the cache directory. This will have the added advantage that if the device is low on storage capacity, the files can be easily deleted.

theScrabi:
Well I'd like database better

wojcik-online:
My first instinct was to use the database as well because it is already implemented.
But why would one want to import/export the feed? It has to be downloaded every few hours anyway. It makes sense to backup or synchronize the subscriptions, but the feed itself is valid only temporarily.

I've been thinking about creating a mockup to cache the feed and to switch to the real DB once it is stable, but now I am not sure. The [description of the internal cache](https://developer.android.com/guide/topics/data/data-storage.html#InternalCache) fits the use case of storing the feed pretty well. It is basically like the cache of a web browser; it speeds up the load time, but there is no problem if the data is missing.
@theScrabi What do you think?

app/src/main/res/values/strings.xml
```diff
@@ -356,4 +354,33 @@
...
     <string name="preferred_player_fetcher_notification_title">Getting info…</string>
     <string name="preferred_player_fetcher_notification_message">"The requested content is loading"</string>
+
+    <!-- "What's New" feed -->
+    <string name="fragment_whats_new">What\'s New</string>
+    <string name="feed_first_time_loading">Loading news for the first time</string>
+    <string name="feed_last_updated">Last updated: <xliff:g id="update_time">%s</xliff:g></string>
+
+    <!-- Parsing YouTube upload dates in the format '2 days ago' -->
+    <string name="downloader_http_accept_language">en</string>
```
TobiGr: I think that's the wrong place to store the language code. But I think that's what you are looking for:
https://github.com/TeamNewPipe/NewPipe/blob/3a8b04e2d1cb2032cc6840ab754ea2ee4c98bd1f/app/src/main/res/values/settings_keys.xml#L585

app/src/main/res/values/strings.xml
```diff
@@ -356,4 +354,33 @@
...
+    <string-array name="upload_date_seconds_ago_phrase">
```
TobiGr: Why do you use string arrays?

wojcik-online: Some languages might not have a single, simple prefix that matches all forms of a word, like in English / Deutsch.

TobiGr: Do you mean differences between singular and plurals form? Then I guess the `<plurals>` tag is the right choice here. Weblate (our translation service) will then take care of the differences in each language.

wojcik-online: Agree. The problems is there is no function `R.plurals.upload_date_phrases -> String[]`.
I would have to iterate over all possible integers that may change the plural form. Like this:
```java
for (int i = 0; i <= 20; i++)
    phrasesSet.add(recources.getQuantityString(R.plurals.upload_date_phrases, i));
```

The code is a little messy, but this will be certainly better for the translators because the plurals have an obvious meaning. I will implement it like this.

app/src/main/res/values/strings.xml
theScrabi: country_codes are not language_codes, these are two different things.

theScrabi:
I this still work in progress, or would you suggest to try a merge?

wojcik-online:
@theScrabi Still work in progress. I want to display consistent relative dates because if the feed was updated 2 hours ago and a video was uploaded 5 hours before, then the feed must show '7 hours ago'.
And @mauriciocolli is gathering the time units in many languages.
Nearly done.

theScrabi:
@wojcik-online sorry for replying so late, I lost side for this PR.

What is the status of this issue?
Please also fix the conflicts :)

TheGoldmasterCorporation:
i have this issue myself on the what's new page where you have a video that's 2 days old above a video that's just been uploaded? it should sorted by date added 1st at the top.

Grammost:
This is my most wanted feature in NewPipe, but sadly it seems like @wojcik-online is not working on this PR anymore. If that is the case, it would be great if someone else continues to work on it.

hqm42:
Has someone thought about using the rss feed for this kind of information? E.g.

https://www.youtube.com/feeds/videos.xml?channel_id=UCckETVOT59aYw80B36aP9vw

theScrabi:
Redirecting to an external reader is suportet for quite some while. However it was cool if we had this directly in this app.

Somethingweirdhere:
@theScrabi Can you make this a new branch? It seems like the only way we could contribute is making a new pull request. I am not too familiar with git though so I might be wrong.

theScrabi:
You clould pull the changes of @wojcik-online into your own repository, we don't necessarily need a new branch.

DevFelixDorn:
This feature wasn't did ?

sentriz:
[invidious](https://github.com/omarroth/invidious) seems to sort subscriptions chronologically. maybe it's [api](https://github.com/omarroth/invidious/wiki/API) could be used?

PR #1067, opened 2018-02-03
Subtitles and Aspect Ratio Switch Support 

karyogamy:
Here's a movie showing this PR in action because I'm too lazy to write a description:

As you can see, we've ran out of room to put the buttons =D 
If anyone has a good UI solution, please don't hesitate to suggest it.

Things to do:

- [x] Improved UI
- [x] Remove unstable aspect ratio options from popup player
- [x] Have subtitles branch merged in extractor and change gradle to use main branch
- [x] Move subtitles dropdown creation into exoplayer track change callback
- [x] Extract hard-coded strings
- [x] Add options to allow different caption font size

Suggestions and feedback welcomed.


speednroll:
wow, this looks amazing !!I can't wait to see this project implemented inside newpipe :sunglasses:

You could make these two new options are within the three points, to make it more orderly

theScrabi:
Awesomest :D

Could you put the two new menus into the main menu, I don't like that they are so deep in the screen.

wb9688:
For UI, maybe something like what RedReader does would work.

karyogamy:
I've thought about the UI a little bit and decided to go with this since adding a secondary menu for caption looks cumbersome. Please let me know what you think.

theScrabi:
Yea that could be something. How does it look like when its in landscape mode?

karyogamy:
Here you go:

theScrabi:
hm ok :)

theScrabi:
All right thanks :)

SkyzohKey:
@karyogamy Could you add an option in the settings to remember the aspect ratio please ? ;3

speednroll:
some videos don't work with the aspect ratio option, has this been reviewed?

karyogamy:
@SkyzohKey Looks like a good idea, probably suitable as a `good first issue` too. Same thing with the caption language. Please open a new separate issue for this.

@speednroll The aspect ratio button is just enabling a feature that's available in ExoPlayer. So if it fails, there's probably not much we can do about it. But my guess at what you are seeing is the video itself have black bars at the border and there really isn't much about that can be done. Can you provide the url so I can take a look at what you are saying?

speednroll:
it happens only with this kind of videos: [link](https://www.youtube.com/watch?v=juR0WfvA6xc)

It works without any problem if the screen is vertically, but the video horizontally has almost no black spaces, except a little on the sides of the top and bottom. when the videos are horizontal they don't adjust or zoom

karyogamy:
I see, yeah that's expected, since the video aspect ratio fits your screen perfectly the options don't do anything.

speednroll:
you can't make it detect the type of size of a video and give different options for each type? or what I propose is very complex?

karyogamy:
Not too complex but we do have to take care of different cases of video and screen aspect ratio changing. Also you have to think of each case and provide a suitable list of options. I just don't see how much value doing this would providing, other than add more code for maintanence. If anyone is interest in creating a PR for this, they are welcome to do so.

Issue #1072, opened 2018-02-06
improvement proposal: control media volume, not ringer volume


ksaerdna:
If you place this call in your activity's _onCreate()_ method:

`
        setVolumeControlStream(AudioManager.STREAM_MUSIC);
`

the volume keys of the device will change media volume, not ringer volume. This is extremely helpful for any app dealing with media playback and works perfectly in all my Android apps.

PR #1077, opened 2018-02-07
Revert "Control media volume with the hardware buttons."

theScrabi:
Reverts TeamNewPipe/NewPipe#1076

Is not compiling.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -22,7 +22,6 @@ package org.schabi.newpipe;
 
 import android.content.Intent;
 import android.content.SharedPreferences;
-import android.media.AudioManager;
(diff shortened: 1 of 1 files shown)


theScrabi:
I merged faster that I should.
Acording to travis this is not compiling.

wojcik-online:
It does. I simply forgot to add the import on the first commit.

michaelblyons:
Can you un-revert if the import fixes the compile issue? I'd really like to have NewPipe control the media volume specifically, rather than the main system volume. It really screws up my ringer volume preferences.

cc #1072 https://github.com/TeamNewPipe/NewPipe/commit/fddcade1fbf60c2988ce515160b40ae0810df8ed

Stypox:
Do not resume old closed PRs/issues. Even if the code worked 5 years ago, today it wouldn't be usable at all today anyway.

PR #1082, opened 2018-02-08
Leak Canary

karyogamy:
This PR adds LeakCanary for detecting memory leaks.

As you can see, we have many things that needs patching up. Even though these leaks are small, they can quickly add up on older devices. The only downside to this is the dependency dumps the heap from time to time and will hog the app for a few seconds, hence why it is debug build only.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -397,5 +397,7 @@
     <string name="playlist_delete_failure">Failed to delete playlist</string>
 
     <!-- Debug Only -->
-    <string name="toggle_leak_canary">Leak Canary</string>
+    <string name="toggle_leak_canary">Monitor Leaks</string>
+    <string name="enable_leak_canary_notice">Memory leak monitoring enabled, app may become unresponsive when heap dumping</string>
+    <string name="disable_leak_canary_notice">Memory leak monitoring disabled</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -395,4 +395,7 @@
     <string name="playlist_add_stream_success">Added to playlist</string>
     <string name="playlist_thumbnail_change_success">Playlist thumbnail changed</string>
     <string name="playlist_delete_failure">Failed to delete playlist</string>
+
+    <!-- Debug Only -->
+    <string name="toggle_leak_canary">Leak Canary</string>
 </resources>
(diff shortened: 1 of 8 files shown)


theScrabi:
All right sounds cool tho. If it causes trouble in the debug build maybe we should consider to add a flavour or build type only for memory leak?
Would that be a good idea?

karyogamy:
Sure, though adding another build flavor might incur more maintenance in the future, since we have 3 different builds already.

Looking through the doc some more, there is a way to toggle on and off the heap dump. Maybe we can create a button on the debug build for this. Would that be an viable alternative?

theScrabi:
> might incur more maintenance

true

> Maybe we can create a button on the debug build for this. Would that be an viable alternative?

Absoloutly. That sounds like a good idea :)

karyogamy:
This should do it.

app/src/main/res/values/strings.xml
```diff
@@ -365,4 +365,7 @@
     
     <string name="preferred_player_fetcher_notification_title">Getting info…</string>
     <string name="preferred_player_fetcher_notification_message">"The requested content is loading"</string>
+
+    <!-- Debug Only -->
+    <string name="toggle_leak_canary">Leak Canary</string>
```
TobiGr: That's a name, so it shouldn't be translatable.

karyogamy: Oops. In this case, perhaps we should change it to "Monitor Memory Leaks"? Or "Monitor Leaks" if that's too long?

TobiGr: Yes, that's better. I'd go with the second one. It also helps the user to understand what he's enabling :)

karyogamy: Fixed.

mauriciocolli:
I think it's better to stick to just "LeakCanary", as the translations [tend to be quite long](https://hosted.weblate.org/translate/newpipe/strings/en/?checksum=5a16931bc1f3f342#translations) and it makes the menu ridiculously large:




Also "LeakCanary" it's the 1st result on Google, so it shouldn't confuse people who use the debug flavor.

karyogamy:
@mauriciocolli Dang. Okay, will be fixed in an upcoming player fixes PR, should be ready in a couple hours.

theScrabi:
I just saw the menu point for memory analysis is also available in beta. There however it should be removed, or at least put into the settings activity.

PR #1092, opened 2018-02-11
move download menu item into detail controls menu

theScrabi:
As seen here:
screenshot_20180211-213341

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -147,6 +147,7 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo> implement
     private TextView detailControlsBackground;
     private TextView detailControlsPopup;
     private TextView detailControlsAddToPlaylist;
+    private TextView detailControlsDownload;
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -335,6 +336,24 @@ public void onClick(View v) {
                             .show(getFragmentManager(), TAG);
                 }
                 break;
+            case R.id.detail_controls_download:
+                if (!PermissionHelper.checkStoragePermissions(activity)) {
```
karyogamy: Might be a good idea to move this out to a method and do a few null checks. `currentInfo` is async and can be null.

app/src/main/res/values/strings.xml
```diff
@@ -13,6 +13,7 @@
     <string name="open_in_popup_mode">Open in popup mode</string>
     <string name="share">Share</string>
     <string name="download">Download</string>
+    <string name="controls_download_desg">Download stream file.</string>
```
karyogamy: Should be `controls_download_desc`?

theScrabi: fixed.

PR #1109, opened 2018-02-16
Enable SoundCloud kiosks as main page fragment for debug and beta

### LINKED ISSUE #1107: NewPipe Beta 0.11.6
SoundCloud trends are the same as YouTube trends :/

TobiGr:

Closes #1107

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/SelectKioskFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/SelectKioskFragment.java
@@ -13,6 +13,7 @@ import android.view.ViewGroup;
 import android.widget.ImageView;
 import android.widget.TextView;
 
+import org.schabi.newpipe.MainActivity;
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.ServiceList;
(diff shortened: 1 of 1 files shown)


wb9688:
I'd show the first kiosk of the service by default, but have some chips at
the top to switch to another kiosk of the service (and then remember the
last used kiosk). I'd then remove the kiosk menu from the menu and on the
settings page I'd just have a selection for kiosk (and not for the specific
kiosk like now).

theScrabi:
> have some chips

I don't know. they would take away a loot of screen space. When we are done with the drawer we don't nee the three dots menu in the main screen any more. So we can dedicate them completely to the kiosk selection.

> and then remember the last used kiosk

Yes

> I'd then remove the kiosk menu from the settings page

definetly

TobiGr:
> When we are done with the drawer we don't nee the three dots menu in the main screen any more. So we can dedicate them completely to the kiosk selection.

I agree with that.

The whole PR was meant rather as a quick fix than providing a good solution for later. So if you want to refactor the kiosk methods in the next days, feel free to do it. I don't have time for it, because I have two exams left and the website has a higher priority for me since it needs some major improvements.

theScrabi:
> I don't have time for it, because I have two exams

I know this :D. I barely finished my exams now i actually have no time since i move to Spain in two weeks.

TobiGr:
> PS: Good luck ;D

Thanks! Moving to Spain sounds interesting :D

PR #1120, opened 2018-02-18
Player Stability Fixes

karyogamy:
- Updated ExoPlayer to 2.6.0. 
- Added view count registration on repeats.
- Added drag reorder speed clamping to play queue list.
- Moved LeakCanary toggle into new debug preference fragment (visible to non-release builds)
- Added undeliverable Rx exception swallowing toggle to debug preference.
- Fixed service player activity memory leak.
- Fixed media source manager sync disposable fallthrough causing uncaught exceptions.
- Fixed thread bouncing during play queue item async stream resolution.
- Fixed video player source loading for audio only streams.
- Renamed "Monitor Leak" to "LeakCanary" as untranslatable.

Addendums:

- Exoplayer is not updated to latest (2.6.1) since the newer version [changed how preferred language strings are normalized](https://github.com/google/ExoPlayer/commit/c1c892f5ec5546257d1da1b125fc4c3466daede2) (ISO 639-2, which does not distinguish language variants, e.g. zh_TW and zh_HK) and broke subtitles by returning nulls when irregular language strings are passed in. This means we may now need to start rolling our own track selector to keep the player up-to-date and will be the next thing I work on.

- This is just something I've observed under stressed memory usage and slowdowns: RxJava (network async) observables that are bound to fragment lifecycles can still be running and fail after the fragment is destroyed and the observable disposed. The exception from failure can trigger an uncaught exception which does not show much usable traceback. This happens mostly with `ExtractionException`, which makes it quite tempting to add it to global Rx exception handler alongside the unhandled network exceptions. @mauriciocolli What do you think we should do with this kind of exceptions?


mauriciocolli:
> which makes it quite tempting to add it to global Rx exception handler alongside the unhandled network exceptions

This is something that has been bothering for a while, and I'm not quite sure what to do because of my limited knowledge/experience with RxJava.

Anyway, if we think about it, the only way to an exception end up in the "main" error handler is:
- Either it has been disposed already and the error couldn't be delivered
    - In NewPipe this happens mainly because RxJava interrupts the thread we were blocking reading from the network, which throws `Interrupted(IO)Thread`
    - Sometimes it throws another exception in some other part of the extractor because of that
- Or it was subscribed without an `errorHandler`

It is mentioned on the section "error handling" of the RxJava's wiki (which I linked in the current implementation), so maybe swallowing `UndeliverableException` is our use case after all, as we don't care about things that we already disposed.

I'll do some tests in the next few days and see if it have some side-effects.

PS: About the string, some languages already have it translated, you should consider to remove them as well.

karyogamy:
I read through the doc again and it seems like `UndeliverableException` should indeed be swallowed in our case, since `OnErrorNotImplementedException` takes care of the case without error handlers. Perhaps we can approach it like LeakCanary and have different build flavors have different handler behaviors (e.g. release swallows and debug retains the same behavior as now). Or we can have swallowing toggleable like the heap dumper.

> About the string, some languages already have it translated, you should consider to remove them as well.

Done.

karyogamy:
@theScrabi New request on #1082 to merge leak canary toggle into settings done. Also added a new fragment that is only visible to non-release builds, which seems to be a better solution for hiding the toggle and new debug settings in the future. 

By the way, I hope we are not completely getting rid of the three-point app menu, since we can put some convenience functions in there, such as one-click disable history, in the future, when need comes ;)

theScrabi:
> By the way, I hope we are not completely getting rid of the three-point app menu

all right, I keep that in mind.

mauriciocolli:
I'd just forget about it and just ignore the `UndeliverableException`, some reasons below.

> I've just discovered that if an exception occurs in onSuccess it would also be thrown at the global Rx exception handler [...]

We don't need to handle exceptions from `onSuccess`, because as the name says, it only handles success cases, and if something can throw (checked or known) exceptions, maybe that's not its place. 

Also, any bug that eventually can occur inside it will ended up in [`RxJavaPlugins.onError`][1], you can notice that it calls [`isBug()`][2] before wrapping, so yes, we can be sure that actual bugs will be noticed.

**PS:** This also applies to other terminal functions (`onComplete, onError...`) of other classes too (`Observable, Flowable, Maybe...`).

[1]: https://github.com/ReactiveX/RxJava/blob/99a620a38c38e5010dc5bcc338a13555b3ccb727/src/main/java/io/reactivex/plugins/RxJavaPlugins.java#L342..L366
[2]: https://github.com/ReactiveX/RxJava/blob/99a620a38c38e5010dc5bcc338a13555b3ccb727/src/main/java/io/reactivex/plugins/RxJavaPlugins.java#L376..L406

karyogamy:
> We don't need to handle exceptions from onSuccess, because as the name says, it only handles success cases, and if something can throw (checked or known) exceptions, maybe that's not its place.

I agree with what you are saying. My only worry is that, without guidelines, future devs might not know that exceptions should not be thrown in terminal functions. It is easy to just code away and not realize something can throw an exception, especially with lambdas, and it is equally difficult to spot this in code reviews. When we swallow this kind of error, the user can experience the app misbehaving but have to now dump the log manually in order to report the problem. Perhaps on another PR we can introduce a snackbar just like the non-critical extraction errors on info list fragments, let me know what you think.

Since we are swallowing these exceptions, I've updated the default behaviour so release build will filter these exceptions out and non-release builds opt-in. Please review. 

Regarding the deprecated translated strings in xml, please notify me to remove them before we merge, since I don't know when weblate is going to push more of these.

theScrabi:
I agree with @karyogamy. We should be aware of any case of occurring error.

PR #1102, opened 2018-02-15
Resolved conflicts with dev branch for new player

avently:
I really hope that you guys, @theScrabi @karyogamy @mauriciocolli @wb9688 will base your future work on the /embed branch. That conflict resolving was like a hell. Mostly because layout conflicts since popup and main players use one layout. 

I didn't tested all new functionality but what i tested is working.

  @karyogamy i have a question. I added screen resizing to popup but couldn't get any difference between the original resize mode and others. Does it work only on limited number of videos? For main player i see the difference.

- [ x ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #1145, opened 2018-02-25
Implement subscriptions import/export

### LINKED ISSUE #698: Subscription Xml Support

So I was able to successfully build NewPipe with subscription support and was wondering if it would be possible to add xml support. What I mean by this is at the bottom of the [subscription manager page](https://www.youtube.com/subscription_manager) there is an "Export to RSS readers" button. I was wondering if it would be possible to import subscriptions that way since the current method takes forever.

mauriciocolli:
- Import subscriptions from YouTube and SoundCloud (all services that the extractor support)
- Includes instructions for each service (in the import fragment/screen)
- Import/export a JSON representation of the subscriptions
- Remove usage of some `javax` annotations in favor of the one provided by the android support library

### Import/export UI

import_export

### Performance

I used a conservative number for extractions in parallel: 8, which in my tests resulted as the more balanced in terms of speed, memory usage and CPU usage.

- In an average phone, it imported 200 subscriptions in about 30-45s
- In the emulator, it imported 200 subscriptions in about 15-25s

---
Closes #698


theScrabi:
All right, that's some huge work :), but why is the import export datatype json? Wouldn't it be better if we supported some more common file formats like [OMPL](https://en.wikipedia.org/wiki/OPML)?

avently:
Sorry for disturbing you in your PR. Do you know about theScrabi's question in #1102 ? I'm waiting your answer but I'm not sure you know about a question. Do you?

karyogamy:
Impressive work =D 

One thing I saw is that the import operation is network intensive. Maybe it would be a good idea to have a confirmation dialog to let the user know before starting the import?

Thanks for cleaning up some of the database stuff. I'll find some time to look into this PR further.

mauriciocolli:
###### Confused and force pushed the head of the dev into my fork, automatically closing this PR.

---

> Wouldn't` it be better if we supported some more common file formats like OMPL?

Is it just because YouTube use it or there are real advantages? Just preference? Some other software would use it?

> Maybe it would be a good idea to have a confirmation dialog 

Done.

theScrabi:
> Is it just because YouTube use it or there are real advantages?

Many people requested to have the ability to migrate the subscriptions from youtube over to NewPipe.

> Some other software would use it?

That to. Its a common file format. This way you could put subscriptions into other apps like ttrss as well.

Maybe we could add OMPL later. This seems like a good base for it.

wb9688:
Isn't OPML just for holding a list of RSS feeds? AFAIK only YouTube has a
RSS fees for a channel, for example SoundCloud doesn't.

mauriciocolli:
> Many people requested to have the ability to migrate the subscriptions from youtube over to NewPipe.

I don't get it, the subscription extraction is done in the extractor. The JSON implementation is just to import/export the already existent ones, that way you can transfer between devices (and other things).

Did you tried to build and use it?

@wb9688 brought a good point that we are not saving the the RSS feed url, just the `serviceId` and the `url` of a given channel, so it shouldn't matter which format we use (although we _could_ save for the services that support it, that would be somewhat inconsistent).

Also, if we implement a better import/export (of the entire app: history, search, subscriptions, playlist) we could save all of it in some format and have finer control on what specifically we want to import/export, then this could be dropped. In the mean time, it's a nice feature/ability to have.

theScrabi:
All right all right. I get it :)

app/src/main/java/org/schabi/newpipe/subscription/services/BaseImportExportService.java
```diff
@@ -0,0 +1,203 @@
...
+    public abstract int getTitle();
+
+    protected void setupNotification() {
+        notificationManager = NotificationManagerCompat.from(this);
+        notificationBuilder = createNotification();
+        startForeground(getNotificationId(), notificationBuilder.build());
+
+        disposables.add(notificationUpdater
+                .filter(s -> !s.isEmpty())
+                .publish(stringFlowable -> stringFlowable.limit(1)
```
karyogamy: Took some eye squinting to get what this is doing. Might be a good idea to extract this and name it `notifyFirstThenThrottle` or something more descriptive.

app/src/main/java/org/schabi/newpipe/fragments/subscription/SubscriptionFragment.java
```diff
@@ -103,9 +131,110 @@ public void onDestroy() {
...
+        if (supportActionBar != null) {
+            supportActionBar.setDisplayShowTitleEnabled(true);
+            setTitle(getString(R.string.tab_subscriptions));
+        }
+
+        inflater.inflate(R.menu.menu_subscriptions, menu);
+        setupImportSubMenu(menu);
+    }
+
+    private void setupImportSubMenu(final Menu menu) {
```
karyogamy: Just a thought, but wouldn't be better if we move the import/export menu into the fragment as part of the list header (along with `What's New`)? Right now, this feature might not be easily discoverable, as users are probably used to the menu staying the same across all pagers.

app/src/main/java/org/schabi/newpipe/subscription/services/BaseImportExportService.java
mauriciocolli: I've chosen `throttleAfterFirstEmission`, I hope that it's easier to understand.

mauriciocolli:
Updated the subscriptions import/export UI, see the [main post](https://github.com/TeamNewPipe/NewPipe/pull/1145#issue-171237220). 

Let me know what  you think.

justanidea:
@mauriciocolli I found it pretty clean !

karyogamy:
Nice animations, I like it :D I was thinking about doing something like this so we can display descriptions on background players. 

The collaps/expand icon seems to have too little margin to the right compared to import/export icon on the left. Might want to fix that.

Also, I'm not sure how scalable the UI will be when we have import/export for many more services. But I suppose we can worry about that later.

mauriciocolli:
> have too little margin to the right

Fixed.

> I'm not sure how scalable the UI

I'd say that, to the services that support it and we planned to have implemented in the near future, it handles reasonably well:



If we eventually have 5~8+ than that, then yes, we have to rethink it (maybe a grid?).

speednroll:
I vote to do it in the form of grid. with horizontal scroll to see all the services (in case there were too many)

drfunkencluck:
Has that ui been added to the newer version on github? So now we can import and export?

mauriciocolli:
@drfunkencluck yes (not just the UI, of course).

I guess there will be a beta before a stable release so we can how it performs.

drfunkencluck:
is there a apk yet? please add the apk download link. then i can see if there are any issues

theScrabi:
we might want to consider to release 0.13 soon. @mauriciocolli however id like to wait until ur work on the extractor is done :)

Issue #1103, opened 2018-02-15
Improve resolution selection


ghost:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

As follow-up of https://github.com/TeamNewPipe/NewPipeExtractor/issues/71, we should add a better resolution selector. It should only show resolutions and choose the best streams for that resolution that the device supports (like YouTube does). We should also add adaptive streaming (e.g. YouTube's Auto resolution). But we should also have an advanced option to choose the specific streams manually.

speednroll:
it's a great idea! especially because I hope this can correct some errors, like watching a video when you have a certain resolution and the video is not compatible with that resolution, it gives an error like "the video is not available". change the resolution to a lower one automatically than if it is available

theScrabi:
Ok souds good.

karyogamy:
There was an attempt on a similar concept in the [play queue PR](https://github.com/TeamNewPipe/NewPipe/pull/680). Though this was done by merging all available stream sources and use ExoPlayer's default track selector to do the work. However, this can have a huge overhead on the network, since Exoplayer wants to load all the headers to make the track switching faster. Also, the default track selector doesn't pass the metadata from extractor internally, so the actual mimetype and framerate data is lost and only 60fps is chosen. We can probably roll our own solution by extending Exoplayer, though I imagine that is going to require some hefty research and maintainance in the future.

We can also use the dash manifest from the extractor, but this is not a universal solution, even on Youtube. Apparently, these manifests are [uncommon](https://github.com/rg3/youtube-dl/issues/3763), and after testing on Youtube-dl, there seems to have manifest for audio only streams in the videos I have tried. 

Both solutions requires the same change to the players, mainly on track selection. The same mechanics are used for subtitles already, so it shouldn't be too difficult implement (this will need to be done eventually due to livestream uses hls manifests). But if we want to continue support for resolutions under both 24 and 60 fps, then it might take some more effort.

litetex:
Closed due 
* to inactivity
* already implemented: Best resolution is already implemented

PR #1147, opened 2018-02-26
Live Streaming and ExoPlayer Shenanigans

### LINKED ISSUE #1139: NewPipe start playing music again after beeing paused if a SMS is received

1. Listen to a music on NewPipe with headphone
2. Screen off, remove the headphone so the music is paused automatically
3. Screen off, receive a SMS (with sound)

Actual :
The NewPipe music is playing again after the SMS sound stops.

Expected : 
The NewPipe music stay paused.

### LINKED ISSUE #1151: Don't show overlay after resuming from buffering
If I'm watching a video and it needs to buffer, the overlay (play/pause, seek bar) keeps appearing over the top of the video after it resumes.
It's annoying having to press the screen to make it disappear.

karyogamy:
- Added cacheless data source to allow HLS livestream playback.
- Added new async loader for media sources to correct HLS livestream loading and enable better control over expired media sources. 
- Added custom cache key for extractor sources to allow more persistent reuse.
- Added seemless shuffling.
- Added toggleable fast inexact seeking.
- Added toggleable auto-queuing when last stream on non-repeating play queue starts.
- Added custom track selector to allow irregular caption text language names, until language normalization NPE in ExoPlayer is fixed. 
- Added buffering strategy to enable faster playback start.
- Fixed inconsistent audio focus when audio becomes noisy through unplugging headsets. Fixes #1139.
- Fixed search fragment long press dialog deletes item prior confirming.
- Fixed potential `TransactionTooLargeException` when passing player intents.
- Fixed main video player state saving on lifecycle events.
- Updated ExoPlayer to 2.7.0.
- Updated video player UI to no longer show interface after buffering and double-tap seeking. Fixes #1151.
- Updated main video player UI to use hidden translucent system UI (nav and status bars), similar to Youtube app.
- Updated Extractor to allow HLS extraction.
- Removed livestream exceptions.
- Removed debouncing time for stream selection on play queue.

Addendum:
- HLS playlists really doesn't like loading and playing on demand, which means we can no longer use the existing `DeferredMediaSource`. Therefore, the old loader is being replaced by a new async loading inside the `MediaSourceManager`. 
It works by first generating a list of placeholder media sources in order to create a playlist media source. When these placeholder are played, they stall the player until they are replaced with a loaded `MediaSource`. When loading is requested, outside the lifecycle of `MediaSource`, the `MediaSourceManager` loads and builds the `MediaSource` asynchronously and replaces the `PlaceholderMediaSource` at the corresponding index on the playlist media source. If the loading fails for a particular media source, then a `FailedMediaSource` will be used for replacement. 
By doing this, HLS livestream is now able to load the playlists properly and we can now have better control on the playlist to evict or update `MediaSource`s with expired urls.

- Regarding the `CustomTrackSelector`, it is mostly a copypasta from `DefaultTrackSelector`. A few lines are changed in order to allow us to use irregular caption language names. This will be removed once ExoPlayer properly nullchecks their language code normalization.

- Currently, I have disabled track selection on livestreams since the manifest allows ExoPlayer to adaptively choose the quality according to the network status. Of course, this adaptiveness can be overriden if needed, but this requires us to create a separate quality menu just for livestreams, which will make the video player more complex. Let me know what you think. 

Todos:
- [x] Add UI for livestreams to seek directly to live edge.
- [ ] Change extractor version to latest once merged.


karyogamy:
A couple updates:
- The player load control is now updated to almost immediately start playback once the buffer is ready. There should be a noticeable performance gain on lower resolutions or a stable connection. After the initial playback start, the buffer should retain the current behaviour of zigzagging between the minimum and maximum buffer time. 
I've also experimented with forcing a constant window of buffer, but this may cause the network radio to constantly turn on or off or stay on for the entire duration of the playback, which may lead to more battery consumption. If anyone has a better buffering strategy, please take a look (here)[] first and let me know.

- The debouncing duration for selecting video or clicking on previous or next button has now been remove. That's 400ms of performance gain, and the user can decide how much they want to load. This reminds me of a story where the project manager tells the programmers to leave some `Thread.sleep` here and there in the code, so that whenever they miss a deadline, they can just remove one of the sleep and call it a performance update =P

- Changed up the UI for video players and removed the system UI (status and navigation bar) after Kitkat. This should make the main video player a bit more like the Youtube app design. However, the code for this is somewhat confusing, so please be sure to go over (it)[]. Though every time I have to deal with api levels, I feel a little more dead inside. #1151 is also included as part of this commit.

karyogamy:
More updates:
- Added seamless shuffling. Now the only sources of non-seamless playback is buffering, player exception and quality switching. Nothing we can do about the first two, and the third is probably non-trivial considering #1103. However, this is a major tech debt that we will have to address eventually, since the current quality selector implementation is still quite hacky.

- Added fast inexact seeking. It might be syncing to the nearest keyframe and off by a few seconds, but boy is it fast. This is another feature that is available in ExoPlayer, which we can simply enable. A setting toggle has been added to let users try it out.

karyogamy:
It's disabled by default.

wb9688:
Oh, I didn't see that part, thanks!

app/src/main/java/org/schabi/newpipe/util/InfoCache.java
```diff
@@ -52,25 +55,27 @@ public static InfoCache getInstance() {
...
 
-    public void putInfo(@NonNull Info info) {
+    public void putInfo(int serviceId, @NonNull String url, @NonNull Info info) {
         if (DEBUG) Log.d(TAG, "putInfo() called with: info = [" + info + "]");
-        synchronized (lruCache) {
-            final CacheData data = new CacheData(info, DEFAULT_TIMEOUT_HOURS, TimeUnit.HOURS);
-            lruCache.put(keyOf(info), data);
+
+        final long expirationMillis;
+        if (info.getServiceId() == SoundCloud.getServiceId()) {
```
karyogamy: @theScrabi
I was hoping to receive some more review and discussions before having the PR merged, since there might be parts that would lead to tech debts.
For example, this part where different services can have different cache expiration. SoundCloud was added here since their API seems to fail sporadically and would cause the `StreamInfo` to have bad stream urls and need to be reloaded. 

This probably should be part of the extractor, or at least `ServiceHelper`. Please consider removing this.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -83,6 +98,58 @@ public static String resizeTypeOf(@NonNull final Context context,
         }
     }
 
+    @NonNull
+    public static String cacheKeyOf(@NonNull final StreamInfo info, @NonNull VideoStream video) {
+        return info.getUrl() + video.getResolution() + video.getFormat().getName();
```
karyogamy: Just for the record. The cache key generation here might be too naive. Hopefully someone can find a better alternative, such as generating the `VideoStream` or `AudioStream` identifier in the extractor.

oceanwaves90:
Hi! Thank you for your works!

By the way, at times the sync of live stream doesn't work. It plays a certain interval of the stream and so the sync button plays always from the same point of the stream. The live stream is [this](https://youtu.be/uqU5iDCZdiw), which has been broadcasted for several days.

New playing also repeats the above problem. It starts from the very point of stream. It never goes to the present live streamed point because the playing ends at the point of 2 hours.

And the live-watching number isn't uptated in subscription window and in the detail window, while the live-watching number of the browsed result of the stream is correct.

Do the problems have to do with caching the stream data?

karyogamy:
@oceanwaves90 
You are right, this is mostly a problem with the cache. We are currently caching the data for the latest 60 Youtube page accesses for 4 hours. So if anything on a cache page updates in the real time, it's likely they aren't reflected. `System.exit` probably obliterated that cache. Also, I don't think live-watching number is even displayed right now, since this requires constant web requests to the stream url which is about 200kb each, and can adds up quickly. 

I also reproduced the sync button problem on the url you gave. I think this is probably due to the livestream being renewed on the uploader end, causing the livestream manifest to automatically transition into a normal stream, as I had no problem streaming [this](https://www.youtube.com/watch?v=XgRjcfKLRdQ) on live-edge for about 8 hours. So, unless this problem becomes rampant, I'd avoid writing extra logic to deal with this kind of cases, since we don't know when a stream url becomes stale.

However, now that we have more dynamic data with livestreaming, we should consider expiring Youtube stream page data sooner, as well as adding a quick button (on the top right dropdown) to let users wipe the cache manually.

oceanwaves90:
I see! That will be great! I'll look forward to the next updates!

app/src/main/java/org/schabi/newpipe/util/InfoCache.java
theScrabi: I see, I merged early, so now we have to take care about this in the aftermath.

> For example, this part where different services can have different cache expiration. SoundCloud was added here since their API seems to fail sporadically and would cause the `StreamInfo` to have bad stream urls and need to be reloaded.

So you`d suggest a new field in `StreamInfo` that tells the URls expiration time? If so we should add this.

>  Please consider removing this.

You mean the part that checks for SoundCloud? Right thats a hack and needs a proper solution.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
theScrabi: So you think we should distinguish between audio and video data as well?

app/src/main/java/org/schabi/newpipe/util/InfoCache.java
karyogamy: I'm currently working on a PR that lets user wipe the `InfoCache`, so we can remove this check. I've also reduce the default cache time to 1 hr instead of 4. I've also moved that check to `ServiceHelper` for now and reduce the cache time for SoundCloud to 5 minutes.

However, this is not a long term solution, as SoundCloud stream url seems to consistently expire after a few minutes if they are not played, at least from my side. We will need to figure out exactly why this is happening, since it's probably a bad idea just slapping on an arbitrary expiration time for each service.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
karyogamy: Sure, I think each normal stream should have a unique id they can use regardless of their url, since it changes all the time.

app/src/main/java/org/schabi/newpipe/util/InfoCache.java
theScrabi: Well we might have similare problems with YouTube as some issues point out. But you are right, we should investigate this.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
theScrabi: Yes that makes better sense.
This is why the UrlIdHandlers exist. Maybe these should be used at this point.

PR #1178, opened 2018-03-10
Media Button Play/Pause, Previous and Next in Background Player

alexandrepa:

I've just added a MediaButtonReceiver which handles Play/Pause Next and Previous actions in the background player.
So basically now you can control the player from headset buttons which I found really conveniant.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
@@ -22,10 +22,13 @@ package org.schabi.newpipe.player;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.app.Service;
+import android.content.BroadcastReceiver;
+import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.graphics.Bitmap;
+import android.media.AudioManager;
 import android.os.Build;
 import android.os.IBinder;
 import android.support.annotation.IntRange;
@@ -33,6 +36,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v4.app.NotificationCompat;
 import android.util.Log;
+import android.view.KeyEvent;
 import android.view.View;
 import android.widget.RemoteViews;
 
@@ -77,6 +81,10 @@ public final class BackgroundPlayer extends Service {
 
     private BasePlayerImpl basePlayerImpl;
     private LockManager lockManager;
+
+    private AudioManager mAudioManager;
+    private ComponentName mReceiverComponent;
+
     /*//////////////////////////////////////////////////////////////////////////
     // Service-Activity Binder
     //////////////////////////////////////////////////////////////////////////*/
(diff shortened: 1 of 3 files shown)


justanidea:
Thank you so much !

karyogamy:
Nice work, thanks. One thing I saw is that the `AudioManager` is created again, but is already available in the [`AudioReactor`](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/helper/AudioReactor.java). So it might be a good idea to add the register and unregister methods to that and call those methods instead.

Also, I must remind you that registering media buttons on the `AudioManager` is deprecated as of SDK 21, so this feature may break at any time in the future without a proper `MediaSession` implementation. @mauriciocolli What do you think about this?

speednroll:
Some time ago I was waiting for this option in Newpipe, hopefully soon they can fix the small mistakes and added to the new release!

Thank you very much to the Newpipe team for doing an incredible job!!

alexandrepa:
Based on what you said @karyogamy, it uses now the AudioManager from AudioReactor.
I know that this way of receiving media buttons is deprecated, but I think the only legit way now is by using MediaSession which seems to be not used currently in the player. (I'm not an android expert)

app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
```diff
@@ -151,6 +160,8 @@ private void onClose() {
         basePlayerImpl = null;
         lockManager = null;
 
+        basePlayerImpl.audioReactor.unregisterMediaButtonEventReceiver(mReceiverComponent);
```
karyogamy: This will definitely cause a NPE, `basePlayerImpl` is set null on line 160.

app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
```diff
@@ -563,4 +574,37 @@ public void onCompleted() {
...
+        }
+
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            if (Intent.ACTION_MEDIA_BUTTON.equals(intent.getAction())) {
+                KeyEvent event = (KeyEvent) intent.getParcelableExtra(Intent.EXTRA_KEY_EVENT);
+                if (event.getAction() == KeyEvent.ACTION_UP) {
+                    int keycode = event.getKeyCode();
+                    PendingIntent pendingIntent = null;
+                    if (keycode == KeyEvent.KEYCODE_MEDIA_NEXT) {
```
karyogamy: Might be a good idea to also add fast forward and rewind, since some media buttons support those when long-pressed.

app/src/main/java/org/schabi/newpipe/player/helper/AudioReactor.java
```diff
@@ -86,6 +87,14 @@ private boolean shouldBuildFocusRequest() {
         return Build.VERSION.SDK_INT >= Build.VERSION_CODES.O;
     }
 
+    public void registerMediaButtonEventReceiver(ComponentName componentName) {
+        audioManager.registerMediaButtonEventReceiver(componentName);
```
karyogamy: We should add a check to disable this (and also log an error) after Oreo, since we don't know if this will be removed in the future.

alexandrepa:
Thanks for the quick response ! I've made fixes based on your review. I hope it's better now !

theScrabi:
> @theScrabi If we are to merge this now, we should at least create an issue that `MediaSession` needs to be implemented soon.

I see, then we should do that :)

theScrabi:
Maybe @alexandrepa you could take a look at implementing media session when you where already working on this issue :)

PR #1182, opened 2018-03-12
v0.12.0

theScrabi:
Things that still need to be done:
- [ ] fix NPE on start when compling as "release"
- [ ] fix NPE because of no banner url give when trying to open playlist


theScrabi:
However in order to accomplish this we may also have to change the workflow on the extractor ass well. :/

karyogamy:
I think the first thing we should do is to rollback the extractor dependency to commit [9dfcb3be](https://github.com/TeamNewPipe/NewPipeExtractor/commit/9dfcb3be068773ab3a97ff1580f687f7760b795a). Beyond that point is the architecture refactoring, which I don't think we should fix in an incompatible release branch. Sure we may miss the fix on offensive content and new features, but I don't think they are required or useful on v12.0.

If we are going to fix the banner exception, then we should create a v12.0 branch on the extractor and a patch for this specific problem, so we can apply it to both the master and v12.0 branch. Then we use the 12.0 branch solely for the release.

theScrabi:
My idea was to merge v0.12.0 into master

karyogamy:
Wait. Just making sure here, do you need a code review on the PR or just an approval to merge so we can fix later?

theScrabi:
Actually I'm not sure my self?
@karyogamy would you suggest to release v0.12.0? Because the more I think about this the more I get the feeling we should skip this version to be release.

However if we release it, or even for releases in the future I would like us all to check weather the version of this branch works stable (do QA) and if not fix thise (cherry-pick). After it works stable we merge into master, and release it.

theScrabi:
serveimage

Let me think I will come up with a better more elaborate idea about how to handle the release flow.

Issue #1186, opened 2018-03-13
Issues with Live Streaming


theScrabi:
I'm sorry I merged the livestream feature before it got tested.
So here are some thoughts and impressions of mine:

- [x] __New instead of Sync__
On the player if you have live stream, on the right side you see the word Sync. However I found it a little more conveniend and intuitive if it was called Now instead.
![sync](https://user-images.githubusercontent.com/1891273/37354284-9381a77e-26e1-11e8-82ae-972f9512e928.png)

- [x] __New streams shouldn't start at the beginning but on "now"__
As a noop I expect a live stream to be "live" when i start it. However it confused me that the live stream starts at the beginning and not at the "now". It took me some time until I realized the picture I saw was not "now", and I had to scroll forward.  However I witnessed that when I lock and unlock the phone we are "live" again.
![start_at_beginning](https://user-images.githubusercontent.com/1891273/37354331-afaa0392-26e1-11e8-8709-dcfa72e44f65.png)

- [x] __Confusing display where we are in the stream__
Also the number showing where we are currently in the stream jumpt arround wildly, even sometimes jumping to a minus value. That confused me.
![confusing_display](https://user-images.githubusercontent.com/1891273/37354353-bf04eea6-26e1-11e8-9486-8964ea1b11a3.png)

Issue #1060, opened 2018-02-02
Disable thumbnails


phly95:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

It would be nice to have the option to disable thumbnails to make the app more light on data. This will be quite helpful for those who are left in slow data after using up thier plan.

speednroll:
Listen, this can be useful... although it should be added to almost everything, including the history and playlists. I think it's going to take a while to implement this if take it into account

phly95:
I wonder if it could be done in a manner that blocks the app from requesting jpg files, therefore taking care of all the different screens in one go?

justanidea:
More over it is great to save data for tiny mobile package

TobiGr:
This has been implemented.

Issue #1129, opened 2018-02-20
Connect with orbot 


herotux:
Please add an option to use proxy or using orbot like fdroid and orfox .
Many users live in countries like iran and govermentes block services and freedom.
Add this for freedom.thanks

theScrabi:
We had that before, but we had to remove it since many data slipt through and where not send via tor. @eightheive back then you said something about making a orbot lib that could redirect all the traffic.
Whats the status about this? Can we use it?

wb9688:
@theScrabi: We'll have to manually add code for supporting SOCKS proxies to the `Downloader` (OkHttp supports them), but AFAIK Universal Image Loader and ExoPlayer don't support SOCKS proxies. We could fork them to add support for SOCKS proxies, but I don't want to do that.

theScrabi:
@wb9688 That's the problem. However I remembered that @eighthave said something about a lib that could somehow intercept the traffic and redirect it so we would not have to take care about that which would make things much easier.. ... Maybe I'm wrong.

wb9688:
@theScrabi: I think he said Orbot has an option to let everything use it transparantly, but that's only possible when you've rooted your device. I'm not 100% sure though.

theScrabi:
Well you can always use vpn even without root, this is already supported by NewPipe.

eighthave:
The NetCipher libs provides easy methods for settng up the proxy, and getting the Orbot status.  You will still need to find all the places where you are using network libs, and apply NetCipher calls.  Universal Image Library works fine with NetCipher, we use both together in F-Droid.  With UIL, just do the NetCipher setup in the downloader class given to UIL.  I know nothing about ExoPlayer.

Perhaps it makes sense to offer a Tor mode, where things that can't be forced over Tor are disabled or hidden.  Downloading and searching always worked, as far as I remember.

wb9688:
When #1136 is merged, it's only one line of code to use a (SOCKS or HTTP) proxy with OkHttp (and some other lines to make it a setting, but you get the point). However the actual playback won't use it then. @mauriciocolli or @karyogamy: do you know if you could set a SOCKS proxy for ExoPlayer?

karyogamy:
[...]

PR #1189, opened 2018-03-14
More Usability Update

karyogamy:
- [#1186] Changed live sync button text to "LIVE".
- [#1186] Fixed livestream not starting from live-edge.
- [#1186] Fixed livestream timeline updates causing negative time position.

@theScrabi I've opted to use "LIVE" instead of "NOW", since this is probably easier to understand when the user is coming from the Youtube app.

- [#1060] Added toggle to disable new thumbnail from loading, if the thumbnail is already in the cache, memory or disk, then they will still be loaded. This should be useful when the device is on mobile data or is low on RAM, since bitmaps add about 10mb+ on memory cache. Currently, this is done by redirecting the image input stream to a dummy stream, causing the image loader to use the default bitmap. If anyone has a better way of doing this, please don't hesitate to comment.

@theScrabi @mauriciocolli I just noticed, when benchmarking the memory, the `background_header.png` and placeholder png for the services takes up about 3mb and 1mb (respectively) of memory when stored as bitmap, as opposed to the 30kb when the are png in the `res` folder. Perhaps should shrink these images down to save RAM usage?

- Added button to wipe metadata cache.
- Added more paddings on player buttons.
- Added new animations to main player secondary controls and play queue expand/collapse.
- Refactored play queue item touch callback for use in all players.
- Improved MediaSourceManager to better handle expired stream reloading.
- Removed MediaSourceManager loader coupling on main players.
- Moved service dependent expiry resolution to ServiceHelper.
- Fixed main player system UI not retracting on playback start on devices before Kitkat.
- Bump support library to 27.1.0 and multidex to 1.0.3.


karyogamy:
Some updates:

- Turns out ExoPlayer has an extension that abstracts most of MediaSession implementations away. So, MediaSession is now implemented on `BasePlayer` and should work across all player. Play queue should now also display on peripherals (Wears OS) as well, though I haven't yet tested that since their emulators are a pain to set up. Probably will try doing it in the next couple days.
- Thumbnail loading have been refactored and should now exhibit the same behaviour across all fragments. Went with fade-in animation since it is quite nice. Also, the image cache should work properly and stay at ~10mb now.
- Thumbnail disabling now uses the dark thumbnail already in `res` for replacement and uses about 11kb per image, so it actually save some RAM as well. However, the toggle now clears the image cache, but this is at least better than the previous hack that fails the image loading to disable it.


@theScrabi Livestream should now always start at the live edge, though it seems to me the startup for livestreams are bit slower now as well. I'll probably look into why this is in a little while. 
@wb9688 Found [this](https://github.com/google/ExoPlayer/tree/release-v2/extensions/okhttp) when going through ExoPlayer's extensions. Might be useful for #1129.

app/build.gradle
```diff
@@ -48,7 +48,12 @@ android {
 }
 
 ext {
-    supportLibVersion = '27.0.2'
+    supportLibVersion = '27.1.0'
+    exoPlayerLibVersion = '2.7.1'
+    roomDbLibVersion = '1.0.0'
+    leakCanaryVersion = '1.5.4'
+    okHttpVersion = '1.5.0'
```
mauriciocolli: Strange usage of this variable name.

Also switching between `"` and `'`.

mauriciocolli:
Some things that maybe you want to fix/improve in this pull request, but just for the record if you don't:

- The duration of the live streams seems a little bit weird for users that don't know the concept of the available "time window" (that when the player is playing it is decreasing/shifting)
- When watching lives with small "available buffer" (can't go more than 30 seconds back) the progress bar don't stay on the end, giving the user the impression that the button "LIVE" doesn't do anything
    - This also happens in lives with greater back support, but the difference is so small (few pixels) that we can't even see
    - Closely related with the previous problem/issue
- The live stream keeps hanging (it loads a part of it and then hang), maybe an issue with the adaptive resolution selection? (I'm not that familiar with the latest player implementation yet so I don't know if I can fix it in a timely manner)

---

The thing is, we talked about replace the `UniversalImageLoader` as it was deprecated some years ago, I still think we need.

>  Perhaps should shrink these images down to save RAM usage?

We have to eliminate all uses of the `drawable-nodpi` or, at least, keep its usage to an absolute minimum. It was an slight oversight by my part in the case of the placeholders.

wb9688:
@mauriciocolli: I'd recommend using Fresco instead of UIL.

theScrabi:
@karyogamy i thought about removing the paceholder banner compleatly, because it looks ugly. We could just make the view br gone if there is nothing to show

wb9688:
@karyogamy: Fresco should only add about 500 kB (according to [its FAQ](http://frescolib.org/docs/faq.html)). It has a lot more features than Glide and it's faster, mainly because it uses native code, but we'd probably wanna have different build variants for the different architectures, but I don't know how good F-Droid supports that and it'd probably be confusing for our users, so now I think about it again, maybe it'd be better to indeed use Glide. When you implement that, make sure to use our `Downloader` class for downloading the images.

wb9688:
>Yeah, this does look strange. Right now, I can't think of a generic UI that works well for both livestreams with small and large back buffer. Though when we do have a solution, we will probably have way too many different UI to handle under different stream types, so it is probably a good time to do a good refactor on the player.

So how does the official YouTube app do it?

karyogamy:
> have different build variants for the different architectures

This might be a problem, since I believe this was the main reason why we stayed away from including ffmpeg support in the app.

> So how does the official YouTube app do it?

For livestream with long back buffer, they remove the current position when its live and only show the current time when seeking to some time in the past. For short back buffers, they get rid of the progress bar altogether. On the video players, this may look fine. But on the play queue activity, this will probably look strange.

wb9688:
>This might be a problem, since I believe this was the main reason why we stayed away from including ffmpeg support in the app.

That's indeed one of the reasons (besides FFmpeg just being big). That's why I said:

>so now I think about it again, maybe it'd be better to indeed use Glide.

---

>But on the play queue activity, this will probably look strange.

Couldn't we disable the play queue for a live stream? I think a play queue wouldn't make sense at all for a live stream.

avently:
@wb9688 I disagree with disabling play queue. 
- some ui changes should be made on player and on videoplayerfragment to get rid of the play queue
- many code changes should be made
- nothing really stops a user from watching a playlist queue with many live streams. I made needed buttons on popup. User can easily switch a video to the next one.
- think about local playlist
- think about search results with play in popup/playlist functionality

karyogamy:
@wb9688

> That's why I said

Ah, sorry. Late nights =\

> Couldn't we disable the play queue for a live stream?

@avently has some valid points. I don't think users would expect this change in consistency either. That being said, we could just use the youtube UI. And on livestream with short buffer, force the seekbar thumb to stay at the end of the bar and have the progress filled, so the UI appears consistent.

Though, because we can't determine the size of the buffer in the extractor, there will be multiple places where the UI strategy can be determined. Since right now this UI change doesn't impact any player functionalities, it's probably a good idea to do this UI change after we pull the player out and have done a good deal of restructuring, so UI elements can be updated from a single source of change.

theScrabi:
@karyogamy 
> Which placeholder banner? The image when thumbnail loading is turned off? Or the service placeholder image?

That banner you mentioned `background_header.png` sounded to me like it was the banner shown in the Channel fragment. That banner is what I like to remove when there is no banner available.

karyogamy:
@theScrabi Oh, yeah. Please do, or at least shrink it down. The video not available image `not_available_monkey.png` is about the same size as well, so we probably want to replace it with something smaller.

theScrabi:
> Please do, or at least shrink it down. 

Do. I'll put it onto my todo list.

> `not_available_monkey.png` is about the same size as well, so we probably want to replace it with something smaller.

Maybe we can use [ShapeDrawable](https://developer.android.com/guide/topics/resources/drawable-resource.html#Shape) here.

theScrabi:
So from the UI I think the refactoring is nice :)

However I found a small thing that can be improved:
For live streams and audio streams the title bar is rather empty. Mabe we can put a tittle like `LIVE` or `AUDIO` in there.

screenshot_20180322-135342
screenshot_20180322-135433

karyogamy:
> Maybe we can use ShapeDrawable here.

On a related issue, maybe we should also replace our existing icons with vector drawables, since right now they look really bad when enlarged (i.e. the play/restart button on the main video player on a large display).

> For live streams and audio streams the title bar is rather empty. Mabe we can put a tittle like LIVE or AUDIO in there.

If I recall correctly, this was added and then removed in favor of the LIVE label (as duration) on the thumbnail. For Soundcloud, the audio icon on the thumbnail should also indicate that this is audio only. So I think having another label on the toolbar is somewhat redundant. But you are right, that space does feel empty. @mauriciocolli what do you think we should do with the toolbar?

theScrabi:
However you removed [WIP] so I suppose you want to get this PR merged already. So I will review the code now, and merge it soon. If we want to fix more issues we can then send a second PR.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -587,7 +588,8 @@ private void initThumbnailViews(StreamInfo info) {
             imageLoader.displayImage(
                     info.getThumbnailUrl(),
                     thumbnailImageView,
-                    DISPLAY_THUMBNAIL_OPTIONS, new SimpleImageLoadingListener() {
+                    ImageDisplayConstants.DISPLAY_THUMBNAIL_OPTIONS,
+                    new SimpleImageLoadingListener() {
```
theScrabi: I guess this can be converted into a lambda expression.

theScrabi:
I like this commit very much, it adds those tiny awesome details that make you like it :D Also the code looks very clean and strait forward :+1: 

And when I'm already there, I'd like to thank you @karyogamy and @wb9688 you two are currently making my days. SoundCloud support and automatic playlist playback are the most vital tools when I do my daily work on the [owncloud](https://github.com/owncloud/android) app :D Thanks guys!

PR #1197, opened 2018-03-21
Remove IP range

### LINKED ISSUE #1148: [data protection] IP address request to third party service leaks information about users
At the moment, the service https://ipv4.icanhazip.com is used by the crash report activity to (whyever) fetch the public IPv4 of a user, which is then "anonymized" by removing the last two bytes.

Not only this information doesn't really add anything to the crash reports (at least nothing I could think of), it also shares the following details with this service (which, so far, nobody checked, regarding data protection and privacy):
  - complete IP address (and therefore an approximation of their location)
  - user's operating system
  - user is using NewPipe (user agent thing)
  - time of the crash (allows for generating stats like frequency, ...)

In my opinion, this "feature" should be removed entirely from NewPipe, following the data protection principle of "Datensparsamkeit" (a.k.a. "data economy", i.e., reduce the amount of data collected to the absolute minimum).

If there are good reasons to continue to collect this information, however, the implementation needs to be improved to ensure users' privacy (otherwise, one of our USPs, "privacy", doesn't apply):
  - validate service's ToS and ensure they won't collect and/or share the information
  - implement multiple providers (I'd say at least 5) and choose a random one every time (round-robin like)
  - ask user whether they want to share this information, maybe with a checkbox they can uncheck, and request service only when the user confirms they want to send a mail (i.e., don't auto fetch it when the activity is shown)

By the way, according to https://major.io/icanhazip-com-faq/, the software is open source, but the requests are all logged and there is no word about deletion of logs. Also, we don't know anything about where it's hosted and how many "service providers" the provider makes use of have access to the data or network streams.

CC @theScrabi

wb9688:
This would close #1148. I haven't tested it myself yet. @TheAssassin: Could you update the Sentry part?

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/report/ErrorActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/report/ErrorActivity.java
@@ -77,12 +77,10 @@ public class ErrorActivity extends AppCompatActivity {
 
     public static final String ERROR_EMAIL_ADDRESS = "crashreport@newpipe.schabi.org";
     public static final String ERROR_EMAIL_SUBJECT = "Exception in NewPipe " + BuildConfig.VERSION_NAME;
-    Thread globIpRangeThread;
     private String[] errorList;
     private ErrorInfo errorInfo;
     private Class returnActivity;
     private String currentTimeStamp;
-    private String globIpRange;
     // views
     private TextView errorView;
     private EditText userCommentBox;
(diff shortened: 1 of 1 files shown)


TheAssassin:
@wb9688 I'll put it on my TODO list. Might take a couple of days.

theScrabi:
Mucho gracia.

speednroll:
it says "muchas gracias" :joy:

TheAssassin:
I thought the same, @speednroll, but as I don't speak Spanish, I didn't want to comment first.

theScrabi:
@speednroll corrected ... Yea my skills in Spanish are still very basic

Issue #1153, opened 2018-03-01
Liberapay for newpipe?


Bubu:
I very much like newpipe and I'd like to donate something to you.

As I'm not using bitcoin I wanted to ask if you have considered creating a liberapay team for newpipe?

Fdroid will have full support for it in the next client and website release as well.

TobiGr:
Hey @Bubu,
thanks for your interest in donating to NewPipe! As discussed in #1127 we are going to set up librepay soon and then publish all information you need to the readme, our [website](https://newpipe.schabi.org/donate) and to F-Droid. If you want, you can also donate via [Bountysource](https://www.bountysource.com/teams/newpipe).

Bubu:
Thanks for your reply, looking forward to it!

theScrabi:
Spread the word, we now have librepay:
https://liberapay.com/TeamNewPipe/

TobiGr:
I'll add it to the [donation page](https://newpipe.schabi.org/donate/) soon.

theScrabi:
@TobiGr @wb9688 @coffeemakr @mauriciocolli @karyogamy @TheAssassin please feel free to sign your self up to the team :)

theScrabi:
@Bubu please think twice if you wan't to donate to NewPipe. I've just seen you are part of the fdroid team, and we use a small part of the NewPipe donations to support fdroid. So if you donate back you might push money in a circle.

Bubu:
@theScrabi that would be indeed rather awkward, yes 😅

Well in that case I have just added it to the f-droid metadata:
https://gitlab.com/fdroid/fdroiddata/commit/6264f1c2c235f6fbc4c06014c1753291139be68d

And thanks for your work, this is now one of the most used apps on my Phone ;-)

TheAssassin:
@Bubu may we quote you on that on our homepage, please?

Bubu:
> @Bubu may we quote you on that on our homepage, please?

Sure, why not.

B0pol:
As schabi mentionned, the liberapay is here: https://liberapay.com/TeamNewPipe/
And the donate button in app, in about, is redirecting the website, which show liberapay https://newpipe.schabi.org/donate/

hxEmre:
Merhabalar ben Newpipe yükledim güncelleme gösteriyor ama güncelleme yok güncelleme olmadığı için kullanamıyorum ugulamayı yardımcı olur musunuz?

B0pol:
English?

Issue #1127, opened 2018-02-20
a tipp for fdroid


theScrabi:
NewPipe exists for about two and half years now, and It would work without fdroid. The guys over there do a great job, and they supported NewPipe from the beginning on. They run fdroid for free, and run their setup with donated money. Therefore I thought we could give them something back :)

The current account value of the NewPipe project over at Bountysource is at bout 900 USD. I'd like to suggest to donate some of that money to fdroid. However I'd like ask you guys what you think about that, and how much we should donate, or even how:
Monthy selery, or one time payment.

I'd suggest 200$ one time or 15-20$ every month. (This month we made [80$](https://www.bountysource.com/teams/newpipe/issues)).

ale5000-git:
I think it is a good idea and you are a good person.

My opinion about donations is that is better less money but regularly, I suggest 10$ every month.

theScrabi:
Well but small prices always come with a higher ratio of transaction fee. Also I still didn't figure out who to get the money form bountysource to librepay other then by hand.
Normally I'd say the same and pay someone an monthy fee, but I don't know about this now.

@wb9688 @TheAssassin @coffeemakr @mauriciocolli @TobiGr since this is basically project money what do you think about the whole idea?

TheAssassin:
I like the idea itself, but I'd rather go for @ale5000-git's suggestion donating something regularly instead of sending them a one-time donation. Being infrastructure provider, I can confirm it's better to have get regular donations to be able to pay the monthly invoices. Also, if we keep some money, we can mitigate fluctuations.
I would suggest 15 USD per month, that's a good start.

For the rest of the money, well, I plan to attend a couple of conferences this year, and was already a bit disappointed we didn't have any merch for 34C3. I'd like to order some labels, posters, etc.
(Perhaps we can even use some of this money to invite some less wealthy developers to a hack weekend or something.)

theScrabi:
A little merge for 35c3 or other conventions (maybe even street lamps ;D) is in general a cool idea. However lets disuse this in another thread.

TheAssassin:
Of course. I'm just saying, we should save some of the money, and invest it in something useful for advertising.

theScrabi:
So I took a look into Bountysource, I could not find anything where we could pay someone monthly.

theScrabi:
I'm afraid we could only claim it once.

theScrabi:
[...]

PR #1201, opened 2018-03-23
Update README.md

TobiGr:

- Replaced the screenshots which was showing a playlist with a bookmark screenshot. Furthermore we now use compressed screenshots. 
- Linked the liberapay site (see #1153 and #1127 ) in the donation section


TheAssassin:
@TobiGr looks good so far, but how about replacing the QR codes with SVG files? They're probably quite a bit smaller, and being scalable is always a plus.

TobiGr:
@TheAssassin I tried SVGs but these files are larger. The PNG files have sizes of 346, 714 and 815 bytes while the corresponding SVG files have sizes of~37kB.  That's why I'd keep the PNG. I can remove the margin/ white border around these images to make them even smaller.
About the scaling: At the moment I use 100px images (87px qr code and 13px margin) which is good enough for phone cameras. We just use the QR codes in the readme, so why do we have to think about scaling?

TheAssassin:
@TobiGr then those SVG files were pretty bad ones. A QR code is a set of `x^2*2` squares and a few lines. I'll give it a try.

TobiGr:
I used https://www.the-qrcode-generator.com/

TheAssassin:
QtQR produces quite large files, too, but they're < 30 kiB already, so quite a bit smaller. However, I can't reach the PNGs' sizes.

TobiGr:
Me either. I tested plenty only QR code generators now :D
Just removed the margin as mentioned above.

TheAssassin:
> Just removed the margin as mentioned above.

That's a bad idea. The QR code specification requires a margin of at least 4 point widths around the code. See http://www.qrcode.com/en/howto/code.html.

TobiGr:
Yes, you're right.

PR #1217, opened 2018-03-26
Fix broken Librepay link in README.md

Grammost:


TobiGr:
Just tested it and the problem is the slash at the end of the url. When we remove it, there is no 404 error. That's bad handling by Liberapay...
@Grammost So could you please just remove the slash?

Issue #1204, opened 2018-03-24
An issue with release 0.13.0


justanidea:
Hi! Amazing release, thanks everyone :) 
However, i just wanted to show that we may need to adapt the text size relative to the translation for the live word: with the french version for instance, when i start a live i should see "DIRECT" at the bottom right corner, but the word is too long, so I only see "DIRE", so its quite consfusing... :')

theScrabi:
Please send a screenshot related to this.

justanidea:
Yup

justanidea:
Here they are :
http://i66.tinypic.com/xaw2et.jpg

http://i66.tinypic.com/25zpcib.png

http://i65.tinypic.com/av6h06.png

theScrabi:
confirmed

karyogamy:
Yeah, I did this to prevent people from using words that are too long for the LIVE button as some translations were over 10 letters and that just looks bad, especially on the popup window. One solution I can think of right now is to simply make LIVE not translatable, so that part of the UI is uniform across languages, but if there's better solution please suggest it.

theScrabi:
> LIVE not translatable

I'd be fine with that.

justanidea:
You solved this problem :)

PR #1226, opened 2018-03-30
Usability Update Follow-up

karyogamy:
- Added "Add to playlist" menu item to service player play queue.
- [#1204] Fixed LIVE button to use untranslatable string.
- Fixed service player play queue not refreshed when existing play queue is replaced.
- Fixed potential ArrayOutOfBound exception when checking if player window is live.
- Fixed strange pausing behavior when main video player goes into multiwindow mode.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -42,7 +42,7 @@ android {
 
 ext {
     supportLibVersion = '27.1.0'
-    exoPlayerLibVersion = '2.7.1'
+    exoPlayerLibVersion = '2.7.2'
     roomDbLibVersion = '1.0.0'
     leakCanaryLibVersion = '1.5.4'
     okHttpLibVersion = '1.5.0'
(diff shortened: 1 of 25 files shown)


theScrabi:
> Made playback speed dialog translucent.

actually I liked it non translucent better

karyogamy:
Well, the reason the dialog is made translucent is to have the video being played in the background still be visible on the main player. I do realize having translucent dialog is inconsistent with the other dialogs on the service player activity, so it's probably better to have an opaque background on that dialog, while keeping the one on main player still translucent. What do you think?

theScrabi:
> What do you think?

Hm I'd make both opaque. It just feels better. However maybe we need the opinion of a third here.

karyogamy:
Just reverted the translucent dialog, I think it's probably better to add a debated change like this in another PR. As for the playlist cleaning, I'll have to do some more testing to figure out a better timing for it.

theScrabi:
So merge or not?

karyogamy:
If you are fine with the changes, go for it. This PR now doesn't introduce anything new, just some refactoring and bugfixes.

theScrabi:
Yea ... I want to release 0.13.1 because everything is better than 0.11.6

theScrabi:
Ok. I wanted to put a little more stable release on fdroid, but i think you are right.

PR #1230, opened 2018-03-31
make click on metadata open info screen

theScrabi:
@karyogamy  

> I wouldn't do this, since the service player activity is a singleTask and is separate from the main activity, so if you open the service player activity while inside the app, and then open the detail fragment, you won't be able to return to the service activity when back button is pressed. Having an easily accessible button open detail fragment will just cause more inconveniences.

> It would be better if we create another button on the player activity to expand a description view, or just make this activity a fragment, so it can be on the fragment stack.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
@@ -438,9 +438,6 @@ public abstract class ServicePlayerActivity extends AppCompatActivity
         } else if (view.getId() == playbackPitchButton.getId()) {
             openPlaybackParameterDialog();
 
-        } else if (view.getId() == metadata.getId()) {
-            scrollToSelected();
-
         } else if (view.getId() == progressLiveSync.getId()) {
             player.seekToDefault();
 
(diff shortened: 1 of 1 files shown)


theScrabi:
> or just make this activity a fragment, so it can be on the fragment stack.

if that can be done.
For me it was pretty usable, but I can only speak about my experience. 
However this or that way I think it was good to have the detail view popup when the metadata is clicked. It just makes sense. Metadata displayed info about the item, when you click it you get more info.

karyogamy:
For now, may I suggest just add another button to the right of the metadata button to open the details, so it doesn't change the existing behavior too drastically.

app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
```diff
@@ -438,9 +438,6 @@ public void onClick(View view) {
         } else if (view.getId() == playbackPitchButton.getId()) {
             openPlaybackParameterDialog();
 
-        } else if (view.getId() == metadata.getId()) {
-            scrollToSelected();
```
karyogamy: Would be better to change this line to `onOpenDetail(info.getServiceId(), info.getUrl(), info.getName())` rather than using a `setOnClickListener`, just to keep consistency.

Issue #1199, opened 2018-03-22
New drawables and icons


theScrabi:
As @karyogamy pointed out [here](https://github.com/TeamNewPipe/NewPipe/pull/1189#issuecomment-375434974) our current icons/drawables use up lots of memor and are not very scalable since they are bitmaps. On top of that I think they don't look not look very awesome. So as he suggested we should replace them with vector desirables.

FaizKhan5:
Close this.

TobiGr:
I just came across [this article](https://android.jlelse.eu/android-vector-drawables-on-pre-lollipop-crash-solution-45c0c34f0160) on a friend's blog. It could solve our problem.

PR #1232, opened 2018-03-31
[WIP] Fix New drawables and icons #1199

michaelsam94:
Fix New drawables and icons #1199


theScrabi:
497 Files changes. How comes?

michaelsam94:
I replace all png's with vectors

theScrabi:
I see. We already have so many of them?

michaelsam94:
can you merge them?

theScrabi:
I need to see.

michaelsam94:
sure I don't change any name all names of vectors are the same png's names

theScrabi:
Nice :D yea I'll check it.

michaelsam94:
the app will crash on pre lollipop devices I fix many problems by using support library but it still there is some issues

theScrabi:
I see. Well that should not happen. Pleas change the name of the isse to have a [WIP] in front.

michaelsam94:
I change title of this pull request there is any thing else I would change

theScrabi:
Jup remove wip once you are done 👌

michaelsam94:
still there is some issues with drawable top I fix some but the method require android api 17 there is anther solution we can make but it requires every vector to have another file which have layerlist tag if there another solutions you can suggest it Thanks

mauriciocolli:
Notice that are multiple issues to fix yet:
- Different sizes
- Icons with different proportions
- Strange vector path, very different than the png images
- Lose the `ldrtl` resources

I don't know how you generated the vectors, but you should have used Android Studio's new vector asset.

And personally I wouldn't replace all of the images, just the ones that we want to scale like the play button and some others (for example, to the popup resizing and the play button animation). I also don't know how stable this is in older versions of Android (pre-lollipop) so I would use vector only where it needs to be.

**PS**: Please, check _ALL_ the usages before replacing any image, as when we created the layouts, we used the images with the specific size/proportions.

**PS 2**: I also don't remember exactly what bug occurred with the preference support library v7, but that's why I used the v14. I will look at this further later on.

michaelsam94:
so we only replace some png's that is better than replace all png's because still vector not well supported I can revert and we can decide which png will be converted and remove other vectors replace them with png's

michaelsam94:
BTW I use Inkscape to convert pngs to vector

theScrabi:
> And personally I wouldn't replace all of the images, just the ones that we want to scale like the play button 

> BTW I use Inkscape to convert pngs to vector

Inkscape is awesome, but probably misplaced here. As @mauriciocolli said, all the png resources we use are already available as svg, so we don't need to convert them :)

PR #1255, opened 2018-04-03
Add distributionSha256Sum property to gradle-wrapper.properties

Poussinou:

Hi,

This small PR adds the `distributionSha256Sum` property to `gradle-wrapper.properties`. It allows for verification of the downloaded Gradle distribution via SHA-256 hash sum comparison.

You can find some documentation about this security improvement [here](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:verification), and the sums for each version of gradle [here](https://services.gradle.org/distributions/).

Note that you need to update this property every time you change your version of gradle.

(Maybe I should move this PR to master? Should I remove the comment at the end of gradle-wrapper.properties?)

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -4,3 +4,5 @@ distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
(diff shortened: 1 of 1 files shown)


Poussinou:
I updated to 4.6 (By the way, I already saw a lot of gradle-wrapper.properties files from various Android project, your gradle is more up to date than theirs gradle...)

I don't have Android Studio on my hardware right now (I'll reinstall it soon). You can report a bug [here](https://developer.android.com/studio/report-bugs.html). You probably know better than me what to say in the report ;)

Just a question : It doesn't change anything for the NewPipe project to update the gradle right? Could you have compatibility problems?

mauriciocolli:
> Could you have compatibility problems?

I don't think so, we don't do anything fancy with gradle.

Issue #861, opened 2017-11-19
NewPipe has started ignoring 't' or time parameter


tgp1994:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.
NP version: 0.11.0

NewPipe used to respect the 't' parameter in URLs and jump to the appropriate time in videos correctly. At some point recently, NewPipe has stopped doing this. Open this URL for example: https://www.youtube.com/watch?t=64&v=YD3fyiVPNDA and note that NewPipe starts at time 0 rather than at 64 seconds like one would expect.

theScrabi:
Confirmed

wb9688:
Confirmed, NewPipe used NewPipeExtractor's `YoutubeStreamExtractor#getTimeStamp()` method for that, but now I don't see any call to that method in NewPipe. (`grep -r getTimeStamp` returns nothing)

tgp1994:
Was a fix for this bug able to make it in to the 11.1 release just now or is it planned for a later release?

stalks:
Having the same issue. Will this be in the next release?

theScrabi:
No seemed not yet fixed.

moshpirit:
New NP user here. Is it related with NP opening the videos in the background/popup from the beginning?

theScrabi:
yes

tgp1994:
@theScrabi Is this a large bug? I see version 0.11.5 was posted to F-Droid but it looks like this bug hasn't been addressed yet.

mauriciocolli:
Fixed in #1268

tgp1994:
Thank you for keeping us informed. It will be nice to see this fixed!

theScrabi:
@mauriciocolli 
If you merge a fix/feature please add a not for it at the [release page](https://github.com/TeamNewPipe/NewPipe/releases)

PR #1268, opened 2018-04-05
Handle links with time specified (ex. &t=xxx)

arispoloway:

Issue #861 

This fixes the starting time for videos opened in the video player. This wasn't working before because:
1. The url was cleaned immediately upon opening, removing the time info. This isn't necessary because NewPipeExtractor should clean urls on its own anyways
2. PlayQueueItem used a transient Single to store the StreamInfo, and it regenerated the Single with only the url, not all of the information.

I fixed these issues by not cleaning the url initially, and storing StreamInfo with PlayQueueItems. This solution works for links opened with the Background, Video, and Popup players.

I've tested this change with a variety of link types and it works with youtube and soundcloud links

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -54,7 +54,7 @@ dependencies {
         exclude module: 'support-annotations'
     }
 
-    implementation 'com.github.TeamNewPipe:NewPipeExtractor:a6b6235644474'
+    implementation 'com.github.TeamNewPipe:NewPipeExtractor:77a74b8'
 
 
     testImplementation 'junit:junit:4.12'
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -122,7 +122,7 @@ public class RouterActivity extends AppCompatActivity {
                         currentService = NewPipe.getServiceByUrl(url);
                         currentServiceId = currentService.getServiceId();
                         currentLinkType = currentService.getLinkTypeByUrl(url);
-                        currentUrl = NavigationHelper.getCleanUrl(currentService, url, currentLinkType);
+                        currentUrl = url;
                     } else {
                         currentService = NewPipe.getService(currentServiceId);
                     }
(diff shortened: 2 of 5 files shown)


karyogamy:
Thanks for looking into this. 

> NewPipeExtractor should clean urls on its own anyways

Looks like it, I think passing the raw url in should be fine. @theScrabi @mauriciocolli Can you please confirm doing this is alright?

> storing StreamInfo with PlayQueueItems

Please don't do this. `StreamInfo` is a relatively large object and its data have a very short lifespan (SoundCloud stream urls expires in less than 5 minutes). `PlayQueueItem` should only be used as a data class for storing the urls and metadata, and the transient `StreamInfo` there is really just for convenience.

> PlayQueueItem used a transient Single to store the StreamInfo, and it regenerated the Single with only the url, not all of the information.

This is the correct behaviour. The constructor `PlayQueueItem(@NonNull final StreamInfo info)` saving the stream in the item should be considered a bug, since this is no longer subject to the cache expiration specified in `ExtractorHelper`. 

However, all is not lost. Since the `StreamInfo` is cached, there is no point in saving the stream info: `this.stream = Single.just(info);` and we can just remove it. While we are in `PlayQueueItem`, there is also a parameter called `recoveryPosition`, which determines the time the player should seek to when the playback for this item starts. Meaning, if we change the line `this.stream = Single.just(info);` to `if (info.getStartPosition() > 0) setRecoveryPosition(info.getStartPosition() * 1000);`, and remove all other changes on `PlayQueueItem`, the time link behavior would remain corrected.

So, if the team is fine with the clean url change, then I think you've just solved this bug in 2 lines. Good job =D

arispoloway:
@karyogamy 

Thanks for the feedback! Doing it this way makes a lot more sense and is much cleaner.

mauriciocolli:
Thanks for the pull request!

I've added a commit that fix this issue when opening links using "Show Info". I had to add a new field to the Info's object so then we can keep track of what we are opening, this makes the `VideoDetailFragment` stack history easier to maintain.

> Can you please confirm doing this is alright?

With the latest commit, it is.

Issue #767, opened 2017-10-15
Show download screen after giving permissiom


wb9688:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

Currently when I install NewPipe and press download for the first time, it asks for permission, but doesn't open the download screen, instead I've to press the download button again, which is slightly annoying and probably easy to fix.

theScrabi:
Yea could be done :) Wed have to listen to a callback there.

TobiGr:
Fixed

PR #1276, opened 2018-04-06
Open downloads after accepting permissions 

arispoloway:

Issue #767 

Listen to the onRequestPermissionsResult in MainActivity, and if it is the correct permissions and they were all granted, open downloads.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -22,11 +22,13 @@ package org.schabi.newpipe;
 
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
 import android.net.Uri;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 import android.preference.PreferenceManager;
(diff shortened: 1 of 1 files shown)


arispoloway:
Ah yes, I've overlooked this. I think the solution is to use different request codes for each possibility. I will work on this and update the PR later today or tomorrow.

theScrabi:
> request codes for each possibility

Yes :D

mauriciocolli:
I think that instead of messing with the activity, we could have implement the permissions directly into the fragments (there are support methods).

This coupling makes it harder to show this fragment in another activity, and I know that the current fragments are not perfect decoupled, but we should aim for that.

theScrabi:
> fragments are not perfect decoupled, but we should aim for that.

@mauriciocolli yes right.

@arispoloway can you please fix this even though it got already merged.

arispoloway:
Sure, I can look into this. Seems like the API version is going to need to be bumped to M for all of these methods though.

theScrabi:
> M for all of these methods though.

but we already surpassed 22/23. I mean if we need an upgrade sure, but only to latest stable versions.

Issue #849, opened 2017-11-14
[Feature request] Ability to import/export settings


tmbgreaves:
[![bounty](https://api.bountysource.com/badge/issue?issue_id=51556592)](https://www.bountysource.com/issues/51556592-feature-request-ability-to-import-export-settings)

Making a specific feature request for use on Bountysource[0], branching from a comment on #729 regarding ability to import/export settings. Original request from @Gitaroony was:

> Would it be possible to add an export/import feature for all settings, favs, history and watched?

+1 - I would very much like to see this added too. Thanks!

[0] Alas, it doesn't allow fine-grained bounties added to comments in issues, at least as far as I can see.

theScrabi:
All right :)

tmbgreaves:
Big thanks for all your work on this - works great for me on Lineage 14.1 (suruzan) and is a significant upgrade on my previous termux+youtube-dl solution :-) Thank you!

theScrabi:
Thank you :)

Somethingweirdhere:
I would give this a try if no one else is currently working on it :)

Somethingweirdhere:
Alright, done. It took longer than expected because another function I was trying to use was broken, but now it all seems to work! 

@theScrabi How long does a dev branch need to get merged into master? I have tested it as well as I could, but I really don't want to screw up peoples exports/backups if I have done something wrong.

theScrabi:
Currently its about a week or two. However did you test importing a zip that has no settings but only the database embodied?

Somethingweirdhere:
Yeah, i added an if statement that checks if the settings file can be extracted from the zip.

theScrabi:
ok

theScrabi:
This got implemented by @Somethingweirdhere. :D You may pick up your reward.

Somethingweirdhere:
@theScrabi Thanks, but how? It says the issue needs to be closed first.

TobiGr:
If I recall correctly, you have to be member of the GitHub team to claim the bounty.

theScrabi:
No, @Somethingweirdhere you just need to sign up at boutysource.

Somethingweirdhere:
@theScrabi Alright, I am, but I still need to wait for it to be accepted by the backer. I will reinvest it in another feature when I get it :)

theScrabi:
@tmbgreaves please remember to do that soon.

tmbgreaves:
[...]

PR #1333, opened 2018-04-18
Settings export

Somethingweirdhere:

Alright, this actually seems to work now! I haven't encountered any issues with this yet, but I will continue testing it because exports feel quite important. I hope someone else can also confirm there are no issues.

Anyways, I had to change how ZipHelper.extractFileFromZip works because it was broken and always returned true and also didn't work for the same ZipInputStream more than once. It now uses the filepath, but as the function isn't used outside of ContentSettingsFragment.java this won't be an issue. 

The exports now contain a new file called newpipe.settings. If while importing it is detected, it will trigger a second dialog asking if to import settings too. If you press Ok it will import the settings from there.

#849

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
@@ -197,7 +197,7 @@ public class ContentSettingsFragment extends BasePreferenceFragment {
                             new FileOutputStream(path)));
             ZipHelper.addFileToZip(outZip, newpipe_db.getPath(), "newpipe.db");
             ZipHelper.addFileToZip(outZip, newpipe_db_journal.getPath(), "newpipe.db-journal");
-
+            //add settings
             outZip.close();
 
             Toast.makeText(getContext(), R.string.export_complete_toast, Toast.LENGTH_SHORT)
--- a/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
@@ -4,7 +4,9 @@ import android.app.Activity;
 import android.app.AlertDialog;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v7.preference.ListPreference;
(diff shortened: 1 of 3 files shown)


Somethingweirdhere:
@theScrabi Is everything alright with this or should I change something with the way it works?

theScrabi:
@Somethingweird I am sorry I didn't find time to check everything yet.
I hope I can do a review at the weekend.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -207,6 +220,29 @@ private void exportDatabase(String path) {
         }
     }
 
+    private void saveSharedPreferencesToFile(File dst) {
```
theScrabi: Why can't we save `/shared_prefs/org.schabi.newpipe.debug_preferences.xml` directly? Is there any advantage generating by generically?

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -223,30 +259,89 @@ private void importDatabase(String filePath) {
...
 
-            // restart app to properly load db
-            //App.restart(getContext());
-            System.exit(0);
         } catch (Exception e) {
             onError(e);
         }
     }
 
+    private void loadSharedPreferences(File src) {
```
theScrabi: Same here.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -223,30 +259,89 @@ private void importDatabase(String filePath) {
...
 
-            zipIn.close();
+            //If settings file exist, ask if it should be imported.
+            if(ZipHelper.extractFileFromZip(filePath, newpipe_settings.getPath(), "newpipe.settings")) {
+                AlertDialog.Builder alert = new AlertDialog.Builder(getContext());
+                alert.setTitle(R.string.import_settings);
+
+                alert.setNegativeButton(android.R.string.no, new DialogInterface.OnClickListener() {
+                    @Override
+                    public void onClick(DialogInterface dialog, int which) {
```
theScrabi: Please use lambda convention here.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -223,30 +259,89 @@ private void importDatabase(String filePath) {
...
+                        dialog.dismiss();
+                        // restart app to properly load db
+                        System.exit(0);
+                    }
+                });
+                alert.setPositiveButton(android.R.string.yes, new DialogInterface.OnClickListener() {
+                    @Override
+                    public void onClick(DialogInterface dialog, int which) {
```
theScrabi: Please use lambda convention here.

Somethingweirdhere:
@theScrabi 

I changed to lambda convention. I am not sure why travis failed, I hope it's nothing serious.

Now about the saving:

I don't see any disadvantages to the way I did it, are there any except that there is more code? I did it this way because of some very minor advantages, but I tested it and it also works like you suggested.

The advantages are:

1. Exporting the preferences one by one and reimporting means that even if the way how preferences work changes (like the xml header, the file where it is stored, or if there is more information in the file than now exists), the exporting/importing will continue working. I don't see why this would change, but it's a possibility I guess.

2. [https://stackoverflow.com/questions/6146106/where-are-shared-preferences-stored/](Here) it is mentioned in the comments of the answers that some devices store their DefaultPreferences in differently names files. I haven't been able to find a device that does this, but I only have a S8+ and the provided emulators, but those only work up from Lollipop for me. I also haven't been able to find information about this topic anywhere else. It seems unlikely to me that there is a device that does not store the preferences in org.schabi.newpipe.debug_preferences.xml, but I didn't want to risk it.

It is up to you which way you prefer, changing the code won't take me long.

theScrabi:
muy bien.

theScrabi:
Don't forget to claim your reward at bountysource:
https://www.bountysource.com/issues/51556592-feature-request-ability-to-import-export-settings

PR #1339, opened 2018-04-21
Added the ability to limit video quality if using mobile data.

### LINKED ISSUE #1259: [Feature request]Setting to limit mobile data usage

When I use NewPipe outside of my home I always switch to 144p to avoid eating up my data plan. This is annoying and I often forget. Could you add an option to limit bandwidth usage when off of wifi?
Instead of adding four additional non-wifi “Default XXXX” options you could add just one:

Settings option:
**Limit mobile data usage**

Choices:

| Label | Action |
| --- | --- |
| No change | Stick with user chosen defaults |
| Play video and audio at a reduced quality | Reduce resolution and format for the lowest bandwidth consumption |
| Play only audio at a reduced quality | Show video thumbnail and reduce audio rate |

What it might look like:
newpipe-mobilelimit

Zenopheus:
Fixes #1259 
This limits the user's default video resolution when not connected to WiFi. The user can choose "No Limit"(default) or a specified resolution ("144p" for example). When active the audio stream with the lowest bitrate is also chosen. This differs from the proposed design due to complexity concerns. For example, supporting a "No Video" option would have required a lot of changes to be implemented cleanly.

- Changes to Settings video & audio settings
  - Added mobile data resolution limit drop down
- Changes to ListHelper:
  - Automatically limits the *default* resolution based on user's resolution limit settings. Direct changes to video resolution (from within the player) will not be limited.
  - Automatically limits the *default* audio bitrate to the lowest available stream if limiting is enabled.
  - Removed some dead code and did some cleanup
  - Make methods private/protected to help understand what was in active use
  - The code now chooses one format over an other using a simple ranking system defined in array constants. I realized I needed to do this in order to choose the most efficient video stream. I did my best to evaluate the video and audio formats based on quality and efficiency. It's not an exact science.
  - Made changes to the tests to support my changes

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
@@ -71,15 +71,11 @@ public final class ListHelper {
 
         // If the user has chosen to limit resolution to conserve mobile data
         // usage then we should also limit our audio usage.
-        int result;
         if (isLimitingDataUsage(context)) {
-            result = getMostCompactAudioIndex(defaultFormat, audioStreams);
+            return getMostCompactAudioIndex(defaultFormat, audioStreams);
+        } else {
+            return getHighestQualityAudioIndex(defaultFormat, audioStreams);
         }
-        else {
-            result = getHighestQualityAudioIndex(defaultFormat, audioStreams);
-        }
-
-        return result;
     }
 
     /**
@@ -109,8 +105,9 @@ public final class ListHelper {
         SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
 
         // Load the prefered resolution otherwise the best available
-        String resolution = preferences != null ? preferences.getString(context.getString(key),
-                context.getString(value)) : context.getString(R.string.best_resolution_key);
+        String resolution = preferences != null
+                ? preferences.getString(context.getString(key), context.getString(value))
+                : context.getString(R.string.best_resolution_key);
 
         String maxResolution = getResolutionLimit(context);
         if (maxResolution != null && compareVideoStreamResolution(maxResolution, resolution) < 1){
--- a/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
@@ -2,6 +2,7 @@ package org.schabi.newpipe.util;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.net.ConnectivityManager;
 import android.preference.PreferenceManager;
 import android.support.annotation.StringRes;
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -13,126 +14,72 @@
...
-        if (audioStreams.size() > 1) for (int i = 1; i < audioStreams.size(); i++) {
-            AudioStream audioStream = audioStreams.get(i);
-            if (audioStream.getAverageBitrate() >= audioStreams.get(highestQualityIndex).getAverageBitrate()) highestQualityIndex = i;
+        MediaFormat defaultFormat = getDefaultFormat(context, R.string.default_audio_format_key,
+                R.string.default_audio_format_value);
+
+        // If the user has chosen to limit resolution to conserve mobile data
+        // usage then we should also limit our audio usage.
+        int result;
+        if (isLimitingDataUsage(context)) {
```
theScrabi: why not
```
if (isLimitingDataUsage(context)) {
   return getMostCompactAudioIndex(defaultFormat, audioStreams);
} else {
   return getHighestQualityAudioIndex(defaultFormat, audioStreams);
}

File: app/src/main/java/org/schabi/newpipe/util/ListHelper.java [hunk 2]
```diff
@@ -154,6 +101,49 @@ public static int getHighestQualityAudioIndex(MediaFormat defaultFormat, List<Au
...
 
+    /*//////////////////////////////////////////////////////////////////////////
+    // Utils
+    //////////////////////////////////////////////////////////////////////////*/
+
+    private static String computeDefaultResolution(Context context, int key, int value) {
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
+
+        // Load the prefered resolution otherwise the best available
+        String resolution = preferences != null ? preferences.getString(context.getString(key),
```
theScrabi: Please format this line like this:
```
String resolution = preferences != null
                             ? preferences.getString(context.getString(key), context.getString(value))
                             : context.getString(R.string.best_resolution_key);

```

app/src/main/res/values/settings_keys.xml
```diff
@@ -851,4 +851,32 @@
         <item>ZM</item>
         <item>ZW</item>
     </string-array>
+
+    <!-- Limit mobile data usage  -->
+    <string name="limit_mobile_data_usage_key" translatable="false">limit_mobile_data_usage</string>
+    <string name="limit_mobile_data_usage_value" translatable="false">@string/limit_data_usage_none_key</string>
+    <string-array name="limit_data_usage_description_list">
```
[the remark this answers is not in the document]
Zenopheus: I thought strings.xml was for stuff that required translation? None of these contain user facing labels. I was mimicking default_resolution_key, default_resolution_value, and resolution_list_values as listed above. Do you want me to move limit_data_usage_description_list into strings.xml? Should I include resolution_list_values also then?

app/src/main/res/values/settings_keys.xml
```diff
@@ -851,4 +851,32 @@
...
+        <item>1080p60</item>
+        <item>1080p</item>
+        <item>720p60</item>
+        <item>720p</item>
+        <item>480p</item>
+        <item>360p</item>
+        <item>240p</item>
+        <item>144p</item>
+    </string-array>
+    <string-array name="limit_data_usage_values_list">
```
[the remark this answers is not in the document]
Zenopheus: Almost, resolution_list_values is mostly identical except for the first item: "Best resolution" vs "No limit". I was also bothered by the duplication but I didn't know of a way to share *some* of the items without more refactoring and I was trying to avoid too many code changes for my first push. We may also want to consider a "No video" item at some point.

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
Zenopheus: I think because my original changes were more complicated (multiple limiting levels). Some people prefer one return at the end of a method so I kept it. It doesn't matter to me either way, I'll throw in the returns.

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
Zenopheus: sure

app/src/main/res/values/settings_keys.xml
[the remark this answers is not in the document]
theScrabi: You can mark them as nontranslatable than they will not get translated.
I want the settings_keys.xml just contain keys.

app/src/main/res/values/settings_keys.xml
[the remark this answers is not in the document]
theScrabi: hm ok.

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
theScrabi: I am an R-Value/functional fan, so I think return would fit well here :)

theScrabi:
If ur done please remove the wip sign

theScrabi:
Question, how does this affect the difference between regular video on popup? Will it affect both when using mobile data?

theScrabi:
all right nice contribution thank you :)

Issue #1307, opened 2018-04-12
Remove The Auto Video Pause When Pressing On The Popup Window


GKid94:
Can you please remove the auto video pause when pressing on the popup window instead make it just to show the player controls when pressing on the popup window?

Somethingweirdhere:
I have tried this but failed. It was easy to prevent the pausing, but I haven't managed to get the controls there.

Stypox:
Fixed in #1342

PR #1342, opened 2018-04-21
PopUpPlayer now also has a Play and Pause button

Somethingweirdhere:

#1307 

Right now, pressing on the popupplayer paused the video and showed a play button. This was very annoying when you just tried to look at the time or something. Now it works more like the normal youtube app, tapping brings up the progress bar, some settings and a pause button. When you press it it pauses the video and becomes a play button. Also, thanks to this you can look at a paused video without all the clutter around.

Additionally, this didn't break the MainPlayer, even though VideoPlayer is used in the PopUpPlayer and the MainPlayer. 

The little issue is that the play/pause/repeat button gets slightly off scale, but only about 10%. I tried fixing this, but I really can't figure out how rescaling works. Maybe someone can help. But in my opinion the fluidity this brings is worth the slightly distorted button.

### DIFF
--- a/app/src/main/res/layout/player_popup.xml
+++ b/app/src/main/res/layout/player_popup.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <FrameLayout
     xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
@@ -218,21 +217,6 @@
         android:weightSum="5.5">
         <!--tools:visibility="gone">-->
 
-        <ImageButton
-            android:id="@+id/videoPlayPause"
-            android:layout_width="30dp"
-            android:layout_height="60dp"
-            android:layout_alignParentRight="true"
-            android:layout_weight="1"
-            android:background="@drawable/ic_pause_white"
-            android:clickable="true"
-            android:focusable="true"
-            android:scaleType="centerInside"
-            android:scaleX="1"
-            android:scaleY="1"
-            android:visibility="invisible"
-            tools:ignore="ContentDescription,RtlHardcoded" />
-
         <ImageView
             android:id="@+id/controlAnimationView"
             android:layout_width="0dp"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/player_popup.xml
```diff
@@ -227,6 +228,26 @@
...
+        android:orientation="horizontal"
+        android:weightSum="5.5">
+        <!--tools:visibility="gone">-->
+
+        <ImageView
+            android:id="@+id/videoPlayPause"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:scaleType="centerInside"
```
theScrabi: Please use something that does not alter the aspect ratio of the play button.

Somethingweirdhere:
@theScrabi I really don't know, it's what I meant with "it goes off scale".

`        @Override
        public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float distanceY) {
            if (isResizing || playerImpl == null) return super.onScroll(e1, e2, distanceX, distanceY);

            if (playerImpl.getCurrentState() != BasePlayer.STATE_BUFFERING
                    && (!isMoving || playerImpl.getControlsRoot().getAlpha() != 1f)) playerImpl.showControls(0);
            isMoving = true;

            float diffX = (int) (e2.getRawX() - e1.getRawX()), posX = (int) (initialPopupX + diffX);
            float diffY = (int) (e2.getRawY() - e1.getRawY()), posY = (int) (initialPopupY + diffY);

            if (posX > (screenWidth - popupWidth)) posX = (int) (screenWidth - popupWidth);
            else if (posX < 0) posX = 0;

            if (posY > (screenHeight - popupHeight)) posY = (int) (screenHeight - popupHeight);
            else if (posY < 0) posY = 0;

            windowLayoutParams.x = (int) posX;
            windowLayoutParams.y = (int) posY;

            //noinspection PointlessBooleanExpression
            if (DEBUG && false) Log.d(TAG, "PopupVideoPlayer.onScroll = " +
                    ", e1.getRaw = [" + e1.getRawX() + ", " + e1.getRawY() + "]" +
                    ", e2.getRaw = [" + e2.getRawX() + ", " + e2.getRawY() + "]" +
                    ", distanceXy = [" + distanceX + ", " + distanceY + "]" +
                    ", posXy = [" + posX + ", " + posY + "]" +
                    ", popupWh = [" + popupWidth + " x " + popupHeight + "]");
            windowManager.updateViewLayout(playerImpl.getRootView(), windowLayoutParams);
            return true;
        }`

This is the code responsible for rescaling, but it seems like it only changes the size of the window. No idea what happens to the contents and why they change size :/

Somethingweirdhere:
Right, the issue is now fixed. 

image

image

PR #1350, opened 2018-04-23
Spelling: Rework new strings

comradekingu:


app/src/main/res/values/strings.xml
```diff
@@ -307,20 +307,20 @@
...
     <string name="give_back">Give back</string>
     <string name="website_title">Website</string>
-    <string name="website_encouragement">To get more information and the latest news about NewPipe visit our website.</string>
+    <string name="website_encouragement">Visit the NewPipe Website for more info and news.</string>
     <string name="website_url" translatable="false">https://newpipe.schabi.org/</string>
     <string name="app_license_title">NewPipe\'s License</string>
-    <string name="app_license" translatable="false">NewPipe is Free Software: You can use, study share and improve it at your will. Specifically you can redistribute and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.</string>
+    <string name="app_license">NewPipe is copyleft libre software: You can use, study share and improve it at your will. Specifically you can redistribute and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.</string>
```
theScrabi: NewPipe is copyleft libre software

Why copyleft? We have a copyright. We just license it unter GPLv3?

theScrabi:
Nice, I like the new strings :)

app/src/main/res/values/strings.xml
theScrabi: I would rather ride free as in freedom.

app/src/main/res/values/strings.xml
```diff
@@ -104,7 +104,7 @@
     <string name="settings_category_player_title">Player</string>
     <string name="settings_category_player_behavior_title">Behavior</string>
     <string name="settings_category_video_audio_title">Video &amp; Audio</string>
-    <string name="settings_category_history_title">History &amp; Cache</string>
+    <string name="settings_category_history_title">History</string>
```
TobiGr: I think that was called "History & Cache" because there is an option to clean stored metadata. When we rename this category, we should also move this option. Otherwise it's hard to find.

app/src/main/res/values/strings.xml
```diff
@@ -307,20 +307,20 @@
     <string name="tab_about">About</string>
     <string name="tab_contributors">Contributors</string>
     <string name="tab_licenses">Licenses</string>
-    <string name="app_description">A free lightweight YouTube frontend for Android.</string>
+    <string name="app_description">A libre lightweight YouTube frontend for Android.</string>
```
TobiGr: We should generally think about replacing YouTube with streaming. SoundCloud is in beta yet, so this suffices for now.
Furthermore we should replace "fronted". NewPipe is more than this.
E.g. I use `NewPipe - the smart streaming solution` and `NewPipe the lightweight YouTube experience for Android` on https://newpipe.schabi.org

app/src/main/res/values/strings.xml
theScrabi: I agree. I wouldn't move it I would keep the string as it is.

app/src/main/res/values/strings.xml
comradekingu: "Free" makes for errors in translation, when people interpret it to mean "gratis software". It will be "free" in every other language, as only English that i know of has this ambiguity.
GPLv3+ is strong copyleft. https://en.wikipedia.org/wiki/Copyleft#Strong_and_weak_copyleft

app/src/main/res/values/strings.xml
comradekingu: "NewPipe, lightweight streaming enjoyment on Android" ?
You cant actually stream to others (yet).

app/src/main/res/values/strings.xml
comradekingu: I reverted it

Issue #1334, opened 2018-04-19
[Feature Request} minimize on exit


droidberg:
currently when you click on a video it expands to a window that then allows you to play the video. when you press on the back button it closes the video and reverts back to the window. any chance you can make it so that the video minimizes to a floating window when you press back? this is one of the things i like about youtube and would love to see it in newpipe.

Phoenix616:
If such a feature is implemented it should definitely be an option. I personally extremely dislike this behaviour about the official app, I expect the back button to go back and actually stop everything that's currently going on.

theScrabi:
I agree with @Phoenix616. Also we already have a button to switch between popup and fullscreen mode.

TimboJones:
Having it as an option would be awesome imo. It's just quality of life really, but more often than not I was just going back to the Homescreen and expected the video to go into pop-up mode automatically. I was thinking of implementing something like that before, maybe I'll get around to it

theScrabi:
@TimboJones I'll merge it :)

droidberg:
obviously, this would be an optional thing but it comes in handy when viewing a video and switching to another task. i also very often look for other videos to watch before the current one ends.

trymeouteh:
I think this should be an option in the settings. You can set the back button to either exit the video (By Default), popup window, or background.

PR #1354, opened 2018-04-27
Minimize main player on exit

karyogamy:
Here is a quick implementation for request #1334, added as toggleable in Settings -> Video & Audio.

However, I'm not particularly fond of this implementation as it performs the switch based on the `onStop` lifecycle event, which will occur on:
- back button pressed -- expected behavior
- home button pressed -- expected behavior
- screen turned off -- expected behavior
- other activity starts (open up settings) -- expected behavior
- navigation/overview button pressed -- maybe unexpected behavior
- entering multiwindow mode (long press overview button) -- unexpected behavior
- (many more possibilities, but mostly when the player activity is actively killed by the user)

Since this implementation uses the existing player switching method, the switching is just going to be as slow performance-wise. If anyone has any idea how this can be improved, please don't hesitate to make suggestions. That being said, I'd like to keep the expected behaviors intact.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
@@ -104,6 +104,7 @@ public final class MainVideoPlayer extends AppCompatActivity
 
     @Nullable private PlayerState playerState;
     private boolean isInMultiWindow;
+    private boolean isBackPressed;
 
     /*//////////////////////////////////////////////////////////////////////////
     // Activity LifeCycle
@@ -192,6 +193,12 @@ public final class MainVideoPlayer extends AppCompatActivity
     }
 
     @Override
+    public void onBackPressed() {
+        super.onBackPressed();
+        isBackPressed = true;
+    }
+
+    @Override
     protected void onSaveInstanceState(Bundle outState) {
         if (DEBUG) Log.d(TAG, "onSaveInstanceState() called");
         super.onSaveInstanceState(outState);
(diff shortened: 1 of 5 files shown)


theScrabi:
what is the status of this. I see that you have two pending PR atm.

justanidea:
Would it be possible to add an unexpected behavior when you double press the home button?

karyogamy:
@theScrabi I was waiting for some advise on a better implementation, but since no one is doing that, I suppose we can go with the current solution. The changes can be easily rolled back anyways. 

I've also decided to follow suit with the VLC app design and have a setting dialog to allow users to specify what action to take on exit: none, resume on background or resume on popup. Will be testing it a bit before removing the WIP prefix.

@justanidea It's possible, but what do you want to do with this action? It generally is a bad idea to bind behavior to some actions no user is expecting.

theScrabi:
Ok I will test it for a while.

theScrabi:
The back pressed didn't work for me tho.

karyogamy:
Sorry, I forgot to update the description. I changed the back button behavior so we still have a direct way of closing the main player, the same way as the vlc app.

TobiGr:
There is also a problem with the pop-up player when the screen overlay permission has not been granted yet. The sound plays, but there is no video.

When you stop a video and the screen turns of, the video starts playing. I guess this is not the intended behaviour.

karyogamy:
Oh crap, nice find. I'll create a fix soon, @theScrabi please don't bump and release the next version yet.

theScrabi:
> Oh crap, nice find. I'll create a fix soon, @theScrabi please don't bump and release the next version yet.

All right.

PR #1355, opened 2018-04-28
Remove GSON lib

acrosca:


theScrabi:
Could you please giva a description, for example why to remove the gson lib.

acrosca:
Removing GSON reduces method count (1345), apk size (188Kb).

The project does not need a large and powerful JSON parser to serialize/deserialize a single class.
Having fewer dependencies is a good thing.

wb9688:
Yeah, and it's also weird that we use another JSON parser in NewPipeExtractor.

theScrabi:
All right.

PR #1356, opened 2018-04-29
Double-tap the middle of the player screen to pause the video

DafabHoid:

This PR changes how double-tap for the main player works:
Now, double-tapping the middle third of the screen will pause the video, while rewind / fast forward are on the left third / right third of the screen, respectively.
The reason is, that currently there is no practical way to pause the video fast: When you want to pause playback, you need to tap the screen, wait for the controls to fade in, and then finally press pause. But if you tap the center (where the pause button fades in) too fast, it will be recognized as a double-tap gesture, which will trigger a **seek**, although you tapped the center of the screen and actually wanted to pause the video. This is probably not what you expect, so this PR attempts to change this behaviour to something more natural.
Another idea, which I did not yet implement, was to allow a double-tap while being paused to resume playback.

Feel free to comment wether this change reflects NewPipe's user experience principles. I am open to any suggestions.


theScrabi:
It's cool :)

Issue #1388, opened 2018-05-09
Cannot play WebM 2160p60 on any device on any network


24RJ6:
- [X] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [X] I checked if the issue/feature exists in the latest version.

I tried this on a couple of top end devices: Note 8, S8, S9+. The first two also both with Nougat and Oreo firmware. I checked the option to always use the highest available resolution and for some reason always end up with WebM 2160p60 even when searching for 8k videos on YouTube. It doesn't seem like an issue with decoding speed, it always plays for like 1-2 seconds and then I get the loading/buffering circle for 1-2 seconds. I tried multiple networks and get about 100-200 Mbit downstream via the phones Wifi. It seems it can't get the data fast enough despite the bandwidth being available on the network level.

The videos play fine when downloaded in MX Player or VLC. These phones also play the [400 Mbit jellyfish stress file](http://jell.yfish.us/media/jellyfish-400-mbps-4k-uhd-hevc-10bit.mkv) smooth in MX Player with HW+ decoder. So there should be more than enough decoding power.

theScrabi:
@karyogamy we have to do something about the buffering. It's not the first time something like this happens. You said you changed some parameters there recently. Could we roll them back for the next release?

karyogamy:
@theScrabi 
Yes, let's do that. I'll create a fix PR soon.

@24RJ6 
> The videos play fine when downloaded in MX Player or VLC.

Did you download the files before playing them in external players? If so, can you please check if the video streams better on-demand (by enabling "Use external video player" option in the "Video & Audio" settings)?

If possible, can you please also use a network monitoring tool, such as [this](https://play.google.com/store/apps/details?id=com.internet.speed.meter.lite&hl=en), to see if the extracted stream url is rate-limited during on-demand?

theScrabi:
@24RJ6 I've created a test build with the changes made in https://github.com/TeamNewPipe/NewPipe/pull/1389. Please tell me if it works:
https://cloud.damken.com/index.php/s/wSyRd2yC15x2ZwT
passwort is: buffertest

24RJ6:
OK, I think I figured why MX is choppy and VLC has issues, they play the 8k version and not the 2160. Let me see if I can get them to play the 2160 one as NewPipe claims to play.

24RJ6:
[...]

Issue #1294, opened 2018-04-09
Slow buffering


azat11:
I have a 50mbps Wi-Fi and videos do still buffer very slowly, especially when it comes to 60fps. So much so that I can not even watch the video for a minute or two without the loading circle appearing.
And with the 0.13.1 version it's now the loading circle for a couple of seconds every couple of seconds of the video.
Is it just me, or is this just how NewPipe is?

azat11:
> However for me it helped to just close the video and restart it.
Seems to help for now. Although it buffers so slowly I can't even see it on the progress bar.

> lucky mann, my connection barely has 10mbps.
Well, it's Russia man. $9 for a month at 50mbps. Double at double the speed.

theScrabi:
xD you've never been to germany. Everyone says we have a shitty net... its because we have a shitty net. You are lucky if you get 50mbps there but you pay up to 40$ for it.

karyogamy:
[This](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/helper/LoadController.java) is our current buffering strategy. In short, once we enter buffering state (at the start or not enough data while playing), and the player have loaded `initialPlaybackBufferMs`, the playback will start immediately. When playing, the buffer will zigzag between the `minimumPlaybackbufferMs` and `optimalPlaybackBufferMs` watermarks. 

> And with the 0.13.1 version it's now the loading circle for a couple of seconds every couple of seconds of the video.

I imagine the constant buffering on 13.1 is because the `initialPlaybackBufferMs` might be too low for unstable network, though it really makes starting up a video faster under stable situations. So, it might be a good idea to create a setting in the player that allows the user to change the `initialPlaybackBufferMs` parameter, to let the player buffer for a little more before start playing.

azat11:
> However for me it helped to just close the video and restart it.

Update. This only helps for a minute or so. After that it's back to couple seconds of loading and playing for 720p60, and a second/half a second of loading and playing for 720p.
With the new version I now can not watch videos at all, though with these speeds it shouldn't be a problem.
BTW, tried SkyTube, and it works fine for the 720p (no 60fps option there). So it's likely not not the connection that is the culprit.

theScrabi:
> BTW, tried SkyTube, and it works fine for the 720p (no 60fps option there). So it's likely not not the connection that is the culprit.

[...]

PR #1389, opened 2018-05-10
Revert Buffering Strategy

karyogamy:
This PR is in response to #1388 and #1294, to revert the buffering strategy and parameter changes to load controller. 

Though after testing on my own device, this only prolongs the slow buffering problem from playing and buffering every second to playing and buffering every several seconds, which, IMO delivers worse user experience.


theScrabi:
I don't like it either. Maybe we shold supply a test build with this parameters to the creators of the issue before we roll it out to the public.

theScrabi:
I don't like it either. Maybe we shold supply a test build with this parameters to the creators of the issue before we roll it out to the public.

karyogamy:
Good idea, thanks =D

Issue #1383, opened 2018-05-08
subtitles option doesn't appear anymore in popup mode


awefartib:
- [x ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x ] I checked if the issue/feature exists in the latest version.

On NewPipe 0.13.3, I can't enable subtitles in popup mode anymore. It used to be possible when increasing the size of the popup, but that doesn't seem to work anymore. The subtitle option never shows up. I'm not sure which NewPipe update caused this setting to disappear from popup mode.  

Edit: This feature got broken on the latest update (0.13.3). Works flawlessly on 0.13.2.

PR #1392, opened 2018-05-12
ExoPlayer 2.8.2 Update

karyogamy:
This update bumps the player library version and:

- Reworked the playback speed control dialog to support different step sizes for faster speed change.
- Added a toggle to fast-forward during silences in playback speed control. This should be helpful for audiobooks and certain music genres, and can bring a true seamless experience (and can break a song with lots of silences =\\).  
- Refactored media source resolution to allow passing metadata alongside media internally in the player, rather than doing so manually. Now we have a single source of metadata and is directly available when playback starts.
- Fixed remote playlist metadata not updating when new metadata is available when playlist fragment is opened.
- Various UI fixes: #1383, background player notification controls now always white, easier to shutdown popup player through flinging

However, there is a potential bug in the current release that breaks livestreams, so we will need to wait until a bugfix or resolution is found before merging.

I've also tried to integrate our players to use the new notification UI, which can reduce all notification code on the background and popup players to about 20 lines and provide dynamic notification coloring from thumbnail on Oreo. Unfortunately, the library hardcoded the broadcast intent strings, which means we can't have two players at the same time. Maybe we can later refactor the notification controls and roll our own solution.

If anyone is interested, [here](https://www.youtube.com/watch?v=svdq1BWl4r8) is the link to the new ExoPlayer Google IO talk.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -42,7 +42,7 @@ android {
 
 ext {
     supportLibVersion = '27.1.1'
-    exoPlayerLibVersion = '2.8.1'
+    exoPlayerLibVersion = '2.8.2'
     roomDbLibVersion = '1.1.1'
     leakCanaryLibVersion = '1.5.4'
     okHttpLibVersion = '3.10.0'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -42,7 +42,7 @@ android {
 
 ext {
     supportLibVersion = '27.1.1'
-    exoPlayerLibVersion = '2.7.3'
+    exoPlayerLibVersion = '2.8.0'
     roomDbLibVersion = '1.1.1'
     leakCanaryLibVersion = '1.5.4'
     okHttpLibVersion = '3.10.0'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -42,7 +42,7 @@ android {
 
 ext {
     supportLibVersion = '27.1.1'
-    exoPlayerLibVersion = '2.8.0'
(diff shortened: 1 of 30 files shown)


Somethingweirdhere:
@karyogamy About the silences: Would there be a way to make an option to only remove them at the beginning and end of tracks?

karyogamy:
It's possible but not with the current ExoPlayer architecture, since the [built-in silence skipper](https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/audio/SilenceSkippingAudioProcessor.java) uses the [AudioProcessor](https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/audio/AudioProcessor.java) interface, which doesn't pass down the track information. 
Later down the road we can rework the base player a bit to expose an interface for adding extra audio processors to allow custom audio plugins, including a track-sensitive silence skipper. But since this is an update PR, I'd prefer not introduce too much complexity.

theScrabi:
Should I still block the release of the next version, or did you fix that issue.

theScrabi:
@karyogamy I grated you advanced writing permissions on the repo, you can now eddid the release note: https://github.com/TeamNewPipe/NewPipe/releases/tag/untagged-30c1321b12d6c8d57604
Be careful with those permissions ;)

> Also we might want to fix the number format exception issue when resolving resolution on mobile data before the next release.

Ok, sounds good :)

Issue #1390, opened 2018-05-11
Fail to open long link


daufinsyd:
- [ ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I checked if the issue/feature exists in the latest version.

If a link is tool long, the hidden part is omitted on click.

For example the second link opens http://www.morgandavidking.com/free-d
Instead of https://morgandavidking.com/free-downloads/
![screenshot_20180511-160745](https://user-images.githubusercontent.com/1218113/39928585-0fd6898e-5536-11e8-994e-a6d02c525873.png)

Somethingweirdhere:
The question is should we make the link longer or just fix the hyperlink?

Somethingweirdhere:
<a href="/redirect?q=http%3A%2F%2Fwww.nbc.com%2Flate-night-with-seth-meyers%2F&amp;redir_token=2vt7Mt9KUxNxV4c9wzaP_VW385h8MTUyNjI2MjA3NkAxNTI2MTc1Njc2&amp;v=h0zMjd5ZAJ4&amp;event=video_description" class="yt-uix-sessionlink  " data-url="/redirect?q=http%3A%2F%2Fwww.nbc.com%2Flate-night-with-seth-meyers%2F&amp;redir_token=2vt7Mt9KUxNxV4c9wzaP_VW385h8MTUyNjI2MjA3NkAxNTI2MTc1Njc2&amp;v=h0zMjd5ZAJ4&amp;event=video_description" data-sessionlink="itct=CDUQ6TgYACITCOH7u7jHgdsCFUx84AodTtMM4Cj4HUiegeTy3ZGzpocB" data-target-new-window="True" rel="nofollow noopener" target="_blank">http://www.nbc.com/late-night-with-se...</a>

This is how a link comes in VideoDetailFragment private void prepareDescription(final String descriptionHtml). Then there is a function that makes all the links in the text clickable. So we need to get the link between the tags before doing Html.fromHtml

Somethingweirdhere:
I have continued investigating this. To fix it, we need to change the String description. Every time we have a link we need to transform

[...]

PR #1399, opened 2018-05-15
Long links now work correctly.

Somethingweirdhere:

#1390 This should now be fixed.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -60,6 +60,7 @@ import org.schabi.newpipe.extractor.stream.StreamInfo;
 import org.schabi.newpipe.extractor.stream.StreamInfoItem;
 import org.schabi.newpipe.extractor.stream.StreamType;
 import org.schabi.newpipe.extractor.stream.VideoStream;
+import org.schabi.newpipe.extractor.utils.Parser;
 import org.schabi.newpipe.fragments.BackPressable;
 import org.schabi.newpipe.fragments.BaseStateFragment;
 import org.schabi.newpipe.util.StreamItemAdapter;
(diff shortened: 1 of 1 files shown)


theScrabi:
Uff, could be done, but you try to get everything by hand. Take a look at the backend, i recently added a Utill with which you can detect liks in a string.

Somethingweirdhere:
@theScrabi I did, but for https://www.youtube.com/watch?v=6UcGyTD8twE I get 11 Links and the longer one is again shortened. Using `URLDecoder.decode` on the whole String before trying this will return the whole href but I would still need to cut the unwanted parts of href by hand.

Additionally there doesn't seem to be an easy way to replace the links back in an easy way.

image

theScrabi:
Hm the thing is that string parsing should be part of the back end, so it makes sense to do semtheing there. I'll think about what.

Somethingweirdhere:
@theScrabi Alright, great!

theScrabi:
Here you go. By Overriding this methode for the YoutubeStreamExtractor you should be able to set the links specialized for youtube:
https://github.com/TeamNewPipe/NewPipeExtractor/pull/85/files#diff-d2e57174c082b2ead102d53de59728baR147

theScrabi:
Maybe you should think about rewriting the 'getDescription()' function so the links get replaced directly in the actual description.

theScrabi:
I think here is what you would want to do. Fix the links in the description html when it gets returned here:
https://github.com/TeamNewPipe/NewPipeExtractor/blob/master/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeStreamExtractor.java#L151

Somethingweirdhere:
https://github.com/TeamNewPipe/NewPipeExtractor/pull/86

Here you go.

Both requests need to be merged for this to work.

theScrabi:
@Somethingweirdhere what effect does it have to remove the autolinkify line?

theScrabi:
lets reopen this once backend is fixed :)

Somethingweirdhere:
@theScrabi alright, but I am not at home for the next month or so, I will take a look at it when Im back and have a PC again :)

theScrabi:
Ok :)

Somethingweirdhere:
@theScrabi Alright, taking a look at backend. The autolinkify makes it so that if there is a link in the text, it is clickable and goes to where the link says. But we don't want that because it overwrites long links (these are invisible) with the shorter one (that is in the text but not complete).

theScrabi:
:) understood.

PR #1420, opened 2018-05-24
Add everything about GDPR

TobiGr:
This adds

- a dialog which needs to be confirmed before using NewPipe.
- a hint in the about fragment about NewPipe's privacy policy.

@theScrabi We need to release a new version until tomorrow. 
@TheAssassin The policy isn't online yet. If I recall correctly, you wanted to provide it at newpipe.schabi.org/legal/privacy/. Is this correct?

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -20,6 +20,9 @@
 
 package org.schabi.newpipe;
 
+import android.app.AlertDialog;
+import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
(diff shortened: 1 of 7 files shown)


TheAssassin:
This is _not_ at all required. And it is really bad UX.

TheAssassin:
I thought it's better to close this quickly before it gets merged.

Only when the crash reporter is used, a privacy policy is required as personal data is going to be processed. Therefore, such a dialog is always required _there_. But _only_ there. We don't collect any data when the user uses the app normally. This makes it look like we do, though.

Please don't add such a dialog, but rather add it to the crash reporter. Also, an "informed consent" can't be really be gained if you don't make them read the policy. Therefore, in such a case, the "read" button must be the top most button.

The link will be https://newpipe.schabi.org/legal/privacy-policy, I guess.  But I'm fine with https://newpipe.schabi.org/legal/privacy as well.

TobiGr:
I've moved the dialog. It now shows up when you hit the "Report via Mail" button to send a crash report.
@TheAssassin When you have any further change requests, please state them. But I won't be able to address them until tonight, so please do it yourself.

PR #1431, opened 2018-05-27
Downloader: Improve performance for weaker CPUs

DafabHoid:

This improves downloading performance dramatically when the download is cpu bound:
For example, a Galaxy S4 with its 2013 high-end CPU only reaches around 1MB/s, with all cores running at its maximum (and overheating).
By inreasing the chunk size, at which download threads notify its progress (and persist its current state) from only 512 Bytes to 64 KiB, the S4 can easily saturate a 50 MBit/s connection (the maximum I could test).

### DIFF
--- a/app/src/main/java/us/shandian/giga/get/DownloadRunnable.java
+++ b/app/src/main/java/us/shandian/giga/get/DownloadRunnable.java
@@ -2,7 +2,6 @@ package us.shandian.giga.get;
 
 import android.util.Log;
 
-import java.io.BufferedInputStream;
 import java.io.RandomAccessFile;
 import java.net.HttpURLConnection;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


theScrabi:
That kind of makes sense, tho I don't know if that would interfeer with the multi threading thing of gigaget. ... It would make sense to deactivate that anyway.

I will test it.

acrosca:
I would use Okio Sink vs reinventing the wheel.

DafabHoid:
@acrosca I did not reinvent any wheel there, I just merely changed the size of a byte array. (This reduces the count of loop interations, which reduces the count of notifyUpdate calls, which finally reduces the amount of state savings for the DownloadMission),
Indeed would adopting the downloader for Okio Sink be a good idea, but I thought GigaGet is about to be replaced soon anyways, so I see no point in it. Instead, this small PR speeds up downloading a lot, especially on older devices like mine, without a rework of the code.

Issue #1423, opened 2018-05-26
Unable to open downloads


Praveen0899:
- [ ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I checked if the issue/feature exists in the latest version.

Unable to open downloads in 0.13.4

Praveen0899:
It's in 0.13.4

theScrabi:
You aren't able to open downloads in v0.13.4. Can you please describe it further. Do you get a stracktrace, if so please send it.

Praveen0899:
When I click the downloads it's not opening and showing the UI  failure. I sent the UI failure to the mail (    crashreport@newpipe.schabi.org    ).can you check the mail and see the problem

rainecheck:
I'm getting this too, though on 0.13.3. It might be something to do with bad data in the downloads folder, as the problem went away after I cleared out files and data, but returned after I downloaded some other files (one of which also failed to download).
Stacktrace:
```
{
"user_action": "ui error",
"request": "App crash, UI failure",
"content_language": "GB",
"service": "none",
"package": "org.schabi.newpipe",
"version": "0.13.3",
"os": "Linux Android 8.1.0 - 27",
"time": "2018-05-30 10:04",
"exceptions": [
[...]

PR #1441, opened 2018-06-01
Downloader: Fix crash on loading unfinished downloads from .giga files (Fixup for release builds)

DafabHoid:

This is a fixup for #1407 , which fixes the crash in release builds, too. It keeps Proguard from removing the new method "private void readObject(java.io.ObjectInputStream)", which is only called by the VM, but not from the code, and therefore considered useless by Proguard in release (minify) mode.

This PR adds some new Proguard rules to keep class members related to serialization for Serializable classes like DownloadMission.


DafabHoid:
This fixes #1423 for me.

Issue #1215, opened 2018-03-25
0.13.0 beta - Polish and English are both displayed at the same time


slyscorpion:
Hi, 

I checked the issue tracker but I couldn't find if my issue exists or my github-fu skills are lacking.

In the 0.13.0 beta for NewPipe there are some Polish words mixed in with English when I open up NewPipe.

https://i.imgur.com/DuhQwM1.jpg

If you look at the attached screenshot, you will see the words "2 dni temu" (2 days ago) mixed in with everything else.

My phone's language is set to English (United States) and my input languages are set English (United States) and Polish. My region in my Android settings is set to Poland. The content country in my install of NewPipe is set to United Kingdom. Figured I should mention that in case any of what I just mentioned has a bearing on this issue.

Let me know if you need any further info.

theScrabi:
2 dni temu is generated localy on your phone, while the rest  of the item is downloaded. If the content country and your language are different this will mix.

slyscorpion:
I just changed my content country to USA in NewPipe to match my language which is English (United States) and I still get the "2 dni temu" text.

theScrabi:
Ah ... well al right are you in Poland right now?

I might have told you wrong, the date is not yet parsed and locally generated as I said.
In fact the date is generated by Youtube, and it depends on your content language. However NewPipe does respect your content countery, which means its which video your content language, which means in which language Youtube presents meta information. 

We discuses this before. Actually NewPipe should respect both.

Bug confirmed.

slyscorpion:
Hi @theScrabi yes I am located in Poland.

Stypox:
Related: #578 #1130 #1132

PR #1465, opened 2018-06-09
Fix upload date field

### LINKED ISSUE #1457: Bad upload date localization
The upload date field is showing a wrong language (maybe similar to #1215):
How? setting **Default content country** to any non-english language (in my case spanish).

Screenshot of **Trending** showing the _upload date_ in english:

Search results also too, but in arab:

kapodamy:
Implement correctly **Accept-Language** HTTP header, also require some changes on **NewPipeExtractor** to work (https://github.com/TeamNewPipe/NewPipeExtractor/pull/91).

Fixes #1457 #1215

✔ I carefully read the contribution guidelines and agree to them.


theScrabi:
I quickly looked over the PR on the extractor.

I think the downloader should be stateless. Putting the language code there does not seem to be a good idea. I think it was better to set it in the newpipe service and then just add a language/locale code to the downloader.download() methode.

app/src/main/java/org/schabi/newpipe/Downloader.java
```diff
@@ -40,10 +40,13 @@
 
 public class Downloader implements org.schabi.newpipe.extractor.Downloader {
     public static final String USER_AGENT = "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:43.0) Gecko/20100101 Firefox/43.0";
+    private static final String ACCEPT_LANGUAGE = "Accept-Language";
 
     private static Downloader instance;
     private String mCookies;
     private OkHttpClient client;
+    private static String countryCode = "GB";
```
theScrabi: Still not stateless. I think the country, and language should be controlled by the back end, not by the downloader, which is part of the frontend.

theScrabi:
Please squash fix1 and fix2 and give it a more meaningfull name :)

lf-:
Some of the comments in this contribution could be improved. For instance, the reference to line numbers will be unhelpful as soon as that file is modified, so it should be referring to something more general or just be self contained in the first place. 

Also, there is a commented out line of code which should probably get deleted.

Issue #1485, opened 2018-06-14
Add download option when sharing video to newpipe?


Somethingweirdhere:
I feel like we could make a new share intent receiver activity for audio and video download, so after setting default downlaod values you could download videos from the youtube app quite easily. Just pressing sharea and selecting Newpipe Audio Downloader. 

And if that's too much, maybe just adding downlaod to the current share intent receiver acitvity would be ok?

theScrabi:
This got requested multiple times now, and I think its a good idea to add a download option to the share receiver activity :)

Somethingweirdhere:
Alright, I will work on it as soon as soon as my other PRs are done!

PR #1498, opened 2018-06-20
Download option in share menu

Somethingweirdhere:
This addresses #1485 and adds an option to download within the share menu.


theScrabi:
Ok, fine so far, but you should also add a option to this menu here, which you will find in the settings.
screenshot_20180625-104558

wb9688:
@theScrabi: Wrong paste?

Somethingweirdhere:
Here we go :)

PR #1510, opened 2018-06-26
Library version update and database backup fix

karyogamy:
This updates bumps the version of a few libraries and was part of #1392, but since it is isolated enough, I've decided to pull it out.

Starting from version 1.1.0, Room db changes its journal file into two separate accessory files, namely db-wal and db-shm. These files do not store any important data after a checkpoint or db connection closes. However, when importing new databases, if any of the accessory files do not match, the existing core database will be wiped. So, in order to provide backward compatibility, I've decided to remove all accessory files when a database is imported. 

The only problem is that, if the database of a newer version needs to be imported to a version before this PR, then the user needs to manually clear the app data from the system settings. @theScrabi What do you think about this?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -43,7 +43,7 @@ android {
 ext {
     supportLibVersion = '27.1.1'
     exoPlayerLibVersion = '2.7.3'
-    roomDbLibVersion = '1.0.0'
+    roomDbLibVersion = '1.1.1'
     leakCanaryLibVersion = '1.5.4'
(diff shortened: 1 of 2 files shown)


theScrabi:
> The only problem is that, if the database of a newer version needs to be imported to a version before this PR, then the user needs to manually clear the app data from the system settings. @theScrabi What do you think about this?

I think we don't really have to support downwards compatibility with NewPipe, so I'd be OK with that.

PR #1516, opened 2018-06-28
Long-tap delete and share in subscriptions

Somethingweirdhere:
I got so far, but I'm kinda stuck. I am calling delete but it doesn't do anything. I would try instantiating a channelfragment, but this would try to get the Channel and that won't work as we know for deleted channels. I guess delete only works if we have the exact same SubscriptionEntity, and creating one from a ChannelInfoItem might be different in some way? Or what might be the problem?


karyogamy:
> I am calling delete but it doesn't do anything

This is likely because the entity created does have a database entry uid, which is what room db uses to uniquely identify entries. Only if the entity is directly obtained from the database, its uid will be valid and can then be used for deletion.

For this task, I suggest looking at how remote playlist is obtained and deleted in `RemotePlaylistManager` (`getPlaylist` and `deletePlaylist`), and then add those methods to `SubscriptionService` for convinience. You might also need to add some database calls in `SubscriptionDAO`.

If you have a bit SQL knowledge, you can simplify this by adding a database call which removes all entry given the service id and the url rather than requiring the uid, which should always be unique as described in `SubscriptionEntity` indices.

Also, please don't use `AsyncTask` as we already have RxJava. If you create a combined database call like above, then it should be easy to write the rx call just by looking around the codebase. Otherwise, you might want to take a look on how to chain rxjava calls to do get and remove in one call.

Somethingweirdhere:
@karyogamy Thanks :) No SQL Knowledge and never heard of RxJava, so I will first need some time to learn all that.

Somethingweirdhere:
Hey how about that, I got it :)

app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.java
```diff
@@ -330,6 +359,63 @@ public void selected(ChannelInfoItem selectedItem) {
...
+
+    }
+
+    private void shareChannel (ChannelInfoItem selectedItem) {
+        shareUrl(selectedItem.getName(), selectedItem.getUrl());
+    }
+
+    @SuppressLint("CheckResult")
+    private void deleteChannel (ChannelInfoItem selectedItem) {
+        ExtractorHelper.getChannelInfo(selectedItem.getServiceId(), selectedItem.getUrl(), true)
```
karyogamy: After you subscribe, a `disposable` will be created to run the code in a different thread which is not killed when the fragment is, so it is a good idea to create a `CompositeDisposable` at the fragment level and put all the disposables in there. When the fragment is destroyed (i.e. `onStop()`), you will need to clean and dispose all the items in the dispoable.

app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.java
```diff
@@ -330,6 +359,63 @@ public void selected(ChannelInfoItem selectedItem) {
...
+                .subscribeOn(Schedulers.io())
+                .observeOn(Schedulers.newThread())
+                .subscribe((@NonNull ChannelInfo result) -> {
+                    List<SubscriptionEntity> toDelete = subscriptionService.subscriptionTable()
+                            .getSubscription(result.getServiceId(), result.getUrl())
+                            .blockingFirst();
```
karyogamy: By convention, the code blocks in `subscribe` should run on `mainThread` since that is the only thread where UI changes can happen. I know that the UI changes for subscriptions occur on a different rx subscription, but this having `subscribe` run main thread is the convention and what devs usually expect it to be. 

What you want to do is chain a `onNext` after `subscribeOn` to run the deletion on io thread, and then use `observeOn` to switch back to main thread to subscribe, perhaps with a toast (which is UI thread only).

app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.java
```diff
@@ -330,6 +359,63 @@ public void selected(ChannelInfoItem selectedItem) {
...
+                    subscriptionService.subscriptionTable().delete(toDelete);
+                });
```
karyogamy: 'subscribe' is also the place where you catch any errors occurring in the async call, so you will also need to implement what happens when any of the logic fails in progress.

karyogamy:
Nice work, but please remember to leave some more informative commit messages and clean up the imports before merge.

Somethingweirdhere:
Alright, thanks for the help :)

mauriciocolli:
Here's some thoughts:

As @karyogamy mentioned, you could have deleted the subscription using only its service id and url. Some resources (aside from those) that may help you with that:
- [`SubscriptionDAO.getSubscription`](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/database/subscription/SubscriptionDAO.java#L34..L37)
- SQLite's [`DELETE`](https://www.sqlite.org/lang_delete.html)

The `ExtractorHelper.getChannelInfo(...)` call will gonna fail if, for example, the channel is deleted or the user have no internet available. We should _not_ make a network request for that (we already have all the information we need). This problem is exactly what prevents the user to unsubscribe from deleted channels right now.

The `disposables` problem that you fixed, are now reverted by the last commit.

So this have to be fixed: 
- Create a delete method and use that, or get and delete in one go.
- Don't use `blockingFirst`, `toObservable` or `scheduleDirect`, look around the codebase to see how it's done.

Somethingweirdhere:
@mauriciocolli 

Hello, and thanks for reviewing :) 

To me the disposables look fine, I just changed the location where I add them to onSubscribe(). Does that work too? Or do I need to add them when calling subscribe()?

Also i now removed the usage of blockingFirst and scheduleDirect(). I hope toObservable() is fine, because it's how ChannelFragment handles subscribing:

```
        final Observable<List<SubscriptionEntity>> observable = subscriptionService.subscriptionTable()
                .getSubscription(info.getServiceId(), info.getUrl())
                .toObservable();
```

mauriciocolli:
> I just changed the location where I add them to onSubscribe()

My bad, didn't saw. And yeah, it's a valid practice.

> I hope toObservable() is fine

I think you should try what was said in the above comments (i.e. create a proper delete method, seeing how it was done in other parts of the app), `getSubscription` is just not meant for that, and it would have to be changed later anyway (simply put: this function returns a `Flowable` that continues to emit items when there are changes in the underlying database).

If you still have trouble adding those functions, tell me. I'll push a commit that will add them if needed.

---

###### **Offtopic**: If you want to have your profile linked properly as the author in commits, [see this](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

Somethingweirdhere:
@mauriciocolli 

Thanks for the profile hint, it seems to work :)

For the rest, yeah, I really have no idea what to do, and would probably need to look into RxJava a lot more. If you could fix what is wrong here it would be great, I hope I can learn from it.

But also, I don't understand the getSubscription thing. Wasn't it the method you suggested earlier to me?

theScrabi:
@mauriciocolli what is the status of you review, is it fixed, or does it still count as blocked?

theScrabi:
@Somethingweirdhere what is the status, is it ready to merge? @mauriciocolli didn't seem to have accepted it yet.

Somethingweirdhere:
@theScrabi You can unsubscribe without a connection now. I still haven't been able to test if non-existant channels work too, but I strongly assume so, because my implementation never accesses the youtube servers.

It might not be the cleanest way, but as you said, better than nothing.

PR #1530, opened 2018-07-13
resolved some errors and updated dependencies

TacoTheDank:
some dependency updates (including build-tools to 28.0.1)

updated .travis.yml

updated gradle-wrapper.properties to 4.8.1

fixed YoutubeStreamExtractor not being resolved, along with an error with kioskId (dunno if the kioskId fix was done right tho)

paceholder -> placeholder because spelling lol

made app movable to sd card because srsly why the hell do people not make their apps movable? people dont necessarily need to, but at least give people the ability to if they want to. storage isnt limitless and, surprise, there are people who use older phones (that dont have as much space internally)


app/build.gradle
```diff
@@ -41,7 +41,7 @@ android {
 }
 
 ext {
-    supportLibVersion = '27.1.1'
+    supportLibVersion = '28.0.0-alpha3'
```
theScrabi: Please don't use an alpha release.

What advantages would 28.0.0 have over 27.1.1?

TacoTheDank:
well alright then... ive had no issues personally with using the alpha releases but if you would rather not use them then thats fine too

theScrabi:
Ey calm down pls, I'm about to move back to germany right now. I don't have to much time right now, but once I do I'll handle this. Pls be patient.

TacoTheDank:
ah nice, i went there during spring break last year

PR #1561, opened 2018-07-28
Link redirection check

kapodamy:
This feature check if the provided link (via "share menu" from another apps) if a youtube url shorted, anonymized or deferred. For example [https://t.co/yi7xTNnPWQ](https://t.co/yi7xTNnPWQ)

**REQURES** Jsoup library for [http-equiv](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-http-equiv) check and TeamNewPipe/NewPipeExtractor#91

This feature can be enabled/disabled from settings:
demo

**✔** I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.


MadderRagax:
Sure good idea but this should definitely go in the extractor. In my opinion this shouldn't even be a setting just make it the default behavior. But that's not really up to me.

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -611,4 +644,103 @@ private String trim(final String input) {
...
     }
+
+    /**
+     * get the redirection url (if exists)
+     *
+     * @param url the url to check
+     * @return the redirection, or {@code null} if there is no such redirection
+     */
+    private static String getRedirectionOf(String url) {
+        HttpURLConnection con = null;
```
theScrabi: We use okhttp not HttpUrlConnection

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -611,4 +644,103 @@ private String trim(final String input) {
...
+                        }
+                    }
+                    break;
+                case HttpURLConnection.HTTP_MOVED_TEMP:
+                case HttpURLConnection.HTTP_MOVED_PERM:
+                case HttpURLConnection.HTTP_SEE_OTHER:
+                    return con.getHeaderField("Location");
+            }
+        } catch (Exception e) {
+            Log.e(TAG, "Failed to check for redirection in ".concat(url), e);
```
theScrabi: We should use ErrorActivity not normal logs.

theScrabi:
I don't know about this. Following redirects is not really or duty. You could make this be done with a browser and that just open the url you get from it there. 

The thing is this is quite complicated code (it even needs an addtional lib), and it only works for sharing the links, not for opening it.

app/src/main/java/org/schabi/newpipe/RouterActivity.java
kapodamy: is for debug only

PR #1581, opened 2018-08-04
Russian strings values-ru-sw330dp

bboa:
Hi, please add this little hack for Russian translation.


theScrabi:
:D interesting solution

TobiGr:
@theScrabi I got this :grinning: I like the idea of using abbreviations and shorter words for smaller screens.

mauriciocolli:
I don't think it's a good solution for this.

A better one would be to use a button instead as it would be more intuitive and easy to spot. Or just use the smaller string till then.

Issue #1306, opened 2018-04-12
Option To Close NewPipe In The Popup Window


GKid94:
Can you please add an option to close NewPipe in the popup window?

karyogamy:
Probably not, since we prefer not adding redundant UI elements and the popup window already can be closed via the notification. Also, although not an obvious feature, the popup window can be closed by hard-swiping the window left or right.

GKid94:
I don't have any notifications and I'm swiping the popup window but it doesn't go away

theScrabi:
Can you show us that via screenshot?

GKid94:
![screenshot_20180421-152515](https://user-images.githubusercontent.com/18757054/39084573-750eb2fe-4578-11e8-9154-4b3aeb1ceb5a.png)

GKid94:
I just found out that I blocked all notifications from New Pipe that's why I don't see the notification, can you improve the swipe popup closing?

karyogamy:
We will probably introduce a parameter in the settings to let users increase the sensitivity of the swipe and implement PIP for Oreo.

Issue #979, opened 2018-01-10
Closing popout


p3lim:
I'm using this mainly to view videos from other apps, and the popout mode is great, but I can't seem to close the video without first going fullscreen, then leaving the full app.
A close button or a swipe action to close would be nice.

wb9688:
There's a close button in it's notification

TobiGr:
You can also swipe the popup to the right side of your screen. This closes it too.

p3lim:
> There's a close button in it's notification

Thanks, although not as convenient as gestures or a dedicated button.
> You can also swipe the popup to the right side of your screen. This closes it too.

First thing I tried, doesn't seem to work for me atleast (EMUI 8, Oreo 8.0).
The only "gestures" that seem to work (basically swiping) is moving the popout up/down on the screen.

PR #1597, opened 2018-08-13
New way to close the popup player

### LINKED ISSUE #658: Pop up swipe to exit.

mauriciocolli:
- Add an overlay at the bottom of the screen that the user can drag the popup to, closing it.

    A little demo showing how it works:
    

- Do not show the player controllers when moving, since the user have to click in the popup
anyway now (#1342).

---
Closes #658
Related #979, #1306

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
@@ -928,9 +928,6 @@ public final class PopupVideoPlayer extends Service {
         public boolean onScroll(MotionEvent initialEvent, MotionEvent movingEvent, float distanceX, float distanceY) {
             if (isResizing || playerImpl == null) return super.onScroll(initialEvent, movingEvent, distanceX, distanceY);
 
-            if (playerImpl.getCurrentState() != BasePlayer.STATE_BUFFERING
-                    && (!isMoving || playerImpl.getControlsRoot().getAlpha() != 1f)) playerImpl.showControls(0);
-
             if (!isMoving) {
                 animateView(closeOverlayButton, true, 200);
             }
--- a/app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
@@ -116,7 +116,6 @@ public final class PopupVideoPlayer extends Service {
     private FloatingActionButton closeOverlayButton;
     private WindowManager.LayoutParams closeOverlayLayoutParams;
 
-    private int shutdownFlingVelocity;
     private int tossFlingVelocity;
 
     private float screenWidth, screenHeight;
@@ -200,7 +199,6 @@ public final class PopupVideoPlayer extends Service {
         View rootView = View.inflate(this, R.layout.player_popup, null);
         playerImpl.setup(rootView);
 
-        shutdownFlingVelocity = PlayerHelper.getShutdownFlingVelocity(this);
         tossFlingVelocity = PlayerHelper.getTossFlingVelocity(this);
 
         updateScreenSize();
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
```diff
@@ -969,5 +1092,43 @@ private boolean handleMultiDrag(final MotionEvent event) {
 
             return true;
         }
+
+        /*//////////////////////////////////////////////////////////////////////////
+        // Utils
+        //////////////////////////////////////////////////////////////////////////*/
+
+        private float map(float value, float start1, float stop1, float start2, float stop2) {
```
karyogamy: Might be a good idea to add some comments here to describe what map does as well as some explanations for the constants `0x88` and `1.15f` used below as well.

app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
```diff
@@ -894,21 +1014,21 @@ public boolean onFling(MotionEvent e1, MotionEvent e2, float velocityX, float ve
             final float absVelocityX = Math.abs(velocityX);
             final float absVelocityY = Math.abs(velocityY);
             if (absVelocityX > shutdownFlingVelocity) {
-                onClose();
+                closePopup();
```
karyogamy: Since now we have a proper quick close option, it'd be a good time to remove this swipe to close (shutdown fling) feature.

mauriciocolli: Removed.

app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
mauriciocolli: The`map()` was removed, and comment made about the percentage.

mauriciocolli:
> disable or make invisible the button when soft keyboard is open.

Is there a way to do that reliably (or at all)?

> move the button on top of the keyboard

Was the chosen solution.

But damn, was it a buggy road. I finally figured out a way to do that work: _do not_ mess with visibility of the view being added to the window manager. Maybe it was messing with the view being measured? Problem with using `MATCH_PARENT`? Well, it all sorted it out at the end...

Yet, there's still a major bug left: the popup don't have the right position if it was positioned in the keyboard area when it appears. Maybe the solution will be something similar to how the the close overlay works, a `MATCH_PARENT` view. We can use it to detect its size changing (the draggable area) and update its boundaries accordingly. Didn't mess with it yet, something for another time/PR...

> It seems excessive to have both the popup overlay turning red and the close button shaking 

I have removed the shaking. It was meant to make the closing action more obvious when the overlay was not that visible, but that doesn't happen very often (basicly just pure red videos, or whatever the color the overlay could be). I also removed the overlay alpha-distance mapping, so it even further make the shaking unnecessary.

app/src/main/java/org/schabi/newpipe/util/AnimationUtils.java
```diff
@@ -241,6 +241,33 @@ public void onAnimationEnd(Animator animation) {
                 }).start();
     }
 
+    public static ValueAnimator shakeView(View view) {
```
karyogamy: This can be removed now.

karyogamy:
> Is there a way to do that reliably (or at all)?

There doesn't seem to be an easy way, I might be talking out of my ass when suggesting this solution =\

> But damn, was it a buggy road.

Didn't expect it to have that much resistance, what was the buggy behaviour?
But in any case, good work, I think this is good to go. 
Will be testing it out on a real device the next couple of days.

mauriciocolli:
> what was the buggy behaviour?

The overlay would not have the correct height, would clip through the keyboard, buggy exit animation and wouldn't have consistency across the APIs versions. Tried multiple combinations of flags and other things.

Basically, it would not have the correct dimensions, and I think the reasons are those that I mentioned earlier (views were `GONE` and wouldn't measure correctly I guess, even though I think it should when it was made `VISIBLE` again). Stupid mistake maybe...

---

> the popup don't have the right position if it was positioned in the keyboard area when it appears

And re-reading what I wrote I just had this "brilliant" idea: why not using the _overlay_ layout changes to correct the popup's position? I'll do some tests and maybe push a commit later.

karyogamy:
I see what you did there, good one.

theScrabi:
Cant you still close through the notification?

adihrustic:
> Cant you still close through the notification?
Yes. But that movement is more awkward using one hand than moving the player to the close button.

PR #1601, opened 2018-08-14
small code clean

BO41:
this should not change the code behavior.
I think an if statement is more fitting than an switch-case with just one case

- [ x ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.


Somethingweirdhere:
Seems to me like for a selection menu, a switch makes more sense. And in the future, we might have to add another menu element.

theScrabi:
I agree to @Somethingweirdhere. However @BO41 however keep it coming :)

Issue #1603, opened 2018-08-14
Improve and shrink resources


TobiGr:
I took a look at the resources and I think that there is potential to minimize NewPipe's size further.

### 1. Remove duplicate resource strings
In main/res/values/strings.xml you will find several identical strings/values with different _names_:

- Play: `play_btn_text`, `view`
- Video preview thumbnail: `list_thumbnail_view_description`, `detail_thumbnail_view_description`
- Popup: `controls_popup_title`, `settings_category_popup_title`
- Error: `general_error`, `msg_error`
- Downloads: `downloads`, `downloads_title`
- Audio: `play_audio`, `audio`
- History: `title_activity_history`, `action_history`
- NewPipe: `app_name`, `preferred_open_action_share_menu_title`
- Download: `download`, `download_dialog_title`, `settings_category_downloads_title`
- Settings: `settings`, `action_settings`
- About: `action_about`, `tab_about`

By removing at least one duplicate occurrence, we'd not only reduce the effort it needs to translate all parts of the app but also the APK size. In some cases it makes sense to not remove the whole string and its value, but replace the value with a reference to another string:
``` XML
<string name="downloads_title">Downloads</string>
=>
<string name="downloads_title" translatable="false">@string/downloads</string>
```
### 2. Remove unused strings

There are several unused strings which we should consider to remove. Unfortunately, Android Studio didn't do a good job while searching for unused resources and removed various required strings. For this reason, I wrote a small programme which is more reliable in finding unused strings. Nevertheless we should discuss what to do with strings which are likely to be used in the future. (e.g everything about TOR).

### 3. Optimize drawables

Our images need quite much memory. We should use vector graphics instead of PNGs (see #1199 for more info). There were some attempts to do this, but they got rejected. Until this is done, we can minimize the images manually or via [ImgBot](https://imgbot.net/).

---

But before I dig a bit deeper in the optimization, I'd like to hear the common opinion on this. @theScrabi @wb9688 @mauriciocolli @karyogamy

TobiGr:
> But remember that we should make our png files compatible with fdroid reproducable build.

Did you have problems with PNG files?

F-Droid's [docs](https://f-droid.org/en/docs/Reproducible_Builds/) say, thee shouldn't be a problem:

[...]

Issue #503, opened 2017-04-15
ExoPlayer volume/brightens too sensitive 


theScrabi:
@mauriciocolli 

The left and right area for brightens, and volume are a little bit to sensitive. Can you please make them less sensitive.

![screenshot_20170415-132720](https://cloud.githubusercontent.com/assets/1891273/25063270/0755532e-21e0-11e7-8151-880118904c70.png)

Generator:
Notice that too, swipe controls way too sensitive.  
Something more smooth like on VLC would be nice.

mauriciocolli:
Just a temporary fix, I guess I'll rewrite the gesture detector when I have more time...

ildar:
devs: logarithmic scale is the key here IMHO. Emm, exponential to be precise.
@theScrabi : please, tOO, not tO

theScrabi:
> @theScrabi : please, tOO, not tO

What do you mean by that?

utack:
He is a grammar comcast, he literally only meant the spelling of the word 'too'

wb9688:
@theScrabi: He meant that you wrote 'to', but it should be 'too'. 'to' is
used before inifinitive verbs, while 'too' is an adverb like 'extremely'.

ildar:
😀 I'm not really that "comcastish" harsh usually. It's just the error
makes the issue title obscure, I had to dive in just to get its meaning.
Sorry for inconvenience.

theScrabi:
Oh I see :P. Well... thank you.

TheMatten:
@theScrabi #1604 introduced new gestures behavior - do you think it resolves this issue?

theScrabi:
Lets se, if not we reopen it.

PR #1604, opened 2018-08-15
Gesture controls

TheMatten:

Current gesture controls work but look like an afterthought. This PR introduces circular design with progress bar and material icons that changes depending on current value.

Demo:
screenrecord-2018-08-15-12-00-51

Changed version seems to work without any problems.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
@@ -36,6 +36,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v4.app.ActivityCompat;
 import android.support.v7.app.AppCompatActivity;
+import android.support.v7.content.res.AppCompatResources;
 import android.support.v7.widget.RecyclerView;
 import android.support.v7.widget.helper.ItemTouchHelper;
 import android.util.DisplayMetrics;
@@ -46,7 +47,9 @@ import android.view.MotionEvent;
 import android.view.View;
 import android.view.WindowManager;
 import android.widget.ImageButton;
+import android.widget.ImageView;
 import android.widget.PopupMenu;
+import android.widget.ProgressBar;
 import android.widget.RelativeLayout;
 import android.widget.SeekBar;
 import android.widget.TextView;
@@ -82,6 +85,7 @@ import java.util.UUID;
 import static org.schabi.newpipe.player.BasePlayer.STATE_PLAYING;
 import static org.schabi.newpipe.player.VideoPlayer.DEFAULT_CONTROLS_DURATION;
 import static org.schabi.newpipe.player.VideoPlayer.DEFAULT_CONTROLS_HIDE_TIME;
+import static org.schabi.newpipe.util.AnimationUtils.Type.SCALE_AND_ALPHA;
 import static org.schabi.newpipe.util.AnimationUtils.Type.SLIDE_AND_ALPHA;
 import static org.schabi.newpipe.util.AnimationUtils.animateRotation;
 import static org.schabi.newpipe.util.AnimationUtils.animateView;
@@ -365,10 +369,16 @@ public final class MainVideoPlayer extends AppCompatActivity
 
     @SuppressWarnings({"unused", "WeakerAccess"})
     private class VideoPlayerImpl extends VideoPlayer {
+        private final float MAX_GESTURE_LENGTH = 0.75f;
+
         private TextView titleTextView;
         private TextView channelTextView;
-        private TextView volumeTextView;
-        private TextView brightnessTextView;
+        private RelativeLayout volumeRelativeLayout;
+        private ProgressBar volumeProgressBar;
+        private ImageView volumeImageView;
+        private RelativeLayout brightnessRelativeLayout;
+        private ProgressBar brightnessProgressBar;
+        private ImageView brightnessImageView;
         private ImageButton queueButton;
         private ImageButton repeatButton;
         private ImageButton shuffleButton;
@@ -392,6 +402,8 @@ public final class MainVideoPlayer extends AppCompatActivity
         private RelativeLayout windowRootLayout;
         private View secondaryControls;
 
+        private int maxGestureLength;
+
         VideoPlayerImpl(final Context context) {
             super("VideoPlayerImpl" + MainVideoPlayer.TAG, context);
         }
@@ -401,8 +413,12 @@ public final class MainVideoPlayer extends AppCompatActivity
             super.initViews(rootView);
             this.titleTextView = rootView.findViewById(R.id.titleTextView);
             this.channelTextView = rootView.findViewById(R.id.channelTextView);
-            this.volumeTextView = rootView.findViewById(R.id.volumeTextView);
-            this.brightnessTextView = rootView.findViewById(R.id.brightnessTextView);
+            this.volumeRelativeLayout = rootView.findViewById(R.id.volumeRelativeLayout);
+            this.volumeProgressBar = rootView.findViewById(R.id.volumeProgressBar);
+            this.volumeImageView = rootView.findViewById(R.id.volumeImageView);
+            this.brightnessRelativeLayout = rootView.findViewById(R.id.brightnessRelativeLayout);
+            this.brightnessProgressBar = rootView.findViewById(R.id.brightnessProgressBar);
+            this.brightnessImageView = rootView.findViewById(R.id.brightnessImageView);
             this.queueButton = rootView.findViewById(R.id.queueButton);
             this.repeatButton = rootView.findViewById(R.id.repeatButton);
             this.shuffleButton = rootView.findViewById(R.id.shuffleButton);
@@ -461,6 +477,20 @@ public final class MainVideoPlayer extends AppCompatActivity
             toggleOrientationButton.setOnClickListener(this);
             switchBackgroundButton.setOnClickListener(this);
             switchPopupButton.setOnClickListener(this);
+
+            getRootView().addOnLayoutChangeListener((view, l, t, r, b, ol, ot, or, ob) -> {
+                if (l != ol || t != ot || r != or || b != ob) {
+                    // Use smaller value to be consistent between screen orientations
+                    // (and to make usage easier)
+                    int width = r - l, height = b - t;
+                    maxGestureLength = (int) (Math.min(width, height) * MAX_GESTURE_LENGTH);
+
+                    if (DEBUG) Log.d(TAG, "maxGestureLength = " + maxGestureLength);
+
+                    volumeProgressBar.setMax(maxGestureLength);
+                    brightnessProgressBar.setMax(maxGestureLength);
+                }
+            });
         }
 
         public void minimize() {
@@ -872,12 +902,28 @@ public final class MainVideoPlayer extends AppCompatActivity
             return channelTextView;
         }
 
-        public TextView getVolumeTextView() {
-            return volumeTextView;
+        public RelativeLayout getVolumeRelativeLayout() {
+            return volumeRelativeLayout;
         }
 
-        public TextView getBrightnessTextView() {
-            return brightnessTextView;
+        public ProgressBar getVolumeProgressBar() {
+            return volumeProgressBar;
+        }
+
+        public ImageView getVolumeImageView() {
+            return volumeImageView;
+        }
+
+        public RelativeLayout getBrightnessRelativeLayout() {
+            return brightnessRelativeLayout;
+        }
+
+        public ProgressBar getBrightnessProgressBar() {
+            return brightnessProgressBar;
+        }
+
+        public ImageView getBrightnessImageView() {
+            return brightnessImageView;
         }
 
         public ImageButton getRepeatButton() {
@@ -886,6 +932,10 @@ public final class MainVideoPlayer extends AppCompatActivity
 
         public ImageButton getPlayPauseButton() {
             return playPauseButton;
+        }
+
+        public int getMaxGestureLength() {
+            return maxGestureLength;
         }
     }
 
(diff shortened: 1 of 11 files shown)


mauriciocolli:
Thanks for the pull request, looks good!

> Current gesture controls work but look like an afterthought

Yep, that's right, there's even a `TODO` there. I think you're talking about just the UI, but the code that handle gestures isn't really the best way to do it and it was meant as just a working solution, but yeah... (#503)

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
```diff
@@ -992,12 +1042,31 @@ public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float d
                 getWindow().setAttributes(lp);
                 if (DEBUG) Log.d(TAG, "onScroll().brightnessControl, currentBrightness = " + currentBrightness);
                 int brightnessNormalized = Math.round(currentBrightness * 100);
+                animateProgressBar(
+                        playerImpl.getBrightnessProgressBar(),
+                        (brightnessNormalized == 1 ? 0 : brightnessNormalized),
+                        100
+                );
+                if (brightnessNormalized < 25) {
+                    playerImpl.getBrightnessImageView().setImageDrawable(
```
mauriciocolli: Crash in old versions of Android.
A very simple solution, thankfully: use `AppCompatResources.getDrawable` instead.

I was also worried about performance, but [it seems to be cached](https://chromium.googlesource.com/android_tools/+/refs/heads/master/sdk/sources/android-25/android/support/v7/widget/AppCompatDrawableManager.java#180), both the Android and compat library versions, so it's all good.

app/src/main/java/org/schabi/newpipe/util/AnimationUtils.java
```diff
@@ -43,6 +46,26 @@
...
     }
 
+    /**
+     * Animate the progress bar's progress
+     *
+     * @param progressBar progress bar that will be animated
+     * @param to          final progress value after animation
+     * @param duration    how long the animation will take, in milliseconds
+     */
+    public static void animateProgressBar(ProgressBar progressBar, int to, long duration) {
```
mauriciocolli: Put this function below the others, but on top of the `Internals` block.

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
TheMatten: Thanks for that - AS stopped complaining when I used it through getResources(), but yeah, I didn't check it manually...

app/src/main/java/org/schabi/newpipe/util/AnimationUtils.java
TheMatten: OK, I'll do

TheMatten:
@mauriciocolli Design seemed like a biggest problem, but while tinkering with it, I thought about implementing some sort of speed-based behavior, so that user can change values more precisely by moving finger slowly. May try it, as well as maybe introduce some "fake" state of progress bar that rounds to the closest value after gesture, instead of that choppy movement.

TheMatten:
@mauriciocolli While talking about gesture controls, there's one more thing that doesn't feel right - when using fast rewind controls, bright background may make icons barely visible:

np-rewind-nobg

E.g. YT web player solves this problem with dark circle as a background:
yt_circle_back_opt

The other thing we could do is temporarily dim the background (as with play/pause), but then rewind would still be in conflict with play button.

TheMatten:
Wow - it seems like GestureDetector's and GestureDetectorCompat's onScroll gets called differently. Luckily, Compat version seems to be more sensitive (it works with slow movement).

After a lot of testing, to me the best solution seems to use event's distance directly as a value for progress incrementation and set max progress value based on screen size. Speed seems to be inconsistent and attempts to "normalize" value before use by e.g division or log result in number rounded down to zero when doing small movements.

I could maybe solve this by some overcomplicated solution I didn't think of or by skipping every Nth event and using progress animation again, but I don't think it would give us any advantage. Maybe if drawing on screen was too expensive, but actually, number of total motion events (including historical) should be <= touchscreen scan rate, which will probably match framerate (I've counted <= 60 per second on my device with `input swipe` command). And in reality, profiler barely shows any difference in load while using new implementation.

At the end, we can think of space used for gesture as an invisible seekbar, which is predictable, precise and smooth solution at the same time (and allows us to get from 0% to 100% at one long swipe). Value from it gets converted to nearest volume/brightness level and applied.

Demo:
screenrecord-2018-08-20-09-41-22

TheMatten:
(Sorry for force push - I reworded one commit message)

mauriciocolli:
> when using fast rewind controls, bright background may make icons barely visible

IMO it doesn't matter much, the user is focused on the video and will certainly notice the rewind/fast forward. Though I don't mind if you want to make more visible, feel free to do it (of course, in a way that doesn't draw too much attention).

And the fact that the icons stay very shortly in screen (video is already loaded) makes the player seems snappy, though the "artifacts" (progress bar visible in a split second, for example) can be seen as a buggy player UI to the user. The optimal solution would be to keep the short animation and fix them.

> we can think of space used for gesture as an invisible seekbar

Exactly, I like this implementation, works and feels much better.

---

Oh, and if you don't mind, please, when this PR is ready remove the commits that cancel each other.

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
```diff
@@ -392,6 +402,18 @@ public void onPlaybackParameterChanged(float playbackTempo, float playbackPitch,
         private RelativeLayout windowRootLayout;
         private View secondaryControls;
 
+        private int maxGestureLength;
+
+        {
+            DisplayMetrics m = getResources().getDisplayMetrics();
```
mauriciocolli: What do you think of using the `rootView`'s width and height?

The code would be simpler, and using this approach, we could already support the multi window mode just by adding a listener to the `rootView` layout changes, that when we detect one, we could simply update the variable and the gesture would work correctly in the new size.

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
```diff
@@ -930,21 +979,9 @@ public boolean onDown(MotionEvent e) {
...
-        private final String brightnessUnicode = new String(Character.toChars(0x2600));
-        private final String volumeUnicode = new String(Character.toChars(0x1F508));
+        private int maxVolume = playerImpl.getAudioReactor().getMaxVolume();
 
         private final int MOVEMENT_THRESHOLD = 40;
-        private final int eventsThreshold = 8;
-        private boolean triggered = false;
-        private int eventsNum;
 
         // TODO: Improve video gesture controls
```
mauriciocolli: I think we can remove this `TODO`.

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
TheMatten: Okay, will do.

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
TheMatten: Great :D

TheMatten:
@mauriciocolli 

> Though I don't mind if you want to make more visible, feel free to do it (of course, in a way that doesn't draw too much attention).

Just a simple background similar to YT or those new gesture controls to make it consistent.

> And the fact that the icons stay very shortly in screen (video is already loaded) makes the player seems snappy, though the "artifacts" (progress bar visible in a split second, for example) can be seen as a buggy player UI to the user.

Otherwise it's fine, I was just interested in "visibility". Frankly, any sort of long animation or effects would just get in the way of users while watching.

Oh, one more thing - #1603 (3. point) and #1199 - those raster images look terrible when scaling, but that's thing for separate PR.

TheMatten:
Okay, cleaned up history - commits changing gesture control are squashed into one big.

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
```diff
@@ -424,6 +440,20 @@ public void initViews(View rootView) {
             titleTextView.setSelected(true);
             channelTextView.setSelected(true);
 
+            getRootView().addOnLayoutChangeListener((view, l, t, r, b, ol, ot, or, ob) -> {
+                // Use smaller value to be consistent between screen orientations
```
mauriciocolli: Little thing that's bothering the micro-optimization part of me, we could return if the size itself didn't changed:
```java
if (left == oldLeft && right == oldRight) return;
```

TheMatten: What about `if (l != ol || t != ot || r != or || b != ob) { // do things... }` ?

TheMatten:
Added check to max length updates - I should be ready for merging.

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
mauriciocolli: It's fine, this piece of code is short enough to be a good a choice.

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
```diff
@@ -956,63 +993,90 @@ public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float d
...
+                float currentProgressPercent =
+                        (float) playerImpl.getBrightnessProgressBar().getProgress() / playerImpl.getMaxGestureLength();
+                WindowManager.LayoutParams layoutParams = getWindow().getAttributes();
+                layoutParams.screenBrightness = currentProgressPercent;
+                getWindow().setAttributes(layoutParams);
+
+                if (DEBUG) Log.d(TAG, "onScroll().brightnessControl, currentBrightness = " + currentProgressPercent);
+
+                if (currentProgressPercent < 0.25) {
+                    playerImpl.getBrightnessImageView().setImageDrawable(
```
mauriciocolli: I think we could improve readibility and be less redudant here (and above) introducing a variable:
```java
// We could initialize at the "else" value, but I think
// this way makes it more readable.
final int iconRes; 
if (currentProgressPercent < 0.25) {
    iconRes = R.drawable.ic_brightness_low_white_72dp;
} else if (currentProgressPercent < 0.75) {
    iconRes = R.drawable.ic_brightness_medium_white_72dp;
} else {
    iconRes = R.drawable.ic_brightness_high_white_72dp;
} 
```

What do you think?

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
```diff
@@ -424,6 +440,20 @@ public void initViews(View rootView) {
             titleTextView.setSelected(true);
             channelTextView.setSelected(true);
 
+            getRootView().addOnLayoutChangeListener((view, l, t, r, b, ol, ot, or, ob) -> {
+                if (l != ol || t != ot || r != or || b != ob) {
+                    // Use smaller value to be consistent between screen orientations
+                    // (and to make usage easier)
+                    int width = r - l, height = b - t;
+                    maxGestureLength = (int) ((width > height ? width : height) * MAX_GESTURE_LENGTH);
```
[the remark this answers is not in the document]
TheMatten: Doesn't it make more sense like this? We're doing update when something changes, not returning when values stayed the same...

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
TheMatten: Not a bad idea, will do.

TheMatten: Hmm, ternary operator is right-associative... :
```
playerImpl.getVolumeImageView().setImageDrawable(
        AppCompatResources.getDrawable(getApplicationContext(),
                currentProgressPercent <= 0
                        ? R.drawable.ic_volume_off_white_72dp
                        : currentProgressPercent < 0.25
                        ? R.drawable.ic_volume_mute_white_72dp
                        : currentProgressPercent < 0.75
                        ? R.drawable.ic_volume_down_white_72dp
                        : R.drawable.ic_volume_up_white_72dp
        )
);
```
Can I use this? :D

mauriciocolli: Definitely hard to read, but if you keep the condition result in the same line, yeah. A variable here can help it again to keep the line shorter and the style closer to what is used currently:

```java
final int resId = currentProgressPercent <= 0 ? R.drawable.ic_volume_off_white_72dp
        : currentProgressPercent < 0.25 ? R.drawable.ic_volume_mute_white_72dp
        : currentProgressPercent < 0.75 ? R.drawable.ic_volume_down_white_72dp
        : R.drawable.ic_volume_up_white_72dp;
```

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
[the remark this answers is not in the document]
mauriciocolli: Just throw a `Math.min(width, height)` in there, even easier to read.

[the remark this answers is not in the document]
TheMatten: Much better idea, thanks.

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
TheMatten: I just wanted to avoid additional variable... but okay, will do.

TheMatten:
Fixed listener and res selection.

TheMatten:
Great! Thanks for your help (and patience) :)

Issue #1531, opened 2018-07-14
Auto upgrade Newpipe


pranavg189:
- [x ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x ] I checked if the issue/feature exists in the latest version.
- [ x] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

It would be great if an Newpipe can auto update itself when a new version is released.

pranavg189:
I can volunteer writing code for this functionality If someone is able to given me pointers on how to get started on this.

alfasi:
This truly seems like a feature F-droid should contain instead of each individual app.
At the most NewPipe should notify when a new version is available, but not take the responsibility of upgrading itself IMO.

Grammost:
F-Droid is already able to auto-update your apps, when you flash the OTA package.

pranavg189:
I think atleast it should notify when a new version is available, because all users like me do not use F-Droid. F-Droid seems very slow to me. I can go ahead and implement this notify feature.

theScrabi:
@pranavg189 +1

crkellen:
To remedy this issue, I have a tab open on my phone's browser to the releases page, which I check semi-regularly.

It would be amazing if I could just get updates through the app. I also don't like F-droid due to it taking weeks or more sometimes to update an app.

krtkush:
I'm working on this as mentioned in #1520. Should have a working version by the end of next weekend.

stonerl:
Some of you apparently don't know how F-Droid works. All versions of Newpipe where available within 24 hours after release. If a maintainer doesn't properly tag releases, F-Droid can't properly provide the latest versions of an app.

Back on topic. The release notifications should be opt-in and/or only available in the version available from the GitHub-Downloadpage and not the one from F-Droid IMHO.

stonerl:
Btw. GitHub provides a newsfeef for releases. For Newpipe you just have to add this URL to the RSS-Reader of you liking:

https://github.com/TeamNewPipe/NewPipe/releases.atom

krtkush:
> The release notifications should be opt-in..

Agreed.

[...]

Issue #1520, opened 2018-07-02
Screen in the app to show and download APKs from github releases.


krtkush:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.
- [x] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

**Feature request/ idea**

Currently, F-Droid takes too long to update its repo with the latest version of the app. Also, there must be users who don't have/ want F-Droid installed but still want to track the latest version of NewPipe. 

For these users and to counter the slowness of F-Droid APK updates, I suggest that we have an inbuilt system in the app which shows and notifies the availability of the latest APKs. We can use the [Releases API](https://developer.github.com/v3/repos/releases/) to get the latest info from github itself. 

I'm willing to take up this task if the general idea is approved.

theScrabi:
Yum I'm at your side here :)

krtkush:
Great! I'll start working on it, then.

TobiGr:
I am not sure which information you want to get, but you can use the api we created for the website to get the current version: [TeamNewPipe/web-api](https://github.com/TeamNewPipe/web-api)
You can fetch the data from https://newpipe.schabi.org/api/data.json

krtkush:
@TobiGr  The API you mentioned does not provide link to download the APK. 

My idea is to provide a functionality to notify the user about a new APK availability and then let the user download it from within the NewPipe app. I, personally, would like this way of updating the app because I find that F-Froid is not quick enough to make the latest APK available for download. 

`https://api.github.com/repos/TeamNewPipe/NewPipe/releases` gives the `assets_url` which further provides `browser_download_url`. This url can be used to download the APK of that release.

TobiGr:
It's no problem to include it into the web-api. We'd only need to make a difference between F-Droid and GitHub releases. @TheAssassin What do you think about this?

krtkush:
> It's no problem to include it into the web-api.

That sounds good.

TheAssassin:
@TobiGr this should be doable, although I'm a bit worried about the GitHub rate limit. I think we should use some cheap web scraping to get the information we need.

TobiGr:
[...]

PR #1607, opened 2018-08-16
app update notification

krtkush:
- [*] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Added feature described in #1520 and #1531.

Currently it is a very basic version. Will further build on this feature in the future.


app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
```diff
@@ -663,6 +664,9 @@ public void onPlaying() {
             lockManager.acquireWifiAndCpu();
 
             hideControls(DEFAULT_CONTROLS_DURATION, DEFAULT_CONTROLS_HIDE_TIME);
+
+            // Check for new version
+            new CheckForNewAppVersionTask().execute();
```
TobiGr: Why do you check here?

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
```diff
@@ -0,0 +1,162 @@
...
+                        stringBuilder.append(line);
+                        stringBuilder.append("\n");
+                    }
+
+                    bufferedReader.close();
+                    response = stringBuilder.toString();
+
+                    return response;
+            }
+        } catch (MalformedURLException ex) {
```
TobiGr: Why do you make a difference between `MalformedURLException` and `IOException` when you execute the same method? I think you can shorten this a little:
``` Java
} catch (MalformedURLException | IOException ex) {
```

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+
+        return null;
+    }
+
+    @Override
+    protected void onPostExecute(String response) {
+
+        Log.i("Response--", response);
```
TobiGr: This entry does not help us to find issues in the future. Please add some additional information or  remove this log entry.

app/src/main/res/values/strings.xml
```diff
@@ -504,4 +508,12 @@
         <item>144p</item>
     </string-array>
 
+    <!-- App update notification -->
+    <string name="app_update_notification_content_title">NewPipe Update Available</string>
+    <string name="app_update_notification_content_text">Tap to download</string>
+
+    <!-- Flavor names -->
+    <string name="app_flavor_github">github</string>
```
TobiGr: I would not place the flavor names here. Making them constants in `App.java` seems to be a good idea.

TobiGr:
You closed this while I wrote the review. 😆 What's the problem? Can we help you?

PR #1608, opened 2018-08-16
App update notification

krtkush:

Added feature described in #1520 and #1531.

Currently it is a very basic version. Will further build on this feature in the future.


app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
```diff
@@ -663,6 +664,9 @@ public void onPlaying() {
             lockManager.acquireWifiAndCpu();
 
             hideControls(DEFAULT_CONTROLS_DURATION, DEFAULT_CONTROLS_HIDE_TIME);
+
+            // Check for new version
+            new CheckForNewAppVersionTask().execute();
```
TobiGr: Why do you run the check here, too?

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+
+        return null;
+    }
+
+    @Override
+    protected void onPostExecute(String response) {
+
+        Log.i("Response--", response);
```
TobiGr: This entry does not help us to find issues in the future. Please add some additional information or remove this log entry.

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
```diff
@@ -0,0 +1,162 @@
...
+                    }
+
+                    bufferedReader.close();
+                    response = stringBuilder.toString();
+
+                    return response;
+            }
+        } catch (MalformedURLException ex) {
+            ex.printStackTrace();
+        } catch (IOException ex) {
```
TobiGr: Why do you make a difference between `MalformedURLException` and `IOException` when you execute the same method? I think you can shorten this a little:
``` Java
} catch (MalformedURLException | IOException ex) {
```

theScrabi:
> My biggest concern about these changes are the additional build flavors. When someone installs NewPipe from F-Droid, he/she will always be forced to use this source to update the application.

Due to different signatures you already have to reisntall the app if you want to use the fdroid version.

> Additionally, with this new config every user needs to uninstall the current version of the app.

How comes? As long as the package name and signing key does not change nothing needs to be re installed.

> Enable / Disable search for new updates

I agree. Although it should be enabled by default.

TobiGr:
> Due to different signatures you already have to reisntall the app if you want to use the fdroid version.

That's true, but I'd like to get back to one version.

> How comes? As long as the package name and signing key does not change nothing needs to be re installed.

``` gradle
flavorDimensions "apkSource"
    productFlavors {
        github {
            dimension "apkSource"
            applicationIdSuffix ".github"

        }

        fdroid {
            dimension "apkSource"
            applicationIdSuffix ".fdroid"
        }
}
```

This means, we now have four different builds:
`githubRelease`
`githubDebug`
`fdroidRelease`
`fdroidDebug`

None of them matches our release build because these new have suffixes. This means an upgrade is not possible.

TobiGr:
The current implementation uses the new build flavors to decide whether to check for an update. As mentioned above, there are other ways to do this (e.g. the signature). So if this is implemented in a different way, we won't need a suffix.

krtkush:
Hey @TobiGr, @theScrabi 

I agree - A single version will be a better approach; I'll look into replacing the flavor with signature check.

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
krtkush: This was not supposed to go into production. I'll remove it.

app/src/main/java/org/schabi/newpipe/player/PopupVideoPlayer.java
krtkush: From what I remember the check was not being performed in the pop-up mode. Hence, I had to put it here. I'll check once again and make changes if needed.

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
krtkush: Aah..

That's just a habit of mine. I'll change it to the suggested one.

theScrabi:
@krtkush @TobiGr so here are some updates. After the youtube amageddon that made us release v0.14.1 a few days ago, I had a conversation with the fdroid people about update speed.

It seems to be a complicated topic which seem to not get a widespread acceptance within their team. Therefore we should try and even deliver updates ourself even if its a version deploid by fdroid.

This on one hand makes things easier for us since we don't have to deploy two separate versions, and on the other hand we need to make NewPipe a reproducable build within fdroid.

@krtkush could you please try to fix the conflicts, and @TobiGr would you please review it after the changes are done.

I am willing to put this feature into the next version if it's possible :)

krtkush:
@theScrabi 
Sure! I'll work on it this weekend. 

Just to be clear, we'll be going with the signature check method as suggested by @TobiGr, right? And not with different builds.

theScrabi:
Yes the signature check method.

krtkush:
@TobiGr @theScrabi Does NewPipe use different KeyStores for the github and f-droid versions respectively? 

I'll need the developer certificate signature of the KeyStore(s), as instructed in the [linked article](https://www.airpair.com/android/posts/adding-tampering-detection-to-your-android-app). Could either of you provide that?

krtkush:
Also, after syncing with upstream, Travis is giving problems - "Task 'lintDebug' not found in root project NewPipe".

TobiGr:
@krtkush Sorry for the silence. We are all quite busy. If I am correct, you can find the fingerprint on @theScrabi's website: https://schabi.org If you need any further information about the signing key, please ask him :)

theScrabi:
The second key is the one of the fdroid version. You can see the cert by downloading the apk, and extracting it. They key is:
`SHA1: 83:10:87:55:C1:3C:C6:D5:5D:46:86:53:C8:2F:CA:9E:25:46:8A:C9`

The issue is, I don't know where you can find the fingerprint for the verification. The only thing I can tell you is, that fdroid is signing the apk with their own key. Therefore you might find this fingerprint on their website.

krtkush:
You're right. I'm having trouble finding the fingerprint.

However, [this](https://stackoverflow.com/questions/9293019/get-certificate-fingerprint-from-android-app#22506133) approach should work. The accepted answer rebuilds the SHA1 key at runtime. Hence, we can confirm that if the retrieved SHA1 key is equal to `B0:2E:90:7C:1C:D6:FC:57:C3:35:F0:88:D0:8F:50:5F:94:E4:D2:15` then it is a GitHub build and therefore we can enable the version update check.

theScrabi:
:D nice solution.

krtkush:
Thanks!

All done from my side.

krtkush:
I made a couple of more changes. Everything seems to be working and I think the feature is ready for a production release; we can add more functionality to it later.

theScrabi:
:D nice i wall qu it soon

app/src/main/res/values/strings.xml
```diff
@@ -511,6 +516,10 @@
         <item>144p</item>
     </string-array>
 
+    <!-- Updates Settings -->
+    <string name="updates_setting_title">Updates</string>
+    <string name="updates_setting_description"> Show a notification to prompt app update when a new version is available</string>
```
TobiGr:
```suggestion
    <string name="updates_setting_description">Show a notification to prompt app update when a new version is available</string>
```

TobiGr:
>:D nice i wall qu it soon

Or even better @theScrabi can do this. I didn't see your comment. GitHub had some problems today 😄

theScrabi:
I don't find time recently, but I'll try my best. After next release this should be highest prio.

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
```diff
@@ -0,0 +1,247 @@
...
+        } catch (PackageManager.NameNotFoundException e) {
+            e.printStackTrace();
+        }
+
+        Signature[] signatures = packageInfo.signatures;
+        byte[] cert = signatures[0].toByteArray();
+        InputStream input = new ByteArrayInputStream(cert);
+        CertificateFactory cf = null;
+
+        try {
```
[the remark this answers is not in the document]
krtkush: Fixed.

krtkush:
Yeah GitHub was down today for a fair bit of time.

app/src/main/java/org/schabi/newpipe/settings/UpdateSettingsFragment.java
```diff
@@ -0,0 +1,36 @@
...
+        findPreference(updateToggleKey).setOnPreferenceChangeListener(updatePreferenceChange);
+    }
+
+    @Override
+    public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
+        addPreferencesFromResource(R.xml.update_settings);
+    }
+
+    private Preference.OnPreferenceChangeListener updatePreferenceChange
+            = new Preference.OnPreferenceChangeListener() {
```
theScrabi: You can use a lambda here.

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
```diff
@@ -0,0 +1,242 @@
...
+import android.content.SharedPreferences;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.content.pm.Signature;
+import android.net.Uri;
+import android.os.AsyncTask;
+import android.preference.PreferenceManager;
+import android.support.v4.app.NotificationCompat;
+import android.support.v4.app.NotificationManagerCompat;
+import android.util.Log;
```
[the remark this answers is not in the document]
krtkush: `import android.util.Log;` is unused. Have removed it. 
But, the other three are needed and being used.

app/src/main/java/org/schabi/newpipe/settings/UpdateSettingsFragment.java
krtkush: Hey, I'm not very familiar with lambda. Could you please suggest how to do I implement that? Or any resource which can?

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
```diff
@@ -0,0 +1,242 @@
...
+                JSONObject githubStableObject = githubObject.getJSONObject("stable");
+
+                String versionName = githubStableObject.getString("version");
+                String versionCode = githubStableObject.getString("version_code");
+                String apkLocationUrl = githubStableObject.getString("apk");
+
+                compareAppVersionAndShowNotification(versionName, apkLocationUrl, versionCode);
+
+            } catch (JSONException ex) {
+                ex.printStackTrace();
```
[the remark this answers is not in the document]
krtkush: I have used ErrorActivity.reportError() in every case as I found it more appropriate against `CheckForNewAppVersionTask`.

app/src/main/java/org/schabi/newpipe/settings/UpdateSettingsFragment.java
theScrabi: Instead of writing:
```
setListener(new Listener() {
   void theEvent(int parameter) {
      ...
   }
```
In java 7+ you can simply write:
```
setListener( (int parameter) -> {
   ...
}
```

Its shorter and simpler to read.

krtkush:
Done! Have made all the changes as requested.

theScrabi:
Please resolve the conflict.

krtkush:
Resolved.

TobiGr:
I removed my blocking because I have no time review the latest changes. Just skimmed the code and saw some vectors. Did you test this with KitKat?

krtkush:
@TobiGr Just did. It's failing on kitkat. Will replace it with a png.

krtkush:
Done. I really hope it is actually done this time 😛

TobiGr:
@krtkush Looks good to merge.
@theScrabi Can you please remove your block?

theScrabi:
Done.

theScrabi:
I wanted to take care about pending features once my semester is over. However the sudden release came because of the full shutdown on Thursday.

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
```diff
@@ -0,0 +1,230 @@
...
+    protected void onPreExecute() {
+
+        mPrefs = PreferenceManager.getDefaultSharedPreferences(app);
+
+        // Check if user has enabled/ disabled update checking
+        // and if the current apk is a github one or not.
+        if (!mPrefs.getBoolean(app.getString(R.string.update_app_key), true)
+                || !isGithubApk()) {
+            this.cancel(true);
+        }
```
snaik20, line 66: Checking this before starting the task would be more efficient, at App#L108

Stypox: This is 4 years old code, is this still the case?

snaik20: Hi @Stypox,

I apologize for my previous comment. I didn't realize that the code was 4 years old, and I see that it has now been deleted. I was going through the code to explore and understand the codebase, but I should have been more careful about commenting on old code.

I'll be more mindful of the age of code in the future and avoid commenting on old code unless I'm sure it's still relevant.

Thanks for your patience and understanding.

PR #1620, opened 2018-08-22
Playback resume

szabolcs-dekany92:
Hey!
First things first,

This is my first android and open source project that I've ever worked on. I implement a video playback resume feature that I've been using on my own. I would like to ask some questions regarding my developed feature and give some feedback. Any feedback regarding android development and open source in general is helpful.

### The motivation behind the feature
I think it would make a great addition to the project since something very similar is already implemented in Youtube cross platform. I developed this for my own use at first, but this could be a really handy feature for every user.

### The current implementation.

- The feature currently resides in the classes MainVideoPlayer and PopupVideoPlayer respectively. Wouldn't it be just easier to move it to the abstract class VideoPlayer so that every implementing video player gets this feature? I wasn't sure where to put it.

- The option the turn this feature on and off currently can be found in Settings > History. Is this an appropriate place for it?

- I'm currently using SharedPreferences to store video URL/playback time pairs. These pairs have to own preferences and I'm not using the default one that is given back by PreferenceManager. Is there a better way to store this info?

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -163,6 +163,7 @@ public class VideoDetailFragment
     private TextView detailControlsDownload;
     private TextView appendControlsDetail;
     private TextView detailDurationView;
+    private TextView detailPlaybackResumeView;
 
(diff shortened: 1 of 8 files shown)


surya-ravishankaran:
I ve been wanting this feature for a long time, would love to see it implemented. For longer videos it's really hard to remember where you left off all the time

PR #1628, opened 2018-08-26
Basic D-Pad support in player

Koitharu:
Support D-Pad and some TV-specified buttons in player

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -49,12 +49,13 @@ ext {
     icepickLibVersion = '3.2.0'
     stethoLibVersion = '1.5.0'
 }
+
 dependencies {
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/schabi/newpipe/fragments/KeyPressable.java
```diff
@@ -0,0 +1,15 @@
+package org.schabi.newpipe.fragments;
+
+import android.view.KeyEvent;
+
+/**
+ * Indicates that the current fragment can handle key presses
+ */
+public interface KeyPressable {
```
theScrabi: Maybe its better to put this into `org.schabi.newpipe.utils`.

PR #1637, opened 2018-08-28
Fix small spelling error

johnnyw3:

This is just a small spelling error - "Chose" should be "Choose" to maintain proper tense.


theScrabi:
It seems another patch fixed this already. 

@johnnyw3 thanks for the help though :)

PR #1644, opened 2018-08-31
Separate options for volume and brightness gestures

### LINKED ISSUE #1582: [Feature Request] Separate switches for gesture controls
- [✔️ ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [✔️ ] I checked if the issue/feature exists in the latest version.
- [✔️ ] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

This is it, I just want the option to be able to activate the volume control gesture and leave the brightness control deactivated, because sometimes I accidentally change the brightness when I just wanted to adjust the volume ... and I prefer to adjust the brightness with the default control of android. Thank you

ritiek:

Closes #1582.

This PR separates "Player gesture controls" options into two separate options; "Volume gesture control" and "Brightness gesture control".

One thing to note is that if only one of these options is turned on; the gesture will work on any area of the screen. I did this purposely as I prefer this instead of only one side of the screen being able to control brightness/volume while the other side does nothing. However, I can do in in the latter way if you guys want.

(Also, this is my first time contributing to an android app. Please be brutal :)


ritiek:
We need to also add translations for these options in other languages.

theScrabi:
@ritiek thank you for your submission, however I think this is a solution to a very specific problem. I try to avoid adding specific solutions.

TobiGr:
@theScrabi I think these changes might be helpful for users with small screens. I often wanted to only disable the volume gesture because I accidentally changed the volume. In my opinion we should reject this PR.
@karyogamy @mauriciocolli What do you think?

theScrabi:
I see, if there is a demand, we can talk about it. Would you mind to do QA on it :)

TobiGr:
I'll review and test it on Friday.:+1:

ritiek:
> But next time, please rebase your commits to keep in sync with our dev branch.

Thanks! I'll keep that in mind.

By the way, we still need to provide translations for [descriptions of these new options](https://github.com/TeamNewPipe/NewPipe/blob/2d17cfaf2ba0795771546bf33533745399c38935/app/src/main/res/values/strings.xml#L86-L89) in other languages.

PR #1648, opened 2018-09-02
Use device language for YouTube (fix #1130)

### LINKED ISSUE #1130: Do not use IP to select app's language

Currently it seems NewPipe uses the IP to determine the language of the app. This is pretty sub-optimal for any user that uses an out-of-country VPN, or for users who wish to use another language (I set all my applications to English, for example). It's also harder to debug translations thi way. Could it be made so that it either has a language selector, or selects the language depending on the system locale?

nv95:

Fix issue #1130
44953357-14fca400-ae9d-11e8-8c60-15471f14a82f

Use new extractor api from [this PR](https://github.com/TeamNewPipe/NewPipeExtractor/pull/104)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ExtractorHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ExtractorHelper.java
@@ -46,6 +46,7 @@ import org.schabi.newpipe.report.UserAction;
 import java.io.IOException;
 import java.io.InterruptedIOException;
 import java.util.List;
+import java.util.Locale;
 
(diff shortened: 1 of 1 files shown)


theScrabi:
We need to be careful with such a fix, since NewPipe extractor currently does not make a difference with country and language. The Language should be the one of the device, true. But the country should be the one the user wants to select.

theScrabi:
Please wait with localisation changes until the localisation of the extractor got refactored. This is because we want to split up the country/language part. Thus we can not say how localisation will look like later in the frontend.

PR #1651, opened 2018-09-03
Spelling: Language rework

comradekingu:


theScrabi:
@comradekingu you may want to check the Kiosk Page, Feed Page etc. strings, since they had a collision with changes on dev.

PR #1748, opened 2018-09-22
store and reload the last used aspect ratio

### LINKED ISSUE #1318: [Feature Request] Set Default Aspect Ratio
This would be useful for people that multiwindow as the fill aspect ratio gives a fullscreen view while in landscape splitscreen and having to set it manually for every video can be such a drag

MaX-Lo:
The last used aspect ratio is now stored in SharedPreferences and gets reloaded when resuming the MainVideoPlayer Activity (similar to storing/reloading the last used screen rotation). 

closes #1318 


theScrabi:
@TobiGr or @karyogamy could you please test this and check if its ok?

Issue #1631, opened 2018-08-27
Add autoplay video at the end of the current video


ReekyMarko:
Right now, the autoplay feature adds the next video as soon as the current one starts playing. This is less than ideal when you want to f.ex add songs to be played in the background one by one using the queuing feature. When I add the first song, the autoplay feature immediately adds the next one. If I want to add another song to be played next, I have to queue that and go to the player and remove the song added by autoplay.

So here is my suggestion: add the autoplay video only when there is no videos coming up next, and preferably towards the end of the video (I'm thinking 5 seconds before the current one ends)

theScrabi:
Sounds like a good idea :)

ReekyMarko:
Forgot this was open. Recent versions have introduced a feature similar to this.

PR #1763, opened 2018-09-26
Move the auto queue from the beginning of a stream to the end of the playlist

NigelTicknor:

This pull request is in reference to #1631 

I tested on regular, background, and popup players.


TobiGr:
Thanks for keeping your forked in sync with the changes on our main branch although you didn't hear from us for quite some time 👍 
As @theScrabi said, code looks fine and your changes work as expected. But there is one thing I'd like to note:
When there is a stream in the queue and the current stream comes to an end, we already fetch the beginning of the next stream. This allows us to continue without an interruption/break.
With your current changes this pre-loading is not possible any more. I am not capable of the code which triggers the pre-loading, but it might worth to take a look at it and adjust it to work with the change done in this PR.
@theScrabi What do you think?

theScrabi:
Well i liked the fact that the video just continues, but i also disliked that any  audio playlist you click together has some unwanted title in the end.

I guess a little pause is aceptable when switching videos/next title.

TobiGr:
Closed in favour of #1878 because this approach enables us to pre-load next streams, too. 
Thanks for your work :)

PR #1773, opened 2018-09-30
[Closes #1772] change 'tab_bookmarks' string to 'Bookmarked Playlists' for UI clarity

### LINKED ISSUE #1772: [Interface Enhacement] Change "Bookmarks" to "Bookmarked Playlists"
# Summary
It seems that within NewPipe user interface, both "Playlists" and "Bookmarks" refer to the same feature in NewPipe.  On the home screen activity of NewPipe, the third "tab" displays "Bookmarks".  However, throughout the rest of the NewPipe UI, most of the "Bookmark" functionality is referred to by "Playlist".

# Why I am filing this bug
When I realized that you could change the content of the home activity, I started investigating what I could add as a tab to the NewPipe home activity.  I saw Bookmarks and was puzzled.  When I went to the Bookmarks tab, I saw "Nothing here but crickets".  "Well, I guess I have to bookmark something.  Let me find out how".

So then I'm looking on different videos in "Detail" view for a "Bookmark" option somewhere.  I couldn't find one - so I started searching the internet.  I came across some NewPipe github issues that gave some context and enlightened me that "Bookmarks" are just saved playlists.  I hadn't used the saved playlist feature much - so it never occurred to me that these were the same.

Some users will probably use the "Save to playlist..." feature and realize this idiosyncrasy.  However, I think NewPipe could improve this aspect of the UI by choosing to use one of these terms.  "Playlist" seems to make more sense to me and would require less changing - but I also realize that user interface design can be *very* subjective.

# Details
NewPipe's strings.xml gives an excellent breakdown of this situation (I don't know the GitHub magic to add a reference, but we all know where to find it).  There are more than a dozen references to "Playlist" here.  However, there are only these three references to "Bookmark":

```
<string name="tab_bookmarks">Bookmarks</string>
<string name="bookmark_playlist">Bookmark Playlist</string>
<string name="unbookmark_playlist">Remove Bookmark</string>
```
I'm sure these could be changed to something using the "Playlist" word - something like "Bookmarks", "Save Playlist", or "Remove Playlist" might be more consistent

edit:  in speaking with comradekingu in IRC, simply changing "tab_bookmarks" string to "Bookmarked Playlists" might be a more straightforward way of resolving this that involves less strings being changed

# Final Thoughts
I realize that a decent amount of effort needs to go into making a change like this - I'm not sure if this applies to other translations.  However, I do think that in English this is partially confusing and think there is room for improvement in this area.  I just wanted to give another thanks to the NewPipe team and all of the hard work that you all do on this excellent project.  

# Issue Checklist
- [ ] (N/A) I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

prg318:
This change makes the heading of the "Bookmarks" tab more intuitive to the user by renaming it to "Bookmarked Playlists".  This addresses https://github.com/TeamNewPipe/NewPipe/issues/1772

Here is an example image of what this change looks like on an emulated Pixel XL:
46259248-035eea00-c4a5-11e8-9c17-25e8649e3e91

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -34,7 +34,7 @@
 
     <string name="tab_main">Main</string>
     <string name="tab_subscriptions">Subscriptions</string>
-    <string name="tab_bookmarks">Bookmarks</string>
+    <string name="tab_bookmarks">Bookmarked Playlists</string>
     <string name="tab_new">New Tab</string>
     <string name="tab_choose">Choose Tab</string>
 


theScrabi:
Sounds good :)

PR #1797, opened 2018-10-08
Music player control

elibroftw:
# Why?
I like the Notification UI in Oreo for music players and the UI for NewPipe always bothered me. The color scheme and layout hurts me a bit everytime I play something in the background. 
# Changes

- gitignore: Android Studio made these files
- buildToolsVersion: "The specified Android SDK Build Tools version (27.0.3) is ignored, as it is below the minimum supported version (28.0.2) for Android Gradle Plugin 3.2.0." I want to PR to target SDK 28 later
- BackgroundPlayerActivity.java: Android Studio said that the code was redundant and could be simplified
- notification XML files: 
  - I moved the stop button beside other buttons
  - The cover is now on the right side just like in Android Oreo+.
  - Aligned the buttons to the left (I think Oreo does this)
  - Progress bar got shortened for small View.
  - Small View is taller (similar to notifications on Oreo, but I'm not sure if they are the exact same height, I eyeballed it)
  - Notification color is same as the dark theme background color (NOTE: I did not make it dependant on theme setting)

# Testing
- Oneplus 3 Android 8.0 (app works with Background Playback)
- Emulators for Pixel XL and Pixel XL 2 (API 27) (app works with Background Playback)
- Eemulator for Nexus (4 API 19) (app works with Background Playback but phone doesn't work if power button is pressed. I'm not sure if that is the fault of the Emulator, my changes, or if that existed on the dev branch)


elibroftw:
I think Travis failed because of different build tools version

elibroftw:
In small View,  the cover does not always stay on the right side. I tried to fix this issue but didn't understand why this issue was happening.

PR #1810, opened 2018-10-13
Fix broken bold

saderror256:
Apologies for these small readme commits, but for some reason the \*\*bold effect\*\* didn't work, so i just changed it to \<b\> instead.
Also removed the multiple "\\" that was deprecated in html5 because well why not, I need something to do :+1: 

- [ yes ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.


TobiGr:
Thanks!

PR #1830, opened 2018-10-20
Add foreground service permission for API 28+ devices

boredomdenied:
As discussed here https://developer.android.com/about/versions/pie/android-9.0-changes-28
Apps which target API 28+ must include foreground permission in the manifest. I first noted this issue today after trying NewPipe master on the Pixel 3 device. Confirmed exists also with API 28 in emulator.

 - Fix foreground playback in android Pie API 28. Tested in emulator & Pixel 3 device.

- [ x ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -9,6 +9,8 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+    
 
     <application
         android:name=".App"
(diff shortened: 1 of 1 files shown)


TobiGr:
> I first noted this issue today after trying NewPipe master on the Pixel 3 device.

@boredomdenied I assume the app crashed. Is that correct?

@theScrabi If this is the case, we should consider to merge this before we release v0.14.2.

theScrabi:
I will merge it.

boredomdenied:
Apologies for the late reply. Yes. It crashed the app. There are still issues when newpipe is put into the background. When trying to bring it back into view an app crash is common. This may be an issue with the pixel 3 aggressive memory and not something for NewPipe to work out. I am a bit too busy to investigate the issue for now

switchtegrax1:
The crash on the main player happens on all pie devices,for now a workaround has been applied to the latest test version of 0.14.2 but this workaround will make the video restart instead of staying where it was left off

PR #1835, opened 2018-10-22
Fix crash with default resolution set to best and limited mobile data resolution

### LINKED ISSUE #1829: Set "Limit resolution when using mobile data" video CRASH
When i set "**Default resolution**" to "**Best resolution**" and "**Limit resolution when using mobile data**" to **480p**(or other XXXp), **turn on mobile data** and try playing video my newpipe always crash can't open video, when i turn OFF mobile data and turn ON WI-FI all is good. 
Can someone else confirm this? 
If needed logs or anything else tell me how to do it.

ritiek:

Fixes #1829.

When default resolution is set to best. The value of `resolution` is set to `best_resolution` here:
https://github.com/TeamNewPipe/NewPipe/blob/6fe3fdce112a617ed94d6b113146df8747ad74bb/app/src/main/java/org/schabi/newpipe/util/ListHelper.java#L108-L116

In the `if` branch however, `compareVideoStreamResolution` function expects values like `144p`, `480p`, etc. (but instead, the string `best_resolution` gets passed in `resolution`) and it is unable to parse it to an integer:
https://github.com/TeamNewPipe/NewPipe/blob/6fe3fdce112a617ed94d6b113146df8747ad74bb/app/src/main/java/org/schabi/newpipe/util/ListHelper.java#L393-L399

This PR workarounds it by checking whether the value of `resolution` is set to `best_resolution` before calling `compareVideoStreamResolution` in the `if` condition.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
@@ -110,7 +110,8 @@ public final class ListHelper {
                 : context.getString(R.string.best_resolution_key);
 
         String maxResolution = getResolutionLimit(context);
-        if (maxResolution != null && compareVideoStreamResolution(maxResolution, resolution) < 1){
+        if (maxResolution != null && (resolution.equals(context.getString(R.string.best_resolution_key))
+                || compareVideoStreamResolution(maxResolution, resolution) < 1)) {
             resolution = maxResolution;
         }
         return resolution;


epitron:
Thank you! I've been unable to use NewPipe on mobile for quite some time because of this! (Easy workaround as well.)

ritiek:
Let me know if this looks good, I'll rebase it.

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -110,7 +110,8 @@ private static String computeDefaultResolution(Context context, int key, int val
                 : context.getString(R.string.best_resolution_key);
 
         String maxResolution = getResolutionLimit(context);
-        if (maxResolution != null && compareVideoStreamResolution(maxResolution, resolution) < 1){
+        if (maxResolution != null && (resolution.equals("best_resolution")
```
TobiGr: Can you please show me, where the "best_resolution" comes from?
I do not have time to look it up and you already managed to find it once :)

ritiek:
```java
System.out.println(preferences.getString(context.getString(R.string.default_resolution_key),
                                         context.getString(R.string.default_resolution_value)));
```

This line prints out `best_resolution` when Default resolution is set to Best resolution in settings.

Same stuff is happening here:
https://github.com/TeamNewPipe/NewPipe/blob/bb00a56de671d5d64b6f3884998b03410746ca2f/app/src/main/java/org/schabi/newpipe/util/ListHelper.java#L109

The `best_resolution` string seems to be arising from here:
https://github.com/TeamNewPipe/NewPipe/blob/fba0a8036bdfe046d2e10d3aa7011fc08592cb25/app/src/main/res/values/settings_keys.xml#L52

theScrabi: Could be that i introduced this line, brcause sometimes the desired resolution is not available, and on different services not even possible. So you should select a resolution closest to the desired one.

Does that sound logic?

TobiGr: > Does that sound logic?

Yes. 

What I wanted to note is that using `"best_resolution"` as String is not a good idea. Please use `context.getString(R.string.best_resolution_key)` instead. Otherwise, this bug might occur when we change the value of `R.string.best_resolution_key`.

ritiek: Yep, that's much better!

PR #1876, opened 2018-11-10
Missed strings

bboa:
Hi guys!
I want said that partial translation looks awful.

HostedWeblate does not contains this strings to translate.
Please add this strings to Hosted Weblate template or merge PR, at your choice.

### DIFF
--- a/app/src/main/res/values-ru/strings.xml
+++ b/app/src/main/res/values-ru/strings.xml
@@ -510,7 +510,7 @@
 
     <string name="brightness_gesture_control_summary">Менять жестом яркость в плеере</string>
     <string name="brightness_gesture_control_title">Жест яркости</string>
-    <string name="download_to_sdcard_error_message">Загрузка на внешний накопитель невозможна. Сбросить местоположение папки для загрузки?</string>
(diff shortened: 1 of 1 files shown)


theScrabi:
Next time please submit something like this using weblate.

bboa:
Hi @theScrabi,
thx for merging..

Probably I need explain that, as not a project owner on Hosted Weblate, I don't have the permissions to _add new_ lines into English strings.xml file, I can only change them. So we (translators) can't see strings to translate, because, even if I upload English strings.xml from Github to Hosted Weblate, this will not add new strings to English and all other languages. Something like this: (22 not found) https://imgur.com/a/UUXXeVX

We simply need your little help to check and setup autoupload of a source file (English strings.xml) from dev branch to Hosted Weblate if it's possible (with adding new strings, currently new strings are dropped). As new strings will be added automatically, translators will have possibility to support translations in complete state, and, as I hope, this may increase app popularity and donations to project. Currently, about all translations are incomplete, so this very destroy app usability and impressions when non-English UI is used.

theScrabi:
I see, although I don't have admin access on weblate. Hance i think if you introduced new strings as pr its ok then :)

PR #1650, opened 2018-09-03
comments support

yausername:
The code is crappy and needs lot of work.

68747470733a2f2f692e696d6775722e636f6d2f62497a784b73702e6a7067

PR #1878, opened 2018-11-11
issue:1336 Fix for inserting new streams when auto queuing is enabled

### LINKED ISSUE #1336: Autonext issue
Please make disapear the autonext video when we enqueue a video after the first one

shivanju:

What I did:
1. When auto queue is enabled and a stream is manually inserted, it shouldn't be appended after the auto queued stream. 
2.  Moving a auto queued stream in playlist will make it a manually inserted song for subsequent insertions.

Fixes #1336


TobiGr:
I quickly tested your changes. This is a nice improvement to the usability / UX of the background player (the others, too) 👍 
Is there any reason why you move the auto-queued stream instead of removing it?

shivanju:
Sometimes a user might like what auto-queued has suggested. And it can be removed if user doesn't like it. Personally I like to use autoqueue alot it's just I want manually queued streams to be played first.

TobiGr:
Yes. I'd also like to use the auto-queue feature. I visualized the behaviour you implemented:

queuing_behaviour_old

I think this approach might be the better one because it follows the streams the user preferred to listen next. @theScrabi what do you think?

queuing_behaviour_new

theScrabi:
how about diabling auto equeing compleatly when you start playint a playlist?

shivanju:
@TobiGr I agree that auto-queued stream should be based on the last stream. And essentially it means that we just need to remove the auto queued stream if a stream is added manually. @theScrabi As for the playlists, disabling auto-queue makes sense. Let me know if we want to go this way, so that I'll start modifying the code. And we are going to disable auto-queue for playlists, any suggestions how should I implement it just for playlist.

TobiGr:
>  I agree that auto-queued stream should be based on the last stream. And essentially it means that we just need to remove the auto queued stream if a stream is added manually.

Correct.

@theScrabi The diagrams I added above apply to the "normal" workflow and not playlists.

> how about diabling auto equeing compleatly when you start playing a playlist?

Good idea!

theScrabi:
I see. Yes ehat you drew makes sense :)

shivanju:
I'll start working on the changes then. I'll try to update the pr before next Monday.

shivanju:
Hi @TobiGr  @theScrabi 
I've updated the auto queue logic to what we discussed earlier i.e. if a new stream gets appended manually, current auto queued stream should be removed. Though I didn't make changes for disabling auto-queue when a playlist is being played. 
As I won't be getting much time for next 2 weeks. What I can think right now is that  I'll create a new issue for that change and work on it separately as this auto-queue change is required for https://github.com/TeamNewPipe/NewPipe/pull/1650 i guess.

theScrabi:
Test/QA pending.

TobiGr:
@shivanju I just noticed a bug. When auto enqueue is enabled and you remove the last auto enqueued stream, it is enqueued right just after it was removed. Can you please take a look at it and open another PR which fixes this behaviour?

TobiGr:
Or is this behaviour wanted?

shivanju:
@TobiGr  I think you are asking for when the stream that is being played is the second last one and the next stream is autoqueued one. In this case if a user deletes the autoqueued song, the deletion itself will trigger the addition of new autoqueued stream since the stream being played will become the last after deletion of autoqueued stream. And that is the expected behavior if autoqueue is ON. 
What's interesting here is that for a particular stream, it's auto queued stream will always be the same. I think it would be better if we choose a random stream from the metadata of currently playing last stream for autoqueued stream instead of always picking up the first one.

theScrabi:
@TobiGr please remember to add a PR to the release nodes after merging:
https://github.com/TeamNewPipe/NewPipe/releases

PR #1884, opened 2018-11-13
Remove support for android 4.1 - 4.3

theScrabi:
This PR will:
- Deprecate Android 4.1, 4.2, 4.3 since these are pourly supported and not tested anymore
- Remove deprecated "Old Player"
- Move "autoplay next" preference element into the player "Behaviour" section.


Brimont:
I have phone with android 4.2.1 i can use newpipe on future? This is my only phone!

gkeegan:
@Brimont You can still use any versions that have been released before this date. Also, if newpipe stops working in the future on your device (due to YouTube changing something in how videos are served) you can probably install a custom rom (if your phone has any available).

SC1040-TS2:
I was away for a number of days, so seeing this PR is a bit of news to me. It would appear that after some thought, the original intent of my issue was valued above more conservative means of deprecation.

I presume that with all things considered, the move away from older forms of Dalvik was especially given value, as well as streamlining development. I suspect that NewPipe will eventually drop support for the runtime environment entirely in two or so version releases, depending on what the actual development timeline looks like.

Other than that, I send my regards for the time being.

Riotous:
@Brimont Time to buy a new phone

bidaian:
@theScrabi, I own a tablet and a phone with 4.2 and I love NewPipe. Getting a new ROM or a new phone is not really an option right now. Would forking NewPipe work for a while at least? I could revert the commits for this issue and follow you.

gkeegan:
Forking and reverting commits would not work, as some recent commits were made to fix the way YouTube gave videos. If you reverted those commits, you would not be able to play or download videos. 

There is some discussion about making a fork that would keep compatibility with older versions of Android, but as far as I know, as of this moment, there is no working build.

Riotous:
@bidaian If you can't buy a new phone and tablet, you should consider flashing these devices with newer versions of Android if you can. If you can't, there's no other solution for you than to buy new devices.  4.3 and below is being deprecated by a lot of software and you should really not be using anything older.

gkeegan:
I would not necessarily say no other option, as there are still alternatives such as using  SkyTube (which as far as I know still has support for 4.2) for your YouTube needs.

orinjus:
I would just like to add that I don't use my old devices for anything sensitive, and I can't buy a new device just for youtube. Previous NewPipe (0.14.1) worked (unstable but adequate) on android 4.0.4 though not "supported".

I appreciate this project deeply, partially _because_ I could use it on older devices.  I don't know how difficult/complex it is to write for older APIs, and I totally understand this is a volunteer project, so... yeah, you do what you need to do.  For my part though, this will limit my options a lot.

Riotous:
@orinjus @gkeegan 
https://dentex.github.io/apps/youtubedownloader/
https://github.com/ram-on/SkyTube

d-g:
Riotous wrote:
> https://dentex.github.io/apps/youtubedownloader/

It’s nonfree.

gkeegan:
I don't understand what you mean. Please elaborate. SkyTube (not the extra version) is FOSS.

d-g:
Keegan Given wrote:
>>> https://dentex.github.io/apps/youtubedownloader/
>>
>> It’s nonfree.
>
> I don't understand what you mean. Please elaborate. SkyTube (not the extra version) is FOSS.

Hmm, I believed, I cited a relevant line while replying; did it get mangled?

SkyTube is free, Dentex’s ‘YouTube Downloader’ is not.

gkeegan:
I believe its something with my app. Looks fine in a browser.

starks:
There are older devices out there (sticks, headunits, etc) that run pre-4.4 Android and cannot be easily upgraded. They need a solution.

d-g:
starks <notifications@github.com> wrote:
> There are older devices out there ... that run pre-4.4 Android and cannot be easily upgraded. They need a solution.

Did not NewPipe maintainers restored support for older Android about a month ago as a separate build [0]?

(And I’d like to take the opportunity to sincerely thank them for that, as Android toys are indeed typically designed to became obsolete quite soon, and it’s not easy for an unsophisticated buyer to predict, how soon.)

[0] https://f-droid.org/en/packages/org.schabi.newpipelegacy

PR #1891, opened 2018-11-19
Move Giga downloader into a separate module.

R4md4c:
I've been following the discussions in #962, and was keen on contributing something to it.

I thought that it would be a good idea to start with modularizing the Downloader code into a separate module so that the `app` module would depend on it. With that we separate the concerns between the actual downloading/scheduling logic and displaying it in the `app` module

I've changed the package id to match NewPipe's id, and made the necessary changes for it. 

If you like these changes and provided that they get merged, I have another [branch](https://github.com/R4md4c/NewPipe/tree/download-new-module2) ready that converts the `DownloadMission` into an interface that is consumed by the fragment so that later we can switch implementations of the `DownloadManager` to be adapted with something like [OkDownload](https://github.com/lingochamp/okdownload), or what ever the preferred Downloading choice.

Let me know please if you have any concerns or questions.

Something unrelated to this PR:

* Are there any reasons why you aren't using a Dependency Injection framework or even the concept itself? as I'm seeing static methods getting called from within the fragments and activities and that makes it hard for them to be tested or even better change the architecture to something like MVVM. 

* Any reasons why until now NewPipe is not using Kotlin yet in the codebase?


theScrabi:
@R4md4c please hold back changes for this until @kapodamy's changes are merged.

theScrabi:
1. NewPipe is an old project, it was started in a time where I didn't know about mvvm etc.
2. it uses java because of the same reason as nr. 1
However at least for the backend i don't want to move to kotlin because it should stay independent and simple to use.

R4md4c:
Thanks for the answers. 

So basically you have nothing against the two points except for point 2 that Kotlin should be applied only to NewPipe's client code (this repository and not the extractor). 

Do you have a preference regarding which DI framework to use? I personally recommend Dagger2 but that comes with expense of using code generation and that could prolong the build time.

PR #1905, opened 2018-11-25
add Warning before playing live stream on mobile data

### LINKED ISSUE #1888: [Feature Request] Warning before playing live stream on mobile data

As an end-user I want to be presented with a warning before application starts playing a live stream when I'm on mobile data so that I don't accidentally use all the data.

trulex:
fixes #1888
- add warning before playing live stream in background
- add setting for enabling/disabling warning before playing live stream in background


app/src/main/res/xml/video_audio_settings.xml
```diff
@@ -33,6 +33,13 @@
         android:summary="@string/show_higher_resolutions_summary"
         android:title="@string/show_higher_resolutions_title"/>
 
+    <SwitchPreference
+        android:defaultValue="true"
+        android:key="show_warning_live_stream_on_mobile"
+        android:summary="Show warning when playing live stream in background on mobile data"
```
TobiGr: Please put the summary and title into the strings.xml file

trulex: Done, thank you

TobiGr:
I tested this quickly. The warning is only shown when pressing the background button. But what about the fact that streams can be minimized to background, too? I'd suggest to show a warning.

trulex:
> I tested this quickly. The warning is only shown when pressing the background button. But what about the fact that streams can be minimized to background, too? I'd suggest to show a warning.

Good point. I will check how to do it since it looks to me that currently background play is initiated differently for each of these two cases.

roazh:
Hello,

Sorry Google traduction, i'm French but it's easier for me.
I installed Newpipe on my Xiaomi Mi box 3. I understand that there was no Android TV version so I took the Android version but despite that we can navigate with the remote when we look a video you can not pause with the remote.

And if not for subscriptions once we import the subscriptions of youtube I have a problem is that it is not in order. We can have a video of a month ago in front of one of another youtubeur from 1h ago. It's mixed. After if we go in youtube youtube channel it works but it would be more convenient. And if not thank you very much for your work it's really nice not to have an ad

TobiGr:
> Good point. I will check how to do it since it looks to me that currently background play is initiated differently for each of these two cases.

@trulex Have you managed to find some time to look at this? If not, don't worry ,we are busy, too.

trulex:
I will need to refactor initiation of background playout first so there won't be duplicated code for this feature. I don't know yet if it's possible to do it in an elegant way.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -849,15 +850,37 @@ private boolean shouldShowComments() {
     //////////////////////////////////////////////////////////////////////////*/
 
     private void openBackgroundPlayer(final boolean append) {
-        AudioStream audioStream = currentInfo.getAudioStreams()
-                .get(ListHelper.getDefaultAudioFormat(activity, currentInfo.getAudioStreams()));
+        if (MobileDataHelper.shouldDisplayWarningForMobileData(currentInfo, getContext())) {
+            displayWarningForMobileData((dialog, which) -> {
+                if (which == DialogInterface.BUTTON_POSITIVE) {
+                    playInBackground(append);
+                    PreferenceManager.getDefaultSharedPreferences(getContext()).edit()
```
theScrabi: Emtpy line

app/src/main/java/org/schabi/newpipe/util/MobileDataHelper.java
```diff
@@ -0,0 +1,44 @@
...
+import android.net.NetworkCapabilities;
+import android.net.NetworkInfo;
+import android.os.Build;
+import android.preference.PreferenceManager;
+
+import org.schabi.newpipe.R;
+import org.schabi.newpipe.extractor.stream.StreamInfo;
+import org.schabi.newpipe.extractor.stream.StreamType;
+
+public class MobileDataHelper {
```
theScrabi: We are already checking mobile data for selecting the stream based on how you connect. So a helper would confuse things up as the already existing solution is not using a helper. As this class is only used once on livestreams I suggest to you to also
- either name it `MobileDataLiveStreamHelper`
- or make the stream select mechanism use your solution as well.

app/src/main/java/org/schabi/newpipe/util/MobileDataHelper.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static boolean shouldDisplayWarningForMobileData(StreamInfo streamInfo, Context context) {
+        boolean showWarningPreference = PreferenceManager.getDefaultSharedPreferences(context)
+                .getBoolean(context.getString(R.string.show_warning_live_stream_on_mobile), true);
```
theScrabi: empty line

app/src/main/java/org/schabi/newpipe/util/MobileDataHelper.java
trulex: I renamed it

Stypox:
I'm closing this, since it's old and the author said he wouldn't work on this anymore in #1888

PR #1916, opened 2018-11-30
Issue 1795:Persisting the repeat and shuffle preference

agarwalakash06:
Issue#1795 Persisting the repeat and shuffle preference


app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -100,6 +102,10 @@
     public static final boolean DEBUG = !BuildConfig.BUILD_TYPE.equals("release");
     @NonNull
     public static final String TAG = "BasePlayer";
+    @NonNull
```
theScrabi: Puth the keys for the shared preference into this file:
https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/res/values/settings_keys.xml

PR #1918, opened 2018-12-01
add stalebot confic

theScrabi:

This will add stalebot config to newpipe.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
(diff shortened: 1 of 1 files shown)


Riotous:
@theScrabi Getting way too many notifications because of this bot. I get notified whenever it marks a thread I spoke in as stale. I suspect others will come to say the same thing.

TobiGr:
IMHO we should remove the bot again. It is not a good idea to close a ticket because of inactivity. There are many important discussions and feature request / enhancements / tips etc which might be important for future tickets. 
If there is no open ticket for a specific request, someone is going to open a new one instead of continuing the discussion. This means good points someone made earlier might be lost in the new discussion. So closing old tickets just because someone had no time to implement the requests seems to be a bad idea. In the end we will have a bunch of duplicates and spend much time with explaining things multiple times.

PR #1953, opened 2018-12-19
Make ducking on temporary audio focus change configurable

Redirion:

My usage scenario: driving from / to work and listening to someone talking on NewPipe that plays in background. This get's interrupted unpleasantly when a notification from another app is played (e.g. new message). NewPipe is humble enough to turn down the volume temporarily to have full notification sound. At the same time I have a hard time to listen what the person is saying.

To overcome this I could of course set to not disturb mode or turn off notifications from every app that sends them. Both workarounds are not practical. So why not adding a feature to disable lowering the volume when a notification is received?

This PR does exactly that. Previous behaviour is preserved, it is just an additional setting.


Redirion:
Tested on Androdi 7.1.2 and works as expected

PR #1969, opened 2018-12-27
Show close button when playing completed

nv95:

Small feature what shows "close" button on main video player when video playback ends. It improve UX, especially on tablets.

screenshot_2018-12-27-19-43-30


TobiGr:
What do you think about using `X` instead of `close`?

theScrabi:
Where to put the X? I think simply an X below the "replay" button looks a little misplaced. If its a word in my opinion it looks better.

Issue #1976, opened 2018-12-30
Simplify navigation to the background player activity


voidc:
In my opinion it should be easier to open the background player activity during background playback.
I suggest to replace the toast, which is shown when starting background playback, with a snackbar allowing to directly open the background player activity.
Additionally a menu item might be added to the top toolbar during playback in the background which opens the  background player activity when clicked.
This would also serve as an better indicator that the background player is active besides the notification.

voidc:
The same could be done for downloading videos. i.e. showing a snackbar allowing to open the downloads activity when starting a new download and showing an icon in the toolbar while downloads are running

MuellerSeb:
I miss this feature, too. I always tap on the tray widget of the background player to access its playlist, which feels a bit inconvenient.

MuellerSeb:
Ping.

Stypox:
Closing in favour of #2477

PR #1977, opened 2018-12-30
Simplify navigation to the background player activity. (Issue #1976)

voidc:
See #1976.
Currently implemented:
- Snackbar which shows when starting background playback

---


TobiGr:
Hey @voidc,
this looks like a good start :)
Our intention was to add a better UI for the background player which is accessible via a a snackbar-like bottom bar like it is used by VLC and others:

This bar should be persistent and contain the most important things like stream name, pause button and length / current position.
Do you think you can implement this? If not, that's not a problem :) In this case, I'd suggest to use `Snackbar.LENGTH_INDEFINITE` instead of `Snackbar.LENGTH_SHORT`. 

Btw. there are multiple ways to trigger the background player. e.g it can be started from many longpress menus.

theScrabi:
We should also directrly show the UI when starting play in background I think.
The reason we don't do it rn is more of a bug rather than a feature :P

PR #2014, opened 2019-01-18
fix decrypt error once again

theScrabi:
This will fix the decrypt issue of NewPipe that occurred lately.


snappyapple632:
When can we expect to have this fix rolled out?

PR #2023, opened 2019-01-18
Add playlist to queue when long clicking on 'Background'

dotvirus:

that button
(Someone had to make NewPipe great again)


TobiGr:
Welccome @dotvirus the code looks good 👍 . Can you do the same with the two other options, too?

Btw, you can use lambda expressions to make your code shorter.

dotvirus:
Added appending to popup queue.
Long click on Play All is useless.

Issue #1981, opened 2019-01-01
Set up own F-Droid repo to push urgent updates to users


AltNico:
At 35C3 there was a F-Droid meetup where @theScrabi showed up and explained your need at NewPipe to sometimes quickly push updates to F-Droid. This is totally understandable and we at F-Droid would really like to offer quicker updates through our infrastructure, but at the moment we can't improve publication time without reducing F-Droid's security standards.

To give some insights: as you all know, at F-Droid everything gets built from source. Unlike Google Play, this means that the building process isn't decentralized at developers' infrastructure, but we rather have a central place where everything is built. This is a great target for people to infiltrate F-Droid's infrastructure and get onto hundreds of thousands of people's devices. Therefore, we have quite high security standards here and run builds only on fully trusted hardware, maintained by fully trusted F-Droid members.

There are plans to decentralize F-Droid's build infrastructure to let builds run on multiple semi-trusted servers run by semi-trusted F-Droid members, but those plans require a lot of work and aren't reachable within the next months. Anyway, what's needed for them are apps to be built reproducibly. So most of the work proposed in this issue needs to be done anyway when F-Droid starts to offer quicker updates.

What I propose to NewPipe (and any other app that wants to distribute quicker updates for apps installed through F-Droid) mainly consists of two parts:
* Set up a F-Droid repo where you publish official builds of NewPipe that can be reproducibly built by F-Droid
* Include [F-Droid's update channels library](https://gitlab.com/fdroid/update-channels) into NewPipe to inform users about urgent updates

Let me explain this a little bit further.

People around F-Droid worked hard to make [reproducible builds](https://f-droid.org/en/docs/Reproducible_Builds/) possible. What this means is that F-Droid tries to build an app and then compares this build with the app provided by upstream. If those builds match, F-Droid publishes the upstream .apk. As of now, there are two reproducibly built apps in F-Droid: Öffi thanks to @schildbach's and Briar thanks to @grote's work.

This alone doesn't solve F-Droid's slow publication time yet. It still needs the same amount of time to publish apps, in some cases it could even take more time if there are problems with apps' reproducibility.

However, one important aspect changed: the apps are now signed by upstream, meaning upstream can publish updates which can be installed seamlessly by users.

[...]

PR #2048, opened 2019-01-21
Add updating section to README

gkeegan:
This will help cull the number of "how to update?" Issues whenever we need to push out a fix, like how recently NewPipe got completely broken.

This attempts to be an all-inclusive guide to updating NewPipe. If there are any (clean/official) links to additional info, please comment them and I will add it.

Also let me know if this section should be relocated within the README. I figured it would be best to keep it up top, so people see it right away, and then don't make a new Issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,20 @@
 
 <b>WARNING: PUTTING NEWPIPE OR ANY FORK OF IT INTO GOOGLE PLAYSTORE VIOLATES THEIR TERMS OF CONDITIONS.</b>
 
+## Updates
+When a change to the NewPipe code occurs (due to either adding features or bug fixing), eventually a release will occur. These are in the format x.xx.x . In order to get this new version, you can:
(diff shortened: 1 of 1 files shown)


theScrabi:
Yea that is ok I guess, ee should put it into te faq as well.

gkeegan:
@theScrabi What location within the Readme should this be put? On second thought it seems odd to have it above the screenshots and other parts. Could it be better to just have a "Having issues updating to the newest version" link at the top like how Screenshots, Description, Features, etc. do?

TobiGr:
> Yea that is ok I guess, ee should put it into te faq as well.

 @theScrabi Put it on my list 👍 
@gkeegan can you please add a link to #1981 which resolves the problem with different signing keys? @theScrabi Is there a way other people except you and theAssasin can help to get reproducible builds running? If yes, we could add an encouragement to maybe find someone who can help.

theScrabi:
Wupwup merged PR 2048. This means until 4096 I am not able to continue.

PR #2056, opened 2019-01-22
Colored Notifications for Background Player

ghost:
- [ yes ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

I added these nice colored notifications, they take a common color in the thumbnail and it changes the background color to match. This is my code contribution to the project, if there is something wrong, I can fix it :-)

This would also be great to add as a setting, but I am not sure how and have to get ready to go somewhere, I am not sure if everyone will like these colorful changes.

Also, grey on Yellow looks a little strange, I will create PR if I can get the Progress bar to match as well. For well its functional and hardly noticeable.
51508567-d459e480-1dc3-11e9-810c-2869034ff8f6 51508568-d4f27b00-1dc3-11e9-893f-8331b9254481

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
@@ -301,7 +301,7 @@ public final class BackgroundPlayer extends Service {
             if (basePlayerImpl == null) return;
             // Generate palette
             Palette coverColor = Palette.from(basePlayerImpl.getThumbnail()).generate();
-            Palette.Swatch swatch = coverColor.getVibrantSwatch();
+            Palette.Swatch swatch = coverColor.getMutedSwatch();
             if (notRemoteView != null) {
                 notRemoteView.setImageViewBitmap(R.id.notificationCover,
                         basePlayerImpl.getThumbnail());


theScrabi:
Thats cool but we must see if this notification code will survive as we have trouble with it on android 8.1.

If you have time would you like to remove the progressbar and the timr view on it? Then we can merge this :)

FaizKhan5:
Timer is important to view how long the actual audio is and how much i listen.

TobiGr:
@theScrabi What is your intention behind removing the progressbar and the timer? Reducing the update frequency?

nekobit1:
Yeah, I use the Progress bar as well. Would suck to remove it (although, I will fix the colored progressbar), although I would be fine with it

FaizKhan5:
Why it brake the app?

theScrabi:
Closing this, because I think in the future we should go with the Android native Player notification. This one is better designed for players, and we don't have trouble with android 8.1 any more. Also since we have the player UI that pops up when you click on the notification I think it is ok if not all functinons that are currently in the notification are not present any more.

screenshot_20190123-123108
screenshot_20190123-123129

theScrabi:
> Why it brake the app?

The UI is flattering when you play something in background. It's flattering because it tries to update the Notification in the UI thread, but since android 8.1 this takes a while.

PR #2064, opened 2019-01-26
Downgrade build tools version to fix F-Droid builds

TheAssassin:
Android broke some licensing stuff related to the SDK and the build
tools. I was told that downgrading to 28.0.2 potentially fixes the
issue.


TheAssassin:
The F-Droid team pushed a fix to the metadata, so we'll see if that fixes the build on their systems. Closing as deprecated.

PR #2086, opened 2019-02-01
Updated search fragment to be amazon fire tv friendly

clockworkant:

I've tried to make as little changes as possible whilst implementing enough functionality that the newpipe search is now usable on amazon fire tv products.

You can now search by clicking on the search box it will display the keyboard.

I've run the junit tests and tested it on a pixel 2 emulator and it appears that the search still works as expected.

Caveats;
The videos play best using an external player. VLC works very very well.


theScrabi:
Nice, I was looking forward to this :D

app/src/main/java/org/schabi/newpipe/firetv/FireTvUtils.java
```diff
@@ -0,0 +1,11 @@
+package org.schabi.newpipe.firetv;
```
theScrabi: Please put this class into `org.schabi.newpipe.utils`.

theScrabi:
Closing this PR since no reply on change request.

PR #2092, opened 2019-02-03
support for youtube-nocookie.com

connectety:

This PR adds "youtube-nocookie.com" as a valid link to the manifest and updates the extractor version.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -183,6 +183,19 @@
                 <category android:name="android.intent.category.DEFAULT"/>
                 <category android:name="android.intent.category.BROWSABLE"/>
 
+                <data android:scheme="http"/>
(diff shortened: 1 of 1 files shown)


theScrabi:
@connectety please remove the line where you update the extractor. As master is already at a newer commit.

app/build.gradle
```diff
@@ -57,6 +57,7 @@ dependencies {
         exclude module: 'support-annotations'
     })
 
+    implementation 'com.github.TeamNewPipe:NewPipeExtractor:8114ce6ae4'
```
theScrabi: Please remove this line as `dev` is already at a newer commit.

connectety: I added this line when NewPipe used an older version. I do not think there were any changes in build.gradle after the first merge. This probably just came along due to me merging and rebasing.

PR #2107, opened 2019-02-08
Fix padding in settings layout (Fixes #1866)

### LINKED ISSUE #1866: Settings layout broken
- I did not use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

In 0.14.2 and later the layout of the sub-setting pages is wrong. It looks like there is an additional column for icons.

chin123:
   
There was previously a gap before every setting option (#1866).
Here's how it looks now:  

52512280-aa792e00-2bc9-11e9-91e1-1dfe60634d55

### DIFF
--- a/app/src/main/res/xml/main_settings.xml
+++ b/app/src/main/res/xml/main_settings.xml
@@ -1,41 +1,49 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PreferenceScreen
     xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:key="general_preferences"
     android:title="@string/settings">
 
     <PreferenceScreen
+        app:iconSpaceReserved="false"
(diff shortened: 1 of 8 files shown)


app/src/main/res/xml/main_settings.xml
```diff
@@ -1,41 +1,49 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PreferenceScreen
     xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:key="general_preferences"
     android:title="@string/settings">
 
     <PreferenceScreen
+        app:iconSpaceReserved="false"
```
TobiGr: Why did you add this here?
In case we want to add this lines to every settings fragment, we there might be a better approach: https://stackoverflow.com/questions/51518758/preferencefragmentcompat-has-padding-on-preferencecategory-that-i-cant-get-rid#answer-52960668

PR #2117, opened 2019-02-12
Small ux improve

nv95:

Removing search history items using swipe & clearing watch history using options menu


theScrabi:
Generally full ack, but have you checked the behaviour of the history page when it is set to be shown in the frontage?

theScrabi:
screenshot_20190212-125337

theScrabi:
I guess you have an overlapping swipe here. A solution to this issue would be to detect weather the page is in frontpage mode and then deactivate the swipe to delete function.

You can use this variable to check:
https://github.com/TeamNewPipe/NewPipe/blob/ce6a5ebf0a2f18a48c70a432b2536074bc509bea/app/src/main/java/org/schabi/newpipe/BaseFragment.java#L27

However it is protected. So it would need to be refactored. That it is private and a getter is used to retrieve its value. Maybe you could please do this as well.

nv95:
> overlapping swipe

Where? O_o

Swipe removes **search** history items, not history)

theScrabi:
Oh :D sorry

TobiGr:
> Removing search history items using swipe

Where can I do this? I've tried this when typing into the search field  But nothing happened. There is only a dialog which asks whether I'd like to remove the item from the search history when doing a long press on an item.
I tried this on an Android 5.0 tablet and Android 6 phone.

PR #2136, opened 2019-02-19
add database doc in dia

theScrabi:
This will add a databasedocumentation created using [dia](http://dia-installer.de/index.html.de). It is stored in the `assets` folder, and is ment to be used in the wiki/documentation.

@TobiGr it would be nice if you could put the svg to it somewhere usefull :).

db


TobiGr:
👍 Will do this next week after I finished my last exam.

theScrabi:
Uff two tables have the name search_history :/. Well the one pointing to streams should be called watch_history. (I think).

TobiGr:
Yes, you're right

TobiGr:
> it would be nice if you could put the svg to it somewhere usefull :).

@theScrabi I'd suggest to put it into the wiki and
a) create a new chapter in the documentation explaining which steps should be followed before/when implementing a new feature or
b) create a new chapter about the NewPipe's. (btw. what about the giga / downloads database which is located [here](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/us/shandian/giga/get/sqlite/DownloadMissionHelper.java)?)

PR #2161, opened 2019-02-26
Fixed CheckForNewAppVersionTask being executed when it shouldn't

Redirion:
1. check if cancel was called in onPrepare
2. if we currently don't have a Connection, don't show crash report dialogue to user


theScrabi:
Sounds usefull.

Issue #2170, opened 2019-03-01
Lag when video plays in background.


EnriUnagi:
The app severally lags when playing a video in the background making it very difficult to do anything while it's playing. I'm using Android 8.1 btw.

theScrabi:
Confirmed. I know the issue i hate it as well, and I said it need to be fixed but someone has to do it.

Th issue is the timestamp update in the notification. It should be removed.

ronidee:
On my device, some videos do not cause UI lag while playing in background. 

One of them is Cheat Codes - SEX (https://www.youtube.com/watch?v=OORoOGY8D2M)

My device: Honor 9 lite, los 15.1, newpipe 16.1

Why is that??? 🤔

SebiderSushi:
The pull request has been merged into the newly released version 0.16.2 so this should be fixed. I'm using Arch btw.

ronidee:
Just wanted to report that I just installed 0.16.2 and this issue is still present.
(My device, look above) 

`Debian sid`

MohamedKiouaz:
Hi,

Same here, J7 2017 with Android 8.1.0.
It's almost not possible to use New Pipe while playing a video in background.
I am listening to [this long music video](https://www.youtube.com/watch?v=snRZHMqnItA) and it triggers the bug sill in 0.16.2.

Thanks for the hard work guys. It's an awesome

EnriUnagi:
Yeah this issue is still present, hope it gets resolved soon. Best of luck!

expobar:
Can confirm. 0.16.2 still has this issue. 
Worse case: the app almost freezes once I am few videos deep while background is playing.
Normal times: the app is laggy while background is playing. 

It's very rare that I trigger the app to almost freeze, once it does 'clear cache' fixes it most of the time.

Issue #1941, opened 2018-12-12
Background play causes UI lag


abhijitparida:
Using background play causes UI lag. The lag goes away if I pause background play, but comes back when I resume playback. (Only the app UI lags; everything outside the app functions normally).

Playing videos in fullscreen or popup mode doesn't cause this issue.

NewPipe version: 0.14.2
Android version: 8.1.0
Device: Moto G5 S Plus

- [x] I checked if the issue/feature exists in the latest version.
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

Octobread:
Can confirm this for android version 8.1.0 an NewPipe version 0.15.1.

PR #2178, opened 2019-03-04
Cache duration String to improve performance

Redirion:
This pull requests intents to improve performance on Backgroundplayer while the screen is on.

In VideoPlayer the Duration String is cached effectively with the help of the playbackSeekBar(getMax() and setMax(). As the playbackSeekBar doesn't exist in BackgroundPlayer, using two additional variables will reduce performance impact of notification updates by almost 50% and thus perform similar to VideoPlayer.

This addresses issue #2170 and issue #1941.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
@@ -275,6 +275,8 @@ public final class BackgroundPlayer extends Service {
     protected class BasePlayerImpl extends BasePlayer {
 
         @NonNull final private AudioPlaybackResolver resolver;
+        private int cachedDuration;
+        private String cachedDurationString;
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
```diff
@@ -351,8 +353,12 @@ public void onUpdateProgress(int currentProgress, int duration, int bufferPercen
             resetNotification();
             if(Build.VERSION.SDK_INT >= 26 /*Oreo*/) updateNotificationThumbnail();
```
TobiGr: I just saw this line while reviewing your PR and didn't want to send a new PR for this single change. Please add it :)
```suggestion
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O /*Oreo*/) updateNotificationThumbnail();
```

PR #2198, opened 2019-03-08
Directplay on Background

### LINKED ISSUE #2187: Extra option in addition of `Enqueue when Backgrounded`
# The problem

One think that will avoid many time for me.

Let's say that a song is already playing in background.

Now i am searching and decided to play another song on Background directly

What i do is Enqueue in background and i go to background player and press next to play the one that i selected to be Enqueued.

# One Solution
Can we have an option like....

If another song is playing in background... replace it and directly play the one i want. As an addition to Enqueue in background.

# Let's have an extra menu option called '' Play directly in Background ''

I swear it saves me a lot of time :)

Thank you team. 

Image

Redirion:

closes #2187 

See the issue description and discussion. (with pictures!)

Regarding difference between "Play on Background" (commit and branch name) vs. "Play in Background" (strings.xml): from developer perspective playback runs on background player while from user perspective playback runs in background. ;)

Tested on Android 7.1.2.


Redirion:
> 
> 
> Code is ok, please check QA.

could you please link or explain what you mean with QA? My change is extremely small and straightforward and tested.

theScrabi:
Quality Asurance. We need to check if everything works as intended.

Redirion:
Ah. I thought you wanted me to check some sort of QA guideline for NewPipe. Thanks for the clarification.

TobiGr:
@theScrabi IMO, we have too many longpress menus. They have a bad UI and UX , becuase
1. You can only interact with one video/stream item at once.
1. These overlay menus block every interaction with other elements.
1. They don't look nice

I think we should replace the menus with some buttons/icons in the app bar. something like this (rather the second draft):
Bild3
In other words, make it possible to select multiple videos from a list and keep every item directly accessible.

Btw. Should I open a new ticket for further discussion on this?

theScrabi:
@TobiGr this proposal looks god and fits better into the material design then the long press menus we have right now :) I think its a good idea.

Stypox:
Oh, I just came across this PR... I saw you added the `direct_on_background` (= Play directly in background) string to the strings.xml... Does it have the same meaning as `start_here_on_background ` (= Start playing in the background)?  I removed it in #2368 because I thought it was a duplicate.

Redirion:
I thought "start here on background" means that this video would be played directly and all videos below would be queued. The option added by me would just play the video without queing others. @Stypox

Stypox:
It seems like that's not the meaning... `start_here_on_background` clears the queue and then adds the stream, just as your option does.

That name is ambiguous

PR #2203, opened 2019-03-10
added default kiosk

yausername:

* Added a default kiosk which changes with the active service.
* If a user is using a custom main page, then this will not be acitvated by default on upgrade.


yausername:
Here is the debug apk for testing. You might have to clear data/uninstall if you use a custom main page.
[app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/2950105/app-debug.apk.zip)

theScrabi:
@yausername stop reading my mind. I just wanted to propose something similar :D

Stypox:
@theScrabi that has nothing to do with this pr: I just reproduced it on the latest dev code (6787d0224cf91ca12fedc2b3731695670ebb00cf) by adding the "SoundCloud/New & Hot" tab as the first one (from "Content of main page" setting) and closing the settings (which brought me directly to the newly added tab).

theScrabi:
All right thanks :) then this feature is ready to be merged :)

PR #2205, opened 2019-03-12
Fit thumbnails size on grid

### LINKED ISSUE #2181: deleted
deleted

nv95:

Fix #2181

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/views/AspectRatioRelativeLayout.java
+++ b/app/src/main/java/org/schabi/newpipe/views/AspectRatioRelativeLayout.java
@@ -1,0 +1,82 @@
+package org.schabi.newpipe.views;
+
+import android.content.Context;
+import android.content.res.TypedArray;
+import android.support.annotation.IdRes;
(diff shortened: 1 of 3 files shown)


TobiGr:
Thanks! Can you please add some comments to the code? Otherwise it's hard to read and understand it when you do not deal with this topic that often. Comments will help in the future when we need to come back to the code again.

I quickly tested the changes and found some buggy behaviour when changing the orientation of my phone (Android 6). Quick steps to reproduce (I'll upload a screencast later this week):
1. Enable the grid view in the settings
1. Open a video list (e.g. trending)
1. Generally, you have the phone in portrait mode. You'll notice one thumbnail in a row.
1. Turn your phone. In landscape mode, I have three smaller thumbnails in each row. Scroll some rows down.
1. Turn the phone back to portrait mode. I now see some thumbnails with the proper size, but also some which didn't change their size. Sometimes they resize when scrolling, sometimes not.

nv95:
@TobiGr this pr isn`t good)
See #2276

PR #2234, opened 2019-03-20
Fix italian translation

Gelma:
Word "ravviati" is wrong.
The correct one is "riavviati".

https://www.collinsdictionary.com/dictionary/italian-english/riavviare

I already subscribed to weblate.org, and suggested the change there. But maybe it takes time to integrate and get lost.

- [X ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.


TobiGr:
Thank your for the fix. I think you've [already corrected](https://hosted.weblate.org/translate/newpipe/strings/it/?checksum=c8832caa064063ee) this via Weblate. So there is no need to do this here. We will pull the translations before the next release.

PR #2237, opened 2019-03-21
Fixed English translations for Play/Enqueue commands

epitron:
The wording for the Play/Enqueue commands is currently a little odd (introduced in commit 6167d5dbfc6de63c50d24fdbfa524ada48432c05):

```diff
-    <string name="enqueue_on_background">Enqueue on Background</string>
+    <string name="enqueue_on_background">Enqueue when backgrounded</string>

-    <string name="enqueue_on_popup">Enqueue on Popup</string>
+    <string name="enqueue_on_popup">Enqueue on new popup</string>

-    <string name="start_here_on_background">Start Here on Background</string>
+    <string name="start_here_on_background">Start here when backgrounded</string>

-    <string name="start_here_on_popup">Start Here on Popup</string>
+    <string name="start_here_on_popup">Start here on new popup</string>
```

This commit fixes them (eg: "Enqueue when backgrounded" is changed to "Enqueue in the background", etc.)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -419,12 +419,12 @@
     <string name="play_queue_stream_detail">Details</string>
     <string name="play_queue_audio_settings">Audio Settings</string>
     <string name="hold_to_append">Hold To enqueue</string>
-    <string name="direct_on_background">Play directly in Background</string>
-    <string name="enqueue_on_background">Enqueue when backgrounded</string>
-    <string name="enqueue_on_popup">Enqueue on new popup</string>
+    <string name="direct_on_background">Play directly in background</string>
+    <string name="enqueue_on_background">Enqueue in the background</string>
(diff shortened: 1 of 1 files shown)


banksio:
What about just "Enqueue in background"? I feel like "backgrounded" sounds a little strange.

epitron:
I did get rid of "backgrounded", which was the weird language. (The diff in my comment is the commit that introduced the weird language. See my pull request's diff to see the fixed version.)

banksio:
My bad, I didn't notice that; I like your changes.

epitron:
I was debating whether to remove the "the" myself... I think it makes it a hair more obvious what's happening with the word "the", and if it fits on the screen, it shouldn't be problematic.

I can change it if anyone feels strongly against "the". :)

TobiGr:
> I think it makes it a hair more obvious what's happening with the word "the"

Yes. That's what I think, too. I guess we can keep it.

Issue #2000, opened 2019-01-11
[Feature Request] Share URL with option to add starting time


Pablohn26:
- [X] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [X] I checked if the issue/feature exists in the latest version.
- [ ] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

Youtube website provides an option to share a video starting at some point in the video.

![seleccion_743](https://user-images.githubusercontent.com/1228418/51044110-a2c16c00-15c0-11e9-937e-1bc06c50fe53.png)

Youtube Android app does not provide that feature either.

Would be possible to add that feature to NewPipe?

Gymcap:
Can we also get the ability to open such links at their specified time? These features together would be awesome.

Somethingweirdhere:
Also, clicking on timestamps in the discription!

dadosch:
I think it would be useful to provide a share option which as available from a playing video then.

Stypox:
I'm implementing this. I added a share button inside the main player, in the "more options" section. When clicking on it, the sharing dialog is opened with the URL of the video starting at the current time. That URL that **always** contains `&t=SECONDS` at the end (where `SECONDS` is the current playback seek, obviously). There is no way to manually change (or unset) the video starting time.
Even though this is a limitation, I don't think it is a good idea to show a dialog that lets users choose, because most of the times the video starting time they need is the current one. On the other hand I think it would be a great idea to add an option to the settings that makes it so that video starting time is **never** added.   
![NewPipe share](https://user-images.githubusercontent.com/36421898/55561725-f5f41800-56f2-11e9-87e8-017f28a4ab7c.jpg)

@Gymcap I think that has already been implemented (tested on 0.16.1)

TobiGr:
That's good to hear :+1:
I think we should not add too much settings and there are enough possibilities to share a video without a timestamp (e.g. from the video detail page or the playlist menu). In other words, I'd vote to not introduce an extra setting.
CC @theScrabi What's your opinion?

Stypox:
[...]

PR #2271, opened 2019-04-06
Add "share at current time" button to the main video player

Stypox:
The button is placed in the upper right corner, under "more settings". No changes has been made to the popup video player (as for other things, there is no space for a share button).
main player share button
The share button generates this text (for the video above at the current time):
```
TEST VIDEO
https://www.youtube.com/watch?v=C0DPdy98e4c&t=3
```
This implements #2000, but it does not give a chance of choosing whether to share the url normally or to at the current time. But as discussed in #2000 this is not a problem because there are already many other ways to do that.

The share utilities (i.e. `shareUrl()` and `openUrlInBrowser()`) were placed under the BaseStateFragment class. But since they have nothing to do with fragments, I moved them to a separate utilities class: `utils.ShareUtils`. This way they can be accessed from anywhere. I could not find any way to access them from the MainVideoPlayer, using the previous setup.

I tested the feature both from a phone (Huawei P9 lite, where the used layout is normal) and from a tablet (Amazon Kindle, where the layout is large-land, I think). Also, since I moved the sharing functions, I tested the "share" buttons in the app and they work as they did before.

Please note that this is my first pull request ever, so if I did something wrong I'm sorry :-)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/BaseStateFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/BaseStateFragment.java
@@ -230,21 +230,4 @@ public abstract class BaseStateFragment<I> extends BaseFragment implements ViewC
         ErrorActivity.reportError(getContext(), exception, MainActivity.class, rootView,
                 ErrorActivity.ErrorInfo.make(userAction, serviceName, request, errorId));
     }
-
-    /*//////////////////////////////////////////////////////////////////////////
-    // Utils
-    //////////////////////////////////////////////////////////////////////////*/
-
-    protected void openUrlInBrowser(String url) {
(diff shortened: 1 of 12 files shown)


TobiGr:
Thanks! Code looks good so far. Will test the changes later this week

TobiGr:
> Please note that this is my first pull request ever, so if I did something wrong I'm sorry :-)

Everything is fine :) It is perfect to have small commits. This allows us to review the changes faster :+1:

theScrabi:
I dont want to play ui nazi, but why add a button here? There is already a share button in the detail view, why have one here? It already looks pretty crowded button wise and now we add another one, this should have a god reason.

PR #2310, opened 2019-04-25
Add send to Kodi button to player next to share button

christophehenry:
[Debug APK](https://webcloud.zaclys.com/index.php/s/drYS1oNG5xpKfKp)


TobiGr:
Thank you!
Code looks good. Will test it later this week.

Ikosihenagon:
Thanks for this very great App and also for adding the Kodi Button - it is very helpful!

Since the YouTube Add-On for Kodi is not working properly the last days it would be very nice to select somewhere in the settings, if the YouTube URL or the "real" video URL will be send to Kodi. Sending the YouTube URL requires that the YouTube Add-on is installed in Kodi, if the "real" video URL can be send, there is no need for an additional Add-on in Kodi.

Thank you.

PR #2317, opened 2019-05-02
#2282, Add channel name and the current date to the downloaded file name

### LINKED ISSUE #2282: Feature Request: Filename include channel name and date in Download Dialog
1. Filename

Right now it only include video title.

Can you also include the channel name and upload date (year-month example 2018-06)?

Two presets for the format:
Channel Name - Video Title - Date
Channel Name - Date - Video Title

Allow the user to select which present or format to use. Include this in the settings to select the default present or format to use.

K1rakishou:

Closes #2282

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
@@ -43,7 +43,9 @@ import org.schabi.newpipe.util.StreamItemAdapter;
 import org.schabi.newpipe.util.StreamItemAdapter.StreamSizeWrapper;
 import org.schabi.newpipe.util.ThemeHelper;
 
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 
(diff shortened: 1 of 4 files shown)


TobiGr:
@K1rakishou Thanks for your work. I am sorry for responding so late...
It might be good to move this setting directly to the DownloadDialog. And some more options would be useful, too. We do not add the date to the filename, so there should be at least one option which does not contain the date. @kapodamy @Stypox  What do you think?

Stypox:
I prefer the original implementation, i.e. to have an option in the settings, since otherwise the user would have to choose what format they want every time they download.

A (better, imo) alternative would be to add the button to the dialog and remember the last format used, so that two issues can be solved: the number of entries in settings does not grow and the user does not have to set the format every time.

TobiGr:
> A (better, imo) alternative would be to add the button to the dialog and remember the last format used, so that two issues can be solved: the number of entries in settings does not grow and the user does not have to set the format every time.

That's what I meant. I should be more precisely

litetex:
No progress since two years, no GitHub actions build and merge conflicts.

frederickjh:
I know this issue is closed due to inactivity, but I wanted to say I would like to see this feature. However, if possible, instead of the current date and time in the filename to put the date and time that the video was published on Youtube or another platform. This combined with the channel name would allow downloaded videos to be grouped by channel and sort in the order that they were published allowing them to be watched in order. Thanks for consideration of this filename format if this issue ever gets revived!

PR #2320, opened 2019-05-04
#1570, Show video thumbnail on the lock screen

### LINKED ISSUE #1570: [Feature Request] Show video thumbnail on lock screen

It would be really cool if - when background music is playing - NewPipe would show the thumbnail of the video on the lockscreen. Many music players do this with album artwork. Here's what I mean:

drawing

K1rakishou:

Closes #1570 
There is a nasty bug, though. When setting a lock screen wallpaper and then killing the app via force stop the wallpaper will stay on the lock screen. Other apps (like Blackplayer) somehow manage to clear the wallpaper when force stopped but I have no idea how they are doing it. I even asked this question on [stackoverflow](https://stackoverflow.com/questions/55981675/lockscreen-wallpaper-stays-after-force-stopping-the-app), but got no answer yet. If anyone knows how to fix this please do tell. For now I added lock screen wallpaper clearing upon app start so it won't stay until user start background player again but until app restart.


TobiGr:
Thanks! This sounds like a cool addition to the backgorund player.
Regarding the force stop problem: I don't think we have to worry about this. IIRC there is no way to intercept a force stop and run some code. If this was possible, a force stop would be quite useless :D 
@theScrabi @Stypox Do you have any idea how to solve the problem? Is it okay if we cannot reset the wallpaper? IMO, only very few people are in the need to kill the app and we do not need to take care of wallpapers in this case.

dhnasa:
i think a better solution would be using [Notification.MediaStyle](https://developer.android.com/reference/android/app/Notification.MediaStyle).
> Finally, if you attach a MediaSession.Token using setMediaSession(MediaSession.Token), the System UI can identify this as a notification representing an active media session and respond accordingly (by showing album artwork in the lockscreen, for example). 
This way the system will handle the wallpaper and will also clear it if newpipe gets killed

K1rakishou:
Ohhh, so this is how they do it. Alright, I'm gonna use Notification.MediaStyle instead of WallpaperManager then.

PR #2322, opened 2019-05-04
Two-finger click to enqueue stream on background

### LINKED ISSUE #2306: [Feature request] swipe while searching to add to queue.
I use this app mainly to listen to music in background, so it will be useful to just look for music and swipe them to the side to add to queue.

Stypox:

Added a two-finger click gesture to the util.OnClickGesture class. It is triggered when the user touches a stream with two fingers, enabling him to quickly enqueue that stream on background.

I tested all the gestures in StreamMiniInfoItemHolder thoroughly and they work (after fixing some complications explained below). [test apk](https://github.com/TeamNewPipe/NewPipe/files/3144692/app-debug_1.zip)

The gesture detector in StreamMiniInfoItemHolder is not very readable, there probably is a way to make it better. The problem is it has to detect single clicks, held clicks and two-finger clicks: without manually handling every gesture there is no way (or at least I couldn't find it) not to block single clicks when a two-finger click or a held click is happening. I'm using the view.onTouchEvent() to detect normal clicks and two-finger clicks and delegating the work to a SimpleGestureDetector to detect held clicks.

Closes #2306

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
@@ -190,6 +190,9 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I> implem
             public void held(StreamInfoItem selectedItem) {
                 showStreamDialog(selectedItem);
             }
+
+            @Override
+            public void twoFingerClick(StreamInfoItem selectedItem) { enqueueStreamOnBackgroundPlayer(selectedItem); }
         });
 
         infoListAdapter.setOnChannelSelectedListener(new OnClickGesture<ChannelInfoItem>() {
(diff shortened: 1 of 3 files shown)


Stypox:
To improve the user experience, I think it would be a good idea to add a label to the long-click menu explaining the existance of the gesture, similarly to the "hold to append" tip. Otherwise almost nobody would use it, even if it is useful.

Stypox:
After reading https://teamnewpipe.github.io/documentation/07_maintainers_view/ and thinking a little bit about it, I'm closing this PR since it does not meet most of the requirements stated in that page of guidelines:
- this feature was requested by 1/2 people not multiple ones
- the code is hacky and I don't know how I could improve it

I think that link should be added to the contribution guidelines, since if I had read that page before I would not have opened this PR :-)

PR #2333, opened 2019-05-07
Fixed setPlayWhenReady behaviour inconsistency

### LINKED ISSUE #2331: paused video plays in background
if i leave a video on pause and the screen times out, it starts auto playing in the background. it does NOT respect the pause setting of the view mode.

Redirion:
this commit aims to fix multiple possible issues by simplifcation:
- multi null checks to return a method early could result in actual available object not being used
- audioreactor request and abandon of audio focus are both leading to setPlayWhenReady calls, so doubled calls have been removed
- simpleExoPlayer.setPlayWhenReady(true) was not respecting isResumeAfterAudioFocusGain (see AudioReactor). It is gone now, as explained in list point above.

Might fix #2331

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -889,27 +889,22 @@ public abstract class BasePlayer implements
 
     public void onPlay() {
         if (DEBUG) Log.d(TAG, "onPlay() called");
-        if (audioReactor == null || playQueue == null || simpleExoPlayer == null) return;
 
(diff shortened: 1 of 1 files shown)


TobiGr:
Looks good, Will test it this weekend.

TobiGr:
@Redirion Something went wrong here. I can't check out your changes. Did you delete the branch? GitHub says: 
> Redirion wants to merge 1 commit into TeamNewPipe:dev from unknown repository

I guess merging this might be difficult :laughing:

Redirion:
How can I solve this? close this PR and open a new one? Otherwise you could try to cherry pick the commit.

PR #2288, opened 2019-04-15
Playback positions in lists

nv95:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Implementation of displaying playback progress in lists

| Screenshot_2019-04-15-22-25-20 | Screenshot_2019-04-15-22-25-54 |
|---|---|

Code based on #2286 ~~and is not ready for release yet~~

PR #2368, opened 2019-05-29
Make long-press menu consistent across views

### LINKED ISSUE #1142: [feature request] add "add to playlist" to the list of related videos

can add to the menu of the related videos the option of "add to playlist" too?
It's strange, this option is in the results but not in the information of a video, in the list of related videos

### LINKED ISSUE #2354: Making the long-press menu consistant across the application

The order of the items in the long-press menu is different in every view (playlist, search, info...). I am willing to fix this inconsistency, but I thought I'd open an issue anyway, just to get some feedback before proceeding. So what do you think? Should I make long-press menus the same across the whole app?

Stypox:
- [v] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

The order of the items in the **stream** long-press menu should be the same in every view (playlist, search, info...), so that the user can subconsciously memorize it, without having to read the same info over and over again. Also, I think consistency is a good thing. See #2354 for more info.

I fixed the menu so that it is the same across the app, and I chose this order:
```
enqueue-background
enqueue-popup
new-background
new-popup
[  ...  specific actions, such as delete-from-history]
add-to-playlist
share
```
The **subscription** long-press menu (in the local subscription list) was also modified to comply with the above order: it is now `unsubscribe; share;` instead of `share; unsubcribe;`.

I decided to remove the `new-on-main-player` option since it was mostly useless. The other play options are useful to (for example) enqueue while carrying on browsing flawlessly. But opening the main player interrupts what the user was doing anyway. And, from a time point of view, one could open the stream details and open the main player in roughly the same time it would take to long-press and then click.

- [X] Before merging I'd like to improve the menu labels: those were inconsistent, too. The new-background option had two different strings used in different views: "Play directly in background" (`direct_on_background`) and "Start playing in the background" (`start_here_on_background`). So the `direct_on_background` string should be removed, along with `start_here_on_main` which is no longer used (as explained above).
- [x] Soundcloud has no video playback, but the entry to play on popup appears in the menu anyway. This has to be fixed.
- [x] Remove the ugly if-else-cascade
- [x] Fix crash when closing a starting/loading popup.

Testing APK: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/3431118/app-debug.zip)

Fixes #1142
Closes #2354


Stypox:
I addressed the duplicate `direct_on_background` string in b6cfb8a3dcd6e1f89bf4ab28b2269853e2b50bfa, but I decided to leave the unused `start_here_on_main` there, so that if it ever becomes useful again it's ready to be used.

Stypox:
There are no popup entries anymore when the long-pressed stream is audio-only. I had to replace the `switch` block with the ugly `if - else if`, but I don't think anything else could be done.

Redirion:
> There are no popup entries anymore when the long-pressed stream is audio-only. I had to replace the `switch` block with the ugly `if - else if`, but I don't think there was anything that could be done.

you could work with enums: https://schneide.blog/2010/12/13/avoid-switch-use-enum/

Stypox:
@Redirion fixed. Is it ok now? I did my best, and I really love Java enums!

Btw 9df27f43de590bf1dcd6e2081925d991b2aa0477 is just to improve safety for any future user of StreamDialogEntry code.

Stypox:
Also, I found a bug: the app crashes when the popup player is closed while starting. I don't know if this problem was caused by my code.

Redirion:
> @Redirion fixed. Is it ok now? I did my best, and I really love Java enums! heart_eyes

It looks beautifully!

You have some unused imports in BaseListFragment that could be cleaned up. 

I will test the code later (in about 8 hours).

Stypox:
@Redirion I added a test APK to the description

Stypox:
Ok, fixed the crash. The problem was caused by the recent resume playback pr #2288: the popup's BasePlayer tried to re-init when closing, since the loaded-listener was called both on video loading completion and when a not-yet-loaded popup is closed.

@nv95 can you please confirm that what I did in 7c9ef58acd0971ef47c78f4d3247d388991044bc was ok?

justanidea:
I tried the Apk and all seems to be good for me

nv95:
@Stypox I'll look on this on weekends

nv95:
@Stypox  I was try it and looks like everything is ok

Redirion:
I have used it on daily basis since you provided the apk and only had two unrelated crashes. This should be fine for merge.

kurobon-jp:
I don’t understand why `Start playing here` and `Start playing in the background` deleted.
Did you decide that you don't need the action that plays continuously from the middle of the list?

Stypox:
I did not delete either of them. They are still there and are being used (see the apk in release 0.17.0). I only deleted `direct_on_background` since it was a duplicate of `start_here_on_background`.

kurobon-jp:
In my environment, after merging the `Stypox:menu-consistency`, the `Start playing here` are gone.

e7b068ed8e87c4163ff09ce2407d4cfdc3557b8f
old_1
4e5a20ec4576b95835b2b2eb2fa984de383a4ea0
new_1

`Start playing in the background` items remain, but it seems to be different from the previous behavior.

e7b068ed8e87c4163ff09ce2407d4cfdc3557b8f
old_2
4e5a20ec4576b95835b2b2eb2fa984de383a4ea0
new_2

Stypox:
Yes, I removed "start playing here" because it seemed mostly useless. See above.

About the differencies about arrows... I don't know why that happens, will look into it in a week (after vacation)

PR #2396, opened 2019-06-13
Fixed setPlayWhenReady behaviour inconsistency

### LINKED ISSUE #2331: paused video plays in background

Redirion:
reopened #2333 for merging
closes #2331


Redirion:
pausing videos is not possible anymore. Will close until inspection is done.

Redirion:
okay. so "setPlayWhenReady" is actually used for pausing and resuming playback of ExoPlayer. When abandonAudioFocus is called, there is no listener event, that audio focus is gone. Therefore "setPlayWhenReady" was not called as expected from me in the AudioFocusReactor. Same is in onPlay. AudioFocus is requested but not gained (maybe as it isn't lost to another app?)

PR #2413, opened 2019-06-21
Add move to next feature in play queue

josephoneill:

In order to make organizing songs within a PlayQueue easier, I have added a "move to next" menu item.

### Feature added:
The move to next menu item appears when a PlayQueueItem is held down, bringing up the menu. The menu item will move this PlayQueueItem to the next-to-play position in the PlayQueue (that is, the index after the current playing song).

### Why? 
I use NewPipe often when commuting or at the gym and it's a wonderful application. One issue I've been running into, however, is organizing the play queue. Often I'll play a large public playlist with the intention of moving the songs I personally enjoy to the front. Furthermore, I'll also often enqueue songs not in the playlist to the queue. Currently this is a time-consuming task due to the time it takes to drag a song from the bottom to the top, not to mention bugs within the ItemTouchHelper that causes the drag to move way too fast after changing direction. By adding a menu option to move a PlayQueueItem to the next position in the PlayQueue, organizing songs can be more time efficient and less painful.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
@@ -32,6 +32,7 @@ import org.schabi.newpipe.fragments.OnScrollBelowItemsListener;
 import org.schabi.newpipe.local.dialog.PlaylistAppendDialog;
 import org.schabi.newpipe.player.event.PlayerEventListener;
 import org.schabi.newpipe.player.helper.PlaybackParameterDialog;
+import org.schabi.newpipe.player.playqueue.PlayQueue;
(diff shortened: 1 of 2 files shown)


Gymcap:
I kinda like this idea. Queues are often done very poorly so if this gets added it's important to make sure the logic of manipulating the queue is consistent. It took PowerAmp like 5 years to finally get right. I highly recommend studying their queue behaviour in preparation for this feature.

litetex:
What's the state of this pr?
Is there still progress?
There are a ton of merge conflicts...

Also there are no GitHub actions builds.

SameenAhnaf:
@litetex Isn't #6872 already merged? Should this PR still remain open?

litetex:
@SameenAhnaf 
Thank you for the hint!

I will close it then 😄

Issue #2358, opened 2019-05-23
NewPipe doesn't remember the video viewed  path when back button is clicked


fightthepower:
**How to reproduce**
1.  Open any video page in newpipe (v 0.16.1)

2.  From that video page select any  recommended  video and go to its page

3.  Repeat the step 2 as many times as you like

**Problem**
Lets say I go from a video page A to video E in this path A->B->C->D->E. When back button is clicked in Eth page I should have gone to Dth page and then back button gets to Cth page similar to this E->D->C->B->A. 

But in my newpipe when I click back button from Eth video page it goes directly to either trending or whats new page and skip all intermediate video pages. Similar thing exist if we go to a channel page also.

- [x ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ x] I checked if the issue/feature exists in the latest version.
- [ x] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

Stypox:
I think this will be fixed as soon as #2286 will be merged.

TobiGr:
There are two back buttons. When using the hardware back button, you should go back one page. But using the back button in the app bar should bring you back to the first level in this case the trending page/home.

yausername:
I too found this non intuitive.
How about using the home icon instead?
![sc2(1)](https://user-images.githubusercontent.com/5203007/58288591-b9c76480-7dd1-11e9-973d-0ab7ce71a5f6.jpg) ![sc1(2)](https://user-images.githubusercontent.com/5203007/58288606-bfbd4580-7dd1-11e9-918a-ea2be6dcd003.jpg)

fightthepower:
@yausername I agree it should have been the home icon, I thought that the back arrow is for going back.
@TobiGr I didn't think of using hardware back button , thanks for letting me know. That said I personally prefer not to use the hardware button and instead I want to use inbuilt app back button or swiping to go  back to page. Please consider this as a feature request for that.

cdwijs:
I confirm that the hardware back button works as expected, the arrow in the top-left goes back too far.

Tested with this version:
NewPipe 0.17.4 (latest release)
Android 6.0.1
CyanogenMod 13.0-20161121-NIGHTLY-ks01lte
Samsung Galaxy S4

opusforlife2:
[...]

PR #2430, opened 2019-06-28
add button to MainPlayer to go to Detail view

csicar:

## Changes
- Added a `<-` button to the top of the `MainVideoPlayer`

## Reason
The reason this button is useful is, that when opening videos via a link (and having selected "Open Video directly"), you can see the comments and related videos. 

## Example
example


app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
```diff
@@ -641,6 +648,9 @@ public void onClick(View v) {
             } else if (v.getId() == repeatButton.getId()) {
                 onRepeatClicked();
                 return;
+            } else if (v.getId() == showDetailButton.getId()){
```
Stypox: A space is missing between ) and {

app/src/main/res/layout/activity_main_player.xml
```diff
@@ -151,49 +151,61 @@
...
-                        android:clickable="true"
-                        android:focusable="true"
-                        tools:ignore="RtlHardcoded"
-                        tools:text="The Video Title LONG very LONG"/>
-
-                    <TextView
-                        android:id="@+id/channelTextView"
-                        android:layout_width="match_parent"
+                        android:layout_marginTop="5dp"
+                        android:background="@drawable/ic_expand_more_white_24dp" />
```
Stypox: The space between " and /> should be removed

csicar:
Requested changes have been implemented. Is there something else missing?

Stypox:
Please wait for a maintainer to review your changes. This could take a while since they are often busy in real life and there are more important and complex PRs on the way to be merged.

AdKiller:
Always wanted this feature. thank you very much for implementing it!

app/src/main/res/layout/activity_main_player.xml
```diff
@@ -151,49 +151,66 @@
...
+                            android:marqueeRepeatLimit="marquee_forever"
+                            android:scrollHorizontally="true"
+                            android:singleLine="true"
+                            android:textColor="@android:color/white"
+                            android:textSize="15sp"
+                            android:textStyle="bold"
+                            android:clickable="true"
+                            android:focusable="true"
+                            tools:ignore="RtlHardcoded"
+                            tools:text="The Video Title LONG very LONG"/>
```
TobiGr: What the hell is that? Is there any reason, why we put this here before? I guess, this is replaced by the correct title autmatically, but why do we use this as the default one 😂

csicar: I think, that `tools:text` is only used by Android Studio to display a value in the layout preview. During runtime, the value will never be used.

Phoenix616:
I feel like making the button look like a back button kinda breaks expectations: Back implies that you actually go back to the previous screen (like with the device back button), not that you always go to the info page.

I feel like a design that clearly shows the info page (e.g. an i in a circle) and a different placement would solve that.

Stypox:
@Phoenix616 you are right, an info button would be better. This expectation-breaking already exists with other back button: #2358. The proposed fix is basically the same as yours: https://github.com/TeamNewPipe/NewPipe/issues/2358#issuecomment-495395844.
@csicar would you mind changing the back button to some type of info button?

csicar:
The bahaviour oft the back butto is how Android UIs are intended to behave (as intended by the ui guidelines):
- the Back Button in the toolbar of an app should navigate hierachically upwards
- the android's system back button navigates chronologically backwards. 


Reference: https://material.io/design/navigation/understanding-navigation.html#reverse-navigation 

I personally perfer to think of the Button as navigating upwards, but I can understand the Argument. Maybe the  down arrow from YouTube's app could be a compromise?

If not, i'll be happy to implement to info button.

Stypox:
For me the back button is ok, then 😉

Phoenix616:
Interesting that they would suggest such a confusing way of handling things. (Duplicate back buttons with different functionality) I feel like the down arrow would be a lot better  and more intuitive. (Especially for people coming from the original app ;))

csicar:
Example of how it looks with the "expand" button:
2019-08-17-214332_398x582_scrot

[patch.txt](https://github.com/TeamNewPipe/NewPipe/files/3512393/patch.txt)

csicar:
@Phoenix616 I guess having very similar symbols for the different actions doesn't help :)

TobiGr:
I am completely fine with the behaviour and icons you used in the example video :+1:

Phoenix616:
I just realised that the downwards arrow causes another usability problem: The arrow at the right side looks the same but it is used to expand the view of available video options instead.

So a backwards arrow like the Android docs suggest might actually be the better option here :S (unless a different icon can e found for either going to the info view or expanding the option view I guess)

TobiGr:
@Stypox is there anything which prevents this PR from being merged?

Stypox:
@TobiGr no, this is ready, I forgot to approve changes at last ;-)

Stypox:
@csicar could you check the bug report from TobiGr?

Stypox:
Closing as this isn't needed anymore after 2907

PR #2434, opened 2019-06-29
Subscriptions shortcuts

nv95:
Adding desktop shortcuts for subscriptions and application shortcuts (on Android 7.1+)

| Screenshot_20190629_182253 | Screenshot_20190629_182417 |
|---|---|


TobiGr:
@nv95 Works good, I'll go through the code soon. Could you take a look at the conflict?
@Stypox changed the order of the actions in all menus and placed `share` at the end. We should do the same here. Maybe we need an offset variable (see very bad version below) or something different as workaround because the shortcut action would be in the middle and therefore cause some trouble with the index in the OnClickListener.
```JAVA
final ArrayList<String> commands = new ArrayList<>(3);
commands.add(context.getResources().getString(R.string.unsubscribe));
if (shortcutSupported) {
    commands.add(context.getResources().getString(R.string.create_shortcut));
}
commands.add(context.getResources().getString(R.string.share));

final DialogInterface.OnClickListener actions = (dialogInterface, i) -> {
    final int commandIndexOffset = shortcutSupported ? 0 : 1;
    if (i == 0) {
        deleteChannel(selectedItem);
    } else if (i + commandIndexOffset == 1) {
        ShortcutsHelper.pinShortcut(context, selectedItem);
    } else if (i + commandIndexOffset == 2) {
        shareChannel(selectedItem);
    }
};

**Comment by Stypox** [comment]
Could you provide an apk? ;-)

**Comment by nv95** [comment]
Release variant, signed by my key
https://www.dropbox.com/s/e0qxydnfk6q2jox/newpipe_shortcuts.apk?dl=0

File: app/src/main/java/org/schabi/newpipe/util/ShortcutsHelper.java [hunk 1]
```diff
@@ -0,0 +1,185 @@
...
+
+public final class ShortcutsHelper {
+
+	public static final String ACTION_OPEN_SHORTCUT = "org.schabi.newpipe.action.OPEN_SHORTCUT";
+
+	private ShortcutsHelper() {
+	}
+
+	@SuppressLint("CheckResult")
+	@SuppressWarnings("ResultOfMethodCallIgnored")
```
Stypox: Is it ok to ignore Disposables? Couldn't this lead to crashes (when an activity is closed but its disposables are not disposed) or leaks?

nv95: I think this is ok, because:
1. This operation is relatively short, so should not be any memory leaks
2. We want to complete this action even if activity was destroyed. Of course, this operation can be placed into separated Service, but, i think, it is too small

app/src/main/java/org/schabi/newpipe/util/ShortcutsHelper.java
```diff
@@ -0,0 +1,185 @@
...
+				.observeOn(AndroidSchedulers.mainThread())
+				.subscribe(builder -> {
+					final String id = getShortcutId(data);
+					final List<ShortcutInfo> shortcuts = manager.getDynamicShortcuts();
+					final int limit = manager.getMaxShortcutCountPerActivity();
+					for (int i = 0; i < shortcuts.size(); i++) {
+						final ShortcutInfo shortcut = shortcuts.get(i);
+						if (id.equals(shortcut.getId())) {
+							// if shortcut already exists - increase rank and update it
+							builder.setRank(shortcut.getRank() + 1);
```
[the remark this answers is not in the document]
nv95: Shortcuts will be sorted by popularity

nv95:
@mauriciocolli I think, use service for shortcuts is not a good idea.

litetex:
No progress since two years, no GitHub actions build and merge conflicts.

Issue #1467, opened 2018-06-10
Background video title not showing on lockscreen and a request 


chris47368:
Hey Newpipe team,well done implementing the media controls on the lock screen that i and others suggested not too long back.

There is an issue however on the lock screen media control where the title of the background video being played is not shown.You may or may not already be aware of this issue but fixing this issue for the next version release would be great :) but im still very happy and grateful about the huge amount of progress your team are making on this app.

I also have another request.Is it possible that while saving a video as a sound file that you can give the user a way to modify the song tags such as title and artist/album .Also allowing to save a sound file at a higher bit rate would be good too.

The app can also randomly crash in the background after a while,its a minor but annoying issue but i have sent the crash logs via email anyway and hopefully its fixed soon in the later versions.

Thanks again Newpipe development team!!!

Edit:
Sorry i forgot to mention this but could you also add support for time-stamped YouTube links so when you open the link into newpipe the video automatically starts on the time within the link? I forgot to mention this while writing this post and did not want to create a second post

Thanks again Newpipe Dev team!!!

theScrabi:
Hello thank you :D. The props for the medio controllger go to @karyogamy.

> I also have another request.Is it possible that while saving a video as a sound file that you can give the user a way to modify the song tags such as title and artist/album .Also allowing to save a sound file at a higher bit rate would be good too.

There where discussions about this already, but so far no code. However please open up a new issue for this.

karyogamy:
> There is an issue however on the lock screen media control where the title of the background video being played is not shown.

Are you referring to the popup _video_ player or the background _audio_ player? 

>  the title of the background video being played is not shown

Does it show any notification or is it just the title not showing? Can you please provide a screenshot if its just the title?

> The app can also randomly crash in the background after a while,its a minor but annoying issue but i have sent the crash logs via email anyway and hopefully its fixed soon in the later versions.

Can you please provide a log or screenshot of the error here? Can't seem to find any recent crash log coming from the background player. Are you on the latest version?

karyogamy:
Thanks for the clarification.

> This issue is with the Newpipe background player as shown below:

I'll have to look into this a little more, but this might have something to do with `MediaBrowser` not implemented at the moment. It's quite strange that the metadata is passed to android wear and auto but not to notification access like in your case.

> I just posted the error report below:

This should be fixed in #1392, though it might take some more time before it is tested and merged.

chris47368:
Thanks :)

[...]

Issue #228, opened 2016-03-08
[Feature request] "Expose" media data


utack:
Android seems to support exposing metadata of the stuff an app is playing, for generic Music widgets and other stuff (lockscreen background?)
Would be cool if you could add support that with the data from the "play in background" notification. Title as Music Track Title, Channel as Artist and the static video previews image as album cover.

theScrabi:
Yes, I already thought about that. I just don't know when I will stumble across it.

Mmedic23:
Not sure if commenting on a 2 year old post is frowned upon but.. Still waiting on this? Video thumbnail as lockscreen background seems like a pretty good feature for background playback mode.

theScrabi:
Yes this is a feature I would also celebrate deeply.

RuthlessRuler:
Any updates on this Issue?

Stypox:
@RuthlessRuler you can see more discussion in #2453, though there are no recent updates

wb9688:
For Android's lockscreen thingie, we need to use https://developer.android.com/reference/android/app/Notification.MediaStyle. I don't think that'd be hard to implement.

PR #2453, opened 2019-07-10
Update play queue metadata

### LINKED ISSUE #2240: Playing title's current time and length is not displayed on car bluetooth player
Hi,

When I connect my Galaxy S8 (Android Pie) to my car media system via bluetooth, the playing title's current time and track length are not displayed.

I'm using version v0.16.1.

With either Android Oreo or previous version of NewPipe, it was working fine.

m0n1ker:

This PR includes
* Bug fix for the duration metadata field. `item.getDuration()` returns the duration in seconds, while the metadata object expects this to be in milliseconds.
* Add the current track number and total number of tracks in the playlist to the metadata. Some Bluetooth players prefer this information be present.

These metadata changes have been tested using a Pioneer car stereo over Bluetooth.

Fixes #2240

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/playback/BasePlayerMediaSession.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playback/BasePlayerMediaSession.java
@@ -58,6 +58,7 @@ public class BasePlayerMediaSession implements MediaSessionCallback {
 
         // set additional metadata for A2DP/AVRCP
         Bundle additionalMetadata = new Bundle();
+        additionalMetadata.putString(MediaMetadataCompat.METADATA_KEY_TITLE, item.getTitle());
         additionalMetadata.putString(MediaMetadataCompat.METADATA_KEY_ARTIST, item.getUploader());
         additionalMetadata.putLong(MediaMetadataCompat.METADATA_KEY_DURATION, item.getDuration() * 1000);
         additionalMetadata.putLong(MediaMetadataCompat.METADATA_KEY_TRACK_NUMBER, index + 1);
--- a/app/src/main/java/org/schabi/newpipe/player/playback/BasePlayerMediaSession.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playback/BasePlayerMediaSession.java
@@ -60,6 +60,8 @@ public class BasePlayerMediaSession implements MediaSessionCallback {
         Bundle additionalMetadata = new Bundle();
         additionalMetadata.putString(MediaMetadataCompat.METADATA_KEY_ARTIST, item.getUploader());
         additionalMetadata.putLong(MediaMetadataCompat.METADATA_KEY_DURATION, item.getDuration() * 1000);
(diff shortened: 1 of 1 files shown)


Stypox:
Could this pr have anything to do with #1467 and #228? By reading the lines you edited it seems like those issues can be closed (after this pr is merged).

Redirion:
@Stypox the title and mediaid are set in the constructor a few lines above.
Sometimes its just a device issue.
I was first excited when I saw this PR as I was the one that introduced the "additionalMetadata" bundle but getDuration didn't work for me. 
But I guess it is also a device issue for me as my car (Kia Stinger 2018) does still not show the correct duration.

But I am happy that it works for you @m0n1ker 
I am just unable to report an improvement for me personally but at least there is no regression :)

m0n1ker:
@Stypox This _might_ help #1467, but I kind of doubt it. I'm running android 9 on a Galaxy S9 and it does not show the media controls as described in the issue. I only see the notification widget, which is working for me, so I'm not able to test this specific issue on my device.

As far as #228, again I think it will make the exposed metadata more reliable but I'm not sure this really resolves the feature request as stated. I looked at adding the album art to the metadata, but it was causing some other issues that I haven't been able to solve. If I can get that working I can submit another PR for it.

@Redirion I'm surprised this didn't change anything in your car. It's new enough it should be able to correctly handle metadata, but bluetooth receivers are weird sometimes. I've verified the metadata being sent with the `bluetooth-player` utility that is a part of Linux's "bluez" bluetooth stack. It prints out metadata and playback status messages as they are sent, and I'm seeing the proper data and playback states being sent now.

I agree that adding `METADATA_KEY_TITLE` to the additional metadata would be a good idea. It doesn't hurt anything, and if it makes it more widely compatible with bluetooth receivers that's an easy win. I'll push another commit after this comment.

m0n1ker:
As an additional note, I think the issue with #228 has more to do with the fact this app is implementing `MediaDescriptionCompat` but not `MediaMetadataCompat`. From what I can tell, the description object is meant for lightweight display-only purposes such as showing items in the play queue. The full `MediaMetadataCompat` object is the preferred one for showing details about an individual item in the queue, such as the currently playing track, and adding support for this may actually help both the bluetooth car audio issues and the issues associated with #228. This is something I'd like to explore more when I have some time.

TobiGr:
@m0n1ker Thanks! The code looks good.

> As an additional note, I think the issue with #228 has more to do with the fact this app is implementing MediaDescriptionCompat but not MediaMetadataCompat. From what I can tell, the description object is meant for lightweight display-only purposes such as showing items in the play queue. The full MediaMetadataCompat object is the preferred one for showing details about an individual item in the queue, such as the currently playing track, and adding support for this may actually help both the bluetooth car audio issues and the issues associated with #228. This is something I'd like to explore more when I have some time.

Yes, that could solve the problems. I am glad that you want to take a look at it.

Is someone capable of testing this with a different car or device? I don't own a car which supports Bluetooth. If not, this shouldn't be a problem, because the changes don't  look like they can break something.

PR #2474, opened 2019-07-23
add support for jedentageinset.de

theScrabi:
This will add support for [jedentageinset.de](https://jedentageinset.de) to the extractor.
jedentag ein set is a list of techno sets which is maintained by some fine duds from my city. Since the do everything based on soundcloud i thought it was a cool idea to add it as available kiosk to the soundcloud service.

This PR relays on a change in the [extractor](https://github.com/TeamNewPipe/NewPipeExtractor/pull/175).

Screenshot_20190723-143903_NewPipe_Debug Screenshot_20190723-143913_NewPipe_Debug
Screenshot_20190723-143950_NewPipe_Debug


TobiGr:
It's a nice idea to give attention to non-famous artists or collections. But before we can merge this, we must clarify some things.

When we put certain channels in a featured position, we should discuss and write down why our intention behind this (e.g. in a blog post). **What are the criteria for a channel to be featured?** Who decides which channels are featured (and more importantly which not).
This can easily look like or be misinterpreted as an advertisement and therefore create confusion about the independence of this project. People even assumed that we had included [advertising in our screenshots](https://forum.f-droid.org/t/advertisement-in-f-droid/2350) (spoiler: we did not).

A more complex, but maybe catchy idea would be a (monthly or so) collection of notable collection/channels. Maybe curated by the community. Nonetheless, we must clarify criteria for an inclusion first.

Apart from that, we should allow the user to remove those featured channels from the side navigation.

@Stypox @mauriciocolli @kapodamy @karyogamy @yausername @nv95 @TheAssassin  What do you generally think about this topic?

gkeegan:
@TobiGr I know you didn't ping me for this, but I think I have a good idea for the below suggestion.

>Apart from that, we should allow the user to remove those featured channels from the side navigation.

It would be good to have a screen that upon first launch of the app to determine which sites you want to be visible, with toggles and a description of the site.

This could be accessed afterwards through the settings.

TobiGr:
@gkeegan Oh, I am sorry, forgot to ping you.

> It would be good to have a screen that upon first launch of the app allows the user to determine which sites you want to be visible, with toggles and a description of the site.

I wanted to do something similar between the semesters. A small config screen at first launch which helps the user to get the most important settings right. I could include a page for this purpose.

gkeegan:
>Oh, I am sorry, forgot to ping you.

Don't worry about it haha.

And yeah I think an issue could be made for this to discuss further what options to include in such a config screen.

theScrabi:
@TobiGr oh i didnt think of that. It was more of a "hey I'd like that function that would be cool", and less a " hey i like these guys i want to help them" in my head. But yea in some way you are right to.

However their side is not a regular channel, in fact you have to scrape the data from an external site, that just contains links to soundcloud. So the best way (and in my understanding) the only way is to implement it as a kiosk, as they allow that. The thing about kiosks is that they are not restricted to the side itself. For example the youtube service could add the kiosk pages of invidio.us or hooktube.

Finally i understand your concern with toggls, but from a ui perspective having to go to the settigs to enable/disable one of three servives seems to complicated for now. I would say it would be ok if we supported 40 or more.

So long story short we should rather leave this feature out before making it on/off switchable.

TheAssassin:
I guess a "real" rule set or so isn't strictly necessary, but I wouldn't make this a first-level citizen. Can't we put that feature into a submenu that might aggregate more of these "curated lists" of stuff. We could add such a point to the menu, and make this "jedentageinset" thingy an entry inside there. We could then establish a policy that suggestions are welcome and will be incorporated after a review. Guidelines for such a review should, as @TobiGr suggests, be written down at first, to make this process transparent. Documentation is key for transparency. We don't have to publish a log or so of discussions of every such list then, in case there's disagreements we can justify our decisions by referring to those guidelines.

By the way, how about privacy there? The most imminent problem IMO is data protection. Random third-party sites might be pinged, and that might be a problem, especially since none of us can constantly validate the data protection statements of such services. The guidelines I mentioned should contain some rules for that as well.

theScrabi:
Site will not be pinged until you use it.

pihug12:
I think it shouldn't be a built-in feature since it "_is not a regular channel, in fact you have to scrape the data from an external site, that just contains links to soundcloud_". It should be provided by a "plug-in/add-on system" that a user could opt in.

**Idea:** server-side, you could expose a JSON that describes all these "not regular channels", and the users could pick (in the settings) the one they are interested. The advantage is you can update the JSON without releasing a new version of the app (kinda what [ClearURLs](https://addons.mozilla.org/en-US/firefox/addon/clearurls/) does: https://gitlab.com/KevinRoebert/ClearUrls/blob/master/core_js/storage.js#L160).

TheAssassin:
> Idea: server-side, you could expose a JSON that describes all these "not regular channels", and the users could pick (in the settings) the one they are interested. The advantage is you can update the JSON without releasing a new version of the app (kinda what ClearURLs does: 

Dependencies on servers are always problematic, since they then require an SLA. Not a fan of this.

wb9688:
Closing due to jedentageinset not being in the extractor anymore and @theScrabi making a separate app for jedentageinset

PR #2476, opened 2019-07-25
Magnetic actionBar for popup player

ronidee:
- [y] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Make the action bar magnetic. The popup player will snap onto it automatically when in range.
The range is 10% of the action bar's height.

Example video
magnetic
Notice: the stuttering before it snaps is not actually there. It's only in the gif. 
~~change behavior of popup player so it will snap onto the actionbar/toolbar on longpress. Also added switch in settings, default: false.
Sometimes popup player jumps a bit after drag&drop, but i can't reproduce that bug consistently and it also happens sometimes in the official app without my modificaiton so I'm not sure if it's also my code that's causing this or not.~~


gkeegan:
Does the jumping still occur when you are in apps other than NewPipe? So say I was on the home screen and long pressed would it go to that position and leave space between the notification bar and the video? Perhaps a long press when not in the NewPipe app could cause the video to jump to the very highest it can go.

Also, there are a few missing spaces in the strings.xml file.

Overall, great work.

ronidee:
Yeah, it jumps to the same position. I first tried to get the toolbar height of each currently opened app, but that didn't work (don't remember why, it's some months ago) and now I just read the systems default actionbar height value (see the method getActionbarHeight or so).

I left it with that little space (when on home screen e.g.) since it's easy to position the player manually at the very top. And it's supposed to snap below the bar on other apps do. Is there a way to detect if the honescreen is open? 

Maybe the long press thing is bad after all and it should use a magnetic actionbar, so that it snaps on to that while dragging it near to it (maybe 5dp or so)

wb9688:
Closing in favor of #2907

PR #2517, opened 2019-08-14
List slowdown fix

### LINKED ISSUE #2507: Reduced performance caused by playback position in list views
Just completely freezes..

Stypox:

This pr changes the way playback state is loaded into lists. Instead of loading all of the states before the list of items is shown, now every item loads its state only when it needs to (i.e. when it becomes visible).
I also did some cleanup of the files I edited.

Testing apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/3501075/app-debug.zip)
Fixes #2507


theScrabi:
Works :)

PR #2518, opened 2019-08-14
Customize "start here" actions in playlist item views

### LINKED ISSUE #2510: Unable to play YouTube playlist from specified track
In last v0.17 there is no option to start playing YouTube playlist from specified track (in background or even in foreground).

Previously I used v0.16.2 and I can open YouTube playlist, long tap on any track in this playlist and start playing my YouTube playlist from this track in backgroud.

Now, when I do the same, the selected track also starts to play, but this time it is enqueued into local playlist alone, losing any relation to the YouTube playlist, from which it was started. And next track in the local queue is not my playlist's next track, but randomly suggested track.

So, it works like if I just enqueued a single track from my playlist, while previously it respected my playlist and played all tracks from it one by one (starting from the selected track). Actually, previously it enqueued all tracks from YouTube playlist (starting from currently selected track) into local queue. So they played in order one by one and looked like my YouTube playlist is playing (while actually it was still a local queue). It was very convinient.

Now the only way to do the same is to manually add all tracks from YouTube playlist into local queue (there is a button for it) and then go to the queue, scroll down to the specified track and tap on it to play from this position. Much more steps involved in the process.

Maybe it is just me, who mainly playing my youtube playlists from the middle. But it is really main mode how I use this application and it perfectly worked before v0.17.

Stypox:

Now those actions start playing the whole playlist from the chosen stream, instead of playing only the chosen stream.

Testing apk: [app-debug_1.zip](https://github.com/TeamNewPipe/NewPipe/files/3501292/app-debug_1.zip)
Closes #2510, see https://github.com/TeamNewPipe/NewPipe/pull/2368#issuecomment-517925997


geekyjazzy:
Hi, i have the latest stable version of the app (0.17.1)and i still dont see the "start playing here" option on my playlists. Using a samsung a8, android pie

Stypox:
Yes, that was not added. See #2583 for the plan and send some feedback there, please :-)

PR #2527, opened 2019-08-17
Fix ReCaptchaActivity

### LINKED ISSUE #2484: ui crashes when google serves a captcha page
- [yes] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
--> I forgot to capture the crash report, error occurs randomly
- [yes] I checked if the issue/feature exists in the latest version. 16.2
- [ ] I did use the [incredible bugreport to markdown converter] (https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports. (next time)

I'm using orbot. Once in a while newpipe crashes when loading a video. By following the share link to external browser I just found out that google asks to solve a captcha because of high site traffic (last tor exit node heavily in use?).
If it happens again I will post the crash report (probably useless) but also the served html (for the extractor to learn to recognize the code).

Stypox:
- [y] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

This pr fixes two problems with the ReCaptchaActivity:
- it crashes on activity startup due to a hidden EditText, see #2484
- it always shows youtube's main page ("https://www.youtube.com"), even though the ReCaptchaException was thrown while loading a different page. The main page sometimes contains the ReCaptcha (so there is no problem), but other times it does not since YouTube only blocks a specific page. The specific page that requested a ReCaptcha challenge is now loaded instead, so that the ReCaptcha is always shown.

Fixes #2484
Corresponding extractor pr: TeamNewPipe/NewPipeExtractor#186
Testing apk: [app-debug_2.zip](https://github.com/TeamNewPipe/NewPipe/files/3511787/app-debug_2.zip)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -57,7 +57,7 @@ dependencies {
         exclude module: 'support-annotations'
     })
 
-    implementation 'com.github.TeamNewPipe:NewPipeExtractor:5f65788a2f89e'
+    implementation 'com.github.Stypox:NewPipeExtractor:06689a2f27edfe83bd4605a1cceed86c06a5ebf8'
 
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.mockito:mockito-core:2.23.0'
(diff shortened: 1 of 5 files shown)


blacklight447-ptio:
I'm still having crashes when recaptcha appears ;-;

Stypox:
Are you able to get a crash report?

blacklight447-ptio:
I will the next time it crashes (I only get captcha's on occasion)

PR #2534, opened 2019-08-18
Fix ":00" being erroneously recognized as a timestamp in comments

nyanpasu64:
On Youtube's web interface, in comments (and descriptions), :00 is not recognized as a timestamp in comments, while 0:00 is.

This does not solve the minor issue that 000:000 is recognized as a comment in NewPipe, but not YouTube.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
@@ -34,7 +34,7 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
     private String commentText;
     private String streamUrl;
 
-    private static final Pattern pattern = Pattern.compile("(\\d+:)?(\\d+)?:(\\d+)");
+    private static final Pattern pattern = Pattern.compile("(\\d+:)?(\\d+):(\\d+)");
 
     private final Linkify.TransformFilter timestampLink = new Linkify.TransformFilter() {
         @Override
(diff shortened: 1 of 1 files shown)


connectety:
try `([0-5]\d:)?([0-5]?\d):([0-5]\d)` it filters out a lot.
tests for the regex are here: https://regex101.com/r/0BcgqH/3/tests
(the regex includes two more characters to only find full matches for tests)

jimbo1qaz:
https://www.youtube.com/watch?v=4cccfDXu1vA&lc=UgwDaGGqe5oCizvtwXJ4AaABAg

- 0:00 is a valid timestamp recognized by Youtube descriptions and comments.
- 0:00:00 is also a valid timestamp.
- Youtube does not recognize 1:1:1 as a timestamp. But 1:01:01 and 01:01:01 are valid.
- Youtube does not recognize 000:00 as a timestamp. But 0:000 is recognized as `0:00` followed by unclickable 0. 😦 
- Youtube does not allow minutes or seconds to be 60 or greater, but hours is OK.
- 100:00:00 is recognized as a timestamp in descriptions, but not comments.

connectety:
100:00:00 still gets falsely matched
fix for that: `(\\d{1,2}:)?([0-5]?\\d):([0-5]\\d)`

jimbo1qaz:
https://www.youtube.com/watch?v=4cccfDXu1vA&lc=Ugzgqjp6Z1rYIzENPkd4AaABAg

- 0:0:0 is not clickable.
- 0:00:0 has clickable 0:00 followed by unclickable :0.
- 0:0:00 is clickable.

Is it worth actually finding the JS code in Youtube classic or polymer, which parses links?

connectety:
Too my understanding YouTube only looks at the front of a timestamps and only at ":" and numbers.
I have taken so long to answer, because the regex to provide this behavior is quiet complex: `(?:^|[^\d:])(\d{1,2}:)?([0-5]?\d):([0-5]\d)`.
The same regex but with an extra group, which does not get used (every group number needs to be incremented by once) `(^|[^\d:])(\d{1,2}:)?([0-5]?\d):([0-5]\d)`

wb9688:
Imho this should be handled by the extractor and we should change the comments to contain a [Description](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/stream/Description.java) instead of just a string.

Stypox:
Yeah, it would be better to move this to the extractor

PR #2535, opened 2019-08-18
Add missing dependencies inherited from NewPipeExtractor into NewPipe

jimbo1qaz:
- NewPipe uses nanojson and jsoup, despite not declaring them.
- NewPipe depends on NewPipeExtractor, which depends on nanojson and jsoup.
- Gradle lets NewPipe use nanojson and jsoup, but this breaks when I use:

```gradle
includeBuild('../NewPipeExtractor') {
    dependencySubstitution {
        substitute module('com.github.TeamNewPipe:NewPipeExtractor') with project(':extractor')
    }
}
```

- Is this the optimal way to build both NewPipe and NewPipeExtractor from source into a single .apk?
- Was there a deliberate design decision to declare the version requirements in NewPipeExtractor and not NewPipe?

Some people do the opposite, and declare "fuzzy dependency versions" in libraries like NewPipeExtractor, and "exact pinned dependency versions, including transitive dependencies" in apps like NewPipe. Maybe someone understands Gradle better than I do. https://docs.gradle.org/current/userguide/managing_transitive_dependencies.html


TobiGr:
>  sorry if I keep pinging you

@Stypox No problem. That's the easiest way to get attention :laughing:

PR #2536, opened 2019-08-18
Fix timestamp links in video descriptions

jimbo1qaz:
Video description timestamp links have link text which is not a URL. I submitted a PR adding the ability to properly parse timestamp links into `<a href="...&t=0">00:00</a>` at https://github.com/TeamNewPipe/NewPipeExtractor/pull/188. However, the NewPipe front-end `videoDescriptionView` erases all Spanned links if their text is not a URL.

`videoDescriptionView.setAutoLinkMask(Linkify.WEB_URLS);` is intended to turn textual URLs into links. For some reason, on my Moto G5 Plus running Android 7.0, it causes `Spanned` hyperlinks to disappear if the text is not a URL. This PR removes that function call.

Will this cause non-timestamp URLs to break? I've tested on https://www.youtube.com/watch?v=801uz0f_UAE and the URLs in the video description still work. I think all URLs delivered to the web interface are already found in HTML and scraped and encoded into `Spanned`, so this function call is unnecessary. But it's possible some URLs fail to survive the pipeline but are regenerated via autolink, and this PR will make them disappear. If so, that's probably a bug hidden by autolink.

I haven't changed `:app/build.gradle` to point to a different `implementation 'com.github.teamnewpipe:NewPipeExtractor:...'` revision yet, since I did all local testing using:

```gradle
includeBuild('../NewPipeExtractor') {
    dependencySubstitution {
        substitute module('com.github.TeamNewPipe:NewPipeExtractor') with project(':extractor')
    }
}
```

Depends on https://github.com/TeamNewPipe/NewPipeExtractor/pull/188.


B0pol:
You should close this. It works, but only for YouTube. It breaks all description links for other services.

I'll will implement it with service checking with the future pr for https://github.com/TeamNewPipe/NewPipeExtractor/pull/239.

Issue #2552, opened 2019-08-23
Change the icon of subscription update


TotalCaesar659:
There are 2 actions with the same icon in NewPipe:
1) going to outer RSS feed has real RSS logo as an icon (https://en.wikipedia.org/wiki/RSS)
![1](https://user-images.githubusercontent.com/14265316/63624537-60a3e280-c605-11e9-83c0-7400343b2faf.png)
2) subscription update has the same icon but opens new video list in NewPipe
![2](https://user-images.githubusercontent.com/14265316/63624544-68fc1d80-c605-11e9-9244-356f6790bde3.png)
I think it would be better to change second icon from real RSS logo to something more abstract (like this one), it should me more comfortable.
![3](https://user-images.githubusercontent.com/14265316/63624481-21759180-c605-11e9-9723-99f528bd9d03.png)

PR #2561, opened 2019-08-26
Change subscription update icon

TotalCaesar659:
RSS logo is using now for subscription RSS feed only (https://github.com/TeamNewPipe/NewPipe/issues/2552).


nv95:
Why not vector drawable?

TotalCaesar659:
It's for uniformity, other icons are [drawn](https://github.com/TeamNewPipe/NewPipe/tree/dev/app/src/main/res/drawable-xxxhdpi) in PNG too.

Stypox:
I think the long-term plan is to convert all pngs to vector drawables, so I don't think it would be a problem to have an inconsistent vector drawable.

TotalCaesar659:
It would be better to merge it now and let vectorizing for a future. I'm not good in this, maybe someone else will do it.

justanidea:
You may be interested in #1232 in some points.

Stypox:
@TotalCaesar659 could you please provide a screenshot?

TotalCaesar659:
Yep.

wb9688:
I'm not sure whether I prefer the update icon. @TobiGr: What do you think?

If we're going to change it, you should use vector drawables though.

gkeegan:
Changing the what's new logo to a loading spinner like that seems like it would update, not take you to a page where you can update your feeds.

TobiGr:
Thank you for the effort. Sorry for getting back to you that late. 
I agree with the others. Apart from that, subscription handling has changed.

PR #2569, opened 2019-08-27
Black navigation bar for black theme

### LINKED ISSUE #1494: Set color of Navigation bar in the bottom
My Samsung A8 has got a white bottom navigation bar (this one with back, home and multitab) and its not very nice next to the dark theme of NewPipe

Just set it to black or something like that

Thank you for this software

marcovr:
As discussed in https://github.com/TeamNewPipe/NewPipe/issues/1494, this changes the software navigation bar to black with light buttons if all of the following points are fulfilled:
1. The black theme is enabled (maybe this should apply to the dark theme as well?)
2. The device supports setting a custom navigation bar color (API Level 21).
3. The device supports setting the navigation buttons color to dark mode (API Level 27).

Fixes #1494

### DIFF
--- a/app/src/main/res/values-v27/styles.xml
+++ b/app/src/main/res/values-v27/styles.xml
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
-    <style name="Base.V21.DarkTheme" parent="Base.V7.DarkTheme">
+    <style name="Base.V27.DarkTheme" parent="Base.V7.DarkTheme">
         <item name="android:navigationBarColor">@color/dark_background_color</item>
         <item name="android:windowLightNavigationBar">false</item>
     </style>
-    <style name="Base.DarkTheme" parent="Base.V21.DarkTheme" />
+    <style name="Base.DarkTheme" parent="Base.V27.DarkTheme" />
 
-    <style name="Base.V21.BlackTheme" parent="Base.V7.BlackTheme">
+    <style name="Base.V27.BlackTheme" parent="Base.V7.BlackTheme">
         <item name="android:navigationBarColor">@android:color/black</item>
     </style>
-    <style name="Base.BlackTheme" parent="Base.V21.BlackTheme" />
+    <style name="Base.BlackTheme" parent="Base.V27.BlackTheme" />
 </resources>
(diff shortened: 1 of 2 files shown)


Redirion:
I would like to propose some (small) changes. Please use the approach described in the answers to this stackoverflow Question: https://stackoverflow.com/questions/32725616/repeating-style-in-v19-v21

It avoids duplicated code and makes the relevant changes easier to see.

Additionally it would be great to have this done similarly to the DarkTheme, not by adding the same black color, but dark themed color to the navigation bar. See styles for the DarkTheme , I think dark_background_color should be fitting.

marcovr:
Alright, I'll check that out

marcovr:
Better?

Just tested it again on the emulator (Android 7 and 9) and on a Samsung device (Android 9), everything seems fine.

This is how the DarkTheme now looks:

Redirion:
why the Base.BlackTheme and Base.DarkTheme though?

why not just add 

<style name="Base.V7.DarkTheme" parent="DarkTheme" /> 

in values/styles.xml

and then

<style name="Base.V27.DarkTheme" parent="Base.V7.DarkTheme">

<style name="DarkTheme" parent="Base.V27.DarkTheme" />

in values-v27/styles.xml?

marcovr:
I agree that it's a bit convoluted (it took me a moment to understand it) but it does make sense:
- The `DarkTheme` contains the actual theme items/values
- The `Base.DarkTheme` is a fixed reference for `DarkTheme` and serves as a pointer to the highest available `Base.XX.DarkTheme`
- `Base.XX.DarkTheme` contains version specific additions and links to the next lower available `Base.XX.DarkTheme`, until at last the lowest one links to the parent style.

Redirion:
okay, then just change the V21 to V27 and its perfect! :)

marcovr:
Oh, how did I miss that? You're right

Redirion:
Thank you! 

@TobiGr 
This would make a nice little addition to 0.17.4

TobiGr:
Sorry for the late reply. Can someone please sum up what this PR changes? A before/after screenshot would be lovely :) I tested this with an Android 9 emulator and didn't see any difference.

marcovr:
See [this comment for before/after](https://github.com/TeamNewPipe/NewPipe/issues/1494#issuecomment-522322635).
In short: it changes the navigation bar color to dark/black in order to match the theme. This only affects phones which (for some reason) use a light navigation bar per default, which looks terrible together with the black theme.

TobiGr:
Thanks! Can you please rebase the commits?

marcovr:
Alright, I'm on it

Issue #2581, opened 2019-08-30
Unable to download files to a dot directory


sohmc:
- [X] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [X] I checked if the issue/feature exists in the latest version.
- [ ] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

I just downloaded 0.17.1 and noticed that this version no longer allows me to save into a dot directory (e.g. `/storage/emulated/0/Movies/.NewPipe`).  I usually save files from YouTube into a dot folder to keep Samsung's photo gallery from picking it up.  I used to be able to type in the directory and location but now it's only showing me a directory navigator and dot directories are hidden.

kapodamy:
The problem is in [NoNonsense-FilePicker](https://github.com/spacecowboy/NoNonsense-FilePicker) library

PR #2591, opened 2019-09-01
show hidden directories in the file picker

kapodamy:
fixes https://github.com/TeamNewPipe/NewPipe/issues/2581 but allows NewPipe pick and navigate on hidden directories


sohmc:
Awesome!  Just tested this out!  Thanks for the quick fix!

PR #2631, opened 2019-09-13
Musicplayerseek fix 

### LINKED ISSUE #2606: Background play mode: Playlist screen: Player in the bottom: Make timestamp label's width dynamic
In live streams, timestamp label's width needs to be dynamic — otherwise it will expand vertically.

IMG_20190908_131651_359

sherlockbeard:

Fixes #2606 
ui before change while running live video
Capturebefore

ui after change in while running live video
Captureafterlive

ui after change in normal video 
Captureafternonlive

@Stypox


app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
```diff
@@ -570,10 +570,10 @@ public void onProgressUpdate(int currentProgress, int duration, int bufferPercen
         }
 
         // this will make shure progressCurrentTime has the same width as progressEndTime
-        final ViewGroup.LayoutParams endTimeParams = progressEndTime.getLayoutParams();
+       /* final ViewGroup.LayoutParams endTimeParams = progressEndTime.getLayoutParams();
```
TobiGr: The idea behind this is to center the seekbar. After your changes, the seekbar will not be centered. Maybe it is better to detect which of both fields is wider and enlarge the other one.

TobiGr: Do you mean [this](https://developer.android.com/guide/topics/ui/look-and-feel/autosizing-textview)? That's only supported on Android 8 and newer.

progressCurrentTime and progressEndTime need to have the same size which centers the seekbar automatically. 
What I meant, is this:
``` JAVA
if (currentTimeParams.getWidth() > progressEndTime.getWidth()) {
     progressEndTime.width = currentTimeParams.getWidth();
} else {
     currentTimeParams.width = progressEndTime.getWidth();
}
```
But I guess there is a proper way to do this by using a linear layout/constaint layout and layout weight.

@Stypox How would you do this?

Stypox: There is one thing to note: theoretically currentTime width is never bigger than endTime, so the first 'if' in the code above could be removed.

I think it would be possible to do something using [a table layout](https://stackoverflow.com/questions/6647177/set-equal-width-of-columns-in-table-layout-in-android), but I'm not sure.

TobiGr: > theoretically currentTime width is never bigger than endTime's, 
@Stypox 
It is only bigger with normal streams. When listening to a livestream, the right view just says _LIVE_ (see the screenshots above). For this reason the view for the current position is that small atm.

Stypox:
@sherlockbeard could you address the requested suggestions? Thank you ;-)

Issue #2406, opened 2019-06-17
App crashes whenever I select MPEG-4 (480p) quality for any video download


chrisbarth01:
### Expected behavior 
I expected the app to download the video in the quality I selected.

---

### Actual behavior 
Whenever I select MPEG-4 (480p) quality for any video download, the video starts downloading then the app crashes.

---

### How to reproduce 
- Download, install and launch the latest version of the app
- Click on a video
- Click on the download icon
- Select the video quality to MPEG-4 (480p)
- Click on the OK button (download)
- The video starts downloading and the app crashes in the process.

---

Device: Gionee f100s 
Android version: v6.0
App version: v0.16.2

---

### Recording of the bug
![20190617_184924](https://user-images.githubusercontent.com/49325171/59628376-70e89e80-9138-11e9-94f6-9fe46bed50da.gif)

---

#### Logcat
```06-17 18:57:26.587 348-8577/? D/DrmMtkUtil/WhiteList: IsDrmTrustedClient: compare [com.android.gallery3d:widgetservice] with [org.schabi.newpipelegacy].
06-17 18:57:26.622 32188-32188/? I/AEE/AED: check process 31678 name:i.newpipelegacy
06-17 18:57:26.624 32188-32188/? I/AEE/AED: [OnPurpose Redunant in void preset_info(aed_report_record*, int, int)] pid: 31678, tid: 32186, name: VideoTrackEncod  >>> org.schabi.newpipelegacy <<<
06-17 18:57:26.675 32188-32188/? I/AEE/AED: pid: 31678, tid: 32186, name: VideoTrackEncod  >>> org.schabi.newpipelegacy <<<
06-17 18:57:27.702 899-32189/? W/ActivityManager:   Force finishing activity org.schabi.newpipelegacy/.MainActivity
06-17 18:57:27.715 899-958/? I/WindowManager: Focus moving from Window{b8cd646 u0 org.schabi.newpipelegacy/org.schabi.newpipelegacy.MainActivity} to null
06-17 18:57:27.715 899-958/? I/WindowManager: Losing focus: Window{b8cd646 u0 org.schabi.newpipelegacy/org.schabi.newpipelegacy.MainActivity}
06-17 18:57:27.745 899-32190/? D/AES:     process : org.schabi.newpipelegacy
06-17 18:57:27.745 899-32190/? D/AES:      module : org.schabi.newpipelegacy v20 (0.16.2)
[...]

PR #2646, opened 2019-09-18
check if last byte in the block is downloaded

kapodamy:
fixes a deadlock in `DownloadRunnable` because youtube is sending in some files one byte **alone** ~~after downloading 26MiB exactly, why... i dont know.~~
possible fix for #2406

- [ ✔ ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

### DIFF
--- a/app/src/main/java/us/shandian/giga/get/DownloadRunnable.java
+++ b/app/src/main/java/us/shandian/giga/get/DownloadRunnable.java
@@ -117,6 +117,8 @@ public class DownloadRunnable extends Thread {
                     byte[] buf = new byte[DownloadMission.BUFFER_SIZE];
                     int len;
 
+                    // use alwways start <= end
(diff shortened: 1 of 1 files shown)


app/src/main/java/us/shandian/giga/get/DownloadRunnable.java
```diff
@@ -117,7 +117,9 @@ public void run() {
                     byte[] buf = new byte[DownloadMission.BUFFER_SIZE];
                     int len;
 
-                    while (start < end && mMission.running && (len = is.read(buf, 0, buf.length)) != -1) {
+                    // use alwways start <= end
```
TobiGr: > alwways

Small typo.

PR #2655, opened 2019-09-22
Improve Accessebility for blind people

theScrabi:
I have worked with @fcnjd on improving the accessibility of NewPipe for blind people. I learned that there are still a lot of pit holes, however we already took on some very annoying things. Mostly they consisted of wrong or missing content descriptions.

Improvements:
- change video preveiw thumbnail text to "play video"
- add tab content description in main page
- add content description and service announcement for the toggle service button


opusforlife2:
Is there currently a way to play audio only by default? Something like a toggle that would remove all video options so that only audio ones remain.

theScrabi:
@fcnjd told me it was not that importand as there is already an audio only function, however one could use the video function to still be able to show videos to others.

Gymcap:
I think if the background player view was given a bit more attention and brought forward a bit, it would meet a lot of these "audio only" needs.

theScrabi:
Yea you are right. This should definitely be done soon.

Stypox:
@TobiGr this can be merged, too ;-)

PR #2666, opened 2019-09-27
Add preference option for unlimited search history

ergor:

This branch adds an option to turn on unlimited search history.

Current behavior: search history is capped at 25 entries when opened, and 3 entries when typing.
New behavior, preference is on: no caps when opening search and no caps when typing.
New behavior, preference is off: as current behavior.
Also, when turning off search history altogether, the new option is disabled.

I have tested and confirmed it works as intended on a virtual device and on my phone.

Screenshot_1569616343

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -90,6 +90,8 @@
     <string name="show_search_suggestions_summary">Show suggestions when searching</string>
     <string name="enable_search_history_title">Search history</string>
     <string name="enable_search_history_summary">Store search queries locally</string>
+    <string name="unlimited_search_history_title">Unlimited search history</string>
+    <string name="unlimited_search_history_summary">Unlimited search history entries</string>
     <string name="enable_watch_history_title">Watch history</string>
     <string name="enable_playback_resume_title">Resume playback</string>
     <string name="enable_playback_resume_summary">Restore last playback position</string>
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
@@ -128,6 +128,7 @@ public class SearchFragment
     private String nextPageUrl;
     private String contentCountry;
     private boolean isSuggestionsEnabled = true;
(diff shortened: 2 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -90,6 +90,8 @@
     <string name="show_search_suggestions_summary">Show suggestions when searching</string>
     <string name="enable_search_history_title">Search history</string>
     <string name="enable_search_history_summary">Store search queries locally</string>
+    <string name="unlimited_search_history_title">Unlimited search history</string>
+    <string name="unlimited_search_history_summary">Unlimited search history entries</string>
```
Stypox: A better suited description would probably be "Show as many search suggestions as possible from history"

TobiGr: Does this show all suggestions or only the related ones? Maybe "Display all (related) entries from search history" does also fit.

TobiGr:
> But is it ok to add another option to settings? @TobiGr

That's what I was thinking about, too. I do not really see a use-case for this option. How many people want and will use that addition? @theScrabi what's your opinion?

ergor:
My use case is, because I use NewPipe disconnected (not logged in), I get no subscriber feed etc. This way I don't have to involve myself with google. I use my local search history to check up on channels I like, and I rely on it to remember channels as well as music. But, because the local history is limited to the last 25 searches, channels and music I haven't checked out recently just vanishes into the ether.

I wouldn't mind expanding the tweakability of this feature. Also, I wonder: why was the limit of 25 chosen in the first place? Because of performance limitations, or simply preference?

gkeegan:
"Subscribing" to channels on NewPipe doesn't tie you to Google or YouTube accounts at all. It's essentially a browser bookmark.

ergor:
What if instead of this being a new option, just be the way search history works?

gkeegan:
You can make your own playlists in NewPipe. Go to a video, press "Add To", then press " New Playlist" to make a new one. Then you can add to that playlist by doing "Add To" and selecting it. 

Search shouldn't be used as a replacement for playlists, and should be limited.

litetex:
No progress since 2 years, no GitHub actions build and merge conflicts.
Also no feedback to our reviews.

Issue #2278, opened 2019-04-10
Possible feature request: Track My Pipe


ale5000-git:
An idea just come to my mind.
Beside personal preference, it could be useful if someone created an app called for "Track My Pipe" (completely separated from NewPipe) implemented similarly to the Google Contacts Sync Adapter but that periodically sync the NewPipe subscription and other things to the Google/YouTube account for the ones that need it.

chopin2712:
Yes ! I want to use New Pipe for replacement to YouTube but I don't have my 200 subscriptions to add one by one. Great Idea !

Stypox:
@chopin2712 you can export your subscriptions from yt and import them in NewPipe

ale5000-git:
@Stypox: Yes but the key here is continual automatic sync.

nv95:
Do anyone know how to fetch user's subscriptions without proprietary api and interaction with user? I think, it is not possible (

gkeegan:
It might be possible with Python's Requests module.

ale5000-git:
@nv95: The code could parse the web interface (without API) like NewPipe.

For the account it could use [microG](https://github.com/microg) (it replace Google Play Service for those that don't know).

gkeegan:
I don't think NewPipe will use microG, even considering that it is open source. It still adds that tie to Google than many people do not want.

From the readme of this repository: "you don't need a YouTube account to use NewPipe". Signing in to a Google account has been shot down many times before here, and won't ever be allowed.

The Requests module would be better, as it could sign in to YouTube like you would in a web browser, which could be done in the background. Then it could download the HTML of the subscriptions page, and parse it for channel links. 

That is the best solution to syncing that I think NewPipe would allow, and I'm not even sure this would be allowed.

@ale5000-git
@theScrabi : how would you feel about this? It could maybe be provided in a separate fork, like the legacy fork is doing. It would need some maintainers though.

ale5000-git:
@gkeegan: If you read the first message I was talking about a separate apk (that use microG).
That separate apk should be able to send subscription to NewPipe but it should be still separate.

ale5000-git:
Since an image is better than 1000 words: ![Track My Pipe](https://user-images.githubusercontent.com/15793015/57249427-2e19ae00-7045-11e9-9b05-a33ed6248147.png)

**PS:** It is just an example, nothing currently exist.

gkeegan:
[...]

Issue #1281, opened 2018-04-07
how does one sync their devices?


simonelthis:
I know this is not a forum but how does one sync suscriptions and bookmarks between playlists?

theScrabi:
There is non yet, you could use the export/import function of the database.

simonelthis:
Is it in the works? It says it on your website?
[Untitled.png](https://postimg.org/image/pb1y31g4l/)

TobiGr:
Oh. I should add a _manually_. ATM you need to go to the settings, export the stuff, get the zip file to the other device and then import it.

theScrabi:
There is no automated sync yet. I think this is was people understand when they hear the word sync.

@TobiGr I think we may not want to call it sync on the website. Call it import/export.

wb9688:
Can't we use Bluetooth? Everything is better with Bluetooth ;)

theScrabi:
I think we should consider a proper cloud solution, so syncing does not involve a user action.

TomJansen:
@theScrabi maybe with webDAV? Then you can support many cloud solutions in one go

theScrabi:
Yes thats the plan :D

theScrabi:
no

trymeouteh:
I am personally not a fan of European cloud storage providers (Except for Switzerland). Is there a list of all the cloud storage providers that support webdav?

theScrabi:
certainly. You can even host a webdav storrage at home if you want ;)

speednroll:
this is directly related to the feature request that I did previously #1163

cocoonkid:
please make webdav happen!

rshadow:
I have a home media server. WebDaw would be the perfect solution for syncing devices at home.

opusforlife2:
Have a look at #3157.

Rik44444:
hello all! Thumbs up for webdav sync! Also a solution like this like Antennapod does might be possible https://gpodder.net

EmanuelLoos:
You could also implement something like [Syncthing](https://en.wikipedia.org/wiki/Syncthing) or do it the way [Briar](https://briarproject.org/) or [OnionShare](https://onionshare.org/) transfer information over Tor.

That way it would be serverless and the user would b in full control of his/her data.

natema:
[...]

Issue #1163, opened 2018-03-05
[Feature Request] save your playlists, subscriptions and settings in the cloud


speednroll:
I'm not sure if this goes a bit against the policy of Newpipe, but I think it would be a very good idea to make a server in the style "firefox-sync" in which you can register and store your data there to be able to synchronize them more easily in other phones, tablets and amazonTvs. maybe even this idea works also for when Newpipe is ported to computers

theScrabi:
Well im currently working with ownCloud on their android app, lets se what the future will bring ;D

speednroll:
great, I'll wait for news :grimacing:

theScrabi:
Well an idea would be some sort of intent interface for the owncloud app, so newpipe could trigger it to sync its database. I'm just not sure if ownCloud would like that 😅

gabriellluz:
It'd be like an OwnCloud app or extension? something like that? it would be cool BUT I don't use Owncloud anymore because they collect your data. So your best best is to create an encrypted file (end to end). Otherwise, OwnCloud exposes our data so that NSA can create a simple tunnel/scraper in order to grab our data.  They do it currently with GMail, Outlook, GDrive and so many other providers according to Edward Snowden. There's a book I read about it.

theScrabi:
![serveimage](https://user-images.githubusercontent.com/1891273/41437579-1878ab8a-7025-11e8-9b25-ac19213757e4.jpeg)

trymeouteh:
I would love this feature since I would not have to worry about losing data if my phone is stolen, lost or stops working since it will constantly sync itself to a cloud and everytime I add a subscription, a bookmark or create a playlists it will be backed up.

Same goes for my Tablet. I would like to have my bookmarks and subscriptions synced between them for easy use instead of having to export and import the data all the time.

The only issue with cloud syncing is there are so many providers and I find that not many are using WebDav and others like myself have their own trusted cloud providers since many cloud providers are not privacy friendly like Google Drive, OneDrive, etc.

Some people host their own NAS and use ownCloud or NextCloud but not everyone has this nor wants this setup for privacy. I personally like using MEGA and Sync.com. But to ensure privacy and to also ensure everyone gets to choose their favorite cloud with ease I think the following needs to be done especially for cloud storage providers that do not support webdav.

-File
[...]

PR #2668, opened 2019-09-28
Account/Cloud sync support

yausername:

This PR adds an option to create newpipe account on server and sync subscriptions, playlists, etc.
Backend code - https://github.com/yausername/newpipe-sync-server

- Login is optional, app will work as usual if one chooses to skip login.
- Latest data is fetched from server on login.
- Data can be updated using pull down to refresh in `BookmarkFragment, SubscriptionFragment, LocalPlaylistFragment, StatisticsPlaylistFragment`
- Data is encrypted before it is sent to server. So the server has no knowledge of your data. It only knows the ciphertext.
- There is no way to recover data is one forgets password.
- Changing password is hard, since all the existing data has to be re-encrypted.
- I will add more information here

**Disclaimer: This is a draft implementation. This is expected to be buggy and might contain some security/crypto related bugs as well. Your local data (of debug app) will be erased if you choose to login.**

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -40,6 +40,7 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+
 }
 
 ext {
@@ -99,4 +100,6 @@ dependencies {
 
     implementation "com.squareup.okhttp3:okhttp:${okHttpLibVersion}"
     debugImplementation "com.facebook.stetho:stetho-okhttp3:${stethoLibVersion}"
+
+    implementation 'com.google.crypto.tink:tink-android:1.3.0-rc1'
 }
--- a/app/src/main/java/org/schabi/newpipe/App.java
+++ b/app/src/main/java/org/schabi/newpipe/App.java
@@ -6,9 +6,9 @@ import android.app.NotificationChannel;
 import android.app.NotificationManager;
 import android.content.Context;
 import android.os.Build;
-import android.preference.PreferenceManager;
 import android.support.annotation.Nullable;
 import android.util.Log;
+import android.widget.Toast;
 
 import com.nostra13.universalimageloader.cache.memory.impl.LRULimitedMemoryCache;
 import com.nostra13.universalimageloader.core.ImageLoader;
@@ -21,9 +21,10 @@ import org.acra.config.ACRAConfiguration;
 import org.acra.config.ACRAConfigurationException;
 import org.acra.config.ConfigurationBuilder;
 import org.acra.sender.ReportSenderFactory;
+import org.schabi.newpipe.database.Database;
+import org.schabi.newpipe.database.RemoteDatabase;
 import org.schabi.newpipe.extractor.Downloader;
 import org.schabi.newpipe.extractor.NewPipe;
-import org.schabi.newpipe.extractor.utils.Localization;
 import org.schabi.newpipe.report.AcraReportSenderFactory;
 import org.schabi.newpipe.report.ErrorActivity;
 import org.schabi.newpipe.report.UserAction;
@@ -37,6 +38,7 @@ import java.net.SocketException;
 import java.util.Collections;
 import java.util.List;
 
+import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.annotations.NonNull;
 import io.reactivex.exceptions.CompositeException;
 import io.reactivex.exceptions.MissingBackpressureException;
@@ -107,6 +109,14 @@ public class App extends Application {
 
         // Check for new version
         new CheckForNewAppVersionTask().execute();
+
+        Database database = NewPipeDatabase.getInstance(this);
+        if(database instanceof RemoteDatabase){
+            ((RemoteDatabase) database).sync().observeOn(AndroidSchedulers.mainThread()).subscribe(() -> {}, e -> {
+                Toast.makeText(this, "Failed to sync data from server", Toast.LENGTH_SHORT).show();
+            });
+        }
+
     }
 
     protected Downloader getDownloader() {
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -49,6 +49,7 @@ import android.widget.Button;
 import android.widget.ImageView;
 import android.widget.TextView;
 
+import org.schabi.newpipe.auth.AuthService;
 import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
@@ -529,7 +530,14 @@ public class MainActivity extends AppCompatActivity {
         StateSaver.clearStateFiles();
         if (getIntent() != null && getIntent().hasExtra(Constants.KEY_LINK_TYPE)) {
             handleIntent(getIntent());
-        } else NavigationHelper.gotoMainFragment(getSupportFragmentManager());
+        } else {
+            AuthService authService = AuthService.getInstance(getApplicationContext());
+            if(!authService.isLoggedIn() && !authService.skipLogin()){
+                NavigationHelper.openLoginFragment(getSupportFragmentManager());
+            }else{
+                NavigationHelper.gotoMainFragment(getSupportFragmentManager());
+            }
+        }
     }
 
     /*//////////////////////////////////////////////////////////////////////////
--- a/app/src/main/java/org/schabi/newpipe/NewPipeDatabase.java
+++ b/app/src/main/java/org/schabi/newpipe/NewPipeDatabase.java
@@ -4,39 +4,71 @@ import android.arch.persistence.room.Room;
 import android.content.Context;
 import android.support.annotation.NonNull;
 
+import org.schabi.newpipe.auth.AuthService;
 import org.schabi.newpipe.database.AppDatabase;
+import org.schabi.newpipe.database.Database;
+import org.schabi.newpipe.database.RemoteDatabaseImpl;
 
 import static org.schabi.newpipe.database.AppDatabase.DATABASE_NAME;
 import static org.schabi.newpipe.database.Migrations.MIGRATION_11_12;
+import static org.schabi.newpipe.database.Migrations.MIGRATION_12_13;
 
 public final class NewPipeDatabase {
 
-    private static volatile AppDatabase databaseInstance;
+    private static volatile Database remoteDatabaseInstance;
+    private static volatile Database localDatabaseInstance;
 
     private NewPipeDatabase() {
         //no instance
     }
 
-    private static AppDatabase getDatabase(Context context) {
+    private static Database getRemoteDatabase(AppDatabase roomDb, Context context) {
+        return new RemoteDatabaseImpl(roomDb, context);
+    }
+
+    private static Database getLocalDatabase(Context context) {
         return Room
                 .databaseBuilder(context.getApplicationContext(), AppDatabase.class, DATABASE_NAME)
-                .addMigrations(MIGRATION_11_12)
+                .addMigrations(MIGRATION_11_12, MIGRATION_12_13)
                 .fallbackToDestructiveMigration()
                 .build();
     }
 
     @NonNull
-    public static AppDatabase getInstance(@NonNull Context context) {
-        AppDatabase result = databaseInstance;
-        if (result == null) {
+    private static Database getRemoteInstance(@NonNull Context context) {
+        AppDatabase roomDb = (AppDatabase) getLocalInstance(context);
+        if (remoteDatabaseInstance == null) {
             synchronized (NewPipeDatabase.class) {
-                result = databaseInstance;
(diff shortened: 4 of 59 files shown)


yausername:
Screenshots

login  signup
settings account-settings

yausername:
https://github.com/TeamNewPipe/NewPipe/issues/1163
https://github.com/TeamNewPipe/NewPipe/issues/1281
https://github.com/TeamNewPipe/NewPipe/issues/2278

TobiGr:
Hey @yausername,
that sounds awesome :heart:
We might need to add some information to the login screen for what that account is used, where the data is stored (one can host their own server), who can access the data, etc.
It would be nice of one could change the instance before creating an account, too.
@TheAssassin I guess these changes require updates to our privacy policy.
Is the data stored in a way, that FreeTube and Invidio.us can benefit from it and cross sync functionality can be added later?

TheAssassin:
Wasn't the idea behind NewPipe always to not implement any form of authentication, whatever kind of?

We have to discuss if we want such a thing first of all, then how it should be hosted and then what privacy and data protection implications it will have.

TheAssassin:
I see that there's some form of user data encryption. If we make this encryption part of our privacy policy (i.e., "we encrypt your data with your passphrase", ...), we'll have to carefully review this in order to be on the safe side. Implementing crypto is always a delicate task.

We're also not legal entity. Liability is a topic that needs to be discussed as well.

Spoiler: not a fan of this. Just because we can doesn't imply we should...

yausername:
>  It would be nice of one could change the instance before creating an account, too.

Yes, makes sense. Currently you can skip the initial login screen, change instance from `settings->account->server instance` and then create account using `settings->account->login`


>  Is the data stored in a way, that FreeTube and Invidio.us can benefit from it and cross sync functionality can be added later?

The schema is similar to that of newpipe internal sqllite db. But since data is encrypted, other services will need to decrypt it before consuming.

TheAssassin:
> > Is the data stored in a way, that FreeTube and Invidio.us can benefit from it and cross sync functionality can be added later?
> 
> The schema is similar to that of newpipe internal sqllite db. But since data is encrypted, other services will need to decrypt it before consuming.

That's not how this kind of encryption works. @yausername already pointed out in the initial post that the data is lost when the passwort is lost, i.e., there's no passphrase reset (which IMO is not so great,  but perhaps we can find a possibility to implement this somehow later).

As said, this is a really delicate kind of service to implement and host in accordance to the laws.

I appreciate the suggestion to provide a self-hosting option.

By the way, do we really need custom service? Perhaps this can be integrated with existing services like {Own,Next}Cloud in some way?

yausername:
> > > Is the data stored in a way, that FreeTube and Invidio.us can benefit from it and cross sync functionality can be added later?
> > 
> > 
> > The schema is similar to that of newpipe internal sqllite db. But since data is encrypted, other services will need to decrypt it before consuming.
> 
> That's not how this kind of encryption works. @yausername already pointed out in the initial post that the data is lost when the passwort is lost, i.e., there's no passphrase reset (which IMO is not so great, but perhaps we can find a possibility to implement this somehow later).

Data can be decrypted with the same username, password combination by other clients like FreeTube, etc. So cross sync is still possible.

yausername:
> By the way, do we really need custom service? Perhaps this can be integrated with existing services like {Own,Next}Cloud in some way?

Sync across devices using flat files is not straightforward I think.

omarroth:
>  there's no passphrase reset (which IMO is not so great, but perhaps we can find a possibility to implement this somehow later).

Generally you would create a master key that is actually used for encryption. The master can then be encrypted with the password to generate _K1_ and a random recovery passphrase to generate _K2_. You can extend this to generate _K3_, etc.).

As always with these things I would recommend finding some other resource that provides some more explanation. In case you haven't already seen it the [OWASP cheatsheet](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cryptographic_Storage_Cheat_Sheet.md) should be useful, as well as [this](https://security.stackexchange.com/a/88989) answer on SO.

PopeRigby:
Why create a new account system if you can use Invidious? I think at least having the option to use that would be useful.

TheAssassin:
Thanks @omarroth, highly appreciated. Of course the solution is a (or multiple) recovery passphrases, although that decreases the overall security again. It's logical that if you want multiple passphrases for a (symmetrically) encrypted storage, you need some sort of intermediate passphrase that is then encrypted with the real passphrases. This method is used widely e.g., by LUKS, which IIRC call that intermediate passphrase the *master key* (they have 8 slots for passphrases IIRC).
Another problem solved by this intermediate passphrase is you can simply add/remove new passphrases. Changing passphrases becomes adding the new one and removing the old one, basically, which is very convenient and much faster and more secure than re-encrypting.

Right now I'm not primarily interested into (web) security, but legal implications (and the need for proper security is a consequence from that).

@PopeRigby does Invidious provide sufficient functionality? I had a glance at their privacy policy, I think it'd need some additional love to be in accordance with EU laws (mainly GDPR related, it lacks some information regarding storage and treatment of data and also requires a few standard texts about users' rights), but we could fix that then and refer to them from our privacy policy.

yausername:
I am using client-side encryption (correct me if this is not a good choice). The encryption key is derived from the username (salt), password (passphrase) combination. Here is the code for that.

```java
    public SecretKey generateKey(char[] passphraseOrPin, byte[] salt) throws NoSuchAlgorithmException, InvalidKeySpecException {
        final int iterations = 1000;
        final int outputKeyLength = 512;
        SecretKeyFactory secretKeyFactory = SecretKeyFactory.getInstance("PBKDF2WithHmacSHA1");
        KeySpec keySpec = new PBEKeySpec(passphraseOrPin, salt, iterations, outputKeyLength);
        SecretKey secretKey = secretKeyFactory.generateSecret(keySpec);
        return secretKey;
    }
```

A username, password combination will generate the same key every time. As @omarroth suggested we can generate a random master key and use that for encryption instead of directly using the password derived key. In this case we can have multiple encrypted versions of keys with different passphrases.

yausername:
> Why create a new account system if you can use Invidious? I think at least having the option to use that would be useful.

I haven't explored that option but my main reason to create a new service was client side data encryption.

PopeRigby:
> @PopeRigby does Invidious provide sufficient functionality? I had a glance at their privacy policy, I think it'd need some additional love to be in accordance with EU laws (mainly GDPR related, it lacks some information regarding storage and treatment of data and also requires a few standard texts about users' rights), but we could fix that then and refer to them from our privacy policy.

@omarroth do you need to do any GDPR stuff to make this work?

dimqua:
> does Invidious provide sufficient functionality? I had a glance at their privacy policy, I think it'd need some additional love to be in accordance with EU laws

It hasn't to be a specific server, there is should be a possibility to specify any server.

omarroth:
> I had a glance at their privacy policy, I think it'd need some additional love to be in accordance with EU laws

As far as I'm aware all code should be GDPR-compliant, changes would only be required for the privacy policy. If there are any resources you'd recommend on this I'd be happy to take a look.

TheAssassin:
I can't give legal advice, sorry. Key points I'm missing that you should add/revise are, as mentioned already, users' rights (you have to inform so-called "data subjects" (people using your service) that they can have their data corrected, deleted, ..., as long as you collected it via informed consent; with your service I doubt you have to collect anything (except for the server logs maybe, depending on your jurisdiction, I don't know where you live nor where the server is hosted), just make sure you make it sound like "you have a right to have it deleted/corrected as per GDPR Art. 16/17"), if you make backups refer to them as well, and, really important, to be GDPR compliant, you ***must*** provide information about who is responsible about the data collection and how they can be reached. I as a user cannot enforce any of my rights if I don't know who to talk to about. I know that's annoying, but like I can't have deleted what I don't know is collected I cannot send a GDPR Art. 17 deletion request if I don't know who to send it to.

Oh, and you support some Google authentication, right? I'm missing a link to Google's privacy policy for the service. Wouldn't hurt to refer to their privacy policy anyway as the videos are coming from YouTube as well.
Also I have big doubts you have a (valid!) contract about commissioned data processing with Google. You need such a thing in order to make use of third-party services. They basically agree on a level of data protection and become responsible for adhering to this. It also protects you a bit, if there's any legal issues with Google *you* as the service provider could be held responsible normally, but with that contract you can basically blame the third party as well.

Other than that, your privacy notice is pretty complete already. I like it. It's better than what I see most of the time. But it's not quite up to being GDPR compliant. And being compliant isn't too hard.

theScrabi:
@omarroth Wouldn't syncing with invidious directly be "cleaner"?
Are you ok if we sync with our own server?
What if you or FreeTube would want to sync as well, would you hook onto this sync server?

TheAssassin:
It appears to me that there's a need for an "inter-app/service video service preference protocol". Standardization ftw!

FetchedAttached:
This sounds wonderful!
I know quite some people are hoping a very long time which lets you sync your history/subscriptions etc between NewPipe and Invidio!
Sometimes you want to watch on a desktop (invidio.us), sometimes on your smartdevice (NewPipe).
Completing this integration would be astounding!

omarroth:
> Wouldn't syncing with invidious directly be "cleaner"?

I think so, as I mentioned in #1163 probably the only thing lacking from the API would be proper support for playlists. I'm in the IRC as well if you'd like to chat there about anything that would need to be added/modified for something like sync.

Since this PR already exists though I think it makes sense to move forward with the proposed design if possible.

> What if you or FreeTube and minirube would want to sync as well, would you hook onto this sync server?

It would be quite a bit of work as it is currently.

For syncing between two web services (newpipe-sync-server and an Invidious instance), it looks like it would be necessary to make calls fairly often (e.g. when viewing preferences, playlists, etc.) to the other service to ensure everything is up-to-date (this is what Invidious does for syncing Google accounts) which can carry a bit of overhead.

From a UI perspective, a user would have two different accounts, potentially depending on how they are registered with newpipe-sync-server (e.g. using same user and pass as Invidious). From a look through the code, my understanding is that Invidious would have to store the session token for newpipe-sync-server, and (potentially) a URL if using a custom instance, although I'm not sure if you're currently looking into this as a possibility.

Hopefully from the above comments I've made it clear I am not opposed to this design. It'll be the for the NewPipe team to decide what works best, and I will look into supporting it.

Obyonekenooby:
> Again, my main issue is that I'm not sure whether we want to offer a centralized sync server instance. 

I get where you are coming from and a lot of this sounds challenging from a legal perspective to say the least.
I have been thinking about this, why not focus on a standardized format, but if the user wants to have his data
shared between multiple devices, he can do so by his own instance of owncloud/nextcloud?
So you won't have to deal with the legal aspect of having to store user data, you store the data of a user at
his/her own place. Such that the user is in full control whether they want to store data.
Maybe you could even let the user choose which data is stored, subscriptions, watched videos, etc.

Two things have to be kept in mind: a syncing service and a data format needs to be developed.
And, the apps/websites should not leak any information of the owncloud/nextcloud instances.
You just want these apps/websites to fetch and store the subscriptions/watched videos/.. of a user, and nothing more.

PopeRigby:
If you just use Nextcloud, you can't sync playlists, history, etc.  with Invidious, which I think is something that's very useful.

ushername:
A better implementation would be to use WebDAV, see #1163.
This gives the possibility to choose on which platform you want to host your data.
Also, you can watch videos and finish them on another device, since it is a possibility
to store at what point a video has been stopped.
Video_progress: {12:55}
or
Video_progress: {0:00}
The first gets triggered when the video hasn't completed, the second when one has completed
the video. This is a bad implementation, one could think of many solutions.

stockholesyndrome:
What is the current status of this discussion? Are there any real ideas to implement account sync between newpipe and invidious using WebDAV?

PopeRigby:
I webDAV along with being able to sign in with Invidious would be a good idea.

vinibali:
Great work, I subscribe :)

trymeouteh:
https://github.com/TeamNewPipe/NewPipe/issues/3198

Would like to suggest adding invidio.us account suppprt in NewPipe.

PopeRigby:
What made you close the pull request?

yausername:
I don't think this as a good idea any more. Storing user data, encrypted or not should be avoided as much as possible. An implementation using owncloud/nextcloud can be considered though.
I think #3157 implements backups in a simpler, more secure albeit not as feature-full way.

BeatLink:
Im so sad to see this pull request closed as I was really looking forward to this feature. With the sync server being self hostable, I don't think this would have compromised user security any more than an implementation of owncloud or nextcloud. In any case, I believe it should be up to the user where their data is stored, and this pull request gives the that option.

trymeouteh:
> I don't think this as a good idea any more. Storing user data, encrypted or not should be avoided as much as possible. An implementation using owncloud/nextcloud can be considered though.
> I think #3157 implements backups in a simpler, more secure albeit not as feature-full way.

Can you add an optional feature to login to a invidious instance which can sync your subscriptions, video feed and playlists? I use invidious on my desktop and it will be cool to be able to access my invidious account in NewPipe.

Again this will be optional, not mandatory.

https://github.com/TeamNewPipe/NewPipe/issues/3198

PR #2690, opened 2019-10-04
Show video thumbnail on the lock screen

### LINKED ISSUE #1570: [Feature Request] Show video thumbnail on lock screen

K1rakishou:
Ahh, sorry had to delete the previous branch together with the PR (https://github.com/TeamNewPipe/NewPipe/pull/2320) because everything kinda broke after I tried to rebase it to the fresh develop. But now this thing works via MediaStyle instead of the WallpaperManager and it works pretty good.

Closes #1570


TobiGr:
Thanks for working on this again :) the code looks good and the changes work smoothly on my phone. It's good to see, that when the app is force closed, the wallpaper disappears :+1: 
Unfortunately, Android Studio and IntelliJ are playing games with me and the emulators don't allow me to open the lockscreen... everytime I lock the phone, the emulator's screen stays black - no matter with or without NewPipe installed ^^
@Stypox @theScrabi Can someone test this with KitKat and >= Pie please?

Stypox:
Maybe I can try on my parents' phone, but I have Android 7.

Stypox:
Here is a debug apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/3694189/app-debug.zip)

Stypox:
Debug apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/3694196/app-debug.zip)
With this on Android 7.0 I don't get any background image on the lock screen. Is there something I am missing?

IMG_20191006_120018

Stypox:
Also, another problem I noticed while debugging is that the `setLockScreenArt` function is called every time the notification is updated. Couldn't this cause slowdowns?

Stypox:
Sorry, apparently the problem I'm encountering is caused by MIUI: https://www.xda-developers.com/how-to-add-third-party-music-apps-emui-lock-screen/

K1rakishou:
Sorry for the late reply. 

So regarding the constant `setLockScreenArt` spam. I've just tried to set the thumbnail only once but it didn't work - the  thumbnail have disappeared. I guess there is just no other way than calling it on every notification update. I guess this (https://github.com/TeamNewPipe/NewPipe/pull/2771) PR will reduce then amount of notification updates which will also reduce the amount of times the thumbnail is set.

Regarding the EMUI not showing the thumbnail at all I guess there is also no other way since it requires root access.

I have also rebased my branch onto the latest dev and updated all appcompat methods to use androidx.

TobiGr:
Thanks for rebasing.

> So regarding the constant `setLockScreenArt` spam. I've just tried to set the thumbnail only once but it didn't work - the thumbnail have disappeared. I guess there is just no other way than calling it on every notification update. I guess this (#2771) PR will reduce then amount of notification updates which will also reduce the amount of times the thumbnail is set.

I'd like to make sure to not make the background player unusable for older devices. Can you please add a setting which allows disabling this feature?

K1rakishou:
@TobiGr Sure, no problem.

K1rakishou:
Alright, the setting is in.

TobiGr:
> Also, another problem I noticed while debugging is that the `setLockScreenArt` function is called every time the notification is updated. Couldn't this cause slowdowns?

@Stypox I merged the other PR limiting notification thumbnail updates. Now the lockscreen is updated ~30 seconds or when a new stream is played. It is not perfect, but better than nothing. And with allowing users to disable the lockscreen thumbnail, we should be save to merge this imho. What do you think?

app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
```diff
@@ -193,18 +201,51 @@ private void resetNotification() {
...
-                .setSmallIcon(R.drawable.ic_newpipe_triangle_white)
-                .setVisibility(NotificationCompat.VISIBILITY_PUBLIC)
-                .setCustomContentView(notRemoteView)
-                .setCustomBigContentView(bigNotRemoteView);
+        NotificationCompat.Builder builder = new NotificationCompat.Builder(this, getString(R.string.notification_channel_id));
+        builder.setOngoing(true);
+        builder.setSmallIcon(R.drawable.ic_newpipe_triangle_white);
+        builder.setVisibility(NotificationCompat.VISIBILITY_PUBLIC);
+        builder.setCustomContentView(notRemoteView);
+        builder.setCustomBigContentView(bigNotRemoteView);
```
Stypox: Why did you convert the method call chain into single calls? Afaik it is good practice to arrange methods in a chain when using Builders/Factories.

app/src/main/java/org/schabi/newpipe/player/helper/MediaSessionManager.java
```diff
@@ -40,13 +49,52 @@ public KeyEvent handleMediaButtonIntent(final Intent intent) {
...
+            NotificationCompat.Builder builder,
+            @Nullable Bitmap thumbnailBitmap
+    ) {
+        if (thumbnailBitmap == null) {
+            return;
+        }
+
+        if (!mediaSession.isActive()) {
+            return;
+        }
```
Stypox: This can be simplified into:
```
if (thumbnailBitmap == null || !!mediaSession.isActive()) {
    return;
}
```

app/src/main/java/org/schabi/newpipe/player/helper/MediaSessionManager.java
```diff
@@ -40,13 +49,52 @@ public KeyEvent handleMediaButtonIntent(final Intent intent) {
         return MediaButtonReceiver.handleIntent(mediaSession, intent);
     }
 
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    public void setLockScreenArt(
+            NotificationCompat.Builder builder,
+            @Nullable Bitmap thumbnailBitmap
+    ) {
```
Stypox: I would prefer if all the function declaration were on the same line, since there are only two arguments, and you don't come across many multiline declarations in the NewPipe codebase. But this is not really important ;-)

app/src/main/java/org/schabi/newpipe/util/BitmapUtils.java
```diff
@@ -0,0 +1,43 @@
...
+        float sourceHeight = inputBitmap.getHeight();
+
+        float xScale = newWidth / sourceWidth;
+        float yScale = newHeight / sourceHeight;
+
+        float newXScale;
+        float newYScale;
+
+        if (yScale > xScale) {
+            newXScale = (1.0f / yScale) * xScale;
```
Stypox:
```suggestion
            newXScale = xScale / yScale;
```

app/src/main/java/org/schabi/newpipe/util/BitmapUtils.java
```diff
@@ -0,0 +1,43 @@
...
+            newYScale = 1.0f;
+        } else {
+            newXScale = 1.0f;
+            newYScale = (1.0f / xScale) * yScale;
```
Stypox:
```suggestion
            newYScale = yScale / xScale;
```

app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
TobiGr: good question. I think we should undo this, too

K1rakishou: Yeah, there was no reason for me to do it like this. Reverted it to how it was initially.

K1rakishou:
I have just noticed that the lockscreen art have become blurred. I remember it wasn't like that when I was testing it the last time and IIRC I used Android 8 emulator that time and now I'm using Android 10 emulator, so maybe that's how the OS handles it now?

TobiGr:
It looks like you cannot change this see [this article](https://www.androidheadlines.com/2019/03/android-q-blur-background-music.html) or [this one](https://www.androidpolice.com/2019/03/13/android-q-uses-blurred-lock-screen-background-while-playing-music)
It seems like the blur effect [is hardcoded](https://android.googlesource.com/platform/frameworks/base/+/android10-release/packages/SystemUI/src/com/android/systemui/statusbar/MediaArtworkProcessor.kt ) when `processArtwork()` is called by the system.

K1rakishou:
Alright, so then it's okay to leave it as it is?

TobiGr:
Yes

app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
```diff
@@ -356,12 +403,11 @@ public void onUpdateProgress(int currentProgress, int duration, int bufferPercen
             updateProgress(currentProgress, duration, bufferPercent);
 
             if (!shouldUpdateOnProgress) return;
-            if (timesNotificationUpdated > NOTIFICATION_UPDATES_BEFORE_RESET) {
-                resetNotification();
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O /*Oreo*/) updateNotificationThumbnail();
-            }
+            if (timesNotificationUpdated > NOTIFICATION_UPDATES_BEFORE_RESET) {resetNotification();
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O /*Oreo*/)
+                updateNotificationThumbnail();}
```
Stypox: This is really messy, why was it refactored this way? ;-)
I think it should be transformed into
```
            if (timesNotificationUpdated > NOTIFICATION_UPDATES_BEFORE_RESET) {
                resetNotification();
                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O /*Oreo*/) {
                    updateNotificationThumbnail();
                }
            }
```

app/src/main/java/org/schabi/newpipe/player/helper/MediaSessionManager.java
```diff
@@ -40,13 +49,45 @@ public KeyEvent handleMediaButtonIntent(final Intent intent) {
...
+        MediaStyle mediaStyle = new MediaStyle()
+                .setMediaSession(mediaSession.getSessionToken());
+
+        builder.setStyle(mediaStyle);
+    }
+
     /**
-     * Should be called on player destruction to prevent leakage.
-     * */
+     * Should be called on player destruction to prevent leakage.BitmapUtils
```
Stypox: Why is there `BitmapUtils` at the end of the comment?
```suggestion
     * Should be called on player destruction to prevent leakage.
```

app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
K1rakishou: Whoops, I guess this has happened after auto merging. Fixed.

app/src/main/java/org/schabi/newpipe/player/helper/MediaSessionManager.java
K1rakishou: Fixed.

TobiGr: I already fixed this in https://github.com/TeamNewPipe/NewPipe/pull/2690/commits/1aa763e86cb24a8dbda55f7a7576c30a7618a440. Something went wrong in your meged commit: https://github.com/TeamNewPipe/NewPipe/pull/2690/commits/835476870b7b4a04af3f79c6c5ecb0c9b70e588f. This branch was on top of the dev branch already

Pdevo:
I am on kitkat 4.4.2 and the video thumbnail on the lockscreen feature seems not to work, tryed also a fresh install on wiped rom.
Can I help you in some way to troubleshoot?

TobiGr:
@Pdevo This feature is available for Lollipop (5.0) and later. However, we should remove or disable the setting on KitKat and below.

TobiGr:
@Pdevo I've created #3355 to keep track of this

PR #2697, opened 2019-10-07
Update to ExoPlayer 2.10.5

Redirion:
- [X] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

As we now finally have AndroidX I was able to update ExoPlayer to the latest version.
This brings numerous of needed fixes and improvements to NewPipe. Have a look at the ReleaseNotes: https://github.com/google/ExoPlayer/blob/release-v2/RELEASENOTES.md

This might close #2257 and #2373 and possible other issues.

Debug apk for testing: https://github.com/Redirion/NewPipe/releases/download/v0.17.3-exo/app-debug.apk

Notes for Review:

- Changes to the MediaSessionManager and PlayQueueNavigator were mainly required due to this commit in ExoPlayer: https://github.com/google/ExoPlayer/commit/063f5704cd80a733bf770375d8fbf402a6c5d958
- I also had to fiddle again in CustomTrackSelector as there have been changes again. I already tried subtitles (it seems to be working fine this time :D)
- I have updated Gradle to 3.4.0 to match ExoPlayer's gradle version
- okhttp has been updated as well to match ExoPlayer (thus deemed stable)
- With the new isPlaying() method in BasePlayer NewPipe makes already use of a new feature added in ExoPlayer 2.10.5

PR #2685, opened 2019-10-04
Migrate to AndroidX

Redirion:
- [X] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Third attempt.
This PR supersedes PR #2598 (seems to be no longer maintained).
This was finally possible after PR #2535 was merged.
I took the stable approach for continued usage of the modified FlingBehavior.java from this commit 0d25645de9e1e5cf6d48d85d6253351ae3e6f8dc

Closes #2641 

Debug apk: https://github.com/Redirion/NewPipe/releases/download/v0.17.4/app-debug.apk

Note that I left everything unchanged, the Extractor still has the NullPointerException on getChannelCount.

PR #2672, opened 2019-09-29
fix scrolling in video detail fragment. fixes #2627

yausername:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Fixes #2627 
Fixes #2437
Also, ParalaxScrollView is removed from dependencies and credits since it is not being used any more.

PR #2637, opened 2019-09-16
Make language of download folder setting more consistent and less ambiguous

opusforlife2:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #2710, opened 2019-10-10
Release 0.17.4

TobiGr:
### Improved
- Add more titles to improve accessibility for blind people #2655
- Make language of download folder setting more consistent and less ambiguous #2637

### Fixed
- Check if last byte in the block is downloaded #2646
- Fixed scrolling in video detail fragment #2672
- Remove double search clear box animations to one #2695

### Development
- Add missing dependencies inherited from NewPipeExtractor into NewPipe #2535
- Migrate to AndroidX #2685
- Update to ExoPlayer 2.10.5 #2697

--------

### APK for testing
Will be uploaded shortly (CC @theScrabi)


mauriciocolli:
@TobiGr what about using a proper version when updating the extractor this time? It's hard to quickly see how updated the library is through commit hashes. It would also encourage to keep both in sync.

TobiGr:
@mauriciocolli Done. I've quickly implemented the head request, which is required since https://github.com/TeamNewPipe/NewPipeExtractor/pull/209, by copy and pasta your code from the extractor's downloader implementation (see 19180c64ff1afeaee19767e3e4da9c5d87dc7f1f).
@theScrabi We are ready for release, aren't we?

app/src/main/java/org/schabi/newpipe/Downloader.java
```diff
@@ -276,6 +278,36 @@ public DownloadResponse post(String siteUrl, DownloadRequest request) throws IOE
...
+             */
+            if (con.getResponseCode() == 429) {
+                throw new ReCaptchaException("reCaptcha Challenge requested", con.getURL().toString());
+            }
+
+            throw new IOException(con.getResponseCode() + " " + con.getResponseMessage(), e);
+        }
+
+        return new DownloadResponse(con.getResponseCode(), null, con.getHeaderFields());
     }
```
mauriciocolli: Even if this is just temporary, we should still use OkHttp's client:

```java
@Override
public DownloadResponse head(String siteUrl) throws IOException, ReCaptchaException {
    final Request request = new Request.Builder()
            .head().url(siteUrl)
            .addHeader("User-Agent", USER_AGENT)
            .build();
    final Response response = client.newCall(request).execute();

    if (response.code() == 429) {
        throw new ReCaptchaException("reCaptcha Challenge requested", siteUrl);
    }

    return new DownloadResponse(response.code(), null, response.headers().toMultimap());
}
```

TobiGr: Oh yes, you are right.

TobiGr: Done. Thanks for the code :)

PR #2713, opened 2019-10-11
Use new Localization and Downloader implementations from extractor

mauriciocolli:
- Make use of new Localization and Downloader implementations from extractor
- Show proper text for live stream items (watching count instead of view count)

Depends on TeamNewPipe/NewPipeExtractor#158.


app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
```diff
@@ -0,0 +1,159 @@
...
+
+import javax.annotation.Nonnull;
+
+import androidx.annotation.Nullable;
+import okhttp3.OkHttpClient;
+import okhttp3.RequestBody;
+import okhttp3.ResponseBody;
+
+public class DownloaderImpl extends Downloader {
+    public static final String USER_AGENT = "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:43.0) Gecko/20100101 Firefox/43.0";
```
cool-student: Is there a reason to use such an old useragent ? I would suggest updating it (unless there's a good reason to use the old one)

```suggestion
    public static final String USER_AGENT = "Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:69.0) Gecko/20100101 Firefox/69.0";
```

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -109,19 +115,34 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
             return true;
         });
 
-        Preference setPreferredLanguage = findPreference(getString(R.string.content_language_key));
+        /*Preference setPreferredLanguage = findPreference(getString(R.string.content_language_key));
```
TobiGr: Can you please add a comment that this is not needed due to the extractor changes and localization changes.

mauriciocolli:
@TobiGr are you sure? [This is how the file looks like](https://github.com/TeamNewPipe/NewPipe/blob/2a15f5fb51d8455e285c0928222b4e5c55a480d3/app/src/main/res/values/settings_keys.xml#L403..L651) as of the last commit.

It seems that currently only the language names are done that way.

TobiGr:
My bad, I mixed content language and country during my review.

mauriciocolli:
@TobiGr Done.

PR #2714, opened 2019-10-11
Update Gradle and introduce Kotlin to NewPipe

mauriciocolli:
### Update Gradle and introduce Kotlin

I think it's time we introduce Kotlin to NewPipe as I consider it very nice to use. The main reasons/opinions as to why I think it'd be good to the project:
- I want to learn more about it, so why not have more contact with it
- Want to learn something new
- Relative ease to migrate _and_ interact with Java code
- Enjoyable coding overall

If you think this is not a good addition for NewPipe, please post your thoughts below.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,4 +1,7 @@
 apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
+apply plugin: 'kotlin-kapt'
 
 android {
     compileSdkVersion 28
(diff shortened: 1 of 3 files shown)


nv95:
I think, we do not need Kotlin in this project, aren`t?

TobiGr:
IMO Kotlin simplifies code a lot: easy null checks, less unnecessary brackets, etc.
@theScrabi We had the discussion somewhere before. I'd approve Kotlin support. What's you opinion?

Apart from that, #2309 is written in Kotlin and I'd like to get this merged asap.

Stypox:
I am in favour of the introduction of Kotlin, since it is faster to program in and the transition from Java is smooth

PR #2720, opened 2019-10-12
Fix Hebrew translation on some devices

haggaie:
Hebrew on android uses two locale codes (iw and he).
See details here: https://stackoverflow.com/a/8470980

For example, Nexus 7 (2013) uses iw, so it shows the English UI even when configuring the tablet to use Hebrew.

Add a symbolic link from values-iw to values-he so both use the same strings.

### DIFF
--- a/app/src/main/res/values-iw
+++ b/app/src/main/res/values-iw
@@ -1,0 +1,1 @@
+values-he/


Redirion:
Did you test it? In your source is a comment stating that symbolic links will not work

haggaie:
I did test it, using a Nexus 7 tablet with factory image (Android 6.0.1). 

I'm not sure why the comment says there's no way to link.

TobiGr:
Thank you, I'll test it with KitKat soon.

TobiGr:
Thanks. I couldn't test it with an emulator which ran Kitkat and Hebrew, but I guess it will be fine because it didn't brake anything and @haggaie said it worked.

PR #2724, opened 2019-10-14
[Bug Fixed] Playlists With No Uploader No Longer Crash The App When Added To Users Library

PeterHindes:


PeterHindes:
Previously playlists with a null uploader (such as album playlists generated by YouTube music) crashed the app (error screen, returns to playlist list). Now the uploader name is left blank and playlists can be used without an uploader. To replicate problem, try adding [this playlist](https://www.youtube.com/playlist?list=OLAK5uy_kimVHgEfenl6FBXq5dQ469NIvR7pfOoE4) (an album generated by YT-Music) to your library.

PeterHindes:
A future commit may be made which fills the uploader with the uploader of the first video/song in the playlist.

app/src/main/java/org/schabi/newpipe/database/playlist/model/PlaylistRemoteEntity.java
```diff
@@ -72,10 +74,18 @@ public PlaylistRemoteEntity(final PlaylistInfo info) {
 
     @Ignore
     public boolean isIdenticalTo(final PlaylistInfo info) {
-        return getServiceId() == info.getServiceId() && getName().equals(info.getName()) &&
+        boolean returnMe;
```
Stypox: The variable `returnMe` is probably useless as it can be replaced by a simple `return` everywhere it is assigned (lines 79 and 84).

PeterHindes: true, it was kinda a sloppy fix, and to be honest uploader will never change so It can just be removed

PeterHindes:
I think that fixes it

PeterHindes:
Just added some doc-strings before final merge. (Wrong branch for that other commit, so I reverted that, it will be a different pr)

Redirion:
what happens, if the channel name gets changed? https://support.google.com/youtube/answer/2657964

is the channelName equal to uploader?

also the import of TextUtils can be removed

PeterHindes:
> what happens, if the channel name gets changed? https://support.google.com/youtube/answer/2657964
> 
> is the channelName equal to uploader?
> 
> also the import of TextUtils can be removed

Thanks for the catch I hadn't considered that. Ill use the == fix instead. Ill push the fix later today

PeterHindes:
Fix pushed

Redirion:
== will most likely never be true (good for null checks, but bad for String comparison)

What about readding the import of TextUtils and use 
`TextUtils.equals(getUploader(), info.getUploaderName())`
?

This is a null safe comparison of those Strings and should fix it.

PeterHindes:
ill check out some solutions and push a fix later today

PeterHindes:
Ok, that should be better. Sorry for the larger change, but it now lists "Auto-Generated" as the uploader when null, and has debug loging!

Redirion:
did you test, whether simply replacing
`getUploader().equals(info.getUploaderName());`
with 
`TextUtils.equals(getUploader(), info.getUploaderName())`
would also work?
I can't see why it wouldn't but maybe I am missing something.
For reference, this is the implementation of TextUtils.equals:
```

    /**
     * Returns true if a and b are equal, including if they are both null.
     * <p><i>Note: In platform versions 1.1 and earlier, this method only worked well if
     * both the arguments were instances of String.</i></p>
     * @param a first CharSequence to check
     * @param b second CharSequence to check
     * @return true if a and b are equal
     */
    public static boolean equals(CharSequence a, CharSequence b) {
        if (a == b) return true;
        int length;
        if (a != null && b != null && (length = a.length()) == b.length()) {
            if (a instanceof String && b instanceof String) {
                return a.equals(b);
            } else {
                for (int i = 0; i < length; i++) {
                    if (a.charAt(i) != b.charAt(i)) return false;
                }
                return true;
            }
        }
        return false;
    }
```

PeterHindes:
Thats awesome, my java is kinda rusty, and this is my first time doing android dev, so thanks for all the awesome tips, i would be pretty lost without the help 😅

PeterHindes:
I'm planing to move the "Auto-Generated" to strings.

Redirion:
KISS (keep it small & simple) ;)

PeterHindes:
Should I just leave it blank? It seems like that would lack polish.

PeterHindes:
which file has the code which puts null as the uploader in the playlist library?

Stypox:
You should probably look for a `LocalPlaylistItemHolder` or something like that inside the `local` folder.

Also, I think leaving the text blank would be ok. Make sure there is no • when the field is blank, though

PeterHindes:
Got it 👍

PeterHindes:
Under what situation would LocalPlaylistFragment be used instead of PlaylistFragment?

Stypox:
Yes

PeterHindes:
[@karyogamy](https://github.com/karyogamy) what is the point of the [local_playlist_header.xml](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/res/layout/local_playlist_header.xml) file

app/src/main/java/org/schabi/newpipe/local/holder/RemotePlaylistItemHolder.java
```diff
@@ -28,8 +28,13 @@ public void updateFromItem(final LocalItem localItem, HistoryRecordManager histo
 
         itemTitleView.setText(item.getName());
         itemStreamCountView.setText(String.valueOf(item.getStreamCount()));
+        // Here is where the uploader name is set in the bookmarked playlists library
         itemUploaderView.setText(Localization.concatenateStrings(item.getUploader(),
                 NewPipe.getNameOfService(item.getServiceId())));
+        if (item.getUploader() == null) {
```
Redirion:
```suggestion
        if (item.getUploader() != null) {
        itemUploaderView.setText(Localization.concatenateStrings(item.getUploader(),
                NewPipe.getNameOfService(item.getServiceId())));
        } else {
            itemUploaderView.setText(NewPipe.getNameOfService(item.getServiceId()));
        }
```

Redirion:
```suggestion
        if (item.getUploader() != null) {
            itemUploaderView.setText(Localization.concatenateStrings(item.getUploader(),
                NewPipe.getNameOfService(item.getServiceId())));
        } else {
```

Stypox:
@PeterHindes you can find its usages by doing Alt-Shift-7 after selecting the file in the android studio file picker.

Anyway, it is the layout of local playlist info items

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -259,7 +259,8 @@ public void handleResult(@NonNull final PlaylistInfo result) {
         animateView(headerRootLayout, true, 100);
         animateView(headerUploaderLayout, true, 300);
         headerUploaderLayout.setOnClickListener(null);
-        if (!TextUtils.isEmpty(result.getUploaderName())) {
+        if (!TextUtils.isEmpty(result.getUploaderName())) { // If we have an uploader : Put them into the ui
+            //headerUploaderLayout.setVisibility(View.VISIBLE);
```
TobiGr: Please remove the unused code

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -273,6 +274,9 @@ public void handleResult(@NonNull final PlaylistInfo result) {
                     }
                 });
             }
+        } else { // Else : hide the uploader section
+            //headerUploaderLayout.setVisibility(View.INVISIBLE);
```
TobiGr: Same here

PeterHindes:
ok done with those suggestions. I will try a build later today.

PeterHindes:
It seems to be working.

TobiGr:
Thanks. Code looks good. Can someone like @Stypox or @Redirion test it, too? I am a bit busy right now, but I'd like to include it in the next release. When testing succeeds, feel free to merge.
Btw. @Redirion You have contributed frequently to NewPipe over a long period of time. Would you like to join the team?

Redirion:
It looks like additionally a change to the NewPipeExtractor is required. as I got a crash when I tried to share [the URL](https://www.youtube.com/playlist?list=OLAK5uy_kimVHgEfenl6FBXq5dQ469NIvR7pfOoE4) to NewPipe with "Show Infos". 
The methods getUploaderUrl, getUploaderName and getUploaderAvatar failed.

My suggestion would be to first only collect errors for these methods and only add them to the actual errors, if not all three failed or if at least one other error was raised.

I have created a PR for that.

@TobiGr I would be honoured to join :)

PeterHindes:
Yes when loading a playlist with no uploader a little error dialogue shows at the bottom of the screen. It's a bit annoying, but it doesn't crash the app like it did before. A fix to the extractor should make the remaining error go away.

app/src/main/res/values/strings.xml
```diff
@@ -444,6 +444,7 @@
     <string name="playlist_add_stream_success">Playlisted</string>
     <string name="playlist_thumbnail_change_success">Playlist thumbnail changed.</string>
     <string name="playlist_delete_failure">Could not delete playlist.</string>
+    <string name="playlist_no_uploader">Auto-Generated (no uploader found)</string>
```
```suggestion
    <string name="playlist_no_uploader">Auto-generated (no uploader found)</string>
```

Stypox: Only the first letter should be uppercase
```suggestion
    <string name="playlist_no_uploader">Auto-generated (no uploader found)</string>
```

PeterHindes:
allright everything looks good. I removed some unnecessary changes. (I had planed do make a more compact layout without the uploader if there was none found, but it didnt work out)

app/src/main/java/org/schabi/newpipe/local/holder/RemotePlaylistItemHolder.java
```diff
@@ -28,8 +28,14 @@ public void updateFromItem(final LocalItem localItem, HistoryRecordManager histo
 
         itemTitleView.setText(item.getName());
         itemStreamCountView.setText(String.valueOf(item.getStreamCount()));
-        itemUploaderView.setText(Localization.concatenateStrings(item.getUploader(),
+        // Here is where the uploader name is set in the bookmarked playlists library
+        if (item.getUploader() != null) {
```
Redirion:
```suggestion
        if (!TextUtils.isEmpty(item.getUploader) {
```

PeterHindes:
Cool so does this mean it will be in the next release target?

Stypox:
@PeterHindes yeah

TobiGr:
@Stypox and @Redirion thanks for reviewing and merging and thanks to @PeterHindes for creating the fix.
@Stypox Please add the PR to the [drafted release notes](https://github.com/TeamNewPipe/NewPipe/releases) after merging it. This helps to keep track of the changes and speeds up creating the first RC PR.
@Redirion I am glad to hear this! I've sent you an invitation.

Stypox:
Ok, thanks for the tip, I didn't know about the drafted release notes 😉

PeterHindes:
The [pr for the extractor fix](https://github.com/TeamNewPipe/NewPipeExtractor/pull/219) should also be reviewed for the next release. Thanks @Redirion for the second half of the fix.

TobiGr:
@PeterHindes I merged the extractor PR and updated to the latest extractor commit (see b883ad165777e505f59388bdc4e361c15add9293 and TeamNewPipe/NewPipeExtractor@8cb3250)

PR #2736, opened 2019-10-19
Bump ExoPlayer to 2.10.6 and OkHttp to 3.12.6

Redirion:
Little dependency update.
ExoPlayer Changelog: https://github.com/google/ExoPlayer/blob/release-v2/RELEASENOTES.md
OkHttp Changelog: https://square.github.io/okhttp/changelog_3x/#version-3126

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,10 +44,10 @@ android {
 
 ext {
     androidxLibVersion = '1.0.0'
-    exoPlayerLibVersion = '2.10.5'
+    exoPlayerLibVersion = '2.10.6'
     roomDbLibVersion = '2.1.0'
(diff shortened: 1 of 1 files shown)


Pentaphon:
That audio focus bugfix on Exoplayer is probably worth doing before the next release.

Redirion:
It looks like this is more a change for ExoPlayer's own audio focus handling, which NewPipe doesn't use.
Speaking of that I think it might be considerable to make use of ExoPlayer's AudioFocusManager instead of NewPipes own AudioReactor. The AudioReactor might also be a source of "SimpleExoPlayer is accessed from the wrong thread" warnings.

PR #2742, opened 2019-10-20
Fix tab handling and enable ignored tests again

mauriciocolli:
Was testing some other PRs and noticed that the main fragment had some problems, the most important one being the way the view pager was being handled.

- Remove usage of hardcoded title for default kiosk.
- Some tests were simply ignored, this PR enables them again.
- Fix broken view pager tabs implementation
    - Fragments were being recreated from scratch (losing their state) every time some configuration change occurred (e.g. screen rotation).
    - Use `FragmentStatePagerAdapter` instead, as it is built to work with them and manage their states.
- Clear the item list when starting loading.
- Enable toolbar title visibility when setting a new one.
- Use tab position from parameters instead of relying on the view pager.
- Make the `KioskFragment` aware of changes in the preferred content country.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
@@ -1,5 +1,6 @@
 package org.schabi.newpipe.fragments;
 
+import android.content.Context;
 import android.os.Bundle;
 import android.util.Log;
 import android.view.LayoutInflater;
@@ -15,7 +16,7 @@ import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
-import androidx.fragment.app.FragmentPagerAdapter;
+import androidx.fragment.app.FragmentStatePagerAdapter;
 import androidx.viewpager.widget.ViewPager;
 
 import com.google.android.material.tabs.TabLayout;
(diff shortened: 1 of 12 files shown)


Stypox:
> Fragments were being recreated from scratch (losing their state) every time some configuration change occurred (e.g. screen rotation).

Does this imply that now when the user changes i.e. the language, the views do not reload?

mauriciocolli:
> Does this imply that now when the user changes i.e. the language, the views do not reload?

Do you mean changes like language of the phone? If you are referring to this, yes, the fragments will reload like any other in the app. Android will recreate and give the correct resources when the fragment is starting up again.

But if you meant the language that is the settings of the app, no, it will not because there is no check to see if it changed or anything. Notice that this isn't the result of this PR, it doesn't work in the current version either. This is a flaw in the current design, things are mixed up in the fragment and the layers (view, data source) are not separated optimally.

And speaking of the language selector, it was decided that pages will be fetched only in english (or whatever the service offer by default), as we don't have to take in account _every_ little change when parsing the page. The content country selector is very useful though.

I will push a commit fixing that in a second.

Stypox:
OK, thank you

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -52,32 +53,21 @@ public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setHasOptionsMenu(true);
 
-        destroyOldFragments();
+        // getChildFragmentManager().enableDebugLogging(true);
```
TobiGr: I think this line was accidentally added.

mauriciocolli:
> Unfortunately you cause a NPE when loading a soundcloud kiosk page which has not the current kioskId. No error with mediaCCC. This does not happen with 0.17.4

The stack trace points to the client id extraction function of the SoundCloud extractor, leading to where the problem is: this branch is outdated so it's still using an old version of the extractor (also in the crash report: `Version: 0.17.3`).

> Can you please rebase and address my comment above?

Done, fixes the other issue as well.

TobiGr:
> The stack trace points to the client id extraction function of the SoundCloud extractor, leading to where the problem is: this branch is outdated so it's still using an old version of the extractor (also in the crash report: Version: 0.17.3).

All right, my bad.

> Done, fixes the other issue as well.

Perfect

Issue #1401, opened 2018-05-15
Double tap options


umunais:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.
please provide an option to adjust the double tap seek timing from at least five seconds onwards.

Somethingweirdhere:
Looks like an easy change. Two questions:

 - Should there be a picker between 5, 10 and 30 seconds or just a field where you can input the desired amount?
 - Double tapping and pressing the fast forward button in the background player notification currently use the same function in Baseplayer, so if you change the double tap seek time it also changes the seek time of the background player. I assume that's alright, or do we need seperate options for both players?

umunais:
Thanks for your response.
Well, personally I think a picker to choose time period less than 5 seconds
in seek option is unnecessary since adding quite a number of multiples of 5
will be more than sufficient, just like in original YouTube API.
also i don't think no modification related to controls is needed so far as
background and main players are concerned.

umunais:
The issue has not been resolved

Keychain-SSB:
Being able to skip just 5 seconds instead of 10 would be ideal. Fingers crossed on this being a thing soon!

PR #2744, opened 2019-10-21
Adjustable seek duration

XiangRongLin:

Added an option in the settings for user to set the duration which gets fast forwarded/rewinded when they double-tap in a video or press the corresponding buttons in background mode.

Related issue: #1401

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -954,14 +954,12 @@ public abstract class BasePlayer implements
 
     public void onFastRewind() {
         if (DEBUG) Log.d(TAG, "onFastRewind() called");
-        final int duration = getSeekDuration();
-        seekBy(-duration);
+        seekBy(-getSeekDuration());
     }
 
     public void onFastForward() {
         if (DEBUG) Log.d(TAG, "onFastForward() called");
-        final int duration = getSeekDuration();
-        seekBy(duration);
+        seekBy(getSeekDuration());
     }
 
     private int getSeekDuration() {
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -954,20 +954,21 @@ public abstract class BasePlayer implements
 
     public void onFastRewind() {
         if (DEBUG) Log.d(TAG, "onFastRewind() called");
-        final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
-        final String key = context.getString(R.string.seek_duration_key);
-        final String value = prefs.getString(key, FAST_FORWARD_REWIND_DEFAULT_AMOUNT_MILLIS);
-        final int duration = Integer.parseInt(value);
+        final int duration = getSeekDuration();
         seekBy(-duration);
     }
 
     public void onFastForward() {
         if (DEBUG) Log.d(TAG, "onFastForward() called");
+        final int duration = getSeekDuration();
+        seekBy(duration);
+    }
+
+    private int getSeekDuration() {
         final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
         final String key = context.getString(R.string.seek_duration_key);
         final String value = prefs.getString(key, FAST_FORWARD_REWIND_DEFAULT_AMOUNT_MILLIS);
-        final int duration = Integer.parseInt(value);
-        seekBy(duration);
+        return Integer.parseInt(value);
     }
 
     public void onPlayPrevious() {
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -178,7 +178,7 @@ public abstract class BasePlayer implements
     // Player
     //////////////////////////////////////////////////////////////////////////*/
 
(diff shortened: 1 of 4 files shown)


TobiGr:
Thanks!
@Stypox or @mauriciocolli do you think we should cache some of the values? Most of the time I use the forward / rewind buttons, I tap on them multiple times. I think this is partly because the old duration was quite short. That's why I'd like to hear a second opinion on this.

Stypox:
@TobiGr What do you mean by "cache some of the values"?
Also @XiangRongLin the name of the setting should probably be something along the lines of "Fast-forward seek duration", otherwise it may be confused with the other ways to seek.

TobiGr:
@stypox Nevermind.

@XiangRongLin Can you please rebase?

XiangRongLin:
@TobiGr rebased it.  
Should the getting of the seek duration be put into its own function? Since i have the exact same 4 lines in forward in rewind [here](https://github.com/XiangRongLin/NewPipe/commit/b2ba0f3052731e834b910790f88e148be64efad2).

TobiGr:
Yes, this might be good.

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -954,12 +954,21 @@ public void onPlayPause() {
 
     public void onFastRewind() {
         if (DEBUG) Log.d(TAG, "onFastRewind() called");
-        seekBy(-FAST_FORWARD_REWIND_AMOUNT_MILLIS);
+        final int duration = getSeekDuration();
```
TobiGr: I think there is no need for a local var here

XiangRongLin:
@TobiGr Rebased it

app/src/main/res/values/settings_keys.xml
```diff
@@ -28,6 +28,25 @@
     <string name="screen_brightness_key" translatable="false">screen_brightness_key</string>
     <string name="screen_brightness_timestamp_key" translatable="false">screen_brightness_timestamp_key</string>
 
+    <string name="seek_duration_key" translatable="false">seek_duration</string>
+    <string name="seek_duration_default_key" translatable="false">10000</string>
```
Stypox: This variable should probably be called `seek_duration_default_value`. You should be good by just committing this and the below suggestion directly.
```suggestion
    <string name="seek_duration_default_value" translatable="false">10000</string>
```

app/src/main/res/xml/video_audio_settings.xml
```diff
@@ -146,5 +146,14 @@
             android:key="@string/use_inexact_seek_key"
             android:summary="@string/use_inexact_seek_summary"
             android:title="@string/use_inexact_seek_title"/>
+
+        <ListPreference
+          app:iconSpaceReserved="false"
+          android:defaultValue="@string/seek_duration_default_key"
```
Stypox:
```suggestion
          android:defaultValue="@string/seek_duration_default_value"
```

app/src/main/res/xml/video_audio_settings.xml
```diff
@@ -146,5 +146,14 @@
...
+          android:entries="@array/seek_duration_description"
+          android:entryValues="@array/seek_duration_value"
+          android:key="@string/seek_duration_key"
+          android:summary="%s"
```
Stypox: Are you sure this is ok? According to [this thread](https://stackoverflow.com/questions/7017082/change-the-summary-of-a-listpreference-with-the-new-value-android) the implementation is buggy

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -954,12 +954,19 @@ public void onPlayPause() {
...
     public void onFastForward() {
         if (DEBUG) Log.d(TAG, "onFastForward() called");
-        seekBy(FAST_FORWARD_REWIND_AMOUNT_MILLIS);
+        seekBy(getSeekDuration());
+    }
+
+    private int getSeekDuration() {
+        final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        final String key = context.getString(R.string.seek_duration_key);
+        final String value = prefs.getString(key, FAST_FORWARD_REWIND_DEFAULT_AMOUNT_MILLIS);
```
Stypox: You could remove `FAST_FORWARD_REWIND_DEFAULT_AMOUNT_MILLIS`, since it is a duplicate of the `seek_duration_default_value` string.
```suggestion
        final String value = prefs.getString(key, context.getString(R.string.seek_duration_default_value));
```

app/src/main/res/values/settings_keys.xml
TobiGr: Correct. Good catch :+1:

app/src/main/res/xml/video_audio_settings.xml
XiangRongLin: To be honest, i think i just copied it from the resolution preference [here](https://github.com/TeamNewPipe/NewPipe/blob/915455d4e5023083e2e2855c52abea17982c291d/app/src/main/res/xml/video_audio_settings.xml#L13). 
It seems to at least work on my API 25 and 29 devices and a 21 emulator.

Stypox: Oh ok, thanks ;-)

TobiGr:
Nice :) Thanks for the review @Stypox 
@XiangRongLin Can you please rebase one last time? I'd like to include your changes in the next release 0.18.0 (#2862)

XiangRongLin:
@TobiGr Rebased it

Issue #2351, opened 2019-05-21
Share video-links as Invidious urls


pred2k:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=74462002)](https://www.bountysource.com/issues/74462002-share-video-links-as-invidio-us-urls?utm_source=74462002&utm_medium=shield&utm_campaign=ISSUE_BADGE)

I imagine a setting to generate `invidio.us/watch?v=...` urls instead of `youtube.com/watch?v=...` when using the share function in NewPipe.

Was also already mentioned [here](https://github.com/TeamNewPipe/NewPipe/issues/1867#issuecomment-481164439), but that Issue is about implementing video playing via Invidious.

As you know opening links will already be supported (#1900) with v0.16.2.

avoidr:
Also use invidio.us for "open in browser" function!

PopeRigby:
This is a good idea, but maybe have it as a user setting. People might get annoyed if they don't want to share an invidio.us link.

pred2k:
> This is a good idea, but maybe have it as a user setting. People might get annoyed if they don't want to share an invidio.us link.

Of course as a optional user setting.

And as extra maybe the possibility to choose between invidious instanes: https://github.com/omarroth/invidious/wiki/Invidious-Instances

PopeRigby:
> And as extra maybe the possibility to choose between invidious instanes

That's a good idea. You could have a field where you can type in your instance, which would be nice for #1163.

avoidr:
I also vote for invidio.us to be the default, with the option to pick youtube. No need to be so modest. invidious is superior technically and in terms of privacy and freedom.

PB020:
Is this issue/bounty still open?

Perflyst:
Yes

src-r-r:
Probably best to shut it down, since...
![image](https://user-images.githubusercontent.com/1633901/97822663-8cf9cb00-1cae-11eb-8c06-83a965d354c6.png)

pred2k:
> Probably best to close it, since...
> ![image](https://user-images.githubusercontent.com/1633901/97822663-8cf9cb00-1cae-11eb-8c06-83a965d354c6.png)
> Unless there's an alternative alternative?

invidio.us has shutdown, but there are other alternative working! instances. I would like leave the issue open and keep the bounty, just to promote these alternatives.

[...]

PR #2748, opened 2019-10-21
Custom share url (ex.: invidio instead of youtube)

XiangRongLin:

Added an option for user to select the way the url are generated for the "share" function. If the custom option is activated then the url will be generated with the given base.
For example: "https://invidio.us/watch?v=abcdefg" instead of "https://youtube.com/watch?v=abcdefg" 

By default turned off. When turned on the default custom website is "https://invidio.us", but can be changed.

Related issue: #2351

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -608,13 +608,13 @@ public class VideoDetailFragment
         switch (id) {
             case R.id.menu_item_share: {
                 if (currentInfo != null) {
-                    ShareUtils.shareUrl(this.getContext(), currentInfo.getName(), currentInfo.getOriginalUrl());
+                    ShareUtils.shareUrl(this.getContext(), currentInfo.getName(), currentInfo.getUrl());
                 }
                 return true;
             }
             case R.id.menu_item_openInBrowser: {
                 if (currentInfo != null) {
-                    ShareUtils.openUrlInBrowser(this.getContext(), currentInfo.getOriginalUrl());
+                    ShareUtils.openUrlInBrowser(this.getContext(), currentInfo.getUrl());
                 }
                 return true;
             }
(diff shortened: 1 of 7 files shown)


XiangRongLin:
I just realised that there is a SoundCloud and MediaCCC section. Will need to check if those sections still work.

Stypox:
This probably needs to be changed in the extractor: you would need to add a command to every extractor (add a base method) that returns all of the possible share URLs. Those share URLs would extend a base share-URL class that has the methods to get a representation of the URL to show to the user and to get the real URL based on the ID.

XiangRongLin:
So after some digging I found out, that the extractor already has the functionality to extract the id. It gets saved into [Info](https://github.com/TeamNewPipe/NewPipeExtractor/blob/50a17afc056cfc9ccb14a72f3ff5a944d6f9418d/extractor/src/main/java/org/schabi/newpipe/extractor/Info.java#L17), but not into [InfoItem](https://github.com/TeamNewPipe/NewPipeExtractor/blob/50a17afc056cfc9ccb14a72f3ff5a944d6f9418d/extractor/src/main/java/org/schabi/newpipe/extractor/InfoItem.java). 
In most cases i have access to a subclass of Info, but there are 2 cases where i only have access to a subclass of InfoItem. [1](https://github.com/TeamNewPipe/NewPipe/blob/bf2a3ca04326442b32a139c1cd94c30596c000e2/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.java#L430) [2](https://github.com/TeamNewPipe/NewPipe/blob/bf2a3ca04326442b32a139c1cd94c30596c000e2/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java#L41). I'm not too sure how to solve it for these cases, without extracting it from the url.

Also my current way breaks sharing of channels, playlists.

pred2k:
How about having that feature **enabled** by default?

TobiGr:
I agree with @Stypox. A proper implementation of this feature requires extractor changes. IMO, this would also require us to list supported URLs in the [YouTubeParsingHelper](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/linkHandler/YoutubeParsingHelper.java). We only use checks whether an URL is a YouTube/Invidio/HooktubeURL. The domains used in these methods should be moved to a list.

XiangRongLin:
To be honest, i can't figure it out where and how to excatly start with changing that in the extractor.

Stypox:
If I find some time in the next days I will try to give you some directions, thank you for the effort ☺

Z3ROGRAV1TYx:
Awesome!

PR #2761, opened 2019-10-25
Fix what Weblate screwed up

enolp:
What it does?

This commit fixes that bad handling of deleted strings on Weblate, all it does is to delete all empty AND non-closed tags, and update translations which I previously uploaded to Weblate.

PR completely safe to accept. Regretting this PR and sync translations with Weblate WON'T fix this problem, it occurred to me on other proyects.


TobiGr:
Did you report the problem to Webate?

enolp:
Nope but it's a good idea

Issue #799, opened 2017-10-30
Parcel too large in Notification


coffeemakr:
Serious issue that keeps occurring.

Steps to reproduce unknown.

Exception example: `java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 570604 bytes` thrown in notificationManager.notify (BackgroundPlayer.updateNotification)

https://sentry.schabi.org/sentry/newpipe/issues/49/

Some user comments:
- Had to fast forward a bunch of times
- listening to music in the background and browsing in chrome. nothing extraordinary
- this issue occurs oftenly since I updated to nougat 7.1... (phone is ZTE Axon 7 with MiFavor overlay), even when the app is freshly restarted

theScrabi:
Uch nasty one :/

karyogamy:
This should be fixed in #680. Indeed, it is a nasty bug. This is because we are not recreating the notification when updating them, which introduces leaks when we update to frequently (especially with a large bitmap). Most other music apps don't update notifications like us, where the time progress is being updated every second. This problem only occurs when you have left the screen on for a long time though AND the notification has been updated a bunch of times. If you leave the screen on for long enough with music playing and look at logcat, you will see the system is trying to gc but to no avail, and eventually crash with parcel to large. 

This is solved by simply recreating the notification builder every time it is being updated, see [BackgroundPlayer](https://github.com/karyogamy/NewPipe/blob/playlist/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java).

coffeemakr:
> This  is because we are not recreating the notification when updating them, which introduces leaks when we update to frequently (especially with a large bitmap).

That's what I thought. 

>  is solved by simply recreating the notification builder

Alright. So commit bd9ee18e562b00b73640525414d771f19ca84f05 and 86c16fa5d8397aca49813d2e2e158a36eb526a6a should fix this. Thanks.

PR #2771, opened 2019-10-30
Limit amount of notification thumbnail updates in background player

### LINKED ISSUE #2170: Lag when video plays in background.

atpamat:
Maybe not the cleanest solution but it seems to fix #2170 while recreating notification often enough to prevent leaks.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
@@ -95,6 +95,9 @@ public final class BackgroundPlayer extends Service {
 
     private boolean shouldUpdateOnProgress;
 
+    private static final int NOTIFICATION_UPDATES_BEFORE_RESET = 60;
+    private int timesNotificationUpdated;
+
     /*//////////////////////////////////////////////////////////////////////////
     // Service's LifeCycle
     //////////////////////////////////////////////////////////////////////////*/
@@ -180,6 +183,7 @@ public final class BackgroundPlayer extends Service {
 
     private void resetNotification() {
         notBuilder = createNotification();
+        timesNotificationUpdated = 0;
     }
 
     private NotificationCompat.Builder createNotification() {
@@ -252,6 +256,7 @@ public final class BackgroundPlayer extends Service {
             if (bigNotRemoteView != null) bigNotRemoteView.setImageViewResource(R.id.notificationPlayPause, drawableId);
         }
         notificationManager.notify(NOTIFICATION_ID, notBuilder.build());
+        timesNotificationUpdated++;
     }
 
     /*//////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


Redirion:
I think you are at the right spot with the resetNotification which is called unnecessarily often due to onUpdateProgress being called every second. But you are also right that your approach is not the cleanest solution.

What about wrapping a resetNotification(); in onUpdateProgress with:
`if(notBuilder == null || bigNotRemoteView == null || notRemoteView == null) {
    resetNotification();
}`

resetNotification should not be required to be called at all in this method, there are plenty of other occasions where it is also called.

atpamat:
> What about wrapping a resetNotification(); in onUpdateProgress with:
> `if(notBuilder == null || bigNotRemoteView == null || notRemoteView == null) {`
> ` resetNotification();`
> `}`
> 
> resetNotification should not be required to be called at all in this method, there are plenty of other occasions where it is also called.

Unfortunately resetNotification has to be called every few calls to notificationManager.notify(). Refer to this discussion https://github.com/TeamNewPipe/NewPipe/pull/1623#discussion_r212509777 and issue #799.

Redirion:
the comment you have highlighted did not refer to onUpdateProgress but another method. As I said, there are plenty of other occasions where this is called. Compare to PopupPlayer for example, that doesn't even have an onUpdateProgress but also the other occasions where onResetNotification is used.
Maybe it's worth trying?

atpamat:
It didn't refer to onUpdateProgress but it's there for the same reason. onUpdateProgress is called twice every second while all the other occurrences of resetNotification are called after user interaction, or video change. With resetNotification removed from onUpdateProgress the UI starts lagging in longer videos and eventually results in the same exception as mentioned in #799.
PopupPlayer doesn't have onUpdateProgress because it doesn't display time in the notification.

Redirion:
> edit: I should have probably mentioned that I've actually tried running it with resetNotification removed and few minutes into the video scrolling comments started lagging very badly and eventually whole UI froze.

thanks for the clarification. In this case I think you might even consider increasing NOTIFICATION_UPDATES_BEFORE_RESET to a value like 120 instead of 20?

atpamat:
I don't want to set it too high because I don't know how will it behave on other devices. 120 works fine on mine, with 240 scrolling comments gets very slightly laggy towards the end of second minute, if I set higher values then around 3:30 lags get more noticeable and at 5 mins I don't think it can be still called scrolling. But I think something around 60 (half a minute) shouldn't be a problem.

atpamat:
I rebased and amended the commit with increased NOTIFICATION_UPDATES_BEFORE_RESET.

TobiGr:
@mauriciocolli Do you have time to review this? I'd like to merge #2690, but we need to get rid of the notification thumbnail update spam first because #2690 uses the same logic to determine when to set the thumbnail on the lock screen.

TobiGr:
I just realized that the pop-up/mini-player has a notification, too. Is it possible to apply this change to it?

atpamat:
> I just realized that the pop-up/mini-player has a notification, too. Is it possible to apply this change to it?

PopupPlayer doesn't call `resetNotification()` in `onUpdateProgress()` which is executed twice every second and I believe all other occurrences of `resetNotification()` are caused either by user input or relatively rare events. By mini-player you used a different word for pop-up or is there something else?

Issue #2392, opened 2019-06-11
[Feature Request] Option to choose default caption


MF-Debug:
an option to choose default language to show as caption whenever a video is playing would be great. thanks!

B0pol:
Closed by #2820

Issue #1329, opened 2018-04-18
[Feature Request] Set subtitles/CCs on by default


plusuncold:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

A quite nice (and likely easy) feature would be to be able to set a default subtitle/CC language and if the video has them available to show them by default.

Sirove:
+1

juanfra684:
If possible, please add also the option to enable the subtitles for different languages including the auto-generated subtitles. For example, the user could sets the languages and the priority: English, Spanish, Auto-generated(EN).

juanfra684:
Bump.

B0pol:
Closed by #2820

PR #2820, opened 2019-11-24
Remember caption option in player, closes #2811

### LINKED ISSUE #2811: Remember caption option in player
- [✔️] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [✔️] I checked if the issue/feature exists in the latest version.
- [I don't have json bugreport since this is feature request so ❌] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

Currently it forgets and always turn off by default. Quite annoying because i have to turn on again and again
Can you make it remember when i turned it on and remember language? Doesn't matter if it's english or english (auto generated)

Redirion:
There had been feature requests like #1329 and #2392.
But instead of adding a new preference which seems to be generally not favourable, the request in #2811 seems actually pretty straight forward.
This PR implements it.

I have also considered the auto-generated subtitles (and hopefully internationalized, I have tested english and german phone languages). If auto-generated language was chosen last time by the user, the non auto-generated language will be preferred, but if the native chosen language is not available, the auto-generated variant will be used as a fallback.

Debug apk for testing: https://github.com/Redirion/NewPipe/releases/tag/v0.17.4cc


app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
```diff
@@ -285,6 +288,12 @@ private void buildCaptionMenu(final List<String> availableLanguages) {
         if (captionPopupMenu == null) return;
         captionPopupMenu.getMenu().removeGroup(captionPopupMenuGroupId);
 
+        String userPreferredLanguage = PreferenceManager.getDefaultSharedPreferences(context)
+                .getString(context.getString(R.string.caption_user_set_key), null);
+        // english (auto-generated)
+        boolean searchForAutogenerated = userPreferredLanguage == null ||
+                !userPreferredLanguage.contains("(");
```
TobiGr: What does `!userPreferredLanguage.contains("(")` do? Please add a comment to ensure people can understand your code later.

app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
```diff
@@ -308,9 +319,26 @@ private void buildCaptionMenu(final List<String> availableLanguages) {
...
                     trackSelector.setParameters(trackSelector.buildUponParameters()
                             .setRendererDisabled(textRendererIndex, false));
+                    final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+                    prefs.edit().putString(context.getString(R.string.caption_user_set_key),
+                            captionLanguage).commit();
                 }
                 return true;
             });
+            // apply caption language from previous user preference
+            if(userPreferredLanguage != null && (captionLanguage.equals(userPreferredLanguage) ||
```
TobiGr:
```suggestion
            if (userPreferredLanguage != null && (captionLanguage.equals(userPreferredLanguage) ||
```

app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
Redirion: if the user already has set the auto-generated language (hence the check for "("), it is not useful to look for that as the first part of the if condition would match the caption language directly.

TobiGr: Thanks

test.diff
```diff
@@ -0,0 +1,20 @@
+diff --git a/app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java b/app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
```
TobiGr: Huh :smile: 
The diff looks a little lost here :D

Redirion: ssht. Nothing to see here :D but you might want to check #2569 as another nice little addition for Android9+ devices

TobiGr: I didn't see anything :D
I'll take a look at #2569 soon. There are so many open PRs at the moment... I need more time and help to review them all

TobiGr: PS: you can squash the commits and nobody will ever know about that file :)

Issue #2759, opened 2019-10-25
Playlist Thumbnail is not updated


532910:
Steps to reproduce:

1. Select item, `Add To Playlist` -> `New Playlist` -> `test1`
(Item Thumbnail will be used as Playlist Thumbnail)

2. Go to `test1` playlist, `Delete` item.
(Playlist Thumbnail will remain, but it should be removed)

3. Select item, `Add To Playlist` -> `test1`
(Playlist Thumbnail will remain, but it should be updated)

It's a dup of #1330 and the expected behaviour is to update Thumbnail. The current behaviour confuses users (I've done tests on four different users.)

- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.
- [x] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

Stypox:
:thinking: yeah, this could be improved

TobiGr:
Maybe we can add a default thumbnail. But why don't we delete playlists with zero items?

decarvalhobo:
Hello! Keeping a playlist with zero elements can be useful when you want to wipe the videos from it in order to replace them with most recent ones. But indeed there is no many use cases.

I am interested in contributing on this project. Can I work on this issue?

532910:
The most used zero-elements playst for me is "watch later".

Stypox:
@decarvalhobo yeah, nobody is working on this AFAIK. If you have any question feel free to ask!
If you want you can take a look at the other "good first issue" issues, so that you can familiarize with the code. ;-)

TobiGr:
Fixed in #2837

PR #2837, opened 2019-12-02
Playlist thumbnail updates

decarvalhobo:

I propose this solution for the issue #2759.
It corrects the use-case described in this issue : 

1. Select item, Add To Playlist -> New Playlist -> test1
    (Item Thumbnail will be used as Playlist Thumbnail)

2. Go to test1 playlist, Delete item.
    (Playlist Thumbnail ~~will remain, but it should be removed~~ is updated with the default thumbnail url : https://i.ytimg.com/ if the playlist is empty)

3. Select item, Add To Playlist -> test1
    (Playlist Thumbnail ~~will remain, but it should be updated~~ is updated with this item thumbnail if playlist was empty)

If the playlist isn't empty in step 2, we just update the playlist thumbnail with one of its item.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
@@ -413,10 +413,24 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
         disposables.add(disposable);
     }
 
+    private void updateThumbnailUrl() {
+        String newThumbnailUrl;
+
+        if(!itemListAdapter.getItemsList().isEmpty()){
+            newThumbnailUrl = ((PlaylistStreamEntry)itemListAdapter.getItemsList().get(0)).thumbnailUrl;
+        }
+        else newThumbnailUrl = "https://i.ytimg.com/";
+
+        changeThumbnailUrl(newThumbnailUrl);
+    }
+
     private void deleteItem(final PlaylistStreamEntry item) {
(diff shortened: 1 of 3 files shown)


Stypox:
Wouldn't it be better to use an Android drawable instead of using a remote image? Just like the thumbnail used for videos before the actual thumbnail loads

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -403,10 +403,24 @@ private void changeThumbnailUrl(final String thumbnailUrl) {
         disposables.add(disposable);
     }
 
+    private void updateThumbnailUrl() {
+        String newThumbnailUrl;
+
+        if(!itemListAdapter.getItemsList().isEmpty()){
+            newThumbnailUrl = ((PlaylistStreamEntry)itemListAdapter.getItemsList().get(0)).thumbnailUrl;
+        }
+        else newThumbnailUrl = "https://i.ytimg.com/";
```
TobiGr: Are we legally allowed to use this image? We need to check its license first. Despite the license issue, the image should be a local drawable as @Stypox pointed out correctly.

decarvalhobo:
In this last commit I use the existing drawable (called:  dummy_thumbnail_playlist) instead of the remote image as you proposed. Is it ok or should I use an  other drawable than this one?

decarvalhobo:
I'm not sure about my getThumbnail function due to the usage of blockingFirst. How should I get the thumbnail in a more correct way (without blocking instructions)? Thanks

Issue #2838, opened 2019-12-03
Bad behavior when downloading videos through native YouTube app share function


exegg:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.
It does
- [ ] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.
It isn't a crash.

I will post a video since I think it is self explanatory, but I'll summary the few issues.

https://streamable.com/4jyw7

- After selecting a video for download, NewPipe stays opened, frozen, with the youtube screen you had opened (00:23). You have to manually switch from NewPipe to YouTube again.

- After downloading a video, when choosing another one, it brings again the latest video you already downloaded instead the one you are fetching. (00:37). You have to repeat the steps in order to get the actual video you want to download.
Ironically when you cancel, the floating window disappears correctly (so NewPipe doesn't stay open, frozen) and you can continue to browse YouTube.

This behavior wasn't present in past versions. I've been using this feature for as long as I can remember using the app.
I'm using the latest stable version from Github.
Samsung Galaxy J7 2018 (j737a) with android 8.0

cool-student:
This could be related to https://github.com/TeamNewPipe/NewPipe/issues/2773 it sounds like you experience problems with the downloading process.

exegg:
Well yes, it is exactly the same issue described by gdevenyi on that issue. The issue is not present in any previous versions as far as I know it.

Nico-late:
I found a way to solve this issue together with issue #2773  and issue #2797 . Youtube seems to freeze as the NewPipe activity is not correctly closed. I added an if statement in DownloadDialog to close the activity when it's not coming from the app itself: 
```java 
toolbar.setOnMenuItemClickListener(item -> {
            if (item.getItemId() == R.id.okay) {
                prepareSelectedDownload();
                if (getActivity() instanceof RouterActivity){
                    getActivity().finish();
                }
                return true;
            }
            return false;
        });
```
 As it finishes the activity, when downloading a second video, the name is reset so it also fixes issue #2797 .

Issue #2773, opened 2019-11-02
application17.4 freezes when downloading audio


jediafr:
sorry to report that i have foremost issue with the new update : the application now freezes when downloading audio (with the share feature).
How to reproduce :
Share a link and select Newpipe.
The download dialog appears and i select audio (i leave the default options for the encoding).
Newpipe then starts downloading but freezes. (previous 17.3 works fine with the same link).

Had to rollback to 17.3 for the time being.

- [ ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I checked if the issue/feature exists in the latest version.
- [ ] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

gdevenyi:
Just experienced the same bug, but I'll add a little more detail.

The UI interface seems to get hung somehow. The download proceeds correctly, but the interface is hung and I need to switch applications to get back to where I was before.

In addition to this but probably related, when I share another link to the newpipe target the original share populates the download dialog. I have to share twice to get the new link.

PR #2858, opened 2019-12-11
Solving issue #2797 by changing RouterActivity

### LINKED ISSUE #2797: Downloading Audio Naming Problem When Shared from YouTube
- [✔] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [✔] I checked if the issue/feature exists in the latest version.
- [✔] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

on version 0.17.4 when sharing a video from youtube to newpipe to download as audio... on the first try the file name that shows up is the file name from the previously downloaded audio... upon the second try to download it then gets it correct name... happen for all downloads... if i remember correctly i think the first time using this version when dwnloading using my method the first file name came up empty..

pchebel:

I managed to fix #2797 which caused the video downloaded by sharing from the YouTube application to be the wrong one.

The problem can be reproduced by following these steps:
1) Downloading a video by sharing from the YouTube app
2) Switching back to YouTube app
3) Sharing a new YouTube video to NewPipe

After adding a OnNewIntend method in RouterActivity, the problem is gone.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -119,6 +119,14 @@ public class RouterActivity extends AppCompatActivity {
         Icepick.saveInstanceState(this, outState);
     }
 
+
+    @Override
+    protected void onNewIntent(Intent intent) {
+        super.onNewIntent(intent);
+
+        currentUrl = getUrl(intent);
+    }
+
     @Override
     protected void onStart() {
         super.onStart();


cool-student:
I believe this could also be the fix for these issues: https://github.com/TeamNewPipe/NewPipe/issues/2773 https://github.com/TeamNewPipe/NewPipe/issues/2838

pchebel:
My change only fixes the problem with the downloaded video but it doesn't fix the freezing issue.

So the issue #2773 isn't solved and #2838 is ony partially solved

TobiGr:
Closing in favor of #2875. Thank you again!

PR #2868, opened 2019-12-14
fix add instance dialog

yausername:

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/PeertubeInstanceListFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/PeertubeInstanceListFragment.java
@@ -5,6 +5,7 @@ import android.content.Context;
 import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
+import android.text.InputType;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/settings/PeertubeInstanceListFragment.java
```diff
@@ -203,17 +204,18 @@ private void initButton(View rootView) {
 
     private void showAddItemDialog(Context c) {
         final EditText urlET = new EditText(c);
+        urlET.setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_URI);
```
TobiGr: We could also add the no suggestions flag to disable auto correction when typing the url. But I guess most people will copy their instance urls and this flag disables the option to paste the first clipboard item from the keyboard.
What's your opinion? @Stypox @yausername 
```suggestion
        urlET.setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_URI);
        urlET.setInputType(InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS);
```

Stypox: I assume most keyboards will stop suggesting things when they see a url text field. Mozilla Preview (fenix) uses [the same flags](https://github.com/mozilla-mobile/fenix/blob/300d8317e1e0726fd66ea18f73543d0dafd3bdd1/app/src/main/res/layout/fragment_edit_bookmark.xml#L59) in the bookmark editor: `textUri` means `InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_URI` ([source](https://developer.android.com/reference/android/widget/TextView.html#attr_android:inputType))

yausername: Auto correction is automatically disabled on `TYPE_TEXT_VARIATION_URI`. I don't see any harm is having suggestions on because many instances use dictionary words.

TobiGr: Ok. I this case my keyboard is behaving shitty :)
When typing, I still get auto-corrected. Never mind, I tested with another keyboard and things work as expected.

PR #2871, opened 2019-12-15
make main page tabs scrollable and hide when there is only a single tab

### LINKED ISSUE #1647: One Item No Tabs in v0.14.0
If you go to Settings->Appearance->Content Of Main Page and only have one item. It would be nice to remove the tabs from the main page to increase the size since there is only one item and the one tab always selected takes up space on the screen.

atpamat:

This PR makes tabs scrollable so they don't get squeezed if there are too many of them and hides tab selector if there is only a single tab. Adds and closes #1647 

Before | After
:---:|:---:
![single tab before][single-pre] | ![single tab after][single-post]
![multiple tabs before][multi-pre] | ![multiple tabs after][multi-post]

Created separate ScrollableTabLayout class to avoid cluttering MainFragment with switching between scrollable and fixed modes.

[single-pre]: https://user-images.githubusercontent.com/10105131/70863722-e647f700-1f42-11ea-95d5-17268d883ad6.jpg "Single tab"
[single-post]: https://user-images.githubusercontent.com/10105131/70863727-efd15f00-1f42-11ea-8ef8-d1edf501fa23.jpg "Single tab"
[multi-pre]: https://user-images.githubusercontent.com/10105131/70863729-f5c74000-1f42-11ea-9064-5aa63f83ffd0.jpg "Multiple tabs"
[multi-post]: https://user-images.githubusercontent.com/10105131/70863733-f9f35d80-1f42-11ea-9445-2ed78edb57fb.jpg "Multiple tabs"

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/views/ScrollableTabLayout.java
+++ b/app/src/main/java/org/schabi/newpipe/views/ScrollableTabLayout.java
@@ -14,6 +14,7 @@ import com.google.android.material.tabs.TabLayout.Tab;
 
 /**
  * A TabLayout that is scrollable when tabs exceed its width.
+ * Hides when there are less than 2 tabs.
  */
 public class ScrollableTabLayout extends TabLayout {
     private static final String TAG = ScrollableTabLayout.class.getSimpleName();
@@ -60,6 +61,13 @@ public class ScrollableTabLayout extends TabLayout {
     }
 
     private void resetMode() {
+        if (getTabCount() < 2) {
+            setVisibility(View.GONE);
+            return;
+        } else {
+            setVisibility(View.VISIBLE);
+        }
+
         if (getTabCount() == 0 || getTabAt(0).view == null) return;
         setTabMode(TabLayout.MODE_FIXED);
 
--- a/app/src/main/java/org/schabi/newpipe/views/ScrollableTabLayout.java
+++ b/app/src/main/java/org/schabi/newpipe/views/ScrollableTabLayout.java
@@ -25,18 +25,21 @@ public class ScrollableTabLayout extends TabLayout {
 
     public ScrollableTabLayout(Context context, AttributeSet attrs) {
         super(context, attrs);
+        setTabMode(TabLayout.MODE_FIXED);
     }
 
     public ScrollableTabLayout(Context context, AttributeSet attrs, int defStyleAttr) {
         super(context, attrs, defStyleAttr);
+        setTabMode(TabLayout.MODE_FIXED);
     }
 
     @Override
-    protected void onAttachedToWindow() {
-        super.onAttachedToWindow();
+    protected void onLayout(boolean changed, int l, int t, int r, int b) {
+        super.onLayout(changed, l, t, r, b);
 
-        setTabMode(TabLayout.MODE_FIXED);
-        resetMode();
+        if (changed) {
+            resetMode();
+        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/views/ScrollableTabLayout.java
```diff
@@ -0,0 +1,91 @@
...
+        int actualWidth = 0;
+        for (int i = 0; i < getTabCount(); ++i) {
+            if (getTabAt(i).view == null) return;
+            actualWidth += ((View) getTabAt(i).view).getWidth();
+            if (actualWidth > layoutWidth) {
+                setTabMode(TabLayout.MODE_SCROLLABLE);
+                return;
+            }
+        }
+    }
```
Stypox: Could you add some comments to this function? For example explaining why `getTabAt(0).view` could be `null` and why you check for width twice (I assume you first do a fast check to improve performance?).

app/src/main/java/org/schabi/newpipe/views/ScrollableTabLayout.java
```diff
@@ -0,0 +1,91 @@
...
+        } else {
+            setVisibility(View.VISIBLE);
+        }
+
+        if (getTabCount() == 0 || getTabAt(0).view == null) return;
+        setTabMode(TabLayout.MODE_FIXED);
+
+        int layoutWidth = getWidth();
+        int minimumWidth = ((View) getTabAt(0).view).getMinimumWidth();
+        if (minimumWidth * getTabCount() > layoutWidth) {
```
Stypox: Aren't you assuming here that the width of every tab is higher than or equal to the width of the first tab?

Take this example:
- there are two tabs of 500 and 300 pixels respectively: ----- ---
- the layout width is 900 pixels: --------

In this case the layout would be made scrollable (since 500*2 > 900), even though the views fit (500+300 <= 900).

I would remove the first layout width check completely, even though it could sometimes increase performance. What do you think? Another solution (that would solve this performance problem) would be to just keep track of the sum of the minimum size of every view by adding in `addTab()` and subtracting in `removeTabAt()`, but I don't know if the widths there would be correct.

atpamat: I checked implementation of `View` and `getWidth()` is constant time so I can remove this.

As for why I've done that - I assumed that `minWidth` of View wouldn't be changed by anything besides parent element (`TabLayout` in this case) so that every tab would have same value but that might not be correct. I would have used `getTabMinWidth()` but it's private. I could try reading it the same way it's done by `TabLayout`:
```java
requestedTabMinWidth =
        a.getDimensionPixelSize(R.styleable.TabLayout_tabMinWidth, INVALID_WIDTH);
```

I don't think it can be done in `addTab()` if we assume that width of some tab may later change.

app/src/main/java/org/schabi/newpipe/views/ScrollableTabLayout.java
atpamat: `getTabAt()` has a `@Nullable` annotation. I'll make this function a little cleaner and account for possible null tabs differently.

atpamat: Nevermind, it returns `null` if index exceeds `getTabCount()`.

There were no checks for whether `Tab.view` is null in some `Tab`'s public methods modifying `view` so I think I can assume that it's not null?

atpamat:
Version `1.1.0-alpha6` of material components adds `MODE_AUTO` to `TabLayout.mode` so once there is a release version and NewPipe migrates to it most of the functionality of `ScrollableTabLayout` will be covered by `TabLayout` (except for auto hiding when there are less than two tabs).

app/src/main/java/org/schabi/newpipe/views/ScrollableTabLayout.java
atpamat: I added a few comments throughout the class and simplified the calculation of tabs' width. I also fixed my mistake where width was calculated based only on minimum width requested by tab even if it's contents were wider.

TobiGr:
The changes look reasonable to me. @Stypox You began to review this. Can you please take a final look?

app/src/main/java/org/schabi/newpipe/views/ScrollableTabLayout.java
```diff
@@ -0,0 +1,128 @@
...
+        } else {
+            setVisibility(View.GONE);
+        }
+    }
+
+    /**
+     * Calculate minimal width required by tabs and set tabMode accordingly
+     */
+    private void remeasureTabs() {
+        if (getVisibility() != View.VISIBLE) return;
```
Stypox: Couldn't you just use `prevVisibility` here?

atpamat: sure, done

PR #2905, opened 2019-12-28
remove TLS 1.1 support for kitkat

### LINKED ISSUE #2898: Deprecate TLS 1.1

#2792 enabled TLS 1.1 and TLS 1.2. However, there are concerns that TLS 1.1 is vulnerable. The whole web has already started phasing out TLS 1.1. If left enabled, it makes it easy to execute a downgrade attack.

Are we sure that we need both TLS 1.1 and TLS 1.2? If we only need the latter then it's better to have only TLS 1.2 enabled.

mqus:

partly fixes #2898 

I wasn't able to test this yet because I'm not at my usual workstation, the soonest I'll be able to is in ~1 week.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/TLSSocketFactoryCompat.java
+++ b/app/src/main/java/org/schabi/newpipe/util/TLSSocketFactoryCompat.java
@@ -18,7 +18,7 @@ import static org.schabi.newpipe.MainActivity.DEBUG;
 
 /**
  * This is an extension of the SSLSocketFactory which enables TLS 1.2.
- * Created for usage on Android 4.1-4.4 devices, which haven't enabled those by default.
+ * Created for usage on Android 4.1-4.4 devices, which do not have it enabled by default.
  */
 public class TLSSocketFactoryCompat extends SSLSocketFactory {
 
--- a/app/src/main/java/org/schabi/newpipe/util/TLSSocketFactoryCompat.java
+++ b/app/src/main/java/org/schabi/newpipe/util/TLSSocketFactoryCompat.java
@@ -17,7 +17,7 @@ import static org.schabi.newpipe.MainActivity.DEBUG;
 
 
 /**
- * This is an extension of the SSLSocketFactory which enables TLS 1.2 and 1.1.
+ * This is an extension of the SSLSocketFactory which enables TLS 1.2.
  * Created for usage on Android 4.1-4.4 devices, which haven't enabled those by default.
  */
 public class TLSSocketFactoryCompat extends SSLSocketFactory {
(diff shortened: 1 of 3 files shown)


Redirion:
I don't think that this PR is necessary yet. As you pointed out in your comment https://github.com/TeamNewPipe/NewPipe/issues/2898#issuecomment-569456396 the browsers will actually drop TLS 1.1 support in March 2020. 
See [here](https://webkit.org/blog/8462/deprecation-of-legacy-tls-1-0-and-1-1-versions/) and [here](https://blog.mozilla.org/security/2018/10/15/removing-old-versions-of-tls/)

We should then come back and remove TLS below 1.2 for all devices alltogether.

TobiGr:
Thanks @mqus. 
I agree with @Redirion. We'll postpone this until March and then drop TLS 1.1 completely.

app/src/main/java/org/schabi/newpipe/util/TLSSocketFactoryCompat.java
```diff
@@ -17,7 +17,7 @@
 
 
 /**
- * This is an extension of the SSLSocketFactory which enables TLS 1.2 and 1.1.
+ * This is an extension of the SSLSocketFactory which enables TLS 1.2.
  * Created for usage on Android 4.1-4.4 devices, which haven't enabled those by default.
```
Stypox: Singular 
```suggestion
 * Created for usage on Android 4.1-4.4 devices, which do not have it enabled by default.
```

Stypox:
@TobiGr see https://github.com/TeamNewPipe/NewPipe/issues/2898#issuecomment-569456396
Is it really useful to disable TLS1.1 on Android 4, while every other user would keep 1.0 and 1.1 enabled? @mqus proposed a solution based on code from another project, maybe we should use that to disable 1.1 everywhere?

Redirion:
while waiting for the next update for Netcipher this PR could be changed in the following way:
remove the check for Kitkat, to use TLSSocketFactoryCompat for all devices and change `((SSLSocket) socket).setEnabledProtocols(new String[]{"TLSv1.2"});` to `((SSLSocket) socket).setEnabledProtocols(Build.VERSION.SDK_INT > 28 ? new String[]{"TLSv1.2", "TLSv1.3"} : new String[]{"TLSv1.2"});`as Android 10 added support for TLSv1.3

mqus:
Which does not really adress the set ciphersuites, which are also different between android versions. I seriously do not want to touch that, this will have to happen in another PR.

mqus:
But i agree that waiting for netcipher is not viable.

Stypox:
At which point is this? Should it be merged now? @Redirion @mqus

litetex:
Converted to draft:
* no progress/feedback since >9 months.
* no GitHub actions build
* merge conflicts

Should be closed in the future if no further progress happens here.
The only thing keeping it currently open is the fact that this is security related.

Redirion:
this is superseded by PR #7175

PR #2928, opened 2020-01-06
Show video preview (draft)

nv95:

Show video preview by long tab on thumbnail (I`ll use swipe gesture instead)

[proof of concept](https://www.dropbox.com/s/6poaxq4k5uvx3m6/device-2020-01-06-204220.mp4?dl=0)


Redirion:
I would like to have this feature in a future release. Could you update your PR to the current source?

nv95:
I completely update this PR, it works, but I got troubles with checkStyle. It fails on code from `dev` branch and also on `dev` branch itself

app/src/main/java/org/schabi/newpipe/info_list/holder/StreamGridInfoItemHolder.java
```diff
@@ -2,11 +2,17 @@
 
 import android.view.ViewGroup;
 
+import androidx.annotation.Nullable;
```
TobiGr, line 5: You also need to remove this unsued import:
```
Error: eckstyle] [ERROR] /home/runner/work/NewPipe/NewPipe/app/src/main/java/org/schabi/newpipe/info_list/holder/StreamGridInfoItemHolder.java:5:8: Unused import - androidx.annotation.Nullable. [UnusedImports]
```

TobiGr (commented):
That should help you: You have added a few lines to InfoListAdapter. That's why, you need to apply this:
``` diff
diff --git a/checkstyle-suppressions.xml b/checkstyle-suppressions.xml
index 2ff7ae68d7..0a304cb25e 100644
--- a/checkstyle-suppressions.xml
+++ b/checkstyle-suppressions.xml
@@ -9,7 +9,7 @@
 
   <suppress checks="FinalParameters"
     files="InfoListAdapter.java"
-    lines="253,325"/>
+    lines="255,327"/>
 
   <suppress checks="FinalParameters"
     files="ListHelper.java"

```

app/src/main/java/org/schabi/newpipe/info_list/holder/StreamGridInfoItemHolder.java
nv95: thanks)

litetex:
Needs rework / rebase as there seem to be build problems.
Converted to draft.

tsiflimagas:
I'd love to see this rebased and eventually in a release. It's such a cool feature!

litetex:
@tsiflimagas 
You may up-vote #6969 or similar issues 😄

Angelk90:
@nv95 : It doesn't go to see the video, but a series of screenshots, right?

nv95:
> a series of screenshots

Yes, this is a series of frames with changing animation

Angelk90:
@nv95 : It would have been nice if that had been the case #6969 , what do you think?
Or with an option in the settings that lets you choose which one to set.

Angelk90:
@litetex : Are there any news regarding this feature?

Angelk90:
@nv95 : Does this new feature also work well on the tablet?

Sagittarisus:
> @nv95 : Does this new feature also work well on the tablet?

Tested it on a tablet, does seem to work

Angelk90:
@Sagittarisus : Could you post a video of how it works?

app/src/main/java/org/schabi/newpipe/info_list/LivePreviewDispatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+            return
+        }
+        val last = target.drawable
+        val res = target.resources
+        target.startAnimation(createLoadingAnimation())
+        activePreview = Observable.zip(
+            RxStreamFrames.from(stream)
+                .buffer(3)
+                .map { x -> x.first() },
+            Observable.interval(500, TimeUnit.MILLISECONDS),
```
Stypox, line 46: Extract the `500` into a constant, it is a little bit cryptic there

app/src/main/java/org/schabi/newpipe/info_list/LivePreviewDispatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    private var activePreview: Disposable? = null
+    private var currentView: WeakReference<ImageView>? = null
+
+    fun show(target: ImageView, stream: StreamInfoItem) {
+        activePreview?.dispose()
+        if (!preferences.getBoolean(preferenceKey, true)) {
+            return
+        }
+        val last = target.drawable
```
Stypox, line 39: `last` is not a meaningful name for this variable. I'd call it `originalThumbnail`

app/src/main/java/org/schabi/newpipe/info_list/LivePreviewDispatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+                }
+            }.doFinally {
+                target.setImageSmooth(last)
+                currentView = null
+            }.subscribe(
+                { bitmap ->
+                    target.setImageSmooth(bitmap)
+                },
+                { throwable ->
+                    throwable.printStackTrace()
```
Stypox, line 65:
```suggestion
                    Log.w(TAG, "Error while loading frames into item thumbnail", throwable);
```

Stypox (changes requested):
Code looks mostly good; I tested a little bit and it seems to work fine.
However I think there are some things to change:
- A single tap on the thumbnail should still open the video, only the long press should be captured.
- If we decide to keep the setting, then when the setting is disabled the long press action should not be set at all. Currently it is set anyway, so when you long press on the thumbnail nothing happens, but the long press menu should open.
- If you long press on the thumbnail of an item where the live preview animation is already running, it should not restart from the beginning, but just continue normally.
- The frames preview should also be implemented for the feed list (see file `local/feed/item/StreamItem,kt`).

Thank you :-)

Angelk90:
@Stypox : 


> * Currently it is set anyway, so when you long press on the thumbnail nothing happens, but the long press menu should open.

For this point, for opening the menu why not do as youtube does which uses the following button (the one with the three dots) to open the menu.
It would be better to use long press to open the menu.

Schermata 2021-11-23 alle 10 56 57

P.s.
I would recommend using a bottomsheet menu instead of a modal.
image


- Instead of using a series of images, how about something similar like youtube does? #6969

Stypox:
>  It would be better to use long press to open the menu.
> I would recommend using a bottomsheet menu instead of a modal.

This is not the place to discuss this

> Instead of using a series of images, how about something similar like youtube does?

The data we have available is a series of images, and they are obtainable my requesting the full data of the video. Doing this for each video on the screen like youtube does would increase data usage too much, and playing the video in the thumbnail would be even worse. So no.

Also, please reduce the image size in your comments, use `` (see how I edited your comment). If you forget how to do that, like I do, just save a GitHub reply with that content.

app/src/main/java/org/schabi/newpipe/info_list/LivePreviewDispatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+                }
```
litetex, line 66: What is happening here?
Some documentation would be appreciated :)

app/src/main/java/org/schabi/newpipe/info_list/LivePreviewDispatcher.kt
```diff
@@ -0,0 +1,95 @@
...
+import io.reactivex.rxjava3.disposables.Disposable
+import io.reactivex.rxjava3.functions.BiFunction
+import io.reactivex.rxjava3.schedulers.Schedulers
+import org.schabi.newpipe.R
+import org.schabi.newpipe.extractor.stream.StreamInfoItem
+import org.schabi.newpipe.util.RxStreamFrames
+import java.lang.ref.WeakReference
+import java.util.concurrent.TimeUnit
+
+class LivePreviewDispatcher(context: Context) {
```
litetex, line 26: Some javadoc here would be great

app/src/main/java/org/schabi/newpipe/util/RxStreamFrames.kt
```diff
@@ -0,0 +1,71 @@
...
+
+import android.graphics.Bitmap
+import io.reactivex.rxjava3.core.Observable
+import io.reactivex.rxjava3.core.ObservableEmitter
+import io.reactivex.rxjava3.core.ObservableOnSubscribe
+import org.schabi.newpipe.database.stream.model.StreamEntity
+import org.schabi.newpipe.extractor.NewPipe
+import org.schabi.newpipe.extractor.stream.StreamInfoItem
+
+class RxStreamFrames(private val serviceId: Int, private val url: String) :
```
litetex, line 11: Some javadoc would be great

app/src/main/java/org/schabi/newpipe/util/RxStreamFrames.kt
```diff
@@ -0,0 +1,71 @@
...
+                                    )
+                                )
+                                if (emitter.isDisposed) {
+                                    return
+                                }
+                            }
+                        }
+                    }
+                }
+            }
```
litetex, line 46: What is happening here?
Some documentation would be appreciated :)

litetex (changes requested):
I have to make use of my veto here:
There is 0 comments and documentation inside the code.

app/src/main/res/xml/appearance_settings.xml
```diff
@@ -41,6 +41,13 @@
         app:singleLineTitle="false"
         app:iconSpaceReserved="false" />
 
+    <SwitchPreference
```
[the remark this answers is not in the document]
opusforlife2: If we remove the setting, then we should have a tooltip show up when the thumbnail is long-pressed so that the user doesn't get confused and think that the menu isn't working.

Something like "Long-press thumbnail to preview video".

Maybe add a visible service-coloured border to the thumbnail as well when the preview is playing to visually highlight it a little.

Also, I don't know what the current status is, but from the Dropbox video in the OP, I it seems that the preview keeps playing forever. I think if it isn't implemented yet, then there should be an obvious way to stop the preview. Maybe by tapping the thumbnail again, or by tapping the list item in general, or tapping elsewhere, or scrolling it out of the screen. Something to make it stop when you're done.

I'm assuming a standard use case will be people browsing their feed and previewing many many videos to decide what to watch. Does playing another preview cause the current one to stop? If not, you could have something like 5 previews playing at the same time, which will be messy.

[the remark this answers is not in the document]
Stypox: > If we remove the setting, then we should have a tooltip show up when the thumbnail is long-pressed so that the user doesn't get confused and think that the menu isn't working.

It would be useful regardless of whether we remove the setting or not

> it seems that the preview keeps playing forever.

No, it stops after the frames are finished

> Does playing another preview cause the current one to stop?

Yes

Angelk90:
@nv95 : is there any news?

nv95:
@Angelk90 I working on it, but have a lack of time(

Angelk90:
@nv95 : is there any news?

Angelk90:
@nv95 : What's still missing?

Stypox:
@SameenAhnaf I think 95% of this PR is completely unrelated to the view mode. Stream previews should work for every mode. But yeah, better to keep this closed until there is more activity.

PR #2935, opened 2020-01-08
Spelling: Could not, PeerTube, HTTPS, URL

comradekingu:


app/src/main/res/values/strings.xml
```diff
@@ -113,13 +113,13 @@
     <string name="service_title">Service</string>
     <string name="content_language_title">Default content language</string>
     <string name="peertube_instance_url_title">PeerTube instances</string>
-    <string name="peertube_instance_url_summary">Set your favorite peertube instances</string>
+    <string name="peertube_instance_url_summary">Set your favorite PeerTube instances</string>
```
Stypox: Maybe this is better?
```suggestion
    <string name="peertube_instance_url_summary">Select your favorite PeerTube instances</string>
```

Stypox:
Thanks!

PR #2938, opened 2020-01-08
Spelling: Language reworked 2

comradekingu:
Some savings, structure, consistency.


app/src/main/res/values/strings.xml
```diff
@@ -43,24 +43,24 @@
     <string name="controls_add_to_playlist_title">Add To</string>
     <string name="download_path_title">Video download folder</string>
     <string name="download_path_summary">Downloaded video files are stored here</string>
-    <string name="download_path_dialog_title">Choose the download folder for video files</string>
+    <string name="download_path_dialog_title">Choose download folder for video files</string>
     <string name="download_path_audio_title">Audio download folder</string>
     <string name="download_path_audio_summary">Downloaded audio files are stored here</string>
-    <string name="download_path_audio_dialog_title">Choose the download folder for audio files</string>
+    <string name="download_path_audio_dialog_title">Choose download folder for audio files</string>
     <string name="download_choose_new_path">Change the download folders to take effect</string>
```
comradekingu: I am not sure what happens here. Change appreciated.

app/src/main/res/values/strings.xml
```diff
@@ -585,7 +585,7 @@
     <string name="downloads_storage_ask_summary">You will be asked where to save each download</string>
     <string name="downloads_storage_ask_summary_kitkat">You will be asked where to save each download.\nChoose SAF if you want to download to an external SD card</string>
     <string name="downloads_storage_use_saf_title">Use SAF</string>
-    <string name="downloads_storage_use_saf_summary">The Storage Access Framework allows downloads to an external SD card.\nNote: some devices are not compatible</string>
+    <string name="downloads_storage_use_saf_summary">The Storage Access Framework allows downloads to an external SD card.\nNote: Some devices are incompatible</string>
```
comradekingu:
```suggestion
    <string name="downloads_storage_use_saf_summary">The \'Storage Access Framework\' allows downloads to an external SD card.\nNote: Some devices are incompatible</string>
```

app/src/main/res/values/strings.xml
snappyapple632: I'm assuming this is the prompt that asks which file you want to save downloaded audio in.

app/src/main/res/values/strings.xml
```diff
@@ -75,20 +75,20 @@
...
-    <string name="show_comments_summary">Disable to stop showing comments</string>
+    <string name="show_comments_summary">Turn off to hide comments</string>
     <string name="download_thumbnail_summary">Turn off to prevent loading thumbnails, saving data and memory usage. Changes clear both in-memory and on-disk image cache.</string>
     <string name="thumbnail_cache_wipe_complete_notice">Image cache wiped</string>
     <string name="metadata_cache_wipe_title">Wipe cached metadata</string>
     <string name="metadata_cache_wipe_summary">Remove all cached webpage data</string>
     <string name="metadata_cache_wipe_complete_notice">Metadata cache wiped</string>
     <string name="auto_queue_title">Auto-queue next stream</string>
-    <string name="auto_queue_summary">Auto-append a related stream when playing the last stream in a non-repeating queue</string>
+    <string name="auto_queue_summary">Auto-append a related stream (to non-repeating queue) after last file</string>
```
TobiGr: I suggest stream because we do not play files. Or did you replace stream to have no repetition?
```suggestion
    <string name="auto_queue_summary">Auto-append a related stream (to non-repeating queue) after last stream</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -138,9 +138,9 @@
     <string name="play_btn_text">Play</string>
     <string name="content">Content</string>
     <string name="show_age_restricted_content_title">Age restricted content</string>
-    <string name="video_is_age_restricted">Show age restricted video. Allowing such material is possible from Settings.</string>
-    <string name="duration_live">LIVE</string>
-    <string name="duration_live_button" translatable="false">LIVE</string>
+    <string name="video_is_age_restricted">Show age restricted video. Future changes are possible from \"Settings\".</string>
+    <string name="duration_live">Live</string>
+    <string name="duration_live_button" translatable="false">Live</string>
```
TobiGr: We should check why there are two strings with the same value and why one is not translatable.
Apart from that, we need to modify the .xml files to force capital letters.

app/src/main/res/values/strings.xml
```diff
@@ -224,15 +224,15 @@
...
-    <string name="restore_defaults_confirmation">Do you want to restore the defaults?</string>
+    <string name="restore_defaults_confirmation">Do you want to restore defaults?</string>
     <!-- error activity -->
     <string name="sorry_string">Sorry, that should not have happened.</string>
     <string name="guru_meditation" translatable="false">Guru Meditation.</string>
-    <string name="error_report_button_text">Report error via e-mail</string>
+    <string name="error_report_button_text">Report this per e-mail</string>
     <string name="error_snackbar_message">Sorry, some errors occurred.</string>
-    <string name="error_snackbar_action">REPORT</string>
+    <string name="error_snackbar_action">Report</string>
```
TobiGr: Same here, we need to alter the code.

app/src/main/res/values/strings.xml
```diff
@@ -529,7 +529,7 @@
     <string name="auto">Auto</string>
     <string name="switch_view">Switch View</string>
     <!-- App update notification -->
-    <string name="app_update_notification_content_title">NewPipe Update Available!</string>
+    <string name="app_update_notification_content_title">There is a fresh NewPipe version available!</string>
```
TobiGr: This is a notification title which should not be too long. Otherwise last part of it will be replaced by  "..."

app/src/main/res/values/strings.xml
TobiGr: Yes. NewPipe allows people to set different download folders for audio and video downloads.

app/src/main/res/values/strings.xml
comradekingu: The shortest alternative is "New NewPipe available" (?)

app/src/main/res/values/strings.xml
comradekingu: I meant the "Change the download folders to take effect"…?

app/src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="auto_queue_summary">Continue ending (non-repeating) playback queue by appending a related stream</string>
```

app/src/main/res/values/strings.xml
cool-student: imo "NewPipe Update Available!" sounds fine. This change should be reverted.

app/src/main/res/values/strings.xml
```diff
@@ -355,7 +355,7 @@
     <string name="github_url" translatable="false">https://github.com/TeamNewPipe/NewPipe</string>
     <string name="view_on_github">View on GitHub</string>
     <string name="donation_title">Donate</string>
-    <string name="donation_encouragement">NewPipe is developed by volunteers spending time bringing you the best experience. Give back to help developers make NewPipe even better while enjoying a cup of coffee.</string>
+    <string name="donation_encouragement">NewPipe is developed by volunteers spending time bringing you the best experience. Give back to help make NewPipe even better.</string>
```
cool-student:
```suggestion
    <string name="donation_encouragement">NewPipe is developed by volunteers spending their free time bringing you the best user experience. Give back to help developers make NewPipe even better while they enjoy a cup of coffee.</string>
```

removing the coffee here is imo not a good idea since the idea is that the cup of coffee IS THE DONATION also it's like that in other languages so it would be odd to only remove the cup of coffee here.

app/src/main/res/values/strings.xml
```diff
@@ -557,7 +557,7 @@
     <string name="error_file_creation">The file can not be created</string>
     <string name="error_path_creation">The destination folder can not be created</string>
     <string name="error_permission_denied">Permission denied by the system</string>
-    <string name="error_ssl_exception">Secure connection failed</string>
+    <string name="error_ssl_exception">Could not connect securely</string>
```
cool-student: I don't like this change

app/src/main/res/values/strings.xml
```diff
@@ -43,24 +43,24 @@
...
-    <string name="kore_not_found">Kore app not found. Install it?</string>
+    <string name="kore_not_found">Install missing Kore app?</string>
     <string name="kore_package" translatable="false">org.xbmc.kore</string>
     <string name="show_play_with_kodi_title">Show \"Play with Kodi\" option</string>
-    <string name="enable_lock_screen_video_thumbnail_title">Enable lock screen video thumbnail</string>
-    <string name="show_play_with_kodi_summary">Display an option to play a video via Kodi media center</string>
-    <string name="enable_lock_screen_video_thumbnail_summary">When using the background player a video thumbnail will be displayed on the lock screen</string>
+    <string name="enable_lock_screen_video_thumbnail_title">Lock screen video thumbnail</string>
+    <string name="show_play_with_kodi_summary">Displayed option to play a video via Kodi media center</string>
+    <string name="enable_lock_screen_video_thumbnail_summary">Using the background player displays a video thumbnail on the lock screen</string>
```
cool-student:
```suggestion
    <string name="enable_lock_screen_video_thumbnail_summary">A video thumbnail will be shown on the lock screen when using the background player</string>
```

I switched the sentence around so that there is more emphasis on the video thumbnail and less emphasis on the background player.

app/src/main/res/values/strings.xml
comradekingu: @cool-student The problem is with update/upgrade. In Debian that is an upgrade, but some are calling it an update.

cool-student: There is no way to determine if it's an Update or Upgrade since one would need to know what changed each version.

comradekingu: That is yet another way to interpret update/upgrade, when "new version" does away with the confusion.

app/src/main/res/values/strings.xml
comradekingu: Why? It is consistent with the other types of errors.

app/src/main/res/values/strings.xml
comradekingu: But then it becomes a garden path sentence. You have to read to the end to get an idea of what is going on.

app/src/main/res/values/strings.xml
Stypox: What about "New NewPipe version available!"

cool-student: I would argue that Update has enough expressive power especially when you consider that most people aren't power users and these casual users will understand what an Update is. There is no such difference of Update or Upgrade for the typical user.

app/src/main/res/values/strings.xml
cool-student: personally "Secure connection failed" sounds better to me but I guess this is subjective.

app/src/main/res/values/strings.xml
cool-student: I think that's the case either way since this sentence has 2 parts that the user needs to understand, so if the user reads only the first half they will be confused but that's the case in your version as well, that is if I stop reading after "Using background player" I as a User will be confused but then why wouldn't I read the whole sentence ? So I restructured the sentence from if(condition) this happens to this happens if (condition) imo perfectly understandable if you don't stop reading midway.

In any case I think that the original line was good too your new version imo makes the sentence slightly awkward unless the purpose was specifically to make it more concise due to size constraints on certain devices with small screens or something like that.

B0pol:
Can you change caption_setting_title from Captions to Subtitles ?
On YouTube image it's named subtitles, and even on [NewPipe website](https://newpipe.schabi.org/#downloads) we used subtitles.

app/src/main/res/values/strings.xml
comradekingu: @cool-student Sounding better or not, you can't really tell from that whether it is a current secure connection that fails, or establishing it that fails. Give that a shake through translation, and now it is the wrong one, only in a different language.

app/src/main/res/values/strings.xml
comradekingu: Because it is tax exempt? I think people understand who are users, and who are developers. I don't understand the psychology you are going for, could you elaborate?

app/src/main/res/values/strings.xml
comradekingu: So since I found a way to express it to all users, that some of them are not affected is a good reason to trip the others up? It is very plausible you could update something else in NewPipe, like the video lists, subscriptions, etc.

app/src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="auto_queue_summary">Auto-append a related stream (to non-repeating queue) after the last one</string>
```

comradekingu:
```suggestion
    <string name="auto_queue_summary">Auto-append a related stream after the last one</string>
```

app/src/main/res/values/strings.xml
cool-student: so the idea is that coffee is cheap and people buy themselves coffee every day so to help out the project they can contribute by giving the equivalent of a coffee as a donation. The user / developers thing is not that important here.

app/src/main/res/values/strings.xml
```diff
@@ -138,9 +138,9 @@
     <string name="play_btn_text">Play</string>
     <string name="content">Content</string>
     <string name="show_age_restricted_content_title">Age restricted content</string>
-    <string name="video_is_age_restricted">Show age restricted video. Allowing such material is possible from Settings.</string>
-    <string name="duration_live">LIVE</string>
-    <string name="duration_live_button" translatable="false">LIVE</string>
+    <string name="video_is_age_restricted">Show age restricted video. Future changes are possible from \"Settings\".</string>
```
TobiGr:
```suggestion
    <string name="video_is_age_restricted">Show age restricted video. Future changes are possible from the settings.</string>
```

app/src/main/res/values/strings.xml
TobiGr: Checked the usages and those changes are fine, because `android:textAllCaps="true"` is already used in XML resource files.

app/src/main/res/values/strings.xml
TobiGr: Please keep the old one, because this is text for a notification and there is only limited space for notification headings.

app/src/main/res/values/strings.xml
TobiGr: I like @cool-student's suggestion. Thanks

TobiGr:
@comradekingu Please grant us maintainers permission to the PR's branch so we can fix the last two things and merge afterwards.

app/src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="error_ssl_exception">Could not establish secure connection</string>
```

app/src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="app_update_notification_content_title">Install the fresh NewPipe version.</string>
```

app/src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="donation_encouragement">Treat the newPipe makers to something you like, by way of funding its development.</string>
```

app/src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="enable_lock_screen_video_thumbnail_summary">A video thumbnail is shown on the lock screen when using the background player</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -529,7 +529,7 @@
     <string name="auto">Auto</string>
     <string name="switch_view">Switch View</string>
     <!-- App update notification -->
-    <string name="app_update_notification_content_title">NewPipe Update Available!</string>
+    <string name="app_update_notification_content_title">Install the fresh NewPipe version.</string>
```
TobiGr: Please revert this change

app/src/main/res/values/strings.xml
```diff
@@ -355,7 +355,7 @@
     <string name="github_url" translatable="false">https://github.com/TeamNewPipe/NewPipe</string>
     <string name="view_on_github">View on GitHub</string>
     <string name="donation_title">Donate</string>
-    <string name="donation_encouragement">NewPipe is developed by volunteers spending time bringing you the best experience. Give back to help developers make NewPipe even better while enjoying a cup of coffee.</string>
+    <string name="donation_encouragement">Treat the newPipe makers to something you like, by way of funding its development.</string>
```
TobiGr: Please use @cool-student's version from https://github.com/TeamNewPipe/NewPipe/pull/2938#discussion_r364598892.

app/src/main/res/values/strings.xml
comradekingu: You got it, but it is not a direct instruction like elsewhere. It is also the type of tiring stuff one sees a lot of in Windows.

app/src/main/res/values/strings.xml
comradekingu: That idea doesn't work on people that have no relation to coffee. I agree that people spend more on things they hold dear, so it is indeed true the cost of a priority diminishes. My idea with that is to allude to it, and then play it into the added value by also tying it to NewPipe. 
"Treat the NewPipe makers to something you like, by way of funding its development."
Could personify "NewPipe makers" more, but I think this is better. Going with @cool-student version for now.

app/src/main/res/values/strings.xml
comradekingu: Done. However it adds two burdens on the user. One is the implied debt of having already received the software. That may work, but I don't like it.
The second is that of reading that long line of text. If done, that incurs an investment of time, which I also don't like, and really grinds my goat. It is not elegant. https://github.com/TeamNewPipe/NewPipe/pull/2938#discussion_r376792572

app/src/main/res/values/strings.xml
```diff
@@ -100,21 +100,21 @@
     <string name="enable_playback_state_lists_summary">Show playback position indicators in lists</string>
     <string name="settings_category_clear_data_title">Clear data</string>
     <string name="enable_watch_history_summary">Keep track of watched videos</string>
-    <string name="resume_on_audio_focus_gain_title">Resume on focus gain</string>
-    <string name="resume_on_audio_focus_gain_summary">Continue playing after interruptions (e.g. phone calls)</string>
+    <string name="resume_on_audio_focus_gain_title">Resumed playing</string>
```
Stypox: I think this should be reverted. It is missing the "focus gain" information

app/src/main/res/values/strings.xml
```diff
@@ -138,9 +138,9 @@
     <string name="play_btn_text">Play</string>
     <string name="content">Content</string>
     <string name="show_age_restricted_content_title">Age restricted content</string>
-    <string name="video_is_age_restricted">Show age restricted video. Allowing such material is possible from Settings.</string>
-    <string name="duration_live">LIVE</string>
-    <string name="duration_live_button" translatable="false">LIVE</string>
+    <string name="video_is_age_restricted">Show age restricted video. Future changes are possible from the settings.</string>
```
```suggestion
    <string name="video_is_age_restricted">Show age restricted video. Future changes are possible from settings.</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -246,7 +246,7 @@
     <string name="detail_dislikes_img_view_description">Dislikes</string>
     <string name="use_tor_title">Use Tor</string>
     <string name="use_tor_summary">(Experimental) Force download traffic through Tor for increased privacy (streaming videos not yet supported).</string>
-    <string name="report_error">Report an Error</string>
+    <string name="report_error">Report Error</string>
```
```suggestion
    <string name="report_error">Report error</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -224,15 +224,15 @@
...
-    <string name="saved_tabs_invalid_json">Using default tabs, error while reading saved tabs</string>
+    <string name="saved_tabs_invalid_json">Could not read saved tabs, so using default ones</string>
     <string name="restore_defaults">Restore defaults</string>
-    <string name="restore_defaults_confirmation">Do you want to restore the defaults?</string>
+    <string name="restore_defaults_confirmation">Do you want to restore defaults?</string>
     <!-- error activity -->
     <string name="sorry_string">Sorry, that should not have happened.</string>
     <string name="guru_meditation" translatable="false">Guru Meditation.</string>
-    <string name="error_report_button_text">Report error via e-mail</string>
+    <string name="error_report_button_text">Report this per e-mail</string>
```
Stypox: I am not sure about this string. What about "Report error by e-mail"?

app/src/main/res/values/strings.xml
TobiGr: Hm. I'd use "from the settings".

app/src/main/res/values/strings.xml
TobiGr: Well, via, by and per are all correct here. Tbh I prefer via (although it is often used formal context).

app/src/main/res/values/strings.xml
Stypox: Mmmh ok

app/src/main/res/values/strings.xml
Stypox: Per is correct, but it is American-ish. So I'd use either by or via

app/src/main/res/values/strings.xml
TobiGr: I prefer @comradekingu's change because I do not think that the "focus gain" info is valuable for the user. What is the difference between "Resume on focus gain" and "Resumed playing" form a user's viewpoint? I think "resumed playing" is a better title for the setting because it tells you what the setting is about: resume a paused video or audio. The info when this is triggered is placed in the description.
But using "Resume on focus gain" can raise several questions when seeing this for the first time: Resume what? What is a focus gain?

app/src/main/res/values/strings.xml
opusforlife2: I suggest "Report this error via e-mail". It is the most descriptive and straightforward.

```suggestion
    <string name="error_report_button_text">Report this error via e-mail</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -585,7 +585,7 @@
     <string name="downloads_storage_ask_summary">You will be asked where to save each download</string>
     <string name="downloads_storage_ask_summary_kitkat">You will be asked where to save each download.\nChoose SAF if you want to download to an external SD card</string>
     <string name="downloads_storage_use_saf_title">Use SAF</string>
-    <string name="downloads_storage_use_saf_summary">The Storage Access Framework allows downloads to an external SD card.\nNote: some devices are not compatible</string>
+    <string name="downloads_storage_use_saf_summary">The \'Storage Access Framework\' allows downloads to an external SD card.\nNote: Some devices are incompatible</string>
```
```suggestion
    <string name="downloads_storage_use_saf_summary">The \'Storage Access Framework\' allows downloads to an external SD card.\nNote: some devices are incompatible</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -557,7 +557,7 @@
     <string name="error_file_creation">The file can not be created</string>
     <string name="error_path_creation">The destination folder can not be created</string>
     <string name="error_permission_denied">Permission denied by the system</string>
-    <string name="error_ssl_exception">Secure connection failed</string>
+    <string name="error_ssl_exception">Could not establish secure connection</string>
```
```suggestion
    <string name="error_ssl_exception">Could not establish a secure connection</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -43,24 +43,24 @@
...
     <string name="play_with_kodi_title">Play with Kodi</string>
-    <string name="kore_not_found">Kore app not found. Install it?</string>
+    <string name="kore_not_found">Install missing Kore app?</string>
     <string name="kore_package" translatable="false">org.xbmc.kore</string>
     <string name="show_play_with_kodi_title">Show \"Play with Kodi\" option</string>
-    <string name="enable_lock_screen_video_thumbnail_title">Enable lock screen video thumbnail</string>
-    <string name="show_play_with_kodi_summary">Display an option to play a video via Kodi media center</string>
-    <string name="enable_lock_screen_video_thumbnail_summary">When using the background player a video thumbnail will be displayed on the lock screen</string>
+    <string name="enable_lock_screen_video_thumbnail_title">Lock screen video thumbnail</string>
+    <string name="show_play_with_kodi_summary">Displayed option to play a video via Kodi media center</string>
```
Stypox: Why `displayed` instead of `display`? I don't think past participle fits here, since the settings says that after being enabled it displays (not displayed) an option [...]
```suggestion
    <string name="show_play_with_kodi_summary">Display an option to play a video via Kodi media center</string>
```

app/src/main/res/values/strings.xml
Stypox: Again for the `-ed`
```suggestion
    <string name="resume_on_audio_focus_gain_title">Resumed playing</string>
```

app/src/main/res/values/strings.xml
Stypox: The other settings in NewPipe do not use `-ed` and other apps (e.g. F-Droid and Aurora Store) use the present tense, too.

TobiGr:
@Stypox In my eyes this is ready to merge.

PR #2944, opened 2020-01-10
Fixed Weblate bugs and other translation related issues

### LINKED ISSUE #3025: "seconds" string is not available for translation

Hey friends, please add ["seconds"](https://github.com/TeamNewPipe/NewPipe/blob/5bf1df9f1403cbd4de9d9cc7d4f8945ac98ef29e/app/src/main/res/values/settings_keys.xml#L33) string to able translate it. [https://imgur.com/r1wAow8](https://imgur.com/r1wAow8)

 #translation #localization

B0pol:
+ Videos in search -> three dots will be translatable in Weblate (it was not translated and not set translatable="false")
+ seek duration (duration for double tap in player) will be translatable in Weblate
+ made default_content_value and default_country_value untranslatable as it should be
+ deleted the strings comments, which was unused and untranslated, to prevent a bug like we have and I just fixed with videos string

Fixes #3025


app/src/main/res/values/strings.xml
```diff
@@ -589,4 +587,11 @@
     <string name="downloads_storage_use_saf_title">Use SAF</string>
     <string name="downloads_storage_use_saf_summary">The Storage Access Framework allows downloads to an external SD card.\nNote: some devices are not compatible</string>
     <string name="choose_instance_prompt">Choose an instance</string>
+    <string name="videos_fixed">Videos</string>
```
TobiGr: Why did you rename this?

app/src/main/res/values/strings.xml
```diff
@@ -589,4 +587,11 @@
...
+    <string name="five_seconds">5 seconds</string>
```
TobiGr: Wouldn't be `%s seconds` better here? We would need to add the seek_duration_description items dynamically, but this shouldn't be a problem.

app/src/main/res/values/strings.xml
B0pol: I tried to explain in the first bullet, I'll retry
This string, which is located here in the app [screenshot](https://i.ibb.co/rtxq5yk/Screenshot-20200128-133137-New-Pipe-PR-2309-1.png), is not translated, and also in weblate, it's not available. So by renaming it, it would be automatically added in Weblate.

If you know how to trigger weblate for a specific string, it's unecessary, but I asked on IRC and TheAssasin answered he have no idea.

app/src/main/res/values/strings.xml
ocean-sicard-baux: I just did it.
The reason why I've chosen this way is: summary is now fine, but when you click on the list, it only shows the value (eg 5, 10, 15…), but not seconds anymore.

B0pol: I firstly done that, but there is a problem: summary is now fine, but when you click on the list, it only shows the value (eg 5, 10, 15…), but not seconds anymore.

 

(I made this with the last commit)

app/src/main/res/values/strings.xml
TobiGr: I think the problem is that another plural string with the same name exists:
https://github.com/TeamNewPipe/NewPipe/blob/8e582aaeddaf8cbed36c920ecff435d24bf5bf39/app/src/main/res/values/strings.xml#L287-L290
`videos_fixed` is a bit misleading. What about `videos_string`?

B0pol: Yeah that's the reason. Also the reason why comments have the same problem. I changed to ``videos_string``.

B0pol:
Closing this pr as commits are screwed up and I don't know how to fix.
Also, as you can see , it's not showing good.
I don't know how to fix this too.

PR #2968, opened 2020-01-18
Select 2k default

### LINKED ISSUE #2941: Ability to select default 2K/4K resolution
It would be good to have 2K and 4K resolution in default Resolution under settings. I know we can have them enabled it and select the best but for devices who have 2k display or people with slower internet where 4K gets a bit overkill, 2K becomes a sweetspot. So please see if this can be done.

harshlele:

Fixes #2941 - added 2K and 4K options to the default resolution settings. also added them to the limit resolution on mobile data settings.

This is my first PR so I have a question. It would be weird to have the option to set default resolution to 2K/4K but the "Allow higher resolutions" option off, which would result in the video not being set to 2K/4K by default. So should I make it so that the 2K/4K options are only visible when the "Allow higher resolutions" option is enabled?


B0pol:
Can you add 1440p & 2160 to popup, and give an apk here ?

harshlele:
popup setting uses the same list as the default video resolution so it's added there too. and the apk, like a debug apk?

B0pol:
Popup also have a selector, in case you want to have a different resolution than the default resolution. And for the apk, yes

harshlele:
i meant that it uses the same XML resource, so the change was also made in the popup settings, like so:
Screenshot_1579360587

The apk(it doesnt support apks so i zipped it): 
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4080873/app-debug.zip)

harshlele:
do i have to make any more changes to this? @B0pol

Stypox:
> So should I make it so that the 2K/4K options are only visible when the "Allow higher resolutions" option is enabled?

I think this is a good idea, otherwise people will get confused. Thanks :-)

harshlele:
@Stypox check this out 

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4250170/app-debug.zip)

wb9688:
What about 1440p60 and 2160p60? Also, could you please rebase on dev? Thanks.

Stypox:
@harshlele sorry, I completely missed this a year ago. Is this still relevant or has something similar already been implemented? Could you rebase? Then I will review

KaKi87:
Hello,
Any news on this ?
Thanks

theoasispc:
Any updates? Thanks

Stypox:
Closing as there has been no activity from the author for a while

Issue #2931, opened 2020-01-08
[Feature Request/Android 10 Bug] Move the video controls at the very bottom of the screen up to prevent accidental touches by fullscreen users


Kizzy1226:
Ever since the Android 10 update, whenever I'm playing a video in the background and have NewPipe open, the video controls such as the time along in the song & etc are at the very bottom of the screen. When I try and swipe up on my home screen (since I have a fullscreen phone), it will always accidently hit the bar and send my current song that I'm listening to into the middle of the song. It's getting to be a bit annoying, and I dunno why this only started with the latest Android 10 update.

Hopefully my description of the issue is accurate and understandable enough.

snappyapple632:
Android 10 officially introduced a new method of navigation that replaces the nav bar via gestures. In this case, the problem is that trying to move the seek bar, which is at the very bottom of the screen, also activates the "back" command (swiping left/right at the bottom of the screen). The first solution that comes to my mind would be to move the seek bar away from the navigation zone, though it doesn't seem like a clean way of doing it.

harshlele:
Also on an Android 10 device here, another problem is that when you try to move the seekbar from left to right, it also activates the Android 10 app switch gesture. 

I think moving the seekbar up is the only way to solve this problem. It's just too close to the "pill"(or whatever its called).

B0pol:
[Here is a video of the problem](https://v.redd.it/wne9fbdri0p41/DASH_1080) (from [reddit](https://www.reddit.com/r/NewPipe/comments/fpauph/request_gestures_interact_with_the_seek_bar_too/))

TobiGr:
Can you please tell us if the problem is solved in this APK?
[app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4623840/app-debug.apk.zip)

PR #3003, opened 2020-01-25
move player controls up on Android 10

harshlele:

Temporary fix for #2931. It moves the player controls up on Android 10 so it doesn't conflict with the gesture navigation on Android 10. 

The SDK version 29  has a special API for moving UI elements away from the navigation area, but the build/compileSdk is 28, so I couldn't use that.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
@@ -33,14 +33,6 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.preference.PreferenceManager;
 import android.provider.Settings;
-import androidx.annotation.ColorInt;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.core.app.ActivityCompat;
-import androidx.appcompat.app.AppCompatActivity;
-import androidx.appcompat.content.res.AppCompatResources;
-import androidx.recyclerview.widget.RecyclerView;
-import androidx.recyclerview.widget.ItemTouchHelper;
 import android.util.DisplayMetrics;
 import android.util.Log;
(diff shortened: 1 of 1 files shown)


avently:
> The SDK version 29 has a special API for moving UI elements away from the navigation area

Why not to switch to API 29 then?

And could you please give a link to that special API? I'll need it in my app soon

harshlele:
also, yeah i think the compileSdk should be moved to 29  @TobiGr

Oizaro:
@harshlele

This fix don't work on my test build. There is any type of changes that I should apply to project to make ot working?

TobiGr:
https://github.com/TeamNewPipe/NewPipe/pull/3337 addressed the same problem, but found a different solution. Can you please tell us if the problem is solved in this APK?
[app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4623840/app-debug.apk.zip)

Oizaro:
@TobiGr

There is a solution also for background player?

TobiGr:
@Oizaro no.

Oizaro:
@TobiGr

If I'm not wrong this PR of harshlele should fix the wrong behaviour of background player:



But not work.

AioiLight:
Galaxy's One UI 2.0 and some custom ROMs implement a true full-screen gesture by setting the height of the navigation bar to zero.
Therefore, even if you calculate from the height of the navigation bar, the gesture cannot be performed correctly on some devices.
The solution is either to move the background player controls up a bit, or to some extent, even on a device with a zero height navigation bar.

My PR also uses the height of the navigation bar, which causes problems on certain devices.

litetex:
No progress since a year, no GitHub actions build and merge conflicts.

opusforlife2:
No need to reopen. Superseded by #6824.

Issue #2922, opened 2020-01-05
Op7/Android 10 This permission is needed to open in popup mode.


famewolf:
- [x ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.
- [ n/a I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

One a oneplus 7 Pro running the latest Android 10 international firmware if set to minimize to the popup player when you hit home a small message appears over the home button saying "This permission is needed to open in popup mode" but it doesn't identify what permission.  Additionally I went into the permissions for Newpipe and the only one available is Storage which has already been allowed.  I assume you need permission to overdraw screens but it doesn't say that nor give me opportunity to enable it.   Since the app did not crash I didn't include that.  Reasonably certain it's either an oxygenos or Android 10 issue but figured I would provide the heads up.

Additionally when I chose the option to play in background it worked after a brief pause in the music however even after the song ended newpipe continued to cause excessive battery drain until Oxygenos flagged it and prompted me to force kill it.

![Screenshot_20200104-232707](https://user-images.githubusercontent.com/4558832/71775225-09b30000-2f4b-11ea-9d73-e2d027fdf32c.jpg)

famewolf:
I did eventually resolve this by finding the permission to display over other apps and enabling newpipe there but in other occasions it's taken me exactly to where I needed to enable it.  I'll leave this open due to that but feel free to close if not of interest.

harshlele:
I think the reason it doesn't send you to the settings is because the user is exiting the app, so forcefully sending them to the settings wouldn't be good.
One way this could be solved is to get the permission from the user right when the setting is enabled.

harshlele:
@Stypox this happens on my device(Nokia 6.1) running Android 10 too, so i dont think its device specific. I tested it on a Android 6 emulator and there when you press the home button, it shows that toast, then takes you to the settings window, but then when you give it permission, it doesnt actually open the popup. I think this happens on any device running android 6 and above. 
I still think the permission should be asked when the user actually sets it to popup when home is pressed.

TobiGr:
Fixed by #3004

PR #3004, opened 2020-01-25
ask for display over other apps permission in settings

harshlele:

#2922 - This issue seems to happen on any device on Android 6 and above. Specifically on Android 10, it seems like it doesn't open the Settings window for the permission when the home button is pressed. But even in cases where it does show the window, it won't open the popup window when you give it the permission. 

So i added a snackbar that lets the user go to the settings when they set the "minimise on home" option to "minimise to popup player".


TobiGr:
Please re-format your code to increase readability.

harshlele:
is this ok?

Issue #1823, opened 2018-10-17
Speed up with touch control


Rockbard:
**Feature request:**
 Speed up with touch control instead of brightness control (as an option)
 **Purpose:**
 Watching  tutorials on youtube will be great if you could quickly speed up video and turn on a normal play speed when needed with convenient touch control
 **Note**
 Volume pitch should be at 100%

Thanks,
Dmitry

bkqtnte10:
What if the speed setting is with swiping left to right (instead up to down)?

Rockbard:
Right to left is filling better as time search function. That's a pretty a
match standard for all players.
 In my opinion brightness with the up and down gesture is arguable, as I
accidentally touching it all the time. But that a standard gesture for most
of the players too.

 Maybe (just an idea) speed is up and down gesture on the left, brightness
in the middle and volume on the right side?

TobiGr:
IMO, three gestures are too much. But @ritiek made it possible in #1644 to disable either the volume or brightness control or even both. So if someone implements a third logic which handles the speed and restricts the maximum number of enabled gestures to two, I'd be okay with this.
@theScrabi @ritiek What do you think?

ritiek:
This should be a good addition. The only caveat I can think of is that three gestures enabled all at once might cause some trouble in triggering the correct behaviour but this shouldn't be a problem as one would be able to disable the gestures they don't need.

> Maybe (just an idea) speed is up and down gesture on the left, brightness in the middle and volume on the right side?

Just my $0.02. IMO as @bkqtnte10 mentioned it would be nicer to have the speed gesture as swiping left-right on any part of the screen. Currently, swiping top-down on the left part of screen affects the brightness and on the right part of screen affects the volume. In the same way, swiping left-right on any part of the screen would affect the playback speed.

In this way, we would be able to have all three gestures (brightness, volume, playback speed) enabled at the same time.

Tikotik:
If someone is looking into this, #1536 is related.
Remember playback speed changes between videos.

androidSwDev099:
How about left/right gesture on top half of the screen to control video play speed/tempo and on left/right gesture on bottom half of the screen for rewind/forward video time?

dardoor:
There's also the double finger gesture, like in MX Player, for example. This way you can easily have three gestures without confusion (in MX Player there's also a zoom gesture, which does get in the way, but you can disable it in the settings).

[...]

Issue #735, opened 2017-09-24
More default settings


Metaception:
Similar to how you can define the default video quality, it would be useful to define default values for other settings.
Particularly, the orientation and playback speed are settings I am interested in.

These two options have dedicated buttons on the playback screen, just like the video quality, so you should still be able to adjust them if the default values do not suffice.

Anyways, Newpipe is a great piece of software!
Also, [newpipe.schabi.org/donate/](https://newpipe.schabi.org/donate/) leads to a 404

wilfriedroset:
Agreed. And this even more needed since recent change when there is no quick access to x2 playback speed. I use to need 3 click to get to x2 (display control, display playback speed, select x2), now I need a lot of them (display control, display playback speed, slider or + button multiple time).

markifi:
Don't close it! Skip silence, timescale speed

theScrabi:
in my opinion having to much settings is not a god design. I would rather want to remember certain configurations rather then overbloading the settings ui.

markifi:
That's fair. Is it possible to do this in a hidden config file edit? All I'm looking for is default speed, skip silence. That's my main reason for using your app, it works rather well.

bill-mcgonigle:
FWIW I'm not currently using the app because I listen to everything at 1.5 and it's too much hassle to do it for every video.  I'm subscribed to this issue so I can know when to switch over (if ever?).

nikolayhg:
+1

syene:
Totally, I would prefer setting my default speed at something greater than 1 as well; and it would be such a timesaver.

I've almost memorised the routine of adjusting the playback speed as I open videos. Those 3-5 seconds definitely add up with how much some of us consume yt now that I think, and it would definitely be an upgrade if the app would let us set a default playback speed.

As for a setting for default orientation, I don't personally find a need for that; the app remembers the last setting and that's works for me.

The app using our last set playback speed for subsequent videos, however, won't work as well: I like the idea of setting something like 1.75x as my default (it would work with the majority of videos) and then adjusting it slightly either ways as the need arises (for the select few where they're going too slow or fast). And I feel this would be the most convenient and least intrusive implementation for all... 

+1 from me.

Korayem:
no need to add options to set default values, just store whatever i chose previously and keep it so
* playback speed
* pitch
* unlink speed and pitch
* skip silence

problem solved

532910:
As I said before there must be an option to save or not 
[...]

PR #3043, opened 2020-02-02
Save and restore whether pitch and tempo are unhooked or not

### LINKED ISSUE #1536: [Feature Request] Make 'Unhook' the default for altered playback speed (or remember it)
I watch most videos on faster than default speed and I have to check 'Unhook' every time, if I don't want a pitch change as well.
Youtube and almost every other video player out there alters the speed without changing the pitch, so it seems like a reasonable thing to do.

I wouldn't even mind if the altered playback speed itself was remembered across videos.

Stypox:

Now the `PlaybackParameterDialog` saves in SharedPreferences the state of the "Unhook" checkbox, so that the user does not have to slide down and click on the unhook checkbox every time. This is one of the proposed solutions in #1536. The other proposed solution was to make the checkbox checked by default, and based on the current pr code it could be implemented just by resetting the saved unhook preference to `true` every time a player is opened, but I think save&restore is better because not every user needs unhook-by-default.
I also fixed the `unhook_checkbox` string, to make it more English-y.

Fixes #1536
Debug apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4144406/app-debug.zip)


gillhash:
Thanks @Stypox for this PR, debug apk working smoothly as intended. I am going to use this for daily basis for days. Remembering speed/pitch-coupling checkbox state & having the unlink checked as default behaviour.

However I might add "If the altered playback speed itself is remembered across videos" (atleast until closing the app) that would be +1.

gillhash:
Possibilities discussed in #735 issue & in #1823 can be touched later as this PR is such a simpler fix for the time being.

I personally feel "If the altered playback speed itself is remembered across videos" (atleast until closing the app) that would be a big +1.
..But that could be done in a different PR.

gillhash:
@TobiGr is this going to be added in 0.18.3 ?

Stypox:
@gillhash No, this will probably be postponed to 0.18.4

PR #3053, opened 2020-02-04
added ability to copy comments on long press

harshlele:

Added ability to copy comments on long pressing them, and entry in the settings to enable/disable it.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
@@ -3,7 +3,6 @@ package org.schabi.newpipe.info_list.holder;
 import android.content.ClipData;
 import android.content.ClipboardManager;
 import android.content.Context;
-import android.preference.PreferenceManager;
 import android.text.util.Linkify;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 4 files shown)


yausername:
I don't think we need a settings option for this feature unless it is causing some performance hit.
We don't want to have a bloated settings menu.

harshlele:
ok i'll remove it

B0pol:
How it works? Is it like description, you can select the area you want, or it copy the whole description?

harshlele:
it copies the whole comment. i actually wanted to make it so that you can select the text, but the root view of the comment is clickable, so the the textview isn't selectable.

app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
```diff
@@ -120,6 +126,21 @@ public void updateFromItem(final InfoItem infoItem, final HistoryRecordManager h
...
         });
+
+
+        itemView.setOnLongClickListener(new View.OnLongClickListener() {
+            @Override
+            public boolean onLongClick(View view) {
+
+                ClipboardManager clipboardManager = (ClipboardManager) itemBuilder.getContext()
+                                                    .getSystemService(Context.CLIPBOARD_SERVICE);
+                clipboardManager.setPrimaryClip(ClipData.newPlainText(null,itemContentView.getText()));
```
yausername: `itemContentView.getText()` can sometimes copy the ellipsized text. Use `commentText` instead.
```suggestion
clipboardManager.setPrimaryClip(ClipData.newPlainText(null,commentText));
```

app/src/main/res/xml/content_settings.xml
```diff
@@ -65,6 +65,7 @@
         android:title="@string/show_comments_title"
         android:summary="@string/show_comments_summary"/>
 
+
```
yausername: remove empty line

PR #3065, opened 2020-02-08
Added the ability to remove all watched videos from a local playlist

GradyClark:
Features:
 - Added the ability to remove all watched videos from a local playlist

Tested on:
 - Pixel

EDIT: Changed Location of 'Remove Watched' button to the three dot menu, see below comments for photos.

Before:
removeWatched_Before

Loading Screen:
removeWatched_Loading

After:
removeWatched_After


TobiGr:
Thanks for the contribution. I guess the idea is to have a dynamic playlist which contains all videos you want to watch later. Is this what you want to achieve? If this is a case, we might want to create a "watch later" playlist which comes with this feature by default.

Otherwise, I'd suggest to move this action to the menu in the toolbar.

GradyClark:
Yeah, a 'watch later' playlist was the idea.

I'll look into making the a 'watch later' playlist.
But a few questions first.

Questions:
 1. What do you think would be better?
     A. A 'watch later' playlist type, like [ local and remote ] playlists.
         So that you could have multiple 'Watch Later' playlists for different occasions.
     B. A single default playlist that everyone gets.
          Like YouTube's watch later playlist.

 2. Do you think the removal of video should be automatic or manual.
     A. Automatic
         Keeps things simple, but people might want to later add the videos to other playlists,
         or to watch them again, and they would have to go through their history to find the videos.
     B. Manual
          Allows the user to come back to it later, 
          but the removal is manual, albeit easier then going video by video.

Stypox:
Here is my opinion:
1. Playlists should all be of the same type (i.e. local type).
2. There should be a button in the dot-menu that removes all the watched videos from any local playlist.

This requires an action by the user, but the automatic alternative has this problem: what if one wanted to watch a video, but then changed his mind after opening it? That video would be removed from the watch-later playlist, which was not intended. So I think we should leave this manual. The other advantage of this approach is that there is no need to implement different types of playlists, or such. Basically you just have to move the button you added to the three-dot menu and we are good to go.

-------

> If this is a case, we might want to create a "watch later" playlist which comes with this feature by default.

@TobiGr For the reasons explained above I think an automatic-removal watch later playlist would be problematic, and the alternative of pressing a button is not that much of a hassle.

GradyClark:
Changed the location of the 'Remove Watched' button, to the menu.<br/>
beforeClick

onClick

Stypox:
@GradyClark could you provide a test apk?

GradyClark:
Here's the zipped apk.

[NewPipe-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4296595/NewPipe-debug.zip)

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -300,6 +316,20 @@ public void onComplete() {}
         };
     }
 
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case R.id.menu_item_removeWatched:
+                //Solution, Scorched Earth, Copy non duplicates, clear playlist, then copy back over
+                //Other options didn't work as intended, or crashed. Like deleteItem(playlist_item) crashes when called in this function.
```
Stypox: `deleteItem(playlist_item)` probably crashes because it has to be called on the main thread (not sure though).

app/src/main/res/layout/local_playlist_header.xml
```diff
@@ -50,7 +50,7 @@
         android:layout_height="wrap_content"
         android:layout_below="@id/playlist_stream_count">
 
-        <include layout="@layout/playlist_control"/>
+        <include layout="@layout/local_playlist_control" />
```
Stypox: Since the local playlist controls have been reset to the old ones, you should revert this change and remove the file `app/src/main/res/layout/local_playlist_control.xml`

app/src/main/res/menu/menu_local_playlist.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools">
+
+    <item
+        android:id="@+id/menu_item_removeWatched"
+        android:title="@string/remove_watched"
+        app:showAsAction="never"/>
```
Stypox: What about adding an icon, and making this line `app:showAsAction="ifRoom"`? The user would then be able to access the option in only one click, if there is enough space. A delete-multiple icon exists in Android Studio's library, which is the same as the Material Design icon library

app/src/main/res/values/strings.xml
```diff
@@ -592,6 +592,7 @@
     <string name="choose_instance_prompt">Choose an instance</string>
     <string name="app_language_title">App language</string>
     <string name="systems_language">System default</string>
+    <string name="remove_watched">Remove Watched</string>
```
```suggestion
    <string name="remove_watched">Remove watched</string>
```

app/src/main/res/menu/menu_local_playlist.xml
GradyClark: I like the idea, but If you accidentally click that button, there is no going back.
With the three dot menu, it's an extra click away, reducing the chance of accidentally removing videos you wanted to keep.
For that reason I would suggest keeping it behind the three dot menu, or at least a 'confirm action' dialog.

Stypox: Oh, there you have a point, then it is ok to keep it behind the three-dot menu

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Converting the AsyncTask to ReactiveX.
Does this seem good?
Any tweaks or should I push the changes?

Added
`private Disposable removeWatchedWorker;`

`onOptionsItemSelected` Looks like

``` 
RemoveWatchedStreams remover = new RemoveWatchedStreams();
// It's the same class, just without extending AsyncTask.

remover.onPreExecute();
removeWatchedWorker = Flowable.just(playlistManager.getPlaylistStreams(playlistId).blockingFirst())
    .subscribeOn(Schedulers.newThread())
    .filter((@NonNull List<PlaylistStreamEntry> playlist) -> {
        remover.doInBackground(playlist);
        return true;
    }
).observeOn(AndroidSchedulers.mainThread())
    .subscribe(playlist -> {
    remover.onPostExecute();
}, (@io.reactivex.annotations.NonNull Throwable throwable) -> {
    onError(throwable);
});
```

It still uses the scorched earth method.
I still haven't figured out how to get the `deleteItem()` to not crash.

Due to the below, I think it's the better method.
With the scorched earth method.
 1. There are less calls to the `itemListAdapter`
 2. if the app crashed, goes wrong, etc, it doesn't effect anything, it only writes once everything is done, and it's quite fast.
 3. less function calls, calling `deleteItem` will call multiple functions that aren't needed until the end.
 For example `updateThumbnailUrl()` doesn't need to be called each time, only at the end.

GradyClark: Switched it over to ReactiveX in commit [fb193ce](https://github.com/TeamNewPipe/NewPipe/pull/3065/commits/fb193ce3622f34b212bff6a645887f98940eb63c)

No problem! 
I really have enjoyed NewPipe, and really thank you, and everyone else who has worked hard to make NewPipe happen!

app/src/main/res/layout/local_playlist_header.xml
GradyClark: I should've caught that. 
Thanks for pointing it out.
I have fixed it in commit [fb193ce](https://github.com/TeamNewPipe/NewPipe/pull/3065/commits/fb193ce3622f34b212bff6a645887f98940eb63c)

app/src/main/res/values/strings.xml
GradyClark: Fixed in commit [fb193ce](https://github.com/TeamNewPipe/NewPipe/pull/3065/commits/fb193ce3622f34b212bff6a645887f98940eb63c)

GradyClark:
Merged it with current 'dev' branch. [8fa29ff](https://github.com/TeamNewPipe/NewPipe/commit/8fa29ffc1985d1a6b31d72dbda3819adee214c4c)
Tested on Pixel, no issues occurred.

APK:
[NewPipe-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4303230/NewPipe-debug.zip)

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -300,6 +321,32 @@ public void onComplete() {}
...
     }
 
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case R.id.menu_item_removeWatched:
+                RemoveWatchedStreams remover = new RemoveWatchedStreams();
+
+                remover.onPreExecute();
+                removeWatchedWorker = Flowable.just(playlistManager.getPlaylistStreams(playlistId).blockingFirst())
```
Stypox: Before this line you should add this code to dispose `removeWatchedWorker` if it is not null:
```java
if (removeWatchedWorker != null) {
    removeWatchedWorker.dispose();
}
```

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -616,5 +663,75 @@ private PlayQueue getPlayQueue(final int index) {
         }
         return new SinglePlayQueue(streamInfoItems, index);
     }
+
+    private class RemoveWatchedStreams{
```
Stypox: I'm not sure this should be done inside a nested class. You could do the same things in three functions without nesting things (at least, that's how I have seen it done in other places in the repository)

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -616,5 +663,75 @@ private PlayQueue getPlayQueue(final int index) {
...
+                    if (history_item.streamId == playlist_item.streamId) {
+                        isNonDuplicate = false;
+                        break;
+                    }
+                }
+                if (isNonDuplicate) {
+                    localItems.add(playlist_item);
+                }
+                else
+                {
```
Stypox: Match the style of the repo: `} else {`
Also below: spaces between `if` and `(`: `if (`

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Fixed in commit [81bcac5](https://github.com/TeamNewPipe/NewPipe/pull/3065/commits/81bcac5d2800b14a4e1917c9ffbfe4064b5fb967)

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Removed the class in commit [https://github.com/TeamNewPipe/NewPipe/pull/3065/commits/81bcac5d2800b14a4e1917c9ffbfe4064b5fb967](https://github.com/TeamNewPipe/NewPipe/pull/3065/commits/81bcac5d2800b14a4e1917c9ffbfe4064b5fb967)

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Fixed in commit 81bcac5

app/src/main/res/values/strings.xml
```diff
@@ -593,6 +593,8 @@
     <string name="choose_instance_prompt">Choose an instance</string>
     <string name="app_language_title">App language</string>
     <string name="systems_language">System default</string>
+    <string name="remove_watched">Remove watched</string>
+    <string name="dynamic_seek_duration_description">%s seconds</string>
```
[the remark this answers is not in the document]
GradyClark: That was a mistake.

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -255,6 +273,9 @@ public void onDestroy() {
         if (debouncedSaveSignal != null) debouncedSaveSignal.onComplete();
         if (disposables != null) disposables.dispose();
 
+        if (removeWatchedWorker != null) removeWatchedWorker.dispose();
+        removeWatchedWorker = null;
+
         debouncedSaveSignal = null;
         playlistManager = null;
         disposables = null;
```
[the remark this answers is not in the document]
GradyClark: 6223c36

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -67,6 +75,8 @@
     private View headerPopupButton;
     private View headerBackgroundButton;
 
+    private Disposable removeWatchedWorker;
```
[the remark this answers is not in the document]
GradyClark: Refactored in 6223c36

GradyClark:
[HistoryRecordManage.loadLocalStreamStateBatch()](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/local/history/HistoryRecordManager.java#L266)
Only returns videos with a, where you left off, marker.
I had to do it manually in 6223c36

Thanks for the advice!
It really speed it up, about 4x faster!

Stypox:
@GradyClark I pushed a commit that really makes the algoritm `O(m*log(n))`: it took ~2s to remove watched videos from a playlist with ~2500 videos and with with a history size of ~2500 videos. I did this by getting the stream ids in ascending order from the database, so that I could apply a binary search on it. Thus, the algorithm uses `log(n)` operations to search for the id every playlisted video inside history.
I would have told you how to do this, but I didn't know either, so I decided to experiment myself and after a while I pushed a commit; hope this is not a problem. :smile:  What do you think about my changes?
Also, sorry for responding so late, there have been 2 versions to review and release lately :sweat_smile:

Stypox:
@TobiGr can I merge this? The work now is polished and fast, as explained in the comment above.

GradyClark:
Sorry for the late reply, I've been busy.

I haven't been able to test the new code, Android Studio has been troublesome. 
It's refusing to compile. saying 
"Unfortunately you can't have non-Gradle Java modules and Android-Gradle modules in one project."

Once I get that sorted out I will test it, and get back with you on that.

As for how the code looks, it looks great!
I especially like the efficiency enhancements you made.

Stypox:
> "Unfortunately you can't have non-Gradle Java modules and Android-Gradle modules in one project."

Try `Build -> Clean Project`, and if even that does not work try `File -> Invalidate caches / restart`

TobiGr:
Two things to note which might be good to be displayed to the user in a dialog:
1. When removing watched videos, videos which have not been completed yet, are also removed. Is this a bug or intended. If it's intended, we should tell the user about this.
2. when a video was watched before it was added to the playlist, it is also removed. In other words, the "remove watched" is global and not restrict to the current playlist.

The dialog should have an option to "do not show again".  @Stypox @GradyClark @wb9688 What do you think?

wb9688:
Here is my comment on this PR's usage of ReactiveX that I sent to @Stypox on IRC:

Screenshot from IRC

GradyClark:
> Try `Build -> Clean Project`

That did the trick
Thanks @Stypox 

I just tested the code, it works without a hitch.

GradyClark:
> 1. When removing watched videos, videos which have not been completed yet, are also removed. Is this a bug or intended. If it's intended, we should tell the user about this.

I don't think there is a way to know for sure that the video watched watched completely.
We can tell at which time they stopped playing the video, and if the video has been played at all.
But we don't have any way of knowing how much of it was watched.
As far as I know.

We could have it ignore any video with a 'where you left off' marker.
But also have the option to remove those as well.

> 2. when a video was watched before it was added to the playlist, it is also removed. In other words, the "remove watched" is global and not restrict to the current playlist.

That was the intended function.
But having a "remove only the videos that were added to the playlist and then watched", I think would be a good additional option.


Perhaps when the "remove watched videos" is pressed a dialog pops up and asks what you want to do

1st dialog
"Do you want to remove video that have been partially watched? Yes / No ?"
2nd dialog
"Do you want to remove only the videos that have been watched after being added to the playlist? Yes / No?"

GradyClark:
> > 1. When removing watched videos, videos which have not been completed yet, are also removed. Is this a bug or intended. If it's intended, we should tell the user about this.
> 
> I don't think there is a way to know for sure that the video watched watched completely.
> We can tell at which time they stopped playing the video, and if the video has been played at all.
> But we don't have any way of knowing how much of it was watched.
> As far as I know.
> 
> We could have it ignore any video with a 'where you left off' marker.
> But also have the option to remove those as well.
> 
> > 1. when a video was watched before it was added to the playlist, it is also removed. In other words, the "remove watched" is global and not restrict to the current playlist.
> 
> That was the intended function.
> But having a "remove only the videos that were added to the playlist and then watched", I think would be a good additional option.
> 
> Perhaps when the "remove watched videos" is pressed a dialog pops up and asks what you want to do
> 
> 1st dialog
> "Do you want to remove video that have been partially watched? Yes / No ?"
> 2nd dialog
> "Do you want to remove only the videos that have been watched after being added to the playlist? Yes / No?"

What are your opinions?
@Stypox, @TobiGr,  @wb9688

wb9688:
Show one "are you sure?" dialog with those 2 things as checkboxes.

Stypox:
But wouldn't the third option require a database restructuring, to keep track of when every video has been added to a playlist?

GradyClark:
That would require changes to the database, perhaps just warning the user will suffice.

GradyClark:
Latest and Greatest
APK: [NewPipe-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4477071/NewPipe-debug.zip)
screen

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -331,6 +358,121 @@ public void onComplete() { }
...
+                            if (indexInHistory < 0) {
+                                notWatchedItems.add(playlistItem);
+                            } else if (!thumbnailVideoRemoved
+                                    && playlistManager.getPlaylistThumbnail(playlistId)
+                                    .equals(playlistItem.getStreamEntity().getThumbnailUrl())) {
+                                thumbnailVideoRemoved = true;
+                            }
+                        }
+                    } else {
+                        boolean hasState = false;
```
Stypox, line 432: Declare variables right where you first use them. Redeclaring a variable every cycle of the while loop does not slow down anything, since declaring is just telling the compiler "I need to have this memory at this point", so memory is really allocated only once.

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -331,6 +358,121 @@ public void onComplete() { }
...
+                        while (playlistIter.hasNext()) {
+                            PlaylistStreamEntry playlistItem = playlistIter.next();
+                            int indexInHistory = Collections.binarySearch(historyStreamIds,
+                                    playlistItem.getStreamId());
+
+                            hasState = streamStatesIter.next() != null;
+                            if (indexInHistory < 0 ||  hasState) {
```
Stypox, line 439: Anyway, I don't think here another variable is needed, you could just go with
```suggestion
                            if (indexInHistory < 0 || streamStatesIter.next() != null) {
```

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -331,6 +358,121 @@ public void onComplete() { }
...
+    }
+
+    public void removeWatchedStreams(final boolean removePartiallyWatched) {
+        if (removeWatchedDisposable != null && !removeWatchedDisposable.isDisposed()) {
+            // already running
+            return;
+        }
+        showLoading();
+
+        removeWatchedDisposable = Flowable.just(Flowable.just(removePartiallyWatched,
```
Stypox, line 391: Do you really need to pass `removePartiallyWatched` to the `Flowable`? I think you can access it just as normal inside the lambda body below

app/src/main/res/values/strings.xml
```diff
@@ -601,6 +601,12 @@
     <string name="choose_instance_prompt">Choose an instance</string>
     <string name="app_language_title">App language</string>
     <string name="systems_language">System default</string>
+    <string name="remove_watched">Remove watched</string>
+    <string name="remove_watched_popup_title">Remove watched videos?</string>
+    <string name="remove_watched_popup_warning">"Videos that have been watched\nbefore and after being added to the playlist will be removed.\nAre you sure? This cannot be undone!</string>
+    <string name="remove_watched_popup_yes">Yes</string>
+    <string name="remove_watched_popup_cancel">Cancel</string>
```
Stypox, line 608: You shouldn't add two new strings: "Yes" and "Cancel" either already exist in `strings.xml` (search with CTRL+F if that's the case) or even are in the default android system strings.

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -331,6 +358,121 @@ public void onComplete() { }
...
+                    boolean localRemovePartiallyWatched = (boolean) flow.blockingFirst();
+                    List<PlaylistStreamEntry> playlist
+                            = (List<PlaylistStreamEntry>) flow.blockingLast();
+                    HistoryRecordManager recordManager = new HistoryRecordManager(getContext());
+                    Iterator<PlaylistStreamEntry> playlistIter = playlist.iterator();
+                    Iterator<StreamHistoryEntry> historyIter = recordManager
+                            .getStreamHistorySortedById().blockingFirst().iterator();
+                    List<PlaylistStreamEntry> notWatchedItems = new ArrayList<>();
+                    Iterator<StreamStateEntity> streamStatesIter = null;
+                    boolean thumbnailVideoRemoved = false;
```
Stypox, line 404: In my opinion it is difficult to read all of these declarations here, maybe they should be distributed right befor the places in code where they are being used (e.g. `streamStatesIter` would be moved on line 405, before the `if (!localRemovePartiallyWatched)` branch where it is used). But I'm not sure if it's the same for others. ;-)

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -331,6 +358,121 @@ public void onComplete() { }
...
+                builder.setMessage(R.string.remove_watched_popup_warning)
+                        .setTitle(R.string.remove_watched_popup_title)
+                        .setPositiveButton(R.string.remove_watched_popup_yes,
+                                (DialogInterface d, int id) -> removeWatchedStreams(false))
+                        .setNeutralButton(
+                                R.string.remove_watched_popup_yes_and_partially_watched_videos,
+                                (DialogInterface d, int id) -> removeWatchedStreams(true))
+                        .setNegativeButton(R.string.remove_watched_popup_cancel,
+                                (DialogInterface d, int id) -> d.cancel());
+                builder.create().show();
```
Stypox, line 376: Usually a "Builder" variable is never saved, since it is needed exactly once. Also, why can't you use `AlertDialog` directly without the full path specification?
```suggestion
                new AlertDialog.Builder(getActivity())
                        .setMessage(R.string.remove_watched_popup_warning)
                        .setTitle(R.string.remove_watched_popup_title)
                        .setPositiveButton(R.string.remove_watched_popup_yes,
                                (DialogInterface d, int id) -> removeWatchedStreams(false))
                        .setNeutralButton(
                                R.string.remove_watched_popup_yes_and_partially_watched_videos,
                                (DialogInterface d, int id) -> removeWatchedStreams(true))
                        .setNegativeButton(R.string.remove_watched_popup_cancel,
                                (DialogInterface d, int id) -> d.cancel());
                        .create()
                        .show();
```

app/src/main/res/values/strings.xml
B0pol: There is a reason we created these strings (yes no cancel ok…) and we aren't using android ones: Android isn't translated to all languages, I can think of Esperanto specifically.

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: streamStatesIter Has to be incremented each time to keep up with the playlistIter
streamStatesIter.next() Doesn't always get called in the "OR" conditional, thus would make them out of sync, (each time) when indexInHistory >= 0

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: I'm really new with ReactiveX, I did that just to make sure that there would be no chance of a weird issue occurring, do you think I should change it?

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: How does this look?
code

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
Stypox: Yes, nothing strange should happen, as `removePartiallyWatched` is declared as final

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
Stypox: Oh, I see, you are right

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
Stypox: Looks better, thanks :-D

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: I tried using that just now, and it causes the removeWatchedDisposable to never dispose.
So you can only use the 'remove watched' ability once, then you have to shut down and reopen the app.

I found a way around it, by force disposing it.

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Fixed in db335d5

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Fixed in db335d5

app/src/main/res/values/strings.xml
GradyClark: Fixed in db335d5

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Fixed in db335d5

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Fixed in db335d5

app/src/main/res/menu/menu_local_playlist.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools">
+
+    <item
+        android:id="@+id/menu_item_removeWatched"
```
wb9688, line 7:
```suggestion
        android:id="@+id/menu_item_remove_watched"
```

app/src/main/res/values/strings.xml
```diff
@@ -601,6 +601,10 @@
     <string name="choose_instance_prompt">Choose an instance</string>
     <string name="app_language_title">App language</string>
     <string name="systems_language">System default</string>
+    <string name="remove_watched">Remove watched</string>
+    <string name="remove_watched_popup_title">Remove watched videos?</string>
+    <string name="remove_watched_popup_warning">"Videos that have been watched\nbefore and after being added to the playlist will be removed.\nAre you sure? This cannot be undone!</string>
```
wb9688, line 606:
```suggestion
    <string name="remove_watched_popup_warning">"Videos that have been watched before and after being added to the playlist will be removed.\nAre you sure? This cannot be undone!</string>
```

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -331,6 +358,127 @@ public void onComplete() { }
...
+                    setVideoCount(videoCount);
+                    if (videoCount == 0) {
+                        showEmptyState();
+                    }
+
+                    hideLoading();
+
+                    //If this is not done, 'removeWatchedDisposable', will never be disposed of.
+                    //Why: Because using the 'removePartiallyWatched' in this functions parms,
+                    // prevents it from disposing. Exact reason for this behavior is unknown
```
wb9688, line 476: They should all have a space after `//`. Also, there probably is a proper way of solving this.

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -244,6 +256,16 @@ public void onPause() {
         saveImmediate();
     }
 
+    @Override
+    public void onCreateOptionsMenu(final Menu menu, final MenuInflater inflater) {
+        if (DEBUG) {
+            Log.d(TAG, "onCreateOptionsMenu() called with: menu = [" + menu
+                    + "], inflater = [" + inflater + "]");
```
wb9688, line 263: I'd personally prefer formatting it something like:

```suggestion
            Log.d(TAG, "onCreateOptionsMenu() called with: "
                    + "menu = [" + menu + "], inflater = [" + inflater + "]");
```

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -331,6 +358,127 @@ public void onComplete() { }
...
+                        .create()
+                        .show();
+                break;
+            default:
+                return super.onOptionsItemSelected(item);
+        }
+        return true;
+    }
+
+    public void removeWatchedStreams(final boolean removePartiallyWatched) {
```
wb9688, line 384: A lot of variables in this function could be `final`

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -331,6 +358,127 @@ public void onComplete() { }
...
+                    Iterator<StreamHistoryEntry> historyIter = recordManager
+                            .getStreamHistorySortedById().blockingFirst().iterator();
+                    Iterator<StreamStateEntity> streamStatesIter = null;
+
+                    //Remove Watched, Functionality data
+                    List<PlaylistStreamEntry> notWatchedItems = new ArrayList<>();
+                    boolean thumbnailVideoRemoved = false;
+
+                    if (!removePartiallyWatched) {
+                        streamStatesIter = recordManager.loadLocalStreamStateBatch(playlist)
```
wb9688, line 408: This could be moved down I think

GradyClark: The space is to separate the variables, and the functions

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: I don't think that is necessary

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: I have looked around, and I'm not sure what the solution would be.

app/src/main/res/menu/menu_local_playlist.xml
GradyClark: Fixed in 776dddd

app/src/main/res/values/strings.xml
GradyClark: Fixed in 776dddd

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Fixed '// ' issue in 776dddd

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Fixed in 776dddd

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
Stypox: It is not necessary, but it is a good idea to do so, @wb9688 is right

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
Stypox: I think he meant to move the
```
streamStatesIter = recordManager.loadLocalStreamStateBatch(playlist)
                                .blockingGet().iterator();
```
block under the `} else {` branch below (where `removePartiallyWatched == false`), i.e. after line 432, so that a check is removed and also less space is used.

Stypox: I think he meant to move the
```
streamStatesIter = recordManager.loadLocalStreamStateBatch(playlist)
                                .blockingGet().iterator();
```
block under the `} else {` branch below (where `removePartiallyWatched == false`), i.e. after line 432, so that a check is removed and also less space is used.

wb9688: Yes, and its initial declaration as well

GradyClark: Ah, I'll fix it in the next commit

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
GradyClark: Fixed in 7361100

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
wb9688: @GradyClark: You should just add this thing to the already existing `CompositeDisposable` in this class (`disposables`) and not do this stuff here at all, see e.g. [here](https://github.com/TeamNewPipe/NewPipe/blob/50046e2a3de419c0583ed6e8c2bc5dd26e13dd4f/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java#L250).

Stypox: @wb9688 but then how can he check if the removal is already going on, and in that case do nothing? See [line 385](https://github.com/TeamNewPipe/NewPipe/pull/3065/files/db335d5cec3b77be37d897279973f1fbf94537af#diff-2ac03147edde19bf93d75ca5d662e7e2R385)

wb9688: @Stypox: Can't you create e.g. a `private boolean isRemovingWatched` for that?

GradyClark:
Your welcome.
NewPipe has been awesome to use, and I wanted to contribute.

Issue #2640, opened 2019-09-16
Cut off m4a downloads


dmoa:
Around the time I switched to F-Droid "version" of NewPipe (Could be coincidence), my audio downloads seem to be cutoff at the beginning.

When playing in google play music some of my favourite songs are cutoff at the beginning.
I also am using NewPipe on the OnePlus 7 Pro (if that helps)

I don't have much information other than this, but I'll be happy to reply to messages.

Any help would be appreciated.

kapodamy:
¿with any video, also in m4a format?

CharAznableLoNZ:
I've found this happens with some videos at random.  It will download about 20-30% then just stop.  Pause and resume doesn't fix it.  Delete video and file then try again it stops at the same point.  I tried downloading the 720 version of this.  In the largest version on F-Droid.
https://www.youtube.com/watch?v=u3c1362zoGg

kapodamy:
confirmed
logcat output:
```
A/libc: heap corruption detected by dlmalloc
    Fatal signal 6 (SIGABRT) at 0x000051ef (code=-6), thread 21912 ( Old Dudes!.mp4)
Process 20975 terminated.
```

kapodamy:
@CharAznableLoNZ 
wait, your issue doent looks similar that OP but is fixed here #2646

CharAznableLoNZ:
I wasn't sure as I didn't want to open a new issue for it if it was the same.

dmoa:
Ok I got it working, super weird.
It doesn't work in Google Play Music, but on other playback clients it does. i.e. no cutoff in other playback clients.

kapodamy:
ok, i will take a look

opusforlife2:
@kapodamy This hasn't been fixed by #2679 from my testing.

Players which don't rely on Android, like VLC, can play these m4a audios just fine. But players which do rely on Android, such as Vanilla and Odyssey, have the first second or two of the audio cut off.

So the file may not be well formed.

kapodamy:
@opusforlife2 
¿youtube url? for testing

opusforlife2:
@kapodamy https://www.youtube.com/watch?v=9IVhjh15ofo
This is a good audio to test as the difference between cut off and full audio is very noticeable.

kapodamy:
@opusforlife2 test this
[...]

Issue #3081, opened 2020-02-12
1080p downloads cannot be opened


gkeegan:
As the title suggests, I can't open/play any 1080p downloads from either SD card or internal storage. I'm using a build of the latest dev commits as of this issue.

This was supposedly fixed by #2959, but somehow isn't?

It was working fine with a build from when 0.18.2 was released, but now doesn't work.

gkeegan:
Example video: https://www.youtube.com/watch?v=6Ljs9_yIiY0

gkeegan:
I contemplated just posting this to #2997, but my issues seem slightly different. For me the video won't open at all or play as audio.

kapodamy:
the `stss` box has less keyframes that a file muxed by ffmpeg (newpipe 77 ffmpeg 96). but plays ok.
¿what player app are using? VLC??? MX Player????

gkeegan:
VLC on Android. It's strange that it works fine for you. I'm on Android 9, but perhaps it is device specific?

kapodamy:
@gkeegan sorry i missed a commit on #2959. thats why works on my device

PR #3095, opened 2020-02-14
m4a cutoff workaround

kapodamy:
fix regression issued by #3081 where the "sbgp" is written in non audio tracks

Also, in order to fix https://github.com/TeamNewPipe/NewPipe/issues/2640, do some """math""" to detect and avoid the cutoff in m4a files. **Work as-is**, i tested many m4a files but no ones have the same problem like https://github.com/TeamNewPipe/NewPipe/issues/2640#issuecomment-575970869

[ ✔ ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/streams/Mp4DashReader.java
+++ b/app/src/main/java/org/schabi/newpipe/streams/Mp4DashReader.java
@@ -195,6 +195,39 @@ public class Mp4DashReader {
         return count;
     }
 
+    /**
+     * find and read the first SIDX related to the selected track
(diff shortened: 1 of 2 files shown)


TobiGr:
Thank you. Is git broken? I cannot see a difference between the first lines O_o Did you change the encoding of the file?

kapodamy:
closing because wrong line ending and github not let me change

PR #3096, opened 2020-02-14
Handle intents for youtube streams in "vnd.youtube:videoId" format

### LINKED ISSUE #3091: java.net.MalformedURLException: unknown protocol: vnd.youtube
If i click a YouTube link or thumbnail in an app like FeedMe or other RSS reader, NewPipe starts but crashes instantly with the error "org.schabi.newpipe.extractor.exceptions.ParsingException: Malformed url: vnd.youtube:thi73589" or some similar error. If I open the same link with standard YouTube app, the right video is opened.

XiangRongLin:

Closes #3091

Tested it by opening "vnd.youtube:Jb8BqYYqDMQ" in the browser and the opening it with NewPipe.

@Offerel Could you check if it works for you, since i don't have those apps you mentioned

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4205823/app-debug.zip)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -632,12 +632,22 @@ public class RouterActivity extends AppCompatActivity {
      */
     private final static String REGEX_REMOVE_FROM_URL = "[\\p{Z}\\p{P}]";
 
+    /**
+     * The official youtube app supports intents in this format, where after the ':' is the videoId.
+     * Accordingly there are other apps sharing streams in this format.
+     */
+    private final static String BASE_YOUTUBE_INTENT_URL = "vnd.youtube:";
+    private final static String BASE_YOUTUBE_VIDEO_URL = "https://www.youtube.com/watch?v=";
+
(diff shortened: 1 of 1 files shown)


TobiGr:
I think this belongs into the extractor. @Stypox What do you think?

XiangRongLin:
I looked there first, but the LinkHandlerFactory already tries to create an URL object from it in the LinkHandlerFactory [here](https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/linkhandler/LinkHandlerFactory.java#L47) before it even calls any methods from the YoutubeStreamLinkHandlerFactory.

And i don't think that something only related to the android front end should go into the extractor, which is supposed to be used for other frontends too. (like the web-api)

Stypox:
This should go into the extractor, since it is something tightly related to the Youtube service, not the app implementation or platform. If one were to build a desktop app, he would need the exact same link handler, even though it is in a different environment

Offerel:
Many thx. Indeed the debug version solves my issue.

Issue #2124, opened 2019-02-15
translate Fastlane app store texts directly on Weblate


eighthave:
- [X] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [X] I checked if the issue/feature exists in the latest version.

Weblate just launched a beta feature for directly translating the Fastlane name/summary/description/whatsnew texts.  Since NewPipe is already on Fastlane and Weblate, it can easily use this. See this in action for F-Droid's Android client:
https://hosted.weblate.org/projects/f-droid/f-droid-metadata/

Follow the beta here: https://github.com/WeblateOrg/weblate/issues/1525

@comradekingu 
#675 
#833

TobiGr:
@theScrabi @TheAssassin  I don't have acces to set this up on WebLate. I guess this is a nice change and will also be required sooner or later. Nevertheless, there will be changes to algo which shows the "latest" app in F-Droid. Apps need to meet certain criteria after a new F-Droid version containing the changes of [this PR](https://gitlab.com/fdroid/fdroidclient/merge_requests/805) is merged.
If I understand the new requiresment correctly, we already fullfill them by providing some German translations. Nevertheless, some translated app descriptions might be nice for F-Droid users.

TheAssassin:
@TobiGr neither do I, @theScrabi will have to do it.

What is Fastlane by the way?

TobiGr:
[Fastlane](https://fastlane.tools/) is an app automation tool. You can do quite interesting things with it. 
IIRC F-Droid tools use it to handle the screenshots, changelogs and descriptions which we store in this repository.

eighthave:
Fastlane is a tool that automates publishing texts and screenshots to
Google Play.  F-Droid also supports the file layout.  So for NewPipe,
you don't need the fastlane tool at all, just use the file layout.

eighthave:
actually, I forgot, fastlane would be useful for you if you want to
generate localized screenshots since it automates that process.

TobiGr:
@theScrabi I know that your busy atm. I'll take care of it when someone grants me the Weblate permissions.

B0pol:
Done

Issue #1588, opened 2018-08-06
Plurals problem in Russian


bboa:
Dear friends, last build have some plurals problems: https://imgur.com/a/m0VryaY

Explanation for screenshot:

str. 2 "6 тыс. подписчика" - expected "many", current "few",
str. 4 "56 тыс. подписчик" - expected "many", current "one".

As I see, the plurals section in strings.xml are correct. How we can help to fix this?

theScrabi:
You could take a look at our weblate.org translation page. Maybe the plurals are wrong there. Also app/src/res/values/strings-ru.xml could help you.

bboa:
hmm, I just check Weblate template and he is also correct, so probably it's a more wide problem, maybe other languages affected too.

I can bypass this by shortening words in translation, but not sure that it's a good idea.

bboa:
Hey! Now I understand.

We have two different view counters.
The one is placed on video details screen, under big thumbnail. It always contains an exact views value, e. g. "194,541 views" and works correctly.

The other is placed on videos list, on right side from little video thumbnails. This counter works dinamically, and here is a problem. If views count is less than 1000, this counter displays exact views count, e.g. "741 views", and all work perfectly too.

But, if views count is more than 999, this counter displays approximate value, e.g. 1K, 2M, 127K etc. And, here is a problem: if we have 194,541 views, in Russian, because last digit is 1, here will be used word form, equal to English "view", not "wiews". But this second counter will be displayed as "194K view", is a English translated eqivalent. Same problem will be reproduced, if exact views count will have last digits as 1, 2, 3, 4, e. g. in 40% cases, for Russian UI.

Looks like for second counter we need to add additional checking to fix non-English UI behaviour, if views count is more than 999 and contains approximated value, then plurals word form (like "views") will be always used, I hope that it's right for all languages.

Issue #1816, opened 2018-10-14
Add more information to crash reports


TobiGr:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.

Sometimes we get error reports, but cannot reproduce crashes due to different versions of the parsed website. To get rid of this difficulty, we should make it possible to add the fetched HTML page to the JSON report. I'd suggest to place an additional checkbox at the top/bottom of the report dialog which is not checked by default. If checked, a new element needs to be added at the end of the JSON. 
@theScrabi @mauriciocolli @karyogamy @TheAssassin I hope my idea is clear. Is there anything I did not take into account which makes this difficult or not recommendable?

I did not look into this yet, but I guess we need to make some changes in the extractor before we can access the fetched document in the frontend. The changes in the crash report converter are quite simple. However, we might want to add this additional information to Sentry, too.

TheAssassin:
That'll be a _lot_ more data than we get now... I'm a bit concerned.

Are you speaking of A/B testing etc.?

theScrabi:
This is hard to implement, since it is not always just html which is part of the extraction process. Also it would require massive changes in the backend. ... Im not sure if the outcome would justify the effort.

Stypox:
I think this would be very useful to be able to fix little nasty bugs that only happen sometimes to some person in some country.
One of the extractor bugs I fixed lately only happened once every ~30 tries, so I had to tell Android Studio to run the same test indefinitely and put a breakpoint in the place the exception was thrown. Only then I was able to catch the page content. Instead if I could access this directly on my phone it would be far simpler, because I would be able to catch random error while using the app normally.

TobiGr:
We needed some more info by some users recently, so I worked on this again. Open a channel page to throw an exception and test the changes: [app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/3692594/app-debug.apk.zip).

<img src="https://user-images.githubusercontent.com/17365767/66245511-982aa200-e70e-11e9-9068-ecfb82b95639.png" width="250">

[...]

PR #3098, opened 2020-02-15
Multiple localization fixes

### LINKED ISSUE #1132: Use the content country of the language set in the Android settings by default

By default, NewPipe sets the content country to United Kingdom, but it'd be better if it looks at Android's language setting (in my case "Nederlands (Nederland)") to set it to the same content country (in my case "Nederland") by default.

### LINKED ISSUE #1558: [Feature Request] add your channels/playlists in the launcher

I would like you to add options in the menus so we can add favorite playlists or channels to the launcher

### LINKED ISSUE #2894: -n  days/hours after update to 18.0
Screenshot_20191226_194109

### LINKED ISSUE #3025: "seconds" string is not available for translation

### LINKED ISSUE #3089: Request: Have video titles and descriptions in the default content language where available
- [X] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

The titles and descriptions for YouTube videos are not always in the language selected as the default content language, even where available. An example is this video: https://www.youtube.com/watch?v=4UDD_yw-LLw. On the YouTube website the title and description will display in English/Japanese depending on the language selected. On NewPipe the title will be in Japanese and the description in English, regardless of whether English or Japanese is selected as the default content language. 

I've also noticed that when Japanese is selected in NewPipe, the titles displayed in the What's New screen and for recommended videos will still be in English when a translated title exists.

I've tried changing the default content country and my system language but this doesn't seem to have had an effect. My issue seems to be similar to the one in https://github.com/TeamNewPipe/NewPipe/issues/1513, but it was closed a while ago and the behaviour persists for me.

Thanks in advance for your help, and apologies if this is a duplicate request.

B0pol:
Hello,
I fixed all the remaining translation related issues on NewPipe, except #2124 (someone with Weblate rigths have to do it).

<details>
<summary>Changelog</summary>
Added a « sytem default » option to content language and country language selector (#1132)
It's the one selected by default (not en-GB anymore then), and correspond to the
language of the system / country of the system
By system I mean phone, tablet, TV…

Fixed russian showing - before time ago (eg 19hrs ago) (#2894)
This is a workaround fix, I opened an issue on prettytime library repo.
https://github.com/ocpsoft/prettytime/issues/182

Fixed russian plurals (#1588)
other was used instead of many for videos and subscribers

Fixed seek_duration english only (#3025)

Fixed views / spectators counts possibly in the wrong language (Bugfix)
Fixed live spectators not showing full count on detail page 
Fixed LIVE shown on players, it shows translated instead
Fixed Videos string in search / three dots not available in Weblate (Bugfix)
(because it was videos, but there already was a plural string named videos, in Weblate)

Subscriber count is always giving the short count (Improvement)
We can't get exact number since this YouTube update: https://support.google.com/youtube/thread/6543166
But only short count (B, M, k), so showing full number, eg for 1.9M: 1,900,000, is wrong because the number could be 1,923,490 or 1,897,789…
</details>
Fixes #3025, fixes #2894, fixes #1558, fixes #1132.


app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
```diff
@@ -21,6 +21,16 @@
     public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        String[] durationsDescriptions = getResources().getStringArray(R.array.seek_duration_description);
```
TobiGr: Please add a comment above that you initialize the seek duration list

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -84,13 +84,19 @@ public static String concatenateStrings(final List<String> strings) {
         final String contentLanguage = PreferenceManager
                 .getDefaultSharedPreferences(context)
                 .getString(context.getString(R.string.content_language_key), context.getString(R.string.default_language_value));
+        if (contentLanguage.equals("system")) {
```
TobiGr: This constant should be in a key

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -84,13 +84,19 @@ public static String concatenateStrings(final List<String> strings) {
...
+            return org.schabi.newpipe.extractor.localization.Localization.fromLocale(Locale.getDefault());
+        }
         return org.schabi.newpipe.extractor.localization.Localization.fromLocalizationCode(contentLanguage);
     }
 
     public static ContentCountry getPreferredContentCountry(final Context context) {
         final String contentCountry = PreferenceManager
                 .getDefaultSharedPreferences(context)
                 .getString(context.getString(R.string.content_country_key), context.getString(R.string.default_country_value));
+        if (contentCountry.equals("system")) {
```
TobiGr: See above

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -226,8 +233,8 @@ private static void changeAppLanguage(Locale loc, Resources res) {
     }
 
     public static Locale getAppLocale(Context context) {
-        SharedPreferences prefs = androidx.preference.PreferenceManager.getDefaultSharedPreferences(context);
-        String lang = prefs.getString("app_language_key", "en");
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        String lang = prefs.getString(context.getString(R.string.app_language_key), "en");
         Locale loc;
         if (lang.equals("system")) {
```
TobiGr: see above

app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
B0pol: done

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -132,14 +137,14 @@ public static String localizeViewCount(Context context, long viewCount) {
         return getQuantity(context, R.plurals.views, R.string.no_views, viewCount, localizeNumber(context, viewCount));
     }
 
-    public static String localizeSubscribersCount(Context context, long subscriberCount) {
```
Stypox: Since this function is exactly the same as `shortSubscriberCount`, wouldn't it be better to rename `shortSubscriberCount`  to `localizeSubscriberCount`? The output of the function is a normal (i.e. with no attributes about size) output, since it can't be considered small when comparing it to the normal output if the normal output does not exist ;-) 

###### Note: removed `s` from subscribers

B0pol: In the first commit
```
Subscriber count is always giving the short count.
We can't get exact number since this YouTube update: https://support.google.com/youtube/thread/6543166
But only short count (B, M, k), so showing full number, eg for 1.9M: 1,900,000, is wrong because the number could be 1,923,490 or 1,897,789…
```
I sticked to this class «convention»
shortSOMETHINGcount is giving the number with the abbreviation at the end (eg 5000000-> 5M in English US)
localizeSOMETHINGcount is formatting the number (eg 5 million is 5,000,000 in English US but 5 000 000 in French).

So no, this is not the same. 5,000,000 != 5M.

For the s, I followed localizeViewCount (not views count)

Then I don't really understand your comment. Can you re-explain please?

Stypox: Oh, my fault, I misunderstood something, this is good 😅

B0pol:
I need to remove the extractor version change first. And this could be merged before the extractor pr, they are not related.

B0pol:
Before merging

app/src/main/res/values/strings.xml
```diff
@@ -594,4 +592,5 @@
     <string name="choose_instance_prompt">Choose an instance</string>
     <string name="app_language_title">App language</string>
     <string name="systems_language">System default</string>
+    <string name="dynamic_seek_duration_description">%s seconds</string>
```
mauriciocolli: This is not a proper value for seconds, it should be using plurals.

To save yet another translation string, maybe we can use a format like "Seconds: %s", or even just the number.

B0pol: It's always plural: 5, 10, 15, 20, 25 seconds are all plurals.

I disagree with only number, the result would be , then:

``Fast forward / rewind seek duration``
- 5
- 10
- 15
…
But 5 what? Seconds, minutes, years? With only number, the user don't know.

And same with Seconds: %s, to me with ``Fast forward / rewind seek duration`` it feels weird, it don't feel like a sentence. If you watch the rest of the menus, ans list preferences, it stays natural, like a sentence.

TobiGr: Well I agree with the mauricio. Some numbers might need special treatment in some languages.
image

mauriciocolli: > It's always plural: 5, 10, 15, 20, 25 seconds are all plurals.

Some languages have different forms depending on the quantity.

> I disagree with only number, the result would be
> But 5 what? Seconds, minutes, years? With only number, the user don't know.

True, the caption of the option would have to include the unit nonetheless.

B0pol: I use plural and included few, many, other as zero, one and two will never be used.

app/src/main/java/org/schabi/newpipe/report/ErrorActivity.java
```diff
@@ -377,8 +378,12 @@ private String getUserActionString(UserAction userAction) {
     }
 
     private String getContentLangString() {
```
B0pol: Three questions:
1. Why is **getContentLangString** giving **content country**?
2. Shouldn't we build a localization and return its toString, with both content Language and content country
3. Should I use Locale.getDefault().toString() or another one from them (thanks [stackoverflow](https://stackoverflow.com/a/23168383))
```
Locale.getDefault().getLanguage()       ---> en      
Locale.getDefault().getISO3Language()   ---> eng 
Locale.getDefault().getCountry()        ---> US 
Locale.getDefault().getISO3Country()    ---> USA 
Locale.getDefault().getDisplayCountry() ---> United States 
Locale.getDefault().getDisplayName()    ---> English (United States) 
Locale.getDefault().toString()          ---> en_US
Locale.getDefault().getDisplayLanguage()---> English
Locale.getDefault().toLanguageTag()     ---> en-US
```

TobiGr: 1. That's a mistake. I already fixed it while working on https://github.com/TeamNewPipe/NewPipe/issues/1816. However, I didn't push the changes. I'll rebase that branch and push it an Friday or Saturday. There was one thing which kept me from creating a PR though: I added extended debugging which allows one to get the documents which could not be parsed and attend them to the crash report. As a result the crash reports could easily be 1BM and more. Copying and pasting them into the a GitHub app or the browser caused serious lagging on my phone. So it might be better to store the documents in a .zip file and ask the user to upload it. I didn't have time to implement this.
2. I added a content country field, too. I guess the best thing is to only print the codes which are stored in the app settings.
3. Should be this. `Locale.getDefault().toString()          ---> en_US`

B0pol: I already put toString so it's ok. Is the array initialization good now?

app/.gitignore
```diff
@@ -1,3 +1,4 @@
 .gitignore
 /build
 app.iml
```
Stypox: This is now irrelevant because there is *.iml
```suggestion
```

app/src/main/res/values/strings.xml
```diff
@@ -594,4 +592,10 @@
...
+    <plurals name="dynamic_seek_duration_description">
+        <item quantity="few">%s seconds</item>
+        <item quantity="many">%s seconds</item>
+        <item quantity="other">%s seconds</item>
+    </plurals>
```
TobiGr: Can you test whether "other" is enough here?
`res.getQuantityString(R.plurals.dynamic_seek_duration_description, currentDurationValue)` should pick the correct one when "few" or "many" is needed by localization.

B0pol: I already tested with esperanto, and yeah other is enough. But for English I left all three, because it's the source string. I don't know if it's necessary, or if weblate will create «many» or «few» for other languages, do you know?

TobiGr: Yes, weblate knows which language has supports the quantities.

B0pol: ok, done

Stypox:
@TobiGr I am not able to merge, I cannot dismiss @mauriciocolli's review even though the changes he requested have been addressed

TobiGr:
@Stypox Done.

PR #3099, opened 2020-02-15
m4a cutoff workaround

kapodamy:
Workaround in order to fix https://github.com/TeamNewPipe/NewPipe/issues/2640, do some """math""" to detect and avoid the cutoff in m4a files. **Work as-is**, i tested many m4a files but no ones have the same problem like https://github.com/TeamNewPipe/NewPipe/issues/2640#issuecomment-575970869

[ ✔ ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/streams/Mp4DashReader.java
+++ b/app/src/main/java/org/schabi/newpipe/streams/Mp4DashReader.java
@@ -195,6 +195,39 @@ public class Mp4DashReader {
         return count;
     }
 
+    /**
+     * find and read the first SIDX related to the selected track
+     * @return the sidx box or null if not found
+     * @throws IOException
+     */
+    public Sidx getSegmentIndexingBox() throws IOException {
+        if (selectedTrack < 0) {
+            throw new IllegalStateException("track no selected");
+        }
+        if (!stream.canRewind()) {
+            throw new IOException("The provided stream doesn't allow seek");
+        }
+
+        Box tmp;
+        Sidx sidx = null;
+
+        stream.rewind();
+
+        while (stream.available())  {
+            tmp = readBox();
+            if (tmp.type == ATOM_SIDX) {
+                sidx = parse_sidx();
+                if (sidx != null) break;
+            }
+            ensure(tmp);
+        }
+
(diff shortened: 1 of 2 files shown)


TobiGr:
Can you please split the regression fix and the new calculation method into two PRs? I'd like to release 0.18.4 asap, but am a bit concerned about 
`        //    following calculation are speculated and based on ffmpeg log's`
I'd like to review that in detail. FYI, you do not need to specualte about what is going on there, you can have a look: https://www.ffmpeg.org/download.html#get-sources

kapodamy:
@TobiGr in #2640 the file has a error in the last samples of the first chunk. I do not know the type of error is.
This problem is detectable using the **Segment Index** box (sidx) embedded in the DASH file. The utility [https://mp4parser.com/](https://mp4parser.com/) shows
sidx_1st_entry
where **438720** is the duration of the first chunk, but the chunk itself has the duration which in this example is **440320**.

This error can be detected if both values are not equal, in this case the **diference** is 1600.

¿what ffmpeg does? nullify these samples with a duration of 1.

The "speculation" is how ffmpeg calculate the "filler" space at the end of the chunk.
after doin various test i concluded with the following formula:
`(Math.ceil(amount_of_no_nullified_samples_on_the_chunk / default_sample_duration) * default_sample_duration) - duration_diference - amount_of_no_nullified_samples_on_the_chunk`

which is implemented as
```java
int filler = ((int)Math.ceil((float)delta / default_sample_duration) * default_sample_duration) - delta - cutoff_chunk;
```
I checked the ffmpeg sourcecode and cannot understand how does. If the "filler" is bad calculated will lead to a little skip of a few miliseconds thats why is a workaround.

kapodamy:
closing becuase this is a device specific bug and fixing the audio stream (like ffmpeg does) :angry: doesnt work.

opusforlife2:
It's alright, man. I've been downloading Opus audios like crazy ever since you fixed it. Out with the old m4a, in with the new opus. :)

PR #3101, opened 2020-02-16
Fixed brightness resetting on first gesture

### LINKED ISSUE #3084: Brightness resets on first gesture brightness change after playback activity open
Thia bug persists since a long time, I didn't come around reporting it.

Stock Android 7.1.1
XT1771

Ullas-Aithal:
- [x]  I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Fixes #3084 

Brightness gesture now works the same ways as volume gesture. The brightness progress bar value is set to previously saved gesture value and increases/decreases value based on relative moment from this value. Previously saved brightness gesture value is calculated by current screen brightness.

Tested different brightness values and verified if the value persists in the next play.
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4209395/app-debug.zip)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
@@ -909,6 +909,9 @@ public final class MainVideoPlayer extends AppCompatActivity
                 final float currentVolumeNormalized = (float) getAudioReactor().getVolume() / getAudioReactor().getMaxVolume();
                 volumeProgressBar.setProgress((int) (volumeProgressBar.getMax() * currentVolumeNormalized));
             }
+            //Set initial value for brightness progress bar based on current screen brightness
+            WindowManager.LayoutParams layoutParams = getWindow().getAttributes();
+            brightnessProgressBar.setProgress((int)(layoutParams.screenBrightness * brightnessProgressBar.getMax()));
         }
 
         @Override


Stypox:
Are you sure that `layoutParams.screenBrightness` is in range `[0.0, 1.0]`? [The reference](https://developer.android.com/reference/android/view/WindowManager.LayoutParams.html#screenBrightness) says:
> A value of less than 0, the default, means to use the preferred screen brightness. 0 to 1 adjusts the brightness from dark to full bright.

Couldn't this go below 0?

Stypox:
Oh, #3102 is another PR that does the same thing, but it also prints some debug info. I'd use that one, what do you think @TobiGr ?

Ullas-Aithal:
> Are you sure that `layoutParams.screenBrightness` is in range `[0.0, 1.0]`? [The reference](https://developer.android.com/reference/android/view/WindowManager.LayoutParams.html#screenBrightness) says:
> 
> > A value of less than 0, the default, means to use the preferred screen brightness. 0 to 1 adjusts the brightness from dark to full bright.
> 
> Couldn't this go below 0?

We do get negative value of -1 if the user has never used brightness gesture. But since we're not using any negative value for ProgressBar using setMin(), the default minimum is 0. So assigning a negative value will default to 0 in our case.

I guess we can set the gesture value to be whatever the system brightness is if the user has never used the gesture?

Stypox:
@Ullas-Aithal see https://github.com/TeamNewPipe/NewPipe/pull/3102#pullrequestreview-359425497

Ullas-Aithal:
@Stypox , oh, its the same change 😄. How does this work? Should I work on my pull request or ?

Stypox:
@Ullas-Aithal #3102 also adds debug information, so I think it is preferable. So you probably shouldn't work on this PR, sorry ;-)
Thank you very much for your contribution, anyway! I will leave the PR open just in case the creator of the other PR does not resolve the submitted reviews.

Ullas-Aithal:
@Stypox that's totally fine. I'll probably take a look at another bug maybe. Thank you for maintaining this app!

Stypox:
@Ullas-Aithal I merged #3102 in the end. Thank you for your contribution, anyway! :-)

PR #3107, opened 2020-02-17
Spelling: Some devices are incompatible

comradekingu:
: smallversal is incompatible. This does away with the potential for errors.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -590,7 +590,7 @@
     <string name="downloads_storage_ask_summary">You will be asked where to save each download</string>
     <string name="downloads_storage_ask_summary_kitkat">You will be asked where to save each download.\nChoose SAF if you want to download to an external SD card</string>
     <string name="downloads_storage_use_saf_title">Use SAF</string>
-    <string name="downloads_storage_use_saf_summary">The \'Storage Access Framework\' allows downloads to an external SD card.\nNote: some devices are incompatible</string>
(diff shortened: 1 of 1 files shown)


Stypox:
What does ": smallversal is incompatible" mean? I am not sure I understand correctly, sorry

comradekingu:
@Stypox The others are : A : B : C etc.

gkeegan:
I'm not sure what this would fix either. Can you explain what the problem is?

comradekingu:
@gkeegan 
Colon bigversal
https://github.com/TeamNewPipe/NewPipe/blob/e14b7851b1b001bfa95635e210d1103335cc82af/app/src/main/res/values/strings.xml#L370
https://github.com/TeamNewPipe/NewPipe/blob/e14b7851b1b001bfa95635e210d1103335cc82af/app/src/main/res/values/strings.xml#L402

Colon smallversal
https://github.com/TeamNewPipe/NewPipe/blob/e14b7851b1b001bfa95635e210d1103335cc82af/app/src/main/res/values/strings.xml#L591

My fix moves everything into the former group, so that it is consistent.
(All the others are ": %1$s")

opusforlife2:
OH! I get it now. Don't make up your own words, man! 🤣 Searching for bigversal gives nothing, and smallversal just links to your profile because of this very issue.

https://en.wiktionary.org/wiki/versal#Noun gives the explanation. One learns new things everyday.

It's the first letter after the colon. That's all. Since the majority are capital letters, the rest should be, too.

Stypox:
Thank you!

Issue #3114, opened 2020-02-19
Release v0.18.4


Stypox:
### New
- Added language selector to change the app language #2921
- Added send to Kodi button to player collapsible menu #2310 
- Added ability to copy comments on long press #3053

### Improved
- Fix ReCaptcha activity and correctly save obtained cookies #3035 
- Remove dot-menu in favour of drawer and hide history button when "Watch history" is not enabled in settings #2960
- Ask for display over other apps permission in settings correctly on Android 6 and later #3004
- Rename local playlist by long-clicking in BookmarkFragment #2954
- Various PeerTube improvements #3044 :
  - Display full description instead of truncated
  - Handle age limit
- Improved several English source strings #2938

### Fixed
- Fixed player starting again although it is paused when option "minimize on app switch" enabled and NewPipe is minimized #2917 
- Fixed `.srt` subtitle downloads containing not all line breaks #2957
- Fixed download to SD card failing because some Android 5 devices are not CTF compliant  #2958
- Fixed downloading on Android KitKat #3032 
- Fixed corrupt video `.mp4` file being recognized as audio file #2959 
- Fixed wrong Chines language codes #2865
- Fix initial brightness value for gesture #3102
- Fixed corrupt mp4 downloads #3117
- [YouTube] Timestamps in description are clickable again #3044

------------------------
## APK for testing: RC1
[app-debug_RC1.zip](https://github.com/TeamNewPipe/NewPipe/files/4227084/app-debug.zip)
**PULL REQUEST #3113**

#### PLEASE REPORT ONLY *NEW* BUGS BELOW

Stypox:
@TobiGr should I also add [NewPipeExtractor changelog](https://github.com/TeamNewPipe/NewPipeExtractor/releases) here?

aaazv:
Do u have a telegram group?

aaazv:
when will this bug of recaptcha get fixed?

garawaa:
Please make it possible to download songs from soundcloud by sharing to
newpipe.

Stypox:
@aaazv no, we don't have a Telegram group, see #2965
Recaptchas have been fixed in this version. Please post here only new issues you find in the provided RC apk.

@garawaa that's not a new bug related to this release and to the provided RC apk, but is just a feature request. There are already issues open for it.

cool-student:
@Stypox 
[...]

PR #2960, opened 2020-01-14
Issue #2254

karkaminski:
Fixes #2254

Removed dot-menu where it is no longer needed and implemented disabling History button according to settings
[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4061629/debug.zip)

- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #3128, opened 2020-02-22
Reintroduce "Settings" button in app bar

Stypox:

As noted by @opusforlife2 in https://github.com/TeamNewPipe/NewPipe/issues/3114#issuecomment-589933537 it is sometimes useful to access settings from activities other than the main one. So this PR readds *some* "Settings" buttons removed in #2960.

In particular, the activities where the button is restored are:
- Queue (background and popup queue activities)
- VideoDetail (see @opusforlife2 usecase)
- Playlist (same as VideoDetail)
- Channel (same as VideoDetail)
- Download (I'm not really sure about this one, because the Download activity is only accessible via the drawer and the user could just press "back", so it would categorize as one of the activities below that do not require the button, but I feel like it would be useful anyway sometimes... **could you give some feedback?**)

Other activities that I think don't need the button, because accessing settings from there is unusual or pressing "back" once is enough:
- MainActivity (just open the drawer)
- Kiosk (press back and the drawer is already open because it was used to open the KioskActivity itself)
- Bookmarks (same as Kiosk)
- What's New (same as Kiosk)
- Subscriptions (same as Kiosk)
- History (same as Kiosk)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -453,15 +453,9 @@ public class MainActivity extends AppCompatActivity {
             NavigationHelper.openMainActivity(this);
         }
 
-        if (sharedPreferences.getBoolean(Constants.KEY_ENABLE_WATCH_HISTORY, true)) {
-            if (DEBUG) Log.d(TAG, "do not show History-menu as its disabled in settings");
-            drawerItems.getMenu().findItem(ITEM_ID_HISTORY).setVisible(true);
-        }
-
-        if (!sharedPreferences.getBoolean(Constants.KEY_ENABLE_WATCH_HISTORY, true)) {
-            if (DEBUG) Log.d(TAG, "show History-menu as its enabled in settings");
-            drawerItems.getMenu().findItem(ITEM_ID_HISTORY).setVisible(false);
-        }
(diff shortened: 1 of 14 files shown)


Stypox:
@opusforlife2 @TobiGr @B0pol what do you think about the Download activity?

opusforlife2:
I was hoping you'd revert this commit 9d773d6e8aa892e929439ba62d2a0985f4e8deb4, actually.

Anyway, I don't think you need the Settings button in the Download activity, as you can't change some Download setting and see its effect there. You'd have to back out of it and download something for that.

gillhash:
@Stypox  As a long time user of NewPipe i urge devs to revert the change which removed dot menu function, Please.

It was so easy to change webm to Mp4 &  vice-versa while loading video.

opusforlife2:
@gillhash This PR fixes exactly that.

Stypox:
@opusforlife2 @gillhash could you test this apk and see if it works correctly?
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4239166/app-debug.zip)

opusforlife2:
Tested all bullets. Everything is fine.

> press back and the drawer is already open because it was used to open the KioskActivity itself

This doesn't check out though. The drawer isn't already open for any of these activities. You have to open it again. I don't see why that matters, anyway.

PR #3133, opened 2020-02-23
Change app id based on current git branch

Stypox:

As stated in https://github.com/TeamNewPipe/NewPipe/issues/3114#issuecomment-589998621 I changed the gradle script so that it adds the current branch name to the application id and to the app name.

Since the application id cannot contain special characters, I remove all non-letter characters from the branch name when appending it to the app id. When the current branch is "master" or "dev", everything is just like before (i.e. the app name is "NewPipe Debug" and the app id is "org.schabi.newpipe.debug"). When building a release apk everything is like before, too (i.e. the app name is "NewPipe" and the app id is "org.schabi.newpipe").
I removed the debug `AndroidManifest.xml` because its only purpose was that of setting the app name to "Android Debug" in debug builds, but this is now done by the gradle script.

Thanks to this change we would be able to **install multiple builds from different branches at once**.

For example take this PR: the branch name is `gradle-app-id-suffix`. Therefore the app name is "NewPipe gradle-app-id-suffix" and the app id is "org.schabi.newpipe.debug.gradleappidsuffix".


B0pol:
> Since the application id cannot contain special characters, I remove all non-letter characters from the branch name when appending it to the app id.

Why not do like #2309 , NewPipe PR#NUMBER ?

wb9688:
@B0pol: Because you can't know what PR number it will be in Gradle.

Stypox:
@B0pol There is no way to get the pr number: it would require making requests to GitHub to check if there is an open pull request from that branch and if that's the case get the number. There are a few issues: I'm not even sure if this can be done; connecting to the internet just to build an app is not a good idea, imo; what if the user wants to build the app before having created a pr?
Yeah, we could end up with conflicts in package names, but they would be rare and this solution would be better that always having conflicts anyway

app/build.gradle
```diff
@@ -28,7 +29,15 @@ android {
         debug {
             multiDexEnabled true
             debuggable true
-            applicationIdSuffix ".debug"
+
+            def workingBranch = "git rev-parse --abbrev-ref HEAD".execute().text.trim()
+            if (workingBranch.isEmpty() || workingBranch == "master" || workingBranch == "dev") {
+                applicationIdSuffix ".debug"
+                resValue "string", "app_name", "NewPipe Debug"
```
B0pol: Uh, won't it add ``.debug`` and ``" Debug" to master? ````|| workingBranch == "master"``

Have you tried to apply your changes in the master branch and launch the app, to see what happens?

B0pol:
> @B0pol There is no way to get the pr number: it would require making requests to GitHub to check if there is an open pull request from that branch and if that's the case get the number. There are a few issues: I'm not even sure if this can be done; connecting to the internet just to build an app is not a good idea, imo; what if the user wants to build the app before having created a pr?
> Yeah, we could end up with conflicts in package names, but they would be rare and this solution would be better that always having conflicts anyway

You're right, I didn't think about you need to upload first to get a pr number.

app/build.gradle
Stypox: The branch name is independent from the build type. So we can have a release build on any branch and a debug build on the master branch. Gradle has no idea whether the current branch is a development branch or not, so it can't do decisions based on that.
So it is ok for a debug build on master branch to have .debug at the end

connectety:
Someone may not name their branches and just code on their dev branch.
What about adding the username to differentiate between branches with the same name?
To make sure the name is not to long, maybe just do it if the branch is `dev` or `master` and the user is not `TeamNewpipe`.

**NOT YET TESTED:**
You can get just the username with:
```gradle
def remoteName = "git remote".execute().text.trim()
def remoteURL = ("git ls-remote --get-url" + remoteName).execute().text.trim()

def pattern = ~/^([^@\n]+@|https:\/\/).*[:\/)](.+)\/.+\.git$/
def matcher = pattern.matcher(remoteURL)
def username = matcher.group(2)
```
You can get the remote username and branch with (this runs one less command and may be faster):
```gradle
def remoteName = "git remote".execute().text.trim()
def remoteURL = ("git ls-remote --get-url" + remoteName).execute().text.trim()

def pattern = ~/^([^@\n]+@|https:\/\/).*[:\/)](.+)\/(.+)\.git$/
def matcher = pattern.matcher(remoteURL)
def username = matcher.group(2)
def branch = matcher.group(3)
```

wb9688:
@connectety: That's IMHO a bad idea and could break more easily, e.g. that `.git` at the end is not required, SSH instead of HTTPS cloning is also possible, etc

connectety:
@wb9688 the regex should not break, i tested with:
```
https://github.com/TeamNewPipe/NewPipe.git
git@github.com:TeamNewPipe/NewPipe.git
https://connectety:EXAMPLETOKEN23123@github.com/TeamNewPipe/NewPipe.git
git@gitlab.com:connectety/NewPipeStub.git
https://gitlab.com/connectety/NewPipeStub.git
https://oauth2:ExampleToken@gitlab.com/connectety/NewPipeStub.git
```

wb9688:
@connectety: What about https://github.com/TeamNewPipe/NewPipe or ssh://git@github.com:22/TeamNewPipe/NewPipe.git or http://localhost:12345/NewPipe.git or file:///some/path/to/NewPipe.git?

connectety:
@wb9688 This is just a PoC to show the idea of adding the username could be implemented. Its not about fixing all the details.
 
If the Author thinks, this is a good idea, I can find the time and the current solution is not used. I might code something using the URI and URL classes in groovy (maybe separate PR).

That  being said, I don't think this should cover all edge-cases, however unlikely they are. The implementation in this PR doesn't cover a good amount of edge cases like:
- A branch name which only consists of Chinese characters
- A branch name which is an issue number
- Git beeing installed, but not in the path
- downloading the repo via the git download button and not having set up git on their PC.

For now this is just a PoC to get an opinion from the Author of the PR and one of the Collaborators @Stypox.

Stypox:
> I don't think this should cover all edge-cases

I agree, the purpose of this pr is to have different apk names most of the times, but it is not something we can't live without. So I think a simple implementation that works most of the times is enough, in case of errors it just falls back to the old `.debug` suffix. (I'm going to add more checks though, as this could fail with strange branch names)

wb9688:
Imho it should just use the branch name like @Stypox already implemented.

Stypox:
I fixed the issues that could have caused a build failure:
- Add function `getGitWorkingBranch` that returns the current working branch, and "" if it could not be determined (either because git is not installed or because the directory is not a git repo).
- Make sure normalizedWorkingBranch is not empty (leading to an invalid app id terminating with `.`)
- Make normalizedWorkingBranch lowercase

The code that adds the suffix should now be safe: in case of problems it defaults to the old `.debug` and `NewPipe Debug` values. So this PR is now ready.

avently:
@Stypox this is cool PR. So useful for everyone

B0pol:
I feel like there is something important missing. Try to share a YouTube link, you'll see they all are newpipe, _except NewPipe PR#2309_ if you've installed it.
Could you please take a look or ask @mauriciocolli and change that?

B0pol:
I found another problem. For me, android studio is launching NewPipe Debug version when I click on run, instead of NewPipe _branchname_. It still install the NewPipe _branchname_ app, but doesn't launch it.

mauriciocolli:
@B0pol 
> Try to share a YouTube link, you'll see they all are newpipe, except NewPipe PR#2309 if you've installed it.

Fixed in #3225.

> For me, android studio is launching NewPipe Debug version when I click on run, instead of NewPipe branchname. It still install the NewPipe branchname app, but doesn't launch it.

I noticed that changing branches with Android Studio (doesn't matter if you change externally) almost always break something, whether is the "launch default activity" (your case) or even the code that is builds from.

Only solution is, with luck, forcing a sync using the options in the "File" menu or, when that doesn't work, restarting the whole IDE.

Surely an area that it could improve on.

opusforlife2:
@Stypox Could you also automatically rename the debug apk that is generated?

TobiGr:
According to [the master of all questions](https://stackoverflow.com/questions/16787493/android-gradle-build-renaming-the-apk), using `project.archivesBaseName` should work.

opusforlife2:
... This might not have been such a good idea. I now have _eight_ of them installed. 😅

B0pol:
There is a button called _uninstall_, it's really useful /s

opusforlife2:
One guy to install them all,
debug apks to try them,
one guy to look for bugs,
and in his testing find them.

TobiGr:
> ... This might not have been such a good idea. I now have eight of them installed. :sweat_smile:

@opusforlife2 An easy way to remove those at once (This feature was added by some lazy devs at Google IIRC):
Go to settings > Backup & reset > Factory reset :rofl:

opusforlife2:
Instructions unclear. Phone exploded. Eyebrows gone. 🤷‍♂

TacoTheDank:
@opusforlife2 Geez, I didn't know the Note 7 was THAT sensitive 😳

Issue #668, opened 2017-09-02
Chromecast support


lacolos:
First off I wanted to say good work on New pipe, it is now my go to YouTube player! The only functionality I miss is the ability to chromecast. I'm not sure if this is in the works or even possible? Thanks

coffeemakr:
Support was already requested (see #351), then closed and not reopened after the discovery of an open-source library to stream to Google Cast.

theScrabi:
I wonder how far they come with this project.

imShara:
Not chromecast, but reversed youtube remote API https://github.com/mutantmonkey/youtube-remote

KeizerDev:
@imShara That's cool, we can implement it but the thing is. The Chromecast audio won't work with it and it would be awesome to get that working with NewPipe. If we can make that work it would be another plus to use this app rather then the original Youtube app as they don't support the Chromecast audio. I want to dig into implementing it from scratch but I don't have much time. I've found a good project who have implemented it, without the Google lib.

KeizerDev:
https://github.com/vitalidze/chromecast-java-api-v2

theScrabi:
That lib actually looks pretty sophisticated. I like it :)

KeizerDev:
Yup, and we have direct links, right? It wouldn't be much of a pain to implement it

theScrabi:
I guess not.

wb9688:
@theScrabi: All we'll have to care about is are [the supported formats of Chromecast](https://developers.google.com/cast/docs/media).

theScrabi:
> Android's layout stuff

Totally understandable xD
> So maybe we should make a PipeCast library

Yea why not :)

wb9688:
@theScrabi: Yeah, but someone will need to do that layout stuff. Btw could
you create a PipeCast repo in TeamNewPipe and give me write access to it?

theScrabi:
I created [PipeCast](https://github.com/TeamNewPipe/PipeCast), and you got write access.
Please make sure the license is gpl3.

wb9688:
@theScrabi: Thanks! Ofcourse I will, that's why I asked if Apache v2 can
become GPL v3 in a derivative work, which I looked up and it looks like
it's possible. I'll try starting to implement it tomorrow.

theScrabi:
Well I'm not that as good as @TheAssassin with legal things, but I think since apche2 is a mandatory license you can change the license when ever you like. You should just keep the copyright.

[...]

Issue #105, opened 2015-11-25
UPnP / DLNA streaming on network receiver


sbaconnais:
Chromecast is a closed source device with closed API.

But UPnP / DLNA are open protocols and many smart TVs implement it natively. It would be great to implement UPnP / DLNA stream to network compatible devices.

You can get source code from YAACC (GPL v3) here http://sourceforge.net/p/yaacc/code/ci/master/tree/

medavox:
Consider it added to the queue.
If you feel like you can implement this yourself, feel free to have a go.

sbaconnais:
Unfortunately i'm not an Android developer ... But i'm patient ;)

medavox:
Fair enough :) I'll add it to the TODO list on the wiki.

ildar:
I suggest the feature should be implemented on the Yaacc side: https://sourceforge.net/p/yaacc/tickets/117/

sbaconnais:
Are you a Yaacc developer  ?

ildar:
Nope

sbaconnais:
Yaacc doesn't seem to be really active. And it's a bit outdated ...

I'm not even sure there are API that could be uses by NewPipe.

Perhaps including source code would be a better solution.

pizzadude:
I know this is an old issue, but just in case anyone's interested, you can share a video from NewPipe to an application like BubbleUPNP or Allcast and it will work with the chromecast. But that's only if you're okay with installing proprietary apps..

(Those apps work with UPnP devices also)

theScrabi:
I think if you use cromecast at the first place you also don't care about using non free software. However I'd like to support upnp but I don't know how, and to be honest, I have other problems to care about right first :/

sbaconnais:
Here is an open source library for Upnp/Dlna : https://github.com/4thline/cling

theScrabi:
All right. I can't do it right now, but if someone wants to implement this for NewPipe I'll add it.

KeizerDev:
If I do a PR with Chromecast support, using https://github.com/googlecast/CastCompanionLibrary-android and the use of the default media controller would this be OK? I would like to support the Chromecast Audio for youtube because the official does not do it. 

If it's not OK i'll start working on my own client with similar support but I would rather like to do a collaboration.

theScrabi:
Depends, is this library free software, or did we have to import binary blobs?

KeizerDev:
[...]

PR #3146, opened 2020-02-26
Chroemcast

au190:
Hi All

Thank you very much for the GREAT app !

Does it support the chroemcast  or it will support latter ?

Best regards!


gkeegan:
May I ask why this is a PR?

TobiGr:
See #668 and #105

PR #3154, opened 2020-02-27
Round at one place for Localization.shortCount()

### LINKED ISSUE #3152: Rounding of number of subscribers
For channels with over a million subscribers, the number of subscribers is rounded down to an integer since v0.18.5, previous versions showed at least tenths of a million.

B0pol:

Fixes #3152


app/src/main/java/org/schabi/newpipe/util/Utils.java
```diff
@@ -0,0 +1,11 @@
+package org.schabi.newpipe.util;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+
+public class Utils {
```
wb9688: Do we really need a separate class just for one function that rounds a number? I'd personally put it as a private function in Localization.. Otherwise this PR looks good to me. @TobiGr: What do you think?

TobiGr: I agree

B0pol: Well, the point was it's about Maths, not Localization. I didn't want to create a MathUtils just for one function, so I created a general Utils file. But ok, I done that change.

PR #3155, opened 2020-02-28
Fixed Bad Static Application Implementation

AmniX:
I have Removed some Bad Implementation of Application which was written in static way.
According to best practices, Context aware components should not be static else they can cause memory leaks

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/App.java
+++ b/app/src/main/java/org/schabi/newpipe/App.java
@@ -67,7 +67,6 @@ import io.reactivex.plugins.RxJavaPlugins;
 public class App extends Application {
     protected static final String TAG = App.class.toString();
     private RefWatcher refWatcher;
-    private static App app;
 
     @SuppressWarnings("unchecked")
     private static final Class<? extends ReportSenderFactory>[]
@@ -91,7 +90,6 @@ public class App extends Application {
         }
         refWatcher = installLeakCanary();
 
-        app = this;
 
         // Initialize settings first because others inits can use its values
         SettingsActivity.initSettings(this);
@@ -112,7 +110,7 @@ public class App extends Application {
         configureRxJavaErrorHandler();
 
         // Check for new version
-        new CheckForNewAppVersionTask().execute();
+        new CheckForNewAppVersionTask(this).execute();
     }
 
     protected Downloader getDownloader() {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/schabi/newpipe/util/FireTvUtils.java
```diff
@@ -1,10 +1,12 @@
 package org.schabi.newpipe.util;
 
+import android.app.Application;
+
 import org.schabi.newpipe.App;
```
Stypox:
```suggestion
```

app/src/main/java/org/schabi/newpipe/App.java
```diff
@@ -91,7 +90,6 @@ public void onCreate() {
         }
         refWatcher = installLeakCanary();
```
Stypox:
```suggestion
```

app/src/main/java/org/schabi/newpipe/util/FireTvUtils.java
```diff
@@ -1,10 +1,12 @@
...
 
 public class FireTvUtils {
```
Stypox:
```suggestion
public class FireTvUtils {
```

avently:
> According to best practices, Context aware components should not be static else they can cause memory leaks

It's an app context saved into the same app class. It can't make a memory leak because the reference will be cleared on app exit. The best practices you are talking about is used for activities, fragments, etc. These components shouldn't be saved in a static or dynamic ways as you mentioned.

Stypox:
So what should we do with this pr? @avently @undeadfox @TobiGr

AmniX:
> > According to best practices, Context aware components should not be static else they can cause memory leaks
> 
> It's an app context saved into the same app class. It can't make a memory leak because the reference will be cleared on app exit. The best practices you are talking about is used for activities, fragments, etc. These components shouldn't be saved in a static or dynamic ways as you mentioned.

But It's a Good Practice not to keep any static variable of Context Aware Components.
And Application Parent Class is ContextWrapper and Context Wrapper Parent Class is Context.

So Best Practice makes the Application Health At Best State.
And On Other hand, If you want to talk about Application Context Memory leak, Pass this Application context to a asyncTask and run it for 5 minutes, Now Exit from your app and Now You will observe that you app will stay active in background and won't release the memory until your Thread is terminated. 
Cannot Explain More :)

Issue #3100, opened 2020-02-16
Feature request: Display Changelog in app


Poolitzer:
I would love to have that implemented. Maybe even with links to the fitting blog posts.

Even better would be to be able to see what the new version brings me from the notification, and then I decide if I want to download it. If we want to link to the blogpost there, we would be required to publish it before the new version, but that would also be nice.

B0pol:
Related #2124.

Poolitzer:
Not sure I see how they are related?

TobiGr:
I create a first prototype for viewing changelogs yesterday :D
However, the changes are stored in a file which is included in the app so viewing the changelog before updating is not possible. I guess only a few people translate these files during the release phase.

I guess @B0pol wants to note that we can get the changelogs directly from the fastlane directory and thus offer translations of them, too. Is this correct?

TobiGr:
I also considered this. We can either link the post before releasing a new version or even better. get that url from our api

B0pol:
> > If we want to link to the blogpost there, we would be required to publish it before the new version, but that would also be nice.
> 
> I also considered this. We can either link the post before releasing a new version or even better. get that url from our api

> I guess @B0pol wants to note that we can get the changelogs directly from the fastlane directory and thus offer translations of them, too. Is this correct?

Yeah, that's what I meant by related, include translation of changelogs. And indeed if you use API, we can gather data after update, therefore get some time for translations (with F-Droid delay + RC phase it should be good).

Poolitzer:
I am not sure I can follow 100%. But it sounds like doubling the work, fastelane and our API. Cant we use only one?

opusforlife2:
- I think the blog post should be linked in the update notification.
- The release notes should be bullet points, nothing else.
- There can be a link to the blog post in the release notes.
- There should be an option to turn off showing release notes automatically.
- The release notes should be viewable at any time by tapping an option in settings.

> I just created a first prototype for viewing changelogs yesterday :D

Any progress on that, @TobiGr?

Stypox:
Related: #859 and #1316 (about app tours and FAQs)

tEctutoRIt:
![8dd9aa51-5f89-43d3-a958-ff163ea84701.jpg](https://user-images.githubusercontent.com/77356742/122400179-1c7f9000-cf99-11eb-8229-fdfd5132c327.jpg)

I too think it be nice have a direct link to the change log in this about page in app.

Issue #2254, opened 2019-03-30
Do not show "History" in context menu if "History & Cache" disabled in "Settings"


Symbian9:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x] I checked if the issue/feature exists in the latest version.
- [ ] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

![pic.1](https://i.imgur.com/3nPj4jO.png)

![pic.2](https://i.imgur.com/lDBK798.png)

TobiGr:
I think the whole menu can be remove as its content is available through the navigation drawer. @theScrabi Is there a reason why it has not been remove yet?

karkaminski:
Hey, I would like to make this one (as my first-ever OS issue). Is it still available?

Stypox:
Yeah, this is still available @karkaminski 
What needs to be done is removing the duplicate three-dot menu and disable the History button in the navigation drawer when the user asked not to store history.
Feel free to ask any question, and thank you :-D

karkaminski:
(I think) I completed the task but need support with git-stuff ;)
This is what I've done so far:
1. Forked and cloned NewPipe repo
2. Created new branch (name: issue-#2254)
3. Completed the task: removed dot-menu where it is no longer needed and implemented disabling History button according to settings
4. commited (2 commits) and pushed to github
What should I do next...? :)

Stypox:
Go [to your branch on GitHub](https://github.com/karkaminski/NewPipe/tree/issue-%232254) and click on the "Pull request" button just above the file browser

mauriciocolli:
If this is to be kept, maybe we should think about making the navigation drawer available application-wide as well?

Very annoying that you have to quit everything for changing a single setting.

PR #3156, opened 2020-02-28
Making drawer accesable for easy navigation 

### LINKED ISSUE #2358: NewPipe doesn't remember the video viewed  path when back button is clicked

### LINKED ISSUE #3135: Make drawer accessible from every activity
Probably related to #3128, could you add other options like downloads to (now-gone) app bar?
IMO this can make navigation bit faster.

Or a potentially better way, adding an option to the settings to return UI back to pre-0.18.3 design?

Hashik-Donthineni:

Fixes #3135
Closes #2358

Additions: 
1. Now Navigation is accessible from both MainFragment, VideoDetailFragment.
2. Added the "Home" tab to the drawer to reach home (MainFragment) easily.

I tested this rigorously and there are no bugs I can find. And, I find the app very easy to navigate now :)
This is the Debug-Apk for testing: [Debug-Apk](https://drive.google.com/file/d/1ii2awCyGs7hh8fZvvkeze_4kAQFE0w8u/view?usp=sharing) (Changed the package name for side by side install) 

ZIP Version of APK: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4292429/app-debug.zip)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -144,11 +144,15 @@ public class MainActivity extends AppCompatActivity {
         drawer = findViewById(R.id.drawer_layout);
         drawerItems = findViewById(R.id.navigation);
 
-        //Home button
-        drawerItems.getMenu().add(R.id.menu_tabs_group, ITEM_ID_HOME, ORDER, R.string.tab_home)
-                .setIcon(ThemeHelper.resolveResourceIdFromAttr(this, R.attr.ic_kiosk_local));
+
 
         //Tabs
+
+        //Home button
+        drawerItems.getMenu()
+                .add(R.id.menu_tabs_group, ITEM_ID_HOME, ORDER, R.string.tab_home)
+                .setIcon(ThemeHelper.resolveResourceIdFromAttr(this, R.attr.ic_kiosk_local));
+
         int currentServiceId = ServiceHelper.getSelectedServiceId(this);
         StreamingService service = NewPipe.getService(currentServiceId);
 
@@ -160,6 +164,7 @@ public class MainActivity extends AppCompatActivity {
                     .setIcon(KioskTranslator.getKioskIcons(ks, this));
             kioskId ++;
         }
+
 
         drawerItems.getMenu()
                 .add(R.id.menu_tabs_group, ITEM_ID_SUBSCRIPTIONS, ORDER, R.string.tab_subscriptions)
@@ -382,6 +387,12 @@ public class MainActivity extends AppCompatActivity {
         serviceArrow.setImageResource(R.drawable.ic_arrow_down_white);
 
         //Tabs
+
+        //Home button
+        drawerItems.getMenu()
+                .add(R.id.menu_tabs_group, ITEM_ID_HOME, ORDER, R.string.tab_home)
+                .setIcon(ThemeHelper.resolveResourceIdFromAttr(this, R.attr.ic_kiosk_local));
+
         int currentServiceId = ServiceHelper.getSelectedServiceId(this);
         StreamingService service = NewPipe.getService(currentServiceId);
 
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -145,14 +145,11 @@ public class MainActivity extends AppCompatActivity {
         drawerItems = findViewById(R.id.navigation);
 
         //Tabs
-        Fragment fragment = getSupportFragmentManager().findFragmentById(R.id.fragment_holder);
-
         //Home button
-        if (!(fragment instanceof MainFragment)) {//We don't need to show the Home button when on home
-            drawerItems.getMenu()
-                    .add(R.id.menu_tabs_group, ITEM_ID_HOME, ORDER, R.string.tab_home)
-                    .setIcon(ThemeHelper.resolveResourceIdFromAttr(this, R.attr.ic_kiosk_local));
-        }
+        drawerItems.getMenu()
+                .add(R.id.menu_tabs_group, ITEM_ID_HOME, ORDER, R.string.tab_home)
+                .setIcon(ThemeHelper.resolveResourceIdFromAttr(this, R.attr.ic_kiosk_local));
+
 
         int currentServiceId = ServiceHelper.getSelectedServiceId(this);
         StreamingService service = NewPipe.getService(currentServiceId);
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -478,6 +478,13 @@ public class MainActivity extends AppCompatActivity {
         if (DEBUG) Log.d(TAG, "onBackPressed() called");
 
         Fragment fragment = getSupportFragmentManager().findFragmentById(R.id.fragment_holder);
+
+        //If Drawer is open close it when back is pressed (PlayStore style)
+        if (drawer.isDrawerOpen(GravityCompat.START)){
+            drawer.closeDrawer(GravityCompat.START);
+            return;
+        }
+
         // If current fragment implements BackPressable (i.e. can/wanna handle back press) delegate the back press to it
         if (fragment instanceof BackPressable) {
             if (((BackPressable) fragment).onBackPressed()) return;
@@ -607,6 +614,8 @@ public class MainActivity extends AppCompatActivity {
         final DrawerLayout drawer = findViewById(R.id.drawer_layout);
 
         final Fragment fragment = getSupportFragmentManager().findFragmentById(R.id.fragment_holder);
+
+        //If it's MainFragment enable the Navigation Drawer again
         if (fragment instanceof MainFragment) {
             getSupportActionBar().setDisplayHomeAsUpEnabled(false);
             if (toggle != null) {
@@ -614,10 +623,18 @@ public class MainActivity extends AppCompatActivity {
                 toolbar.setNavigationOnClickListener(v -> drawer.openDrawer(GravityCompat.START));
                 drawer.setDrawerLockMode(DrawerLayout.LOCK_MODE_UNDEFINED);
             }
-        } else {
-            drawer.setDrawerLockMode(DrawerLayout.LOCK_MODE_LOCKED_CLOSED);
-            getSupportActionBar().setDisplayHomeAsUpEnabled(true);
-            toolbar.setNavigationOnClickListener(v -> onHomeButtonPressed());
+            return;
+        }
+
+        //We need to enable the home button for the search fragment to go back to MainFragment
+        if (fragment instanceof SearchFragment) {
+            getSupportActionBar().setDisplayHomeAsUpEnabled(false);
+            if (toggle != null) {
+                drawer.setDrawerLockMode(DrawerLayout.LOCK_MODE_LOCKED_CLOSED);
+                getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+                toolbar.setNavigationOnClickListener(v -> onHomeButtonPressed());
+            }
+            return;
         }
     }
 
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -641,9 +641,9 @@ public class MainActivity extends AppCompatActivity {
                 toolbar.setNavigationOnClickListener(v -> onHomeButtonPressed());
                 return;
             }
-        }else{
-            getSupportActionBar().setDisplayHomeAsUpEnabled(false);
+        } else {
             if (toggle != null) {
+                getSupportActionBar().setDisplayHomeAsUpEnabled(false);
                 toggle.syncState();
                 toolbar.setNavigationOnClickListener(v -> drawer.openDrawer(GravityCompat.START));
                 drawer.setDrawerLockMode(DrawerLayout.LOCK_MODE_UNDEFINED);
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -145,7 +145,8 @@ public class MainActivity extends AppCompatActivity {
         drawerItems = findViewById(R.id.navigation);
 
         //Home button
-        drawerItems.getMenu().add(R.id.menu_tabs_group, ITEM_ID_HOME, ORDER, R.string.tab_home);
+        drawerItems.getMenu().add(R.id.menu_tabs_group, ITEM_ID_HOME, ORDER, R.string.tab_home)
+                .setIcon(ThemeHelper.resolveResourceIdFromAttr(this, R.attr.ic_kiosk_local));
 
         //Tabs
         int currentServiceId = ServiceHelper.getSelectedServiceId(this);
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -92,11 +92,13 @@ public class MainActivity extends AppCompatActivity {
     private boolean servicesShown = false;
     private ImageView serviceArrow;
 
-    private static final int ITEM_ID_SUBSCRIPTIONS = - 1;
-    private static final int ITEM_ID_FEED = - 2;
-    private static final int ITEM_ID_BOOKMARKS = - 3;
-    private static final int ITEM_ID_DOWNLOADS = - 4;
-    private static final int ITEM_ID_HISTORY = - 5;
+
+    private static final int ITEM_ID_HOME = - 1;
+    private static final int ITEM_ID_SUBSCRIPTIONS = - 2;
+    private static final int ITEM_ID_FEED = - 3;
+    private static final int ITEM_ID_BOOKMARKS = - 4;
+    private static final int ITEM_ID_DOWNLOADS = - 5;
+    private static final int ITEM_ID_HISTORY = - 6;
     private static final int ITEM_ID_SETTINGS = 0;
     private static final int ITEM_ID_ABOUT = 1;
 
@@ -141,6 +143,9 @@ public class MainActivity extends AppCompatActivity {
         final Toolbar toolbar = findViewById(R.id.toolbar);
         drawer = findViewById(R.id.drawer_layout);
         drawerItems = findViewById(R.id.navigation);
+
+        //Home button
+        drawerItems.getMenu().add(R.id.menu_tabs_group, ITEM_ID_HOME, ORDER, R.string.tab_home);
 
         //Tabs
         int currentServiceId = ServiceHelper.getSelectedServiceId(this);
@@ -236,6 +241,9 @@ public class MainActivity extends AppCompatActivity {
 
     private void tabSelected(MenuItem item) throws ExtractionException {
         switch(item.getItemId()) {
+            case  ITEM_ID_HOME:
+                NavigationHelper.openMainFragment(getSupportFragmentManager());
+                break;
             case ITEM_ID_SUBSCRIPTIONS:
                 NavigationHelper.openSubscriptionFragment(getSupportFragmentManager());
                 break;
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -92,15 +92,15 @@ public class MainActivity extends AppCompatActivity {
     private boolean servicesShown = false;
     private ImageView serviceArrow;
 
-
+    private static final int ITEM_ID_SETTINGS = 0;
+    private static final int ITEM_ID_ABOUT = 1;
     private static final int ITEM_ID_HOME = 2;
     private static final int ITEM_ID_SUBSCRIPTIONS = 3;
     private static final int ITEM_ID_FEED = 4;
     private static final int ITEM_ID_BOOKMARKS = 5;
     private static final int ITEM_ID_DOWNLOADS = 6;
     private static final int ITEM_ID_HISTORY = 7;
-    private static final int ITEM_ID_SETTINGS = 0;
-    private static final int ITEM_ID_ABOUT = 1;
+
 
     private static final int ORDER = 0;
 
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -93,12 +93,12 @@ public class MainActivity extends AppCompatActivity {
     private ImageView serviceArrow;
 
 
-    private static final int ITEM_ID_HOME = - 1;
-    private static final int ITEM_ID_SUBSCRIPTIONS = - 2;
-    private static final int ITEM_ID_FEED = - 3;
-    private static final int ITEM_ID_BOOKMARKS = - 4;
-    private static final int ITEM_ID_DOWNLOADS = - 5;
-    private static final int ITEM_ID_HISTORY = - 6;
+    private static final int ITEM_ID_HOME = 2;
+    private static final int ITEM_ID_SUBSCRIPTIONS = 3;
+    private static final int ITEM_ID_FEED = 4;
+    private static final int ITEM_ID_BOOKMARKS = 5;
+    private static final int ITEM_ID_DOWNLOADS = 6;
+    private static final int ITEM_ID_HISTORY = 7;
     private static final int ITEM_ID_SETTINGS = 0;
     private static final int ITEM_ID_ABOUT = 1;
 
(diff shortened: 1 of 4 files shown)


TobiGr:
Thank you, please keep the `dev` branch. We should look into this soon

wb9688:
You could use icons from https://material.io/resources/icons/, which are iirc included in Android Studio.

opusforlife2:
Debug apk please! (If you could rename the package to org.schabi.newpipe.drawernav or something it would be swell!)

opusforlife2:
Oh no.

I just realised avently removed the top navigation bar from the VideoDetailFragment, so he had to enable the side drawer anyway. This PR would be made redundant when that one gets merged. :(

Until then, the settings button restored by @Stypox could work as a substitute.

Hashik-Donthineni:
@opusforlife2 Can you please explain mate? I mean I already implemented the drawer here anyway, right? Why did he remove the navigation and where can we access the drawer if the navigation is gone?

wb9688:
@opusforlife2: You mean his unified player PR? If so, that probably won't be merged for a while, and then this PR would still be useful in the meantime.

Hashik-Donthineni:
I added the Debug-apk, please check.

opusforlife2:
@wb9688 True. 

@Hashik-Donthineni as he said, the unified player PR implements the side drawer. But it would be great to have this functionality now instead of possibly a month later (or even more). :)

Gonna test the apk now. Thanks!

opusforlife2:
Tested. Looks and works great! The hamburger icon makes it clear that the side drawer is available, and the back arrow's functionality is shifted to Home, so no loss there. You correctly discard the backstack when tapping Home too, just like the erstwhile back arrow. No complaints with this PR!

Hashik-Donthineni:
@opusforlife2 I uploaded to GDrive because it's easier to check for upload version in Gdrive and we can update the same link with the newer version if we want and I am commuting sorry that I didn't change the package name I'll do it when I get the time and update the same link :)

Hashik-Donthineni:
@wb9688 Added the home icon as well, used ic_kiosk_local :)

opusforlife2:
Bug found! When you tap on the service selector drop down then dismiss it, the Home button disappears. This happens on both Fragments.

Hashik-Donthineni:
@opusforlife2 Fixed the issue :)

opusforlife2:
Confirmed. :)

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -236,6 +247,9 @@ private  void changeService(MenuItem item) {
 
     private void tabSelected(MenuItem item) throws ExtractionException {
         switch(item.getItemId()) {
+            case  ITEM_ID_HOME:
```
Stypox:
```suggestion
            case ITEM_ID_HOME:
```

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -373,6 +387,12 @@ private void showTabs() throws ExtractionException {
         serviceArrow.setImageResource(R.drawable.ic_arrow_down_white);
 
         //Tabs
+
+        //Home button
+        drawerItems.getMenu()
```
Stypox: Shouldn't the menu button be hidden in the MainActivity? There is no need for a "home" button if we are already at home ;-)

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -607,17 +634,27 @@ private void updateDrawerNavigation() {
...
-            toolbar.setNavigationOnClickListener(v -> onHomeButtonPressed());
+            return;
+        }
+
+        //We need to enable the home button for the search fragment to go back to MainFragment
+        if (fragment instanceof SearchFragment) {
+            getSupportActionBar().setDisplayHomeAsUpEnabled(false);
+            if (toggle != null) {
+                drawer.setDrawerLockMode(DrawerLayout.LOCK_MODE_LOCKED_CLOSED);
+                getSupportActionBar().setDisplayHomeAsUpEnabled(true);
```
Stypox: Question: why are you first setting `HomeAsUpEnabled` to `false` and then to `true`?

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -92,11 +92,13 @@
...
-    private static final int ITEM_ID_BOOKMARKS = - 3;
-    private static final int ITEM_ID_DOWNLOADS = - 4;
-    private static final int ITEM_ID_HISTORY = - 5;
+
+    private static final int ITEM_ID_HOME = - 1;
+    private static final int ITEM_ID_SUBSCRIPTIONS = - 2;
+    private static final int ITEM_ID_FEED = - 3;
+    private static final int ITEM_ID_BOOKMARKS = - 4;
+    private static final int ITEM_ID_DOWNLOADS = - 5;
+    private static final int ITEM_ID_HISTORY = - 6;
```
Stypox: Why were these values like that?
```suggestion
    private static final int ITEM_ID_HOME = -1;
    private static final int ITEM_ID_SUBSCRIPTIONS = -2;
    private static final int ITEM_ID_FEED = -3;
    private static final int ITEM_ID_BOOKMARKS = -4;
    private static final int ITEM_ID_DOWNLOADS = -5;
    private static final int ITEM_ID_HISTORY = -6;
```

app/src/main/java/org/schabi/newpipe/MainActivity.java
Hashik-Donthineni: Why tho? I mean most apps have it I guess, If you want me to take it off I'll do it :)

Stypox: Oh ok, if it's normal to leave it in, then feel free to ignore my request ;-)

app/src/main/java/org/schabi/newpipe/MainActivity.java
Hashik-Donthineni: Are you talking about the negative values? That I ain't sure about. It's already there so I let it stay.

Stypox: Yeah, I know ;-) I'd just make them normal anyway

Hashik-Donthineni: Done

app/src/main/java/org/schabi/newpipe/MainActivity.java
Hashik-Donthineni: Changed it :) my bad

Hashik-Donthineni:
@Stypox In my opinion, it's better to make the Home button available on the MainFragment aswell because the user will get accustomed to using and seeing that button there and it'll get into muscle memory if we keep moving it, I think it'll contribute to bad UX, unknowingly.

app/src/main/java/org/schabi/newpipe/MainActivity.java
wb9688: I agree that the home button should also be visible on the home page.

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -92,11 +92,13 @@
...
-    private static final int ITEM_ID_HISTORY = - 5;
+
+    private static final int ITEM_ID_HOME = 2;
+    private static final int ITEM_ID_SUBSCRIPTIONS = 3;
+    private static final int ITEM_ID_FEED = 4;
+    private static final int ITEM_ID_BOOKMARKS = 5;
+    private static final int ITEM_ID_DOWNLOADS = 6;
+    private static final int ITEM_ID_HISTORY = 7;
     private static final int ITEM_ID_SETTINGS = 0;
     private static final int ITEM_ID_ABOUT = 1;
```
Stypox: While you are at it, let's also reorder things a little bit ;-)
```suggestion
    private static final int ITEM_ID_HOME = 0;
    private static final int ITEM_ID_SUBSCRIPTIONS = 1;
    private static final int ITEM_ID_FEED = 2;
    private static final int ITEM_ID_BOOKMARKS = 3;
    private static final int ITEM_ID_DOWNLOADS = 4;
    private static final int ITEM_ID_HISTORY = 5;
    private static final int ITEM_ID_SETTINGS = 6;
    private static final int ITEM_ID_ABOUT = 7;
```

Hashik-Donthineni: Done!

opusforlife2:
> I just realised avently removed the top navigation bar from the VideoDetailFragment, so he had to enable the side drawer anyway. This PR would be made redundant when that one gets merged. :(

Update on this. I was mistaken. He didn't add it. That was just me swiping up the paused video from the home screen. The drawer naturally works there. But it doesn't work if you open the video details screen and try there.

So, this PR is much needed, anyway. 👍

Stypox:
@Hashik-Donthineni could you please provide another APK?

Stypox:
@Hashik-Donthineni could you please provide another APK and fix conflicts?

Hashik-Donthineni:
@Stypox Done! updated the same link. BTW why is the build failing?

Hashik-Donthineni:
@Stypox Merge conflict resolved.

Stypox:
I found an issue: when opening a video from the search fragment the drawer is missing

Hashik-Donthineni:
> Code looks good, thank you!
> Could you provide another APK? After a last row of testing I think this is ready to go :-D
> Btw, you can upload apks on GitHub by zipping them

Do you want the package name to be reset, so that it'll rewrite on the existing app?

Stypox:
For now there is no need that you do it manually, this will be automatic once my PR is merged, but I don't mind if you don't change id

Stypox:
> I found an issue: when opening a video from the search fragment the drawer is missing

This also happens with playlists and channels, when opened from search fragment

Hashik-Donthineni:
Will check that and update.

Hashik-Donthineni:
@Stypox Please check now :) updated the apk on the same drive link.

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -633,23 +633,20 @@ private void updateDrawerNavigation() {
...
-
         //We need to enable the home button for the search fragment to go back to MainFragment
         if (fragment instanceof SearchFragment) {
             if (toggle != null) {
                 drawer.setDrawerLockMode(DrawerLayout.LOCK_MODE_LOCKED_CLOSED);
                 getSupportActionBar().setDisplayHomeAsUpEnabled(true);
                 toolbar.setNavigationOnClickListener(v -> onHomeButtonPressed());
+                return;
+            }
+        }else{
```
Stypox:
```suggestion
        } else {
```

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -633,23 +633,20 @@ private void updateDrawerNavigation() {
...
+            getSupportActionBar().setDisplayHomeAsUpEnabled(false);
```
Stypox: Why is this outside of the if here, but inside in the part above?

surrodox2001:
> 
> 
> Almost ready, the apk works well :-D

So that means we're gonna see 0.18.7 coming in the near future?

Stypox:
> So that means we're gonna see 0.18.7 coming in the near future?
No, version releases are unrelated. I don't even know if we are going to release 0.18.7 or 0.19.0 directly

app/src/main/java/org/schabi/newpipe/MainActivity.java
Hashik-Donthineni: Which part @Stypox you meant before the bug araised?

surrodox2001:
> 
> 
> > So that means we're gonna see 0.18.7 coming in the near future?
> 
> No, version releases are unrelated. I don't even know if we are going to release 0.18.7 or 0.19.0 directly @surrodox2001

From https://newpipe.schabi.org/blog/release/pinned/newpipe-0.18.6-released/:
>Another release???
>Yes, another release. The developers told me they are deeply sorry. And now it is on me to try to >explain why you have to update (and I have to write a blogpost…) again.

I've never seen the subversion goes up so high right now. Maybe a roll to new version?

app/src/main/java/org/schabi/newpipe/MainActivity.java
Hashik-Donthineni: Before the bug, I assumed that to go to VideoDetailsFragment the MainFragment is the only way, but when search fragment is called, the video details fragment is gone. This time, I activated the back button only to search, enabled the menu to the rest of the fragments. But settings and About will not be affected.

Hashik-Donthineni:
Happy to contribute :)

Stypox:
@TobiGr may I merge this? I tested many activity paths and the drawer with its new button seem to work reliably everywhere

Hashik-Donthineni:
> the drawer button is only visible in the parent screen.

Yeah, very true but can you tell me the distinction between the back and up button in play store when you are deep into apps how will you go home? why do we need two buttons doing the same thing?

> Check out the "Upward navigation" section from the material design guidelines.

It still preserved in the back button, the back button is almost exactly doing what the up button is doing, as it's mentioned in that link....

> It can move users chronologically through their recent screen history, or upwards through an app’s hierarchy
 
the back button still serves its purpose.

> parent screen and the action is not considered a navigation from the user.

What is "Parent screen" can you please explain? which action?

> it should be kept as it is
The drawer available from every screen is very useful, I thought about adding it as well.

Which one is it mate? you like it or you don't like it? I don't understand xD

mauriciocolli:
> Yeah, very true but can you tell me the distinction between the back and up button in play store when you are deep into apps how will you go home? why do we need two buttons doing the same thing?

They are not doing the same thing, test in your end these steps to see it:

- Open an app page.
- Navigate to multiple related apps.
- Click the up button.

You should be home, because that's what that button does, navigate a level up. It's subtle, but there _is_ a difference.

> That's because they have the bottom navigation. We have much handy side navigation bar, they have different scenario than ours, it's not very suitable to compare.

It's very suitable because they also have the same navigation view and it's enabled at all times, _even when you are deep navigating through items_, that's why I talked about that app.

###### Of course, doing the same thing that Play Store does isn't good just because Google does it, but because I think it makes sense (and you can find other other apps doing that now).

> It still preserved in the back button, the back button is almost exactly doing what the up button is doing, as it's mentioned in that link....

It's not the same, with the back button you'd have to keep pressing it multiple times. The up button is a _level_ upwards navigation.

Here's some relevant parts:
>> Upward navigation allows users to navigate one level upwards within a single app’s hierarchy, until the app’s home or top-level screen is reached. For example, the Up arrow in a top app bar is a form of upwards reverse navigation.

>> Clearly indicate child screens’ relationship with screens above them in the hierarchy. For example, if a user moves directly to a child screen in your app, they should be able to identify the parent screen to which they can navigate upward.

This was discussed before, the solution that we ended up with was:

https://github.com/TeamNewPipe/NewPipe/blob/d5c29bf1b540e8e9243651b6d43baa177884c29a/app/src/main/java/org/schabi/newpipe/MainActivity.java#L512-L542

> What is "Parent screen" can you please explain? which action?

Take in consideration that I mentioned the case where keeping the drawer icon would make sense.

Another way to call it would be "Home screen", the one that is visible when you open the app from a closed state. The action in this case would be the user pressing a button in said drawer.

Usually, it is used in apps that have a flat navigation, that is, there's no child screens, similar to how the bottom navigation bar does it (though in the case of the bottom navigation you'd still have to indicate to the user that they aren't in the "home screen" by showing the up button when they navigate away from home, again, how it's done in the Play Store).

> Which one is it mate? you like it or you don't like it? I don't understand xD

>> The drawer available from every screen is very useful, I thought about adding it as well.
>>
>>>>>>> ### About the up button:
>>
>> IMO, it should be kept as it is, some reasons why:

xDDDDDDDDDDDDDDDDDDDDDD

----

Even if we disagree about the up button behavior, what the icon represents it's what is important. Just keep it.

**TL;DR**: Keep the drawer active in all screens, keep the up icon the way it is now.

opusforlife2:
@Hashik-Donthineni Where art thou?

Hashik-Donthineni:
@opusforlife2 sorry what?

opusforlife2:
Have you addressed the requested changes? It can't be merged without that.

Hashik-Donthineni:
@opusforlife2 there are no changes @mauriciocolli is asking to close the whole PR, he didn't like the idea it seems.

opusforlife2:
> TL;DR: Keep the drawer active in all screens, keep the up icon the way it is now.

How? 🤔

surrodox2001:
> 
> 
> > TL;DR: Keep the drawer active in all screens, keep the up icon the way it is now.
> 
> How? thinking

I never saw any up button in Newpipe.

surrodox2001:
Also, I think the triple dot bar should be included in the main screen and download & history should be included in the video window as navigation could be faster. (personal opinion)

opusforlife2:
> I never saw any up button in Newpipe.

@surrodox2001 Read this part from mauricio's [review](https://github.com/TeamNewPipe/NewPipe/pull/3156#pullrequestreview-370067620):

>  Up button doesn't reflect the user navigation "diagram" any more.

and
> Killed the up navigation handling. 
The behavior is slightly different from a normal back button event.
Check out the "Upward navigation" section from the material design guidelines.

Open the link to the material design guidelines and read about Reverse chronological navigation and Upward navigation.

surrodox2001:
> 
> 
> > I never saw any up button in Newpipe.
> 
> @surrodox2001 Read this part from mauricio's [review](https://github.com/TeamNewPipe/NewPipe/pull/3156#pullrequestreview-370067620):
> 
> > Up button doesn't reflect the user navigation "diagram" any more.
> 
> and
> 
> > Killed the up navigation handling.
> > The behavior is slightly different from a normal back button event.
> > Check out the "Upward navigation" section from the material design guidelines.
> 
> Open the link to the material design guidelines and read about Reverse chronological navigation and Upward navigation.
> 
> Edit: specifically this guideline:
> 
> > Upward navigation should be implemented for all child screens in an app and follow platform guidance.

Don't know all about this, can a UX designer enlighten me?

surrodox2001:
And why someone wants to make changes to the essence of UI that's accustomed by many users in 0.18.4?

opusforlife2:
> And why someone wants to make changes to the essence of UI that's accustomed by many users in 0.18.4?

Because this PR gives you a better option?

Did you open that page? It has photos explaining what they mean. You don't need anyone to explain anything.

surrodox2001:
I saw #2254 before and it originally requests to track the history on/off with option in 3dot menu and then escalates to remove the entire thing altogether to dedupe. Why we need that though? As we've accustomed to the interface early on. Everyones opinion will be different, so YMMV.

surrodox2001:
@Hashik-Donthineni I seen your work and it's nice. Maybe you can add d/l and history to 3dot menu in the video ui?

Hashik-Donthineni:
@surrodox2001 I would, but I'm not sure about the stance of this PR. I worked on this PR reading the issue, and it took me time to understand and write proper code that adheres to the app codebase. @mauriciocolli didn't respond on the issue, but after making the PR, he's asking to close it.

opusforlife2:
@Hashik-Donthineni I still don't get it. Can you point out where he says that?

Hashik-Donthineni:
@opusforlife2 Did I get the wrong impression then? Here is what that made me believe so,

>  IMO it should be kept as it is, some reasons why:

> TL;DR: Keep the drawer active in all screens, keep the up icon the way it is now.
How??? The only way is to kill this PR and redirect the up button to the "Home." I don't see any other way.

opusforlife2:
Are you saying that the back button in the top bar _has_ to be changed to be able to use the home button in the drawer?

Hashik-Donthineni:
@opusforlife2 sorry what? I am saying we can't use drawer icon and up icon at the same time? not the the home button inside the drawer.

opusforlife2:
Oh. That's it? Just use the Up icon instead of the drawer icon. The drawer can be used by swiping from the left anyway. The button is redundant.

opusforlife2:
@TobiGr @Stypox In #2907, the video detail fragment currently doesn't have a top bar.

This means that if you are on a screen where the drawer cannot be accessed (I can think of the search fragment, channel fragment and playlist fragment right now), you can't access Settings without navigating a bit. This is not much of a problem for the playlist and channel fragments, as they have the three dot menu still (so you can swipe down the video and access it), but the search fragment changes that to a filter.

I think the drawer should definitely be available in the search fragment, but it could be made available in the other two as well.

Hashik-Donthineni:
@opusforlife2 So, finally you want me to remove the drawer button and preserve the functionality of up button (and restore the up button of course) and make drawer accessible with a swipe.. that right?

opusforlife2:
It is already accessible with a swipe. But in a nutshell, yes. I think that is what Mauricio was asking for as well.

surrodox2001:
> 
> 
> @opusforlife2 So, finally you want me to remove the drawer button and preserve the functionality of up button (and restore the up button of course) and make drawer accessible with a swipe.. that right?

Is that @mauriciocolli wanted you to do? I think that will be unintuitive as it's not apparent for the users.

opusforlife2:
> I think that will be unintuitive as it's not apparent for the users. (user dont know you can swipe to reveal side menu)

Agreed. F-Droid users can see the changelog when updating, but Github users can't. But that is for #3100 to solve, not this PR.

Hashik-Donthineni:
>  I think that will be unintuitive
Absolutely. 

@opusforlife2 This is what I am really confused about.

surrodox2001:
Can @Stypox @TobiGr comment on the @mauriciocolli objection to this PR?

opusforlife2:
@Hashik-Donthineni It's not your problem to solve, bro. Don't worry about it. ;)

Stypox:
There are some apps where the drawer is hidden but can be accessed from anywhere (e.g. OctoDroid). So this might not be completely unintuitive. Also, it is really likely that at some point the user swipes left by accident, thus finding the hidden drawer. But [Material guidelines](https://material.io/components/navigation-drawer/#modal-drawer) say that "a modal drawer is always opened by a navigation menu icon". 

------

*off-topic*
> F-Droid users can see the changelog when updating

I never saw a changelog for NewPipe in F-Droid, I thought it was badly configured, but do you see them?

opusforlife2:
> But Material guidelines say that "a modal drawer is always opened by a navigation menu icon".

That's okay. Other apps usually don't have the need for them on every screen. We do. (Plus Play Store violates this guideline anyway. 🙄 )

> I never saw a changelog for NewPipe in F-Droid, I thought it was badly configured, but do you see them?

It has been doing so for quite a few versions. It's just that the links aren't clickable. For that, the user can just further open the changelog link.

Hashik-Donthineni:
Tbh with yall, in my opinion, material design isn't like a law book, we can take what works for our app. Heck, even google's own apps don't obey material design sometimes. 

So, what I suggest is keeping the drawer icon and dropping the up button, the drawer can be shorthand for many things like downloads, home, trending, etc. Up button will take you to home, I think drawer gives more functionality. "Home" in drawer still emulates the up button, it's 2 presses away rather than 1. But, it's a good trade given navigation drawer is giving a variety of functionality.

surrodox2001:
> 
> 
> Tbh with yall, in my opinion, material design isn't like a law book, we can take what works for our app. Heck, even google's own apps don't obey material design sometimes.
> 
> So, what I suggest is keeping the drawer icon and dropping the up button, the drawer can be shorthand for many things like downloads, home, trending, etc. Up button will take you to home, I think drawer gives more functionality. "Home" in drawer still emulates the up button, it's 2 presses away rather than 1. But, it's a good trade given navigation drawer is giving a variety of functionality.

I agree will your decision to stay with your way as almost every android device have back buttons or touch points on them to make the "up" (back) button redundant.

wb9688:
Up is not the same as back. I think it shouldn't show the hamburger, but keep the up button, and then just allow opening the drawer from swiping.

surrodox2001:
> 
> 
> Up is not the same as back. I think it shouldn't show the hamburger, but keep the up button, and then just allow opening the drawer from swiping.

Seems you've the same conclusions as @mauriciocolli isn't it?

I think the up buttton should be removed and hamburger reintroduced as most android devices have back buttons on them that have the same function as up buttons in app UI.

opusforlife2:
> Up button will take you to home, I think drawer gives more functionality.

That's the thing. The up button is not 'just another home button' or a duplicate of the back button. It goes one level up in the app screen hierarchy. For example, if you go to the search fragment, then open a search result, tapping the up button takes you one level up to the search fragment, and _not_ the home screen, which needs another tap of the button.

> I think the up buttton should be removed and hamburger reintroduced as most android devices have back buttons on them that have the same function as up buttons in app UI.

Another thing to note here is that you can open several further videos after that first one, but the up button will _still_ take you to the search fragment, showing how it's different from the OS back button.

Personally, I don't use it. But I can imagine a situation where someone has been opening several videos in a chain, reached somewhere unrelated, and has forgotten the search term. Having to go back all the way with the OS back button would be irritating, if the up button were not there anymore.

Anyway, at this point, in my head I'm going "Just put both the buttons side by side! I don't care anymore! (╯°□°）╯︵ ┻━┻ " because I really want the drawer.

surrodox2001:
> Anyway, at this point, in my head I'm going "Just put both the buttons side by side! I don't care 
> anymore! (╯°□°）╯︵ ┻━┻ " because I really want the drawer.

Best choice really.

Stypox:
Now that Unified UI was merged, and the drawer is accessible from the video detail fragment, is this still relevant?

surrodox2001:
> 
> 
> Now that Unified UI was merged, and the drawer is accessible from the video detail fragment, is this still relevant?

I think this pull is effectively 'closed' because the things has done in an 'another' way, so... (and this' for the old ui anyway, so more irrelevant?)

PR #3160, opened 2020-02-29
Hide 5, 15, 25 second seek options if inexact seek is enabled

### LINKED ISSUE #3127: [Bug] Seek duration of 5 seconds makes audio rewind impossible

- Set the seek duration to 5 seconds.
- Play an audio till 30 seconds or so.
- Try rewinding.

It will do so once until the timestamp is a multiple of 10 seconds. Then tapping rewind just keeps taking you to that timestamp, not allowing you to go further back.

If you rapidly jam the rewind button, though, you overcome this 'barrier' of sorts and get taken quite far back.

This issue exists only for the 5 second duration.

XiangRongLin:

closes #3127 

Hides 5, 15, 25 second seek options if inexact seek is enabled. If one of those values was selected while activating inexact seek, it automatically gets set to 10 seconds.

Will the edited string resource make any problems concerning translation with weblate, since it was not a rewrite but adding additional info.

I tested it on my Android 10 and 7.1 devices.
Updated apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4283025/app-debug.zip)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
@@ -7,6 +7,7 @@ import android.os.Bundle;
 import android.provider.Settings;
 
 import android.text.format.DateUtils;
+import android.widget.Toast;
 import androidx.annotation.Nullable;
 import androidx.preference.ListPreference;
 
@@ -86,8 +87,16 @@ public class VideoAudioSettingsFragment extends BasePreferenceFragment {
         durations.setEntries(displayedDescriptionValues.toArray(new CharSequence[0]));
         final int selectedDuration = Integer.parseInt(durations.getValue());
         if (selectedDuration / (int) DateUtils.SECOND_IN_MILLIS % 10 == 5) {
-            durations.setValue(
-                Integer.toString(selectedDuration + 5 * (int) DateUtils.SECOND_IN_MILLIS));
+            final int newDuration = selectedDuration / (int) DateUtils.SECOND_IN_MILLIS + 5;
+            durations.setValue(Integer.toString(newDuration * (int) DateUtils.SECOND_IN_MILLIS));
+
+            Toast toast = Toast.makeText(getContext(),
+                getString(R.string.new_seek_duration_toast) + " " + String.format(
+                    res.getQuantityString(R.plurals.dynamic_seek_duration_description,
+                        newDuration),
+                    newDuration),
+                Toast.LENGTH_LONG);
+            toast.show();
         }
     }
 
--- a/app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
@@ -6,6 +6,7 @@ import android.os.Build;
 import android.os.Bundle;
 import android.provider.Settings;
 
+import android.text.format.DateUtils;
 import androidx.annotation.Nullable;
 import androidx.preference.ListPreference;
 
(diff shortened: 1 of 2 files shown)


opusforlife2:
Works as expected. 👍

Stypox:
>  If one of those values was selected while activating inexact seek, it automatically gets set to 10 seconds.

I think it should be rounded up instead of always being set to 10:
- 5 -> 10
- 15 -> 20
- 25 -> 30

Stypox:
Also, would it be possible to add a description to "Fast-forward/-rewind seek duration", too, so that the user knows why 5,15,25 seek durations are not available?

XiangRongLin:
> I think it should be rounded up instead of always being set to 10:
>
>    5 -> 10
>   15 -> 20
>    25 -> 30

Can be done.

>Also, would it be possible to add a description to "Fast-forward/-rewind seek duration", too, so that the user knows why 5,15,25 seek durations are not available?

The description field is already used to display the selected value.

> Another option would be to show a Toast to the user when the "Fast-forward/-rewind seek duration" setting was changed automatically after he activated "Use fast inexact seek".

I would do this then.

app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
```diff
@@ -58,10 +46,59 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+                displayedDurationValues.add(durationsValue);
+                displayedDescriptionValues.add(String.format(
+                    res.getQuantityString(R.plurals.dynamic_seek_duration_description,
+                        currentDurationValue),
+                    currentDurationValue));
+            } catch (Resources.NotFoundException ignored) {
+                //if this happens, the translation is missing, and the english string will be displayed instead
+            }
+        }
+        final ListPreference durations = (ListPreference) findPreference(getString(R.string.seek_duration_key));
```
Stypox: Add a newline here:
```suggestion

        final ListPreference durations = (ListPreference) findPreference(getString(R.string.seek_duration_key));
```

app/src/main/res/values/strings.xml
```diff
@@ -593,6 +593,7 @@
     <string name="app_language_title">App language</string>
     <string name="systems_language">System default</string>
     <string name="dynamic_seek_duration_description">%s seconds</string>
+    <string name="new_seek_duration_toast">Due to ExoPlayer contraints the seek duration was set to</string>
```
Stypox: Shouldn't this have a `%s` indicating where the "X seconds" string should be inserted? For English it is ok for this to be at the end, but for other languages this could be different.

XiangRongLin: Good point, didn't think of other languages having a different arrangement

XiangRongLin:
I'll update the apk tomorrow, I'm on mobile now.

Stypox:
No problem, thank you for the time you are putting into contributing ;-)

XiangRongLin:
@Stypox I updated the apk and rebased the branch onto dev

app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
```diff
@@ -58,10 +46,58 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+            .getBoolean(res.getString(R.string.use_inexact_seek_key), false);
+
+        for (String durationsValue : durationsValues) {
+            currentDurationValue =
+                Integer.parseInt(durationsValue) / (int) DateUtils.SECOND_IN_MILLIS;
+            if (inexactSeek && currentDurationValue % 10 == 5) {
+                continue;
+            }
+            try {
+                displayedDurationValues.add(durationsValue);
```
Stypox: Even though this is not relative to this PR, I think this line should be before the try-catch, since it can't throw an exception, and for a moment it led me to thinking that for missing translations the value would not be added.
Also note the added newline.
```suggestion

            displayedDurationValues.add(durationsValue);
            try {
```

app/src/main/res/values/strings.xml
```diff
@@ -593,6 +593,7 @@
     <string name="app_language_title">App language</string>
     <string name="systems_language">System default</string>
     <string name="dynamic_seek_duration_description">%s seconds</string>
+    <string name="new_seek_duration_toast">Due to ExoPlayer contraints the seek duration was set to %d seconds</string>
```
```suggestion
    <string name="new_seek_duration_toast">Due to ExoPlayer constraints the seek duration was set to %d seconds</string>
```

PR #3165, opened 2020-03-01
Mute button

### LINKED ISSUE #2876: Mute button
It would be amazing if you could put a button to mute the app regardless of the volume of the smartphone, similar to what it has in MX Player, for example.  This function would be sensational.

karkaminski:
Closes #2876. 

Added "Mute button" for MainPlayer, BackgroundPlayer and PopupPlayer. 

[mute_button-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4271745/mute_button-debug.zip)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
@@ -397,8 +397,8 @@ public final class BackgroundPlayer extends Service {
         }
 
         @Override
-        public void onMuteUnmuteButtonClicled() {
-            super.onMuteUnmuteButtonClicled();
+        public void onMuteUnmuteButtonClicked() {
+            super.onMuteUnmuteButtonClicked();
             updatePlayback();
         }
 
--- a/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
@@ -30,15 +30,15 @@ import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.os.Build;
 import android.os.IBinder;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.annotation.RequiresApi;
-import androidx.core.app.NotificationCompat;
-
 import android.preference.PreferenceManager;
 import android.util.Log;
 import android.view.View;
 import android.widget.RemoteViews;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
+import androidx.core.app.NotificationCompat;
 
 import com.google.android.exoplayer2.PlaybackParameters;
 import com.google.android.exoplayer2.Player;
@@ -341,7 +341,6 @@ public final class BackgroundPlayer extends Service {
         @Override
         public void handleIntent(final Intent intent) {
             super.handleIntent(intent);
-
             resetNotification();
             if (bigNotRemoteView != null)
                 bigNotRemoteView.setProgressBar(R.id.notificationProgressBar, 100, 0, false);
@@ -389,7 +388,6 @@ public final class BackgroundPlayer extends Service {
         @Override
         public void onPrepared(boolean playWhenReady) {
             super.onPrepared(playWhenReady);
-            simpleExoPlayer.setVolume(1f);
         }
 
         @Override
--- a/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
@@ -399,6 +399,12 @@ public final class BackgroundPlayer extends Service {
         }
 
         @Override
+        public void onMuteUnmuteButtonClicled() {
+            super.onMuteUnmuteButtonClicled();
+            updatePlayback();
+        }
+
+        @Override
         public void onUpdateProgress(int currentProgress, int duration, int bufferPercent) {
             updateProgress(currentProgress, duration, bufferPercent);
 
--- a/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
@@ -341,6 +341,7 @@ public final class BackgroundPlayer extends Service {
         @Override
         public void handleIntent(final Intent intent) {
             super.handleIntent(intent);
+            
             resetNotification();
             if (bigNotRemoteView != null)
                 bigNotRemoteView.setProgressBar(R.id.notificationProgressBar, 100, 0, false);
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -542,7 +542,7 @@ public abstract class BasePlayer implements
     // Mute / Unmute
     //////////////////////////////////////////////////////////////////////////*/
 
-    public void onMuteUnmuteButtonClicled() {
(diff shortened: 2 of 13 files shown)


app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -532,6 +538,18 @@ public void onShuffleClicked() {
         if (simpleExoPlayer == null) return;
         simpleExoPlayer.setShuffleModeEnabled(!simpleExoPlayer.getShuffleModeEnabled());
     }
+    /*//////////////////////////////////////////////////////////////////////////
+    // Mute / Unmute
+    //////////////////////////////////////////////////////////////////////////*/
+
+    public void onMuteUnmuteButtonClicled() {
```
B0pol: Rename this here and in other players
```suggestion
    public void onMuteUnmuteButtonClicked() {
```

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
```diff
@@ -394,6 +400,16 @@ protected void setShuffleButton(final ImageButton shuffleButton, final boolean s
         shuffleButton.setImageAlpha(shuffleAlpha);
     }
 
+    protected void setMuteButton(final ImageButton muteButton, final boolean isMuted) {
+        if (isMuted) {
+            muteButton.setColorFilter(ContextCompat.getColor(getApplicationContext(), R.color.white));
+        } else {
+            muteButton.setColorFilter(ContextCompat.getColor(getApplicationContext(), R.color.gray));
+
+        }
```
Stypox:
```suggestion
        muteButton.setColorFilter(ContextCompat.getColor(getApplicationContext(), isMuted ? R.color.white : R.color.gray));
```

app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
```diff
@@ -676,4 +685,13 @@ private void onMaybePlaybackAdapterChanged() {
...
     }
+
+    private void onMaybeMuteChanged(){
+        if (player.isMuted()) {
+            muteButton.setColorFilter(ContextCompat.getColor(getApplicationContext(), R.color.white));
+        }
+        else {
+            muteButton.setColorFilter(ContextCompat.getColor(getApplicationContext(), R.color.gray));
+        }
+    }
```
Stypox:
```suggestion
        muteButton.setColorFilter(ContextCompat.getColor(getApplicationContext(), isMuted ? R.color.white : R.color.gray));
    }
```

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -327,6 +331,8 @@ protected void initPlayback(@NonNull final PlayQueue queue,
 
         if (playQueueAdapter != null) playQueueAdapter.dispose();
         playQueueAdapter = new PlayQueueAdapter(context, playQueue);
+
+        if (isMuted) simpleExoPlayer.setVolume(0);
```
Stypox: Just to make sure the volume is at maximum, since you removed this line: https://github.com/TeamNewPipe/NewPipe/pull/3165/files#diff-0b87dfa694cf6a7720aee912b16ab32dL392
```suggestion
        simpleExoPlayer.setVolume(isMuted ? 0 : 1);
```

app/src/main/java/org/schabi/newpipe/player/BackgroundPlayer.java
```diff
@@ -341,7 +341,6 @@ public void initPlayer(boolean playOnReady) {
         @Override
         public void handleIntent(final Intent intent) {
             super.handleIntent(intent);
-
```
Stypox: Readd this newline

Stypox:
I am not sure if the mute button should be in the place you put it into. It is not symmetric and looks out of place imho. I think a better place would be the top bar, with a mute menu item that changes **name** and icon color based on whether mute is active or not. I would also show the menu item with the option "ifRoom", so that if there is no place it is shown in the three-dot menu.
Also, for the icon, make sure the color follows the theme (i.e. the "active" color with the white theme should be black). 
IMG_20200302_160734

I tested the current apk and it works correctly, thanks ;-)

karkaminski:
Done! I think this functionality may require translation for other languages now ;) 
[mute_button-acitonbar.zip](https://github.com/TeamNewPipe/NewPipe/files/4288898/mute_button-acitonbar.zip)

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -275,6 +277,7 @@ public void handleIntent(Intent intent) {
         final float playbackPitch = intent.getFloatExtra(PLAYBACK_PITCH, getPlaybackPitch());
         final boolean playbackSkipSilence = intent.getBooleanExtra(PLAYBACK_SKIP_SILENCE,
                 getPlaybackSkipSilence());
+        final boolean isMuted = intent.getBooleanExtra(IS_MUTED, isMuted());
```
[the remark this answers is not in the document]
karkaminski: To be honest I haven't thought about that risk while writting it. And now I think there is no "null risk"... I think so because handleIntent() is only called when intent is received -> which happens only while switching between base/popup/bckgrnd player -> which have to be != null ass Intent comes from other player :) But maybe I am missing something. Please correct me if I am wrong.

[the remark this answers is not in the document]
Stypox: No, that's reasonable, you are right. But I'd put a null check anyway just to be sure, something like ()```suggestion
        final boolean isMuted = intent.getBooleanExtra(IS_MUTED, simpleExoPlayer == null ? false : isMuted());
```

File: app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java [hunk 7]
```diff
@@ -394,6 +400,11 @@ protected void setShuffleButton(final ImageButton shuffleButton, final boolean s
         shuffleButton.setImageAlpha(shuffleAlpha);
     }
 
+    protected void setMuteButton(final ImageButton muteButton, final boolean isMuted) {
+        muteButton.setColorFilter(ContextCompat.getColor(getApplicationContext(), isMuted ? R.color.white : R.color.gray));
```
Stypox: What if the theme is white theme? The active button color should be black in that case. I'd suggest adding two attributes in `attrs.xml` named something along the lines of `button_active` and `button_inactive` and giving them different color values in the `themes.xml` file, so that they are themed.

Stypox: What if the theme is white theme? The active button color should be black in that case. I'd suggest adding two attributes in `attrs.xml` named something along the lines of `button_active` and `button_inactive` and giving them different color values in the `themes.xml` file, so that they are themed.

karkaminski: White theme does not affect background color for Main video player. Background is always black for this player.

Stypox: Oh you are right, nevermind :man_facepalming: 
Then there should be different code for every player, at this point

karkaminski: For MainVideoPlayer icon change happens in setMuteButton (place you've mentioned above). 
For PopUp and Background, icon change happens for queue in ServicePlayerActivity in onMaybeMuteChanged. I did it similarly to other buttons, (e.g onPlayModeChanged also in ServicePlayerActivity)

opusforlife2:
The grey color of the icon breaks consistency with all the others.

Is it not possible to have 2 white icons that switch to one another when toggled? One, a speaker with sound waves, and the other with no sound waves?

PR #3195, opened 2020-03-07
Fix seek duration not showing

### LINKED ISSUE #3169: Seek duration not shown in languages other than english
Hello. I apologize in advance if I am not writing there. in Russian localization, for some reason, the text stopped showing, although in the main language (English), it is shown. why it happens?
Screenshot_2020-03-02-01-51-32-053_org schabi newpipe
Screenshot_2020-03-02-01-51-37-095_org schabi newpipe
#2738 #2740

B0pol:

Fixes #3169.

I put the already translated string into all cases where the plural is needed. Therefore, the translation may be innacurate, but I'll explain:

+ People translated for "%s seconds", in general, so it's already plural, even if it can be innacurate, it's better than blank.
+ If we don't put everywhere, sometimes it's blank. Same, innacurate is better than blank.
+ Translators will be notified that the strings has changed, and either they will confirm it's ok, or correct the string.

I also corrected the «one» plural translation for some languages I know how plural works (eg french, german, spanish…)

For Chinese, for all plurals, from IDEA: one or zero is never called, it's always other.
So with the same reasoning, I put the already translated word.

BTW, it may happen in the future to other languages that didn't made the translation for 0.18.6, but did it before we release the next version (0.19.0 hopefully).
So if someone is releasing a new version @TobiGr @Stypox  — and after Weblate sync — please check if the problem happens too, and do the same fix, or ping me and I'll quickly fix it. (It should not happen if you sync weblate twice, one after this is merged, and one right before release as usual).

### DIFF
--- a/app/src/main/res/values-ar/strings.xml
+++ b/app/src/main/res/values-ar/strings.xml
@@ -569,10 +569,10 @@
     <string name="videos_string">الفيديوهات</string>
     <plurals name="dynamic_seek_duration_description">
         <item quantity="zero">%s ثوانٍ</item>
-        <item quantity="one"/>
-        <item quantity="two"/>
-        <item quantity="few"/>
-        <item quantity="many"/>
-        <item quantity="other"/>
+        <item quantity="one">%s ثوانٍ</item>
+        <item quantity="two">%s ثوانٍ</item>
+        <item quantity="few">%s ثوانٍ</item>
+        <item quantity="many">%s ثوانٍ</item>
+        <item quantity="other">%s ثوانٍ</item>
     </plurals>
 </resources>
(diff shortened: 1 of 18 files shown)


B0pol:
> I think you should remove all of the `<item quantity="one">` plurals, since there is no "one second" in dynamic seek. Otherwise this seems good.

I don't think it's necessary. I guess Weblate will recreate them, or at least for languages that didn't translated this string. 
Plus that means someone can see this string in the future, and possibly use it for 1 second, but that would not work.
What's the point of restricting it? Performance issue? (No, and if it's the case we should delete all the unused strings in non english files)

Stypox:
Oh, ok, then I think this pr is ok

PR #3196, opened 2020-03-07
Fix visual glitch when exiting the app

### LINKED ISSUE #1173: Closing app animation glitch

Whenever i exit newpipe sometimes the last screen of the app pops in for a split second during the exit animation

mauriciocolli:
This is a pretty old issue, I'm surprised nobody was annoyed enough to fix it.

It was even added by me, 2 years ago. The reason was to provide smooth transitions when changing the theme, and it worked, but in recent versions of Android it doesn't seem to make a difference other than trigger this annoying glitch.

---

Closes #1173.

### DIFF
--- a/app/src/main/res/values-v21/styles_services.xml
+++ b/app/src/main/res/values-v21/styles_services.xml
@@ -1,69 +1,69 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <!-- YouTube -->
-    <style name="LightTheme.YouTube" parent="LightTheme.Switchable">
+    <style name="LightTheme.YouTube" parent="LightTheme">
         <item name="colorPrimaryDark">@color/light_youtube_statusbar_color</item>
     </style>
 
-    <style name="DarkTheme.YouTube" parent="DarkTheme.Switchable">
+    <style name="DarkTheme.YouTube" parent="DarkTheme">
         <item name="colorPrimaryDark">@color/dark_youtube_statusbar_color</item>
     </style>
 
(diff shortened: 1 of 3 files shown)


Stypox:
Could you provide an apk?

mauriciocolli:
Hey @Stypox @TobiGr, what do you guys think of automating this process?

Travis is doing that already for every pull request, so when the tests succeed, we could just make it upload the resulting APK in a repository.

Would save a lot of time for testers and us.

Stypox:
Yeah, that'd be great! IMO this should be done after #3133 is merged, so that we also have packages with non-conflicting ids

TobiGr:
> Travis is doing that already for every pull request, so when the tests succeed, we could just make it upload the resulting APK in a repository.

Good idea. But as @wb9688 pointed out in #3211, Travis does not allow access to secret keys in PRs, so it is difficult to upload them somewhere.

We might want to take a look into GitHub actions. Maybe we can use their artifacts feature. I've already tested this out in January as GitHub actions was new and I was able to upload artifacts for each PR build. See https://github.com/TobiGr/NewPipe/actions/runs/3169663 for the workflow file.  I did not look into this further, because it was not possible at this time to retrieve the artifact link. It is possible now: https://developer.github.com/v3/actions/artifacts/

bboa:
@mauriciocolli thanks a lot, looks perfect.

Poolitzer:
Btw @TobiGr I had a good experience switching to Github Actions from Travis. It should be easy enough to have a little bot throwing out the archive link as a new comment (otherwise it will be hidden in the test tab)

PR #3211, opened 2020-03-11
added apk upload and comment on pr in travis

yausername:
Implemented the feature requested in https://github.com/TeamNewPipe/NewPipe/pull/3196#issuecomment-596200424
Travis will put a comment like [this](https://github.com/yausername/NewPipe/pull/1#issuecomment-597382609) on every build of a PR.
`CI_USER_PASSWORD` environment variable has to be added in travis settings. (I can share this with someone who has access to travis settings.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,11 @@ android:
 
 before_install: 
   - yes | sdkmanager "platforms;android-28"
+  - echo -e "machine github.com\n  login newpipe-machine-user\n  password $CI_USER_PASSWORD" > ~/.netrc
+  - echo -e "machine api.github.com\n  login newpipe-machine-user\n  password $CI_USER_PASSWORD" >> ~/.netrc
+  - chmod 600 ~/.netrc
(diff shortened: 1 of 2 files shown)


wb9688:
Does this even work? Travis doesn't give access to secrets in cases of PRs, as someone could change the build scripts to print that secret and make a PR, and then they'll know the secret if that wasn't the case.

yausername:
> Does this even work? Travis doesn't give access to secrets in cases of PRs, as someone could change the build scripts to print that secret and make a PR, and then they'll know the secret if that wasn't the case.

Ok. I didn't think of that.

TobiGr:
Thanks for the effort. See my https://github.com/TeamNewPipe/NewPipe/pull/3196#issuecomment-598222796. If you want, feel free to use my workflow file and finish my draft.

yausername:
@TobiGr I wanted to add apk download link as a comment in PR. But it is not possible even with github actions because the `secrets.GITHUB_TOKEN` is read-only for PRs from forked repos.

Anyway, with your workflow we are able to upload the artifact which is useful nonetheless.

PR #3220, opened 2020-03-12
fix for #3204

### LINKED ISSUE #3204: Some WebM videos are still not seekable in 0.18.6
Back in 0.17.3, I created an issue #2676 reporting that YouTube WebM videos didn't have seeking support. Now in 0.18.6 (github build), I noticed that most WebM videos have are seekable, but occasionally some aren't. 

I explained there that just for comparision's sake, Windows application jDownloader2 is able to generate seekable WebM and mkv videos just fine, so this issue is on NewPipe's muxer. I prefer WebM container because most videos in this format are way smaller that in MP4, I'd say 80% of them are, so you might understand why fixing this issue is so important for me.

In #2676 issue thread I got a reply that this was fixed in pull request #2679, but checking that, it says the change was still required and that it was a regression from #2149. The change was listed there as "Webm muxer fixes and strings.xml changes".

I wonder if it's a problem with my phone? My device has VP9 hardware encoding, but I don't see how that would be causing this, as it doesn't dictate how the muxer works. Anyway, I think it's worth reporting this issue again, as it was only partially fixed. Thanks in advance.

kapodamy:
fixes https://github.com/TeamNewPipe/NewPipe/issues/3204
There was a division (using floating point) that was not precise, consequently some offstets was wrong written on the file. I had to refactor several things until I spot the bug.

- [ ✔ ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/streams/WebMReader.java
+++ b/app/src/main/java/org/schabi/newpipe/streams/WebMReader.java
@@ -37,6 +37,7 @@ public class WebMReader {
     private final static int ID_DefaultDuration = 0x3E383;
     private final static int ID_FlagLacing = 0x1C;
     private final static int ID_CodecDelay = 0x16AA;
+    private final static int ID_SeekPreRoll = 0x16BB;
 
     private final static int ID_Cluster = 0x0F43B675;
     private final static int ID_Timecode = 0x67;
@@ -332,6 +333,10 @@ public class WebMReader {
                         break;
                     case ID_CodecDelay:
                         entry.codecDelay = readNumber(elem);
+                        break;
+                    case ID_SeekPreRoll:
+                        entry.seekPreRoll = readNumber(elem);
+                        break;
                     default:
                         break;
                 }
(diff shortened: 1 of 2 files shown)


wb9688:
@TobiGr: I tested it in AVD with Android 21, 25, 27 and 29 (a 43" 2160p monitor is so handy for putting 4 emulators besides each other), and on my Samsung phone with Android 28. In all cases, it works in both https://www.youtube.com/watch?v=QrlgAvJI3WU and https://www.youtube.com/watch?v=5ufOfV40fx0.

ksio89:
> @TobiGr: I tested it in AVD with Android 21, 25, 27 and 29 (a 43" 2160p monitor is so handy for putting 4 emulators besides each other), and on my Samsung phone with Android 28. In all cases, it works in both https://www.youtube.com/watch?v=QrlgAvJI3WU and https://www.youtube.com/watch?v=5ufOfV40fx0.
> 
> Edit: from a quick look at the code, it seems OK, other than that `make_EBML_void()` shouldn't have underscores in the name. However, I don't know anything about muxing and stuff.

Guess it's some issue with my device then :( thanks for testing anyway.

PR #3230, opened 2020-03-15
Merge duplicated duration plurals and introduce new plurals for selected count 

B0pol:
Follows the discussion we had on IRC, and [this message](https://github.com/TeamNewPipe/NewPipe/pull/2309#issuecomment-599117006)


wb9688:
I'd prefer doing something like `<string name="seconds" translatable="false">@string/dynamic_seek_duration_description</string>` instead.

Stypox:
Why not renaming `dynamic_seek_duration_description` to just `seconds`?

B0pol:
Because it will cause problems with Weblate

Stypox:
Oh, ok

B0pol:
> I'd prefer doing something like `<string name="seconds" translatable="false">@string/dynamic_seek_duration_description</string>` instead.

I tested with subscribers to see what happens, it crashes, it's not possible with plurals. For that case, as we catch ResourcesNotFoundException, it will use english instead.

app/src/main/java/org/schabi/newpipe/local/subscription/dialog/FeedGroupDialog.kt
```diff
@@ -189,7 +189,8 @@ class FeedGroupDialog : DialogFragment() {
         val groupAdapter = GroupAdapter<GroupieViewHolder>()
         groupAdapter.spanCount = if (useGridLayout) 4 else 1
 
-        val selectedCountText = getString(R.string.feed_group_dialog_selection_count, this.selectedSubscriptions.size)
+        var subscriptionsCount = this.selectedSubscriptions.size
```
mauriciocolli: A `val` should be used here.

app/src/main/java/org/schabi/newpipe/local/subscription/dialog/FeedGroupDialog.kt
```diff
@@ -234,7 +235,8 @@ class FeedGroupDialog : DialogFragment() {
                     item.isSelected = isSelected
                     item.notifyChanged(PickerSubscriptionItem.UPDATE_SELECTED)
 
-                    val updateSelectedCountText = getString(R.string.feed_group_dialog_selection_count, this.selectedSubscriptions.size)
+                    subscriptionsCount = this.selectedSubscriptions.size
```
mauriciocolli: A `val` should be used here as well, as keeping a reference to that variable outside is not necessary.

mauriciocolli:
I missed this: `dynamic_seek_duration_description` was originally added missing cases, so [the translations that followed it are broken](https://hosted.weblate.org/translate/newpipe/strings/en/?q=dynamic_seek_duration_description#translations).

Maybe we should just delete it and let people submit them correctly? Or will we just let Weblate flag them all out of date? We'd have to wait to release after they are fixed though.

app/src/main/res/values/strings.xml
```diff
@@ -623,8 +623,11 @@
     <string name="feed_notification_loading">Loading feed…</string>
     <string name="feed_processing_message">Processing feed…</string>
     <string name="feed_group_dialog_select_subscriptions">Select subscriptions</string>
-    <string name="feed_group_dialog_selection_count">Selected: %d</string>
     <string name="feed_group_dialog_empty_selection">No subscription selected</string>
+    <plurals name="feed_group_dialog_selection_count">
+        <item quantity="one">Selected: %d</item>
```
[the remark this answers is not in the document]
B0pol: I just changed to plural, and you are right.

B0pol:
> I missed this: `dynamic_seek_duration_description` was originally added missing cases, so [the translations that followed it are broken](https://hosted.weblate.org/translate/newpipe/strings/en/?q=dynamic_seek_duration_description#translations).
> 
> Maybe we should just delete it and let people submit them correctly? Or will we just let Weblate flag them all out of date? We'd have to wait to release after they are fixed though.

Yeah, you're right because https://github.com/TeamNewPipe/NewPipe/commit/29edb8c8a1b08a64aee5019ea3efd44b82b40be2 broke it
But it was fixed in merged PR https://github.com/TeamNewPipe/NewPipe/pull/3195/

mauriciocolli:
The fixes doesn't seem to be happening though, [there's multiple languages that are broken yet](https://hosted.weblate.org/translate/newpipe/strings/en/?q=dynamic_seek_duration_description#translations).

What about the https://github.com/TeamNewPipe/NewPipe/pull/3230#pullrequestreview-374832908 comment, will you address it or there's something to it?

B0pol:
> The fixes doesn't seem to be happening though, there's multiple languages that are broken yet.

It's not fixed yet, because it hasn't been synchronised yet, it will be fixed when synchronised.
That's one of the problem with Weblate, the way it works currently is long if problems occur, or even translations, they usually are shipped in the next update (and it can take months).

> What about the #3230 (review) comment, will you address it or there's something to it?

>> A optimal way of doing this would involve blocking translation from Weblate, doing the changes, and then unlocking.

I can't do anything, I don't have weblate rights.

mauriciocolli:
> It's not fixed yet, because it hasn't been synchronised yet

🤦 I missed it, sorry. Yet another merge conflict.

> Look at history, it's false. More than 800 Strings have been translated in the last 6 days.

I meant that people had stopped translating for +15 hours when I commented, so we could take advantage of that time window.

> I can't do anything, I don't have Weblate rights.

I remembered about an old shared API key for Weblate and found it, so now I can issue commands to it.

As Weblate can't push here anymore (was disabled, some people thought that were too many frequent commits) and I think I can't push it to `dev` directly, we could use this PR to do the merge and fix it once and for all. @TobiGr already merged Weblate into `dev` 5 days ago (9f91043131535c739462d12b7046e4a3489ee42c), the difference now is only something like 8 commits.

It should just follow the repo automatically then (if it doesn't I can trigger a Weblate pull then, but it really should).

Let me know if you're available to do it.

B0pol:
> Let me know if you're available to do it.

I'm

PR #3240, opened 2020-03-17
YouTube Music search stuff

wb9688:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

See TeamNewPipe/NewPipeExtractor#291.

PR #3243, opened 2020-03-17
Show radio icon as uploader avatar in Youtube mixes 

Stypox:

~~Shows "100+ videos" when the stream count of a playlist is `ListExtractor.ITEM_COUNT_MORE_THAN_100` and shows "∞ videos" when the stream count of a playlist is `ListExtractor.ITEM_COUNT_INFINITE`.
Uses the global `utils.Localization.localizeStreamCount()` function everywhere, i.e. in `PlaylistFragment`, `PlaylistInfoItem` and `ChannelMiniInfoItem` to correctly localize the item count (before it was not used in Playlist-related classes).~~
Now only does as described in the title. As discussed below using YouTube's logo is too risky, so we are using a radio icon instead.

Depends on TeamNewPipe/NewPipeExtractor#280 @XiangRongLin

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
@@ -11,12 +11,12 @@ import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.ImageView;
 import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.appcompat.content.res.AppCompatResources;
 
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
@@ -26,8 +26,10 @@ import org.schabi.newpipe.database.playlist.model.PlaylistRemoteEntity;
 import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.ListExtractor;
 import org.schabi.newpipe.extractor.NewPipe;
+import org.schabi.newpipe.extractor.ServiceList;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.playlist.PlaylistInfo;
+import org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper;
 import org.schabi.newpipe.extractor.stream.StreamInfoItem;
 import org.schabi.newpipe.extractor.stream.StreamType;
 import org.schabi.newpipe.fragments.list.BaseListInfoFragment;
@@ -44,13 +46,13 @@ import org.schabi.newpipe.util.Localization;
 import org.schabi.newpipe.util.NavigationHelper;
 import org.schabi.newpipe.util.ShareUtils;
 import org.schabi.newpipe.util.StreamDialogEntry;
-import org.schabi.newpipe.util.ThemeHelper;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import de.hdodenhof.circleimageview.CircleImageView;
 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
 import io.reactivex.rxjava3.core.Flowable;
 import io.reactivex.rxjava3.core.Single;
@@ -58,6 +60,7 @@ import io.reactivex.rxjava3.disposables.CompositeDisposable;
 import io.reactivex.rxjava3.disposables.Disposable;
 
 import static org.schabi.newpipe.util.AnimationUtils.animateView;
+import static org.schabi.newpipe.util.ThemeHelper.resolveResourceIdFromAttr;
 
 public class PlaylistFragment extends BaseListInfoFragment<PlaylistInfo> {
     private CompositeDisposable disposables;
@@ -74,7 +77,7 @@ public class PlaylistFragment extends BaseListInfoFragment<PlaylistInfo> {
     private TextView headerTitleView;
     private View headerUploaderLayout;
     private TextView headerUploaderName;
-    private ImageView headerUploaderAvatar;
+    private CircleImageView headerUploaderAvatar;
     private TextView headerStreamCount;
     private View playlistCtrl;
 
@@ -301,8 +304,22 @@ public class PlaylistFragment extends BaseListInfoFragment<PlaylistInfo> {
 
         playlistCtrl.setVisibility(View.VISIBLE);
 
-        IMAGE_LOADER.displayImage(result.getUploaderAvatarUrl(), headerUploaderAvatar,
-                ImageDisplayConstants.DISPLAY_AVATAR_OPTIONS);
+        final String avatarUrl = result.getUploaderAvatarUrl();
+        if (result.getServiceId() == ServiceList.YouTube.getServiceId()
+                && (YoutubeParsingHelper.isYoutubeMixId(result.getId())
+                || YoutubeParsingHelper.isYoutubeMusicMixId(result.getId()))) {
+            // this is an auto-generated playlist (e.g. Youtube mix), so a radio is shown
+            headerUploaderAvatar.setDisableCircularTransformation(true);
+            headerUploaderAvatar.setBorderColor(
+                    getResources().getColor(R.color.transparent_background_color));
+            headerUploaderAvatar.setImageDrawable(AppCompatResources.getDrawable(requireContext(),
+                    resolveResourceIdFromAttr(requireContext(), R.attr.ic_radio)));
+
+        } else {
+            IMAGE_LOADER.displayImage(avatarUrl, headerUploaderAvatar,
+                    ImageDisplayConstants.DISPLAY_AVATAR_OPTIONS);
+        }
+
         headerStreamCount.setText(Localization
                 .localizeStreamCount(getContext(), result.getStreamCount()));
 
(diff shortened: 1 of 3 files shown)


wb9688:
This isn't for the InfoItem, you'll have to do that in PlaylistMiniInfoItemHolder. Also INFINITY should be separate from MORE_THAN_100_ITEMS imho.

Stypox:
@wb9688 done, and updated APK

Stypox:
@wb9688 done, and updated APK. Use `PlaylistExtractor.MORE_THAN_100_ITEMS` in the `PlaylistMiniInfoItem` when you will implement it, even though the constant is in a different class I think it relates to playlists anyway, so it is ok to keep it in `PlaylistExtractor` imo

wb9688:
@Stypox: I implemented the `PlaylistMiniInfoItemHolder` part in #3240. Also, you should rebase on dev.

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -284,8 +285,15 @@ public void handleResult(@NonNull final PlaylistInfo result) {
 
         imageLoader.displayImage(result.getUploaderAvatarUrl(), headerUploaderAvatar,
                 ImageDisplayConstants.DISPLAY_AVATAR_OPTIONS);
-        headerStreamCount.setText(getResources().getQuantityString(R.plurals.videos,
-                (int) result.getStreamCount(), (int) result.getStreamCount()));
+
+        int streamCount = (int) result.getStreamCount();
+        if (streamCount == PlaylistExtractor.MORE_THAN_100_ITEMS) {
```
XiangRongLin: I would prefer a switch case here

Stypox: Mmmh you are right

XiangRongLin:
@Stypox the -2 appears in the saved playlists too

wb9688:
@XiangRongLin: Like I said above, that's fixed in #3240

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
Stypox: Done, also moved everything to a global `localizeStreamCount` function that is also used in PlaylistInfoItem and ChannelInfoItem and correctly localizes the item count.

Stypox:
Updated apk that also fixes the problem with PlaylistInfoItem: [app-debug_2.zip](https://github.com/TeamNewPipe/NewPipe/files/4361609/app-debug_2.zip)

app/src/main/java/org/schabi/newpipe/info_list/holder/PlaylistMiniInfoItemHolder.java
```diff
@@ -36,7 +37,8 @@ public void updateFromItem(final InfoItem infoItem, final HistoryRecordManager h
         final PlaylistInfoItem item = (PlaylistInfoItem) infoItem;
 
         itemTitleView.setText(item.getName());
-        itemStreamCountView.setText(String.valueOf(item.getStreamCount()));
+        itemStreamCountView.setText(Localization.localizeStreamCount(itemStreamCountView.getContext(),
```
wb9688: Here it should say just the count like in my PR. So e.g. "100+" instead of "100+ videos".

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -145,7 +148,14 @@ public static String localizeViewCount(Context context, long viewCount) {
     }
 
     public static String localizeStreamCount(Context context, long streamCount) {
-        return getQuantity(context, R.plurals.videos, R.string.no_videos, streamCount, localizeNumber(context, streamCount));
+        switch ((int) streamCount) {
```
wb9688: Why are you casting to `int`?

app/src/main/java/org/schabi/newpipe/info_list/holder/PlaylistMiniInfoItemHolder.java
Stypox: No, to keep consistenct, the string should contain "videos":
"No videos"
"1 video"
"2 videos"
...
"100+ videos"
"∞ videos"

app/src/main/java/org/schabi/newpipe/util/Localization.java
Stypox: Because Java `switch` does not allow the usage of `long`

app/src/main/java/org/schabi/newpipe/info_list/holder/PlaylistMiniInfoItemHolder.java
wb9688: Do you really think there's enough space for that there? 
Screenshot

Stypox: Oh, sorry, that's right ;-)

app/src/main/java/org/schabi/newpipe/util/Localization.java
wb9688: Then just use an `if`, as this would cause issues when the `streamCount` Is e.g. 4294967294.

Stypox:
Another apk: [app-debug_3.zip](https://github.com/TeamNewPipe/NewPipe/files/4361743/app-debug_3.zip)

app/src/main/java/org/schabi/newpipe/util/Localization.java
Stypox: Yeah but I don't think a stream count can reach 2^31 - 1 or  2147483647 or 2*10^10, that's really big. Also, it would be pointless to prevent issues here, if `streamCount` has to be cast to `int` anyway in order to pass it to `getQuantity`.

wb9688: I didn't see `getQuantity` needed an `int`, so in that case, that's fine.

wb9688:
I think this PR is good now, however it depends on the YT mix NPE PR, which hasn't been reviewed yet.

Stypox:
@XiangRongLin @TobiGr here I detected the uploader url is Youtube and replaced the image with the radio, while also hiding the white border. What do you think?

TobiGr:
@Stypox looks good to me :+1:

opusforlife2:
The radio icon looks weird, cut off from the bottom two vertices as it is. Is that unavoidable?

wb9688:
@opusforlife2: That's because avatars are round. We could make the radio icon smaller though, or maybe programatically replace the circle with a normal image view in this case.

Stypox:
> The radio icon looks weird, cut off from the bottom two vertices as it is. Is that unavoidable?

I saw that, but I didn't think it was really a problem

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -282,10 +284,21 @@ public void handleResult(@NonNull final PlaylistInfo result) {
 
         playlistCtrl.setVisibility(View.VISIBLE);
 
-        imageLoader.displayImage(result.getUploaderAvatarUrl(), headerUploaderAvatar,
-                ImageDisplayConstants.DISPLAY_AVATAR_OPTIONS);
-        headerStreamCount.setText(getResources().getQuantityString(R.plurals.videos,
-                (int) result.getStreamCount(), (int) result.getStreamCount()));
+        String avatarUrl = result.getUploaderAvatarUrl();
+        if (avatarUrl.contains("upload.wikimedia.org")) {
```
TobiGr: I think checking whether the uploader name is "YouTube" and the uploaderUrl is better here

TobiGr:
```suggestion
        if (result instanceof YoutubeMixPlaylistExtractor) {
```

TobiGr:
```suggestion
        if (YouTubeParsingHelper.isYouTubeMixId(result.getId())) {
```

wb9688: That's not a good idea, unless you first check whether `result.getService()` equals YouTube's service idea.

Stypox: @wb9688 is `result.getServiceId() == ServiceList.YouTube.getServiceId()` ok?

wb9688: `result.getServiceId() == ServiceList.YouTube.getServiceId() && YouTubeParsingHelper.isYouTubeMixId(result.getId())` is OK.

Stypox:
@opusforlife fixed; @TobiGr fixed;
Another test apk: [app-debug_4.zip](https://github.com/TeamNewPipe/NewPipe/files/4377440/app-debug_4.zip)

opusforlife2:
While testing, I found that Newpipe doesn't show private videos in the playlist, but counts them in the tally.

For example: https://www.youtube.com/playlist?list=PL3CB4CB7A114C1070

Invidious shows the private videos in the playlist. Newpipe should as well.

opusforlife2:
Wait, what is the reason for showing "100+"? Searching for "huge playlist" on Youtube gives a 5000 video playlist. Search for the same on Youtube Music using wb9688's apk shows the _same_ playlist having 100+ videos.

wb9688:
@opusforlife2: NewPipe loads the count at the search screen from the service's search results page. When you search for a playlist at music.youtube.com, you'll see 100+ in the search results. Once you click on the playlist in NewPipe, it will indeed be loaded from youtube.com, where it does show the actual count.

opusforlife2:
Alright. I get that.

But I still don't know what problem this PR is addressing. "huge playlist" still shows the 5000 video one in this apk.

Stypox:
@opusforlife2 this PR introduces the constants, even though only two of them are really used (unknown and infinite). I put the third one (100+) in here, too, because @wb9688 needed it for his yt music PR and the change was very strictly related to this one, so I was better off just putting related things together ;-).

opusforlife2:
Oh. 🤦‍♂ 
> But I still don't know what problem this PR is addressing. "huge playlist" still shows the 5000 video one in this apk.

I was testing this PR's apk when I should have been looking at the YT Music apk. I can indeed see that the 5000 video playlist shows as 100+ when the search is directed towards the YT Music front end.

Stypox:
I tested this and it works on API 19, API 29 (TV) and API 30

PR #3262, opened 2020-03-23
Use clean URL for sharing and opening in browser

### LINKED ISSUE #3247: "Clean" URL when sharing (remove all parameters that are not video ID)

It would be great, if NewPipe would only use video ID when sharing the link (`v=…`) instead of the whole URL that was used to open the video (often including source from which we got the video).

This is somewhat related to #1556

opusforlife2:
#### What is it?
- [x] Bug fix

- The share and open in browser options use a clean URL, free of extraneous URL parameters.

- fixes #3247

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
@@ -64,6 +64,8 @@ public class PlaylistFragment extends BaseListInfoFragment<PlaylistInfo> {
 
     private RemotePlaylistManager remotePlaylistManager;
     private PlaylistRemoteEntity playlistEntity;
+
+    private String cleanUrl;
     /*//////////////////////////////////////////////////////////////////////////
     // Views
     //////////////////////////////////////////////////////////////////////////*/
@@ -226,10 +228,10 @@ public class PlaylistFragment extends BaseListInfoFragment<PlaylistInfo> {
                 NavigationHelper.openSettings(requireContext());
                 break;
             case R.id.menu_item_openInBrowser:
-                ShareUtils.openUrlInBrowser(requireContext(), url);
+                ShareUtils.openUrlInBrowser(requireContext(), cleanUrl == null ? url : cleanUrl);
                 break;
             case R.id.menu_item_share:
-                ShareUtils.shareUrl(requireContext(), name, url);
+                ShareUtils.shareUrl(requireContext(), name, cleanUrl == null ? url : cleanUrl);
                 break;
             case R.id.menu_item_bookmark:
                 onBookmarkClicked();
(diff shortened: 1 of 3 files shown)


Stypox:
Have you checked this for every call of `ShareUtils.shareUrl()` and `ShareUtils.openUrlInBrowser()`?

opusforlife2:
I actually don't know where to find these "dirty" URLs to test. 😅 I just went by what Mauricio said here https://github.com/TeamNewPipe/NewPipe/issues/3247#issuecomment-602880441.

Apart from that I've tested all calls, and they all work. Testing both Share and Open in Browser is redundant, though. It's the same URL. If it doesn't work for one, it wouldn't work for the other case either.

Stypox:
@opusforlife2 you should also edit the `PlaylistFragment` class:
- in `handleResult()` store `result.getUrl()` in a class field (e.g. `cleanUrl`)
- in `onOptionsItemSelected()` share `cleanUrl` if it is not null, otherwise share the old `url`

opusforlife2:
You mean something like:
```java
public class PlaylistFragment extends BaseListInfoFragment<PlaylistInfo> {

    private String cleanUrl;
```
and
```java
    public void handleResult(@NonNull final PlaylistInfo result) {

        cleanUrl = result.getUrl();
```
and
```java
    case R.id.menu_item_openInBrowser:
        ShareUtils.openUrlInBrowser(requireContext(), (cleanUrl != null)? cleanUrl : url);
        break;
    case R.id.menu_item_share:
        ShareUtils.shareUrl(requireContext(), name, (cleanUrl != null)? cleanUrl : url);
        break;
```
?

Stypox:
Yes! But invert the `?:`: `cleanUrl == null ? url : cleanUrl`

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -226,10 +228,10 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 NavigationHelper.openSettings(requireContext());
                 break;
             case R.id.menu_item_openInBrowser:
-                ShareUtils.openUrlInBrowser(requireContext(), url);
+                ShareUtils.openUrlInBrowser(requireContext(), cleanUrl == null ? url : cleanUrl);
                 break;
             case R.id.menu_item_share:
-                ShareUtils.shareUrl(requireContext(), name, url);
+                ShareUtils.shareUrl(requireContext(), name, cleanUrl == null ? url : cleanUrl);
```
wb9688, line 234: You could just use `currentInfo.getUrl()` here as well

opusforlife2: So I should replace `cleanUrl == null ? url : cleanUrl` with `currentInfo.getUrl()`? I can basically revert the last commit and make this change instead, right?

opusforlife2:
Whoops. Forgot this was open.

Issue #3266, opened 2020-03-25
Release 0.19.0


TobiGr:
# Changelog
### New
- Correctly sorted feed and subscription groups #2309 
- Add mute button to main, background and popup players #3165, #3224, #3252, #3283

### Improved
- Handle content not available exception more comprehensively #3279
- Improve consistency with versal after colon #3107
- Hide 5, 15, 25 second seek options if inexact seek is enabled #3160
- Relocate two settings from Appearance to Content #3184 
- Allow opening music.youtube.com and media.ccc.de links in NewPipe #3177, #3090
- Remove search icon in search mode #3193 

### Fixed
- Fix bug in main screen tabs state management #3197
- Temporary fix for main screen menu visibility #3281 
- Fix visual glitch when exiting the app on some devices #3196 
- Fix seek duration not showing in some languages #3195
- Fix database backup on Android P and later not exporting remote playlists #3120 
- Fix downloading large files #3273
- Fix some WebM videos not being seekable due to a not precise floating point division #3220 
- Fix crash when sharing a downloaded file #3046 
- Fix URL parsing from shared content #3278 

### Development
- Change application id based on the current branch #3133, #3225
- New pull request and issue templates #3132

-------------------------------------
# DEBUG APK FOR TESTING

[NewPipe_v0.19.0_RC5.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4397798/NewPipe_v0.19.0_RC5.apk.zip)

Corresponding PR #3267

TobiGr:
RC2 with updated translations: 
[NewPipe_v0.19.0_RC2_debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4380633/NewPipe_v0.19.0_RC2_debug.apk.zip)

snappyapple632:
One subscription in my feed is not loading:

## Exception
* __User Action:__ requested feed
* __Request:__ Loading feed
* __Content Language:__ US
* __Service:__ none
* __Version:__ 0.19.0
* __OS:__ Linux Android 10 - 29


<details><summary><b>Crash log</b></summary><p>

```
org.schabi.newpipe.local.feed.service.FeedLoadService$RequestException: 0:https://www.youtube.com/channel/UCUgTZLBQUOE14NOu9iclvtA
	at org.schabi.newpipe.local.feed.service.FeedLoadService$RequestException$Companion.wrapList(FeedLoadService.kt:162)
[...]

PR #3252, opened 2020-03-21
Change mute button color for more visibility

B0pol:
#### What is it?
- [x] Bug fix ??
- [ ] Feature

#### Long description of the changes in your PR

The recently introduced mute button #3165 is not really visible the sound is unmuted, it's in gray. I made it white everytime, therefore it's more visible.

Screenshots:
Before: 
Screenshot_20200321-170322_NewPipe_Debug
After:
Screenshot_20200321-165830_NewPipe_mute_button

When it's muted, it stays the same icon, same color (it's the icon on the first screenshot but in white).

#### Agreement
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #3193, opened 2020-03-07
Remove search icon in search mode. Issue: #3183

jonas-skywalker:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

This is a preview of the changes that match the requested change in issue #3183 

newpipe

PR #3184, opened 2020-03-04
Relocate two settings from Appearance to Content

opusforlife2:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

"Show 'Next' and 'Similar' videos" and "Content of main page" have been relocated from Appearance to Content settings.

APK for testing: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4296606/app-debug.zip)

Closes #3166. Closes #3176.

PR #3267, opened 2020-03-25
Release 0.19.0

TobiGr:
# Changelog
### New
- Correctly sorted feed and subscription groups #2309 
- Add mute button to main, background and popup players #3165, #3224, #3252, #3283

### Improved
- Handle content not available exception more comprehensively #3279
- Improve consistency with versal after colon #3107
- Hide 5, 15, 25 second seek options if inexact seek is enabled #3160
- Relocate two settings from Appearance to Content #3184 
- Allow opening music.youtube.com and media.ccc.de links in NewPipe #3177, #3090
- Remove search icon in search mode #3193 

### Fixed
- Fix bug in main screen tabs state management #3197
- Temporary fix for main screen menu visibility #3281 
- Fix visual glitch when exiting the app on some devices #3196 
- Fix seek duration not showing in some languages #3195
- Fix database backup on Android P and later not exporting remote playlists #3120 
- Fix downloading large files #3273
- Fix some WebM videos not being seekable due to a not precise floating point division #3220 
- Fix crash when sharing a downloaded file #3046 
- Fix URL parsing from shared content #3278 

### Development
- Change application id based on the current branch #3133, #3225
- New pull request and issue templates #3132

# APK FOR TESTING

[NewPipe_v0.19.0_RC5.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4397798/NewPipe_v0.19.0_RC5.apk.zip)

Please report issue in #3266


app/build.gradle
```diff
@@ -12,8 +12,8 @@ android {
         resValue "string", "app_name", "NewPipe"
         minSdkVersion 19
         targetSdkVersion 28
-        versionCode 870
-        versionName "0.18.7"
+        versionCode 900
```
wb9688: Why increase by 30? It was a coincidence that e.g. 0.18.7 was 870 and it should just be increased by 10 like always imho. However now it's already too late.

app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
```diff
@@ -58,10 +46,59 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+            } catch (Resources.NotFoundException ignored) {
+                //if this happens, the translation is missing, and the english string will be displayed instead
+            }
+        }
+
+        final ListPreference durations = (ListPreference) findPreference(getString(R.string.seek_duration_key));
+        durations.setEntryValues(displayedDurationValues.toArray(new CharSequence[0]));
+        durations.setEntries(displayedDescriptionValues.toArray(new CharSequence[0]));
+        final int selectedDuration = Integer.parseInt(durations.getValue());
+        if (selectedDuration / (int) DateUtils.SECOND_IN_MILLIS % 10 == 5) {
```
XiangRongLin: I just noticed an error in one of my PR, it needs to be
`if (inexactSeek && selectedDuration / (int) DateUtils.SECOND_IN_MILLIS % 10 == 5) {`
Otherwise the seek durations get rounded up, even if inexact seek is disabled

app/build.gradle
opusforlife2: Why though? It's nice to be consistent. In fact, I would prefer 1910 for 0.19.1 and eventually, 10000 for 1.0.0. It's just symbolic.

app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
TobiGr: Thanks. I'll fix it in a minute

app/build.gradle
TobiGr: Yeah. I agree. 
I think we can do this when going 1.0.0 which should be soon after #2907 is merged

wb9688: Copy-paste of my argument on IRC:

>With his 10000 example it's still not possible to give a "correct" version code for e.g. 1.0.10. I don't think we should have ridiculously long version codes, though that doesn't really matter since we probably won't hit the 32-bit limit that soon. Also, version codes are purely for Android to check whether an APK is newer than the currently installed one. Version codes aren't for users to see, that's what version names are for. Google says: " Typically, you would release the first version of your app with versionCode set to 1, then monotonically increase the value with each release, regardless of whether the release constitutes a major or minor release. This means that the versionCode value does not necessarily have a strong resemblance to the app release version that is visible to the user (see versionName, below). Apps and publishing services should not display this version value to users." Iirc TheAssassin once argued that we shouldn't just increase by 1 every time, but by 10, which he had a reason for, so we switched to that. Imho it just doesn't make sense to somehow tie it to the version names.

app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
TobiGr: Fixed in https://github.com/TeamNewPipe/NewPipe/pull/3267/commits/55c577e76ec96fad3a52279eaea2201f8f05c2a7

app/build.gradle
TheAssassin: It's much easier and foreseeable IMO to just increase by the ever same amount. The interval of 10 isn't even really needed any more, but we should keep it. As @wb9688 showed, any "binding" to a version code doesn't work, there's always cases in which it breaks. A version code is 3 integers, not 1.

Issue #3277, opened 2020-03-26
Crash when sharing a certain playlist to Newpipe


opusforlife2:
### Version

- 0.18.7
### Steps to reproduce the bug

Share this playlist with Newpipe: https://www.youtube.com/playlist?list=PLoHg1UZiy8oG_j8mN7byXpybTONmYDXt_
### Expected behavior
Newpipe should open the playlist normally. (It opens normally if you paste this URL into the search bar.)

### Actual behaviour
Error.

### Logs
If your bug includes a crash, please head over to the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/). Copy the result. Paste it here:

## Exception
* __User Action:__ requested playlist
* __Request:__ https://www.youtube.com/playlist?list=PLoHg1UZiy8oGj8mN7byXpybTONmYDXt
* __Content Language:__ en_IN
* __Service:__ YouTube
* __Version:__ 0.18.7
* __OS:__ Linux Android 9 - 28

<details><summary><b>Crash log</b></summary><p>

```
org.schabi.newpipe.extractor.exceptions.ContentNotAvailableException: Got error: "The playlist does not exist."
	at org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeParsingHelper.defaultAlertsCheck(YoutubeParsingHelper.java:402)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubePlaylistExtractor.onFetchPage(YoutubePlaylistExtractor.java:43)
	at org.schabi.newpipe.extractor.Extractor.fetchPage(Extractor.java:56)
	at org.schabi.newpipe.extractor.playlist.PlaylistInfo.getInfo(PlaylistInfo.java:29)
	at org.schabi.newpipe.util.ExtractorHelper.lambda$getPlaylistInfo$8(ExtractorHelper.java:155)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$oXX6037sP0Rn784EpEbKFh6xH48.call(Unknown Source:4)
	at io.reactivex.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:44)
	at io.reactivex.Single.subscribe(Single.java:3438)
	at io.reactivex.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
	at io.reactivex.Single.subscribe(Single.java:3438)
[...]

PR #3279, opened 2020-03-27
Handle content not available exception more comprehensively

mauriciocolli:
#### What is it?
- [x] Bug fix
- [ ] Feature

Some fragments were not handling these types of exceptions.

We should improve the display of `ContentNotAvailableException` exceptions in the future by showing more specific details, like not found.

For now, a "Content unavailable" is better than a crash.

#### Relates to the following issue(s)
- #3277

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -51,7 +51,6 @@ import org.schabi.newpipe.download.DownloadDialog;
 import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.ServiceList;
-import org.schabi.newpipe.extractor.exceptions.ContentNotAvailableException;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1220,20 +1219,16 @@ public void openDownloadDialog() {
...
+        int errorId = exception instanceof YoutubeStreamExtractor.DecryptException
+                ? R.string.youtube_signature_decryption_error
+                : exception instanceof ParsingException
+                ? R.string.parsing_error
+                : R.string.general_error;
+        onUnrecoverableError(exception,
+                UserAction.REQUESTED_STREAM,
+                NewPipe.getNameOfService(serviceId),
+                url,
+                errorId);
```
TobiGr:
```suggestion
        int errorId = exception instanceof YoutubeStreamExtractor.DecryptException
                ? R.string.youtube_signature_decryption_error
                : exception instanceof ParsingException
                ? R.string.parsing_error
                : R.string.general_error;

        onUnrecoverableError(exception,
                UserAction.REQUESTED_STREAM,
                NewPipe.getNameOfService(serviceId),
                url,
                errorId);
```

mauriciocolli: Done.

PR #3281, opened 2020-03-27
Temporary fix for main screen menu visibility

mauriciocolli:
#### What is it?
- [x] Bug fix
- [ ] Feature

When restoring the state of the adapter, all the fragments' menu visibility were set to false, effectively disabling the menu from the user until he switched pages or another event that triggered the menu to be visible again happened.

As the bug happens deep inside the adapter restore method and the require field to fix it is private, a copy of the class was needed.

`FragmentStatePagerAdapter` is deprecated and should be replaced with its `ViewPager2` counterpart, until then, this should do it.

#### Testing apk

[fix-main-screen-menu-creation.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4394064/fix-main-screen-menu-creation.apk.zip)


XiangRongLin:
Your apk does solves the issue

PR #3287, opened 2020-03-28
what's new in chronological order

Eddy9517:
Hi, I'm really enjoying your app and I want to thank you for your amazing work!
But if can suggest an improvement, it would be really nice to have an option to set the "what's new" section in chronological order.
Thank you, have a great day!


wb9688:
1. Why this PR?
2. That's already implemented for v0.19.0, which will be released soon.

PR #3288, opened 2020-03-28
Save and restore playback parameters into/from preferences

XiangRongLin:
#### What is it?
- [x] Feature

- Playback parameters are speed, pitch and skip silence
- Save playback parameters into shared preferences whenever they are set
- Retrieve playback parameters from shared preferences when starting any player
- Remove playback parameters from being passed from one player to another through intents. Instead they can be retrieved from the shared preferences.

- solves #735

#### Testing apk
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4398458/app-debug.zip)

#### Agreement

#### Note
I noticed that changing the playback parameters while the video is paused gets ignored. At first I thought i broke something, until I noticed that it was like this before too.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -118,12 +118,6 @@ public abstract class BasePlayer implements
     @NonNull
     public static final String REPEAT_MODE = "repeat_mode";
     @NonNull
-    public static final String PLAYBACK_PITCH = "playback_pitch";
-    @NonNull
-    public static final String PLAYBACK_SPEED = "playback_speed";
-    @NonNull
-    public static final String PLAYBACK_SKIP_SILENCE = "playback_skip_silence";
-    @NonNull
     public static final String PLAYBACK_QUALITY = "playback_quality";
     @NonNull
     public static final String PLAY_QUEUE_KEY = "play_queue_key";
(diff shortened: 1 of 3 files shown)


XiangRongLin:
After seeing the other PR for the same issue I realized that the formatting/parsing of the float when saving/retrieving is completely unnecessary.

I'll remove that tomorrow.

XiangRongLin:
@wb9688 
I just removed the unnecessary formatting/parsing.

New apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4398458/app-debug.zip)
Data needs to be cleared, if you are just installing over my previous apk

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -272,11 +266,12 @@ public void handleIntent(Intent intent) {
             return;
         }
 
+        PlaybackParameters savedParameters = retrievePlaybackParametersFromPreferences();
```
wb9688: Could be `final`.

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -312,6 +307,19 @@ public void handleIntent(Intent intent) {
...
+    private PlaybackParameters retrievePlaybackParametersFromPreferences() {
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
+
+        float speed = preferences
+            .getFloat(context.getString(R.string.playback_speed_key), getPlaybackSpeed());
+        float pitch = preferences.getFloat(context.getString(R.string.playback_pitch_key),
+            getPlaybackPitch());
+        boolean skipSilence = preferences
+            .getBoolean(context.getString(R.string.playback_skip_silence_key),
+                getPlaybackSkipSilence());
```
wb9688: Could also all be `final`.

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -312,6 +307,19 @@ public void handleIntent(Intent intent) {
                 /*playOnInit=*/!intent.getBooleanExtra(START_PAUSED, false), isMuted);
     }
 
+    private PlaybackParameters retrievePlaybackParametersFromPreferences() {
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(context);
```
wb9688: Couldn't this be `final`?

Stypox:
I'm not completely sure if this is ok because in the background player there is a completely different probability of speeding up playback (I do it once in a blue moon) from other players (where I speed up things 95% of the times). So in the end I would have to tweak parameters anyway every time I switch players.

Stypox:
Though this is surely better than the current behaviour, so if no one comes up with a better solution this is the way to go.

Maybe making it so that:
- there is a different list of saved play

TobiGr:
@Stypox Our three players will be unified into one in #2907, so there won't be three players any more soon.

PR #3300, opened 2020-03-30
handle ContentNotSupportedException

B0pol:
#### What is it?
- [ ] Bug fix
- [ ] Feature
- [x] improvement

See https://github.com/TeamNewPipe/NewPipeExtractor/pull/300 and https://github.com/TeamNewPipe/NewPipeExtractor/pull/250
Copy paste from https://github.com/TeamNewPipe/NewPipeExtractor/pull/300
> Follows the discussion we had in https://github.com/TeamNewPipe/NewPipeExtractor/pull/250
It will greatly improve SoundCloud usability, because instead of randomly crashing, it will show the error with an appropriated UI, not a crash report, then the average user will understand why, and he can go back to play another track.
Before, the user was confused, and it went back to main kiosk each time, so it was really annoying.

How to Test: use debug apk below, search for "Dua Lipa" in SoundCloud, go to her channel page, and play the newest track, or go to youtube music.

Screenshots:
 

https://github.com/TeamNewPipe/NewPipeExtractor/pull/300

#### Testing apk
[NewPipe_contentnotsupported-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4444808/NewPipe_contentnotsupported-debug.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
@@ -6,6 +6,7 @@ import android.net.Uri;
 import android.os.Bundle;
 import android.text.TextUtils;
 import android.util.Log;
+import android.util.TypedValue;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
@@ -30,6 +31,7 @@ import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.ListExtractor;
 import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.channel.ChannelInfo;
+import org.schabi.newpipe.extractor.exceptions.ContentNotSupportedException;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.stream.StreamInfoItem;
 import org.schabi.newpipe.fragments.list.BaseListInfoFragment;
@@ -45,6 +47,7 @@ import org.schabi.newpipe.util.NavigationHelper;
 import org.schabi.newpipe.util.ShareUtils;
 
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
@@ -83,6 +86,9 @@ public class ChannelFragment extends BaseListInfoFragment<ChannelInfo> {
     private LinearLayout headerPopupButton;
     private LinearLayout headerBackgroundButton;
     private MenuItem menuRssButton;
+    private TextView contentNotSupportedTextView;
+    private TextView kaomojiTextView;
+    private TextView noVideosTextView;
 
     public static ChannelFragment getInstance(final int serviceId, final String url,
                                               final String name) {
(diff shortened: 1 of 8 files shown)


B0pol:
Travis fails because I used local NewPipeExtractor folder to compile, but travis doesn't have this folder.
Should I edit build.gradle?

B0pol:
Screenshot_20200408-205423_NewPipe_content_not_supported
@wb9688 I updated

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
```diff
@@ -439,6 +468,13 @@ public void handleResult(@NonNull final ChannelInfo result) {
                 .playOnBackgroundPlayer(activity, getPlayQueue(), false));
     }
 
+    private void updateNoVideoDisplay() {
```
wb9688, line 471: Rename this function to `showContentNotSupported`.

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
```diff
@@ -417,9 +431,24 @@ public void handleResult(@NonNull final ChannelInfo result) {
...
+            // handling ContentNotSupportedException not to show the error but an appropriate string
+            // so that crashes won't be sent uselessly and the user will understand what happened
+            Iterator<Throwable> it = errors.iterator();
+            while (it.hasNext()) {
+                Throwable throwable = it.next();
+                if (throwable instanceof ContentNotSupportedException) {
+                    updateNoVideoDisplay();
+                    it.remove();
+                }
+            }
```
wb9688, line 446: Can't you just use `for (Throwable throwable : errors)`?

B0pol: Is it `ConcurrentModificationException` free? I was told that I should use iterator if I want to remove or add things while iterating because it's always safe.

wb9688: @B0pol: You're right. <sup>I'm used to Python.</sup> You should still remove the outer `if` though, and I still prefer making a copy of the list, though that uses more memory. You're also currently assuming there could only be one `ContentNotSupportedException`, while you don't when just making a copy of the list.

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
```diff
@@ -417,9 +431,32 @@ public void handleResult(@NonNull final ChannelInfo result) {
...
+
+            if (!errors.isEmpty()) {
+                showSnackBarError(result.getErrors(), UserAction.REQUESTED_CHANNEL,
+                        NewPipe.getNameOfService(result.getServiceId()), result.getUrl(), 0);
+            }
+
+            if (contentNotSupported != null) {
```
TobiGr, line 455: Instead of adding the error at the end, it might be good to keep the original order. We should create a copy the error list.

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
TobiGr: I'd keep the `if`, but also agree to use a copy of the list to keep the original order.

PR #3344, opened 2020-04-03
Use DownloaderImpl in CheckForNewAppVersionTask

wb9688:
This way we only have to change code in DownloadImpl and the ExoPlayer stuff when we want to add support for e.g. setting a SOCKS proxy in the settings. I tested that this works successfully.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
+++ b/app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
@@ -19,6 +19,7 @@ import androidx.core.app.NotificationManagerCompat;
 
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.schabi.newpipe.extractor.exceptions.ReCaptchaException;
 import org.schabi.newpipe.report.ErrorActivity;
 import org.schabi.newpipe.report.UserAction;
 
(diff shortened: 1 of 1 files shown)


TobiGr:
Could you also reduce this to one case: 
https://github.com/TeamNewPipe/NewPipe/blob/41061d0289941c528cc1ccbcbe7d26fea9b722b7/app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java#L96-L104

``` Java
        } catch (NoSuchAlgorithmException | CertificateEncodingException ex) {
            ErrorActivity.reportError(APP, ex, null, null,
                    ErrorActivity.ErrorInfo.make(UserAction.SOMETHING_ELSE, "none",
                            "Could not retrieve SHA1 key", R.string.app_ui_crash));
        }
```

app/src/main/java/org/schabi/newpipe/CheckForNewAppVersionTask.java
```diff
@@ -178,21 +159,22 @@ protected void onPostExecute(final String response) {
...
             try {
-                JSONObject mainObject = new JSONObject(response);
-                JSONObject flavoursObject = mainObject.getJSONObject("flavors");
-                JSONObject githubObject = flavoursObject.getJSONObject("github");
-                JSONObject githubStableObject = githubObject.getJSONObject("stable");
+                final JsonObject githubStableObject = JsonParser.object().from(response)
+                        .getObject("flavors").getObject("github").getObject("stable");
+
+                System.err.println("appelflapsausautomaat");
+                System.err.println(JsonWriter.string(githubStableObject));
```
TobiGr, line 166: What's up with these two lines?

wb9688: What are you talking about? :P

PR #3373, opened 2020-04-05
Add search for subscription picker in the feed group dialog

mauriciocolli:
#### What is it?
- [ ] Bug fix
- [x] Feature
- [ ] Code base improvement
- [ ] Meta improvement to the project

Let the user search for subscriptions in the feed group dialog.

#### Testing apk

[feed-add-search-sub-list.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4840678/feed-add-search-sub-list.apk.zip)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/database/subscription/SubscriptionDAO.kt
+++ b/app/src/main/java/org/schabi/newpipe/database/subscription/SubscriptionDAO.kt
@@ -20,6 +20,13 @@ abstract class SubscriptionDAO : BasicDAO<SubscriptionEntity> {
     @Query("SELECT * FROM subscriptions ORDER BY name COLLATE NOCASE ASC")
     abstract override fun getAll(): Flowable<List<SubscriptionEntity>>
 
+    @Query("""
+        SELECT * FROM subscriptions
+        WHERE name LIKE '%' || :filter || '%'
+        ORDER BY name COLLATE NOCASE ASC
+        """)
+    abstract fun filterByName(filter: String): Flowable<List<SubscriptionEntity>>
+
     @Query("SELECT * FROM subscriptions WHERE url LIKE :url AND service_id = :serviceId")
     abstract fun getSubscriptionFlowable(serviceId: Int, url: String): Flowable<List<SubscriptionEntity>>
 
--- a/app/src/main/java/org/schabi/newpipe/database/subscription/SubscriptionEntity.java
+++ b/app/src/main/java/org/schabi/newpipe/database/subscription/SubscriptionEntity.java
@@ -130,4 +130,55 @@ public class SubscriptionEntity {
         item.setDescription(getDescription());
         return item;
     }
+
+
+    // TODO: Remove these generated methods by migrating this class to a data class from Kotlin.
+    @Override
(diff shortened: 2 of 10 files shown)


justanidea:
Oh damn! That's cool! Also, now that you are working on it again, what would you think about a checkbox to show only the subs that still arent in a group? I think it could be handy when you have a lot of subscritptions (>100). Any idea about it :) ?

app/src/main/java/org/schabi/newpipe/database/subscription/SubscriptionDAO.kt
```diff
@@ -16,6 +16,13 @@ abstract class SubscriptionDAO : BasicDAO<SubscriptionEntity> {
     @Query("SELECT * FROM subscriptions ORDER BY name COLLATE NOCASE ASC")
     abstract override fun getAll(): Flowable<List<SubscriptionEntity>>
 
+    @Query("""
+        SELECT * FROM subscriptions
+        WHERE name LIKE '%' || :filter || '%'
+        ORDER BY name COLLATE NOCASE ASC
```
B0pol, line 22: You made it diacritic sensitive, and it's really annoying for languages with diacritics, because you never know if people have respected diacritics or not.

From [this SO](https://stackoverflow.com/questions/2461522/how-do-i-perform-an-accent-insensitive-compare-e-with-%C3%A8-%C3%A9-%C3%AA-and-%C3%AB-in-sql-ser), collate Latin1_general_CI_AI should do the job
```suggestion
        ORDER BY name COLLATE Latin1_general_CI_AI
```

mauriciocolli: Yep, I did noticed that.

However, I don't think it is that easy to fix it. Android's SQLite doesn't have a collation available for this use case and I'm not sure [if SQLite itself even supports using it for the `LIKE` statement](https://www.sqlite.org/lang_expr.html#like).

Some solutions that I found is to either have a column (in some table) where you normalize these characters when inserting to it, or transforming the search query into a expression that treat those special characters as normal ones (e.g. `[aáàãAÁÀÃ...]`). We could also add support for the accent insensitive `LIKE` natively, but I'm not so sure about that...

Maybe I missed something about this in Android, so let me know if you find some native solution for this.

Still, for the order, `UNICODE` or `LOCALIZED` would've been better here (available in Android).

I'll leave these issues for another PR because that's how it's done in multiple queries across the app, would have to change them as well.

Stypox:
I found an issue while testing on Android 7.0: while filtering items with a search query, when clicking on those items nothing happens (i.e. no tick appears), but they are correctly added to the selected ones.

TobiGr:
@mauriciocolli What's the status of this PR and #3404?

mauriciocolli:
> @Stypox: I found an issue while testing on Android 7.0: while filtering items with a search query, when clicking on those items nothing happens (i.e. no tick appears), but they are correctly added to the selected ones. The checkbox appears as soon as the query changes.

I think I forgot to update the test APK when I last pushed, already fixed.

> @TobiGr: What's the status of this PR and #3404? 

Completed, just needed a rebase.

PR #3374, opened 2020-04-05
Fix visibility of group sort button in the subscriptions fragment

mauriciocolli:
#### What is it?
- [x] Bug fix
- [ ] Feature
- [ ] Code base improvement
- [ ] Meta improvement to the project

The visibility of the group sort button in the subscriptions fragment was not the right one when a rebind happened (e.g. return from another fragment that destroyed its host's views).


mauriciocolli:
> @TobiGr: I cannot reproduce the buggy behaviour with 0.19.2.

The bug is definitely there, here's the steps to reproduce it:

1. Have less than 2 groups.
2. Scroll down (enough so the header item is not visible when you return to the screen).
3. Go to another fragment, so the current one destroy its views (e.g. open a channel).
4. Go back.

The update message is not processed by the item because it isn't visible when the check happens in the fragment, that message is also not replayed so it is essentially lost.

Holding its state in a variable fix that because when the item do a rebind (it is visible again), it will check it and find the correct value.

PR #3394, opened 2020-04-08
Add EditorConfig

wb9688:
#### What is it?
- [ ] Bug fix (user facing)
- [ ] Feature (user facing)

This PR adds an [EditorConfig](https://editorconfig.org/). EditorConfig is some standard to set certain code style related settings in different editors and IDEs. I've also changed all files (except `gradlew.bat`) to use LF line ending, replaced all tabs with 4 spaces, have a newline at the end, and to not have trailing whitespace. I didn't look at whether files that were formatted with spaces are actually formatted with 4 spaces, or whether the indentation is correct.

#### Agreement


.editorconfig
```diff
@@ -0,0 +1,15 @@
+root = true
```
TobiGr, line 1: Please add a comment about the root thing here. IMO this is a little confusing. Something similar which is used in the official docs should be good:
```suggestion
# top-most EditorConfig file
root = true
```

wb9688:
>I'm a little bit worried if changing all of the strings.xml files could cause problems with weblate.

Maybe, cc @B0pol and @TobiGr. There are some issues that could actually be fixed through Weblate, however some others have to be fixed this way. I also saw some translations that don't have `%1$s` written correctly.

>Also, is it ok to edit changelogs of older versions (I assume it is, I'm pointing out this just to make sure)?

Yes, F-Droid only looks at the changelog in the commit that was tagged for that release.

B0pol:
I've a question: does this automatically set the rule for imports? Because often you remove the wildcard imports .*. But as it's automatically done by intellij, people revert that.
If it's possible to configure imports, please do it.

wb9688:
@B0pol: No, that's not possible in EditorConfig, as that's some IntelliJ-specific setting, however we could add `ij_java_class_count_to_use_import_on_demand = 999` and `ij_java_names_count_to_use_import_on_demand = 999` to it as IntelliJ supports their own non-standard settings in EditorConfig, which is technically invalid, but other EditorConfig implementations ignore it anyway, so we could safely add it. You could also manually set it in Settings -> Editor -> Code Style -> Java -> Imports, and we enforce not using wildcard imports through Checkstyle and Ktlint, and having alphabetically sorted imports in Kotlin through Ktlint (which IntelliJ doesn't do fully automatically, though that will be fixed by JetBrains/kotlin#3336 and for now we can just manually run the `formatKtlint` task when writing Kotlin).

PR #3406, opened 2020-04-09
Add possibility to open PeerTube links in NP for some instances (without needing to share to NP)

B0pol:
#### What is it?
- [ ] Bug fix (user facing)
- [x] improvement
- [ ] Feature (user facing)

Add support for opening PeerTube links in NewPipe without needing to share it to newpipe (see in manifest for the instances choosen).

No issue fixed, we discussed previously that the default instance should be added, I also added instances we are using for tests in extractor, and media assassinate you which is the instance where some NewPipe videos are hosted (cf [this video](https://media.assassinate-you.net/videos/watch/957fedb7-316f-4591-8293-b105b53d822e)), and some other popular peertube instances.

https://github.com/TeamNewPipe/NewPipeExtractor/pull/307 : for playlist part. It's really needed because otherwise it will cause lots of crashes "could not get name" as playlist follow this pattern : videos/watch/playlist, then current extractor think it's a video and not playlist

#### Testing apk
[peertube_playlists.zip](https://github.com/TeamNewPipe/NewPipe/files/4581137/peertube_playlists.zip)

#### Agreement


TobiGr:
@wb9688 That are ~ 4900 instances. I am not sure if we should add them all.

wb9688:
@TobiGr: It's "only" 449 instances. Why should we not add them to the manifest?

app/src/main/AndroidManifest.xml
```diff
@@ -135,189 +134,632 @@
...
+                <data android:host="v.mbius.io" />
+                <data android:host="vod.ksite.de" />
+                <data android:host="vod.mochi.academy" />
+                <data android:host="watch.44con.com" />
+                <data android:host="watch.snoot.tube" />
+                <data android:host="wetube.ojamajo.moe" />
+                <data android:host="widemus.de" />
+                <data android:host="www.videos-libr.es" />
+                <data android:host="www.yiny.org" />
+                <data android:host="xxxporn.co.uk" />
```
TobiGr, line 752: @wb9688 For example because of this. We just decided to not add support for a pornsite. But now we support opening porn URLs.

gkeegan:
Some of these instances host only pornography. Is putting link support for all of them without checking a great idea?

app/src/main/AndroidManifest.xml
wb9688: However it was already possible to add that instance to NewPipe, and those URLs aren't displayed anywhere in NewPipe itself anyway.

gkeegan: There are many others that are porn sites as well, such as doby.io. Literally the third one I checked randomly.

gkeegan: By adding native support it is different than allowing users to add their own.

wb9688: @gkeegan: The only thing this adds is when you go to e.g. doby.io in your browser, it will allow you to open it in NewPipe. We aren't adding it as a default instance or whatever. Also note that it isn't possible to dynamically add URLs to the list of sites the app can open.

gkeegan: It also adds it to the "supported links" page, which is going to be clogged with a bunch of sites that nobody cares about, and also will have porn sites in it. 

NewPipe is not and should not ever endorse such content by adding any native support for it or references in the code, either by adding link support or by adding as a service.

TobiGr: Okay. What do we do now? Which instances should we add to the manifest?
two suggestions:
- take the ~10 most popular
- add the ones @B0pol suggested (could you list them again, please?)

Apart from that we should put a note somewhere that we are willing to add support for opening links from more instances on request as long as they match our inclusion criteria.

B0pol: IMO:
+ most popular, definitely
+ the ones used for tests
+ user wanted instances, updated from time to time.

and you'll need to update [inclusion criteria](https://teamnewpipe.github.io/documentation/00_Prepare_everything/) because technically from this, this instance could be added.

To add instance, we should check mainly check if videos are marked as NSFW to avoid such cases(It should be blurred).
I'll push tomorrow a first list.

opusforlife2: Which kid is going to check the "supported links" list, of all things? I don't get the brouhaha around supporting Peertube links with adult content in them. We literally have a toggle that allows us to watch age restricted content on Youtube.

Newpipe shouldn't become a censor, IMO. Newpipe doesn't show adult content by default, which is why the toggle is off when you first open the app. The user has the freedom to enable it.

Which should be the case with Peertube as well. Discriminating between instances like this is unnecessary. As @wb9688 said, this just allows links from a site to be opened in the app, nothing more. You have to first go to the site and click on a video, at which point, who exactly are we "protecting"?

Anyway, the instances list has a toggle called "Sensitive videos" which can be set to Hide. This only shows instances that completely hide adult content by default. Maybe that list can be fetched programmatically.

gkeegan: The difference between YouTube's "restricted content" and the content on these sites is VERY large. I agree NewPipe shouldn't be a filter, but if someone wants to watch it, they can add it in the instances list, but we shouldn't add it into the code for link support and thus endorse it.

Personally this is less about protecting people and more about the ethics of supporting it natively. The industry preys on people who are vulnerable and then chews them up and spits them out broken and addicted to drugs. I want no part of a project that has anything to do with that.

opusforlife2: Adding it to the list of instances without link support would mean that the user could open the site but not share links to Newpipe, which could generate bug reports, since this behaviour would be different from YT, SC and MediaCCC.

Regarding the ethics part: Newpipe, being just a client, has nothing to do with the websites it fetches from. It doesn't give them ad revenue, or any other kind of benefit. In fact, you could argue that since it accesses their servers without giving anything back, it's kind of a drain on their resources, and hence their money, hurting them economically, on however small a scale. :P

Anyway, I don't think it really matters on a practical level. I've never even heard of anyone accessing Peertube to watch porn. Philosophically, I'm against censorship.

TobiGr: @opusforlife2 I agree with you about censorship. We are not in a position - either ethically or scientifically - to decide which content is good or bad, true or false, to be looked at or preferred, or not to be viewed. I am aware of the problematic that arises when opening not all instances is supported or worse, when some instances are deliberately blocked. Pornography was only one example, which immediately caught my eye by the URL. The question is whether we simply support opening all instances. We are in no way responsible for their content and users should be aware of it because they already got the URL.

They main problem I see, when adding all 449 instances, is crashing old devices, or at least create lags. Every shared URL would be checked against 449 patterns for PeerTube and some others for YouTube (+ invidio + hooktube), SoundCloud, mediaCCC. And that's only for NewPipe. I should have communicated that clearer.
For this reason, I suggested to only add a few instances to the manifest. We can add more on request or even ask our users in a blog post for their most used instances. Adding the most popular for now does not prevent us from adding unknown or small instances later.

B0pol: More clarification: now, since PeerTube was added, since more than two years for other links, and for future not-in-manifest instances, you can still open a link with NewPipe by sharing to it. Then the linkHandlers will try to open or say Unsupported URL if NewPipe can't open.

The difference is: if it's in manifest, you will be able to set NewPipe to automatically open this link, without needing to share it to NewPipe.

opusforlife2: Well, that makes sense. Technical reasons supersede everything (well, mostly) so that is probably for the best.

> The difference is: if it's in manifest, you will be able to set NewPipe to automatically open this link, without needing to share it to NewPipe.

Oh. I see. I thought it would affect the share functionality. That clears it up.

TobiGr:
@B0pol added some instances now. Please suggest more instances if you are missing one.

B0pol:
Can you add this PR to 0.19.3 milestone?

PR #3410, opened 2020-04-10
Show error when video has age limit and setting is disabled

### LINKED ISSUE #776: Age restricted content option not working
When I turn off 'show age restricted content' I still get the same results as if it were enabled and NSFW items appear in my results.

wb9688:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

Now it will show an error for videos with an age limit when the option to view age restricted videos is disabled in the settings:

Screenshot of video detail fragment with error

Apparently there already was a bug in NewPipe that it will still show the tabs when an error occured, and this PR doesn't fix that yet, though it makes it much more noticable.

This PR will fix #776.

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -816,7 +816,7 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
                     isLoading.set(false);
                     if (result.getAgeLimit() != NO_AGE_LIMIT && !prefs.getBoolean(
                             getString(R.string.show_age_restricted_content), false)) {
-                        showError(getString(R.string.restricted_video), false);
+                        hideAgeRestrictedContent();
                     } else {
                         currentInfo = result;
                         handleResult(result);
@@ -1239,6 +1239,16 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
         }
     }
 
+    private void hideAgeRestrictedContent() {
+        showError(getString(R.string.restricted_video), false);
+
+        if (relatedStreamsLayout != null) { // tablet
+            relatedStreamsLayout.setVisibility(View.INVISIBLE);
+        }
+
+        viewPager.setVisibility(View.GONE);
+        tabLayout.setVisibility(View.GONE);
+    }
 
     public void openDownloadDialog() {
         try {
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -103,6 +103,7 @@ import io.reactivex.disposables.Disposable;
 import io.reactivex.schedulers.Schedulers;
 
 import static org.schabi.newpipe.extractor.StreamingService.ServiceInfo.MediaCapability.COMMENTS;
+import static org.schabi.newpipe.extractor.stream.StreamExtractor.NO_AGE_LIMIT;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1325,4 +1335,6 @@ private void updateProgressInfo(@NonNull final StreamInfo info) {
                     animateView(detailPositionView, false, 500);
                 });
     }
+
+    private static class RestrictedVideoException extends Exception { }
```
Stypox, line 1339: Why create a new exception type to just use it to transfer information to a method? Why not just creating another method to handle restricted video infos?

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -811,9 +812,13 @@ public void startLoading(final boolean forceLoad) {
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe((@NonNull StreamInfo result) -> {
                     isLoading.set(false);
-                    currentInfo = result;
-                    handleResult(result);
-                    showContent();
+                    if (result.getAgeLimit() != NO_AGE_LIMIT) {
```
Stypox, line 815: I don't understand where you are checking if the user has disabled restricted videos

Stypox: Actually, enabling or disabling the option does nothing

wb9688: I forgot to commit that ;)

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
wb9688: Because I was having some issue when calling `showError` directly, but it seems to work fine now.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -806,19 +807,25 @@ public void startLoading(final boolean forceLoad) {
...
-                .subscribe((@NonNull StreamInfo result) -> {
+                .subscribe((@NonNull final StreamInfo result) -> {
                     isLoading.set(false);
-                    currentInfo = result;
-                    handleResult(result);
-                    showContent();
-                }, (@NonNull Throwable throwable) -> {
+                    if (result.getAgeLimit() != NO_AGE_LIMIT && !prefs.getBoolean(
+                            getString(R.string.show_age_restricted_content), false)) {
+                        showError(getString(R.string.restricted_video), false);
```
Stypox, line 819: Add a function at line 1243, this should fix the tab hiding issues. Comments are shown until the video is loaded, and are then hidden, but there is not much that can be done about this since when comments are loaded we do not yet know if the video is age restricted or not
```java
    private void hideAgeRestrictedContent() {
        showError(getString(R.string.restricted_video), false);

        if (relatedStreamsLayout != null) { // tablet
            relatedStreamsLayout.setVisibility(View.INVISIBLE);
        }

        viewPager.setVisibility(View.GONE);
        tabLayout.setVisibility(View.GONE);
    }
```
```suggestion
                        hideAgeRestrictedContent();
```

B0pol: hum, not sure whether we should hide comments.
Nor PeerTube neither YouTube do that.

PR #3414, opened 2020-04-10
Fix ReCaptcha Activity for another type of recaptcha page

### LINKED ISSUE #3134: Non Authorized Apps Captcha Verification

Screenshot_2020-02-23-19-02-26

I know it's a very old issue, maybe you are guys working on it, i can see now you guys added done (tik) button. This button goes back to video startup list. That's good, before it was stuck at white screen. But why this captcha behaving like that and why non-authorized issue continuosly appearing even after completing captcha.

I am now using latest version.

Let me know how to deal with it

### LINKED ISSUE #3272: ReCaptcha
Continous recaptcha requests with 18.7 and 19rc2
Does not seem to be a done button
Screenshots attached
Screenshot_20200325-093759
Screenshot_20200325-093523
Screenshot_20200325-093449

### LINKED ISSUE #3310: ReCAPTCHA works only with stock Android System Webview
On the latest (v.19.0) version of NewPipe, ReCAPTCHAs work only when the `WebView implementation` option under `Developer Settings` of Android is set to `Android System WebView` aka the stock Android webview.

Setting this to `Chrome`, and thus using the WebView provided by an installed Chrome version (like Beta/Dev) makes it so NewPipe endlessly gets captchas over and over again.

In both scenarios the captcha can be solved, but using any non-stock WebView implementation makes it so the captcha appears again in an endless loop.

I haven't looked into the code yet, but could this be related to how NewPipe gets the cookies from the WebView?

### LINKED ISSUE #3418: Create a Tor friendly app, please
# Describe the feature you want
A clear description would be. Please dear NewPipe Team, I was wondering if you could make the current NewPipe app able to handle the Tor Network better/friendlier. What I want, is for NewPipe to work with the Tor Network as in "flawless execution".

I was hoping for you guys to make it so whenever I change Tor Identity it refreshes the app, internally. Not so much as it restarts the app, but allows me to clear the network cache. So I can get right back into browsing. Meaning I still want to be able to load a video,  and whenever it fails, I press change Tor Identity and I continue browsing right away.

# Is your feature request related to a problem? Please describe it
Whenever I use Tor, I get reCAPTCHAs every ~5 minutes. A very annoying procedure to fix as I have to reQuit the app, both the background player and the app itself. Change Tor Identity, and reOpen the app. Sometimes multiple times before it works. And then refind the video I watched, or research the thing I was searching for.

# How will you/everyone benefit from this feature?
As Google works to fingerprint our lives, Tor is the only thing that is free to use and heavily protects users. Not a thing we should take lightly. I remember when I watched NewPipe videos without Tor enabled. Later that day I opened the third-party official YouTube app(Vanced). With my Google account logged in. I got recommendations of the videos I had just watched on NewPipe. Which is NOT what I want. NewPipe is the private client I wish for it to be. But without the Tor network I will never be able to enjoy it fully. :'c 

I had always wished for NewPipe to be my #1 YouTube client and so I can ditch the official one. As the features gets merged, I really really want to switch over fully. But not as long as this annoying reCAPTCHA interrupts everything, everytime. Please dear NewPipe Team, make my life easier. I beg of you!

Stypox:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

- Try to get cookies from pages even when they start loading, because non-html pages like ones with `pbj=1` never stop loading.
- Try to extract the cookie directly from the redirection url, by looking at the field `google_abuse=`.
- Add `GOOGLE_ABUSE_EXEMPTION=` to the youtube recaptcha cookies.
- Store cookies into shared preferences so that the user does not have to solve a recaptcha every time he opens the app

Fixes #3134 fixes #3272 fixes #3418 fixes #3310

#### Testing apk
@mehedihasanziku @baraa272 @LargePrime @tusch001 could you test this, if recaptchas still pop up for you? Also, @Ristovski, even though I didn't change anything about webviews, could you test, too?
[NewPipe_recaptcha-debug_1.zip](https://github.com/TeamNewPipe/NewPipe/files/4464377/NewPipe_recaptcha-debug_1.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
@@ -206,7 +206,7 @@ public class ReCaptchaActivity extends AppCompatActivity {
             Log.d(TAG, "handleCookies: cookies=" + (cookies == null ? "null" : cookies));
         }
 
-        if (cookies == null || foundCookies.contains(cookies)) {
+        if (cookies == null) {
             return;
         }
 
--- a/app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
@@ -1,6 +1,7 @@
 package org.schabi.newpipe;
 
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.os.Build;
 import android.os.Bundle;
 import android.util.Log;
@@ -19,6 +20,7 @@ import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.widget.Toolbar;
 import androidx.core.app.NavUtils;
+import androidx.preference.PreferenceManager;
 
 import org.schabi.newpipe.util.ThemeHelper;
 
@@ -159,6 +161,12 @@ public class ReCaptchaActivity extends AppCompatActivity {
         }
 
         if (!foundCookies.isEmpty()) {
+            // save cookies to preferences
+            final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(
+                    getApplicationContext());
+            final String key = getApplicationContext().getString(R.string.recaptcha_cookies_key);
+            prefs.edit().putString(key, foundCookies).apply();
+
             // give cookies to Downloader class
             DownloaderImpl.getInstance().setCookies(foundCookies);
             setResult(RESULT_OK);
@@ -170,7 +178,7 @@ public class ReCaptchaActivity extends AppCompatActivity {
     }
 
 
-    private void handleCookiesFromUrl(final @Nullable String url) {
+    private void handleCookiesFromUrl(@Nullable final String url) {
         if (MainActivity.DEBUG) {
             Log.d(TAG, "handleCookiesFromUrl: url=" + (url == null ? "null" : url));
         }
@@ -201,7 +209,7 @@ public class ReCaptchaActivity extends AppCompatActivity {
         }
     }
 
-    private void handleCookies(final @Nullable String cookies) {
+    private void handleCookies(@Nullable final String cookies) {
         if (MainActivity.DEBUG) {
             Log.d(TAG, "handleCookies: cookies=" + (cookies == null ? "null" : cookies));
         }
@@ -214,7 +222,7 @@ public class ReCaptchaActivity extends AppCompatActivity {
         // add here methods to extract cookies for other services
     }
 
-    private void addYoutubeCookies(final @NonNull String cookies) {
+    private void addYoutubeCookies(@NonNull final String cookies) {
         if (cookies.contains("s_gl=") || cookies.contains("goojf=")
                 || cookies.contains("VISITOR_INFO1_LIVE=")
                 || cookies.contains("GOOGLE_ABUSE_EXEMPTION=")) {
--- a/app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
@@ -7,17 +7,23 @@ import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.webkit.CookieManager;
+import android.webkit.WebResourceRequest;
 import android.webkit.WebSettings;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.widget.Toolbar;
 import androidx.core.app.NavUtils;
 
 import org.schabi.newpipe.util.ThemeHelper;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
 
 /*
  * Created by beneth <bmauduit@beneth.fr> on 06.12.16.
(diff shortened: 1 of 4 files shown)


Stypox:
Newest apk: [NewPipe_recaptcha-debug_1.zip](https://github.com/TeamNewPipe/NewPipe/files/4464377/NewPipe_recaptcha-debug_1.zip)

Ristovski:
@Stypox Just tested and it does indeed appear to fix the issue with non-native WebViews (#3310)

app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
```diff
@@ -137,23 +178,54 @@ private void saveCookiesAndFinish() {
...
+
+            try {
+                String abuseCookie = url.substring(abuseStart + 13, abuseEnd);
+                abuseCookie = URLDecoder.decode(abuseCookie, "UTF-8");
+                handleCookies(abuseCookie);
+            } catch (UnsupportedEncodingException | StringIndexOutOfBoundsException e) {
+                if (MainActivity.DEBUG) {
+                    e.printStackTrace();
+                    Log.d(TAG, "handleCookiesFromUrl: invalid google abuse starting at "
+                            + abuseStart + " and ending at " + abuseEnd + " for url " + url);
```
TobiGr, line 206: Is that information useful to find further cookies? We might want to show a snackbar to allow reporting the error and more importantly, the url

jniniinoinoino:
> Newest apk: [NewPipe_recaptcha-debug_1.zip](https://github.com/TeamNewPipe/NewPipe/files/4464377/NewPipe_recaptcha-debug_1.zip)

Hello. I tried this apk. First thing that happened was it gave me a CAPTCHA when I launched the app. I pressed the button to complete it and the reCAPTCHA box disappeared. I cannot do anything after that other than press :heavy_check_mark: mark. Then the CAPTCHA will reappear obviously. Pressing the reCAPTCHA button again does the same thing and disappears.

I don't really know what you did in this PR. I watched videos until I finally got another CAPTCHA and pressed back and changed Tor networks without quitting out of the app. Still got the CAPTCHA after reEntering the video. Please help me to understand what this PR is doing thanks.

app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
```diff
@@ -137,23 +178,54 @@ private void saveCookiesAndFinish() {
...
             return;
         }
 
         addYoutubeCookies(cookies);
-        // add other methods to extract cookies here
+        // add here methods to extract cookies for other services
     }
 
-    private void addYoutubeCookies(@NonNull final String cookies) {
+    private void addYoutubeCookies(final @NonNull String cookies) {
```
wb9688, line 225: Why this change? We have `final` after the `@`s everywhere…

Stypox: I don't know ;-)
Fixed

app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
Stypox: This piece of code has never been useful in my testing, since it just obtains a cookie that would anyway appear as a cookie in one of the subsequent requests. Though to prevent issues under different environment where the webview behaves strangely (who knows?), it doesn't hurt to obtain the cookie twice in a different way and call it a day.

I left a Log there just so that when I send people an apk to debug, I don't have to make a custom one that throws a random Exception with the log message inside (as I did with the other cookie testing apk), but I can just tell them to install a logcat viewer and I have the information needed.

Anyway, if the catch block is ever reached it means the url isn't useful to obtain a cookie, so there is no need to report anything to a normal user imho.

wb9688: However logcat apps need root access since Jelly Bean.

Stypox: Oh, do they? I didn't know ;-)
But what could I implement then, that would log everything and display it to the user after he presses "Ok" on a snackbar? Should I do some kind of bug report, but with logs instead of stack trace?

Stypox:
@jniniinoinoino from https://github.com/TeamNewPipe/NewPipe/issues/3418#issuecomment-612635224
Unfortunately there is nothing that can be done to prevent recaptchas and tracking through the cookie being saved, because without that cookie nothing would work. Maybe an option to delete the current cookie could be created, though there are already far too many options, so I don't feel like adding another one.

---
> the reCAPTCHA box disappeared

There has to be yet another different behaviour with recaptchas that only comes up sometimes. The recaptcha-solved page is never loaded and due to this no cookie is ever obtained. I'm investingating

mehedihasanziku:
Yes I am using this latest debug apk, but i tried it so many captcha page not loading at all completely blank white screen with top tik (√) bar. I don't think it's a network issue, because i am running everything else very smoothly even youtube also in browser.

wb9688:
I think this should be merged now, and just leave the other reCAPTCHA page to another PR.

Stypox:
I tried to understand in many ways what the problem is with the other recaptcha behaviour, but I still haven't understood anything because there seems to be no difference.
- I tried to reproduce by connecting my pc to the Tor network and opening the app in an emulator, and the recaptcha could always be solved even after changing network multiple times.
- I tried to reproduce by enabling the socks5 proxy using NetGuard, but that didn't trigger any recaptcha even though the IP had changed.
- I was given (by someone facing this problem) the HAR file containing all of the connections made by Firefox during the resolution of a recaptcha under the same conditions where it didn't work in the app, but I couldn't find any meaningful information: everything seems to be the same as with a resolution that works in NewPipe
- I tried to solve other manually-loaded recaptchas from the app (e.g. this: https://www.google.com/recaptcha/api2/demo), but it always worked

PR #3265, opened 2020-03-25
Change APK base name depending on the branch name for debug builds

TobiGr:
#### What is it?
- [ ] Bug fix
- [ ] Feature
- [x] Development improvement

#### Long description of the changes in your PR

Debug APKs are now named after their branch name as suggested by @opusforlife2  e.g. the debug APK for this branch is named `NewPipe_apk_base_name-debug.apk`
Release builds are not touched, as well as debug builds on dev and master branch.

#### Testing apk

not needed

#### Agreement
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #3415, opened 2020-04-10
Fix F-Droid build

TobiGr:
Fixes the behaviour described in https://github.com/TeamNewPipe/NewPipe/pull/3265#issuecomment-612102349

#### What is it?
- [x] Bug fix (build facing)
- [ ] Feature (user facing)

@Bubu told us:
> when doing a release build the apk is now called:
`NewPipe_head-release-unsigned.apk`. Also a file in the `META-INF` dir now ends up being called `NewPipe_head_release.kotlin_module` instead of `app_release.kotlin_module`.

For some reason the `archivesBaseName` Gradle property from debug builds was used in release builds. Forcing it to be `app` and changing the order of build flavors fixed the problem for me

- https://github.com/TeamNewPipe/NewPipe/pull/3265#issuecomment-612102349

#### Testing apk

@Bubu With the changes I could not reproduce the behaviour you mentioned. Can you verify this by running a build on your machine, please?
[app-release-unsigned.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4463031/app-release-unsigned.apk.zip)

#### Agreement

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -27,12 +27,6 @@ android {
     }
 
     buildTypes {
-        release {
-            minifyEnabled true
-            shrinkResources true
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
-        }
-
         debug {
             multiDexEnabled true
             debuggable true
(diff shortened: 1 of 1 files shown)


Bubu:
Looks, good. Seems that solves the problem I encountered.

B0pol:
I found another bug with apk name: if the branch has `_`, the `_` will be dropped for the apk name. See the debug apk for #3406.

Stypox:
@B0pol that's intended, as to be sure not to create issues with names

TobiGr:
@Bubu Thank you for testing.

I added a comment to ensure that the release build config is kept at the end of the list. Should be good to go now, I think.

opusforlife2:
I'm guessing it was this PR since I can see no other reason why: all debug apks generated have the name `app-debug`. Is there a way to fix it to have branch names again and still have F-Droid builds not break?

PR #3430, opened 2020-04-13
Display parent channel details

Royosef:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

Add sub-channel details to channel fragment & video detail fragment

https://github.com/TeamNewPipe/NewPipeExtractor/pull/313

#### Testing apk

[NewPipe_displayparentchanneldetails-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4472065/NewPipe_displayparentchanneldetails-debug.zip)

#### Agreement


settings.gradle
```diff
@@ -1 +1,7 @@
 include ':app'
+
+includeBuild('../NewPipeExtractor') {
+    dependencySubstitution {
+        substitute module('com.github.TeamNewPipe:NewPipeExtractor') with project(':extractor')
+    }
+}
```
B0pol, line 7: This should be used only for testing locally. When you publish the PR, you should put in [app/build.gradle](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/build.gradle#L139) `implementation 'com.github.TeamNewPipe:NewPipeExtractor:48b397567813a8b78a21d47e830e879b49e7036c' //your last newpipe extractor commit`.

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
```diff
@@ -425,6 +472,15 @@ public void handleResult(@NonNull final ChannelInfo result) {
             headerSubscribersTextView.setText(R.string.subscribers_count_not_available);
         }
 
+        if (!TextUtils.isEmpty(currentInfo.getParentChannelName())) {
+            headerParentChannelTitleView.setText(
+                    "Created by " + currentInfo.getParentChannelName());
```
B0pol, line 477: Extract this string to strings.xml

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1249,6 +1282,28 @@ private void hideAgeRestrictedContent() {
...
+        uploaderTextView.setVisibility(View.GONE);
+    }
+
+    private void displayBothUploaderAndParentChannel(final StreamInfo info) {
+        parentChannelTextView.setText(info.getParentChannelName());
+        parentChannelTextView.setVisibility(View.VISIBLE);
+        parentChannelTextView.setSelected(true);
+
+        if (!TextUtils.isEmpty(info.getUploaderName())) {
+            uploaderTextView.setText("By " + info.getUploaderName());
```
B0pol, line 1299: Extract this string to strings.xml

B0pol, line 1299: Extract this string to strings.xml

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
B0pol, line 477: Extract this string to strings.xml

settings.gradle
```diff
@@ -1 +1,7 @@
 include ':app'
+
```
B0pol, line 2: This should be used only for testing locally. When you publish the PR, you should put in [app/build.gradle](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/build.gradle#L139) `implementation 'com.github.TeamNewPipe:NewPipeExtractor:48b397567813a8b78a21d47e830e879b49e7036c' //your last newpipe extractor commit`.

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
```diff
@@ -394,6 +408,36 @@ private void updateSubscribeButton(final boolean isSubscribed) {
         return ExtractorHelper.getChannelInfo(serviceId, url, forceLoad);
     }
 
+     /*//////////////////////////////////////////////////////////////////////////
```
wb9688, line 411: Why does this have an extra space before it?

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
```diff
@@ -394,6 +408,36 @@ private void updateSubscribeButton(final boolean isSubscribed) {
...
+    // OnClick
+    //////////////////////////////////////////////////////////////////////////*/
+
+     @Override
```
wb9688, line 415: Why does this have an extra space before it?

Royosef:
> There is one thing I've noticed while testing: I'm not a fan of the possibility to go to parent channel page directly in video page. The room is already tiny, and it's now divided by two, on my large phone it's hard to go on the one you want, on my small phone it's simply impossible to choose.
> 
> Words are better than images: red is the separation between channel (up) and parent (down)
> .

@B0pol So, will the whole section navigate to the channel page? (`Les vidéos de Framasoft` in this case)

B0pol:
> > There is one thing I've noticed while testing: I'm not a fan of the possibility to go to parent channel page directly in video page. The room is already tiny, and it's now divided by two, on my large phone it's hard to go on the one you want, on my small phone it's simply impossible to choose.
> > 
> > Words are better than images: red is the separation between channel (up) and parent (down)
> > .
> 
> @B0pol So, will the whole section navigate to the channel page? (`Les vidéos de Framasoft` in this case) 

@wb9688 asked me to reply but we've already talked about that via IRC, I just make it public:

Yes, the whole section will go to `Les vidéos de Framasoft` in this case. Unless someone find a good way to access seamlessly to both, it should go to the (sub)channel page.

But don't remove this part from the extractor because it's not only used by NewPipe, or someone in another PR will success to integrate it seamlessly…

wb9688:
>Unless someone find a good way to access seamlessly to both, it should go to the (sub)channel page.

Maybe use long press for parent channel while using a normal tap for (sub)channel.

Royosef:
> > Unless someone find a good way to access seamlessly to both, it should go to the (sub)channel page.
> 
> Maybe use long press for parent channel while using a normal tap for (sub)channel.

Honestly, I don't know how much intuitive it will be for the users.

wb9688:
@Royosef: It's better than nothing, and we have long presses on other buttons on that page as well.

@TobiGr: What do you think?

TobiGr:
I agree with @wb9688. Long-press is better than nothing.

opusforlife2:
There is already a setting to show a tip for background and popup buttons. It could be modified to also show one for this so that users get to know of the change.

wb9688:
@opusforlife2: I don't think that's needed as it's currently only a PeerTube thing, and we don't have a tip for the download button either.

opusforlife2:
Alright. If it's a Peertube specific feature and not app wide, I can imagine it would be tougher to implement.

Yeah, I got to know of the download button from your reddit comments. There really should be a tool tip for it.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -463,6 +474,15 @@ public boolean onLongClick(final View v) {
             case R.id.detail_controls_download:
                 NavigationHelper.openDownloads(getActivity());
                 break;
+
+            case R.id.detail_uploader_root_layout:
+                if (TextUtils.isEmpty(currentInfo.getSubChannelUrl())) {
```
Stypox, line 479: Why is this checking whether a subchannel exists, but then the parent uploader url is used?

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1127,13 +1155,17 @@ public void handleResult(@NonNull final StreamInfo info) {
...
-            uploaderTextView.setVisibility(View.VISIBLE);
-            uploaderTextView.setSelected(true);
+        if (!TextUtils.isEmpty(info.getSubChannelName())) {
+            displayBothUploaderAndSubChannel(info);
+            subChannelThumb.setVisibility(View.VISIBLE);
+        } else if (!TextUtils.isEmpty(info.getUploaderName())) {
+            displayUploaderAsSubChannel(info);
         } else {
             uploaderTextView.setVisibility(View.GONE);
+            uploaderThumb.setVisibility(View.GONE);
```
Stypox, line 1165: Shouldn't you also hide subChannelThumb? just to be sure that it is gone, because I think it could happen (i.e. when restoring instance) that `handleResult` is called twice and with different results, thus possibly resulting in wrongly showing the thumbnail left behind from the previous call. Not sure though if this is really an issue, but this is already the implemented behaviour with uploaderThumb

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
```diff
@@ -425,6 +474,17 @@ public void handleResult(@NonNull final ChannelInfo result) {
...
 
+        if (!TextUtils.isEmpty(currentInfo.getSubChannelName())) {
+            headerSubChannelTitleView.setText(String.format(
+                            getString(R.string.channel_created_by),
+                            currentInfo.getSubChannelName())
+            );
+            headerSubChannelTitleView.setVisibility(View.VISIBLE);
+            headerSubChannelAvatarView.setVisibility(View.VISIBLE);
+        } else {
+            headerSubChannelTitleView.setVisibility(View.GONE);
```
Stypox, line 485: Also hide headerSubChannelAvatarView

wb9688: I didn't do that, as it's gone by default in the .xml, so I thought that's not needed.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
wb9688: We open the sub-channel when it exists on click. On long-click we want to open the parent channel, but only if it has a sub-channel, as you could just click when it doesn't have a sub-channel to open the parent channel.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
wb9688: I didn't do that, as it's gone by default in the .xml, so I thought that's not needed.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
Stypox: Oh, I see, sorry

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
Stypox: Mmmh ok, it not the best to have hardcoded default values for things that could possibly be used multiple times, but maybe we can ignore it since if we were to make videodetailfragment reusable it would mean changing many things, so I guess this is just not important

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -463,6 +474,15 @@ public boolean onLongClick(final View v) {
             case R.id.detail_controls_download:
                 NavigationHelper.openDownloads(getActivity());
                 break;
+
+            case R.id.detail_uploader_root_layout:
+                if (TextUtils.isEmpty(currentInfo.getSubChannelUrl())) {
+                    Log.w(TAG,
```
TobiGr, line 480: why is this a warning? Apart from that, this info should only be put in the log in debug apks

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
```diff
@@ -394,6 +410,36 @@ private void updateSubscribeButton(final boolean isSubscribed) {
...
+    public void onClick(final View v) {
+        if (isLoading.get() || currentInfo == null) {
+            return;
+        }
+
+        switch (v.getId()) {
+            case R.id.sub_channel_avatar_view:
+            case R.id.sub_channel_title_view:
+                if (TextUtils.isEmpty(currentInfo.getSubChannelUrl())) {
+                    Log.w(TAG, "Can't open sub-channel because we got no channel URL");
```
TobiGr, line 427: same here

Issue #3275, opened 2020-03-25
Make the mute button faster accessible for pop-up and background player


Poolitzer:
this is related to #3165.

### Pop-up Player:

Currently, to mute a video playing in pop-up, I need to scroll down my notifications, press on the NewPipe one, then it's in the upper corner. If the pop-up window is big enough though, there is space after the captions string. Someone could use that to implement the mute button.

<img src="https://user-images.githubusercontent.com/25934244/77593630-bd27f280-6eb1-11ea-8662-5fb0a97c33f6.jpg" alt="screenshot of the pop-up player" width="300px">

I hope the red arrow makes clear what I mean.

### Background Player:

Currently, to mute a video which is playing in the Background, I need to scroll down my notifications, press on the NewPipe one, then it's in the upper right corner. This process can be simplified if we add the mute button between the repeat and the back control directly in the notification.

<img src="https://user-images.githubusercontent.com/25934244/77594358-94a0f800-6eb3-11ea-9eee-6a6f42c3b5cc.png" alt="screenshot of the pop-up player" width="300px">

I hope the red arrow makes clear what I mean.

Stypox:
I don't think this should be done. In the popup player there are already too many icons, and when it is not full-sized it looks strange. In the background notification there is almost not enough space for the repeat button on phones with lower `dpi`s. Also, the background notification will be replaced soon.
@TobiGr what is your opinion? I don't think this should be implemented, also because the mute button is not used by so many people

Poolitzer:
Both of your issues can be countered with "only display the button if enough place is there". At least for the pop up window this works, not sure about the notification.

Stypox:
Yeah but there are more important buttons that could be fit instead

wb9688:
About putting that in the notification: not possible after #3178

B0pol:
> About putting that in the notification: not possible after #3178

no, that's false. It will even be better, because you can choose in settings which one to choose. Now there are more important buttons that could be in notifications, but if user can choose, there is no problem to add it via settings.

opusforlife2:
What's the use of having a mute button for background audio, anyway? The PR linked issue just asked for MX Player like functionality, which I can understand as it's for video. For example, you could watch a math problem being solved while another app is playing music. What about audio, though?

Poolitzer:
I am not sure, I just wanted to provide feedback for a way to easier access it :)

opusforlife2:
[...]

PR #3437, opened 2020-04-14
Fast rewind forward in background activity

### LINKED ISSUE #2722: Fast forward/rewind buttons in background mode
In background mode:
implementation of buttons or gestures that permits to go forward or backward of some seconds, like it's yet developed in standard mode tapping twice on left (-10") or right (+10") screen side.

As is implemented in podcasts apps where there are 2 main buttons:
+30 seconds
-10 seconds
That should be useful to resume easily some audio content that is lost during the listening.

Another helpful implementation could be a finest scrolling of timeline for long contents, putting time in manual mode or timebar in widescreen when mobile is tilted..

Thanks and btw... AMAZING APP!!

Screenshot_2019-10-11-11-21-22-324_com google android googlequicksearchbox

TheLastGimbus:
### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

- Add fast rewind / fast forward in BackgroundPlayerActivity (activity where you manage videos playing in background)

Honestly, 9 icons in one row is maybe too much? We could move shuffle and repeat upward. And we could remove (one of) playbackSpeedButton/playbackPitchButton, since they both do the same.

fixes #2722 

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
@@ -164,6 +164,9 @@ public abstract class ServicePlayerActivity extends AppCompatActivity
             case R.id.action_append_playlist:
                 appendAllToPlaylist();
                 return true;
+            case R.id.action_playback_speed:
+                openPlaybackParameterDialog();
+                return true;
             case R.id.action_mute:
                 player.onMuteUnmuteButtonClicked();
                 return true;
@@ -688,7 +691,10 @@ public abstract class ServicePlayerActivity extends AppCompatActivity
 
     private void onPlaybackParameterChanged(final PlaybackParameters parameters) {
         if (parameters != null) {
-            Log.d("Dupa", "12");
+            if (menu != null && player != null) {
+                MenuItem item = menu.findItem(R.id.action_playback_speed);
+                item.setTitle(formatSpeed(parameters.speed));
+            }
         }
     }
 
--- a/app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
@@ -45,7 +45,6 @@ import org.schabi.newpipe.util.ThemeHelper;
 import java.util.Collections;
 import java.util.List;
 
-import static org.schabi.newpipe.player.helper.PlayerHelper.formatPitch;
 import static org.schabi.newpipe.player.helper.PlayerHelper.formatSpeed;
 import static org.schabi.newpipe.util.Localization.assureCorrectAppLanguage;
 
@@ -90,9 +89,6 @@ public abstract class ServicePlayerActivity extends AppCompatActivity
     private ImageButton forwardButton;
     private ImageButton shuffleButton;
     private ProgressBar progressBar;
-
-    private TextView playbackSpeedButton;
-    private TextView playbackPitchButton;
 
     private Menu menu;
 
@@ -317,8 +313,6 @@ public abstract class ServicePlayerActivity extends AppCompatActivity
         fastForwardButton = rootView.findViewById(R.id.control_fast_forward);
         forwardButton = rootView.findViewById(R.id.control_forward);
         shuffleButton = rootView.findViewById(R.id.control_shuffle);
-        playbackSpeedButton = rootView.findViewById(R.id.control_playback_speed);
-        playbackPitchButton = rootView.findViewById(R.id.control_playback_pitch);
         progressBar = rootView.findViewById(R.id.control_progress_bar);
 
         repeatButton.setOnClickListener(this);
@@ -328,8 +322,6 @@ public abstract class ServicePlayerActivity extends AppCompatActivity
         fastForwardButton.setOnClickListener(this);
         forwardButton.setOnClickListener(this);
         shuffleButton.setOnClickListener(this);
-        playbackSpeedButton.setOnClickListener(this);
-        playbackPitchButton.setOnClickListener(this);
     }
 
     private void buildItemPopupMenu(final PlayQueueItem item, final View view) {
@@ -489,10 +481,6 @@ public abstract class ServicePlayerActivity extends AppCompatActivity
             player.onPlayNext();
         } else if (view.getId() == shuffleButton.getId()) {
             player.onShuffleClicked();
-        } else if (view.getId() == playbackSpeedButton.getId()) {
-            openPlaybackParameterDialog();
-        } else if (view.getId() == playbackPitchButton.getId()) {
-            openPlaybackParameterDialog();
         } else if (view.getId() == metadata.getId()) {
             scrollToSelected();
         } else if (view.getId() == progressLiveSync.getId()) {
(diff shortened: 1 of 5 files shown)


Stypox:
Only the playbackSpeedButton should be kept, to stay consistent with players.
Maybe you could add only the fast-forward button and put it where playbackPitchButton is currently. This would not create many problems imho, since fast-forwarding is more used than fast-rewinding, and the latter option can be obtained anyway by moving back the progress bar by a bit and then fast-forwarding until the correct point is reached, thus not wasting much time.

TheLastGimbus:
I would totally use fast rewind all the time. Like when you listen to some long talk, and missed what the guy just said, and you just want to repeat that - not mess with scroll bar. As for fast forward - I would almost never use that.

Listening in background is either for music or some talks - and you totally need fast rewind for talks.

I would lift pich/speed control up - you are probably not going to click that many times.

Stypox:
> I would totally use fast rewind all the time. 

I see, sorry for my wrong assumptions ;-)

> I would lift pich/speed control up - you are probably not going to click that many times.

What about moving it to the app bar at the top, where there is also the mute button?

TheLastGimbus:
Progress bar now updates when you press fast forward/rewind and it's paused.
Sorry but android xml layout gives me a headache :/ (in term of moving speed control on right of video title, where it's most intuitive)
But I can try to move it to app bar as you said, but I don't think it will look well/intuitive, and the app bar's title won't fit.
I think we could replace mute button itself. I don't see any use for it in background player - don't know why it was added in first place (?).

Stypox:
Sorry for the late response. Apparently the mute button is present in the background player because some people find it useful: #3275 https://github.com/TeamNewPipe/NewPipe/issues/2876#issuecomment-575153729
But since it pretty much is a niche feature, I would be ok moving it to the three-dot menu (or showing it `ifRoom`). Then its place could be taken by the `1x` button, which is not that big (i.e. at maximum 5 characters: `1.11x`). And the current place of the `100%` and `1x` would be taken by fast forward and rewind.

RickyM7:
@Stypox In my opinion, as long as it keeps working as it is now, I don't care if the button is moved, the important thing is to work well.

Stypox:
@TheLastGimbus could you implement the requested changes, if you have time? ;-)

TheLastGimbus:
Sorry, some other stuffed poped up (I've fell into home automation rabbit hole :P )
Thanks for reminding, I'll try to do it today <3

TheLastGimbus:
Okay soo... that's it.
I'm a little worried about the mute button now - as it is hidden in three dots menu - if someone presses it by accident, and then is like "why the hell video is not playing?"
But NewPipe users should be techy enough to find out :P
What do you think?

app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
```diff
@@ -690,8 +691,10 @@ private void onPlayModeChanged(final int repeatMode, final boolean shuffled) {
 
     private void onPlaybackParameterChanged(final PlaybackParameters parameters) {
         if (parameters != null) {
-            playbackSpeedButton.setText(formatSpeed(parameters.speed));
-            playbackPitchButton.setText(formatPitch(parameters.pitch));
+            if (menu != null && player != null) {
+                MenuItem item = menu.findItem(R.id.action_playback_speed);
```
wb9688, line 695: Please make this variable `final`

TheLastGimbus: done

TheLastGimbus:
It does crash...
```
org.schabi.newpipe.report.ErrorActivity: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.widget.ImageButton.setOnClickListener(android.view.View$OnClickListener)' on a null object reference
        at org.schabi.newpipe.player.ServicePlayerActivity.buildControls(ServicePlayerActivity.java:323)
```
It's a null object on `fastRewindButton.setOnClickListener(this);`
But fastRewindButton exitst... just like other buttons... or does it?

wb9688:
@TheLastGimbus: No, it doesn't, as you didn't make the appropriate changes to the landscape version of the layout: https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/res/layout-land/activity_player_queue_control.xml

app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
wb9688: Instead of saying "done", please click on "Resolve conversation" next time ;)

TheLastGimbus:
Whoopsie, sorry. Now it works :)

TheLastGimbus:
Sorry that all of this took so long. All of those Java `findViewById's` and `setOnClickListener's` made the code somewhat confusing to me as a begginer on this project.
Do you maybe plan to migrate it to Kotlin? It would make the code soo much cleaner and easier to maintain.
@Stypox said in #2714:
> I am in favour of the introduction of Kotlin, since it is faster to program in and the transition from Java is smooth

and I totally agree

wb9688:
>Sorry that all of this took so long. All of those Java findViewById's and setOnClickListener's made the code somewhat confusing to me as a begginer on this project.

That's just Android's API. I don't see how Kotlin would help with that. We're not actively converting the code to Kotlin, but we accept new code written in Kotlin in NewPipe. We're going to fully move to view binding (or data binding when view binding isn't enough for that fragment) though, see #3697. Also, NewPipe's code is currently just pretty crappy.

>I've fell into home automation rabbit hole :P

What are you using for that? I'm using Home Assistant and have it connected to my Hue lights, some Sonoff S20 switches, my UniFi APs for presence detection, and some other small stuff.

TheLastGimbus:
> That's just Android's API. I don't see how Kotlin would help with that

Oh, a lot! Kotlin pretty much does that for you. You don't need to do anything like
```java
private ImageButton fastRewindButton;
fastRewindButton = rootView.findViewById(R.id.control_fast_rewind);
fastRewindButton.setOnClickListener(this);
...
// some global method onClick
@Override
public void onClick(final View view) {
    switch(view.getId()) // need to check which is it
}
...
fastRewindButton.setText("Something");
```
you just do
```kotlin
fastRewindButton.text = "Something"
fastRewindButton.setOnClickLister {
    // stuff
}
```
and it's done.
But I've also heard that view bindings are recommended now and kida better.
Also no problems with null checking, null exceptions etc. Kotlin really saves a lot of time, space and readability.

> What are you using for that? I'm using Home Assistant 

Yass, HASS is *amazing*. I use ESP8266's flashed with ESPHome. It enables me to connect anything in my home, with any sensor I want, and then HASS allows me to bind them with each other to do crazy stuff, almost like my home is alive.
I can have my plants watered when a new person gets coronavirus.
It's mindblowing

TheLastGimbus:
I just discovered that when I use some other way to click backward/forward - that is, buttons on headphones, or volume buttons on LineageOS - it activates next/previous video instead of fast forward/rewind.
It is counter intuitive, because the notification has only "fast buttons", so it suggests that this is the _main option_ or something.

Spotify in "podcast mode" has those external buttons connected to forward/rewind - could we do the same, or this is intentional?

wb9688:
>You don't need to do anything like

Neither do you on Java:

```java
button.setOnClickListener((View v) -> {
    v.doWhatever();
});
```

That's also much clearer to me than:

```kt
button.setOnClickListener {
    it.doWhatever()
}
```

View binding also isn't specific to Kotlin, but works on Java as well. Null is also not really a problem if you properly annotate everything.

Btw I also use some ESP8266s flashed with ESPHome for e.g. temperature, though I have some Hue lights as well.

Issue #3448, opened 2020-04-17
Ability to share the stream URL externally


jadkik:
#### Describe the feature you want

It would be nice if we could easily share the stream URL (i.e. the actual media URL extracted from YouTube) to some external source. A NewPipe developer on Reddit told me something similar is done when you enable the external video player.

#### Is your feature request related to a problem? Please describe it

My use case is to be able to share (instead of the [VIEW](https://github.com/TeamNewPipe/NewPipe/blob/65cd9751d87044ffaa96a27eed945098cefcf546/app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java#L227) intent) the stream URL to be able to watch it on Kodi (via the Kore app).

#### Additional context

I am aware of the Play on Kodi setting but it just shares the youtube URL AFAIK and that requires the Kodi YouTube plugin which often fails because of the whack-a-mole with YouTube API ToS.

#### How will you/everyone benefit from this feature?

I imagine an additional menu item like "open in browser" could benefit not just this use case specifically. It could be useful to share it to get the media link on a PC (e.g. sharing via KDE Connect) or send a watchable link to other people. In a sense the idea would be similar to what [the Web Media Share app on F-Droid](https://f-droid.org/en/packages/com.tobykurien.webmediashare/) does.

jadkik:
I made this quick draft PR #3449 (still work in progress) to check if it would work, and it did to an extent. Some problems:

  * the video name does not appear when shared to Kodi (not sure if there's some EXTRA we can specify)
  * maybe a better way would be to add a setting that would change the behavior of the Cast to Kodi menu item?
  * the item I added would need an icon (or set to always show in the "...") (and translated I guess)

B0pol:
Little off-topic, but still related

> the Kodi YouTube plugin which often fails because of the whack-a-mole with YouTube API ToS.

You might be interested in this fresh new plugin https://github.com/TheAssassin/kodi-invidious-plugin and the todo list

> be able to open YouTube videos from NewPipe

jadkik:
@B0pol thanks for the info, I'll keep an eye on it and maybe try it out.

Re: the benefit of the feature; aside from being useful to share the URL to Kodi (via Kore), the URL can be shared to "Play with VLC" which means it can be used effectively as a one-off "play with external player" (instead of changing the setting).

jadkik:
> the video name does not appear when shared to Kodi (not sure if there's some EXTRA we can specify)

[...]

PR #3449, opened 2020-04-17
Add ability to share stream URL externally

jadkik:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

- add additional menu item in video detail page to share the URL that would be played in external player

- Implements feature requested in #3448 

#### Testing apk

~~[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4493521/debug.zip)~~

[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4604987/debug.zip) (updated 2020-05-10)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -941,7 +941,8 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
     private void openBackgroundPlayer(final boolean append) {
         final AudioStream audioStream = getDefaultAudioStream();
 
-        final boolean useExternalAudioPlayer = PreferenceManager.getDefaultSharedPreferences(activity)
+        final boolean useExternalAudioPlayer = PreferenceManager
+                .getDefaultSharedPreferences(activity)
                 .getBoolean(activity.getString(R.string.use_external_audio_player_key), false);
 
         if (!useExternalAudioPlayer && android.os.Build.VERSION.SDK_INT >= 16) {
@@ -1034,8 +1035,9 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
      * @return AudioStream instance according to user settings
      */
     private AudioStream getDefaultAudioStream() {
-        final int streamIndex = ListHelper.getDefaultAudioFormat(activity, currentInfo.getAudioStreams());
-        return currentInfo.getAudioStreams().get(streamIndex);
+        final List<AudioStream> audioStreams = currentInfo.getAudioStreams();
+        final int streamIndex = ListHelper.getDefaultAudioFormat(activity, audioStreams);
+        return audioStreams.get(streamIndex);
     }
 
     private void prepareDescription(final Description description) {
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -696,7 +696,7 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
                 return true;
             case R.id.menu_item_share_stream:
                 if (currentInfo != null) {
-                    Stream stream;
+                    final Stream stream;
                     if (currentInfo.getVideoStreams().isEmpty()
                             && currentInfo.getVideoOnlyStreams().isEmpty()) {
                         stream = getDefaultAudioStream();
@@ -939,9 +939,9 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
     //////////////////////////////////////////////////////////////////////////*/
 
     private void openBackgroundPlayer(final boolean append) {
-        AudioStream audioStream = getDefaultAudioStream();
+        final AudioStream audioStream = getDefaultAudioStream();
 
-        boolean useExternalAudioPlayer = PreferenceManager.getDefaultSharedPreferences(activity)
+        final boolean useExternalAudioPlayer = PreferenceManager.getDefaultSharedPreferences(activity)
                 .getBoolean(activity.getString(R.string.use_external_audio_player_key), false);
 
         if (!useExternalAudioPlayer && android.os.Build.VERSION.SDK_INT >= 16) {
@@ -1033,9 +1033,8 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
      *
      * @return AudioStream instance according to user settings
      */
-    @Nullable
     private AudioStream getDefaultAudioStream() {
-        int streamIndex = ListHelper.getDefaultAudioFormat(activity, currentInfo.getAudioStreams());
+        final int streamIndex = ListHelper.getDefaultAudioFormat(activity, currentInfo.getAudioStreams());
         return currentInfo.getAudioStreams().get(streamIndex);
     }
 
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -696,11 +696,14 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
                 return true;
             case R.id.menu_item_share_stream:
                 if (currentInfo != null) {
-                    VideoStream selectedVideoStream = getSelectedVideoStream();
-                    if (selectedVideoStream != null) {
-                        ShareUtils.shareUrl(requireContext(), currentInfo.getName(),
-                                selectedVideoStream.getUrl());
+                    Stream stream;
+                    if (currentInfo.getVideoStreams().isEmpty()
+                            && currentInfo.getVideoOnlyStreams().isEmpty()) {
+                        stream = getSelectedAudioStream();
+                    } else {
+                        stream = getSelectedVideoStream();
                     }
+                    ShareUtils.shareUrl(requireContext(), currentInfo.getName(), stream.getUrl());
                 }
                 return true;
             case R.id.menu_item_openInBrowser:
@@ -936,8 +939,7 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
     //////////////////////////////////////////////////////////////////////////*/
 
     private void openBackgroundPlayer(final boolean append) {
-        AudioStream audioStream = currentInfo.getAudioStreams()
-                .get(ListHelper.getDefaultAudioFormat(activity, currentInfo.getAudioStreams()));
+        AudioStream audioStream = getSelectedAudioStream();
 
         boolean useExternalAudioPlayer = PreferenceManager.getDefaultSharedPreferences(activity)
                 .getBoolean(activity.getString(R.string.use_external_audio_player_key), false);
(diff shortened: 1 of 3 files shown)


jadkik:
You are right, I'm not very familiar with the code base yet.

I also moved and renamed the menu item and placed it below "Open in browser".

How do you handle translation? Should I add a placeholder entry for the menu item in all `strings.xml` files or is it managed separately?

wb9688:
>How do you handle translation? Should I add a placeholder entry for the menu item in all strings.xml files or is it managed separately?

You should only touch the `values` one, like you currently do. The rest will be handled by Weblate.

jadkik:
Sorry, marked this as ready for review, but I just noticed it still breaks with SoundCloud items because `getSelectedVideoStream()` fails.

Will look into it later to try and find a proper fix.

jadkik:
Updated to work with audio and video streams. Tested it with some Soundcloud links, YouTube links and MediaCCC link.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -964,6 +975,12 @@ private VideoStream getSelectedVideoStream() {
         return sortedVideoStreams != null ? sortedVideoStreams.get(selectedVideoStreamIndex) : null;
     }
 
+    @Nullable
+    private AudioStream getSelectedAudioStream() {
```
Stypox, line 979: Rename this to a something like `getDefaultAudioStream()`, the current name is misleading, as this method does not really pick what the user selected. Maybe add a comment about the fact that it is going to be used in association with `getSelectedVideoStream()`.

app/src/main/res/values/strings.xml
```diff
@@ -11,6 +11,7 @@
     <string name="open_in_browser">Open in browser</string>
     <string name="open_in_popup_mode">Open in popup mode</string>
     <string name="share">Share</string>
+    <string name="share_stream">Share link to stream file</string>
```
Stypox, line 14: I don't like this string much. What about "Share stream file link" or "Share stream link"? @B0pol @wb9688 @opusforlife2 do you have any other idea?

opusforlife2: My vote goes to Share Video Link. I'm assuming this doesn't work for background and popup.

jadkik: I picked this name to be consistent with `controls_download_desc` (`Download stream file`). Looking into it, that string is only used for accessibility so it may not even make sense to follow the same naming scheme.

In any case, I agree that the string is not ideal. Personally, I would go with `Share stream link`.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
jadkik: I pushed a change to this. Renamed and added a javadoc, let me know if it makes more sense.

(btw I also rebased to latest `dev`)

app/src/main/res/values/strings.xml
jadkik: @opusforlife2 It works for audio streams too (e.g. soundcloud)

opusforlife2: What about youtube audio stream?

jadkik: I haven't checked against youtube audio streams. If you can provide such a link I could try it.

I will also check it with YouTube livestreams (haven't checked those yet).

opusforlife2: No idea how Newpipe gets those. One of the devs could help you there.

I'm thinking that there should be some way to choose between sharing the video stream or audio stream to send to the external player.

Stypox: I think @opusforlife2 is talking about YouTube audio streams (those played with the background player). Maybe two buttons should be added, one to share the video stream and another to share the audio stream? I am not sure though, since it would mean adding two buttons used only by a small group of people, so do not implement this yet, I'll ask for some opinions internally.
Btw, I am ok with `Share stream link`

Stypox: For now, let's only add the ability to share the video stream link (and the audio one for SoundCloud, i.e. the current behaviour), to be consistent with the fact that there is no way to open an audio stream link with the external player. So keep the current changes and just change the value of the string to `Share stream link`

jadkik: Ah got it. I somehow thought YouTube had audio-only streams. In the meantime, I tried playing some YouTube live streams, and the link is shared correctly but it won't play in VLC. I guess that's a problem with the external player rather than NewPipe itself.

As for choosing between audio-only vs. video, I guess it could be made into a submenu, similar to the one used to pick the stream quality in the action bar. IMO it's overkill unless there is demand for this feature. I don't mind implementing that if/when it's needed.

The string should be updated now.

B0pol: what about share stream URL?

Stypox: I also thought about "url", but then I realized it is more of a technical word, and "link" is more used by the average non-tech-savy person. For example, Google uses "link" in Meet invites, and other people I know use that term, too (this could be biased by the fact that it is used in the italian language, though). So I'd keep "link".

B0pol: You are right. Though URL is often used in newpipe strings.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -694,6 +694,18 @@ public boolean onOptionsItemSelected(final MenuItem item) {
                             currentInfo.getOriginalUrl());
                 }
                 return true;
+            case R.id.menu_item_share_stream:
+                if (currentInfo != null) {
+                    Stream stream;
```
wb9688, line 699: I think this could be `final`

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1014,6 +1025,20 @@ private VideoStream getSelectedVideoStream() {
...
+     * Get the stream to play when the current stream is an audio-only stream.
+     *
+     * This is the audio-only equivalent of getSelectedVideoStream,
+     * without the ability for the user to select a custom stream quality.
+     *
+     * @return AudioStream instance according to user settings
+     */
+    @Nullable
+    private AudioStream getDefaultAudioStream() {
+        int streamIndex = ListHelper.getDefaultAudioFormat(activity, currentInfo.getAudioStreams());
```
wb9688, line 1038: This could be `final`

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -927,8 +939,7 @@ private boolean shouldShowComments() {
     //////////////////////////////////////////////////////////////////////////*/
 
     private void openBackgroundPlayer(final boolean append) {
-        AudioStream audioStream = currentInfo.getAudioStreams()
-                .get(ListHelper.getDefaultAudioFormat(activity, currentInfo.getAudioStreams()));
+        AudioStream audioStream = getDefaultAudioStream();
```
wb9688, line 942: This and 2 lines below could be `final`

wb9688:
@Stypox: The reason I said URL is that I think we use URL instead of link throughout the app

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
B0pol: Nope, its reference is changed below. Here the reference is null, and after it will be either poiting on an object return by `getDefaultAudioStream()` or `getSelectedVideoStream()`

wb9688: @B0pol: No, it isn't assigned here yet, and it's afterwards only assigned once: in `if` or in `else`. (With e.g. `try`-`catch` it would've been different.)

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
jadkik: I updated it. This line and the next one were not final in the original code though.

opusforlife2:
Well, there is also the chance that people confuse stream to mean live stream, because that's the usual connotation.

I suggest "Share media URL" to be unambiguous about what the option does.

I'd prefer "Share video/audio URL" based on whether it's SoundCloud or one of the others, though. It's extra code, but worth it, if it can be implemented.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
jadkik: Yeah it is never assigned to null, it's only assigned in two different branches.

I also removed the `@Nullable` annotation on `getDefaultAudioStream()` because it cannot be null. If there's no audio stream or no appropriate format, it would fail by calling `get(-1)` eventually and throwing an index out of bounds exception.

And while theoretically `getSelectedVideoStream()` could return null, given that we only call it inside this conditional, it would never happen. But the IDE still complains 🤷 . An assertion would silence it... I don't know if it's worth it.

jadkik:
> I'd prefer "Share video/audio URL" based on whether it's SoundCloud or one of the others, though. It's extra code, but worth it, if it can be implemented.

I assume it wouldn't be hard to have both `Share video URL`/`Share audio URL` and hide one (or show both where applicable). It looks like it would be a few simple changes in `setupActionBar` or `updateMenuItemVisibility`?

B0pol:
How can people know the resolution is audio free? With the external player setting there is that mute button, but no such thing here

Stypox:
> How can people know the resolution is audio free?

The selector shows whether the resolution has audio or not

B0pol:
Only if you've set external player in settings, yes that's what I was saying. Otherwise people won't know

litetex:
No progress since a year, no GitHub actions build and merge conflicts.

jadkik:
@litetex I'm willing to resolve the conflicts and get this merged but I need feedback from the reviewers about how to proceed.

app/src/main/res/values/strings.xml
ShareASmile: > No idea how Newpipe gets those. One of the devs could help you there.
> 
> I'm thinking that there should be some way to choose between sharing the video stream or audio stream to send to the external player.

I think @ AudricV can shed some light on that

dandare1980:
> @litetex I'm willing to resolve the conflicts and get this merged but I need feedback from the reviewers about how to proceed.

@jadkik would you consider reopening this pull request?  There's still interest in seeing this feature in NewPipe.

ShareASmile:
@dandare1980 Let member of the TeamNewPipe decide about it as rewrite of the app is on the cards.

> would you consider reopening this

Only a Team member or a maintainer having rights to do that can reopen issues/PR's if closed by a member/maintainer.

radasbona:
> > but are you still interested in this feature?

Oh yes! @opusforlife2 
I have still a Raspberry Pi 1B at my living room TV. Latest possible Kodi Version is 18, no available YT or Inidious Plugins any more.
Only way to watch YT is to go into the settings and switch external player on. I wait a long time for a share stream url button. The play with kodi button is useless for me unfortunately.

jadkik:
> @jadkik would you consider reopening this pull request?

> @jadkik I apologise for your PR getting ignored. It's been 2 years, but are you still interested in this feature?

I haven't used my Kodi/RPi setup in a few months, so I don't immediately need this myself anymore.

> If so, maybe it would be better to wait for the rewrite, because it will be sure to contain massive changes to the player code.

I haven't kept up with the developments here - so I wouldn't know. I'll try to rebase this and see how messy it gets.

> @jadkik How would this PR handle the separate A/V streams btw ? Caz external players can only grab a single stream source.

Good point, I do not know. I think I had made it pick the first option or some kind of preferred option based on existing code. I was building on top of what the "play in external player" did, so if that didn't work with some videos it would also fail when sharing that stream URL.

boernsen-development:
This feature would be awesome to have. A real pity that it is closed!

Issue #3038, opened 2020-01-30
Standardize view animation duration 


Stypox:
If you check the usages of `animateView` you will see that the typical range will be 0 all the way to 500, with most in the 100-250 range.

It should've been done with a default value in the first place. It is pretty much arbitrary right now, it was whatever the person felt like at the time of writing.

I'll leave this task to another PR, which the goal will be to standardize the used duration of animations.

_Originally posted by @mauriciocolli in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MzQxNTczMDc1/pull_request_reviews/more_threads_

ludugz:
Hi @Stypox ,
I check the source code and see that most values are from 0-500 as you mentioned above. 
Before fixing this, I want to make sure that is it OK if I change all the values from 0-500 to 250 (average)? 
In the case of value bigger than 500 ( f.e 800), should I just leave it there?

wb9688:
I vote for 420 ms :rofl:

If you want to find _all_ magic numbers in our code, you could uncomment https://github.com/TeamNewPipe/NewPipe/blob/dev/checkstyle.xml#L148. We should reduce the amount of magic numbers in general.

Stypox:
> Hi @Stypox ,
> I check the source code and see that most values are from 0-500 as you mentioned above.
> Before fixing this, I want to make sure that is it OK if I change all the values from 0-500 to 250 (average)?
> In the case of value bigger than 500 ( f.e 800), should I just leave it there?

250 seems good to me :-D
Anyway, if that value doesn't turn out to be good it will be changeable with only one edit, since it would be a constant. So go for 250ms ;-)

ludugz:
I also vote for 250ms :D, probably have the issue fixed this weekend.

Novmbrain:
Hello @TobiGr ! We are back
We already set up the environment needed and launch Newpipe in our machine and we have watched the last @ludugz's works. 
Firstly, by searching key word 'animate(' in the project. We found all magic numbers passed to this (But we are not sure about it). 

Secondly, wee found that ludugz defined two constants in AnimationsUtils
```
import static org.schabi.newpipe.util.AnimationUtils.DEFAULT_LONG_ANIM_DURATION;
import static org.schabi.newpipe.util.AnimationUtils.DEFAULT_SHORT_ANIM_DURATION;
```
[...]

PR #3458, opened 2020-04-19
Standardize view animation duration #3038

ludugz:
#### What is it?
- [ ] Bug fix (user facing)
- [ ] Feature (user facing)
- [X] Code base improvement (dev facing)

- Standardize view animation for 3 methods: animateView, showControls, hideControls
- Replace VideoPlayer/DEFAULT_CONTROLS_DURATION with AnimationUtils/DEFAULT_SHORT_ANIM_DURATION ( For 0-500 ms in previous version)
- Create constant DEFAULT_LONG_ANIM_DURATION (for 800ms & 1500ms in previous version)
- Refactor code format some places

- #3038 

#### Testing apk

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4503018/app-debug.zip)

#### Agreement


B0pol:
Pausing feels so slow now. I don't think it should be that slow.

app/src/main/java/org/schabi/newpipe/streams/Mp4FromDashWriter.java
```diff
@@ -640,19 +640,20 @@ private int makeFtyp() throws IOException {
         return size;
     }
 
-    private byte[] makeMdat(long refSize, final boolean is64) {
+    private byte[] makeMdat(final long refSize, final boolean is64) {
```
wb9688, line 643: Why even change this file? That's unrelated to your PR and not needed. Plus if you're gonna fix a suppressed error, at least remove it from the suppressions.

app/src/main/java/org/schabi/newpipe/streams/Mp4FromDashWriter.java
```diff
@@ -711,7 +712,8 @@ private int makeMoov(final int[] defaultMediaTime, final TablesInfo[] tablesInfo
 
         for (int i = 0; i < tracks.length; i++) {
             if (tracks[i].trak.tkhd.matrix.length != 36) {
-                throw new RuntimeException("bad track matrix length (expected 36) in track n°" + i);
+                throw new RuntimeException("bad track matrix length (expected 36) in"
+                        + " track n°" + i);
```
wb9688, line 716: Why split it into 2 lines? That line isn't even too long…

XiangRongLin: My guess is that his build also fails like mine, because his system also sees it as 101 characters in a single line (because of the °, being recognized as 2 characters).

wb9688: @XiangRongLin: Then you have weird encoding settings and should fix that, not this line. That also still doesn't explain the other changes in this file.

ludugz: @XiangRongLin  nailed it. Maybe something wrong with my Android Studio, I will return it back soon.

app/src/main/java/org/schabi/newpipe/streams/Mp4FromDashWriter.java
ludugz: My build could not succeed and the log prints out that it should be changed to "final" variable because of the CheckStyle. That's why I fixed it.

app/src/main/java/org/schabi/newpipe/streams/Mp4FromDashWriter.java
wb9688: @ludugz: Could you also try whether #3460 fixes this?

app/src/main/java/org/schabi/newpipe/streams/Mp4FromDashWriter.java
wb9688: @ludugz: That's actually mpossible afaik, since line 643 is suppressed, as you could see [here](https://github.com/TeamNewPipe/NewPipe/blob/dev/checkstyle-suppressions.xml#L19).

ludugz: @wb9688 Yeah you are right. I don't know why but this time I build without any problem related to this, I will put this file back as before.

app/src/main/java/org/schabi/newpipe/streams/Mp4FromDashWriter.java
ludugz: @wb9688 It worked. Thank you!

PR #3460, opened 2020-04-19
Force UTF-8 encoding for Gradle

wb9688:
@XiangRongLin: Could you test whether this works for your Checkstyle issue?


XiangRongLin:
Yeah, that fixes it

Issue #3419, opened 2020-04-11
Search suggestions


wb9688:
#### Describe the feature you want
Show search suggestions, i.e. "Did you mean: …".

#### Additional context
NewPipeExtractor already has the relevant code to extract it, but apparently that's never called in the app.

B0pol:
It requires extractor changes. Currently, the provided method doesn't work for youtube or youtube music (and for other services it returns `null`)

wb9688:
@B0pol: What's https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeSearchExtractor.java#L66 and https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/extractors/YoutubeMusicSearchExtractor.java#L130 then? I'm pretty sure they both work.

B0pol:
I found there are in fact two way to have suggestion. For example, test on youtube music "duo lipa", it will show you "dua lipa". If you try with Newpipe extractor and getSearchSuggestion, it will be empty.

PR #3471, opened 2020-04-22
Display search suggestion: did you mean & showing result for

Royosef:
#### What is it?
- [ ] Bug fix (user facing)
- [X] Feature (user facing)

Display search suggestions: "did you mean" and "showing result for" (in YouTube & YouTube Music).

<table><tr><td></td><td></td></tr></table>

- https://github.com/TeamNewPipe/NewPipe/issues/3419

- https://github.com/TeamNewPipe/NewPipeExtractor/pull/311

#### Testing apk

[NewPipe_gisplaysearchsuggestion-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4517507/NewPipe_gisplaysearchsuggestion-debug.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
@@ -119,7 +119,7 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
     String lastSearchedString;
 
     @State
-    String searchSuggestionString;
+    String searchSuggestion;
 
     @State
     boolean isCorrectedSearch;
@@ -266,9 +266,7 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
             }
         }
 
-        if (!TextUtils.isEmpty(searchSuggestionString)) {
-            handleSearchSuggestion(searchSuggestionString, isCorrectedSearch);
-        }
+        handleSearchSuggestion();
 
         if (suggestionDisposable == null || suggestionDisposable.isDisposed()) {
             initSuggestionObserver();
@@ -981,10 +979,11 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
                     NewPipe.getNameOfService(serviceId), searchString, 0);
         }
 
-        handleSearchSuggestion(result.getSearchSuggestion(), result.isCorrectedSearch());
-
-        searchSuggestionString = result.getSearchSuggestion();
+        searchSuggestion = result.getSearchSuggestion();
         isCorrectedSearch = result.isCorrectedSearch();
+
+        handleSearchSuggestion();
+
         lastSearchedString = searchString;
         nextPageUrl = result.getNextPageUrl();
         currentPageUrl = result.getUrl();
@@ -1002,10 +1001,9 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
         super.handleResult(result);
     }
 
-    private void handleSearchSuggestion(@NonNull final String searchSuggestion,
-                                        @NonNull final Boolean isCorrected) {
+    private void handleSearchSuggestion() {
         if (!TextUtils.isEmpty(searchSuggestion)) {
-            String helperText = getString(isCorrected
+            final String helperText = getString(isCorrectedSearch
                     ? R.string.search_showing_result_for
                     : R.string.did_you_mean);
 
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
@@ -149,8 +149,7 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
     private EditText searchEditText;
     private View searchClear;
 
-    private View correctSuggestionPanel;
-    private TextView correctSuggestionText;
+    private TextView correctSuggestion;
 
     private View suggestionsPanel;
     private RecyclerView suggestionsRecyclerView;
@@ -357,8 +356,7 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
         searchEditText = searchToolbarContainer.findViewById(R.id.toolbar_search_edit_text);
         searchClear = searchToolbarContainer.findViewById(R.id.toolbar_search_clear);
 
-        correctSuggestionPanel = rootView.findViewById(R.id.correct_suggestion_panel);
-        correctSuggestionText = rootView.findViewById(R.id.correct_suggestion_text);
+        correctSuggestion = rootView.findViewById(R.id.correct_suggestion);
     }
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -511,7 +509,7 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
                 return;
             }
 
-            correctSuggestionPanel.setVisibility(View.GONE);
+            correctSuggestion.setVisibility(View.GONE);
 
             searchEditText.setText("");
             suggestionListAdapter.setItems(new ArrayList<>());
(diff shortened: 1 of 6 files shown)


wb9688:
You haven't changed `app/build.gradle` to have `implementation 'com.github.B0pol:NewPipeExtractor:621be5ba3b7d72e486111555a9a0e4400c8beedd'`.

app/src/main/res/layout/fragment_search.xml
```diff
@@ -6,13 +6,46 @@
...
+            android:id="@+id/correct_suggestion_text"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center_vertical"
+            android:layout_marginStart="5dp"
+            android:layout_weight="1"
+            android:textColor="@color/background_title_color"
+            android:textSize="@dimen/search_suggestion_text_size"
+            android:textStyle="bold|italic"
+            tools:text="z101bara" />
```
wb9688, line 38: Why? We pretty much never use `tools:text`, and "Did you mean: z101bara" is of course bs, as "z101bara" is the incorrect search query, which it would obviously never suggest.

app/src/main/res/values/dimens.xml
```diff
@@ -112,4 +112,7 @@
     <dimen name="feed_group_carousel_top_bottom_margin">2dp</dimen>
     <dimen name="feed_group_carousel_between_items_margin">4dp</dimen>
 
+    <dimen name="search_suggestion_text_size">16sp</dimen>
+    <dimen name="search_suggestion_helper_text_size">14sp</dimen>
```
wb9688, line 116: What did you base the text size on?

app/src/main/res/values/strings.xml
```diff
@@ -646,4 +646,6 @@
     <string name="feed_use_dedicated_fetch_method_disable_button">Disable fast mode</string>
     <string name="feed_use_dedicated_fetch_method_help_text">Do you think feed loading is too slow? If so, try enabling fast loading (you can change it in settings or by pressing the button below).\n\nNewPipe offers two feed loading strategies:\n• Fetching the whole subscription channel, which is slow but complete.\n• Using a dedicated service endpoint, which is fast but usually not complete.\n\nThe difference between the two is that the fast one usually lacks some information, like the item\'s duration or type (can\'t distinguish between live videos and normal ones) and it may return less items.\n\nYouTube is an example of a service that offers this fast method with its RSS feed.\n\nSo the choice boils down to what you prefer: speed or precise information.</string>
     <string name="content_not_supported">This content is not yet supported by NewPipe.\n\nIt will hopefully be supported in a future version.</string>
+    <string name="search_showing_result_for">Showing results for</string>
+    <string name="search_did_you_mean">Did you mean:</string>
```
B0pol, line 650: you shouldn't do like that, format using `%s` or `%1$s`: some languages could be like: 
**newpipe**: did you mean, or 
**newpipe** showing results for.

wb9688: That's true indeed, however how will you do the differently styled suggestion then?

Royosef:
> You haven't changed `app/build.gradle` to have `implementation 'com.github.B0pol:NewPipeExtractor:621be5ba3b7d72e486111555a9a0e4400c8beedd'`.

Oh, I did. It's just I wasn't sure I'm supposed to include it in the PR, I will add it.

app/src/main/res/layout/fragment_search.xml
Royosef: Haha, I know it's BS, it was written for the sake of preview in the Android Studio designer. Would you like me to remove it?

app/src/main/res/values/dimens.xml
Royosef: Nothing specific, I tried some numbers until it was looking good.

app/src/main/res/values/strings.xml
Royosef: Yeah, but I think this would happen specifically in languages that are RTL (like Hebrew or Arabic), and in this case I assume the whole layout is mirrored to RTL (and at least for Hebrew and Arabic, I know this is the way it's supposed to be).

app/src/main/res/layout/fragment_search.xml
wb9688: If you're going to put something there, do something that's not BS such as "newpipe"

app/src/main/res/values/strings.xml
B0pol: @wb9688 
1. Using Markwon, a markdown library we already use, and surround the word in with `**` (bold in markdown), like: `Showing results for **%1$s**`. See VideoDetailFragment#prepareDescription() for an usage. It's the easiest solution.
2. Using spannable, see https://stackoverflow.com/questions/20936901/combining-spannable-with-string-format . But it seems harder, and not better.

@Royosef It's false, go in Settings -> Content -> App language, pick Arabic, the second language (actually the first after system default), you'll see nothing is changed except text. As it's two different textviews, it won't work as you think, and that's why you need to format.

By the way
```suggestion
```
edit the already existing did_you_mean string, it's unused.

wb9688: @B0pol: Not really a good idea, as you could have service return weird characters that would make the formatting weird. In this case I'd probably choose for just having e.g. `Did you mean "%s"?` and `Showing results for "%s"` with normal formatting.

Royosef: So, @ B0pol, are you cool with a plain string format?

B0pol: I don't see how like that could happen. But yes

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -257,6 +266,10 @@ public void onResume() {
             }
         }
 
+        if (!TextUtils.isEmpty(searchSuggestionString)) {
+            handleSearchSuggestion(searchSuggestionString, isCorrectedSearch);
```
B0pol, line 270: handleSearchSuggestion already checks `if (!TextUtils.isEmpty(searchSuggestionString))`

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -961,6 +981,10 @@ public void handleResult(@NonNull final SearchInfo result) {
                     NewPipe.getNameOfService(serviceId), searchString, 0);
         }
 
+        handleSearchSuggestion(result.getSearchSuggestion(), result.isCorrectedSearch());
+
+        searchSuggestionString = result.getSearchSuggestion();
+        isCorrectedSearch = result.isCorrectedSearch();
```
B0pol, line 987: why do you save it *after* calling handleSearchSuggestion()? here you call result.getSearchSuggestion() and result.isCorrectedSearch() twice.

B0pol:
I rebased and used a switch preference instead of "follow … (theme 1)"  "follow … (theme 2)"

Stypox:
@B0pol I think you picked the wrong PR ;-D
What you meant was #3632, wasn't it?

B0pol:
Yes 🤦‍♂️

app/src/main/res/layout/fragment_search.xml
```diff
@@ -6,13 +6,34 @@
...
+        <TextView
+            android:id="@+id/correct_suggestion_text"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_marginStart="10dp"
+            android:layout_marginTop="10dp"
+            android:layout_marginBottom="10dp"
+            android:layout_weight="1"
+            android:textColor="@color/background_title_color"
+            android:textSize="@dimen/search_suggestion_text_size" />
```
Stypox, line 26:
```suggestion
            android:textSize="@dimen/search_suggestion_text_size"
            tools:text="Showing results for lorem ipsum dolor sit amet consectetur adipisci elit" />
```

app/src/main/res/layout/fragment_search.xml
```diff
@@ -6,13 +6,34 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
+    <LinearLayout
```
Stypox, line 9: Why nesting inside a linearlayout?

Stypox:
I rebased, applied the suggestions above, and changed commit names to a better format.
I added a selectable item background to the suggestion panel, and I also made it long clickable: on a long click the current search query is replaced with the suggested one and the keyboard is opened, but the search is not performed, to let the user edit the query before searching.

Test apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4874124/app-debug.zip)

wb9688:
>I added a selectable item background to the suggestion panel

@Stypox: What do you mean?

>on a long click the current search query is replaced with the suggested one and the keyboard is opened

It doesn't open the keyboard for me. It also would be nice if it'd have the cursor at the end of the query instead of the start.

Stypox:
I rebased again and fixed the problems reported by @wb9688 

> What do you mean?

The background that makes bubbles around touches: `?attr:selectableItemBackground`

Another testing apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4875485/app-debug.zip)

wb9688:
@Stypox: Ah, you mean ripple?

TobiGr:
Looks good to me. Btw. is there a reason, why the bold and italic font effects were removed from the search terms?

B0pol:
@TobiGr yes, read this thread: https://github.com/TeamNewPipe/NewPipe/pull/3471#discussion_r413152629

TobiGr:
@B0pol Thank you. acccc6e0ad5e7d6f29f0f12c933eb8ed1613ea3e should solve the problem. Or do you expect something else to be displayed?

PR #3271, opened 2020-03-25
Use vector drawables instead of PNGs for material icons

Stypox:
### What is it?
- [ ] Bug fix
- [ ] Feature
- [x] Enhancement under the hood

### Long description of the changes in your PR
For all manually-created images PNG have been kept. Debug apk size decreased by 232,030 bytes.
- Rename all icon attrs to have a **`ic_` prefix**
- Always use **`_24dp` icons**, because there is no real difference, since they are vector drawables
- Always use the **original name found on material.io for icon** drawables, as to not create confusion and possibly duplicates. Icon names can still be different from real drawable names, though I have made some of them compliant to this or maybe more meaningul.
- Use standard icons for **`expand_more` and `expand_less`** instead of triangles
- Use `play_button_outline` instead of custom PNG as **play button in VideoDetailFragment** (questionable, as there is no shadow anymore)
- **Replace the search-insert-icon** (on the button that is right to every search suggestion, and when pressed replaces what is in the search field with the value of that suggestion), that was an arrow pointing up-left and didn't have a meaning, with a custom hand-made icon (`ic_search_add`)
- Keep `ic_close` and `ic_replay` as PNGs: needed for notifications in RemoteViews

Other changes:
- Remove duplicate `getIconByAttr()` in ThemeHelper (use `resolveResourceIdFromAttr()` instead)
- Improve the **VideoDetailFragment layout** by a little by forcing the "expand description" icon to always stay in the same place even after expanding. I tested the new layout on various screens and it always looked ok.
- **Update AndroidX** to `1.1.0` to fix icon crashes in preferences on API 19: use `androidx.preference:preference` instead of `androidx.legacy:legacy-preference-v14`
- **Remove legacy libraries** (`androidx.legacy:legacy-support-v4`): this made no effect, and allowed me not to use `androidx.appcompat.widget.*` everywhere, tested on various Android versions
- Fix `SelectChannelFragment` and `SelectKioskFragment` (those used to set main screen tabs): in API 19 they had a **duplicate title** always empty.
- Fix pause icon used instead of play when seeking in a paused popup player

### Fixes the following issue(s)
- Fixes #1199

### Testing apk
[...]

PR #3482, opened 2020-04-25
Fix toolbar icon and text color in white theme

### LINKED ISSUE #3042: Polish white theme: make app bar text and icons white
This is how the white theme looks like:

As you can see text and icons in the title bar are black, even though they are on the exact same red background as in the dark theme. Black on red looks strange, so text and icons should be white, as in the Retry button

IAmSSH:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

- Create a new theme `ToolbarTheme` in `styles.xml` to set it on the toolbar to get the desired color and icons.
- Modify `Tab.java` and `KioskTranslator.java` to always return icons having white color.

- Fixes #3042 

### Testing apk
[NewPipe.zip](https://github.com/TeamNewPipe/NewPipe/files/4534541/NewPipe.zip)

#### Agreement


IAmSSH:
@Stypox 
Request your review please.
Thanks.

app/src/main/res/values/styles.xml
```diff
@@ -115,6 +115,97 @@
         <item name="preferenceTheme">@style/PreferenceThemeOverlay.v14.Material</item>
     </style>
 
+    <!-- DrawerArrowStyle is used to change the color of hamburger-icon to white -->
+    <style name="DrawerArrowStyle" parent="Widget.AppCompat.DrawerArrowToggle">
+        <item name="color">@android:color/white</item>
+    </style>
+
+    <style name="ToolbarTheme" parent="ThemeOverlay.AppCompat.ActionBar">
```
Stypox, line 123: Isn't there a way to inherit all of the icon definitions from the dark theme, so that they don't have to be specified two times?

IAmSSH:
I know its not good practice, I tried doing that, but there's a problem.
If we inherit `DarkTheme`, we essentially also inherit its parent the `Theme.AppCompat.NoActionBar`. There are some styles in `Theme.AppCompat.NoActionBar` that make the text in the spinner that is used to select the quality of a video (in video detail fragment) appear black along with its background, so it becomes unreadable. 
I tried overriding some styles but it didn't work out.
I'll try some more and let you know.

IAmSSH:
I changed the spinner items to always be of red colour, irrespective of the theme selected.

Stypox:
I just realized one thing: maybe there is no need to create a new style in `styles.xml` just for icons, but it is enough to force the white drawables (instead of using `?attr/`) in all menus, since there is only one global style for toolbars. Anyway, the new code looks better, and if my above assumption can't be implemented it is good to go after some testing ;-)

IAmSSH:
I think the hamburger icon's colour cannot be changed without using the `DrawerArrowStyle` style, since it does not belong to our set of style icons (defined in Dark and Light themes), but is preset. So, it might be necessary to keep the style.

Stypox:
Ok, then this is correct approach.

I found some issues (tested on an emulator with Android 4.4 API 19):
- The background color of the toolbar does not change according to the service (i.e. try with soundcloud). It should have the same background color as the tab selector in the main page.
- The background color of the stream selector in the video detail activity should not be red, but should follow the chosen app theme (i.e. dark if dark theme, white if light theme, etc.), just like three-dot menus do.
- The color of the `______` line under the currently selected tab in the main page should always be white, but instead it changes color when changing theme.
- In download activity the color of the "List" toolbar button is black, while it should be white.
- The toolbar of the about activity still has black icons and text when using white theme.

IAmSSH:
Okay, I'll look into it

IAmSSH:
Regarding the last comment:
>The background color of the stream selector in the video detail activity should not be red, but should follow the chosen app theme (i.e. dark if dark theme, white if light theme, etc.), just like three-dot menus do.

You mean 


The menu to select stream quality is red but it should be like the colour of dot menu popup, i.e 


The dot menu popup is always of the colour as shoown in the image irrespective of the stream or the theme, so should I make it similar to that (gray background and white text), or you meant something else?

Stypox:
Oh, that's another issue. Both menus should have dark background with the dark theme and light background with the light theme. I guess you need to do something similar to what you did with `DrawerArrowStyle`

IAmSSH:
The app is working fine now. All issues are resolved.
The only thing remaining is the colour of the spinner in video detail fragment. It does not change according to Light and Dark themes, but is consistent with the colours of different services.
I think it appears fine like this. If you want it to be changed do let me know.

Stypox:
- The "save playlist" toolbar button in the playlist fragment is black in white theme, while it should always be white
- The download dialog (opened from the video detail fragment) has two issues: the stream selector has a red background color, while its background should follow the theme, and the back button (top left) is black in white theme, while it should always be white
- The toolbar in background/popup player activities still has black icons and title in white theme, while they should be white

These two issues you posted screenshots above are still not solved:
- The three-dot menu has a dark background independently of the theme, while it should follow the theme
- The stream selector in video detail fragment has a red background, while it should follow the theme

IAmSSH:
There were some problems with the current approach. Since the Toolbar had the DarkTheme as the parent, so the styles of DarkTheme were getting applied to it statically and they were not changing with the Themes or the Services and it created problems everywhere.

So I had to revert back to the earlier approach that involved copying icons in the ToolbarTheme, it solved all the problems.

IAmSSH:
Hey,
I removed the dead code.
Is there anything else that is remaining?

wb9688:
@IAmSSH: #3271 changed the names of some icons, so you have to change your PR accordingly.

PS: Please stay longer on IRC until we've seen your question and answered, or use e.g. Matrix.

app/src/main/res/values/styles.xml
```diff
@@ -115,6 +115,107 @@
...
+        <item name="drawerArrowStyle">@style/DrawerArrowStyle</item>
+        <item name="colorControlNormal">@color/white</item>
+
+        <item name="thumbs_up">@drawable/ic_thumb_up_white_24dp</item>
```
Stypox, line 127: Now the build is failing because some icon attrs were renamed (e.g. thumbs_up is now ic_thumbs_up). So please copy-paste this list of attributes again

IAmSSH: Done. All checks passing.
Thanks

IAmSSH:
The required changes are done.

PS: @wb9688 Sorry about the IRC. I was really having a hard time figuring out how to work with it. I have set up Matrix now.

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -138,7 +138,15 @@ protected void onCreate(final Bundle savedInstanceState) {
...
         }
 
-        setSupportActionBar(findViewById(R.id.toolbar));
+        Toolbar toolbar = findViewById(R.id.toolbar);
+
+        toolbar.setPopupTheme(
+                ThemeHelper.isLightThemeSelected(this)
+                ? R.style.PopupThemeLight
+                : R.style.PopupThemeDark
+        );
```
Stypox: Why not use `R.style.ToolbarTheme` here, too?

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -743,6 +743,9 @@ private void setupActionBar(final StreamInfo info) {
             public void onItemSelected(final AdapterView<?> parent, final View view,
                                        final int position, final long id) {
                 selectedVideoStreamIndex = position;
+                RelativeLayout r = (RelativeLayout) parent.getChildAt(0);
+                ((TextView) r.getChildAt(1)).setTextColor(getResources().getColor(R.color.white));
+                ((TextView) r.getChildAt(2)).setTextColor(getResources().getColor(R.color.white));
```
Stypox, line 748: Isn't there any other way to do this? It seems to be a workaround to me, though if there is no other way then ok, I guess the toolbar inflated layout is always the same.

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -471,12 +470,13 @@ private void updateBookmarkButtons() {
...
-                ? R.attr.ic_playlist_add : R.attr.ic_playlist_check;
+                ? R.drawable.ic_playlist_add_white_24dp
+                : R.drawable.ic_playlist_add_check_white_24dp;
 
         final int titleRes = playlistEntity == null
                 ? R.string.bookmark_playlist : R.string.unbookmark_playlist;
 
-        playlistBookmarkButton.setIcon(ThemeHelper.resolveResourceIdFromAttr(activity, iconAttr));
+        playlistBookmarkButton.setIcon(iconAttr);
         playlistBookmarkButton.setTitle(titleRes);
```
Stypox, line 480: While you are at it, you could move the variables above directly here:
```suggestion
        playlistBookmarkButton.setIcon(playlistEntity == null
                ? R.drawable.ic_playlist_add_white_24dp
                : R.drawable.ic_playlist_add_check_white_24dp);
        playlistBookmarkButton.setTitle(playlistEntity == null
                ? R.string.bookmark_playlist
                : R.string.unbookmark_playlist);
```

app/src/main/java/us/shandian/giga/ui/fragment/MissionsFragment.java
```diff
@@ -224,9 +224,15 @@ private void updateList() {
...
-                    : ThemeHelper.resolveResourceIdFromAttr(requireContext(), R.attr.ic_list));
+            boolean isLight = ThemeHelper.isLightThemeSelected(mContext);
+            int icon;
+
+            if (mLinear)
+                icon = R.drawable.ic_grid_white_24dp;
+            else
+                icon = R.drawable.ic_list_white_24dp;
+
+            mSwitch.setIcon(icon);
```
Stypox: Here, too. Also, `isLight` is unused.
```suggestion
            mSwitch.setIcon(mLinear
                    ? R.drawable.ic_grid_white_24dp
                    : R.drawable.ic_list_white_24dp);
```

app/src/main/res/layout/error_retry.xml
```diff
@@ -27,6 +27,7 @@
         android:textAlignment="center"
         android:textAllCaps="true"
         android:textAppearance="@style/TextAppearance.AppCompat.Body1"
+        android:background="?attr/colorPrimary"
```
Stypox, line 30: Why was this added? I don't think it is needed after @B0pol introduced `ServiceColoredButton`

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
IAmSSH: I couldn't find any direct way of styling just the first element of a spinner, so had to use this approach.

IAmSSH:
Done @Stypox

wb9688:
Should we move this to v0.20.0 seeing how @IAmSSH hasn't replied for 25 days and it not being a bug fix (so not really worth it fixing it ourself imho)?

TiA4f8R:
Closing in favor of #5927.

PR #3501, opened 2020-04-29
Open in browser button now really opens in browser

B0pol:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

- Open in browser buttons open in default browser for android 6+ (otherwise no change).

- Open in browser button not opening in browser, but either showing a chooser, or worse: opening in NewPipe if NewPipe is set as default for youtube links. I just wanna open in browser

It will only work for android 6+ since it uses default apps, available from settings -> apps -> more -> default apps -> browser app, [a new feature in android 6](https://www.tomsguide.com/us/change-default-apps-in-android,review-3309.html), otherwise it will not change. It will also not change if you have no browser set as default. It's a rare case, because if you have only one browser that does not happen, if you have two, as soon as you choose one browser as default for opening a link, it will be set default in that selector.
Therefore there are only two such cases: fresh new browser install, or not any link set as default.

I've tested with Firefox, Firefox preview and a chromium based browser and had no troubles with this code (I first had a problem with Firefox browsers but changed the way to send the intent, following [playWithKore()](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java#L602) and now it works).

#### Testing apk
[NewPipe_openInBrowser.zip](https://github.com/TeamNewPipe/NewPipe/files/4551024/NewPipe_openInBrowser.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
@@ -16,36 +16,33 @@ public final class ShareUtils {
      * Open the url with the system default browser.
      * <p>
      * If no browser is set as default, fallbacks to
-     * {@link ShareUtils#openUrlBrowserChooser(Context, String)}
+     * {@link ShareUtils#openInDefaultApp(Context, String)}
      *
      * @param context the context to use
      * @param url     the url to browse
      */
     public static void openUrlInBrowser(final Context context, final String url) {
-        final Intent intent = new Intent(Intent.ACTION_VIEW);
-        intent.setData(Uri.parse(url));
-
         final String defaultBrowserPackageName = getDefaultBrowserPackageName(context);
 
         if (defaultBrowserPackageName.equals("android")) {
             // no browser set as default
-            openUrlBrowserChooser(context, url);
+            openInDefaultApp(context, url);
         } else {
+            final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
             intent.setPackage(defaultBrowserPackageName);
             context.startActivity(intent);
         }
     }
 
     /**
-     * Open the url with application chooser, including browser and apps able to open url.
+     * Open the url in the default app set to open this type of link
      * <p>
-     * If any app (except browser, typically NewPipe) is set as default,
-     * it will nor open in browser, neither open the chooser, but just the default app.
+     * If no app is set as default, it will open a chooser
      *
      * @param context the context to use
      * @param url     the url to browse
      */
-    private static void openUrlBrowserChooser(final Context context, final String url) {
+    private static void openInDefaultApp(final Context context, final String url) {
         final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
         context.startActivity(Intent.createChooser(
                 intent, context.getString(R.string.share_dialog_title)));
--- a/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
@@ -2,19 +2,77 @@ package org.schabi.newpipe.util;
 
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.content.pm.ResolveInfo;
 import android.net.Uri;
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -2,19 +2,77 @@
...
+     * If any app (except browser, typically NewPipe) is set as default,
+     * it will nor open in browser, neither open the chooser, but just the default app.
+     *
+     * @param context the context to use
+     * @param url     the url to browse
+     */
+    private static void openUrlBrowserChooser(final Context context, final String url) {
+        final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
         context.startActivity(Intent.createChooser(
                 intent, context.getString(R.string.share_dialog_title)));
```
Stypox, line 51: Maybe the title should be changed to something like "Open in browser"

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -2,19 +2,77 @@
...
+        final Intent intent = new Intent(Intent.ACTION_VIEW);
+        intent.setData(Uri.parse(url));
+
+        final String defaultBrowserPackageName = getDefaultBrowserPackageName(context);
+
+        if (defaultBrowserPackageName.equals("android")) {
+            // no browser set as default
+            openUrlBrowserChooser(context, url);
+        } else {
+            intent.setPackage(defaultBrowserPackageName);
```
Stypox, line 34: Move the intent instantiation code inside the else branch, as it is not needed otherwise

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -2,19 +2,77 @@
...
+     * If no browser is set as default, fallbacks to
+     * {@link ShareUtils#openUrlBrowserChooser(Context, String)}
+     *
+     * @param context the context to use
+     * @param url     the url to browse
+     */
     public static void openUrlInBrowser(final Context context, final String url) {
-        Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
+        final Intent intent = new Intent(Intent.ACTION_VIEW);
+        intent.setData(Uri.parse(url));
```
Stypox, line 26:
```suggestion
        final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
```

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
B0pol: Nope, that's why i did this PR: it doesn't open in browser, but in favorite / default app for the type of link (in NewPipe case, "open in Browser" doesn't open browser but newpipe, so it's pointless) , and if there is no default app set for this type of link, it opens a chooser (with browser and NewPipe).

What about `openInDefaultApp()`?

Stypox: Oh right sorry ;-)
Yeah, `openInDefaultApp()` is maybe a better name

TobiGr:
@B0pol When merging PRs, please add them to the release notes.

Younes-L:
I dislike this change :(
This was useful to open videos in any app, such as VLC or YouTube app. YouTube website is pretty bad in comparison.
Any way to bypass the "newpipe is default app for YouTube" without limiting this to browser ?

B0pol:
The fix comes in version 0.19.4

Younes-L:
> The fix comes in version 0.19.4

I only originally commented thinking the behaviour change affects me, but after more testing it seemed to not affect my usecase. 

To find out what actually changed, I tried to reproduce the problem using 0.19.3 and compare the results with 0.19.4, but saw no difference.
Using 0.19.3, I was unable to have newpipe open videos with the "open in browser" button.
(Could not reproduce this issue : "worse: opening in NewPipe if NewPipe is set as default for youtube links.")
I set newpipe 0.19.3 as default for www.youtube , m.youtube, youtu.be , and channel links.
Excuse my confusing comment.

Issue #3312, opened 2020-03-30
Main page tabs aren't what they should be after import before restart


PeterHindes:
Imported data is only updated  after restart of NewPipe.
After importing data from an older version with a different tab arrangement, the taps will not contain the screen they should. Haven't tested with data from the current version. Fix is simple. Close the app and relaunch after importing. The app can do this its self to prevent user jank.

Different tab arrangement:
![image](https://user-images.githubusercontent.com/19994487/77962663-3ef49300-7299-11ea-8206-37b1ff8be467.png)

mauriciocolli:
I'm not able to replicate this, can you post more detailed steps to reproduce it [like the bug report template indicates](https://raw.githubusercontent.com/TeamNewPipe/NewPipe/5f549a8fc688157dfea8329408311e52e7e8c88f/.github/ISSUE_TEMPLATE/bug_report.md)?

PeterHindes:
Sorry about that.
Steps
  1. Install older version of newpipe
  2. Use settings to change which tabs show on top
  3. Export settings from old version of newpipe
  4. Install newest newpipe.
  5. import settings
  6. return to the home screen and see that they are not what the icon says they are. Unitll
  7. Restart the app
  8. They behave normaly now.

gkeegan:
I have noticed odd behavior with the import as well when changing the tabs. Can confirm this is (or at least was, but I don't see commits that say it was fixed) a bug.

PeterHindes:
I tested it on a build of the code from the dev branch

PeterHindes:
Should be a simple fix to restart the app on import of settings.

mauriciocolli:
@PeterHindes

You can see in this [absolute bad practice of a class](https://github.com/TeamNewPipe/NewPipe/blob/5f549a8fc688157dfea8329408311e52e7e8c88f/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java#L47) that the app is, indeed, [restarted](https://github.com/TeamNewPipe/NewPipe/blob/5f549a8fc688157dfea8329408311e52e7e8c88f/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java#L254-L269). In my case, the app is restarted and all the tabs are what they should be, in various Android versions as well.

[...]

PR #3509, opened 2020-05-01
Upgrade some dependencies

wb9688:
#### What is it?
- [ ] Bug fix (user facing)
- [ ] Feature (user facing)

#### Agreement


Stypox:
Have you tested on 4.4?

wb9688:
@Stypox: Not thoroughly, but I was going to.

wb9688:
Reading licenses is broken

app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.java
```diff
@@ -26,8 +28,8 @@ public LicenseFragmentHelper(@Nullable final Activity activity) {
         weakReference = new WeakReference<>(activity);
     }
 
-    private static String getFinishString(final Activity activity) {
-        return activity.getApplicationContext().getResources().getString(R.string.finish);
+    private static String getFinishString(final Context context) {
+        return context.getString(R.string.finish);
```
Stypox, line 32: Nothing to do with dependency updates, but what's the purpose of this function? `context.getString(R.string.finish)` is already explicative, readable and not long, so why encapsulate it in a function?

wb9688: I was wondering that as well… I'll fix it later today

TobiGr:
> is there a protocol that can be followed to fully test such app-wide changes?

@Stypox Test _everything_ on KitKat, Lollipop, latest and a version (maybe Oreo) in between. With so many upgraded dependencies almost everything is somehow affected by this change.

wb9688:
@TobiGr: I think he meant more like what is _"everything"_? NewPipe has so many features that you could easily forget one.

app/build.gradle
```diff
@@ -79,22 +79,23 @@ android {
...
+    icepickVersion = '3.2.0'
+    checkstyleVersion = '8.32'
+    stethoVersion = '1.5.1'
+    leakCanaryVersion = '2.2'
+    exoPlayerVersion = '2.10.8'
+    androidxLegacyVersion = '1.0.0'
+    androidxLifecycleVersion = '2.2.0'
+    androidxRoomVersion = '2.2.5'
+    groupieVersion = '2.8.0'
+    markwonVersion = '4.3.1'
```
B0pol, line 91: Why did you change the order? And if we change the order, it should be for lexicographical order.

wb9688: I changed the order to be the same as down there. Down there I changed the order in a way that seems more logical to me.

Stypox:
Btw I also tested on my 7.0 phone and everything works fine

B0pol:
Tested on 9.0 phone and Android TV 7.1.2 and I haven't found an issue.

wb9688:
I tested it on AVD API 19 and on my phone with Android 10. I'll rebase and do what @Stypox suggested.

app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.java
```diff
@@ -26,28 +28,24 @@ public LicenseFragmentHelper(@Nullable final Activity activity) {
...
-    public static String getFormattedLicense(final Context context, final License license) {
+    private static String getFormattedLicense(final Context context, final License license) {
         if (context == null) {
-            throw new NullPointerException("context is null");
+            throw new IllegalArgumentException("context is null");
         }
         if (license == null) {
-            throw new NullPointerException("license is null");
+            throw new IllegalArgumentException("license is null");
         }
```
Stypox, line 43: This is also unrelated to dependency updates, but this block of code should be replaced by `@NotNull` parameters, or just removed alltoghether, because a NullPointerException would be thrown anyway crashing the app, so it makes no sense to check for nullity and then do not do anything different.

TobiGr:
On KitKat main page tabs are not loaded again when new tabs are added through the settings.
E.g. I added the history tab and moved it to the top of the list. Instead of seeing the history when going back, I see trending content.

app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.java
wb9688: That's why I made it an `IllegalArgumentException`, though making it `@NotNull` is probably better indeed

wb9688:
@TobiGr: I just rebased, but I can't reproduce your bug.

Stypox:
@TobiGr I think this has to do with #3312 and is unrelated from this PR.

app/src/debug/java/org/schabi/newpipe/DebugApp.kt
```diff
@@ -0,0 +1,58 @@
...
+    override fun attachBaseContext(base: Context) {
+        super.attachBaseContext(base)
+        MultiDex.install(this)
+    }
+
+    override fun onCreate() {
+        super.onCreate()
+        initStetho()
+
+        AppWatcher.config = AppWatcher.config.copy(watchDurationMillis = 10000)
```
TobiGr, line 23: Please (re-)add the comment about the delay. The new naming is less understandable.

app/build.gradle
```diff
@@ -163,10 +163,10 @@ dependencies {
         exclude module: 'support-annotations'
     }
 
-    implementation 'com.github.TeamNewPipe:NewPipeExtractor:f3913e241e379adf0091319091e8f895c5fcfd07'
+    implementation 'com.github.wb9688:NewPipeExtractor:a28cf8a26cf1dce8b46b1be049b00265aa971189'
```
TobiGr, line 166: We should not forget to change this line once the extractor PR is merged.

PR #2806, opened 2019-11-16
Android TV support

Alexander--:
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Hi guys, it is me again.

This PR improves upon the previous one in many areas.

Long video descriptions were made properly scrollable (and actually convenient to read!) on TV devices.

NewPipeRecyclerView works much better. The new version is so good, it can actually be used as drop-in replacement for RecyclerView by itself — the separate patched LayoutManager is no longer necessary!

I have identified a bunch of bugs in FocusFinder on various Android versions. Those bugs are no big deal, but they make focus transitions in lists unpredictable if list items have different size (to be precise, when _focusable areas_ have different size). I have implemented a fix in SuperScrollLayoutManager. This is not the same as previous situation, when LayoutManager was necessary to avoid focus loss during scroll — you can now use NewPipeRecuclerView without SuperScrollLayoutManager, having later just improves experience further.

@TobiGr I have tested on Android 9 emulator and found out, why it was hanging for you — it was trying to rotate into portrait mode and apparently failing to do so.

In my experience, real Android TV devices have patched framework classes, that make switching to portrait mode utterly impossible (both automatically and manually). Android emulator does not contain those fixes, so when running on it the player still tried to enter portrait mode... This is not really problem with NewPipe or my changes, — just emulator doing poor job at emulating Android TV boxes. But I still added a change to use portrait mode for player by default when Android TV device is detected.

In general I don't advise to test this on Android 9 — it has several mis-features, that make NewPipe a pain to use. I will send fixes for those later. For now please test on Android 7 and older versions (most TV boxes don't use Android 9 anyway).

supplants #2701 

fixes #1363 fixes #899 fixes #717 fixes #3428

PR #3537, opened 2020-05-05
Always hide controls if video is playing [Android TV]

### LINKED ISSUE #3403: The player controls don't hide when using external controls
### Version

0.19.2

1. Play video with a bluetooth speaker connected
2. Press "pause" on the bluetooth speaker
3. Press "play" on the bluetooth speaker
4. The video starts playing again

### Expected behavior

The player controls should hide after a few seconds
### Actual behaviour

The player controls remain on screen and disappear only when it is tapped. This means that the play/pause controls on the bluetooth speaker become useless.

Stypox:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

In #2806 a new `safeHideControls` function was added, that would hide controls only if they were opened using a touch gesture. This is obviously never the case when using a tv, so after  `safeHideControls` was used (i.e. after every click on the interface), nothing would happen, preventing controls from ever being hidden. I assume this was added because on a tv it takes some time to press the correct buttons, thus having the ui close while selecting things would have been annoying, so now I made it so that `safeHideControls` hides controls while the video is playing, too, and otherwise the old behaviour is preserved. Also see https://github.com/TeamNewPipe/NewPipe/pull/2806#discussion_r420373214

- fixes #3403

#### Testing apk
@Poomex @B0pol @Generator could you test if the usability on TV devices is preserved, even if the ui is hidden when the video is playing?
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4588183/app-debug.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
@@ -1098,7 +1098,7 @@ public final class MainVideoPlayer extends AppCompatActivity
             }
 
             View controlsRoot = getControlsRoot();
-            if (controlsRoot.isInTouchMode()) {
+            if (controlsRoot.isInTouchMode() || isPlaying()) {
                 getControlsVisibilityHandler().removeCallbacksAndMessages(null);
                 getControlsVisibilityHandler().postDelayed(() ->
                         animateView(controlsRoot, false, duration, 0,
--- a/app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
@@ -98,8 +98,8 @@ public abstract class VideoPlayer extends BasePlayer
     //////////////////////////////////////////////////////////////////////////*/
 
     public static final int DEFAULT_CONTROLS_DURATION = 300; // 300 millis
-    public static final int DEFAULT_CONTROLS_HIDE_TIME = 2000;  // 2 Seconds
-    public static final int DPAD_CONTROLS_HIDE_TIME = 7000;  // 7 Seconds
+    public static final int DEFAULT_CONTROLS_HIDE_TIME = 2000; // 2 seconds
+    public static final int DPAD_CONTROLS_HIDE_TIME = 5000; // 5 seconds
 
     protected static final int RENDERER_UNAVAILABLE = -1;
 


B0pol:
It's fixed for me. Though I've one concern: the speed is not the same as usual, i.e. [**SEVEN seconds**](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java#L102)
We definitely should standardize. But why it's so high on TV??? I would remove this and just let normal time on all other devices, which is 2 seconds (line just above).

Stypox:
d29e0aa1a7443c928f46b749da0692e45a2463ef is the commit which increased the delay, I guess for usability reasons (on some TVs maybe 2s is too short to be able to do anything?)

Stypox:
@B0pol I reduced to 5s the hide time. New debug apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4588183/app-debug.zip)

Alexander--:
@B0pol

> I would remove this and just let normal time on all other devices, which is 2 seconds

Sounds like you have lighting-fast reaction. Have you actually used NewPipe on TV?

Quickly looking over controls, reading all labels, and pressing the navigation buttons — all of that in under 2 seconds — is impressive feat (especially since infrared TV remotes have additional ~100 ms of lag due to limitatons of IR protocols).

@Stypox

Right now method `safeHideControls` is being used from onClick() only. Hiding controls in onClick() makes some sense on touch-based devices, because showing them again is cheap, and the user is probably done interacting with screen anyway. It does not make sense on TV devices, because re-summoning and navigating around them with dpad is a massive hassle.

On TV devices `safeHideControls` does nothing (unless they are in touch mode). It does not abort current timeout, if present. So if controls were going to be hidden, they will still get hidden.

If something (`showControls`/`showControlsThenHide`) does not hide controls properly on TV devices, that should be changed. Method `safeHideControls` currently works as designed — it should not hide controls any quicker just because user has clicked on one of them.

Alexander--:
@B0pol 

Note, that 7 second timeout isn't really expected to be triggered. You are meant to dismiss controls with BACK button after you are done with them.

I agree, that media buttons (PLAY,PAUSE,REWIND etc.) should have different treatment.

Stypox:
> Quickly looking over controls, reading all labels, and pressing the navigation buttons — all of that in under 2 seconds — is impressive feat (especially since infrared TV remotes have additional ~100 ms of lag due to limitatons of IR protocols).
Anyway, as soon as something is pressed, the player resets the countdown before closing

Alexander--:
@Stypox the point is — in some circumstances there is no cooldown. When controls are summoned by showControls (such as when playback is paused), they do not disappear. But if you click on something — for example, change aspect ratio — the controls quickly get hidden again. This makes no sense to me.

I believe, that controls should stay permanently summoned after playback is paused — to remind user that playback was indeed paused and not just hanged. Resuming from pause via media buttons — including the buttons on headset — should quickly hide controls, because headsets do not have BACK button to quickly dismiss them.

I have sent #3547 with those changes.

B0pol:
@Poomex please test the apk in the original post

Poomex:
Unfortunately, the issue hasn't been fixed for me. Pressing play on my bluetooth speaker keeps the interface visible until I tap the screen.

MD77MD:
>I believe, that controls should stay permanently summoned after playback is paused — to remind user that playback was indeed paused and not just hanged.

although sometimes showing controls is needed... I disagree with you in saying hiding them don't make sense, hiding player controls is also useful. sometimes a quick pause and play action is needed such as for reading a table or a chart, for comparing two pics or even for taking a screenshot because it keeps the controls out of the way.

my suggestion regarding showing controls after pause or hiding them is divided in two parts as fallow:

for phones:
A. when to show controls after pause: when pause is triggered by tapping the pause button, that is:
  1. while video is playing
  2. tap screen to show controls
  3. tap pause button
  4. video is paused while showing controls
B. when to hide controls after pause: when using double tap to pause gesture. that is:
  1. while video is playing
  2. double tap to pause
  3. player controls are are kept hidden
  4. perform action the pause was intended for 
      (look above for examples)
  5. double tap to play

For TVs:
A. when to show controls after pause: when pause is triggered by pressing the ok button on the remote
B. when to hide controls after pause: when pause is triggered by pressing the play/pause button on the remote

This unifies the double tap to seek and pause/play gestures behavior related to player controls showing or hiding.. it also gives newPipe a cleaner UI feeling for both TVs  and smartphones whenever a quick pause and play action is needed like what I explained above

Stypox:
@Alexander-- should I close this in favour of #3547?

Alexander--:
@Stypox

Yes, I suggest you do

PR #3544, opened 2020-05-06
Consider 503 response code a ReCaptcha challenge

### LINKED ISSUE #1119: no internet connection through a vpn
hi, using bitmask apk to connect through a vpn, no matter with which provider, i can't connect to youtube through newpipe/newpipe-beta...

Stypox:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

Another small pull request: in #1119 @robohobbit got a recaptcha, but the page response code was not 429, but rather 503. @mauriciocolli proposed to "check if the response code isn't successful and the redirected url contains `https://ipv4.google.com/sorry`", but I think it would be enough to just consider code 503 as a "Too many requests" indicator.

From wikipedia, it seems like what I said above is correct:
> **503 Service Unavailable**
> The server cannot handle the request (because it is overloaded or down for maintenance). Generally, this is a temporary state.

- fixes #1119

#### Testing apk
*None, since the issue is not reproducible*

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
@@ -208,7 +208,7 @@ public final class DownloaderImpl extends Downloader {
             final okhttp3.Response response = client.newCall(request).execute();
             final ResponseBody body = response.body();
 
-            if (response.code() == 429) {
+            if (response.code() == 429 || response.code() == 503) {
                 throw new ReCaptchaException("reCaptcha Challenge requested", siteUrl);
             }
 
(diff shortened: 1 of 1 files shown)


TobiGr:
Should we consider 503 as ReCaptcha challenge in all services or just when requesting YouTube?

Stypox:
In my opinion it should be considered a ReCaptcha in all services, because Wikipedia states that 503 are returned when the service is overloaded / flooded with requests, which is similar to 429.

wb9688:
@Stypox: No, as 503 is actually intended for when a website is down for maintanance or e.g. because of a DDOS attack. I think our current CAPTCHA implementation only works properly for YouTube though.

Stypox:
Oh ok, then this pr is useless since the reported issue is very old and probably now youtube does not raise 503 anymore. I will close both this and the issue

PR #3545, opened 2020-05-06
 Fix Kodi button showing up with unsupported services

### LINKED ISSUE #1166: soundcloud kore not found

If I try to stream soundcloud on kore the error (kore not found) will popup. When I stream/share youtube link with kore it works perfectly.

### LINKED ISSUE #1930: NewPipe fails to see that Kore (Kodi remote) is installed
While it has previously worked on this installation (same version, 0.14.2 from F-Droid), NewPipe now fails to register that Kore is installed, so it's not possible to queue videos Kodi.
I've tried reinstalling both NewPipe and Kore but even with Kore open in the background, it's not recognised as being there. It used to occasionally work but has now stopped, even if I restart my phone (Mi4C, running the latest MIUI). 

- [ yes] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [yes] I checked if the issue/feature exists in the latest version.
-

Stypox:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

This PR disables the share to Kodi button when the current service is not supported (see the Kore  [AndroidManifest.xml](https://github.com/xbmc/Kore/blob/master/app/src/main/AndroidManifest.xml)). Supported services are Youtube and Soundcloud. From that manifest I could also deduce that replacing `https` with `http` is not needed anymore.

- fixes #1166
- closes #1930

#### Testing apk
I tested if the urls correctly matched Kore's manifest, by installing the app (even without connecting it to anything) and sharing urls. Before this fix sharing MediaCCC urls would have generated an error (thus leading to a "Missing Kore app, install it?" dialog), so this method of testing is good. 
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4588833/app-debug.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -632,9 +632,10 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
     }
 
     private void updateMenuItemVisibility() {
-        // show kodi if set in settings
+        // show kodi button if it supports the current service and it is enabled in settings
         menu.findItem(R.id.action_play_with_kodi).setVisible(
-                PreferenceManager.getDefaultSharedPreferences(activity).getBoolean(
+                KoreUtil.isServiceSupportedByKore(serviceId)
+                && PreferenceManager.getDefaultSharedPreferences(activity).getBoolean(
                         activity.getString(R.string.show_play_with_kodi_key), false));
     }
 
@@ -665,8 +666,7 @@ public class VideoDetailFragment extends BaseStateFragment<StreamInfo>
                 return true;
             case R.id.action_play_with_kodi:
                 try {
-                    NavigationHelper.playWithKore(activity, Uri.parse(
-                            url.replace("https", "http")));
+                    NavigationHelper.playWithKore(activity, Uri.parse(currentInfo.getUrl()));
                 } catch (Exception e) {
                     if (DEBUG) {
                         Log.i(TAG, "Failed to start kore", e);
--- a/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
@@ -593,9 +593,6 @@ public final class MainVideoPlayer extends AppCompatActivity
 
             titleTextView.setSelected(true);
             channelTextView.setSelected(true);
-            boolean showKodiButton = PreferenceManager.getDefaultSharedPreferences(this.context)
-                    .getBoolean(this.context.getString(R.string.show_play_with_kodi_key), false);
-            kodiButton.setVisibility(showKodiButton ? View.VISIBLE : View.GONE);
 
             getRootView().setKeepScreenOn(true);
         }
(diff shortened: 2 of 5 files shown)


Stypox:
@TobiGr done, and tested again, by enqueueing videos from different services and going backward and forward in the main player.

TobiGr:
According to https://octodon.social/@dennix/104122037269553457 there is a kodi plugin which allows to open media.ccc videos with kodi.

Stypox:
@TobiGr is that an app with a standalone AndroidManifest.xml?

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -665,8 +666,7 @@ public boolean onOptionsItemSelected(final MenuItem item) {
                 return true;
             case R.id.action_play_with_kodi:
                 try {
-                    NavigationHelper.playWithKore(activity, Uri.parse(
-                            url.replace("https", "http")));
+                    NavigationHelper.playWithKore(activity, Uri.parse(currentInfo.getUrl()));
```
wb9688, line 669:
```suggestion
                    NavigationHelper.playWithKore(activity, Uri.parse(url));
```

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
```diff
@@ -712,6 +709,13 @@ public void onShuffleClicked() {
         protected void onMetadataChanged(@NonNull final MediaSourceTag tag) {
             super.onMetadataChanged(tag);
 
+            // show kodi button if it supports the current service and it is enabled in settings
+            boolean showKodiButton =
```
wb9688, line 713: `final`

wb9688:
@TobiGr: There indeed is a [Kodi plug-in](https://github.com/voc/plugin.video.media-ccc-de), however Kore doesn't seem to have support for it in [its AndroidManifest.xml](https://github.com/xbmc/Kore/blob/master/app/src/main/AndroidManifest.xml).

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
Stypox: I changed this because url is the originalUrl, that can be in a non-standard form, while getUrl() gets the standard url

app/src/main/java/org/schabi/newpipe/player/MainVideoPlayer.java
Stypox: Yes 😫😂

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
wb9688: Oh, you're right, I thought I had checked that…

TobiGr:
@wb9688 Thank you for checking that and opening the ticket!

We can add mediaCCC to `isServiceSupportedByKore()` once they changed their manifest. Looks good to merge :+1:

Stypox:
@wb9688 @TobiGr fixed

PR #3546, opened 2020-05-06
Allow choosing which types of search suggestions to show

### LINKED ISSUE #3540: Show search history even when search suggestions are disabled
### Version

- 0.19.3

1) enable Search history
2) disable Search suggestions
3) enter some search queries
4) check search history
5) search history is empty :(

### Expected behavior

store search queries locally

### Actual behaviour

search history is empty

Stypox:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

<table>
<tr>
<td></td>
<td>This PR replaces the switch preference to enable/disable all search suggestions with a multiselect one, so that the user can choose which types of search suggestions he likes: local, remote, both, none.
Since the previous preference stored a <code>boolean</code> while the new one stores a <code>Set&lt;String&gt;</code>, I had to create a migration function and call it in <code>NewPipeSettings.init()</code>. If the previous value was <code>true</code>, both local and remote suggestions are activated, otherwise they are both deactivated. I testes this and it works both when updating the app and when importing a database with the old setting.</td>
</tr>
</table>

- fixes #3540 

#### Testing apk
@anpic could you test if this option suits you?
[app-debug_1.zip](https://github.com/TeamNewPipe/NewPipe/files/4842389/app-debug_1.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
@@ -55,8 +55,8 @@ import org.schabi.newpipe.extractor.services.youtube.linkHandler.YoutubeSearchQu
 import org.schabi.newpipe.fragments.BackPressable;
 import org.schabi.newpipe.fragments.list.BaseListFragment;
 import org.schabi.newpipe.ktx.AnimationType;
-import org.schabi.newpipe.ktx.ExceptionUtils;
 import org.schabi.newpipe.local.history.HistoryRecordManager;
+import org.schabi.newpipe.settings.NewPipeSettings;
 import org.schabi.newpipe.util.Constants;
 import org.schabi.newpipe.util.DeviceUtils;
 import org.schabi.newpipe.util.ExtractorHelper;
@@ -65,16 +65,19 @@ import org.schabi.newpipe.util.ServiceHelper;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import icepick.State;
 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
-import io.reactivex.rxjava3.core.Flowable;
 import io.reactivex.rxjava3.core.Observable;
+import io.reactivex.rxjava3.core.Single;
 import io.reactivex.rxjava3.disposables.CompositeDisposable;
 import io.reactivex.rxjava3.disposables.Disposable;
 import io.reactivex.rxjava3.schedulers.Schedulers;
@@ -143,7 +146,8 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
     @Nullable private Map<Integer, String> menuItemToFilterName = null;
     private StreamingService service;
     private Page nextPage;
-    private boolean isSuggestionsEnabled = true;
+    private boolean showLocalSuggestions = true;
+    private boolean showRemoteSuggestions = true;
 
     private Disposable searchDisposable;
     private Disposable suggestionDisposable;
@@ -194,24 +198,12 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
     public void onAttach(@NonNull final Context context) {
         super.onAttach(context);
 
+        final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activity);
+        showLocalSuggestions = NewPipeSettings.showLocalSearchSuggestions(activity, prefs);
+        showRemoteSuggestions = NewPipeSettings.showRemoteSearchSuggestions(activity, prefs);
+
         suggestionListAdapter = new SuggestionListAdapter(activity);
-        final SharedPreferences preferences
-                = PreferenceManager.getDefaultSharedPreferences(activity);
-        final boolean isSearchHistoryEnabled = preferences
-                .getBoolean(getString(R.string.enable_search_history_key), true);
-        suggestionListAdapter.setShowSuggestionHistory(isSearchHistoryEnabled);
-
         historyRecordManager = new HistoryRecordManager(context);
-    }
-
-    @Override
-    public void onCreate(final Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-
-        final SharedPreferences preferences
-                = PreferenceManager.getDefaultSharedPreferences(activity);
-        isSuggestionsEnabled = preferences
-                .getBoolean(getString(R.string.show_search_suggestions_key), true);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


app/src/main/res/values/settings_keys.xml
```diff
@@ -158,7 +158,7 @@
     <string name="caption_user_set_key" translatable="false">caption_user_set_key</string>
 
     <!-- Content & History -->
-    <string name="show_search_suggestions_key" translatable="false">show_search_suggestions</string>
+    <string name="show_remote_search_suggestions_key" translatable="false">show_search_suggestions</string>
```
wb9688, line 161: Why change this? For all of them, it's whatever_key without changes in whatever. We couldn't change the whatever either, as then a previously set setting would be back to its default.

Stypox: Yeah, I only changed the string name to make it more clear that the option only refers to remote suggestions, but I left the value as it was before as not to reset settings

TobiGr:
I think we should also allow the user to stop showing local search suggestions. E.g. when showing something to another persion and using the search, one might not want to expose the search history.

Stypox:
What about replacing the current on-off option in settings with a spinner containing "All", "Only remote", "Only local", "None"?

TobiGr:
Sounds good

opusforlife2:
> What about replacing the current on-off option in settings with a spinner containing "All", "Only remote", "Only local", "None"?

I just checked to see what a spinner is. Wouldn't it be weird to have one odd spinner menu standing out when the rest of the settings all contain pop up dialogues?

Stypox:
@TobiGr done & updated the PR description
@opusforlife2 see screenshot

opusforlife2:
_That's_ a spinner? The Android documentation is misleading, then. It looked like a Kitkat option selector, there.

Stypox:
@opusforlife2 no, that's not a spinner, that's a multiselect menu, I said spinner just to show that I wanted to make the menu entry contain multiple things (probably also because no other name came to my mind ;-) )

app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
```diff
@@ -50,6 +55,32 @@ public static void initSettings(final Context context) {
         getAudioDownloadFolder(context);
     }
 
+    /**
+     * Backward-compatibility: convert boolean search suggestions value to new multi-select list.
+     * @param context the context to use to get shared preferences
+     */
+    private static void migrateSearchSuggestionsIfNeeded(final Context context) {
```
TobiGr, line 62: I think we should do the migration process a little different. First of all, it is good to store the last used/started NewPipe version (or better version code). That should also help us when creating a welcome tour and showing changelogs in future versions.
Depending on the last used version, we could do something like this:
``` Java
int lastStartedVersionCode = preferences.getInt(context.getString(R.string.last_started_newpipe_version_code), -1);
if (lastStartedVersionCode < 951) { // version code for 0.19.6
    // do migration here or call a method
}
// register later migrations here
```
The only thing we should figure out is how to differentiate between a new installation and a user updating. Both have not set `last_started_newpipe_version_code`, but a new installation does not need to run migrations. @wb9688 @B0pol @Stypox Any thoughts on this or suggestions for a different structure of the migration process?
 
We should update the last used version code if necessary at the end of https://github.com/TeamNewPipe/NewPipe/blob/049c8f70cdb12fa95ce6d61bfc70e64584fbf063/app/src/main/java/org/schabi/newpipe/App.java#L85

avently:
I found this PR in #3540, have the same problem. What's the state of the PR? Is something stops from merging?

Stypox:
This PR is blocked by the settings migration code, as explained by @TobiGr above, so his pr needs to be merged before

app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
Stypox: Done

Stypox:
Rebased and added settings migration. Here are the test apks; `before.apk` is from `dev`, and `after.apk` is from this PR; try to update from `before.apk` to `after.apk` after having changed the search suggestions setting.
[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5324281/debug.zip)

Stypox:
Rebased again. @TobiGr @opusforlife2 @B0pol @anpic could you test with the instructions above?
[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5470547/debug.zip)

Stypox:
@TobiGr @B0pol I rebased again. As always, here are the test apks; `before.apk` is from `dev`, and `after.apk` is from this PR; try to update from `before.apk` to `after.apk` after having changed the search suggestions setting. [debug.zip](https://github.com/TeamNewPipe/NewPipe/files/6159417/debug.zip)

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -723,70 +745,45 @@ private void initSuggestionObserver() {
...
                 .subscribe(listNotification -> {
                     if (listNotification.isOnNext()) {
-                        handleSuggestions(listNotification.getValue());
-                    } else if (listNotification.isOnError()) {
-                        showError(new ErrorInfo(listNotification.getError(),
+                        if (listNotification.getValue() != null) {
+                            handleSuggestions(listNotification.getValue());
+                        }
+                    } else if (listNotification.isOnError() && listNotification.getError() != null
+                            && !ExceptionUtils.isInterruptedCaused(listNotification.getError())) {
```
TobiGr, line 785:
```suggestion
                    } else if (listNotification.isOnError()
                            && listNotification.getError() != null
                            && !ExceptionUtils.isInterruptedCaused(listNotification.getError())) {
```

Stypox:
- First run migrations, then call the various `setDefaultValues`, since the latter requires the correct types. There was a crash with the "Content settings"' `setDefaultValues` call, on the search suggestion setting, since the `boolean` type, used on `dev`, does not match the new type of the (default) value, that is `StringSet`. Running the migrations before setting default values makes sense, so that's what I did.
- Show a snackbar for a suggestion error instead of the error panel.
- Do not show any error for interrupted I/O. Interrupted I/Os are expected since when the user inputs something new the previous suggestion fetch is interrupted.

I tested updating in the three ways possible (suggestions enabled/disabled in the before app and no before app installed at all) and for all of them I tried the 4 combinations of the search suggestions setting, everything worked fine.
As always, here are the test apks; `before.apk` is from `dev`, and `after.apk` is from this PR; you could change the search suggestions setting, then try to update from `before.apk` to `after.apk`.
[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/6159690/debug.zip)

TobiGr (dismissed):
Code looks good.

Stypox:
@TobiGr done in a9a11f99ddda4e764000348ab8d31e0b16ed5b1c

ShareASmile:
Is something that stops it from being merged?

Stypox:
I rebased and added a comment to explain how to add a migration to `SettingMigrations.java`, @TobiGr please merge if you think it is ready

Stypox:
I rebased again

TobiGr (approved):

TobiGr:
Damn, how many times did i forget about this PR? Sorry

PR #3567, opened 2020-05-10
Replace PrettyTime with DateUtils.getRelativeTimeSpanString()

sryze:
#### What is it?
- [X] Bug fix (user facing)
- [ ] Feature (user facing)

Remove dependency on PrettyTime by replacing it with a call to `DateUtils.getRelativeTimeSpanString`, which formats relative time in a similar way (although slightly differently):
https://developer.android.com/reference/android/text/format/DateUtils#getRelativeTimeSpanString(long)

One downside is that DateUtils abbreviates time unit names despite that the corresponding flag is not set (FORMAT_ABBREVIATE_RELATIVE). I didn't find a way to disable that...

PrettyTime has a bug with formatting relative time in Russian  - https://github.com/ocpsoft/prettytime/issues/182
For example, instead of "2 дня назад" (2 days ago) PrettyTime produces "2 дней назад", which is incorrect plural of "day" in Russian (same with other time units).

#### Testing apk
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4605511/app-debug.zip)

#### Agreement
- [ X I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -197,7 +197,6 @@ dependencies {
     implementation 'io.reactivex.rxjava2:rxjava:2.2.2'
     implementation 'io.reactivex.rxjava2:rxandroid:2.1.0'
     implementation 'com.jakewharton.rxbinding2:rxbinding:2.1.1'
-    implementation 'org.ocpsoft.prettytime:prettytime:4.0.3.Final'
 
     implementation "androidx.room:room-runtime:${roomDbLibVersion}"
     implementation "androidx.room:room-rxjava2:${roomDbLibVersion}"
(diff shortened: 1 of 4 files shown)


sryze:
I didn't know that the Android version didn't support as many languages as PrettytTime, sorry. In that case, it's probably better to fix PrettyTime itself like you mentioned.  I made a PR to fix this issue - https://github.com/ocpsoft/prettytime/pull/189

sryze:
FYU the PrettyTime PR was merged but the latest release (from April 2020) doesn't have it... so the bug is still present in NewPipe

B0pol:
Thanks for the information

@lincolnthree said there will be a release ASAP [here](https://github.com/ocpsoft/prettytime/pull/192#issuecomment-698531239)

I hope it'll be released before the next newpipe version that should come really soon too.

Issue #3420, opened 2020-04-12
Add "System default" theme option.


B0pol:
#### Describe the feature you want

I would like NewPipe theme to follow Android theme
#### Is your feature request related to a problem? Please describe it

If I change Android theme, it won't change NewPipe theme.
#### How will you/everyone benefit from this feature?

It will need less effort to change theme of my apps (apps without system theme option, including NewPipe).
It would also fix https://github.com/TeamNewPipe/NewPipe/issues/1535 for people who can set a system-wide theme settings (Correct me if I'm wrong, Android 10+ users, LineageOS 16 (and custom ROMs based on Lineage 16), and other custom ROMs implementing system-wide theme settings).

xmha97:
+1

opusforlife2:
@xmha97 To show agreement you give a thumbs up on the top comment, not add another comment.

julian-fairfax:
I'm just dropping in to say that I would also greatly appreciate this feature being added, as it would improve the integration with the android system, and would thus be a welcome change for me. Thanks for all the work you've put into this wonderful app!

LongJohn-Silver:
Will this be added or something or why was it closed, @Redirion ?

Redirion:
Yes as PR #3632 added this option. It will be part of the next NewPipe version

opusforlife2:
@LongJohn-Silver "X closed this in #abcd" means that issue was closed by a linked PR which fixes the bug/implements the feature request.

LongJohn-Silver:
@opusforlife2 Sorry I am using a third party client which did not show all the details.

Issue #1535, opened 2018-07-17
DayNight Mode


trymeouteh:
VLC offers light and dark themes but also allows it so that in the daytime the theme is light and at night the theme changes to dark so it is easier on the eyes. I would suggest looking at the VLC Andriod source code to understand how it is done in VLC to make it easier to be developed.

switchtegrax1:
it would also be nice if we could set the time manually if possible

wb9688:
>I would suggest looking at the VLC Andriod source code to understand how it is done in VLC to make it easier to be developed.

No need for that, it'd actually be pretty easy to do I think. VLC's source code is probably hard to read since they're using a lot of native code.

AnubrataDS:
I'm working on this

moneytoo:
The new option should be probably called ``System default`` and switch between ``Light`` and ``Dark`` themes.

However NewPipe actually has two non light themes: ``Dark`` and ``Black``. Shouldn't be there therefore two options: ``System default (dark)`` and ``System default (black)``?

I should be able to send PR if we agree on the option(s).

B0pol:
I think system default (dark) and system default (black) would be the best, yes.

PR #3632, opened 2020-05-22
Add settings to match device's theme (dark & black)

### LINKED ISSUE #687: Add full black theme.
The addition of the black theme was one of my favourite recent additions.

I wonder if you might consider adding a full black theme; doing away with the red bars. I find them quite distracting; they clash with my all-black theme elsewhere.

B0pol:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

Added two device theme settings:
+ Device theme (Dark)
+ Device theme (Black)

Device theme (Dark) and Device theme (Black) will set the theme to light if device's theme is light or unknown, otherwise it will set dark or black.

Available for Android 10+ only, LineageOS 16+ and other ROMs providing dark theme.

Here is a GIF:
newpipe_device_theme_gif

- Fixes https://github.com/TeamNewPipe/NewPipe/issues/3420
- Fixes #687
- Fixes **partially** #1535 : fixes for people who can set Automatic (time of the day) theme in Android settings.

#### Testing apk

[NewPipe_theme.zip](https://github.com/TeamNewPipe/NewPipe/files/5360974/NewPipe_theme.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ThemeHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ThemeHelper.java
@@ -21,6 +21,8 @@ package org.schabi.newpipe.util;
 
 import android.app.Activity;
 import android.content.Context;
+import android.content.res.Configuration;
+import android.content.res.Resources;
 import android.content.res.TypedArray;
 import android.util.TypedValue;
 import android.view.ContextThemeWrapper;
@@ -39,7 +41,8 @@ import org.schabi.newpipe.extractor.StreamingService;
 import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 
 public final class ThemeHelper {
-    private ThemeHelper() { }
+    private ThemeHelper() {
+    }
 
     /**
      * Apply the selected theme (on NewPipe settings) in the context
@@ -70,8 +73,13 @@ public final class ThemeHelper {
      * @return whether the light theme is selected
      */
     public static boolean isLightThemeSelected(final Context context) {
-        return getSelectedThemeString(context).equals(context.getResources()
-                .getString(R.string.light_theme_key));
+        final String selectedThemeString = getSelectedThemeString(context);
+        final Resources res = context.getResources();
+
+        return selectedThemeString.equals(res.getString(R.string.light_theme_key))
+                || (selectedThemeString.equals(res.getString(R.string.device_dark_theme_key))
+                || selectedThemeString.equals(res.getString(R.string.device_black_theme_key))
+                && !isDeviceDarkThemeEnabled(context));
     }
 
 
@@ -130,9 +138,12 @@ public final class ThemeHelper {
      */
     @StyleRes
     public static int getThemeForService(final Context context, final int serviceId) {
-        final String lightTheme = context.getResources().getString(R.string.light_theme_key);
-        final String darkTheme = context.getResources().getString(R.string.dark_theme_key);
-        final String blackTheme = context.getResources().getString(R.string.black_theme_key);
+        final Resources res = context.getResources();
+        final String lightTheme = res.getString(R.string.light_theme_key);
+        final String darkTheme = res.getString(R.string.dark_theme_key);
+        final String blackTheme = res.getString(R.string.black_theme_key);
+        final String deviceDarkTheme = res.getString(R.string.device_dark_theme_key);
+        final String deviceBlackTheme = res.getString(R.string.device_black_theme_key);
 
         final String selectedTheme = getSelectedThemeString(context);
 
@@ -141,8 +152,18 @@ public final class ThemeHelper {
             defaultTheme = R.style.LightTheme;
         } else if (selectedTheme.equals(blackTheme)) {
             defaultTheme = R.style.BlackTheme;
-        } else if (selectedTheme.equals(darkTheme)) {
-            defaultTheme = R.style.DarkTheme;
+        } else if (selectedTheme.equals(deviceDarkTheme)) {
+            if (isDeviceDarkThemeEnabled(context)) {
+                defaultTheme = R.style.DarkTheme;
+            } else {
+                defaultTheme = R.style.LightTheme;
+            }
+        } else if (selectedTheme.equals(deviceBlackTheme)) {
+            if (isDeviceDarkThemeEnabled(context)) {
+                defaultTheme = R.style.BlackTheme;
+            } else {
+                defaultTheme = R.style.LightTheme;
+            }
         }
 
         if (serviceId <= -1) {
@@ -157,12 +178,10 @@ public final class ThemeHelper {
         }
 
         String themeName = "DarkTheme";
-        if (selectedTheme.equals(lightTheme)) {
+        if (defaultTheme == R.style.LightTheme) {
             themeName = "LightTheme";
-        } else if (selectedTheme.equals(blackTheme)) {
+        } else if (defaultTheme == R.style.BlackTheme) {
             themeName = "BlackTheme";
-        } else if (selectedTheme.equals(darkTheme)) {
-            themeName = "DarkTheme";
         }
 
         themeName += "." + service.getServiceInfo().getName();
@@ -179,9 +198,12 @@ public final class ThemeHelper {
 
     @StyleRes
     public static int getSettingsThemeStyle(final Context context) {
-        final String lightTheme = context.getResources().getString(R.string.light_theme_key);
-        final String darkTheme = context.getResources().getString(R.string.dark_theme_key);
-        final String blackTheme = context.getResources().getString(R.string.black_theme_key);
+        final Resources res = context.getResources();
+        final String lightTheme = res.getString(R.string.light_theme_key);
+        final String darkTheme = res.getString(R.string.dark_theme_key);
+        final String blackTheme = res.getString(R.string.black_theme_key);
+        final String deviceDarkTheme = res.getString(R.string.device_dark_theme_key);
+        final String deviceBlackTheme = res.getString(R.string.device_black_theme_key);
 
         final String selectedTheme = getSelectedThemeString(context);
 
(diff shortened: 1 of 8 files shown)


B0pol:
I've rebased and fixed a bug around `isLightThemeSelected`.

app/src/main/java/org/schabi/newpipe/util/ThemeHelper.java
```diff
@@ -231,4 +269,28 @@ private static String getSelectedThemeString(final Context context) {
...
+     * Get the device theme
+     * <p>
+     * It will return true if the device's theme is dark, false otherwise.
+     * <p>
+     * From https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#java
+     *
+     * @param context the context to use
+     * @return true: dark theme, false: light or unknown
+     */
+    private static boolean getDeviceTheme(final Context context) {
```
TobiGr, line 283: The name of this method is irritating. What about
```suggestion
    private static boolean isDeviceNightModeEnabled(final Context context) {
```

app/src/main/res/values/strings.xml
```diff
@@ -66,6 +66,8 @@
     <string name="light_theme_title">Light</string>
     <string name="dark_theme_title">Dark</string>
     <string name="black_theme_title">Black</string>
+    <string name="device_dark_theme_title">Device theme (Dark)</string>
+    <string name="device_black_theme_title">Device theme (Black)</string>
```
TobiGr, line 70: To be honest, when testing the APK, I did not know what the two settings were supposed to change.

We need to make clear that the dark or black theme is used when the night mode is enabled, light theme otherwise. However, I have no idea how to put that in a few words.

app/src/main/java/org/schabi/newpipe/util/ThemeHelper.java
```diff
@@ -139,6 +150,18 @@ public static int getThemeForService(final Context context, final int serviceId)
             defaultTheme = R.style.BlackTheme;
         } else if (selectedTheme.equals(darkTheme)) {
             defaultTheme = R.style.DarkTheme;
```
TobiGr, line 152: Why are these two lines necessary?

app/src/main/java/org/schabi/newpipe/util/ThemeHelper.java
B0pol: isDeviceDarkThemeEnabled

app/src/main/res/values/strings.xml
B0pol: I don't know neither. I've only seen this way when there is dark black and light theme, but I agree it's not ideal

B0pol: @Stypox @wb9688 @Poolitzer @opusforlife2 what do you think about that? ^

opusforlife2: I've seen on various apps: Use/Follow system/device theme.

So you could go with: Follow system theme (dark/black).

Alternatively, the "follow system theme" option could be a separate toggle which works only with Dark and Black radio buttons. FairEmail uses such a toggle.

B0pol: I really like the idea of switch.

Poolitzer: I like how GitHubs android app does it:
Screenshot_20200606-142731 Screenshot_20200606-142735

wb9688: I also prefer "Follow system theme (dark/black)"

wb9688:
Imho we should hide the follow system theme options when the user's Android version doesn't support that.

B0pol:
> Imho we should hide the follow system theme options when the user's Android version doesn't support that.

I won't hide it for android 9 because of LineageOS supporting themes anyway. And SAF settings was doing this for a long time

app/src/main/java/org/schabi/newpipe/util/ThemeHelper.java
```diff
@@ -231,4 +265,28 @@ private static String getSelectedThemeString(final Context context) {
...
+     * <p>
+     * It will return true if the device's theme is dark, false otherwise.
+     * <p>
+     * From https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#java
+     *
+     * @param context the context to use
+     * @return true: dark theme, false: light or unknown
+     */
+    private static boolean isDeviceDarkThemeEnabled(final Context context) {
+        int deviceTheme = context.getResources().getConfiguration().uiMode
```
wb9688, line 280: `final`

Stypox:
From @B0pol accidentally posted in https://github.com/TeamNewPipe/NewPipe/pull/3471#issuecomment-653621955
I rebased and used a switch preference instead of "follow … (theme 1)" "follow … (theme 2)"

wb9688:
@B0pol: Your switch preference is flawed, because then it won't actually follow device theme when you've selected the light theme.

B0pol:
> @B0pol: Your switch preference is flawed, because then it won't actually follow device theme when you've selected the light theme.

that's intended. With what would you switch? Dark or black? If we had only dark and light I'd have make a simple third theme: auto. The point is to select between dark and black here

wb9688:
@B0pol: Yes, but why is it showing the light option when you could only choose between dark and black for the follow device theme functionality? I still prefer your previous UI though.

TobiGr:
I agree with @wb9688. What about replacing the switch with a list

**Follow device theme**
May not work on Android 9 and below. Device theme ignored.
May not work on Android 9 and below. Use (dark|black) theme on device's night mode.

opusforlife2:
When the follow system theme toggle is switched on, the light theme should be greyed out so only the dark and black options are selectable. That's what I had suggested previously.

B0pol:
I have another idea: 

- light
- dark
- black
- [maybe? Terminal theme]
- automatic (system)
- [in the future] automatic (battery saver)
- [in the future] automatic (day / night)

And after that two other lists
[In the future, if it exists)

- light 1
- light and cool 2

The second list:
- dark
- black
- [terminal?]

And if automatic is set, we switch between light and the selected night theme (default dark).

opusforlife2:
@B0pol See https://github.com/TeamNewPipe/NewPipe/issues/4148#issuecomment-674807664

Stypox:
@B0pol I think your comment above is a good idea. The list [The two lists] below should appear only if system theme is selected, and the user should be made aware of its [their] existance with a toast.

opusforlife2:
Ummm... I can't review the code. However, if you update the test apk... I can test the shit out of it. ✌️

B0pol:
Updated to use a list!
New APK: 
[NewPipe_theme.zip](https://github.com/TeamNewPipe/NewPipe/files/5360974/NewPipe_theme.zip)

Screenshots: 
 
When automatic is not selected, Night theme selector is removed:

opusforlife2:
Hmmm. I would have gone for:
- Light
- Dark
- Automatic

With the Night Theme selector showing for both Dark and Automatic. That's how I've generally seen it in other apps.

B0pol:
These apps don't have both dark and black themes.

Don't you think by removing black theme, some users will be angry? 😒

opusforlife2:
Wut? I'm not talking about removing anything. If you select Dark, then you should still be able to select "Dark" or "Black" from the Night Theme selector.

Stypox:
@opusforlife2 the night theme selector would be hidden if "Automatic" is not chosen

opusforlife2:
🤦 That is precisely what I'm suggesting to change.

> With the Night Theme selector showing for both Dark and Automatic.

What is happening here?

app/src/main/java/org/schabi/newpipe/util/ThemeHelper.java
```diff
@@ -141,8 +150,19 @@ public static int getThemeForService(final Context context, final int serviceId)
             defaultTheme = R.style.LightTheme;
         } else if (selectedTheme.equals(blackTheme)) {
             defaultTheme = R.style.BlackTheme;
-        } else if (selectedTheme.equals(darkTheme)) {
-            defaultTheme = R.style.DarkTheme;
+        } else if (selectedTheme.equals(automaticDeviceTheme)) {
+
+            if (!isDeviceDarkThemeEnabled(context)) {
```
Stypox, line 155: Invert this `if`

app/src/main/java/org/schabi/newpipe/util/ThemeHelper.java
```diff
@@ -191,6 +212,18 @@ public static int getSettingsThemeStyle(final Context context) {
             return R.style.BlackSettingsTheme;
         } else if (selectedTheme.equals(darkTheme)) {
             return R.style.DarkSettingsTheme;
+        } else if (selectedTheme.equals(automaticDeviceTheme)) {
+            if (!isDeviceDarkThemeEnabled(context)) {
```
Stypox, line 216: Also here

Stypox (changes requested):
Code looks good. But shouldn't there be something checking whether the phone NewPipe is running on supports automatic themes? E.g. my Huawei P9 lite with Android 7.0 does not have system themes, but the option is shown nonetheless, and for some reason it defaults to night.

opusforlife2 (commented):
I posted a comment, not a review, so I'm just restating it. Since you can choose between Dark and Black themes in the second menu, you should only see Light, Dark (or Night, since the other menu is called that?) and Automatic in the first menu.

Consequently, the second menu should show for both Automatic and Dark/Night.

Redirion:
@B0pol can you update your PR? I see no reason why this isn't yet merged.

B0pol:
I can't update it right now

yagoplx:
I was about to suggest this feature, why is this still open?

julian-fairfax:
I'm just dropping in to say that I would also greatly appreciate this feature being added, as it would improve the integration with the android system, and would thus be a welcome change for me. Thanks for all the work you've put into this wonderful app!

j-james:
Hey @B0pol, you might consider replacing the two-state dark mode list with a toggle switch. Maybe something like this?

mockup

Stypox:
I rebased, squashed some commits, added 731c65cd59e72921389c114f5f5ba4da8374336c which refactors ThemeHelper and tested thorougly on API 19 (emulated, where there is no device theme system setting) and on API 29 (Fairphone 3). Everything works fine and smoothly. @TobiGr This has also been tested before many times and I think it should be merged. @B0pol could you review my changes?

Redirion:
Tested on Pixel 4a (API 30). LGTM!

Redirion (approved):

PR #3704, opened 2020-05-31
Do not remove items generating errors form queue

### LINKED ISSUE #2268: stream cannot be played [wishlist]

Hi !
I often add musics to the queue however if there is no network for some time (eg 30s) for whatever reason, newpipe stops reading the stream (and display "cannot read stream") but also remove the stream from the queue.
It's really annoying since the stream is "lost" except if you remember the name / url of the stream(s).

It would be great if newpipe kept the stream in the queue (and play the next stream) :)

### LINKED ISSUE #3097: Store unplayed items in queue
If I had a nickel for every time I had a long queue lined up and accidentally pressed the back button while in the fullscreen player, thus losing the whole queue...

And just now, I had a long queue in the popup player, and pressed the fullscreen button, only the fullscreen player never appeared, it just ate my queue.

Oh and sometimes, my connection's not that great, so NewPipe just decides "can't play this stream" so that video gets deleted out of my queue, not to be found in my history or anything.  Even better when NewPipe discards like 10 videos before I realize anything's up.

Other scenarios include crashes and the like.  It would prevent a lot of grief if NewPipe would save the queue and offer to restore it on boot or somewhere in the UI

### LINKED ISSUE #3725: Option to disable unplayable stream removal
Seeing as I cannot reopen #3665 I am posting once again on the topic. I struggle with connectivity issues in many places both at home and during my commute, and if my active playlist transitions between songs during these connectivity hiccups my playlist will begin rapidly discarding streams until it is empty or it wraps around to streams that are still loaded. The last time I made this request it was closed as being derived from #3576; however, I have installed the latest update for NewPipe that was meant to fix #3576 and verified that my issue still remains in the most current version. As I stated in my reply to the previous issue, a "feature" that empties my playlist at the slightest sign of a bad connection is not one I want to have active, and while I can understand its existence in the case of deleted videos and the like that truly cannot ever be played, it is much easier for me to troubleshoot a single song not loading because there's nothing to load than it is for me to reassemble my entire playlist because I walked under the wrong shingle.

Stypox:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

<table>
<tr>
<td></td>
<td>When an item in the play queue failed to load, only some kind of errors were previously considered as "skippable" (i.e. go to the next stream without removing the current one). But other errors were not (e.g. network errors), so it was annoying in case of a temporary internet instability, because many streams would be removed from a possibly carefully handcrafted queue. This PR keeps the old behaviour when a stream cannot be played, except that every item is considered as "skippable". So if an erroring stream is encountered, the queue index is incremented without removing the old item. If the end of the queue is reached, the index never resets to 0 (even in repeating mode), to prevent "Could not play this stream" loops in case every stream fails.</td>
</tr>
</table>

- fixes #2268
- fixes #3097
- fixes #3725

#### Testing apk
@barudaret @gillhash @daufinsyd
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4707625/app-debug.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -64,7 +64,6 @@ import org.schabi.newpipe.player.helper.LoadController;
 import org.schabi.newpipe.player.helper.MediaSessionManager;
 import org.schabi.newpipe.player.helper.PlayerDataSource;
 import org.schabi.newpipe.player.helper.PlayerHelper;
-import org.schabi.newpipe.player.mediasource.FailedMediaSource;
 import org.schabi.newpipe.player.playback.BasePlayerMediaSession;
 import org.schabi.newpipe.player.playback.CustomTrackSelector;
 import org.schabi.newpipe.player.playback.MediaSourceManager;
@@ -77,7 +76,6 @@ import org.schabi.newpipe.util.ImageDisplayConstants;
 import org.schabi.newpipe.util.SerializedCache;
 
 import java.io.IOException;
-import java.net.UnknownHostException;
 
 import io.reactivex.Observable;
 import io.reactivex.disposables.CompositeDisposable;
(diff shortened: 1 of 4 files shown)


opusforlife2:
Yes! Finally! Praise be to our lord and saviour Stypox!

wb9688:
Why did that code even exist in the first place?

Stypox:
@wb9688 I have no idea 🤷‍♂

androidSwDev099:
I can't import or export settings into this debug version, but downloading audio files works. Is there a reason for this?

androidSwDev099:
Settings import/export with the debug version works fine one my Android 6 device.  I'm using 0.19.5 on both devices (Android 10 and Android 6) with no issues.

Stypox:
I don't remember for sure, but probably the apk I built was not on 0.19.5 but on the buggy 0.19.4. Also, that problem is completely unrelated from the changes made here

androidSwDev099:
@Stypox,
Debug app information has 19.3, so it seems to imply that the debug version was built with 19.3?  Guess I'll wait for another release. The bug has been causing me some misery with dropping videos in a 3000+ video list - cannot wait to be using a version with the fix :)

opusforlife2:
Is there a way (using Github Actions or something) that when a new version is released, the debug apks in various open PRs are automatically updated?

PR #3771, opened 2020-06-10
Completely remove autoplay preference (now always false)

### LINKED ISSUE #2232: Re-add ability to autoplay video when directly opened

There used to be a setting with which videos would automatically start playing if you opened the info page from another app. This functionality seems to have been removed in a recent update and I would like to request it to be added back in. (It was optional to begin with so wouldn't hurt anyone)

I realise that a similar behaviour is possible by selecting "Video Player" as the "Preferred 'open'  option" but that has one major drawback: You are not able to quickly go to view the video info activity by pressing the back button e.g. to go to the channel page or view related videos.

There doesn't even seem to be any way to get to the video info from the player as far as I can tell so an extra info button might be useful there too. (But I would prefer the old way because although a new button might work it would still break my "[workflow](https://xkcd.com/1172/)" ;P)

### LINKED ISSUE #2366: 'Show info' does not show info.
'Show info' does not show info but loads the video directly.
From all places where it appears (ie. from Downloads context menu, from What's new, etc.).

This is destructive when one has a limited dataplan.

Android 6
Huawei P8 Lite
```
android:versionCode="730" android:versionName="0.16.1" android:compileSdkVersion="28" android:compileSdkVersionCodename="9" package="org.schabi.newpipe" platformBuildVersionCode="730" platformBuildVersionName="0.16.1">
<uses-sdk android:minSdkVersion="19" android:targetSdkVersion="28"/>
```

Stypox:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

Completes the removal of the autoplay feature. Fixes #2366 and also closes #2232 since this is the chosen UX. Thanks to @nyanpasu64 for the investigation.

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
@@ -7,7 +7,6 @@ import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
 import android.os.Build;
-import android.preference.PreferenceManager;
 import android.util.Log;
 import android.widget.Toast;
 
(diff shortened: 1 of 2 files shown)


Stypox:
This is used, after #2907

Issue #3013, opened 2020-01-26
(Feature request) Status bar color


phnikola:
Hi all - Status bar color looks very outdated compared to Action bar.

Can this be offered as an option in menus that Status bar color is the same color as Action bar? Visually it looks alot more appealing.

Alternatively - would be great if someone could point to code line that renders this, I looked through it but without luck.

Stypox:
The status bar is currently a little bit darker than the action bar... isn't that OK and conforming to Material Design?

CharlieBrown12:
Thanks. I work in design for years and that darker shade in NewPipe is just not right.

Indeed - Material design guidelines recommend this but: 1. Many apps in PlayStore offer the setting to color the status bar in the exact tone of action bar. 2. Some Google apps like Gmail (versions from 2019 with red action bar) also have aligned statusbar and action bar (see screenshot from my phone).


3. NewPipe is not on Google's recommended apps list.
![Screenshot_20200127-100202_Chrome](https://user-images.githubusercontent.com/38468934/73161923-e82dda00-40ec-11ea-8d5a-6fca82279822.jpg)

I dont mean to cause any delay - you may try for yourself and see how NewPipe design looks Status bar in same tone as Action bar.
Meanwhile if you point me to the code line which does this - I would be grateful to try it.

CharlieBrown12:
![Screenshot_20200127-100202_Chrome](https://user-images.githubusercontent.com/38468934/73162037-30e59300-40ed-11ea-9e04-9f50720aedd5.jpg)

moneytoo:
I'm witch Charlie on that. I cannot find it now but I believe that later revisions of MD allowed it. Yes, Gmail and some other 1st party apps switched to single color (also some samples) before going white. Chrome also dropped generating darker status bar color from themeColor and uses single color now.

moneytoo:
Just trying to get rid of darker ``colorPrimaryDark`` results in looks like this:

![device-2020-02-03-225722](https://user-images.githubusercontent.com/1071643/73694749-02f9e300-46d9-11ea-8a07-d938aa16aefd.png)

I see that @Stypox already tracks that (make text white on red) in #3042 :)

PR #3774, opened 2020-06-12
Update status color to match toolbar color

eames-palmer:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

- Remove translucent status bar
- Change status bar color to match toolbar color across services

- Implements: (Feature request) Status bar color #3013

#### Testing apk
[app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4768501/app-debug.apk.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -34,8 +34,6 @@ import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
-import android.view.Window;
-import android.view.WindowManager;
 import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
 import android.widget.Button;
@@ -126,12 +124,6 @@ public class MainActivity extends AppCompatActivity {
         assureCorrectAppLanguage(this);
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_main);
-
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-            Window w = getWindow();
-            w.setFlags(WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS,
-                    WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS);
-        }
 
         if (getSupportFragmentManager() != null
                 && getSupportFragmentManager().getBackStackEntryCount() == 0) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -127,12 +125,6 @@ protected void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_main);
 
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-            Window w = getWindow();
-            w.setFlags(WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS,
-                    WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS);
-        }
-
```
Stypox, line 135: Since you removed this code, according to https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_TRANSLUCENT_STATUS you should use a half-translucent color for the toolbar, to get the same result as before.

eames-palmer: ### Clarifications:
1. Do we want the status bar to look the same as it did previously? I interpreted #3013 as looking for the same color status and toolbar instead of having the transparent dark overlay.
2. If we change the alpha of the status bar color it will blend with the background colors \#EEEEEE for the light theme, \#222222 for the dark theme, and \#000000 for the black theme which causes it to look even further from the toolbar color. In this case we could balance the color so that the change of alpha does not affect visual appearance, however, this requires balancing for every background color and service color option (it is also not very straightforward.)
3.	Lastly, looking at Google’s own apps that use the matching status/toolbar colors (e.g. clock, hangouts, and photos) their status bars appear to have alpha values of one.

### Google Apps:

Google Photos status bar color

Google Clock status bar color

### Without Matching: 

Newpipe status bar color with half alpha and side menu open

Newpipe status bar color with half alpha on trending tab

### Rough Match:

Newpipe status bar color with half alpha, attempted match, and side menu open

Newpipe status bar color with half alpha and attempted match on trending tab

eames-palmer:
Thank you for the review! After some thorough research/testing, this commit should contain updates for all the status bars present within NewPipe. Feel free to critique any of the changes I have made or leave further suggestions! I did have to update the `colorPrimaryDark` for the themes, however, this seemed like the best option as directly changing the status bar color with `android:statusBarColor` is not supported in NewPipe's minimum SDK.

eames-palmer:
@Stypox Yes! I triple checked and ran some global searches just to make sure. It looks like `colorPrimaryDark` is only for the contextual app bar's in NewPipe which, from my testing, only impacted the status bar color!

Issue #3148, opened 2020-02-27
Newpipe v0.18.5 not load youtube channel :(


Domiuns:
- [x ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [x ] I checked if the issue/feature exists in the latest version.
- [x] I did use the [incredible bugreport to markdown converter](https://teamnewpipe.github.io/CrashReportToMarkdown/) to paste bug reports.

Hi, the new version of newpipe not load channel videos (i have all video channel on screening but when i click to open it newpipe is loading for unlimited time) only comment section load, I think is a problem with new extractor. I have android 4.4
![IMG_20200227_141004](https://user-images.githubusercontent.com/53193434/75448314-414a9100-596b-11ea-8919-11874e06b689.JPG)
![IMG_20200227_140728](https://user-images.githubusercontent.com/53193434/75448429-7b1b9780-596b-11ea-9f53-17da954c8b13.JPG)

Stypox:
Are you sure this only happens when opening videos from a channel, and it happens every time? I followed all the steps you provided but I couldn't reproduce this issue unfortunately.
@wb9688

wb9688:
@Stypox: There is, in NewPipeExtractor at least (not sure about NewPipe), no difference between opening a video from a channel or otherwise. I can't reproduce it either

@Domiuns: What countries/languages have you set in Settings->Content? Are you using a proxy or VPN? If not, in what country are you located?

Domiuns:
I have tried and it happens with any video, even if a video in the trend I managed to upload it. I am Italian and I have the feed Italy

Domiuns:
I have finnaly load an video for test all. Buth other video not load.
This is worked
![IMG_20200227_161614](https://user-images.githubusercontent.com/53193434/75457385-b377a180-597c-11ea-9c45-5d79b9bd4e2b.JPG)
![IMG_20200227_161625](https://user-images.githubusercontent.com/53193434/75457399-b7a3bf00-597c-11ea-932b-77707cfde971.JPG)
![IMG_20200227_161647](https://user-images.githubusercontent.com/53193434/75457415-bbcfdc80-597c-11ea-900b-27d910d61c98.JPG)
With 0.18.4 i not have problem. (I not sue vpn)

Domiuns:
I have Sniff traffic and i not receive nothing from some youtube channel.

[...]

PR #3777, opened 2020-06-13
Support SAF properly

### LINKED ISSUE #3465: Native File Picker For importing/exporting backup
### Describe the feature you want
Native file picker for selecting file for import/export.
### Is your feature request related to a problem? Please describe it
While selecting file to import/export, the app shows custom file picker which can't be used to select file from cloud service like nextcloud.
### How will you/everyone benefit from this feature
In case of native file picker the nextcloud option will be visible in there only. So selection will be possible. 
Screenshot_20200421-143038_1_1

wb9688:
#### What is it?
- [ ] Bug fix (user facing)
- [ ] Feature (user facing)

1. Use SAF or NoNonsense-FilePicker *everywhere* based on the setting.
2. Don't have `requestLegacyExternalStorage` anymore in the manifest.
3. Only allow NoNonsense-FilePicker for versions below Android 5 and only allow SAF starting at Android 10.
4. Move `SharpInputStream` into NewPipe and add `SharpOutputStream`.
5. Show folder picker in download dialog if no download folder set yet.

I probably need to clean up some code and we should definitely do some more testing.

Fixes #3465 and perhaps some others.

#### Testing apk
[app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4778503/app-debug.apk.zip)

#### Agreement


app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -175,30 +173,43 @@ public void onActivityResult(final int requestCode, final int resultCode,
...
+                if (FilePickerActivityHelper.isOwnFileUri(getContext(), uri)) {
+                    uri = Uri.fromFile(Utils.getFileForUri(uri));
+                }
+                if (requestCode == REQUEST_EXPORT_PATH) {
+                    final StoredDirectoryHelper directory
+                            = new StoredDirectoryHelper(getContext(), uri, null);
+                    final SimpleDateFormat sdf
+                            = new SimpleDateFormat("yyyyMMdd_HHmmss", Locale.US);
+                    exportDatabase(directory.createFile("NewPipeData-"
+                            + sdf.format(new Date()) + ".zip", "application/zip"));
```
wb9688, line 187: Do we really want a directory picker here and not a file picker with a pre-filled name?

app/src/main/java/us/shandian/giga/io/StoredFileHelper.java
```diff
@@ -380,4 +381,64 @@ private boolean stringMismatch(String str1, String str2) {
...
+            i = new Intent(Intent.ACTION_CREATE_DOCUMENT)
+                    .putExtra("android.content.extra.SHOW_ADVANCED", true)
+                    .setType("*/*")
+                    .addCategory(Intent.CATEGORY_OPENABLE)
+                    .addFlags(Intent.FLAG_GRANT_PERSISTABLE_URI_PERMISSION
+                            | StoredDirectoryHelper.PERMISSION_FLAGS);
+
+            if (startPath != null && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                i.putExtra(DocumentsContract.EXTRA_INITIAL_URI, Uri.parse(startPath));
+            }
```
wb9688, line 416: Currently we're not using this

app/src/main/java/us/shandian/giga/io/StoredFileHelper.java
```diff
@@ -37,6 +41,19 @@
     private String srcName;
     private String srcType;
 
+    public StoredFileHelper(final Context context, final Uri uri, final String mime) {
```
wb9688, line 44: Perhaps we should move `StoredFileHelper` and `StoredDirectoryHelper` to NewPipe as well.

app/src/main/res/values/strings.xml
```diff
@@ -324,6 +324,7 @@
     <string name="msg_wait">Please wait…</string>
     <string name="msg_copied">Copied to clipboard</string>
     <string name="no_available_dir">Please define a download folder later in settings</string>
```
wb9688, line 326: Should I remove this now that we're not using this anymore?

wb9688:
@imcmjha, @Domiuns, @opusforlife2: Would you mind testing this? You should test whether all file pickers (import/export DB, import/export subscription, download) show the correct file picker (SAF or NoNonsense-FilePicker) and whether saving or reading that file works correctly.

On Android 4.4-, SAF should be disabled and impossible to enable.

On Android 5-9, SAF should be enabled by default.

On Android 10+, SAF should be enabled and impossible to disable.

opusforlife2:
I have Android 9:
1. Export/Import subscriptions works perfectly fine with both SAF and NNFP.
2. Instead of showing the full content URI, maybe the text up to "/tree/primary:" should be replaced by "Internal storage: " followed by the user's selected path?
3. SAF is enabled by default on first run, which is correct.
    - However, SAF is _also_ automatically enabled whenever you open the Download menu. Sure enough, the app uses SAF when you try to import/export. The preference should stay off if the user toggles it off. If you _don't_ open the Download menu after toggling it off, the app uses NNFP.
    - Even though SAF is automatically toggled on, the download location strings still show the NNFP locations. Toggling it twice shows the default SAF strings ("Downloaded video/audio files are stored here").
    - Repeatedly toggle the pref a few times. Notice that the app sets a default location in case of NNFP, but not in case of SAF. Either the same folders should be set as content URIs by default in case of SAF too, or the app should ask for new default locations for NNFP as well.
    - I suspect you should also check from logs if SAF is toggled on if you close then open the app. I can't check this from just the UI because of this bug.

wb9688:
@opusforlife2:

>Instead of showing the full content URI, maybe the text up to "/tree/primary:" should be replaced by "Internal storage: " followed by the user's selected path?

No, as different storage providers will start with a different thing.

>However, SAF is also automatically enabled whenever you open the Download menu. Sure enough, the app uses SAF when you try to import/export. The preference should stay off if the user toggles it off. If you don't open the Download menu after toggling it off, the app uses NNFP.

You mean when you open the downloads activity where you could see all the things you downloaded? Hmm… that setting shouldn't be changed unless the user changes it (or you upgrade to Android 10+, as then only SAF will be possible).

>Even though SAF is automatically toggled on, the download location strings still show the NNFP locations. Toggling it twice shows the default SAF strings ("Downloaded video/audio files are stored here").

I'm pretty sure I fixed that bug, though maybe only for Android 10 or maybe I forgot to update the APK in this issue.

>Repeatedly toggle the pref a few times. Notice that the app sets a default location in case of NNFP, but not in case of SAF. Either the same folders should be set as content URIs by default in case of SAF too, or the app should ask for new default locations for NNFP as well.

That can't be done with SAF, because you don't know what that path would be and you have to acquire permission to be able to use that (only possible by opening the directory picker for that).

>I suspect you should also check from logs if SAF is toggled on if you close then open the app. I can't check this from just the UI because of this bug.

What do you mean?

opusforlife2:
Some questions that came up while bug testing:
1. Is Subscription Import a full replacement operation? Can/Shouldn't there be an additional option to compare the JSON to be imported with one's already existing Subscription database so that only new channels are imported?
2. Why is the default download location for NNFP "Music/Movies > Newpipe"? I think it should be "Newpipe > Videos/Audios". We can't assume the user wants media downloaded through Newpipe to be added to their music/video player libraries by default. I tend to think of it as 'contamination', because those two folders are usually exclusively accessed by music/video player apps. It's perfectly alright if the user wants to download straight to those folders, but it shouldn't be the default behaviour.
3. Is this a bug or expected behaviour? : Delete a file (not just its entry) from the Downloads activity. It will still be present in the file system and even be playable. It only gets deleted from the file system when you back out of the app entirely.

opusforlife2:
> You mean when you open the downloads activity where you could see all the things you downloaded?

No, I mean the download settings menu. Sorry, that was ambiguous.

> What do you mean?

Leave it. It's just a precaution. I'll test it once the above bug is fixed.

opusforlife2:
Export/Import DB also seems to work fine with both SAF and NNFP.

Question: Does export/import DB work like a _complete_ app backup, or does it leave something out?

wb9688:
@opusforlife2: It should be a complete app backup, other than the download history.

opusforlife2:
Alright. Then everything works. 👍

app/src/main/java/us/shandian/giga/io/StoredFileHelper.java
wb9688: See #3781

imcmjha:
I tested in Android 9 and it is working perfectly. [Bughttps://github.com/TeamNewPipe/NewPipe/issues/3465) resolved. Thanks a lot.

wb9688:
@opusforlife2: I fixed the bug regarding SAF looking like it was enabled again when opening the download settings. Could you try the new APK?

Also, did you notice how the download dialog now shows a directory picker when you haven't set a download folder yet?

opusforlife2:
Confirmed fixed.

> Also, did you notice how the download dialog now shows a directory picker when you haven't set a download folder yet?

I didn't know that it didn't before. This is the first time I'm using SAF. Good, though. Adds to the seamlessness.

xibr:
Everything seems to work correctly on Android 10

app/src/main/res/values/strings.xml
B0pol: Which one?
Please wait is being used [here](https://github.com/wb9688/NewPipe/blob/use-saf-by-default/app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java#L982)
Copied to clipboard [here](https://github.com/wb9688/NewPipe/blob/use-saf-by-default/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java#L104) and [here](https://github.com/wb9688/NewPipe/blob/use-saf-by-default/app/src/main/java/us/shandian/giga/util/Utility.java#L212)
Please define … [here](https://github.com/wb9688/NewPipe/blob/use-saf-by-default/app/src/main/java/org/schabi/newpipe/settings/DownloadSettingsFragment.java#L278) and [here](https://github.com/wb9688/NewPipe/blob/use-saf-by-default/app/src/main/java/us/shandian/giga/service/DownloadManagerService.java#L566)

B0pol:
With SAF it works fine on android 10 … but the import / export will pick the last used folder, both import or download

Let's say you export in NewPipe/Save, download musics in NewPipe/Music, videos in NewPipe/Videos. Music will always be downloaded in NewPipe/Music, and videos in NewPipe/Videos. 
Monday, you export your database (in NewPipe/Save). You download music tuesday (NewPipe/Music), make a playlist wesnesday therefore you want to export your database. But it will show the last used folder, i.e. NewPipe/Music. That make no sense, you want NewPipe to show up NewPipe/Save as you picked it on monday.

app/src/main/res/values/strings.xml
wb9688: >no_available_dir here and here

You're right, sorry, though I don't think that's the most appropriate string to use in those places.

wb9688:
>Let's say you export in NewPipe/Save, download musics in NewPipe/Music, videos in NewPipe/Videos. Music will always be downloaded in NewPipe/Music, and videos in NewPipe/Videos.
>Monday, you export your database (in NewPipe/Save). You download music tuesday (NewPipe/Music), make a playlist wesnesday therefore you want to export your database. But it will show the last used folder, i.e. NewPipe/Music. That make no sense, you want NewPipe to show up NewPipe/Save as you picked it on monday.

@B0pol: That's just default behavior of SAF. Default behavior of NNFP is to always start at /sdcard/. The only other thing we could do is add a new preference for that (not necessarily user-visible in the settings) and use that [here](https://github.com/TeamNewPipe/NewPipe/pull/3777#discussion_r439819577).

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
wb9688: See #3785

TobiGr: A dir picker seems good to me

wb9688: @TobiGr: Why would you even prefer a directory picker over a file picker here?

Domiuns:
@wb9688 when try to import all setting as zip app crash.

wb9688:
@Domiuns: On 4.4? What error are you getting?

Domiuns:
@wb9688 yes android 4.4 but log not give info is empty

{"user_action":"ui error","request":"App crash, UI failure","content_language":"it_IT","service":"none","package":"org.schabi.newpipe.debug.refactorstoredhelpers","version":"0.19.5","os":"Linux Android 4.4.4 - 19","time":"2020-07-29 15:30","exceptions":[""],"user_comment":""}

I not understand why is empty the exceptions

wb9688:
@Domiuns: I can't reproduce on an AVD. That ACRA bug is known and should be fixed by #3982. Could you try the following APK that also contains that fix, #3781 and #3785? [app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4995813/app-debug.apk.zip)

Domiuns:
@wb9688 yes now i have log, thanks
## Exception
* __User Action:__ ui error
* __Request:__ App crash, UI failure
* __Content Language:__ it-IT
* __Service:__ none
* __Version:__ 0.19.8
* __OS:__ Linux Android 4.4.4 - 19


```
java.lang.NullPointerException
	at java.io.RandomAccessFile.<init>(RandomAccessFile.java:118)
	at us.shandian.giga.io.FileStream.<init>(FileStream.java:20)
	at org.schabi.newpipe.streams.io.StoredFileHelper.getStream(StoredFileHelper.java:188)
	at org.schabi.newpipe.settings.ContentSettingsFragment.importDatabase(ContentSettingsFragment.java:243)
	at org.schabi.newpipe.settings.ContentSettingsFragment.lambda$onActivityResult$2$ContentSettingsFragment(ContentSettingsFragment.java:187)
	at org.schabi.newpipe.settings.-$$Lambda$ContentSettingsFragment$DothoYKAlQb2nBRMlrmwCmdVF4k.onClick(lambda)
	at com.android.internal.app.AlertController$ButtonHandler.handleMessage(AlertController.java:169)
	at android.os.Looper.loop(Looper.java:153)
	at android.app.ActivityThread.main(ActivityThread.java:5297)
	at java.lang.reflect.Method.invokeNative(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:511)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:833)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:600)
	at dalvik.system.NativeStart.main(Native Method)

```
</p></details>
<hr>

Domiuns:
Normal
IMG_20200729_190946
This
IMG_20200729_190959

wb9688:
>This PR not have check box

@Domiuns: That's intended, as you should only be able to select a single file.

Does exporting database work? And does importing a YouTube OPML work? For example, [subscription_manager.txt](https://github.com/TeamNewPipe/NewPipe/files/4996600/subscription_manager.txt) (in case you don't have a YouTube account, like me).

Domiuns:
- Import OPML work.
-Import zip file now work well.
-Export not work.
@wb9688 Thanks for all help and work and sorry for disturb.

wb9688:
@Domiuns: No, problem, I found the problem. The issue was that some code that changes the path to something usable will be executed when we initialize the `StoredFileHelper` in a certain way, but the code that created that `StoredFileHelper` executed that code as well, and it doesn't work when you (try to) change that path twice.

>I have never used logcat sorry.

Do you have a rooted phone? Or a PC and a USB cable? Not that that matter for now as (I think) I found the problem, but for next time such an issue occurs maybe.

---

Could you confirm that all of the following things in [this APK](https://github.com/TeamNewPipe/NewPipe/files/4999449/app-debug.apk.zip) work?

- Export database
- Import database
- Export subscriptions
- Import subscriptions from previous export
- Import subscriptions from YouTube
- Setting default audio download path
- Setting default video download path
- Downloading audio
- Downloading video
- Downloading audio when "Ask where to download" is enabled
- Downloading video when "Ask where to download" is enabled

Thanks in advance.

Stypox:
@wb9688 everything works on Android 7.0, both with and without SAF. The only difference with SAF is that files are never overwritten, but a ` (1)` is appended instead (which I think is ok).

Domiuns:
All now work i have test all list.
I have only this problem when i try to download subtitle.

## Exception
* __User Action:__ ui error
* __Request:__ App crash, UI failure
* __Content Country:__ IT
* __Content Language:__ it-IT
* __App Language:__ it_IT
* __Service:__ none
* __Version:__ 0.19.8
* __OS:__ Linux Android 4.4.4 - 19

```
java.lang.NoClassDefFoundError: java.nio.charset.StandardCharsets
	at org.schabi.newpipe.streams.SrtFromTtmlWriter.<init>(SrtFromTtmlWriter.java:25)
	at us.shandian.giga.postprocessing.TtmlConverter.process(TtmlConverter.java:28)
	at us.shandian.giga.postprocessing.Postprocessing.run(Postprocessing.java:174)
	at us.shandian.giga.get.DownloadMission.doPostprocessing(DownloadMission.java:702)
	at us.shandian.giga.get.DownloadMission.lambda$1gGnKY4qsIM8af3LqOOlrMGS3J4(DownloadMission.java)
	at us.shandian.giga.get.-$$Lambda$DownloadMission$1gGnKY4qsIM8af3LqOOlrMGS3J4.run(lambda)
	at java.lang.Thread.run(Thread.java:838)

```
</details>
<hr>


Yes i have a pc and USB to micro USB cable. I need linux for use logcat?

I want to thank you for fixing all the problems.

wb9688:
@Domiuns: You don't need Linux for logcat. That error is the same as the one you had in #3148 and as I could see it already is at a step after what this PR changes, that shouldn't be caused by this PR.

Stypox:
@fajis1 you have Android 11, don't you? Could you please check if the actions above work correctly for you?

wb9688:
@Stypox: On Android 10+ it could only be enabled (I edited your comment to reflect that) ;)

Anyway, I merged the other 2 PRs into this one and uploaded the fix for @Domiuns' issues.

fajis1:
@Stypox I installed and tried... But I don't have an external SD card slot on my phone. It's a Pixel 4 XL. 

So it didn't work for me. 


It was just grayed out.

wb9688:
@fajis1: What do you mean? You should be able to download stuff without having a SD card.

fajis1:
@wb9688 when I followed the directions to go to settings and enable SAF it is grayed out, and the subtext for the SAF button talks about being able to download to SD cards. 

See the link for the picture. 
https://photos.app.goo.gl/jF51pBd8KZ8F2Axj7

wb9688:
@fajis1: Ah, that should indeed be the case on Android 10+, that's what @Stypox misunderstood, sorry for the confusion. You should just test whether all things mentioned in https://github.com/TeamNewPipe/NewPipe/pull/3777#issuecomment-666176793 work.

wb9688:
This could not be merged yet, due to some Android 11 user saying he has issue with downloading, see https://www.reddit.com/r/NewPipe/comments/hr21tf/newpipe_v0196_has_been_released/fzyq7ql/

wb9688:
I rebased on `dev` and changed it so that it doesn't unnecessarily ask for the storage permission when using SAF.

New APK: [app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/5011902/app-debug.apk.zip)

opusforlife2:
No change for me to test on LOS 16, right?

wb9688:
@opusforlife2: Not really, other than that it won't ask for storage permission on download when using SAF, so maybe test that downloading still works (and that it asks for permission when you disable SAF)? First uninstall previous version of NewPipe use-saf-by-default though.

opusforlife2:
Tested what you mentioned. Works fine.

Question: is SAF better than NNFP? Is there any advantage to either of them? (SAF already has a +1 in that it doesn't need storage permission now.)

wb9688:
@opusforlife2: SAF is the only thing that works on Android 10+ (though you could request a legacy mode in Android 10 to work around that) and since a certain version the only thing that allows access to external storage (like SD cards). SAF allows apps to provide storage, so that you could save directly to your NextCloud instance or whatever. SAF is more secure in that it can't access anything a user doesn't explicitly give access to, however SAF is a bit slower.

TobiGr:
@Domiuns Could you please uninstall the app for this PR and install the latest APK again? When testing in a **KitKat** emulator, I am not asked to give any storage permissions. Thus the download is failing. When downloading audio, I get the message that "the destination folder cannot be created". But when I download a video, the error message "no space left on device" is shown after some parts of the video were downloaded (I guess into the app's storage).

Did not find any problems with other Android versions.

wb9688:
@TobiGr: Hmm… I'll look into that. Thanks for pointing that out!

wb9688:
@TobiGr: It indeed doesn't ask storage permission on Android 4.4, but that's a feature that has been introduced in Android 6.0. Downloading also works fine for me in an emulator though.

Domiuns:
@TobiGr android 4.x-5.x not ask per missione because is all permission allowed, permission has been introduced on android 6.x

I have import issues with your last apk i have try to make a log of problem but log is empty

wb9688:
@Domiuns: Hmm… what kind of issues are you having exactly?

Domiuns:
I have install the apk uploaded on post

Testing apk

app-debug.apk.zip.  <----- this i have install
Agreement

Not work zip setting import
Export subscription json file

ingingin:
What the fuck? What happened to the v.0.20.0??

Stypox:
@Domiuns could you further explain what you mean by "not works"? Does the app crash? If that's the case and by "log is empty" you mean that there is no stacktrace in the bugreport screen, then @wb9688 you might need to provide a rebased apk with the bugreport fix

Domiuns:
@Srypox this apk work https://github.com/TeamNewPipe/NewPipe/pull/3777#issuecomment-667650552
But the apk uploaded on start of conversation not work (from this instruction https://github.com/TeamNewPipe/NewPipe/pull/3777#issuecomment-667987837 )

Stypox:
After merging this we should add a setting migration that enables SAF on every device possible, and then let the user disable it if they want, otherwise almost nobody would know the switch exists

opusforlife2:
> Only allow NoNonsense-FilePicker for versions below Android 5 and only allow SAF starting at Android 10.

Is there any downside to only allowing SAF on Android 5+ instead of 10+?

Stypox:
@opusforlife2 I think the "only" is misplaced there. What he means is that users with Android 10+ will only see saf, and users with <5 will only see nononsense, while all versions in between will see both.

opusforlife2:
I know what he meant. What I'm asking is if there is any downside of having SAF only for Android 5+. It will remove a setting and make things more consistent and code easier to maintain: NNFP only for Kitkat, SAF only for 5+.

Stypox:
Oh ok
@wb9688 is there a reason not to do this?

Issue #3871, opened 2020-07-10
East-asian-friendly number format wanted


chr56:
When people meet large numbers,such as 1234567890,we separate them into groups.European and North Amerian separate them by 3 digits ,like 1,234,567,890.But East-asian traditionally separates them by 4 digits,like12,3456,7890.
So does the language: In English,dividing it by 3 digits ,first comma from right side stands for"thousand",second stands for "million"...But,some east-asian ,such as Chinese, dividing it by 4 digits,first comma from right side ,in "4-digits-mode",has its own name(not like"ten thousand“ but a unique name)...
However,NewPipe uses first form,and not localizes the numbers for East-asians.
So,there are only "thousand""million"billion" to be translated in resource.

AudricV:
Is it still an issue with 0.26.1?

PR #3872, opened 2020-07-10
Localize numbers in NewPipe for east-asian

chr56:
#### What is it? 
- [X] Feature (user facing)

As #3871 said, this PR
- change codes to localize Chinese's number system.
  including large number dividing ,introducing large number names in Chinese (like thousand ,million,billion)...

- #3871 

#### Testing apk

[app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/4903204/app-debug.apk.zip)

#### Agreement


chr56:
These are demo pictures
Before
After

I know you may not speak English ,so here is English original：
en

chr56:
However I just did it for Simplifid Chinese ,maybe I can complish Traditional Chinese or more...

B0pol:
- why did you remove zh hans CN filey
- doesn't an android method or library exist for that
- you should not use settings_keys.xml unless it's a key for a preference.
- you should add other languages (Japan for example)
- you should use identifiers to make it simpler to understand. E.g. 万 = tenThousandsAbbreviation, 億 = oneHundredMillionsAbbreviation…

chr56:
> * why did you remove zh hans CN file?
> * doesn't an android method or library exist for that
> * you should not use settings_keys.xml unless it's a key for a preference.
> * you should add other languages (Japan for example)
> * you should use identifiers to make it simpler to understand instead of the original name in Chinese, eg YI (it could be different in Indian for example).
>   E.g. 万 = tenThousandsAbbreviation, 億 = oneHundredMillionsAbbreviation…

1.values-b+zh+HANS+CN is a duplicate of values-zh-rCN,and the latter one is more "formal".
2.It seems that there is not one for that
3.So, should I put them in values-zh-rCN/Strings.xml (or somewhere else）？
4.I would add more lauguage for this "4-digits-number-system"
5.Ok,I`d rename strings.

Are these above passible?

chr56:
Wait,it seems that there is android.icu.NumberingSystem offical liberary which could help. @B0pol

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -185,17 +198,38 @@ public static String localizeWatchingCount(final Context context, final long wat
 
     public static String shortCount(final Context context, final long count) {
```
mitosagi, line 199:
```suggestion
    public static String shortCount(final Context context, final long count) {
        if (Build.VERSION.SDK_INT >= 24) {
            return CompactDecimalFormat.getInstance(getAppLocale(context),
                    CompactDecimalFormat.CompactStyle.SHORT).format(count);
        }
```
You can easily localize if you focus on Android 7.0 or later, which has a built-in ICU library.
Even if it is below Android 7.0, it will work if you bundle the library, but I want to avoid it because it is about 12MB.

Also in this case, I think the `short_billion`, `short_million` and `short_thousand` on Weblate need a note that they are only used for below Android 7.0.

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -134,7 +135,19 @@ public static String localizeNumber(final Context context, final long number) {
 
     public static String localizeNumber(final Context context, final double number) {
         NumberFormat nf = NumberFormat.getInstance(getAppLocale(context));
-        return nf.format(number);
+        switch ((getAppLocale(context).toString())) {
+            case "zh_CN":
```
mitosagi, line 139: In Japanese and Korean, numbers are read every 4 digits, but are separated by a comma every 3 digits. So, I'd like to add that it's correct that it's not included in this case classification.

Issue #3490, opened 2020-04-26
Allow user to specify a default download type (Video, Audio, or Captions, etc)


androidSwDev099:
#### Describe the feature you want

Currently when a users wants to download contents of a video, the download type in the Download dialogue box always defaults to the Video radio button being selected.

I propose an enhancement that allows the user to specify the default download type radio button selection (the three types I'm aware of are Video, Audio, and Captions - are there others?).   

This could  be accomplished by:
1. Create a new "Default download type" setting option under Settings -> Download which lets the user specify default download type as Video, Audio, or Captions. 

or

2. The last download type the user selected in the download dialogue would be the default selected for the next download.  For example if the user downloads Audio for the current video selected, then the default download type would now be Audio so that the next time the user initiates a download for a video, the Audio radio button would be selected by default in the Download dialogue box.  

Of course if the current default download type is not applicable for the next video download request (e.g. default is Captions and the current video doesn't support captions download), then software logic would have to detect this condition and select a valid download type for this video download request and show a valid radio button selected for the download.

In all cases the user could still change the download type before selecting the OK button in the Download dialogue box just as things currently work.

#### How will you/everyone benefit from this feature?

The enhancement will allow users who want to download a sequence of videos with the same download type other than Video to do so with fewer clicks. I mostly use NewPipe to download the audio content of a video, so this enhancement will significantly reduce the number of times I have to change the download type radio button.

For users who exclusively download Video, this enhancement would have no impact.

TobiGr:
The better approach is to store the last used download type and auto-select it.

PR #3901, opened 2020-07-18
Add settings for default download media type.

vmazoyer:
#### What is it?
- [x] Feature (user facing)

- Users can now select a default media type for their downloads.

- Closes feature request #3490. This implementation follows the suggestion (1): a default value in preferences.

- Added a new setting in the download preferences.

#### Agreement
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them. Please note that I do not plan to maintain this feature.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
@@ -519,6 +519,11 @@ public class DownloadDialog extends DialogFragment
         String defaultMedia = prefs.getString(getString(R.string.default_download_type),
                 getString(R.string.default_download_type_default));
 
+        if (defaultMedia.equals(getString(R.string.select_last_used_download_type))) {
+            defaultMedia = prefs.getString(getString(R.string.last_used_download_type),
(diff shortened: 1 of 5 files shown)


TobiGr:
The better approach is to store the last used download type and auto-select it.

vmazoyer:
How about adding "Last used" as a 4th option?
In my use case I prefer to be able to select a media type rather than "last used".

opusforlife2:
> How about adding "Last used" as a 4th option?

Best of both worlds.

Stypox:
> Best of both worlds.

Nope, @TobiGr suggested to always use the last configuration in order not to add yet another niche option.

opusforlife2:
> Nope, @TobiGr suggested to always use the last configuration in order not to add yet another niche option.

Throw it in a new Miscellaneous Settings menu we've discussed earlier, I say!

vmazoyer:
I made the changes according to the feedback here: no additional setting; just remember the last option used.
To keep things clean I decided to rebase and open a new PR: #4038.

PR #3928, opened 2020-07-23
Disable shrinkResources

wb9688:
#### What is it?
- [ ] Bug fix (user facing)
- [ ] Feature (user facing)

Needed for F-Droid reproducible builds. Will make the APK about 20 kB bigger.

#### Agreement

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -50,7 +50,7 @@ android {
         // TODO: update Gradle version
         release {
             minifyEnabled true
-            shrinkResources true
+            shrinkResources false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
             archivesBaseName = 'app'
         }


opusforlife2:
Question: as of 0.19.8, is it possible to install all three versions (F-Droid, Newpipe's F-Droid repo, Github releases) over each other?

wb9688:
@opusforlife2: Now F-Droid publishes 2 APKs. F-Droid's F-Droid ≠ F-Droid's F-Droid with our signature ≈ NewPipe's F-Droid = NewPipe's GitHub.

opusforlife2:
Both the F-Droid app and website show just one apk. Which one is it? And where is the other?

opusforlife2:
Oh wait. The website shows 2. What's up with the app, then?

wb9688:
@opusforlife2: F-Droid's app does show both, notice how v0.19.8 appears twice under versions.

opusforlife2:
Ah. Yes. I had to enable "Include incompatible versions". So, just to clarify once more, the apk you publish on Github, the one you publish on your repo, and the one _you send_ to F-Droid (not the one they build themselves) all have the same signature, right?

wb9688:
@opusforlife2: To clarify:

The one we publish on GitHub and our F-Droid are exactly the same. They are signed with v1 and v2 signatures.

The one we send the signatures of to F-Droid is signed with the same signature, but only v1 as F-Droid doesn't support v2 yet.

F-Droid then compiles our APK and checks whether the signature we sent them is correct for that APK they built. If that's the case, they sign it with their signature and publish that APK. Then they also replace their signature with our signature and publish that as well.

opusforlife2:
Thanks. I think this info should be put somewhere on the homepage of the Newpipe website.

opusforlife2:
Oh, by the way, apart from the v2 signature, there's another difference between the F-Droid built apk and yours, right? The update functionality.

wb9688:
@opusforlife2: Nope, that's 100% the same in all APKs, but it checks the signature and only runs with our signature. That does mean it does indeed also run on the one in F-Droid's F-Droid repo with our v1 signature.

opusforlife2:
Should the download URL be configurable now that you can get the update from 3 places?

opusforlife2:
@DI555 here you go. Difference between apks.

PR #3931, opened 2020-07-24
Improve energy efficiency by applying Dynamic Retry Delay Energy Pattern + Cache Energy Pattern 

anasofiagribeiro:
This improves the energy efficiency of SecondScreen by applying the Dynamic Retry Delay Energy Pattern and the Cache Energy Pattern for mobile applications.

The energy pattern was applied in RouterActivity.java and AndroidManifest.xml (for the Dynamic Retry Delay Energy Pattern) and TLSSocketFactoryCompat.java (for the Cache Energy Pattern). The general idea for the first one is to avoid accessing an internet connection if the connection is down . In particular, a class NetworkStateReceiver will be created in the java file that checks for an internet connection. The general idea for the second one is to avoid performing unnecessary operations by using cache mechanisms. In particular, setting the cache size to have no limit.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -4,8 +4,13 @@ import android.annotation.SuppressLint;
 import android.app.AlarmManager;
 import android.app.IntentService;
 import android.app.PendingIntent;
-import android.content.*;
 import android.content.pm.PackageManager;
+import android.content.BroadcastReceiver;
+import android.content.SharedPreferences;
+import android.content.Intent;
+import android.content.DialogInterface;
+import android.content.Context;
+import android.content.ComponentName;
 import android.net.ConnectivityManager;
 import android.net.Network;
 import android.os.Bundle;
@@ -638,8 +643,7 @@ public class RouterActivity extends AppCompatActivity {
                 }
                 Choice playerChoice = (Choice) serializable;
                 handleChoice(playerChoice);
-            }
-            else {
+            } else {
                 NetworkStateReceiver.enable(getApplicationContext());
             }
         }
@@ -754,7 +758,8 @@ public class RouterActivity extends AppCompatActivity {
 
         //The method hasActiveNetwork() checks whether the network connection is active
         protected boolean hasActiveNetwork() {
-            final ConnectivityManager connManager = (ConnectivityManager) getApplicationContext().getSystemService(Context.CONNECTIVITY_SERVICE);
+            final ConnectivityManager connManager = (ConnectivityManager)
+                    getApplicationContext().getSystemService(Context.CONNECTIVITY_SERVICE);
             Network activeNetwork = connManager.getActiveNetwork();
             return (activeNetwork != null);
         }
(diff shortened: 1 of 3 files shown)


wb9688:
Please follow our code style, so that Checkstyle won't complain.

Redirion:
Sorry but this pattern appears to be outdated since Android 7, as the CONNECTIVITY_CHANGE cannot be received anymore from BroadcastReceivers registered in AndroidManifest.

Please see the Android doc: https://developer.android.com/reference/android/net/ConnectivityManager#CONNECTIVITY_ACTION

anasofiagribeiro:
Hello. I have a question, if Android version 7 has API Level 24 (https://source.android.com/setup/start/build-numbers), how come "registerDefaultNetworkCallback", that was added in API Level 24, doesn't work? Thank you for your time.

anasofiagribeiro:
Hello. Since the solution I proposed only works for Android 7 or newer, would a solution where this change only applies on Android 7 or newer be acceptable? This guarantees that for Android 7 or older it doesn’t break. 
“if (Build.VERSION.SDK_INT < 7 || hasActiveNetwork() ) {”

Stypox:
> would a solution where this change only applies on Android 7 or newer be acceptable?

Yes, that's the way to go, thank you ;-)

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -626,16 +631,26 @@ public void onCreate() {
 
         @Override
         protected void onHandleIntent(@Nullable final Intent intent) {
-            if (intent == null) {
-                return;
-            }
+            if (Build.VERSION.SDK_INT < 7 || hasActiveNetwork()) {
```
Redirion, line 634:
```suggestion
            if (Build.VERSION.SDK_INT < 24 || hasActiveNetwork()) {
```

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -745,6 +760,71 @@ public void onDestroy() {
...
+                    final AlarmManager alarmManager = (AlarmManager)
+                            context.getSystemService(Context.ALARM_SERVICE);
+                    final Intent innerIntent = new Intent(context, FetcherService.class);
+                    final PendingIntent pendingIntent =
+                            PendingIntent.getService(context, 0, innerIntent, 0);
+
+                    SharedPreferences preferences =
+                            context.getSharedPreferences(context.getPackageName(),
+                                    Context.MODE_PRIVATE);
+                    preferences.edit();
```
Stypox, line 796: Why `.edit()`?

Stypox (changes requested):
It doesn't build because of checkstyle, see https://travis-ci.org/github/TeamNewPipe/NewPipe/builds/716366001
Also, could you further explain what you did? Is it safe to put the ssl cache to infinity? And are we sure we want to do nothing silently when the user shares something to NewPipe until the connection is active (if I understood correctly that's what the code does)?

app/src/main/java/org/schabi/newpipe/RouterActivity.java
Redirion:
```suggestion
```
unintente obviosly

mhmdanas:
@anasofiagribeiro  I'm closing this PR, but please make a comment when you want to resume work on this so I would re-open it. Thanks!

PR #3930, opened 2020-07-24
Improve playlist dialogs, use only one layout

Stypox:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)
- [x] Code base improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
Use the same class and the same layout both to append an item to a local playlist and to choose a local/remote playlist tab. The resulting `PlaylistDialog` class is the result of a merge between the old `local.dialog.Playlist*Dialog` and `SelectPlaylistFragment` classes. In particular:
- The custom adapter found in `SelectPlaylistFragment` has been kept, to allow for both local and remote items to exist in the list
- The correct state saving behaviour is taken from `local.dialog.PlaylistDialog`
Fixes ui issues with the select playlist tab dialog, which didn't show uploaders and stream counts, looked strange and had double title on Android 4.4.

#### Depends on
#3982 (due to bug report fix)

#### Testing apk
Tested on emulated Android 4.4 and device Android 7.0.
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4987201/app-debug.zip)

#### Agreement
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Issue #3076, opened 2020-02-11
Change text of lonpgress menu


bkqtnte10:
I often find myself confused what to tap when I want to play a video in the queue.

I now realised that the wording of the long press popup is unintuitive:
Especially "Enqueue in a new popup" is confusing. Is it played in a new popup or later?

Proposed solution is to change wording and what is displayed:
- If nothing is currently playing only show "Play in background" and "Play in popup"
- If something is playing, additionaly show "Enqueue in background" and "Enqueu in popup"

Maybe the icons for the respective functions could also be added: Headphones for background picture-in-picture for popup.

OffifialMITX:
Additionally a "play triangle" on the bottom left corner for starting playback now and a plus on said corner on queue buttons.

IMO the whole menu, including the share function could be replaced with an icon-only (with text on long-press) one.

Stypox:
The plan is #2583

opusforlife2:
Hence close.

PR #3936, opened 2020-07-25
Add toolbars to list info items with fast actions

### LINKED ISSUE #2583: Fast actions from mini info item - UI / UX ideas
# Problem
I have been thinking about the long press menu lately, and I realized it is inconvenient because:
1. Entries have no corresponding icon, so our eyes can't distinguish between two entries at the first sight.
2. Entries are not always in the same position on the screen, so we can't use muscle memory. Even though this was partially addressed by #2354 & #2368, the position of entries depends anyway on the menu length.
3. There are too many entries, making the menu too crowded, and there are many requests for new entries (#437 #2198 #2510 #1142).
4. There is no way to select multiple streams and apply an action on all of them.

# Solution
This is the solution I propose, sorry for the bad quality ;-)

menu

Screen A contains normal info item list. By tapping on the blue *stream* info item we get to screen B. Yes, video info is not loaded, instead the item expands to show some action buttons above and below it! Here is why I think this is a good idea:
- Most of the times when I click on a video I just want to play it, and currently I have to wait for video info to load before being able to play anything (#2306). With this solution it would take less time.
- Video info can be loaded by clicking again on the item.
- Icons are always in the same place (relative to the clicked stream) and with the same icon (no text, as icons are representative enough). This solves problems 1. and 2. above. Also, actions about stream content (play&download) are separated from the specific ones that only use stream info (add-to-playlist, share, delete-from-playlist, set-as-playlist-thumbnail, etc.).
- Since play-actions are displayed where the item was before, and the item is shifted down (see how the above separator is not moved in screen B), one can remember where an option is and double-tap in the correct place directly from screen A. This would take no time.
- Background and popup play buttons can be long-pressed to enqueue (consistent with the video info activity).
- There is place for new actions, in case we ever need them. We would just need to add a button to the bar below the item.

By long-pressing on the blue *stream* info item on screen A, the multiple-select options are shown in the action bar (similar to https://github.com/TeamNewPipe/NewPipe/issues/2354#issuecomment-494871457), solving the problem 4. above. By tapping once on any other *stream* info items we can select them, getting to situation C. A tick is displayed above selected items' thumbnails and the action bar is populated with actions about stream content.
The three-dot menu cointains the other actions and also two options useful for selection purposes: "select below" and "select above", only available in non-infinite lists. These can be considered a replacement to a "select all" button, but are more powerful and would solve #2510 #2559. If a playlist is long and is not fully loaded, the "select below" button would lazily select all items so that they are loaded only if needed. 

# Feedback
I would implement this but I need feedback both from designers and from users. Would this idea suit your needs? Would it be pleasant to the eye? If you disagree with any of the assumptions I made above, be sure to tell me ;-)

@theScrabi @TobiGr @TheAssassin @nv95 @Gymcap @karyogamy @justanidea @Lirlisa @gkeegan @BartFarley @xak2000 @jojoxl01

Stypox:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

<table><tbody><tr><td></td><td>
This unfinished PR introduces the item toolbars promised in #2583, as shown in the images on the left. Even though all buttons are in the layout, only share, add to playlist, and play/enqueue have been implemented so far.<br>
The issue is, I realized this could not be the best UI for non-techy people, as it is not usual that clicking on a list item does not open it but only shows some related actions. So I would like to get some feedback from you: do you find this ui comfortable or did you prefer the long-press menu?<br>
Btw, even though in #2583 I said I would make two different actions for thumbnail and other data double-press, I realised it is better not to put too many things into a small item without a clear indication. So for now clicking twice anywhere on the item first opens the toolbar and then shows info.
</td></tr></tbody></table>

- Closes #2583 
- OTHERS (will specify later)

- #3930

#### Testing apk
This apk is not complete at all. There are many issues (one of them is: adding local playlists to other local playlists is disabled because strange loops would arise). So don't expect a pleasant experience, but please tell me if the overall way of doing things is in your opinion better or worse.
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4975306/app-debug.zip)

#### Agreement


opusforlife2:
Oh my! What is it with this Apple-like secretiveness and flashy reveal? XD

opusforlife2:
Right off the bat, I noticed that long press on a list item does nothing. I think rather than a single tap to expand and another to open, it would be more intuitive to open on single press and expand on long press instead.

heberjeur:
Can you separate actions, when I click to text It appears icons list and when I click to image it open directly and with one press the video.
shot_01

opusforlife2:
@heberjeur
> Btw, even though in #2583 I said I would make two different actions for thumbnail and other data double-press, I realised it is better not to put too many things into a small item without a clear indication.

tomjuggler:
I have to say I just tried this and I am a big fan of the one-click menu popup. Maybe making the play button more prominent would be a suggestion I could make (simply putting it closer to the middle of the screen would work). To be honest I never used the long press anyway. If you just added the download button in there I would be very happy with this change myself.

opusforlife2:
Crash: Malformed URL when opening a channel page on any service.

MD77MD:
bug: "add to playlist" button is showing in playlists toolbar when it should not.

MD77MD:
I love how long pressing acts as enqueue however this should be mentioned in like a tips section or how to use introduction when using newPipe for the first time.. it

also long pressing on the play button could act as enqueue next.

I also like previous suggestion. having the toolbar while pressing on thumbnail to play video directly is a great combination.

opusforlife2:
#3076 reminded me: Since text has been replaced with just icons, you should absolutely show tooltips for all of them. The current "hold to append" tooltip for background and popup icons should show for the list item icons as well, if the toggle is on.

MD77MD:
in your proposal, you mentioned multi select feature... I hope that is not gone. this online is considered a big update. imagine how much less clicking is needed whenever y want to add or remove a bunch of videos from playlists or queues.

heretoenhance:
> also long pressing on the play button could act as enqueue next (play next). this would be even more useful after updating to unified player.

I was going to open a feature request for this, having "play next" would be an awesome addition to this already ultra awesome app. Thank you all guys for your work!

MD77MD:
https://github.com/TeamNewPipe/NewPipe/issues/4405#issuecomment-710826700

Stypox:
I will not continue thir PR, as I changed my mind about #2583: having such a menu popping up below items, with rather small buttons, would feel strange and probably also difficult to use.

PR #3982, opened 2020-07-28
Fix ACRA bug reports not containing stack trace

Stypox:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

This PR fixes blank reports not containing the stack trace (as seen in #3930). This was caused by `MultiDex.install` not being called before ACRA initialization, so ACRA could not load some of its classes and failed to initialize the stack trace collector (among others). See ACRA/acra#619 and [ACRA wiki](https://github.com/ACRA/acra/wiki/Troubleshooting-Guide#legacy-multidex).

This PR also contains two other improvements to ACRA's initialization code:
- Skip ACRA initialization if the current process ends with `:acra` (i.e. is created by ACRA): https://github.com/ACRA/acra/wiki/Troubleshooting-Guide#applicationoncreate
- Remove deprecated calls to set the `ReportSender` class in the configuration: `setReportSenderFactoryClasses()` is deprecated, now extensions (`ReportSenderFactory` is an extension) should be registered using `@AutoService`. This commit adds two compile-time dependencies and only fixes a deprecation, so it can be skipped if needed. See https://github.com/ACRA/acra/wiki/Custom-Extensions#by-annotation

I think this should be merged as soon as possible, as otherwise it is not possible debug APKs by giving them to people, as there would be no stack trace in case of errors.

<sup>It took me ages to find the root cause of this 😩</sup>

#### Testing apk
This apk crashes every time a video is opened, use that to check if ACRA correctly catches the unhandled exception.
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4987144/app-debug.zip)

#### Agreement

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -89,6 +89,7 @@ ext {
     androidxRoomVersion = '2.2.5'
     groupieVersion = '2.8.0'
     markwonVersion = '4.3.1'
+    googleAutoServiceVersion = '1.0-rc7'
 }
 
 configurations {
@@ -174,6 +175,9 @@ dependencies {
     implementation "com.google.android.exoplayer:extension-mediasession:${exoPlayerVersion}"
 
     implementation "com.google.android.material:material:1.1.0"
+
+    compileOnly "com.google.auto.service:auto-service-annotations:${googleAutoServiceVersion}"
+    kapt "com.google.auto.service:auto-service:${googleAutoServiceVersion}"
 
     implementation "androidx.appcompat:appcompat:1.1.0"
     implementation "androidx.preference:preference:1.1.1"
--- a/app/src/debug/java/org/schabi/newpipe/DebugApp.kt
+++ b/app/src/debug/java/org/schabi/newpipe/DebugApp.kt
@@ -1,6 +1,5 @@
 package org.schabi.newpipe
 
-import android.content.Context
 import androidx.multidex.MultiDex
 import androidx.preference.PreferenceManager
 import com.facebook.stetho.Stetho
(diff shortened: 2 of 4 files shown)


app/build.gradle
```diff
@@ -201,6 +201,8 @@ dependencies {
     implementation "com.nononsenseapps:filepicker:4.2.1"
 
     implementation "ch.acra:acra-core:5.5.0"
+    compileOnly "com.google.auto.service:auto-service-annotations:1.0-rc7"
+    kapt "com.google.auto.service:auto-service:1.0-rc7"
```
wb9688, line 205: Could you move this up to the other `com.google` dependencies? And could you put the version in `ext` instead? That's what we do if we depend on multiple parts of the same library that thus needs the same version.

Stypox: > Could you move this up to the other com.google dependencies?

I put them here since they are related to ACRA, but ok

> And why depend on a RC?

Apparently ACRA's (probably a little bit outdated) documentation suggests `1.0-rc6`, so I chose the latest RC, `1.0-rc7`. The version suggested in e.g. google/auto#665 (again, outdated) is also an rc, and in the [releases page](https://github.com/google/auto/releases) there is no AutoService version which is not an RC (with `rc1` dating back to 2014).

Stypox: Done

Stypox: Done

wb9688: >I put them here since they are related to ACRA, but ok

No, they're not related to ACRA, we just currently happen to only use it in that code.

>Apparently ACRA's (probably a little bit outdated) documentation suggests 1.0-rc6, so I chose the latest RC, 1.0-rc7. The version suggested in e.g. google/auto#665 (again, outdated) is also an rc, and in the releases page there is no AutoService version which is not an RC (with rc1 dating back to 2014).

I found out about there not being a non-RC version just after I commented, so I think that's OK.

Issue #1778, opened 2018-10-02
zero brightness turns off the screen


lingislong:
- [X] I checked if the issue/feature exists in the latest version.
Hello , A bug that appears in : ver 0.14.0 and **0.14.1** 
on **androin 4.1.2** ,I checked on 2 different devices **galaxy s3** and **lg optimus 4x**

When the brightness is lowest -using the gesture , the screen turns off, brightness can not be increased back ...
To "fix" this, you need to **clean all** the application's caches..(and actually resat the brightness level)
I will be very happy to fix it soon ..

PS This is the best app in the universe!!!

delightfulagony:
Is this Issue still open? Is there anyone working on it?

mhmdanas:
Is anybody still able to reproduce this?

efb4f5ff-1298-471a-8973-3d47447115dc:
Do u guys think this i relatable? Suddenly my brightness wont update through the use of gestures.

https://user-images.githubusercontent.com/73130443/115885413-77848300-a43f-11eb-9eae-f67e0394788d.mp4

Imericxu:
I just tried and couldn’t reproduce on the most recent release (0.21.2).
Using OnePlus 6 testing both with and without automatic brightness.

PR #4037, opened 2020-08-01
Don't allow brightness percentage below 1% when using gesture control

XiangRongLin:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

When the gesture control for brightness results in a value below 0.01f (1%), then increase it to 0.01f
Even though the [doc](https://developer.android.com/reference/android/view/WindowManager.LayoutParams#screenBrightness) suggests otherwise, this avoids a sudden spike in brightness when scrolling down to the minimum. 

- #1778
- #4169 

#### Testing apk
after rebase: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5360763/app-debug.zip)

@delightfulagony @lingislong
Could you check out if it works?

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
+++ b/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
@@ -271,7 +271,10 @@ public class PlayerGestureListener
             bar.setProgress((int) (bar.getMax() * Math.max(0, Math.min(1, oldBrightness))));
             bar.incrementProgressBy((int) distanceY);
 
-            final float currentProgressPercent = (float) bar.getProgress() / bar.getMax();
+            float currentProgressPercent = (float) bar.getProgress() / bar.getMax();
+            if (currentProgressPercent <= 0.01) {
+                currentProgressPercent = 0.01f;
+            }
             layoutParams.screenBrightness = currentProgressPercent;
             window.setAttributes(layoutParams);
 


opusforlife2:
I thought this was Newpipe reverting to system brightness when the slider hit zero? If you have automatic brightness on, there is no spike if you're in the dark. It only happens when there is some light.

Maybe a better alternative is that the slider only controls custom brightness from 0 to 100, and there is a separate button (like you have in camera apps with the exposure slider) to reset to system brightness? Otherwise the only way to get system brightness again would be to swipe away the app from memory.

XiangRongLin:
> Otherwise the only way to get system brightness again would be to swipe away the app from memory.

I just checked and yes, with the unified player, it does not return to system brightness when closing the stream, before it did. There already is an issue for it #4043, so i would not merge this until that is resolved.

I don't like the current behaviour since if i want it be really dark, then i have to slide down slowly to get close to 0. I don't use automatic brightness, so sliding all the way down does not work

avently:
> I don't like the current behaviour since if i want it be really dark, then i have to slide down slowly to get close to 0

I don't like it too but it was made like this before me so I just didn't touch it.
Also the most confusing thing is this: 
https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java#L422

So the brightness will be changed unexpectedly and it's annoying. I would like to see this be configurable at least. @XiangRongLin, can you make this behavior configurable? 
Something like:
```java
if (useSmartBrightnessSwitch && (System.currentTimeMillis() - timestamp) > TimeUnit.HOURS.toMillis(4)) {
....
} else {
return true brightness value without all this "smart" things
}
```

opusforlife2:
@XiangRongLin So what do you think of a button to revert to system brightness? It should only show when using the brightness gesture, of course.

XiangRongLin:
> 
> 
@opusforlife2 
> So what do you think of a button to revert to system brightness? It should only show when using the brightness gesture, of course.

I personaly see no use for reverting to system brightness, but since it was possible before one can consider adding a button for it.

But only showing it while changing brightness will be really hard to use. After you adjust the brightness the brightness indication disappears really quickly, which is good since it is blocking the video.
So you would have to be really fast to click it. Additionally you have to find an icon which represent "revert to system brightness" that everyone understands on first view, otherwise it will just create confusion.

opusforlife2:
Oh wow. VLC doesn't even let you go below 1%.

XiangRongLin:
> 
> 
> Oh wow. VLC doesn't even let you go below 1%.

yeah that's where i got this. Just with nicer code, since they use kotlin `coerceIn` (something like that).

opusforlife2:
Alright. A permanent button to defer to system brightness would be too niche. Let's drop the idea then, I suppose.

app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
```diff
@@ -271,7 +271,10 @@ private boolean onScrollInMain(final MotionEvent initialEvent, final MotionEvent
             bar.setProgress((int) (bar.getMax() * Math.max(0, Math.min(1, oldBrightness))));
```
XiangRongLin, line 271: After i rebased it, when you scroll all the way down and continue downwards the wheel jumps around a bit, while the brightness stays at 1%. My guess is that because here a float gets converted to an int. On my phone `bar.getMax()` returns 810, so with that conversion you lose something (810 * 0.01 = 8.1 => 8)

Changing it to `Math.max(0.01 , Math....)` doesn't help.

XiangRongLin:
I notices that the brightness spike doesn't occur anymore on my android 11 phone.
Im going to close this for now. If someone still has that problem please comment.

B0pol:
I still have the problem of minimum brightness brighter than just above minimum brightness. I have an android 10 phone

PR #4038, opened 2020-08-01
Remember last selected media type for downloads.

### LINKED ISSUE #3490: Allow user to specify a default download type (Video, Audio, or Captions, etc)

vmazoyer:
#### What is it?
- [x] Feature (user facing)

- NewPipe now remembers which media type was downloaded last and will select that option the next time the user wants to download something.

- Closes #3490. This implementation follows the suggestion (2): remember last choice.
- Replaces PR #3901.

#### Testing apk

[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5010087/debug.zip)

#### Agreement
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them. Please note that I do not plan to maintain this feature.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
@@ -516,7 +516,23 @@ public class DownloadDialog extends DialogFragment
         videoButton.setVisibility(isVideoStreamsAvailable ? View.VISIBLE : View.GONE);
         subtitleButton.setVisibility(isSubtitleStreamsAvailable ? View.VISIBLE : View.GONE);
 
-        if (isVideoStreamsAvailable) {
+        prefs = PreferenceManager.getDefaultSharedPreferences(getContext());
+        final String defaultMedia = prefs.getString(getString(R.string.last_used_download_type),
+                    getString(R.string.last_download_type_video_key));
+
+        if (isVideoStreamsAvailable
+                && (defaultMedia.equals(getString(R.string.last_download_type_video_key)))) {
+            videoButton.setChecked(true);
+            setupVideoSpinner();
+        } else if (isAudioStreamsAvailable
+                && (defaultMedia.equals(getString(R.string.last_download_type_audio_key)))) {
+            audioButton.setChecked(true);
+            setupAudioSpinner();
+        } else if (isSubtitleStreamsAvailable
+                && (defaultMedia.equals(getString(R.string.last_download_type_subtitle_key)))) {
+            subtitleButton.setChecked(true);
+            setupSubtitleSpinner();
+        } else if (isVideoStreamsAvailable) {
             videoButton.setChecked(true);
             setupVideoSpinner();
         } else if (isAudioStreamsAvailable) {
@@ -595,6 +611,7 @@ public class DownloadDialog extends DialogFragment
         final StoredDirectoryHelper mainStorage;
         final MediaFormat format;
         final String mime;
+        final String selectedMediaType;
 
         // first, build the filename and get the output folder (if possible)
         // later, run a very very very large file checking logic
@@ -603,6 +620,7 @@ public class DownloadDialog extends DialogFragment
 
         switch (radioStreamsGroup.getCheckedRadioButtonId()) {
             case R.id.audio_button:
+                selectedMediaType = getString(R.string.last_download_type_audio_key);
                 mainStorage = mainStorageAudio;
                 format = audioStreamsAdapter.getItem(selectedAudioIndex).getFormat();
                 switch (format) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
```diff
@@ -515,7 +515,22 @@ protected void setupDownloadOptions() {
         videoButton.setVisibility(isVideoStreamsAvailable ? View.VISIBLE : View.GONE);
         subtitleButton.setVisibility(isSubtitleStreamsAvailable ? View.VISIBLE : View.GONE);
 
-        if (isVideoStreamsAvailable) {
+        prefs = PreferenceManager.getDefaultSharedPreferences(getContext());
+        String defaultMedia = prefs.getString(getString(R.string.last_used_download_type),
+                    getString(R.string.video));
+
+        assert defaultMedia != null;
+        if (isVideoStreamsAvailable && (defaultMedia.equals(getString(R.string.video)))) {
```
TobiGr, line 523: `(defaultMedia.equals(getString(R.string.video))` this will fail to detect the last used type when the user changed tha pplication language. Please create three keys in settings_keys.xml to avoid that.

app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
```diff
@@ -663,6 +682,11 @@ private void prepareSelectedDownload() {
 
         // check for existing file with the same name
         checkSelectedDownload(mainStorage, mainStorage.findFile(filename), filename, mime);
+
+        // remember the last media type downloaded by the user
+        SharedPreferences.Editor editor = prefs.edit();
```
TobiGr, line 687:
```suggestion
        final SharedPreferences.Editor editor = prefs.edit();
```

opusforlife2:
I still think that we should make this a setting and put it under "Miscellaneous Settings". Once that happens, it'll open up the floodgates for a lot of niche toggles and settings that people want to implement. Then, after that, based on user feedback, some of those settings might be popular enough to put in regular menus. We can only experiment with them if they exist, after all.

The defaults for all such settings would be as if that menu didn't exist at all, naturally.

vmazoyer:
Thank you for the review @TobiGr. The suggested changes are implemented in https://github.com/TeamNewPipe/NewPipe/pull/4038/commits/4f40bd6ae58c9d8732f8ba2f3c842a2a8b3264bf.

app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
```diff
@@ -515,7 +515,24 @@ protected void setupDownloadOptions() {
         videoButton.setVisibility(isVideoStreamsAvailable ? View.VISIBLE : View.GONE);
         subtitleButton.setVisibility(isSubtitleStreamsAvailable ? View.VISIBLE : View.GONE);
 
-        if (isVideoStreamsAvailable) {
+        prefs = PreferenceManager.getDefaultSharedPreferences(getContext());
+        final String defaultMedia = prefs.getString(getString(R.string.last_used_download_type),
+                    getString(R.string.last_download_type_video_key));
+
+        assert defaultMedia != null;
```
TobiGr, line 522: According to the [docs](https://developer.android.com/reference/android/content/Context#getString(int)), `getString(resInt)` cannot be `null`, thus defaultMedia cannot be `null` (as long as we don't put anything stupid into the preferences, which we clearly do not). I think, the assertion can be removed.

TobiGr:
looks good to me! Please squash the commits, I'll merge afterwards.

Stypox (approved):
This is good, thank you :-D

I'm squashing the commits myself, so that this can be merged right away

Stypox:
Done, @TobiGr merge ;-)

TobiGr (approved):

PR #4054, opened 2020-08-02
Service extensions

wb9688:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

This allows loading a .zip file that will be added as a service. That .zip file should contain the following files:

- `about.json`: a JSON file containing:
  - `name`: name of the extension
  - `author`: author of the extension
  - `class`: class that extends `StreamingService`
  - `major_version` and `minor_version`: internal version of NewPipeExtractor that it's compatible with
  - `replaces`: service ID that it replaces (optional)
- `classes.dex`: this should be compiled as a Java library depending on NewPipeExtractor as a compile-only dependency, the resulting .jar should go through `d8` to turn it into a .dex
- `icon.png`: icon for the service

You could find an example implementations of extensions at https://github.com/wb9688/bandcamp-extension and https://github.com/wb9688/craptube-extension.

Note how the DB uses service ID, which could change, so that doesn't work well if you add/remove extensions or update to newer NewPipe versions that have more services built-in.

##### To do
- Show a bigger warning explaining the dangers
- Add an in-database service ID <-> service name table
- Compare whole signature
- Handle failed upgrades correctly
- Implement some kind of safe mode for when an extension causes crashes

TeamNewPipe/NewPipeExtractor#379 and #3777

#### Testing apk
[app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/5012566/app-debug.apk.zip)
[bandcamp.zip](https://github.com/TeamNewPipe/NewPipe/files/5012462/bandcamp.zip)
[craptube.zip](https://github.com/TeamNewPipe/NewPipe/files/5012463/craptube.zip)

#### Agreement


wb9688:
@opusforlife2: In case you wanted to test, I attached the .apk and the 2 demo extensions. Know that this is pretty WIP though.

wb9688:
@opusforlife2: You should be able to test it now as I implemented a UI for managing installed extensions. The only issues currently are that it doesn't know how to handle incorrectly programmed extensions and what happens when you export a DB and import it in a NewPipe that doesn't have the same extensions. I'll hopefully be able to fix those tomorrow.

opusforlife2:
I was already testing. XD I'll just install your new one over it.

opusforlife2:
Werks. Both Bandcamp and Craptube work (if the only thing craptube does is change all newly fetched video titles to CRAP).

wb9688:
@opusforlife2: [Here](https://github.com/TeamNewPipe/NewPipe/files/5022908/newpipe-service-extensions.zip) is a better version, but please uninstall the previous version first. It contains the following extensions in the .zip:

- `bandcamp-tnp.jar`: It's signed by Team NewPipe's key and should work by default.
- `bandcamp.zip`: It's not signed, so should never work.
- `bandcamp.jar`: It's signed by me, so it should work after adding the fingerprint that's in `fingerprint.txt`, unless you have `bandcamp-tnp.jar` installed.
- `craptube-name.jar`: The extension name is different from the service name, so it should fail after having accepted installation.
- `craptube.jar`: It now has even more `CRAP`.

Please keep this version installed and add subscriptions and remote playlists from various services. When I post the next version, it will have a different DB scheme, so the migration for that should be tested.

opusforlife2:
- [x] bandcamp-tnp.jar: It's signed by Team NewPipe's key and should work by default.

Werks. Installing it over itself says it's an upgrade. Maybe it should say "it's the same version, wanna reinstall?"

- [x] bandcamp.zip: It's not signed, so should never work.

Doesn't.

- [x] bandcamp.jar: It's signed by me, so it should work after adding the fingerprint that's in fingerprint.txt, unless you have bandcamp-tnp.jar installed.

Werks. And either one doesn't install when the other is already installed. This, too installs over itself in the same way (upgrade).

About your fingerprint: there should be a way to select that text file just like with the JARs, so that users don't have to do anything different.

Also, the Team Newpipe default signature the apk comes with is removable. Maybe you want to grey out that entry and make it irremovable?

- [x] craptube-name.jar: The extension name is different from the service name, so it should fail after having accepted installation.

Fails.

- [x] craptube.jar: It now has even more CRAP.

Werks. Now with 2x more crap!

wb9688:
@yashx: You might be interested in this. If you have any issues, please let me know. However you should know that this is still a work in progress.

theScrabi:
Uff I somehow have a weird feeling about this. This might allow to add any kind of code to the app even harmful one.
Maybe you should think about adding some sort of signature to the zip so an addon can at least be checked as valid or not by the main app itself.

Also please remember to add a service color to the about.json. As every service can change the color of the NewPipe ui.

TheAssassin:
This doesn't seem like the best idea in an app like NewPipe. You have to be extremely careful, this is security sensitive. It will allow arbitrary code to be run within NewPipe (i.e., NewPipe might easily be abused for spying on the users, e.g., by injecting modules through exploitable bugs). Modules might crash the entire app, flooding us with bug reports which we don't want to even see, etc. etc. I don't think it's a good idea to have such a system inside this app.

TobiGr:
We could start a new VM for the loaded extensions and enable a security manager for it. That should solve the problem.

PR #3948, opened 2020-07-27
Add basic resize functionality

TobiGr:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)
- [ ] Code base improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
Allow resizing of app when OS + device support it.

#### Fixes the following issue(s)
Closes #3947
See https://codeberg.org/gitnex/GitNex/commit/9e352df1ed7d1d285fde61632822e2db5d128f16
See https://developer.samsung.com/samsung-dex/modify-optimizing.html#Enabling-Multi-Window-support

#### Testing apk
[NewPipe_basic_resize_APK.zip](https://github.com/TeamNewPipe/NewPipe/files/5017065/app-debug-1.zip)

#### Agreement
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

@mqus @Domiuns @MD77MD Can you please test if the app works properly when changing the orientation of your device, opening the keyboard when typing something in the searchbar and in dialogs (e.g. when creating a new feed group)?

PR #4080, opened 2020-08-05
Reduced CPU usage when playing a video by 7-10%

avently:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

After reading this comment https://github.com/TeamNewPipe/NewPipe/pull/3948#issuecomment-668241710  
> this new unified player ... drains the bettery twice as much

I checked how much CPU the player uses. And found out that @MD77MD said incorrect information but even if it's not true (about double difference) the new player definitely uses more CPU than 0.19.8. Difference is about 15%. So I started seaching for the cause and found that this line https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L1774
adds 7-10% of CPU load. Actually there is no need to animate the progress bar when it's hidden under the video but it should be updated anyway (without animation) to reflect playing position changes. 

Next. After this change the difference is still here but small: around 5%. Please, note, that I tested it on 5-year old phone while it had 1% of battery and worked on a charger with screen recorder impact so absolute values in your case would be less as well as difference.

I couldn't find a cause until I realized that MainVideoPlayer (old video player in 0.19.8) doesn't use notification. So it doesn't need to update it twice every second. This was it! After commenting out everything after this line: https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java#L640   the CPU usage was the same compared to 0.19.8 (0-1% difference).

Here is a screenrecord of comparison (with commented code related to notification) of new and old apps:
https://www.dropbox.com/s/l2sdo5rsjfmjusc/2020-08-05-12-01-09.mp4?dl=0

Of course this shouldn't be commented out in the current codebase but it gives us information about why this little difference exists: it's needed to update notification. 

#### Testing apk
[test-apk.zip](https://github.com/TeamNewPipe/NewPipe/files/5027736/test-apk.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -1769,10 +1769,11 @@ public class VideoDetailFragment
     private void showPlaybackProgress(final long progress, final long duration) {
         final int progressSeconds = (int) TimeUnit.MILLISECONDS.toSeconds(progress);
         final int durationSeconds = (int) TimeUnit.MILLISECONDS.toSeconds(duration);
+        // If the old and the new progress values have a big difference then use
+        // animation. Otherwise don't because it affects CPU
+        final boolean shouldAnimate = Math.abs(positionView.getProgress() - progressSeconds) > 2;
         positionView.setMax(durationSeconds);
-        // If there is no player inside fragment use animation, otherwise don't because
-        // it affects CPU
-        if (playerPlaceholder.getChildCount() == 0) {
+        if (shouldAnimate) {
             positionView.setProgressAnimated(progressSeconds);
         } else {
             positionView.setProgress(progressSeconds);
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -202,6 +202,7 @@ public class VideoDetailFragment
     private ImageView thumbnailImageView;
     private ImageView thumbnailPlayButton;
     private AnimatedProgressBar positionView;
+    private ViewGroup playerPlaceholder;
 
     private View videoTitleRoot;
     private TextView videoTitleTextView;
@@ -705,6 +706,7 @@ public class VideoDetailFragment
         thumbnailBackgroundButton = rootView.findViewById(R.id.detail_thumbnail_root_layout);
         thumbnailImageView = rootView.findViewById(R.id.detail_thumbnail_image_view);
         thumbnailPlayButton = rootView.findViewById(R.id.detail_thumbnail_play_button);
+        playerPlaceholder = rootView.findViewById(R.id.player_placeholder);
 
         contentRootLayoutHiding = rootView.findViewById(R.id.detail_content_root_hiding);
 
@@ -1265,17 +1267,15 @@ public class VideoDetailFragment
             return;
         }
 
-        final FrameLayout viewHolder = getView().findViewById(R.id.player_placeholder);
-
         // Check if viewHolder already contains a child
-        if (player.getRootView().getParent() != viewHolder) {
+        if (player.getRootView().getParent() != playerPlaceholder) {
             removeVideoPlayerView();
         }
         setHeightThumbnail();
 
         // Prevent from re-adding a view multiple times
         if (player.getRootView().getParent() == null) {
-            viewHolder.addView(player.getRootView());
+            playerPlaceholder.addView(player.getRootView());
         }
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1771,8 +1770,17 @@ private void showPlaybackProgress(final long progress, final long duration) {
         final int progressSeconds = (int) TimeUnit.MILLISECONDS.toSeconds(progress);
         final int durationSeconds = (int) TimeUnit.MILLISECONDS.toSeconds(duration);
         positionView.setMax(durationSeconds);
-        positionView.setProgressAnimated(progressSeconds);
-        detailPositionView.setText(Localization.getDurationString(progressSeconds));
+        // If there is no player inside fragment use animation, otherwise don't because
+        // it affects CPU
```
wb9688, line 1774: I prefer line breaks at a logical place, i.e. start the next line at "otherwise". That's not really an issue though (so don't waste time changing it unless you happen to need to change something else in this PR as well). I'll take an actual look at the code later today (in like 2 hours).

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1771,8 +1770,17 @@ private void showPlaybackProgress(final long progress, final long duration) {
         final int progressSeconds = (int) TimeUnit.MILLISECONDS.toSeconds(progress);
         final int durationSeconds = (int) TimeUnit.MILLISECONDS.toSeconds(duration);
         positionView.setMax(durationSeconds);
-        positionView.setProgressAnimated(progressSeconds);
-        detailPositionView.setText(Localization.getDurationString(progressSeconds));
+        // If there is no player inside fragment use animation, otherwise don't because
```
wb9688, line 1773: So I took a look, but I don't understand in what case there would be no player in the fragment. I also think we shouldn't update the progress in the player when the controls aren't being shown.

avently:
Start playing video in fragment, press popup button, voilà - player is not in fragment and this progress bar still shows current position.
Another situation: play video, close it via notification. Same thing. 
Same with background playback. So this red progress bar at the bottom of thumbnail and it should be in sync with playback position from player.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
avently: Start playing video in fragment, press popup button, voilà - player is not in fragment and this progress bar still shows current position.
Another situation: play video, close it via notification. Same thing.
Same with background playback. So this red progress bar at the bottom of thumbnail and it should be in sync with playback position from player.

MD77MD:
@avently 
I have been actively comparing between newPipe 19.8 and new player... and found it's more like 30% (not double). however this was on my Samsung s6, if you think it might be different on S4 I'll do a test.. let me know if you need me to.

another thing, I wanted to ask your thought on an idea and weather I sould open a request for it.. now the player starts loading comments then starts to play video (thus introducing some waiting depending on internet speed)...I was wandering if we can make the player start playing the video first while the comments and related videos sections can catch-up... please  let me know if I should open a request to discuss this.

avently:
> I have been actively comparing between newPipe 19.8 and new player... and found it's more like 30% (not double

Did you compare test apk from this PR? Is there any improvements? I would like to see real numbers instead of "I think it's around 30% in difference". I used `Simple System Monitor` for showing popup with graph of cpu usage on the same video with the same quality. Compare only CPU, it the only thing matters because your difference in battery life can be result of a difference in brightness, quality, video format, etc. The only thing I could do bad is to make computation harder but as showed in the screenrecord, the only difference makes worse CPU usage is updating notification which can not be removed. I can send a test apk with disabled notification updating just to ensure that you will get near equals result with 0.19.8 (it didn't have notification so this process didn't add a cpu pressure)

> now the player starts loading comments then starts to play video

I'm not sure you will get any improvements in terms of start time of video loading because amount of comments is so small. And skipping comments from initial loading will probably require additional call to the website. But it's only my expectations I don't know what happens in extractor under the hood. So you can open an issue and to discuss with guys who made the extractor.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
wb9688: Ah, I didn't think of that, but why would we still want an animation in those cases?

avently: It is how it worked before me. And I think it looks cool. Open a video that you watched before and stopped on 50% of total duration. You'll see an animation. Isn't it cool?
Why do you want to remove it?

wb9688: Ah, yes, that's cool, but wom't this mean the additional CPU usage will still be there for e.g. the background player?

avently: Made better logic in this place. Now it's smart:)

Issue #4098, opened 2020-08-08
The timestamp shouldn't be included when sharing an ongoing livestream


MrHeating:
# Version
0.19.8

### Steps to reproduce the bug
1. Go to an active livestream
2. Share to anywhere from the player

### Expected behavior
The time shouldn't be shared. Not only is it unnecessary, but NewPipe's time doesn't correspond to anything once the stream is >30s long

### Actual behaviour
Time is shared

nmurali94:
Hey! First time contributor here, can I take this one?

wb9688:
@nmurali94: Sure.

Also join #newpipe on Freenode if you have any questions regarding NewPipe development. (If you don't know about IRC, we suggest using Matrix through app.element.io and then joining #freenode_#newpipe:matrix.org)

PR #4115, opened 2020-08-11
Remove timestamp from url when sharing a live stream

nmurali94:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

- Remove timestamp from url when sharing a live stream

- https://github.com/TeamNewPipe/NewPipe/issues/4098

#### Testing apk
Verified on Pixel 1

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
@@ -880,9 +880,11 @@ public class VideoPlayerImpl extends VideoPlayer
 
     private void onShareClicked() {
         // share video at the current time (youtube.com/watch?v=ID&t=SECONDS)
+        // Timestamp doesn't make sense in a live stream so drop it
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
```diff
@@ -880,9 +880,11 @@ private void onMoreOptionsClicked() {
 
     private void onShareClicked() {
         // share video at the current time (youtube.com/watch?v=ID&t=SECONDS)
+        // Timestamp doesn't make sense in a live stream so drop it
+        String timestamp = !isLive() ? "&t=" + getPlaybackSeekBar().getProgress() / 1000 : "";
```
wb9688, line 884: Could you make this variable `final`? Could you also reverse the ternary statement so that it isn't negative, i.e. `isLive ? "" : "&t=" + …`?

Thanks for your contribution.

nmurali94: Done!

PR #3178, opened 2020-03-03
Notification Improvements

cool-student:
- add MediaStyle notifications for Background and Popup playback
- reduce excessive notification updating ( / recreating of Notification.Builder object) when playing background / popup media
- add new buffering state indicator (can be disabled)
- upscale close icon / downscale replay icon
- add notification slot settings
- move notification settings to appearance
- fix Metadata (song title, artist and album art) sometimes not being set correctly
- other misc notification fixes

------------------------------

debug apk (@Stypox 22/09/2020 b4e073cde7d737ea6c6230ecb17f47acac1da0b8): [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5262867/app-debug.zip)

closes #690 closes #1660 closes #2845 closes #1104 closes #1207 closes #3149 closes #3170
closes #1527 closes #1242 closes #1568 closes #1574 closes #1596 closes #2007
closes #3436 closes #3071 closes #4031 closes #4033 closes #2976 closes #2397 closes #228
closes #4286

- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #4130, opened 2020-08-15
[regression] Close player in onPlaybackShutdown()

### LINKED ISSUE #4129: [unified player] Swiping away the last stream in the queue does not close player
### Version

- Apk in the regression kanban: d2e6700dd1f746ba5aba59d39a4f0f53d56ec89e (2020-08-04 13:18 UTC)

1. Play a (single) stream in the background
2. Open the background play queue through the notification
3. Swipe away the stream

### Expected behavior
The player should stop and close

### Actual behaviour
The stream keeps playing as if nothing happened, controls are still working, but there is no stream in the play queue.

### Analysis
When the only stream playing is removed, [player.onPlaybackShutdown() is called](https://github.com/TeamNewPipe/NewPipe/blob/d1d942f3fdd6c803502d75eaf8a3dd1b778f025b/app/src/main/java/org/schabi/newpipe/player/playback/MediaSourceManager.java#L221). But [that function](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java#L621-L626) does absolutely nothing, and there is even a comment asserting that it should do nothing:
```java
    @Override
    public void onPlaybackShutdown() {
        if (DEBUG) {
            Log.d(TAG, "onPlaybackShutdown() called");
        }
        // Override it because we don't want playerImpl destroyed
    }
```
@avently why is this the case? Shouldn't that function be removed alltoghether if it does nothing? Anyway,  `onPlaybackShutdown()` is used in the situation discussed above and when an unrecoverable error occours. Is it safe to have that function to what it ought to, i.e. shutting down everything?

Stypox:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

See #4129. Since this is such a simple change I think it should be merged asap, as to make testing #3178 simpler
@avently is this ok?

Fixes #4129 

#### Testing apk
@opusforlife2 you were encountering this problem in #3178, could you check if it is fixed in this apk?
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5078760/app-debug.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
@@ -1068,14 +1068,6 @@ public abstract class BasePlayer implements
         registerView();
     }
 
-    @Override
-    public void onPlaybackShutdown() {
-        if (DEBUG) {
-            Log.d(TAG, "Shutting down...");
-        }
-        destroy();
-    }
-
     /*//////////////////////////////////////////////////////////////////////////
     // General Player
     //////////////////////////////////////////////////////////////////////////*/
--- a/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
@@ -622,7 +622,7 @@ public class VideoPlayerImpl extends VideoPlayer
         if (DEBUG) {
             Log.d(TAG, "onPlaybackShutdown() called");
         }
-        // Override it because we don't want playerImpl destroyed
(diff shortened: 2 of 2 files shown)


avently:
> @avently is this ok?

Looks ok, I tested and it works correctly even when stream completes. This indeed smarter solution than mine because `onPlaybackShutdown` called only in two places: `onPlayQueueChanged` without any items in queue and in case of undetermined error. So yes, it's cool.

opusforlife2:
Yup. Fixed.

PR #4138, opened 2020-08-16
Add checkstyle rule to show final local variable violations as warning

XiangRongLin:
#### What is it?

Add checkstyle rule to check for final local variable violations. 
Since fixing everything at once would be very bothersome, it is shown as warning and not error. Instead it can be gradually fixed when someone touches the files.
The reason i want it is, that on every PR one has to manually search for the violations.

Doc: https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/FinalLocalVariableCheck.html

#### Agreement


checkstyle.xml
```diff
@@ -154,6 +154,11 @@
         <module name="MultipleVariableDeclarations"/>
         <module name="SimplifyBooleanExpression"/>
         <module name="SimplifyBooleanReturn"/>
+        <module name="FinalLocalVariable">
+            <property name="tokens" value="VARIABLE_DEF,PARAMETER_DEF"/>
```
wb9688, line 158: Remove this line, it's pointless, as we require all parameters to be final with that other rule anyway.

wb9688: Hmm… it seems to make a significant difference in amount of warnings

TobiGr:
Happy rebasing to everyone!?

wb9688:
@TobiGr: I guess so… ¯\_(ツ)_/¯

PR #4154, opened 2020-08-17
Prevent jumping of the player and wrong padding on devices with cutout

### LINKED ISSUE #4040: Content not centered with 18:9 screen -> weird behaviour with navigation bar
### Version

- d8b5549fd98e43f18b993c5f36f2385cf9a80620 

1. Look for a vertical video, eg [this one](https://www.youtube.com/watch?v=KRntP-q_R9s)
2. Launch fullscreen

### Expected behavior

It is centered, and there the video doesn't "jump" to give space for navigation bar
### Actual behaviour

It's not centered, then when you click, the video goes up so that there is space for navigation bar

0.19.8 
0 19 8_nav_bar_1 0 19 8_nav_bar_2

unified
unified_nav_bar_1 unified_nav_bar_2

gif for this "jump" with unified

avently:
#### What is it?
- [x] Bug fix (user facing)

On devices with cutout there is a "jump" when you are in multiWindow mode or when you're watching vertical video in fullscreen. This PR fixes that jump (the first commit).

On the same devices there is a problem related to top padding on vertical videos in portrait in fullscreen. This is more interesting because a couple hours ago I thought I will not be able to fix it. This problem makes situation when the center of the player is not in the center of the screen. After hours trying I installed the app on my phone and the issue is fixed actually (same Android version but different result). So I have no idea will my method (from the second commit) work on user's devices or not. I think the issue is still here but on devices with non-standard screen ratio. But I don't know how to fix it. You can merge the PR as is or merge only first commit, or drop it.

In short:
- in multiwindow mode all videos work great
- there is no jump that happened before when a user shows/hides controls in the player
- there is a correct padding on top of the player (but I'm sure it isn't on devices with non-standart ratio).

Everyone who has a device with a hole, cutout, whatever check this build in vertical videos in landscape, in portrait, in multi window mode, check non-vertical videos too.
 
partially fixes #4040

#### Testing apk
[app-release-final.zip](https://github.com/TeamNewPipe/NewPipe/files/5085735/app-release-final.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -1982,6 +1982,11 @@ public class VideoDetailFragment
             return;
         }
 
+        // Prevent jumping of the player on devices with cutout
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            activity.getWindow().getAttributes().layoutInDisplayCutoutMode =
+                    WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_DEFAULT;
+        }
         activity.getWindow().getDecorView().setSystemUiVisibility(0);
         activity.getWindow().clearFlags(WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS);
     }
@@ -1995,6 +2000,11 @@ public class VideoDetailFragment
             return;
         }
 
+        // Prevent jumping of the player on devices with cutout
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            activity.getWindow().getAttributes().layoutInDisplayCutoutMode =
+                    WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_NEVER;
+        }
         final int visibility = View.SYSTEM_UI_FLAG_LAYOUT_STABLE
                 | View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN
                 | View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION
--- a/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
@@ -1538,8 +1538,9 @@ public class VideoPlayerImpl extends VideoPlayer
                 && navBarAtTheBottom ? size.y : ViewGroup.LayoutParams.MATCH_PARENT;
         controlsRoot.requestLayout();
 
+        final DisplayMetrics metrics = getRootView().getResources().getDisplayMetrics();
         int topPadding = isFullscreen && !isInMultiWindow() ? getStatusBarHeight() : 0;
-        topPadding = !isLandscape && hasCutout(topPadding) ? 0 : topPadding;
+        topPadding = !isLandscape && DeviceUtils.hasCutout(topPadding, metrics) ? 0 : topPadding;
         getRootView().findViewById(R.id.playbackWindowRoot).setTranslationY(topPadding);
         getBottomControlsRoot().setTranslationY(-topPadding);
     }
@@ -1567,20 +1568,6 @@ public class VideoPlayerImpl extends VideoPlayer
                     TypedValue.COMPLEX_UNIT_DIP, 24, metrics);
         }
         return statusBarHeight;
-    }
-
-    /*
-    * Compares current status bar height with default status bar height in Android and decides,
-    * does the device has cutout or not
-    * */
-    private boolean hasCutout(final float statusBarHeight) {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-            final DisplayMetrics metrics = getRootView().getResources().getDisplayMetrics();
-            final float defaultStatusBarHeight = TypedValue.applyDimension(
-                    TypedValue.COMPLEX_UNIT_DIP, 25, metrics);
-            return statusBarHeight > defaultStatusBarHeight;
-        }
-        return false;
     }
 
     protected void setMuteButton(final ImageButton button, final boolean isMuted) {
--- a/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
@@ -915,8 +915,8 @@ public class VideoPlayerImpl extends VideoPlayer
         // show kodi button if it supports the current service and it is enabled in settings
         final boolean showKodiButton = playQueue != null && playQueue.getItem() != null
                 && KoreUtil.isServiceSupportedByKore(playQueue.getItem().getServiceId())
-                        && PreferenceManager.getDefaultSharedPreferences(context)
-                        .getBoolean(context.getString(R.string.show_play_with_kodi_key), false);
+                && PreferenceManager.getDefaultSharedPreferences(context)
+                .getBoolean(context.getString(R.string.show_play_with_kodi_key), false);
         playWithKodi.setVisibility(videoPlayerSelected() && kodiEnabled && showKodiButton
                 ? View.VISIBLE : View.GONE);
     }
@@ -1499,9 +1499,10 @@ public class VideoPlayerImpl extends VideoPlayer
         // It doesn't include NavigationBar, notches, etc.
         display.getSize(size);
 
+        final boolean isLandscape = service.isLandscape();
         final int width = isFullscreen
-                ? (service.isLandscape()
-                ? size.x : size.y) : ViewGroup.LayoutParams.MATCH_PARENT;
+                ? (isLandscape ? size.x : size.y)
+                : ViewGroup.LayoutParams.MATCH_PARENT;
         final int gravity = isFullscreen
                 ? (display.getRotation() == Surface.ROTATION_90
                 ? Gravity.START : Gravity.END)
@@ -1532,14 +1533,15 @@ public class VideoPlayerImpl extends VideoPlayer
         // And the situations when we need to set custom height is
         // in fullscreen mode in tablet in non-multiWindow mode or with vertical video.
         // Other than that MATCH_PARENT is good
-        final boolean navBarAtTheBottom = DeviceUtils.isTablet(service) || !service.isLandscape();
+        final boolean navBarAtTheBottom = DeviceUtils.isTablet(service) || !isLandscape;
         controlsRoot.getLayoutParams().height = isFullscreen && !isInMultiWindow()
                 && navBarAtTheBottom ? size.y : ViewGroup.LayoutParams.MATCH_PARENT;
         controlsRoot.requestLayout();
 
-        final int topPadding = isFullscreen && !isInMultiWindow() ? getStatusBarHeight() : 0;
-        getRootView().findViewById(R.id.playbackWindowRoot).setPadding(0, topPadding, 0, 0);
-        getRootView().findViewById(R.id.playbackWindowRoot).requestLayout();
+        int topPadding = isFullscreen && !isInMultiWindow() ? getStatusBarHeight() : 0;
+        topPadding = !isLandscape && hasCutout(topPadding) ? 0 : topPadding;
+        getRootView().findViewById(R.id.playbackWindowRoot).setTranslationY(topPadding);
+        getBottomControlsRoot().setTranslationY(-topPadding);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
```diff
@@ -1565,6 +1571,20 @@ private int getStatusBarHeight() {
...
+    private boolean hasCutout(final float statusBarHeight) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            final DisplayMetrics metrics = getRootView().getResources().getDisplayMetrics();
+            final float defaultStatusBarHeight = TypedValue.applyDimension(
+                    TypedValue.COMPLEX_UNIT_DIP, 25, metrics);
+            return statusBarHeight > defaultStatusBarHeight;
+        }
+        return false;
+    }
+
```
TobiGr, line 1587: That looks like it should be in DeviceUtils.

blackbox87:
> This problem makes situation when the center of the player is not in the center of the screen. After hours trying I installed the app on my phone and the issue is fixed actually (same Android version but different result). So I have no idea will my method (from the second commit) work on user's devices or not. I think the issue is still here but on devices with non-standard screen ratio. But I don't know how to fix it. You can merge the PR as is or merge only first commit, or drop it.

That commit actually makes the player off-center for me while using a Poco F1. It was perfect before the commit as it matched the official YouTube app, minus pinch to zoom.

YouTube:
YouTube

NewPipe v0.19.8:
NewPipe

NewPipe Cutout:
NewPipe_Cutout

blackbox87:
> This problem makes situation when the center of the player is not in the center of the screen. After hours trying I installed the app on my phone and the issue is fixed actually (same Android version but different result). So I have no idea will my method (from the second commit) work on user's devices or not. I think the issue is still here but on devices with non-standard screen ratio. But I don't know how to fix it. You can merge the PR as is or merge only first commit, or drop it.

That commit actually makes the player off-center for me while using a Poco F1. It was perfect before the commit as it was almost identical to the official YouTube app, minus pinch to zoom.

Aligning to the notch means you'll probably lose some of the image. Like in my case it's now off-center and I lose some of the image on the right side of my screen due to the phones rounded corners.

YouTube:
YouTube

NewPipe v0.19.8:
NewPipe

NewPipe Cutout:
NewPipe_Cutout

avently:
@blackbox87 
Thank you for testing. Actually we need to compare the apk from this PR with the apk from this post: https://github.com/TeamNewPipe/NewPipe/projects/17#card-42918515

You compared with 0.19.8 which doesn't make sense. In 0.19.8 the app uses default mechanism of handling the cutout, everything is done by Android. On the newest version there is no way to give this job to Android because of the need to use Fragment instead of an Activity. 

So, just compare with the latest apk from the link. More info about the problem the PR trying to fix can be found in #4040

blackbox87:
> You compared with 0.19.8 which doesn't make sense. In 0.19.8 the app uses default mechanism of handling the cutout, everything is done by Android. On the newest version there is no way to give this job to Android because of the need to use Fragment instead of an Activity.

Sorry about that. I haven't been testing debug versions for a while now, so I didn't think to compare it to a different debug build. But I can confirm that the builds by @B0pol all have the same issue. It's awful.

When a device is in landscape you're not supposed to align to the cutcout because they come in many different sizes. If you do then you'll likely end up with off-centered videos and only having rounded corners on one side of your screen. That's why YouTube, VLC & many other players don't do anything about the notch and always center the videos to fit.

blackbox87:
> You compared with 0.19.8 which doesn't make sense. In 0.19.8 the app uses default mechanism of handling the cutout, everything is done by Android. On the newest version there is no way to give this job to Android because of the need to use Fragment instead of an Activity.

Sorry about that. I haven't been testing debug versions for a while now, so I didn't think to compare it to a different debug build. But I can confirm that the builds by @B0pol all have the same issue. It's awful.

When a device is in landscape you're not supposed to align videos to the cutcout because cutcouts come in many different sizes. If you do then you'll likely end up with off-centered videos and only having rounded corners on one side of your screen. That's why YouTube, VLC & many other players don't do anything about the notch and always center the videos to fit.

blackbox87:
> You compared with 0.19.8 which doesn't make sense. In 0.19.8 the app uses default mechanism of handling the cutout, everything is done by Android. On the newest version there is no way to give this job to Android because of the need to use Fragment instead of an Activity.

Sorry about that. I haven't been testing debug versions for a while now, so I didn't think to compare it to a different debug build. But I can confirm that the builds by @B0pol all have the same issue. It's awful.

You're not supposed to align videos to the cutcout while using landscape because cutcouts come in many different sizes. If you do then you'll likely end up with off-centered videos and only having rounded corners on only one side of your screen. That's why YouTube, VLC & many other players don't do anything about the notch and always center the videos to fit.

blackbox87:
> You compared with 0.19.8 which doesn't make sense. In 0.19.8 the app uses default mechanism of handling the cutout, everything is done by Android. On the newest version there is no way to give this job to Android because of the need to use Fragment instead of an Activity.

Sorry about that. I haven't been testing debug versions for a while now, so I didn't think to compare it to a different debug build. But I can confirm that the builds by @B0pol all have the same issue. It's awful.

You're not supposed to align videos to the cutcout while using landscape because cutcouts come in many different sizes. If you do then you'll likely end up with off-centered videos and only having rounded corners on one side of your screen. That's why YouTube, VLC & many other players don't do anything about the notch and always center the videos to fit.

app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
avently: Force-pushed (which was a bad idea but anyway) a commit with a new place for this code

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1995,6 +2000,11 @@ private void hideSystemUi() {
             return;
         }
 
+        // Prevent jumping of the player on devices with cutout
```
blackbox87, line 2003: Change to the following and it fixes landscape too.
```java
        // Prevent jumping of the player on devices with cutout
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
            if (isLandscape()) {
                activity.getWindow().getAttributes().layoutInDisplayCutoutMode =
                        WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES;
            } else {
                activity.getWindow().getAttributes().layoutInDisplayCutoutMode =
                        WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_NEVER;
            }
        }
```

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1995,6 +2000,11 @@ private void hideSystemUi() {
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            activity.getWindow().getAttributes().layoutInDisplayCutoutMode =
+                    WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_NEVER;
+        }
```
blackbox87, line 2007: Change to the following and it fixes landscape too.
```java
        // Prevent jumping of the player on devices with cutout
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
            if (isLandscape()) {
                activity.getWindow().getAttributes().layoutInDisplayCutoutMode =
                        WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES;
            } else {
                activity.getWindow().getAttributes().layoutInDisplayCutoutMode =
                        WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_NEVER;
            }
        }
```

avently: I can do this but in this case a title and a description will be hidden under possible cutout from top right corner (in landscape orientation it will be in top left corner). I'm not sure about how much devices have such cutout but it's not an ideal choice anyway.

blackbox87: That's actually a separate issue that's caused by the status bar height and navbar height not being accounted for while viewing a video in landscape. So I'm currently trying to work out the best way to center the controls.

blackbox87: That's actually a separate issue that's caused by the status bar height (and navbar height?) not being accounted for while viewing a video in landscape. So I'm currently trying to work out the best way to center the controls.

avently: Good to know that someone cares about it. Not only landscape mode has the problem with the centering but vertical videos too.

blackbox87:
@avently [Check out my commit](https://github.com/blackbox87/NewPipe/commits/video-placement). It should give you perfectly centered videos and all of the controls will have the correct padding.

And although it's unrelated to this problem I noticed that screen rotation doesn't always behave correctly.

1. Enable auto-rotate and watch a video in landscape
2. Disable auto-rotate while watching the video in landscape

Previously NewPipe would return you to your default orientation, but now you'll remain stuck in landscape and NewPipe's resize button won't do anything no matter how many times you press on it. So as it seems buggy I made [this change](https://github.com/blackbox87/NewPipe/commits/auto-rotate-fix) which restores the original behaviour.

blackbox87:
@avently [Check out my commit](https://github.com/blackbox87/NewPipe/commits/video-placement). It should give you perfectly centered videos and all of the controls will have the correct padding.

And although it's unrelated to the bug reports mentioned here I noticed that screen rotation doesn't always behave correctly.

1. Enable auto-rotate and watch a video in landscape
2. Disable auto-rotate while watching the video in landscape

Previously NewPipe would return you to your default orientation, but now you'll remain stuck in landscape and NewPipe's resize button won't do anything no matter how many times you press on it. Maybe that's the intended behaviour now, but as it seems buggy to me I made [this change](https://github.com/blackbox87/NewPipe/commits/auto-rotate-fix) which restores the original behaviour.

avently:
@blackbox87 I didn't look closely on your changes but looks like the changes depend on this PR. And you say now everything is fine. So we can merge this PR and then to wait your PR, right? 

About autorotation issue: can you make a PR too?

blackbox87:
> So we can merge this PR and then to wait your PR, right?

It could be done that way or you could just copy my changes into your branch. I'm happy with whatever as long as this gets fixed.

> About autorotation issue: can you make a PR too?

I'd like to know if it's the intended behaviour or not first. Maybe @Stypox or @TobiGr can comment?

It's possible that they want NewPipe to remain in landscape mode, but having the button not do anything when you disable auto-rotate seems like a bug either way. It should always be able to toggle between portrait and landscape.

avently:
@blackbox87 I prefer to maintain only my code. So better to have your code inside your PR.

About autorotation: this is my code that makes such situation so if your solution fixes it just make a PR. For me it was like a rare case (when you lock orientation while in landscape) and I just skipped that case

blackbox87:
> @blackbox87 I prefer to maintain only my code. So better to have your code inside your PR.

Fair enough. If someone merges this in then I'll update my fork and create a pull request.

> About autorotation: this is my code that makes such situation so if your solution fixes it just make a PR. For me it was like a rare case (when you lock orientation while in landscape) and I just skipped that case

Well there's two ways to fix the issue.

1. Set the orientation to portrait if auto-rotate is disabled and certain conditions are met.
2. Fix the resize button so that you can always toggle between portrait and landscape.

I'm not familiar with the code, which is why I went with the first option and made NewPipe behave as it used to. But I think fixing the button might actually be the better solution.

avently:
@blackbox87 second way is not an option at all. So, yeah, the first way is what we need.

Stypox (approved):
Code looks good, though I don't have a device to test this on. But if @blackbox87 says he has a fully working solution based on this one I'd merge right away, also to prevent rebasing issues. Thank you @avently ;-) 

@blackbox87 thank you for your investigations, as soon as this PR will get merged feel free to create two separate PRs for cutout and rotation. I think your rotation approach is right, obviously having a button doing nothing is a bad idea and the previous behaviour was ok. Thanks :-D

blackbox87:
@Stypox 
> I think your rotation approach is right, obviously having a button doing nothing is a bad idea and the previous behaviour was ok.

I've actually figured out how to fix the button, although @avently said that it wouldn't be an option. So we could have NewPipe remain in landscape and only change to portrait when the fullscreen button is pressed again. Personally I think that might be the better solution too.

wiidev (approved):

Stypox:
@blackbox87 yeah, do as you said ;-)

B0pol:
thank you

avently:
> thank you

You're welcome. Can wait to see a fix for centering from @blackbox87

PR #4224, opened 2020-09-03
Retain dialogs on screen rotation - 'What's New' help option, and chan…

nmurali94:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

- Prevent popups from being dismissed on screen rotation like  'What's New' help option and others.
These are the first steps towards consolidating how some of these popups are created. The current implementation works for items in regular streams, channels and subscriptions.

#### Testing apk

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5171910/app-debug.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
@@ -1,54 +1,85 @@
 package org.schabi.newpipe.info_list;
 
-import android.app.Activity;
 import android.app.AlertDialog;
-import android.content.DialogInterface;
+import android.app.Dialog;
+import android.content.DialogInterface.OnClickListener;
+import android.os.Bundle;
 import android.view.View;
 import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.fragment.app.DialogFragment;
+import androidx.fragment.app.FragmentManager;
 
 import org.schabi.newpipe.R;
-import org.schabi.newpipe.extractor.stream.StreamInfoItem;
 
-public class InfoItemDialog {
-    private final AlertDialog dialog;
+import java.io.Serializable;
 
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
```diff
@@ -1,54 +1,86 @@
...
+public class InfoItemDialog  {
+
+    private final ItemDialogFragment idf;
+
+    public InfoItemDialog(@NonNull final String title,
+                          final String additionalInfo,
                           @NonNull final String[] commands,
-                          @NonNull final DialogInterface.OnClickListener actions) {
-        this(activity, commands, actions, info.getName(), info.getUploaderName());
+                          @NonNull final OnClickListener actions) {
```
nmurali94, line 26: These are the first steps towards consolidating how some of these popups are created. 
The current implementation works for items in regular streams, channels and subscriptions.

avently:
It's hard to imaging that this pr will ever be merged
By making retainedInstance you just ruined every fragment's lifecycle. So the fragments will not be recreated on configuration change. 
And maybe you can disagree, but there is no any problem with a situation when you have missing dialogs on orientation change. Nothing bad happens.

nmurali94:
I think it's always good if we can keep screen elements after rotation. By putting the popups in the dialog fragment we've got that even if we potentially lose some actions in the popup (the parent fragment is destroyed)

I can certainly understand that retainInstance isn't desirable, it was lazy solution. I'll spend some more time when I get the chance to see if I work around it.

Stypox:
@nmurali94 could you explain what you did technically in this PR? I am worried by what @avently said, but do not understand what it means ;-)

mhmdanas:
@nmurali94 I'm closing this PR, but please make a comment when you want to resume work on this so I would re-open it. Thanks!

PR #4246, opened 2020-09-07
Disabled preloading when switching streams

### LINKED ISSUE #4232: [unified player] stop current video from preloading when trying to play another one

i think it would be nice to stop playing the actual video when choosing to play another not after it's fully loaded this way we won't be looking at the comments section of a video and another is being played on the background.

i don't know maybe?

#### Additional context

nothing..

every other app does this so newpipe should as well? 
well i guess it's helpful because it's what everyone expects to happen !

avently:
#### What is it?
- [x] Feature (user facing)

When you play one video and then select another the player will stop loading the first stream. Which theoretically saves some megabytes while next video is loading.  So the first video will NOT stop playing but will stop loading more parts that will never be viewed because the second video shows up.

closes #4232 

#### Testing apk
[disabled-preloading.zip](https://github.com/TeamNewPipe/NewPipe/files/5184067/disabled-preloading.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -958,6 +958,9 @@ public class VideoDetailFragment
             return;
         }
         setInitialData(sid, videoUrl, title, queue);
+        if (player != null) {
+            player.disablePreloadingOfCurrentTrack();
+        }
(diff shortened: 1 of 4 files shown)


TobiGr (approved):
Thanks!
This PR does not only save data, but also speeds up switching streams :+1:

Stypox (approved):
Looks good to me, too, and by the test I made it works :-D

PR #4251, opened 2020-09-08
Drop LoadController to just rely on DefaultLoadControl of ExoPlayer

### LINKED ISSUE #3251: Video preloading
I think that video preloading is too short. I am not talking about fetching before the video even start but when video stop (lag) and start to load. Newpipe is trying to load video fast so it load 1 second or little more forward and than start to play video, but after 1 second or more the video stop (lag) again and that repeat and repeat. You maybe say that i should use lower quality, but i say that video can be pretty stable in this quaility and strenght of internet connection. Maybe it will really help if newpipe would preload more time, for example  10 seconds. Yeah, it would take more time but it would be worth it because video would be much more stable with some bigger preloaded video time. I am not some kind of expert, but i am just normal user and maybe I say total stupidities here, but I tried to play the same video, in the same resolution, format, fps, etc. in other youtube alternative called Gettube. I don't know if you know it, but that doesn't matter, and video playback was much smoother because when video stop (lag) app preloaded much more time than it again start to play. I suggest to make video preloading time bigger or at least add option to make it bigger.

If I say stupidity, please instruct me a little.

Thank you for your support
Jiří Král

Redirion:
#### What is it?
- [X] Bug fix (user facing)
- [ ] Feature (user facing)
- [X] Code base improvement (dev facing)

In the future I would prefer to rely on Google to find reasonable values for buffers as they do in their DefaultLoadControl.
The last change to buffer values NewPipe did was 2,5 Years ago with this commit: https://github.com/TeamNewPipe/NewPipe/commit/b4668367c6640986cd78f2ac06436520fddb9399
The commit aimed to speed up start of initial playback.
However as issues like #3251 show, on an unstable connection, this can lead to buffer under-runs.
On the other hand I don't think that users with a high speed connection will notice this change as the high speed connection will fill the 5 times longer start buffer duration quick anyway. This is also the most significant change by dropping NewPipe's LoadController. The minimum buffer is increased from 500ms to 2500ms.

Fixes #3251 

#### Testing apk
https://github.com/Redirion/NewPipe/releases/tag/removeloadcontroller

#### Agreement


opusforlife2:
@Redirion Doesn't this conflict with #4246?

avently:
@Redirion if you want to change some default values why not to change them without removing LoadController? If you think it's reasonable choice

Issue #4071, opened 2020-08-04
Video doesn't autoplay if shared to Newpipe


opusforlife2:
### Version

- dev branch after 2907

### Steps to reproduce the bug

1. Turn on Autoplay (on Wi-Fi or Always).
2. Swipe Newpipe away from Recents.
3. Share a link to Newpipe.

### Expected behavior

Video details open up and video starts playing automatically.
### Actual behaviour

Video details open up, but that's it. That's the same behaviour as if autoplay was off.

Found while investigating #2232.

ingingin:
Do you mean the video should start or autoplay doesn't work properly. That title is messy

opusforlife2:
@ingingin You're thinking of auto-**queuing** of the next stream, which is an option in settings and also a toggle on the video details page. It's incorrectly named, especially now that it conflicts with the new setting added by the unified player PR.

opusforlife2:
I think the toggle should be renamed to "Auto-play Next" or "Auto-play Next Stream". What do you think, @B0pol, @Stypox?

avently:
Actually the issue is a little bit wrong and here is why. There are couple of choices you can use right now when you share a video to NewPipe: Show Info, Video player, Popup Player, Background Player, Download. 
First option (show info) just do what it should do: it just shows the info. Auto play is not working here which is expected.
Video player option will autoplay the video if auto play is enabled globally in the app which is expected too.
Popup and background player will always play automatically which is how they should work too.

In order to enable auto play for `show info` this line should be added before this line: https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/RouterActivity.java#L459 

```java
intent.putExtra(VideoDetailFragment.AUTO_PLAY, true);
```

But it's not logical choice at all based on how options are named.

opusforlife2:
Well, yeah. Show info is incorrect if autoplay is on. I think the visibility of the Show Info option should change based on if autoplay is on or off.

With autoplay off it can continue to be:
> Show Info, Video player, Popup Player, Background Player, Download.

But with autoplay on it should be:

> Video player, Popup Player, Background Player, Download.

opusforlife2:
Ok, wait, I just tested with the Always Ask option in the latest debug apk. Both Show Info and Video Player have the exact same behaviour for autoplay off: they open the video details but don't play it. For autoplay on, both open video details, and the video additionally plays if Video Player was selected.

[...]

Issue #3477, opened 2020-04-24
Move to a single activity app


wb9688:
Currently we have some activities besides `MainActivity`, such as `AboutActivity` or `SettingsActivity`, while they should be fragments imho.

opusforlife2:
Would solve #4135?

opusforlife2:
Downloads as well.

wb9688:
@opusforlife2: Yes, but not worth it to move that into the single activity if we want to make a new UI soon-ish.

opusforlife2:
There's a new UI in the works? O_O

PR #4259, opened 2020-09-09
Add settings migration, remove "Detail page" option from share dialog and minimize to background by default

### LINKED ISSUE #4070: Fix video player option from preferred open action (open in fullscreen)
### Version

- https://github.com/TeamNewPipe/NewPipe/commit/b7f50c3e124365b75fa28df5af64f94f32b10ab2 (after unified player, #2907)

1. In audio-video settings, set "video player"
2. Open a video (from an external url?)

### Expected behavior
It open the video in fullscreen, either portrait (vertical video) or landscape mode (horizontal video).

### Actual behaviour
It open the video just like "Show info" would do.

### Additional note

"Video player" should be renamed to fullscreen ~ or something like that
People which have chosen "Video player" in settings should have fullscreen ~ preference selected

TobiGr:
#### What is it?
- [x] Bug fix (user facing)
- [x] Feature (user facing)

The difference between "detail page" and the video player is only the phone's orientation. Therefore, we decided to remove the open "Detail Page" option from the share menu. To handle entries from older versions, I added a migration system for the settings / shared preferences which is quite similar to the one used in for the DB.
With the new app workflow introduced by the unified player PR, it is possible to minimize the video player to the bottom while doing other stuff in the app. However, this does only apply to the main fragment. When a user opens the settings or the about screen, the player stops. That's bad UX. For this reason, I set the "Minimize on app switch" setting to "background" instead of "None" to allow seamless transitions within the app as well as when closing NewPipe.

Closes #4070

###### note: I did not have much time for testing. 
Please install the current dev version from the zip first and change the "Minimize on app switch" setting or keep it as it is to see if the migration works when updating with the other APK.

[apps.zip](https://github.com/TeamNewPipe/NewPipe/files/5214440/apps.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -39,6 +39,7 @@ import org.schabi.newpipe.extractor.exceptions.ExtractionException;
 import org.schabi.newpipe.extractor.playlist.PlaylistInfo;
 import org.schabi.newpipe.extractor.stream.StreamInfo;
 import org.schabi.newpipe.extractor.stream.VideoStream;
+import org.schabi.newpipe.player.helper.PlayerHelper;
 import org.schabi.newpipe.player.playqueue.ChannelPlayQueue;
 import org.schabi.newpipe.player.playqueue.PlayQueue;
 import org.schabi.newpipe.player.playqueue.PlaylistPlayQueue;
@@ -278,6 +279,7 @@ public class RouterActivity extends AppCompatActivity {
 
             handleChoice(choice.key);
 
+            // open future streams always like this one, because "always" button was used by user
             if (which == DialogInterface.BUTTON_POSITIVE) {
                 preferences.edit()
                         .putString(getString(R.string.preferred_open_action_key), choice.key)
@@ -377,23 +379,50 @@ public class RouterActivity extends AppCompatActivity {
         final boolean isExtAudioEnabled = preferences.getBoolean(
                 getString(R.string.use_external_audio_player_key), false);
 
-        returnList.add(new AdapterChoiceItem(getString(R.string.show_info_key),
-                getString(R.string.show_info),
-                resolveResourceIdFromAttr(context, R.attr.ic_info_outline)));
+        final AdapterChoiceItem videoPlayer = new AdapterChoiceItem(
+                getString(R.string.video_player_key), getString(R.string.video_player),
+                resolveResourceIdFromAttr(context, R.attr.ic_play_arrow));
+        final AdapterChoiceItem showInfo = new AdapterChoiceItem(
+                getString(R.string.show_info_key), getString(R.string.show_info),
+                resolveResourceIdFromAttr(context, R.attr.ic_info_outline));
+        final AdapterChoiceItem popupPlayer = new AdapterChoiceItem(
+                getString(R.string.popup_player_key), getString(R.string.popup_player),
+                resolveResourceIdFromAttr(context, R.attr.ic_popup));
+        final AdapterChoiceItem backgroundPlayer = new AdapterChoiceItem(
+                getString(R.string.background_player_key), getString(R.string.background_player),
+                resolveResourceIdFromAttr(context, R.attr.ic_headset));
 
-        if (capabilities.contains(VIDEO) && !(isExtVideoEnabled && linkType != LinkType.STREAM)) {
-            returnList.add(new AdapterChoiceItem(getString(R.string.video_player_key),
-                    getString(R.string.video_player),
-                    resolveResourceIdFromAttr(context, R.attr.ic_play_arrow)));
-            returnList.add(new AdapterChoiceItem(getString(R.string.popup_player_key),
-                    getString(R.string.popup_player),
-                    resolveResourceIdFromAttr(context, R.attr.ic_popup)));
-        }
+        if (linkType == LinkType.STREAM) {
+            if (isExtVideoEnabled) {
+                // show both "show info" and "video player", they are two different activities
+                returnList.add(showInfo);
+                returnList.add(videoPlayer);
+            } else if (capabilities.contains(VIDEO)
+                    && PlayerHelper.isAutoplayAllowedByUser(context)) {
+                // show only "video player" since the details activity will be opened and the video
+                // will be autoplayed there and "show info" would do the exact same thing
+                returnList.add(videoPlayer);
+            } else {
+                // show only "show info" if video player is not applicable or autoplay is disabled
+                returnList.add(showInfo);
+            }
 
-        if (capabilities.contains(AUDIO) && !(isExtAudioEnabled && linkType != LinkType.STREAM)) {
-            returnList.add(new AdapterChoiceItem(getString(R.string.background_player_key),
-                    getString(R.string.background_player),
-                    resolveResourceIdFromAttr(context, R.attr.ic_headset)));
+            if (capabilities.contains(VIDEO)) {
+                returnList.add(popupPlayer);
+            }
+            if (capabilities.contains(AUDIO)) {
+                returnList.add(backgroundPlayer);
+            }
+
+        } else {
+            returnList.add(showInfo);
+            if (capabilities.contains(VIDEO) && !isExtVideoEnabled) {
+                returnList.add(videoPlayer);
+                returnList.add(popupPlayer);
+            }
+            if (capabilities.contains(AUDIO) && !isExtAudioEnabled) {
+                returnList.add(backgroundPlayer);
+            }
         }
 
         returnList.add(new AdapterChoiceItem(getString(R.string.download_key),
--- a/app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
@@ -18,8 +18,22 @@ public final class SettingMigrations {
     /**
      * Version number for preferences. Must be incremented every time a migration is necessary.
      */
-    public static final int VERSION = 0;
+    public static final int VERSION = 1;
     private static SharedPreferences sp;
+
+    public static final Migration MIGRATION_0_1 = new Migration(0, 1) {
+        @Override
+        public void migrate(final Context context) {
+            // We changed the content of the dialog which opens when sharing a link to NewPipe
+            // by removing the "open detail page" option.
+            // Therefore, show the dialog once again to ensure users need to choose again and are
+            // aware of the changed dialog.
+            final SharedPreferences.Editor editor = sp.edit();
+            editor.putString(context.getString(R.string.preferred_open_action_key),
+                    context.getString(R.string.always_ask_open_action_key));
+            editor.apply();
+        }
+    };
 
     /**
      * List of all implemented migrations.
@@ -28,7 +42,7 @@ public final class SettingMigrations {
      * If not sorted correctly, migrations which depend on each other, may fail.
      */
     private static final Migration[] SETTING_MIGRATIONS = {
-
+        MIGRATION_0_1
     };
 
 
--- a/app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
@@ -18,7 +18,7 @@ public final class SettingMigrations {
     /**
      * Version number for preferences. Must be incremented every time a migration is necessary.
      */
-    public static final int VERSION = 1;
+    public static final int VERSION = 2;
     private static SharedPreferences sp;
 
     public static final Migration MIGRATION_0_1 = new Migration(0, 1) {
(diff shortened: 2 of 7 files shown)


opusforlife2:
https://github.com/TeamNewPipe/NewPipe/issues/4071#issuecomment-686742703

opusforlife2:
>  When a user opens the settings or the about screen, the player stops.

The proper solution to this is still #3477, right?

TobiGr:
yes, but that's out of scope for this version

MD77MD:
could this be done for this as well https://github.com/TeamNewPipe/NewPipe/pull/3178#issuecomment-689826763

app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
```diff
@@ -0,0 +1,126 @@
...
+import android.content.SharedPreferences;
+
+import androidx.preference.PreferenceManager;
+
+import org.schabi.newpipe.R;
+import org.schabi.newpipe.report.ErrorActivity;
+import org.schabi.newpipe.report.ErrorActivity.ErrorInfo;
+import org.schabi.newpipe.report.UserAction;
+
+public final class SettingMigrations {
```
Stypox, line 13: Not in this PR, but we may consider adding junit tests for migrations, using some kind of mocked shared preferences

Stypox (changes requested):
I tested updating from the dev apk to the other under many configurations and everything worked just fine. Thank you!

Stypox:
I was told of an inconsistency with the minimize on app switch options.
- tap and play a video, and wait for it to start playing
- press the back hardware button until the app is closed
If minimize is set to popup, a popup will start, but is set to background the app just closes. I think this should be fixed before releasing 0.20.0, it shouldn't be difficult, so I'll look into it tomorrow.

opusforlife2 (commented):
Reminder regarding the autoplay discussion yesterday: https://github.com/TeamNewPipe/NewPipe/issues/4071#issuecomment-689779956

opusforlife2:
Do you intend for this to be a temporary fix for
> When a user opens the settings or the about screen, the player stops. That's bad UX.

until #3477 is fixed or will this be a permanent change?

I think it might be better off as a permanent change. Minimizing is a feature Newpipe has that makes it superior to Youtube. Better that we alert the user that such a feature exists and let them turn it off manually, than to hide it and risk that most users don't even realise it is there.

xxxxxxxxxxxxxxxxx

> I set the "Minimize on app switch" setting to "background" instead of "None" to allow seamless transitions within the app as well as when closing NewPipe.

Should there be a toast the first 2-3 times this happens? A user who is new to Newpipe (New to the Pipe? 🤭) might get surprised or think it is a bug.

Something like "Minimizing to background/popup. This can be changed in settings."

If the user does change it in settings, this toast could be turned off immediately. Otherwise, after it has been shown twice or thrice.

MD77MD:
>If the user does change it in settings, this toast could be turned off immediately. Otherwise, after it has been shown twice or thrice.

I would leave it until changed in settings... no need to hide it

opusforlife2:
> I would leave it until changed in settings... no need to hide it

It would be very irritating to see the toast every time when you _know_ you can change it in settings but you _want_ to continue with the default setting. Having to change a setting, then change it back again to get rid of a toast is bad UX.

app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
```diff
@@ -0,0 +1,126 @@
...
+    };
+
+
+    public static void initMigrations(final Context context) {
+        // setup migrations and check if there is something to do
+        sp = PreferenceManager.getDefaultSharedPreferences(context);
+        final String lastPrefVersionKey = context.getString(R.string.last_used_preferences_version);
+        final int lastPrefVersion = sp.getInt(lastPrefVersionKey, 0);
+        if (lastPrefVersion == VERSION) { // no migration to run, already up to date
+            return;
```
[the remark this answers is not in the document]
TobiGr: Ah yes. I only focused on migrating old versions, but did not think about new releases.
Checking if there are any prefs seems like a good solution. I'll just need to check whether it is okay to move the call to injt thr migrations before the other stuff for the download died in done.

MD77MD:
@opusforlife2
sorry... i thought you meant to disable background playback after three times... I agree with you on the toasta

app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
TobiGr: Completely agree with that.

TobiGr:
Update APKs: [apps.zip](https://github.com/TeamNewPipe/NewPipe/files/5214440/apps.zip)

I only fixed https://github.com/TeamNewPipe/NewPipe/pull/4259#discussion_r486629567

app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
```diff
@@ -0,0 +1,131 @@
...
+        // setup migrations and check if there is something to do
+        sp = PreferenceManager.getDefaultSharedPreferences(context);
+        final String lastPrefVersionKey = context.getString(R.string.last_used_preferences_version);
+        final int lastPrefVersion = sp.getInt(lastPrefVersionKey, 0);
+
+        // no migration to run, already up to date
+        if (isFirstRun) {
+            sp.edit().putInt(lastPrefVersionKey, VERSION).apply();
+            return;
+        } else if (lastPrefVersion == VERSION) {
```
Stypox, line 75: Isn't VERSION equal to SETTINGS_MIGRATIONS[-1].newVersion? Even though fixing this would probably reduce redundancy, I am not sure if it should be done, since what you have done looks clearer

Stypox (commented):
Thanks, the first-time check looks good now

app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
TobiGr: That's correct, but as you said, this makes the code harder to read.

TobiGr:
Fixed a typo

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -377,10 +376,6 @@ private void showDialog(final List<AdapterChoiceItem> choices) {
         final boolean isExtAudioEnabled = preferences.getBoolean(
                 getString(R.string.use_external_audio_player_key), false);
 
-        returnList.add(new AdapterChoiceItem(getString(R.string.show_info_key),
-                getString(R.string.show_info),
-                resolveResourceIdFromAttr(context, R.attr.ic_info_outline)));
-
         if (capabilities.contains(VIDEO) && !(isExtVideoEnabled && linkType != LinkType.STREAM)) {
             returnList.add(new AdapterChoiceItem(getString(R.string.video_player_key),
                     getString(R.string.video_player),
```
Stypox, line 381: In order to achieve what was discussed in #4071, I'd just add a condition here that shows "R.string.show_info" if autoplay is off, but still uses "R.string.video_player_key" as the key in any case. At the end of the day both options open the same activity, even though the outcome is different, so I think users would expect "Show info" to be selected in the dialog if they had used the "Video player" choice before disabling autoplay, and viceversa. 

I don't think you would need to do anything else in order to achieve what I described above, since VideoDetailFragment already handles autoplay by itself.

Stypox (changes requested):
Please review lines 452 to 470 of RouterActivity.java, I think that piece of code can be deleted, as "show_info_key" is not a choice anymore.

opusforlife2:
> When chosen, auto-start the playback, even when the device is in portrait mode and auto-play is disabled.

Can you clarify this a little bit? If I have autoplay set to off, and Preferred open action is set to Video player (after Show info is removed) then will opening links from other apps autoplay the video?

Stypox:
@opusforlife2 I think commit messages are still the old ones, so don't look at them now ;-)

TobiGr:
I did not change anything regarding @Stypox's suggestion yet. Last pushs were just fixes to the migration process. In the first implementations, all migrations were run and not only the ones for newer pref versions.

Stypox:
> Remove "Detail Page" open action from share dialog under certain circumstances
> With the new application workflow and unified player, video detail page and video player are the same activity. So show only one of these options based on whether autoplay is enabled or not, and show both if using external player

New APKs with the changes quoted above: [apks.zip](https://github.com/TeamNewPipe/NewPipe/files/5264066/apks.zip)

Please test this, everything should work fine under many conditions:
- upgrading from `before.apk` to `after.apk`
- importing files from old newpipe versions
- sharing videos, playlists, channels to newpipe and the correct options turn up. In particular for videos, where the behaviour was changed, make sure the statements quoted above hold true.
@opusforlife2 is the new behaviour suitable?

opusforlife2:
`before.apk` is identical to the latest debug?

TobiGr (commented):
@Stypox Your changes look good :+1:

opusforlife2:
When sharing to 'after.apk', you first see a notification saying "Getting info - loading requested content", and only after it is loaded does the app open. This isn't mentioned in the changes above. Also, this delay might be perceived as a bug or Newpipe being slow by the user. Maybe a toast should be shown additionally?

> sharing videos, playlists, channels to newpipe and the correct options turn up in after.apk

Correct options do show up for videos, but what has changed for playlists or channels? 

---
Depending on the autoplay setting, the Always ask menu shows 'Show info' or 'Video player'. Shouldn't this also happen for the selection menu in settings? That menu always shows 'Video player', even if autoplay is off. Could be confusing or misleading for users.

Both minimize and preferred open action settings migrations happen correctly, for upgrading as well as importing. 👍

Stypox:
> This isn't mentioned in the changes above. 

That's always been there

> Correct options do show up for videos, but what has changed for playlists or channels? 

Both show info and video player should always be shown for channels and playlist, since they actually do different things

> Shouldn't this also happen for the selection menu in settings? 

No, since as I said above both info and player are shown for playlists and channels

> Both minimize and preferred open action settings migrations happen correctly, for upgrading as well as importing. 👍 

Great :-D

opusforlife2:
> That's always been there

Doesn't happen for me in either 0.19.8 or the latest unified debug. Only in this apk. Release and debug first open, then load the video details. This apk first loads video details, then opens with them already showing.

> Both show info and video player should always be shown for channels and playlist

Wow, before this I didn't even know it was possible to play a queue externally. :P My bad. I didn't select Always Ask before testing playlists and channels.

> No, since as I said above both info and player are shown for playlists and channels

I'm not talking about the Always Ask menu. I'm talking about the Preferred open action menu with radio buttons. It doesn't have Show Info when Autoplay is off, but still shows Video Player.

In any case, the radio buttons seem to be for the case of a single video link being opened from another app. I think we might need separate Preferred 'open' action settings for playlists and channels. Because the current setting isn't geared towards them. Playlists and channels being affected seems more like a byproduct of the setting.

Stypox:
> I think we might need separate Preferred 'open' action settings for playlists and channels

Yes

> I'm talking about the Preferred open action menu with radio buttons
> Doesn't happen for me in either 0.19.8 or the latest unified debug. Only in this apk. Release and debug first open, then load the video details. This apk first loads video details, then opens with them already showing.

Does this apk work as expected? I'm not providing before and after apks, since the only thing I did was restoring some things unrelated to migration [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5287261/app-debug.zip).

opusforlife2:
Oh no. The package is named org.schabi.newpipe.debug.pr, which is also the name of the notification PR apk. It installed over that one. xD

opusforlife2:
The Getting requested info notification is gone now, and the behaviour is as it was before. Is that what we want, though? I kind of liked the new behaviour. Didn't like the idea of a toast?

---
Now 'Preferred open action' shows both Show Info and Video player. That's not what I meant. When I set Autoplay to Off, I should see only Show Info. When I set it to On or Wifi, I should see only Video player. At least, that's what I thought up when I originally wrote the comment in that issue.

Stypox:
@opusforlife2 that menu for now also applies to channels and playlists, so both "show info" and "video player" should be there. In a future PR we could add separated settings, and at that point I'd agree with you. But for now both have to be kept.

Stypox:
> Didn't like the idea of a toast?

This PR has nothing to do with that notification, so that won't be changed here

opusforlife2:
Okay. Then selecting either Show info or Video player would mean the same thing for a video link, opening the video details page, right? Whether or not the video autoplays would then depend on that setting?

opusforlife2:
> This PR has nothing to do with that notification, so that won't be changed here

Where was it added? O.o

Stypox:
@TobiGr do you agree with what I said above? If so, this is ready ;-)

Stypox:
> Then selecting either Show info or Video player would mean the same thing for a video link, opening the video details page, right? Whether or not the video autoplays would then depend on that setting?

Yes, it would mean the same thing. I realised this myself the other day, but since this part of code (i.e. `RouterActivity`) is rather crappy, it would take too much time to make this work as expected. So I'd prefer to put it off to a future PR, containing a `RouterActivity` refactor and different open options for channel, playlist and video.

Issue #4285, opened 2020-09-16
Inconsistent player behaviour with both landscape and portrait videos in queue (unified player)


opusforlife2:
### Version

- https://github.com/TeamNewPipe/NewPipe/commit/530f745e44a95e74f3587c429805f147ced4e73c

### Steps to reproduce the bug

Enqueue a landscape video and a portrait video in the main player and try switching between them with various initial orientations and rotation locked/unlocked.

### Actual behaviour
The next video (or next video's frame, if paused) shows in the aspect ratio of the previous one, then after half a second it regains the correct aspect ratio.
Moreover...
- Rotation unlocked: No rotation occurs. True for landscape, portrait, and fullscreen portrait.
- Rotation locked in portrait:
  - Portrait (video details): no rotation on video change.
  - Landscape: playing portrait video next rotates phone to portrait (video details). After that, no rotation.
  - Fullscreen portrait: playing landscape video next rotates phone to landscape, which is good, but going back to the portrait video opens portrait video details instead of fullscreen portrait. After that it stays in portrait video details on further video switches, no rotation.
- Rotation locked in landscape:
  - Landscape: Landscape video plays fine. Playing portrait video next _doesn't_ rotate screen. The player stays in landscape on further video switches.
  - Still landscape: when on landscape video, fullscreen button does nothing (@blackbox87 turns out you just need rotation locked in landscape to trigger the bug your PR fixes, no need to toggle it while watching).      When on portrait video, fullscreen button opens video details in landscape, and video continues playing. Pausing and resuming the video opens fullscreen again, as does changing the video.
  - Portrait: nothing. There is no way to switch to portrait mode, even when video is portrait.

0.19.8 (for comparison) instantly shows the video in the correct aspect ratio, regardless of orientation, and doesn't rotate regardless of rotation lock status.

### Expected behaviour
- Once the fullscreen button has been tapped to go into fullscreen, then the player should stay in fullscreen until the button is tapped again to exit fullscreen. (or back button is pressed)
- When in fullscreen, the player should rotate to whatever orientation the video is, but this rotation should happen in fullscreen (no rotation back to video details).
- When viewing video details in portrait, tapping the video plays it right there. But when viewing video details in landscape, tapping the video plays it in fullscreen. This behaviour should be consistent for both orientations.

Thanks to @blackbox87 for notifying me about queue issues involving vertical videos.

Issue #4286, opened 2020-09-17
[Unified player] Heavy work on Main Thread during seeking on almost all videos


vkay94:
### Version

530f745 2020-09-08

### Steps to reproduce the bug
1. Open any video (99% videos have this bug)
2. Show the controller and change the position multiple times by dragging/tapping on the seekbar
3. Wait ...

### Expected behavior
The video should load shortly and start the playback and response to other actions.

### Actual behaviour
The seekbar doesn't update the position for a long time. During this time the UI doesn't react to gestures ("freeze"). After a while the app crashes.

### Screenshots/Screen recordings
Video demonstration: https://streamable.com/9gf4tv
The first video is the expected behavior, the second video shows the bug.

### What's the reason?
During debugging I saw in the logcat that the device tries to seek to all positions which were tapped, so the doesn't discard "unneeded" positions (positions which are older than the current).

vkay94:
I might have found a workaround for this ... kind of. It's not as smooth as the first video then (keep in mind that the video is recorded, on the device it's more fluid), but definitely faster responsive.

Just remove / comment out these lines: https://github.com/TeamNewPipe/NewPipe/blob/4b7311bafd919bdb9e6df8b8a865d14bc3908081/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java#L1094
https://github.com/TeamNewPipe/NewPipe/blob/4b7311bafd919bdb9e6df8b8a865d14bc3908081/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java#L1150

I don't see a reason for recreating the whole notification when the video is sought. This has a high impact on the main thread if it's occurs too often.

Stypox:
@vkay94 please check if the apk provided in #3178 works well

vkay94:
@Stypox It's slightly faster on the video which works well but on the others it's still freezing the UI/seekbar.

avently:
> It's slightly faster on the video which works well but on the others it's still freezing the UI/seekbar

Are you sure that the delay is caused by @Stypox's notifications? Please, try to measure using logcat timestamps or Android Profiler

vkay94:
@avently I think so. Probably not the actual recreation as I looked into the code of the pr (`forceRefresh`-parameter) but it has an impact. Maybe it's the source. I don't know :(

![image](https://user-images.githubusercontent.com/6626210/93513132-af659e00-f925-11ea-9308-0cce670afe3c.png)

[...]

PR #4254, opened 2020-09-08
Fix rotation when auto-rotate is disabled

blackbox87:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)
- [ ] Code base improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

1. Enable auto-rotate and watch a video in landscape
2. Disable auto-rotate while watching the video in landscape

Previously NewPipe would return you to your default orientation, but now you'll remain stuck in landscape and NewPipe's fullscreen button won't do anything no matter how many times you press on it.

With this change NewPipe will remain in landscape mode, but when you press on the fullscreen button it'll correctly change the orientation to portrait.

#### Testing apk
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5196379/app-debug.zip)

#### Agreement
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Issue #3753, opened 2020-06-06
ParsingException: Could not load decryption code for the Youtube service


eladkarako:
## Exception
* __User Action:__ requested stream
* __Request:__ https://www.youtube.com/watch?v=EeOxyMGe6H8
* __Content Language:__ GB
* __Service:__ YouTube
* __Version:__ 0.19.5
* __OS:__ Linux samsung/j4ltejx/j4lte:9/PPR1.180610.011/J400FXXU4BSJ1:user/release-keys 9 - 28
just scrolling the what's new section clicking through the video, going back, viewing another video, then going back and viewing the first one again. only happen sometimes. when going back and forth to the video it is usually loading fine. it looks like a sync between how fast the data arrives and maybe some sort of timeout protection in the code that isn't initialize properly on a new view (perhaps related to remembering the history of the last playing point) 

<details><summary><b>Crash log</b></summary><p>

```
org.schabi.newpipe.extractor.exceptions.ParsingException: Could not load decryption code for the Youtube service.
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.getPlayerUrl(YoutubeStreamExtractor.java:669)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.onFetchPage(YoutubeStreamExtractor.java:619)
	at org.schabi.newpipe.extractor.Extractor.fetchPage(Extractor.java:56)
	at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:68)
	at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:64)
	at org.schabi.newpipe.util.ExtractorHelper.lambda$getStreamInfo$3(ExtractorHelper.java:116)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$5fJcha6Sq5APJBLdG6osaJby-mc.call(Unknown Source:4)
	at io.reactivex.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:44)
	at io.reactivex.Single.subscribe(Single.java:3666)
	at io.reactivex.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
	at io.reactivex.Single.subscribe(Single.java:3666)
[...]

PR #4288, opened 2020-09-17
Performance increase

avently:
#### What is it?
- [x] Feature (user facing)

This PR for devs and for users.

### For devs:

I made optimizations for managing the player service. Now the player service is bound to static class and there is static listener for all events from the player service. VideoDetailFragment can subscribe/unsubscribe from events dynamically without the need to bind/unbind to the service. Fragment only need to subscribe/unsubscribe to/from events via static PlayerHolder. It gives possibility to skip binding process on configuration change and to speed up orientation change.

Also I discovered that `prepareAndHandleInfo` is terribly slow, something around 1 second on slow device. Whenever currently playing item in playlist changes this method is called and freezes main thread for a long time. Ideal choice would be to rewrite some methods that populates view groups but it's hard to do and time consuming task. Instead I call this method (`prepareAndHandleInfo`) after timeout which makes main thread free for some time which allows to start playback on next/prev item in playlists fast.

Also I found that SerializedCache clones every item in it which takes some time (100-200 ms). So I removed that part from fragment and saving playQueue, streamInfo and stack inside static variables.

### For users:
Faster loading time of streams, faster orientation change. 
Also when you have horizontal + vertical + horizontal videos in queue this is what happens:
1. In locked global orientation (portrait):
    - in non-fullscreen mode you play all videos tapping on `next arrow` button and orientation will be the same
    - in fullscreen mode you play tapping on `next arrow` button from landscape video to vertical will change orientation to portrait, then after vertical video will change to landscape again for landscape video.

2. In locked global orientation (landscape):
    - every video will be in landscape

3. In unlocked global orientation:
    - video will be in the same orientation you have your device rotated
    - if orientation is portrait and the video is vertical you will be able to play it in fullscreen (fullscreen button appears).
    - if orientation is landscape and you are watching landscape video it will be in fullscreen without an ability to make it non-fulscreen (you're able to rotate your device to portrait and the video will be in non-fullscreen mode)

If you read correctly you noticed that the player will rotate the video from landscape to portrait and back if you selecting vertical video after/before landscape with locked global orientation. It's useful thing but with a cost. And the cost is wrong size of the video for about 0.1-1 second based on performance of the device. This happens because the app still doing many things in main thread. Mostly because of notifications. Yes, they are terrible slow and nothing we can do in this area except to wait until this PR #3178 will be merged. I hope it fixes the notifications' performance but if not the delay is still acceptable. But even if the notifications will be fast, it doesn't help to reduce the delay to 0 because when orientation changes from one to another there is a delay related to reconstruction of the whole fragment. I made this process faster but the delay is still visible on slow devices.

Here we have two APKs: the first one is what this PR gives. The second one shows what it means to have all notifications disabled. So I just tested with disabled notifications and the app is soooooo fast when orientation changes and between videos in queue. Of course, we still need the notifications but it's just for educational purposes:)

Based on #4272

#### Testing apk
This PR: 
[performance.zip](https://github.com/TeamNewPipe/NewPipe/files/5241696/performance.zip)

This PR + disabled notifications (just for experiment)
[performance-without-notifications.zip](https://github.com/TeamNewPipe/NewPipe/files/5241698/performance-without-notifications.zip)

#### Agreement

This PR starts at https://github.com/TeamNewPipe/NewPipe/pull/4288/commits/5b8eda48053ddfc2f732cd7e0e793e7f636ea53b


vkay94:
Some feedback regarding the orientation flow:

- Orientation locked (portrait):
  - If the video is set to fullscreen by using the button: 
    - back press during active playback switches the video to portrait but the playback stops (intended?)
    - dragging the video to collapsed state keeps the device on landscape mode even if the orientation is set to portrait in the first place (the orientation should be set to portrait after the drag)
- Orientation locked (landscape)
  - the fullscreen button has no effect (should be set to `GONE`)

opusforlife2:
@avently Wow, dude! PR after PR. You're on fire!

> Faster loading time of streams, faster orientation change.

Quite a difference from the unified debug!

> Also when you have horizontal + vertical + horizontal videos in queue this is what happens:

Much better and more intuitive behaviour!

Since this PR (along with the notification PR, as you said) basically fixes #4285, shall I link it?

@vkay94 

> (intended?)

Yes. The back button pauses, the fullscreen button doesn't.

> (the orientation should be set to portrait after the drag)

Yeah, after the user lifts their finger, the app should go back to portrait.

> the fullscreen button has no effect (should be set to GONE)

Fixed in blackbox's rotation fix PR.

avently:
@vkay94
> back press during active playback switches the video to portrait but the playback stops (intended?)

Yep. So you have to options: you can press on fullscreen button and it will rotate a screen but not pause; you can press back button and it will rotate the screen and pause the playback. Something that I use often.

> dragging the video to collapsed state keeps the device on landscape mode even if the orientation is set to portrait in the first place (the orientation should be set to portrait after the drag)

If I made like you're saying it will completelly broke the use-case with swiping the player down and selecting another video without changing the orientation. So now you can browse other videos while current video is playing. If you want to switch to portrait you'll click on fullscreen button. So I just don't see any profits from making like you said. 

> Orientation locked (landscape) the fullscreen button has no effect

#4254 will make rotation possible while it's locked in landscape-only. Maybe it's better than hiding button, don't you think so? Since you can in portrait to swich to landscape via button than you should be able to do it from landscape to portrait. So I'm fine with the linked PR.

Thanks for the feedback!

@opusforlife2 
> Since this PR (along with the notification PR, as you said) basically fixes #4285, shall I link it?

You could if you think that the PR in it's current state fixes your issue.

vkay94:
>Yep.

Alright ;)

>If I made like you're saying it will completelly broke the use-case with swiping the player down and selecting another video without changing the orientation.

I understand the use-case and it makes sense. I wanted to say that the whole app may be locked in this orientation when the collapsed player is dismissed until a new video is opened and something like that could prevent it :) 

>#4254 will make rotation possible while it's locked in landscape-only

Nice ;) It just felt odd without functionality but nice to know it's already done :D

blackbox87:
> Fixed in blackbox's rotation fix PR.

Yeah. If you lock the rotation while watching a video then the button should remain on screen, but when you press the button it should revert you to your devices default orientation. That's what my PR does, although the behaviour between the back button and fullscreen/rotation button is different as one pauses the video and the other doesn't.

I often disable auto-rotate, so with the current code if I have a playlist that contains 5 vertical videos and 5 horizontal videos then I'll be flipped between portrait and landscape, which is annoying as they should all be played with my devices current locked orientation unless I specifically click on the button within NewPipe to change the orientation. And has anyone checked how an Android TV box handles this?

Personally, I prefer the way v0.19.8 handles orientation as it gives you greater control of the videos that you're watching. Like with these dev versions if you lock your devices orientation to portrait and try to watch a vertical video in landscape then you can't do it anymore due to the buttons being merged. So in my opinion it would be better to have 1 button that toggles fullscreen and a separate button that toggles the orientation.

avently:
> And has anyone checked how an Android TV box handles this?

Once vertical video starts after landscape in fullscreen the player exists from fullscreen but orientation is still landscape. I fixed it locally, so the player will stay in fullscreen on vertical video too. One line of code, will send with other changes if any of them be needed. 

> which is annoying as they should all be played with my devices current locked orientation unless I specifically click on a button within NewPipe to change the orientation.

I think it's the matter of taste. For me it's ok to see rotation because without that rotation I will need to do many different steps to make the vertical video in portrait & fullscreen since I don't want to see the vertical video in landscape. I always like to see how my code doing something for me automatically:)

If you have another use-case, you can make this behavior switchable from settings maybe it will be useful for others too.

> it would be better to have 1 button that toggles fullscreen and a separate button that toggles the orientation

Having two buttons when you can have just one is kind of unoptimized behavior and not good for UI and usabilty.

Why do you want to look at vertical videos in landscape?

avently:
@opusforlife2 
> Wow, dude! PR after PR. You're on fire!
> Quite a difference from the unified debug!

Do you see improvements in terms of performance of your device? Does the rotation and prev/next videos switching works better and how much better (on % maybe)?

blackbox87:
> I think it's the matter of taste. For me it's ok to see rotation because without that rotation I will need to do many different steps to make the vertical video in portrait & fullscreen since I don't want to see the vertical video in landscape. I always like to see how my code doing something for me automatically:)

I think that if someone locks their device to a specific orientation then they expect everything to use that orientation, unless they specifically press on a button within an app to change it. And then it should automatically restore your original locked orientation when you close the video.

> Why do you want to look at vertical videos in landscape?

If I locked my device to landscape, started a playlist and then put my device down then I'd expect every video to honor my orientation choice and play the video in landscape. If it doesn't do that then it might eventually get to one that's portrait and suddenly the image is sideways, which requires me to alter how I'm holding the device. That doesn't respect my orientation choice and it's going to annoy people.

Use NewPipe v0.19.8 with your orientation locked to portait, start a mixed playlist and then press on NewPipe's button to change the orientation to landscape. It'll then respect your orientation choice and play every video in landscape until you exit.

> Having two buttons when you can have just one is kind of unoptimized behavior and not good for UI and usabilty.

I disagree. For with auto-rotate enabled it might not be necessary, but with auto-rotate disabled there should be another button so that you have full control like you do with NewPipe v0.19.8.

avently:
@blackbox87 when I read your posts I see myself in your posts. I mean, you talk like I would talk - plain, straightforwardly, unceremoniously. You know what and why you need, you have a use-case for every idea. It's interesting. Where are you from? 

I think we cannot be in agreement in multiple places because we have our use-case which will not work when we hear other's use-case. So that's why I often saying about a PR from you because there is no universal solution of how we can handle some situations. It's completely different vision on some thing, and you can choose only one or another, but not both. Fortunately we have settings in the app which allows to select one or another way for you and me. What if you make your ideas configurable from settings?

blackbox87:
> when I read your posts I see myself in your posts. I mean, you talk like I would talk - plain, straightforwardly, unceremoniously. You know what and why you need, you have a use-case for every idea. It's interesting. Where are you from?

I like to get to the point instead of beating around the bush. Sometimes it's appreciated and sometimes people don't like it, but at the end of the day I'm only trying to help.

I'm currently living in the US, but I'm not originally from here.

> I think we cannot be in agreement in multiple places because we have our use-case which will not work when we hear other's use-case. So that's why I often saying about a PR from you because there is no universal solution of how we can handle some situations. It's completely different vision on some thing, and you can choose only one or another, but not both. Fortunately we have settings in the app which allows to select one or another way for you and me.

I just prefer the way NewPipe v0.19.8 does it since that already respects your orientation choice, and you know what they say about people not liking change? So I really do believe that if the current method is pushed to the master then there's going to be some complaints.

> What if you make your ideas configurable from settings?

I agree that there could be a setting so that NewPipe will either respect your orientation choice or automatically changes the orientation depending on if it's playing a vertical or horizontal video. It's not something that I can setup though, since I'm not familiar enough with the existing code and I honestly don't have a lot of free time right to keep contributing.

opusforlife2:
> Do you see improvements in terms of performance of your device? Does the rotation and prev/next videos switching works better and how much better (on % maybe)?

Single video loads are roughly 2-3x faster.
Playing the next video in a queue is roughly 2x faster.
Rotating a video is roughly 2x faster.

Now the unified player feels just as fast as 0.19.8!

opusforlife2:
> If I started a playlist, locked my device to landscape and then put my device down then I'd expect every video to honor my orientation choice by playing in landscape.

@blackbox87 That is what happens with this PR. So isn't your issue solved? It's only when you've locked to portrait that the player rotates when a vertical video plays.

blackbox87:
@opusforlife2 For landscape yes, but not for portrat. Both should honor your locked orientation IMO, otherwise you're forced to rotate your device depending on a video is horizontal or vertical.

There's also a minor issue if you've got auto-rotate enabled.

- Play a vertical video in fullscreen
- Press the next button to go to a horizontal video

It's missing a button to take you out of fullscreen, so you need to use your navbar instead and that pauses the video.

avently:
@opusforlife2 
> That is what happens with this PR. So isn't your issue solved? It's only when you've locked to portrait that the player rotates when a vertical video plays.

Actually it's not rotating automatically when going from non-vertical to vertical for the same reason as why the player is not responding to click on fullscreen button with locked landscape orientation. So after #4254 the player will probably rotate the screen anyway (based on how this mechanism is implemented in #4254, I didn't check the code)

opusforlife2:
I thought of something just now. I was thinking of how some media players have a 'lock rotation' button. Having yet another button would be unoptimal, though.

Could we instead allow the full screen button to be long pressed to lock orientation? (There is no forced rotation when global orientation is unlocked, so that case doesn't need to be handled.)

avently:
> Could we instead allow the full screen button to be long pressed to lock orientation?

Good idea, I like it. The info about such possibility should be shown somehow. But why it's needed? Why not to just make an option in settings? I don't think that someone will need to switch between these two scenarios often

vkay94:
>Could we instead allow the full screen button to be long pressed to lock orientation? (There is no forced rotation when global orientation is unlocked, so that case doesn't need to be handled.) Maybe with a toast saying "Locked to portrait/landscape".

As much as I like the idea there might be no "simple way" to explain this feature to the user (the user wouldn't think about long pressing it, it's a toggle button in the end if you understand what I mean)

opusforlife2:
When a user taps the full screen button for the first time, a toast could be shown telling them about the long press functionality.

Regarding an option in settings, in general, the developers don't want to add a setting for every single use case. A setting would generally be added when it would be used by a large number of people. There have often been suggestions to have a Miscellaneous menu in Settings, but that hasn't been done yet, for whatever reason.

blackbox87:
I agree with having an option in the settings, since it's easier to understand. I'd label it something like "Orientation changes to suit aspect ratio", which is what this dev branch currently does. Disabling it should restore the behaviour that I'm talking about, as shown in v0.19.8.

heberjeur:
I find this issue, it happens when I slide down to open some videos in the tab channel 
You can see in this video
https://youtu.be/IeARZgj0YnU

## Exception
* __User Action:__ requested stream
* __Request:__ https://www.youtube.com/watch?v=r5_xaaHiUpU
* __Content Country:__ FR
* __Content Language:__ fr-FR
* __App Language:__ fr_FR
* __Service:__ YouTube
* __Version:__ 0.19.8

```
org.schabi.newpipe.extractor.exceptions.ParsingException: Could not load decryption code for the Youtube service.
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.getPlayerUrl(YoutubeStreamExtractor.java:705)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.onFetchPage(YoutubeStreamExtractor.java:655)
	at org.schabi.newpipe.extractor.Extractor.fetchPage(Extractor.java:56)
	at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:68)
	at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:64)
	at org.schabi.newpipe.util.ExtractorHelper.lambda$getStreamInfo$3(ExtractorHelper.java:116)
	at org.schabi.newpipe.util.-$$Lambda$ExtractorHelper$5fJcha6Sq5APJBLdG6osaJby-mc.call(Unknown Source:4)
	at io.reactivex.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:44)
	at io.reactivex.Single.subscribe(Single.java:3666)
	at io.reactivex.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
	at io.reactivex.Single.subscribe(Single.java:3666)
	at io.reactivex.internal.operators.maybe.MaybeFromSingle.subscribeActual(MaybeFromSingle.java:41)
	at io.reactivex.Maybe.subscribe(Maybe.java:4290)
	at io.reactivex.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.drain(MaybeConcatArray.java:153)
	at io.reactivex.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.request(MaybeConcatArray.java:78)
	at io.reactivex.internal.operators.flowable.FlowableElementAtMaybe$ElementAtSubscriber.onSubscribe(FlowableElementAtMaybe.java:66)
	at io.reactivex.internal.operators.maybe.MaybeConcatArray.subscribeActual(MaybeConcatArray.java:42)
	at io.reactivex.Flowable.subscribe(Flowable.java:14935)
	at io.reactivex.internal.operators.flowable.FlowableElementAtMaybe.subscribeActual(FlowableElementAtMaybe.java:36)
	at io.reactivex.Maybe.subscribe(Maybe.java:4290)
	at io.reactivex.internal.operators.maybe.MaybeToSingle.subscribeActual(MaybeToSingle.java:46)
	at io.reactivex.Single.subscribe(Single.java:3666)
	at io.reactivex.internal.operators.single.SingleSubscribeOn$SubscribeOnObserver.run(SingleSubscribeOn.java:89)
	at io.reactivex.Scheduler$DisposeTask.run(Scheduler.java:578)
	at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:66)
	at io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:301)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.lang.Thread.run(Thread.java:919)
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.startsWith(java.lang.String)' on a null object reference
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.getPlayerUrl(YoutubeStreamExtractor.java:700)
	... 30 more

```
</details>
<hr>

opusforlife2:
@heberjeur If it isn't 100% reproducible, then it's probably #3753.

avently:
@heberjeur unrelated to this PR, it's a NewPipe extractor problem

avently:
> And has anyone checked how an Android TV box handles this?

> One line of code, will send with other changes if any of them be needed.

Added this change inside https://github.com/TeamNewPipe/NewPipe/pull/4272/commits/f41549ccf17fb12e8c38de5a5d049291159ceead   for #4272

Domiuns:
> performance.zip

i have tried the apk performance.zip on android 7.0 and i didn't notice any difference the loading improved slightly, but when i rotate the screen it takes 2-3 seconds to show the video

avently:
> but when i rotate the screen it takes 2-3 seconds to show the video

Is it better or worse than before? Is this result constant or happened once?

Domiuns:
> > ma quando ruoto lo schermo ci vogliono 2-3 secondi per mostrare il video
> 
> È meglio o peggio di prima? Questo risultato è costante o è accaduto una volta?


the rotation always remains the same as before, only loads the video first
[record phone.zip](https://github.com/TeamNewPipe/NewPipe/files/5252509/record.phone.zip)

avently:
Didn't understand your answer. You're saying:
> when i rotate the screen it takes 2-3 seconds to show the video

How much time (in seconds) it takes on non-performance apk? For example, on any other unified player apk

Domiuns:
> How much time (in seconds) it takes on non-performance apk? For example, on any other unified player apk

in version 0.19.8 when I rotate the screen is fast to show the video and there is no black screen left with the audio of the video that continues. In the unified player when I press play it takes a few seconds to show the video (i.e. to load the video source) and when I rotate the screen has the same delay shown in the previously loaded video.

> Didn't understand your answer. You're saying:
>> when i rotate the screen it takes 2-3 seconds to show the video

I wanted to say that I have the same situation of the unified player that is when I rotate the screen the video continues to "move" feeling the audio "walk" but the screen remains black for 2-3 seconds

avently:
Did you tried to test [performance-without-notifications.zip](https://github.com/TeamNewPipe/NewPipe/files/5241698/performance-without-notifications.zip)  ? I'm just curious, will you have a better performance?

Domiuns:
> Did you tried to test [performance-without-notifications.zip](https://github.com/TeamNewPipe/NewPipe/files/5241698/performance-without-notifications.zip) ? I'm just curious, will you have a better performance?

I must say that it is much better, you always see a delay but it is acceptable, I can not complain
[phone recorder 2.zip](https://github.com/TeamNewPipe/NewPipe/files/5252589/phone.recorder.2.zip)

avently:
> I must say that it is much better, you always see a delay but it is acceptable, I can not complain

Cool, thank you for testing. We can assume that the cause of the delay is notifications. And I hope #3178 will fix it.

Domiuns:
> Cool, thank you for testing. We can assume that the cause of the delay is notifications. And I hope #3178 will fix it.

thanks to you contributing to this project and trying to help us users. sorry for the long conversation and for the inconvenience I've caused you.

vkay94:
>We can assume that the cause of the delay is notifications. And I hope #3178 will fix it.

I found the reason for it: https://github.com/TeamNewPipe/NewPipe/issues/4286#issuecomment-696199305

@avently Could you try to adjust it with your `PlayerHolder` changes and test it? Just pass `newUrl` instead of `url` in the line:

https://github.com/TeamNewPipe/NewPipe/blob/b69e477ecdf576d0406c0de87db3c8a2866ce741/app/src/main/java/org/schabi/newpipe/player/BasePlayer.java#L469

```
String[] urlParts = getVideoUrl().split("=");
String newUrl = "https://i.ytimg.com/vi/" + urlParts[urlParts.length - 1] + "/hqdefault.jpg";
```

Stypox:
@vkay94 that would not work for services other than Youtube

avently:
@vkay94  this apk with prescaled precached thumbnail with width 200dp. It's not needed to load a small image, it can be just cached with a small width/height after loading.
[perf-thumbnail.zip](https://github.com/TeamNewPipe/NewPipe/files/5265225/perf-thumbnail.zip)

opusforlife2:
Slightly tangential: in the formula px = dp * (dpi / 160), do we input the physical dpi of the device or the value of "Smallest width" in Developer Options?

vkay94:
@avently You're right, this is the best choice here (your perf-performance build reacts flawlessly (also because of the signing ^^)).

avently:
@opusforlife2 from developer settings

avently:
@vkay94  can I link [your issue](https://github.com/TeamNewPipe/NewPipe/issues/4286) for closing after this PR is merged or it's incomplete solution?

vkay94:
@avently Hmm, basically the issue consists of unnecessary update frequency (basically solved by https://github.com/TeamNewPipe/NewPipe/pull/3178#issuecomment-696844911) and high work on the main thread (core, image loading, solved by your changes). 

It's your choice.

avently:
@vkay94 let's wait for rebase of this PR which will be based on notification's PR. And then we will decide.

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
```diff
@@ -497,12 +497,19 @@ public void onLoadingFailed(final String imageUri, final View view,
...
         if (DEBUG) {
             Log.d(TAG, "Thumbnail - onLoadingComplete() called with: "
                     + "imageUri = [" + imageUri + "], view = [" + view + "], "
-                    + "loadedImage = [" + loadedImage + "]");
+                    + "loadedImage = [" + loadedImage + "], "
+                    + loadedImage.getWidth() + "x" + loadedImage.getHeight()
+            + ", scaled width = " + width);
         }
-        currentThumbnail = loadedImage;
     }
```
TobiGr, line 513: Is this change needed with the MedidaStyle notifications?

Stypox: I think so, as doing this in the background prevents it from being done by Android on the main thread. Also, I think the slowdown is not related to MediaStyle, but rather to the MediaSession, so this code probably helps independently of the type of notifications used. @vkay94 correct me if I'm wrong ;-)

avently: In this place
```java
Bitmap.createScaledBitmap
```
I'm scaling down the image only once before assigning it to `currentThumbnail`. Without it Android will resize the thumbnail on every notification update which is a heavy task if the image is large (on YouTube it's large). In this case I made it smaller (much smaller) and if Android needs to resize it the time needed for the resize will be small. I'm not sure about the quality of the thumbnail but for me it looks ok compared to previous thumbnail impelementation. You can then change the width from 200dp (player_notification_thumbnail_width) to something else if the quality is not as good as you want.

The main thing that eats main thread's CPU time is an image resize but after this change there is no such problem. Actually I didn't tested in profiler after the change but visually it doesn't lock main thread as before on rotation change and after other changes.

avently: With media style notifications the thumbnail can be even smaller because it use less space than previous notifications.

avently:
Merged dev. Didn't tested after merge, there were no merge conflicts. 

@vkay94 now you can test performance with all optimizations (from this PR and notifications PR)
@opusforlife2 @blackbox87 switching in fullscreen mode while playing a queue from non-vertical to vertical to non-vertical videos works without leaving fullscreen
@Stypox something I said about tablets will work here with full power after merge Small Fixes 2.  

[perf-final.zip](https://github.com/TeamNewPipe/NewPipe/files/5288507/perf-final.zip)

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
avently: Would be better to somehow get the size for thumbnail inside media style notification and apply this size instead of 200dp. Does anyone know the exact size Android uses for notification?

opusforlife2:
> switching in fullscreen mode while playing a queue from non-vertical to vertical to non-vertical videos works without leaving fullscreen

Awesome! It's noticeably faster now! Thanks!

avently:
> Awesome! It's noticeably faster now! Thanks!

Yep, I like the result. We do here many things that Youtube or other apps don't do. For example, we have an ability to view video description while in landscape fullsreen, we have a notification updates. Without it (like in Youtube) the rotation will be almost instant but I don't want to lose the ability to view description in landscape as well as the features provided by notifications (like background playback, controls, etc). There is a place for optimizaions (like always) so maybe the performance will be even better after someone's PR.

app/src/main/java/org/schabi/newpipe/player/BasePlayer.java
TobiGr: no. I assume that varies from screen size and density.

blackbox87:
> @opusforlife2 @blackbox87 switching in fullscreen mode while playing a queue from non-vertical to vertical to non-vertical videos works without leaving fullscreen

That seems to work, but while testing this I noticed something that happens with your test version and the release candidate.

Watch any video and then select "enqueue in the background" on another video. The audio will pause for a brief second, but then it starts playing the audio again without the video. The player displays a play icon and the notification shows a pause icon. Shouldn't the video remain playing?

avently:
@blackbox87 why do you expect the video to be played when you choose to enqueue it in *background*? So why this enqueue exists from your point of view if the video will continue to play in main player?

opusforlife2:
He means if you're playing video 1, and you enqueue video 2 in background player, video 1 will switch to playing in the background as well. Video 1 shouldn't be affected at all, right?

blackbox87:
> @blackbox87 why do you expect the video to be played when you choose to enqueue it in _background_? So why this enqueue exists from your point of view if the video will continue to play in main player?

Because I'm selecting to enqueue the other video in the background, not the video that's already playing.

@opusforlife2 Exactly.

avently:
@blackbox87 you talk about per-video player's type. Now there are three types of the player: audio (everyone except me call it background), video, popup. Changing the type happens in `handleIntent` of VideoPlayerImpl. So if you want to change that you can make per-video player type and handle the type change in onMetadataUpdate or somewhere else.

I don't think it's useful because it's really unexpected to see a video inside main player and then (after video ends) to see player change without any user interaction. Or even worse to see like player's type changes when you click next/previous video in the queue. For me it looks super unintuitive.

All those buttons (like start playing in background, enqueue in popup and so) should be replaced with a better alternative but no one knows what alternative is.

blackbox87:
@avently Well it's not just me who seems to think that it should behave differently. But it's probably something you'd need more input on from others like @TobiGr and @Stypox.

To me it seems obvious that if I'm watching a video and I click on "enqueue in the background" then I want to watch that video next and not stop the existing video 🤷‍♂️

> All those buttons (like start playing in background, enqueue in popup and so) should be replaced with a better alternative but no one knows what alternative is.

I think "in the background" should be removed and it should instead work as a **temporary session queue**. So for example, while watching a video you might see a few other videos that you'd like to watch next, so you'd could click on "enqueue" to add them to the session queue. And if you then wanted to start playing them all in the background all you'd need to do is press on the existing "background" button.

opusforlife2:
Wait. I had written a comment, where did it go? Or did I just close the tab without posting it? -.-

Anyway, I had written something similar to what blackbox has written above, that since the unified player is meant to have only one queue, enqueuing on popup or audio (there you go, avently!) should just add to the common queue and not affect which player it is.

For example, if video 1 is playing in audio and you enqueue video 2 in popup, it should ignore the popup part and keep playing in audio mode. Same for any other player and enqueuing combinations.

This is a band-aid solution until we can rethink queuing UI and behaviour for the unified player as a whole.

avently:
@blackbox87 this is a possible solution, yes. But it limits current possibilities. So now you can drop the whole queue by starting to play one stream ('start in background') or you can switch the player by choosing to enqueue in popup. So the solution you are talking about limits the possibilities. I don't know is this acceptable or not. It's not hard to do so someone could make a PR and everyone can say their opinion.

Since closing of performance PR I'll not be an active contributor anymore. My life changed so much so I'm not interested in doing something large for youtube app. If you expect that I'll start making your ideas alive, don't.

opusforlife2:
You've done more than enough for Newpipe and its users. Thank you so much! 🥳

avently:
@opusforlife2 you're welcome and thank you too:) You're doing so much as well as other maintainers. This is hard job but looks like you like it

blackbox87:
@avently Honestly, my suggestion was just a quick idea. I figured I'd put it out there since @opusforlife2 seems to agree that the existing behaviour is most likely unexpected and unwanted. But to come up with a solution it's going to require input from a few people.

I don't expect you to fix every issue, if you even see this as an issue. But regardless we've all got other things going on in our lives and we can only contribute so much at a time. It's not like we're getting paid to do this.

Good luck to you.

TobiGr:
i think we should replace "enqueue in backgorund / popup" with one option: "enqueue". Switching to other players is very unexpected behaviour.

TobiGr:
> Since closing of performance PR I'll not be an active contributor anymore. My life changed so much so I'm not interested in doing something large for youtube app. If you expect that I'll start making your ideas alive, don't.

Thank you for your contributions! Good luck in real life :)

avently:
At least 3 + 1 (me) people agree to remove different player's types from menu. This is a good start.

>Good luck to you.

I'm not dying, lol.

vkay94:
@avently 
>now you can test performance with all optimizations (from this PR and notifications PR)

Looks great. The delay is gone 👍

TobiGr (approved):

Stypox (approved):
Thank you again for your contributions, and good luck with life :-D
@opusforlife2 you can merge ;-)

avently:
First steps of @opusforlife2 as a maintainer :)

opusforlife2:
٩( ᐛ )و

PR #4315, opened 2020-09-25
String improvements

comradekingu:
#### What is it?
- [x] Bug fix (user facing)
- [x] Feature (user facing)

- Spelling

Not sure what "Change the download folders to take effect" means…

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -8,7 +8,7 @@
     <string name="install">Install</string>
     <string name="cancel">Cancel</string>
     <string name="fdroid_vlc_url" translatable="false">https://f-droid.org/repository/browse/?fdfilter=vlc&amp;fdid=org.videolan.vlc</string>
-    <string name="open_in_browser">Open in web-browser</string>
+    <string name="open_in_browser">Open in browser</string>
     <string name="open_in_popup_mode">Open in popup mode</string>
     <string name="share">Share</string>
     <string name="download">Download</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -8,14 +8,14 @@
     <string name="install">Install</string>
     <string name="cancel">Cancel</string>
     <string name="fdroid_vlc_url" translatable="false">https://f-droid.org/repository/browse/?fdfilter=vlc&amp;fdid=org.videolan.vlc</string>
-    <string name="open_in_browser">Open in browser</string>
+    <string name="open_in_browser">Open in web-browser</string>
     <string name="open_in_popup_mode">Open in popup mode</string>
     <string name="share">Share</string>
     <string name="download">Download</string>
     <string name="controls_download_desc">Download stream file</string>
     <string name="search">Search</string>
     <string name="settings">Settings</string>
-    <string name="did_you_mean">Did you mean: %1$s?</string>
+    <string name="did_you_mean">Did you mean \"%1$s\"?</string>
     <string name="search_showing_result_for">Showing results for: %s</string>
     <string name="share_dialog_title">Share with</string>
     <string name="choose_browser">Choose browser</string>
@@ -73,8 +73,8 @@
     <string name="use_inexact_seek_summary">Inexact seek allows the player to seek to positions faster with reduced precision. Seeking for 5, 15 or 25 seconds doesn\'t work with this.</string>
     <string name="seek_duration_title">Fast-forward/-rewind seek duration</string>
     <string name="clear_queue_confirmation_title">Ask for confirmation before clearing a queue</string>
-    <string name="clear_queue_confirmation_summary">After switching from one player to another your queue may be replaced</string>
-    <string name="clear_queue_confirmation_description">Queue from the active player will be replaced</string>
+    <string name="clear_queue_confirmation_summary">Switching from one player to another may replace your queue</string>
+    <string name="clear_queue_confirmation_description">The active player queue will be replaced</string>
     <string name="download_thumbnail_title">Load thumbnails</string>
     <string name="show_comments_title">Show comments</string>
     <string name="show_comments_summary">Turn off to hide comments</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -8,14 +8,14 @@
     <string name="install">Install</string>
     <string name="cancel">Cancel</string>
     <string name="fdroid_vlc_url" translatable="false">https://f-droid.org/repository/browse/?fdfilter=vlc&amp;fdid=org.videolan.vlc</string>
-    <string name="open_in_browser">Open in browser</string>
+    <string name="open_in_browser">Open in web-browser</string>
```
comradekingu, line 11: ?

app/src/main/res/values/strings.xml
```diff
@@ -127,8 +127,8 @@
     <string name="peertube_instance_add_exists">Instance already exists</string>
     <string name="settings_category_player_title">Player</string>
     <string name="settings_category_player_behavior_title">Behavior</string>
-    <string name="settings_category_video_audio_title">Video &amp; audio</string>
-    <string name="settings_category_history_title">History &amp; cache</string>
+    <string name="settings_category_video_audio_title">Video and audio</string>
+    <string name="settings_category_history_title">History and cache</string>
```
TobiGr, line 131: The old version looks good to me. Why do you want to change it?

app/src/main/res/values/strings.xml
```diff
@@ -423,7 +423,7 @@
     <string name="kiosk">Kiosk</string>
     <string name="trending">Trending</string>
     <string name="top_50">Top 50</string>
-    <string name="new_and_hot">New &amp; hot</string>
+    <string name="new_and_hot">New and hot</string>
```
TobiGr, line 426: That's SoundCloud's original name / spelling. We should not change it.

app/src/main/res/values/strings.xml
TobiGr: We should keep the old version, because it is shorter and there should be any room for confusion with a file browser.

TobiGr (commented):
Thanks for another contribution! You make NewPipe look more professional and easier to understand.

app/src/main/res/values/strings.xml
comradekingu: Because it is not how an ampersand works.

app/src/main/res/values/strings.xml
comradekingu: @TobiGr The problem that arises when there "is any room for confusion" is sometimes you will have specifically web-browser as a translation, and other times you end up with specifically file-browser. Thus removing the desired(?) ambiguity in languages that don't have it.

app/src/main/res/values/strings.xml
opusforlife2: Could you explain that further?

app/src/main/res/values/strings.xml
TobiGr: This context should be clarified through weblate.

opusforlife2: > and there should be any room for confusion

He meant "shouldn't"

app/src/main/res/values/strings.xml
TobiGr: That's how my phone settings look like:

comradekingu: @opusforlife2 
If you look at what the categories are in "Settings" for Android, it will say "and", not "&", between shared categories. Much like it says "and" for mentioning anything above one thing in NewPipe already.
If needing to specifically make it two distinct things, 
"Option1, option 2" instead of 
"Option1 option 2" is the way to go.

It makes it into languages where it doesn't belong at all 
https://hosted.weblate.org/translate/newpipe/strings/nb_NO/?checksum=35c864b620356a41 https://hosted.weblate.org/translate/newpipe/strings/nb_NO/?checksum=b402d2ded65c5cd6 https://hosted.weblate.org/translate/newpipe/strings/nb_NO/?checksum=8260572532e87eb2

opusforlife2: Oh. So it's causing problems for translations. Makes sense to change it then. I asked because it's fine grammar-wise.

comradekingu: @opusforlife2 I tried to Wolfram Alpha an answer, but didn't work. However "and" is unproblematic.
Also consistency is better to every mention of "and", but specifically
<string name="charset_letters_and_digits">Letters and digits</string> was and remains "and", where it is exactly the same convention as "&" is for the changed strings.

app/src/main/res/values/strings.xml
comradekingu: @opusforlife2 I wasn't 100% sure myself, and I translated it from English.
"string name="open_in_browser" is of no help, but at-least that has to change for it have any hopes of landing on web-.

app/src/main/res/values/strings.xml
opusforlife2: I don't think there are any strict rules around it. From what I know, you generally use 'and' everywhere, and only use '&' when you have less space available.

app/src/main/res/values/strings.xml
opusforlife2: Sorry, did you mean to address Tobi here? I have no idea about any of this.

TobiGr: Looks like this now:
https://hosted.weblate.org/translate/newpipe/strings/en/?checksum=7f6f1617954bdd58

TobiGr: I posted a comment to notify all translators, too. @comradekingu Are you okay with that? If yes, I'd suggest to  revert your change back to "browser" and merge this pr.

comradekingu, line 11:
```suggestion
    <string name="open_in_browser">Open in browser</string>
```

comradekingu: @TobiGr Done. If you can add https://hosted.weblate.org/user/kingu/ as admin in https://hosted.weblate.org/access/newpipe/ I can add source comments and upload screenshots so that no confusion arises.

TobiGr (approved):

app/src/main/res/values/strings.xml
TobiGr: Thank you for the offer. Done.
Btw. do you know, if i can restrict your permissions? I can only choose admin and review. But [according to the docs](https://docs.weblate.org/en/latest/admin/access.html#groups), there should be more groups.

comradekingu: @TobiGr I think those are available as top level (Weblate) admins to do (on Hosted), so only for self-hosted instances. There is "review" access, but I don't think there are any more rights attached to it. I got admin access for the ones in https://hosted.weblate.org/user/kingu/#managed
Can get the setup working a bit better some time later, right now there isn't fastlane support and a few other things could be better.

Issue #4141, opened 2020-08-16
Add branch info into release builds


avently:
#### Describe the feature you want
I want to see support of custom branches (i.e. branches for PR makers) inside release part of build.gradle file. For now there is only such support for debug build type but not for release.

#### Is your feature request related to a problem? Please describe it
Right now when I make some code changes for PR I have to specify 
```groovy
resValue "string", "app_name", "NewPipe Feature name"
applicationIdSuffix ".mysuffix"
```

here: https://github.com/TeamNewPipe/NewPipe/blob/dev/app/build.gradle#L55

Because users asks me about release build instead of debug build. Debug build is terribly slow because of enabled debugging (this is expected behaviour for all apps).

#### Additional context
Just make the same thing from debug build with release build (set resValue and applicationIdSuffix if workingBranch != "master" && workingBranch != "dev")

#### How will you/everyone benefit from this feature?
Developers will not need to write this info for every PR they do after you make proposed change.

wb9688:
Note how it'll also still need to work for F-Droid, which will check out a certain tag, which might not be the latest commit on master.

opusforlife2:
> Developers will not need to write this info for every PR they do after you make proposed change.

At the very least, in case this change risks disrupting anything, as a last resort the build.gradle could have this code in comments
```gradle
/*
applicationIdSuffix ".debug." + normalizedWorkingBranch
resValue "string", "app_name", "NewPipe " + workingBranch
archivesBaseName = 'NewPipe_' + normalizedWorkingBranch
*/
```
so that it takes marginally less work than before.

avently:
@wb9688 then there is another way: just to add a new build configuration which can be used by devs. It should be inherited from "release". Something like "releaseLocally".

opusforlife2:
@TobiGr Ja?

TobiGr:
``` Gradle
    pr {
           multiDexEnabled true

           // suffix the app id and the app name with git branch name
           def workingBranch = getGitWorkingBranch()
           def normalizedWorkingBranch = workingBranch.replaceFirst("^[^A-Za-z]+", "").replaceAll("[^0-9A-Za-z]+", "")
           if (normalizedWorkingBranch.isEmpty() || workingBranch == "master" || workingBranch == "dev") {
               // default values when branch name could not be determined or is master or dev
               applicationIdSuffix ".releaseTest"
               resValue "string", "app_name", "NewPipe Test"
           } else {
[...]

PR #4326, opened 2020-09-26
Allow numbers and uppercase letters in app package id

Stypox:
#### What is it?
- [ ] Bug fix (user facing)
- [ ] Feature (user facing)

When I work on PRs from other people I am used to calling the local branches `prNUMBER`, i.e. `pr3178`. But all such names resulted the same app package id, since numbers were stripped completely, leading to conflicting APKs (e.g. https://github.com/TeamNewPipe/NewPipe/pull/4259#issuecomment-699543016 ). This PR fixes the issue by only stripping the numbers at the beginning of the git branch, [as required by Android](https://stackoverflow.com/questions/12041857/what-is-the-rule-of-the-name-of-the-packages-that-start-with-number#12042111), but not all of the others. While I was at it I also removed `toLowercase()` since package ids needn't be only lowercase, according to [the same SO](https://stackoverflow.com/questions/12041857/what-is-the-rule-of-the-name-of-the-packages-that-start-with-number#12042111).

#### Agreement

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -33,7 +33,7 @@ android {
 
             // suffix the app id and the app name with git branch name
             def workingBranch = getGitWorkingBranch()
-            def normalizedWorkingBranch = workingBranch.replaceAll("[^A-Za-z]+", "").toLowerCase()
+            def normalizedWorkingBranch = workingBranch.replaceFirst("^[^A-Za-z]+", "").replaceAll("[^0-9A-Za-z]+", "")
             if (normalizedWorkingBranch.isEmpty() || workingBranch == "master" || workingBranch == "dev") {
                 // default values when branch name could not be determined or is master or dev
                 applicationIdSuffix ".debug"


avently:
Since you are changing build.gradle could you please consider adding info I asked a couple of weeks ago in #4141 ?

Something like that is enough:
 ```
pr {
            multiDexEnabled true

            // suffix the app id and the app name with git branch name
            def workingBranch = getGitWorkingBranch()
            def normalizedWorkingBranch = workingBranch.replaceAll("[^A-Za-z]+", "").toLowerCase()
            if (normalizedWorkingBranch.isEmpty() || workingBranch == "master" || workingBranch == "dev") {
                // default values when branch name could not be determined or is master or dev
                applicationIdSuffix ".debug"
                resValue "string", "app_name", "NewPipe Debug"
            } else {
                applicationIdSuffix "." + normalizedWorkingBranch
                resValue "string", "app_name", "NewPipe " + workingBranch
                archivesBaseName = 'NewPipe_' + normalizedWorkingBranch
            }
        }
```
Here: https://github.com/TeamNewPipe/NewPipe/blob/dev/app/build.gradle#L47

It will allow to build non-debuggable apk (i.e. fast) with predefined suffix and app name by just invoking: 
```./gradlew assemblePr```


Of couse with the changes you do for debug build in this PR

**Comment by TobiGr** [comment]
@Stypox your change looks good.
@avently I think naming release builds "Debug" is misleading.
I'd suggest to do something like this:
``` Gradle
    pr {
           multiDexEnabled true

           // suffix the app id and the app name with git branch name
           def workingBranch = getGitWorkingBranch()
           def normalizedWorkingBranch = workingBranch.replaceFirst("^[^A-Za-z]+", "").replaceAll("[^0-9A-Za-z]+", "")
           if (normalizedWorkingBranch.isEmpty() || workingBranch == "master" || workingBranch == "dev") {
               // default values when branch name could not be determined or is master or dev
               applicationIdSuffix ".releaseTest"
               resValue "string", "app_name", "NewPipe Test"
           } else {
               applicationIdSuffix "." + normalizedWorkingBranch
               resValue "string", "app_name", "NewPipe " + workingBranch
               archivesBaseName = 'NewPipe_' + normalizedWorkingBranch
           }
       }
```
Apart from that, we also need to change 
https://github.com/TeamNewPipe/NewPipe/blob/14486782dcfb3ac99965b995c1e45c6368c05ce5/app/src/main/java/org/schabi/newpipe/MainActivity.java#L92
I am not sure if there are more changes necessary. This needs to be tested properly and I'd therefore suggest to do it in another PR.

TobiGr (approved):

B0pol:
yes, it should be done in another PR

PR #4345, opened 2020-09-28
Fix PlaylistAppendDialog showing when no local playlists exist

### LINKED ISSUE #4339: Add to playlist results in flashing if no playlist created before
### Version
- v0.19.8 / 0.20.0 RC

1. Do a clean install or delete all created local playlists.
2. Search for a stream and long press on it to open the menu.
3a. Choose "Add to playlist" (or whatever it is called in the English version) or
3b Schedule the stream into the background player and click on the "Add to playlist"-button in the toolbar (more clearly visible)

### Expected behavior
The "create new playlist"-dialog should appear directly.

### Actual behaviour
**UI**: Before the actual dialog is visible there is a flash (sometimes)
**Code**: The app opens `PlaylistAppendDialog` and inflates the view. If no playlist exists it calls `dismiss` and opens `PlaylistCreationDialog`.

vkay94:
#### What is it?
- [x] Bug fix (user facing)
- [ ] Feature (user facing)

- Added methods to `PlaylistDAO` and `LocalPlaylistManager` to get an observable for checking if there is a playlist
- Added method `checkPlaylists` to `PlaylistAppendDialog` which gives the opportunity to react on existing playlists
- Inserted the check into the three corresponding places (menus + background player)

- fixes #4339 

#### Testing apk

[app-debug-fix-createplaylist.zip](https://github.com/TeamNewPipe/NewPipe/files/5295452/app-debug-fix-createplaylist.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
@@ -40,8 +40,8 @@ public final class PlaylistAppendDialog extends PlaylistDialog {
 
     private CompositeDisposable playlistDisposables = new CompositeDisposable();
 
-    public static Disposable checkPlaylists(
-            final Context context, final Invoker onSuccess, final Invoker onFailed
+    public static Disposable onPlaylistFound(
+            final Context context, final Runnable onSuccess, final Runnable onFailed
     ) {
         final LocalPlaylistManager playlistManager =
                 new LocalPlaylistManager(NewPipeDatabase.getInstance(context));
@@ -50,15 +50,11 @@ public final class PlaylistAppendDialog extends PlaylistDialog {
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe(hasPlaylists -> {
                     if (hasPlaylists) {
-                        onSuccess.invoke();
+                        onSuccess.run();
                     } else {
-                        onFailed.invoke();
+                        onFailed.run();
                     }
                 });
-    }
-
-    public interface Invoker {
-        void invoke();
     }
 
     public static PlaylistAppendDialog fromStreamInfo(final StreamInfo info) {
--- a/app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
@@ -1,5 +1,6 @@
 package org.schabi.newpipe.local.dialog;
 
+import android.content.Context;
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -29,6 +30,7 @@ import java.util.List;
 
 import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.disposables.CompositeDisposable;
+import io.reactivex.disposables.Disposable;
 
 public final class PlaylistAppendDialog extends PlaylistDialog {
     private static final String TAG = PlaylistAppendDialog.class.getCanonicalName();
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
```diff
@@ -38,6 +40,27 @@
 
     private CompositeDisposable playlistDisposables = new CompositeDisposable();
 
+    public static Disposable checkPlaylists(
+            final Context context, final Invoker onSuccess, final Invoker onFailed
```
Stypox, line 44: Maybe use different names here, like onPlaylistFound

app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
```diff
@@ -38,6 +40,27 @@
...
+                .subscribe(hasPlaylists -> {
+                    if (hasPlaylists) {
+                        onSuccess.invoke();
+                    } else {
+                        onFailed.invoke();
+                    }
+                });
+    }
+
+    public interface Invoker {
```
Stypox, line 60: You can just use a Runnable instead

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistManager.java
```diff
@@ -126,4 +126,10 @@ public String getPlaylistThumbnail(final long playlistId) {
                 }).subscribeOn(Schedulers.io());
     }
 
+    public Maybe<Boolean> hasPlaylists() {
```
Stypox, line 129: Should't this return a Single? Is it impossible for the database getCount to return a Single? I might be wrong

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -40,8 +41,14 @@
 
     append_playlist(R.string.append_playlist, (fragment, item) -> {
         if (fragment.getFragmentManager() != null) {
-            PlaylistAppendDialog.fromStreamInfoItems(Collections.singletonList(item))
-                    .show(fragment.getFragmentManager(), "StreamDialogEntry@append_playlist");
+            final PlaylistAppendDialog d = PlaylistAppendDialog
+                    .fromStreamInfoItems(Collections.singletonList(item));
```
Stypox, line 45: Maybe extract this code to a function so that it can be reused above

Stypox (commented):
Thank you for looking into this! The changes look mostly good. :-D

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistManager.java
vkay94: I've tried `Single` (`.single(0L)` on `Flowable`) but the `subscribe`-block won't be called (I logged it).

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
vkay94: I'm not sure what you mean by "extract" in this context. Probably a wrapper for the check- and show-logic? Something like one signature:

`static Disposable safeShow(context, fm, tag, <input>)`

where input is either `StreamInfo` or `List<StreamInfoItem>` or 'List<PlayQueueItem>' (by overriding).

Stypox: Oh, sorry, since the two blocks of code actually do different things you can ignore this suggestion

vkay94:
@Stypox I pushed your recommendations to the branch. Thanks for them! (Why didn't I think about the Runnable? xD - I'm too spoiled with Kotlin's built-in lambda as parameters ^^)

Stypox (approved):
Now code looks good to me, thank you again!

TobiGr (approved):
Fixed a merge conflict

PR #4347, opened 2020-09-29
Player rebind

avently:
#### What is it?
- [x] Bug fix (user facing)
- [x] Feature (user facing)

- each time something starts to play in any player VideoDetailFragment will be started (if not yet started) and mini player will show up. It makes possible to see a playing stream in mini player even if the stream was started without using fragment or after player service was closed somehow
- play/next/queue buttons will be updated in realtime when stream was added/removed from queue instead of waiting for a onPlay/onPause action to happen. It needed when you have opened popup and doing something with queue in background activiry or fragment
- when popup or background players start the stream will start playing only if paused state wasn't requested. Which means, for example, if a user opens popup it will be started when START_PAUSED is false. If, for example, the stream was played in main player and then popup was started the stream will still be playing, but if it was paused it still be paused in popup (or background) in APPEND_ONLY mode (but will be playing on new queue initialization). Who read these sentences and understood please tell me where I can learn more about how to use different tenses in such hard way but correctly
- the play/pause button in mini player will be set as paused once player service dies.

Maybe I missed something since it's quite serious change in logic and may broke other functionality but it works ok for me. Some weak places I already changed. It was something I should do in initial PR but better now than never. I'm sure that other changes to unified player architecture after this PR can be made by someone without deep knowledge of it.

https://github.com/TeamNewPipe/NewPipe/issues/4332#issuecomment-700256016
https://github.com/TeamNewPipe/NewPipe/issues/4332#issuecomment-700325604

#### Testing apk
[rebind.zip](https://github.com/TeamNewPipe/NewPipe/files/5296172/rebind.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -20,7 +20,10 @@
 
 package org.schabi.newpipe;
 
+import android.content.BroadcastReceiver;
+import android.content.Context;
 import android.content.Intent;
+import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
 import android.os.Build;
@@ -101,6 +104,8 @@ public class MainActivity extends AppCompatActivity {
     private boolean servicesShown = false;
     private ImageView serviceArrow;
 
+    private BroadcastReceiver broadcastReceiver;
+
     private static final int ITEM_ID_SUBSCRIPTIONS = -1;
     private static final int ITEM_ID_FEED = -2;
     private static final int ITEM_ID_BOOKMARKS = -3;
@@ -147,6 +152,7 @@ public class MainActivity extends AppCompatActivity {
         if (DeviceUtils.isTv(this)) {
             FocusOverlayView.setupFocusObserver(this);
         }
+        setupBroadcastReceiver();
     }
 
     private void setupDrawer() throws Exception {
@@ -454,6 +460,7 @@ public class MainActivity extends AppCompatActivity {
         if (!isChangingConfigurations()) {
             StateSaver.clearStateFiles();
         }
+        unregisterReceiver(broadcastReceiver);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


opusforlife2:
> Who read these sentences and understood please tell me where I can learn more about how to use different tenses in such hard way but correctly

The only solution is to read books in English. Or use smaller sentences. :P

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -302,6 +304,12 @@ public static VideoDetailFragment getInstance(final int serviceId, final String
         return instance;
     }
 
+    public static VideoDetailFragment getInstanceCollapsed() {
```
Stypox, line 307: I'd call this `newCollapsedInstance`

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -795,9 +802,24 @@ private void handleIntent(final Intent intent) {
...
+    private void setupBroadcastReceiver() {
+        broadcastReceiver = new BroadcastReceiver() {
+            @Override
+            public void onReceive(final Context context, final Intent intent) {
+                if (intent.getAction().equals(VideoDetailFragment.ACTION_PLAYER_STARTED)) {
+                    final Fragment fragmentPlayer = getSupportFragmentManager()
+                            .findFragmentById(R.id.fragment_player_holder);
+                    if (fragmentPlayer == null) {
+                        NavigationHelper.showMiniPlayer(getSupportFragmentManager());
+                    }
```
Stypox, line 815: Could you add a comment explaining what this piece of code does? 
Is the following interpretation correct? "If a player is started when there is no `VideoDetailFragment` active, the bottom sheet is initialized."

Stypox (commented):
Thank you! I only took a look at the code, I will test later

Stypox:
I tested this and it seems to work as expected. It also fixes the issue with swiping away the notification on Android 11, since now the player is also accessible from within the app. The only unexpected behaviour, which though is probably intended and shouldn't be changed, is the fact that after closing the ~~background~~ audio <sub>as you prefer ;-)</sub> player via the notification I expected the play button in the collapsed bottom sheet to reopen the audio player instead of the video one.

avently:
@Stypox interesting idea. Let's say you closed the player via notification or closed popup via swiping. Then you hid the app and after some hours you entered in it again and pressed play in mini player. And you'll see popup or background player launched. It's unexpected for me at least. If you want similar behavior I have another idea. Maybe long click on play in mini player can be bound to showing alertDialog with possible options like audio, video, popup, external player, etc. And it can work like this even with currently playing stream and allows to switch players. I think it's useful and interesting feature. If you want, you can create it:)

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
avently: `new` means that can be another instance, but it's not possible here. Instance can not be collapsed. So I just named it `getInstanceInCollapsedState`. Ok?

app/src/main/java/org/schabi/newpipe/MainActivity.java
avently: Added comments with some contextual explanation

Stypox (approved):
Thank you, the code looks good to me, though I haven't tested toroughly

TobiGr (approved):

PR #4354, opened 2020-09-29
Improve age restriction and yt restricted content strings

Stypox:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

This PR changes the titles for "Show age restricted content" and "Enable YouTube's Restricted Mode" and adds a summary to them, since they have been confused by many users lately. In order to reduce redundancy in translations, I also changed the restricted-video-error string to include a `%1$s` instead of the duplicate "Show age restricted content" string. If the latter change causes prolems with translation syncing I can revert it. @wb9688 @opusforlife2 @comradekingu what do you think of these strings?

#### Testing apk
I also tested whether the latter change cause problems with languages different than English, where `%1$s` is not there yet, but it isn't creating problems.
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5301355/app-debug.zip)

#### Agreement

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -154,10 +154,11 @@
     <string name="popup_playing_append">Queued on popup player</string>
     <string name="c3s_url" translatable="false">https://www.c3s.cc/</string>
     <string name="content">Content</string>
-    <string name="show_age_restricted_content_title">Age restricted content</string>
-    <string name="video_is_age_restricted">Show age restricted video. Future changes are possible from the settings.</string>
-    <string name="youtube_restricted_mode_enabled_title">YouTube restricted mode</string>
-    <string name="restricted_video">This video is age restricted.\n\nTurn on \"Age restricted content\" in the settings if you want to see it.</string>
+    <string name="show_age_restricted_content_title">Show age restricted content</string>
+    <string name="show_age_restricted_content_summary">Show content possibly unsuitable for children because it has an age limit (like 18+).</string>
+    <string name="youtube_restricted_mode_enabled_title">Turn on YouTube\'s \"Restricted Mode\"</string>
+    <string name="youtube_restricted_mode_enabled_summary">YouTube provides a \"Restricted Mode\" which hides potentially mature content.</string>
+    <string name="restricted_video">This video is age restricted.\n\nTurn on \"%1$s\" in the settings if you want to see it.</string>
     <string name="duration_live">Live</string>
     <string name="downloads">Downloads</string>
     <string name="downloads_title">Downloads</string>
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -1551,7 +1551,8 @@ public class VideoDetailFragment
     }
 
     private void hideAgeRestrictedContent() {
-        showError(getString(R.string.restricted_video), false);
+        showError(getString(R.string.restricted_video,
+                getString(R.string.show_age_restricted_content_title)), false);
 
         if (relatedStreamsLayout != null) { // tablet
             relatedStreamsLayout.setVisibility(View.INVISIBLE);
--- a/app/src/main/res/xml/content_settings.xml
+++ b/app/src/main/res/xml/content_settings.xml
@@ -49,13 +49,15 @@
         app:iconSpaceReserved="false"
         android:defaultValue="false"
         android:key="@string/show_age_restricted_content"
-        android:title="@string/show_age_restricted_content_title"/>
+        android:title="@string/show_age_restricted_content_title"
+        android:summary="@string/show_age_restricted_content_summary"/>
 
     <SwitchPreferenceCompat
         app:iconSpaceReserved="false"
         android:defaultValue="false"
         android:key="@string/youtube_restricted_mode_enabled"
-        android:title="@string/youtube_restricted_mode_enabled_title"/>
+        android:title="@string/youtube_restricted_mode_enabled_title"
+        android:summary="@string/youtube_restricted_mode_enabled_summary"/>
 
     <SwitchPreferenceCompat
         app:iconSpaceReserved="false"


app/src/main/res/values/strings.xml
```diff
@@ -154,10 +154,11 @@
     <string name="popup_playing_append">Queued on popup player</string>
     <string name="c3s_url" translatable="false">https://www.c3s.cc/</string>
     <string name="content">Content</string>
-    <string name="show_age_restricted_content_title">Age restricted content</string>
-    <string name="video_is_age_restricted">Show age restricted video. Future changes are possible from the settings.</string>
-    <string name="youtube_restricted_mode_enabled_title">YouTube restricted mode</string>
-    <string name="restricted_video">This video is age restricted.\n\nTurn on \"Age restricted content\" in the settings if you want to see it.</string>
+    <string name="show_age_restricted_content_title">Show age restricted content</string>
+    <string name="show_age_restricted_content_summary">Show content with an age limit (like 18+) and thus being probably unsafe for children.</string>
```
opusforlife2, line 158:
```suggestion
    <string name="show_age_restricted_content_summary">Show content suitable for ages 18 and above.</string>
```

Stypox: All content is **suitable** for ages 18 and above (i.e. the statement is ill-formed, as it is always true); Newpipe also restricts content 14+, 12+, 6+, ... (you name it, even though no service has such specific labels so far afaik); I wanted to talk about children safety just to be more clear, but this can indeed be removed.

opusforlife2: Wat really? So if a video is marked 6+ Newpipe would block it if the setting was off?

Added "only".

Stypox: What about "show content only suitable..."

opusforlife2: It would mean we show only 18+ content, not anything else. :P

opusforlife2: How about just directly saying "Show NSFW content"?

comradekingu, line 158:
```suggestion
    <string name="show_age_restricted_content_summary">Show content only suitable for ages 18 and above.</string>
```
@opusforlife2 Age restricted content is an addition, not an exclusive filter (?)

opusforlife2: The "only" there is ambiguous.

opusforlife2: Ooh. We could diss Youtube a bit.

"Show content Youtube deems unsafe for children below 18".

Stypox: This is not only for Youtube and NSFW does not imply age restricted. Also, one would have to go look for NSFW on urban dictionary (like me a while ago). I'd keep your current suggestion then, it is clear what it means imo

comradekingu: How so?

comradekingu: Would be people below 18. Isn't the limit used to bury content too?

opusforlife2: In the middle, a user might get confused which part it applies to. If it was "and above only" or "Only show content" then they would know clearly either way.

app/src/main/res/values/strings.xml
```diff
@@ -154,10 +154,11 @@
     <string name="popup_playing_append">Queued on popup player</string>
     <string name="c3s_url" translatable="false">https://www.c3s.cc/</string>
     <string name="content">Content</string>
-    <string name="show_age_restricted_content_title">Age restricted content</string>
-    <string name="video_is_age_restricted">Show age restricted video. Future changes are possible from the settings.</string>
-    <string name="youtube_restricted_mode_enabled_title">YouTube restricted mode</string>
-    <string name="restricted_video">This video is age restricted.\n\nTurn on \"Age restricted content\" in the settings if you want to see it.</string>
+    <string name="show_age_restricted_content_title">Show age restricted content</string>
```
comradekingu, line 157:
```suggestion
    <string name="show_age_restricted_content_title">18+ content</string>
```
?

opusforlife2: "Show 18+ content"

comradekingu:
@Stypox Not a fan of "<string name="show_age_restricted_content_summary>", I'll read up on what is actually "age restricted". Right now it explains where it shouldn't.

app/src/main/res/values/strings.xml
Stypox: `Show` should not be removed. And as I said above that setting would block all videos with an age restriction (i.e. even 6+, if they exist), not only 18+ ones

Stypox:
@comradekingu thank you

opusforlife2:
@Stypox It doesn't make sense to have a universal age restriction setting. It should be per-service.

Stypox:
@opusforlife2 well, if someone just wants to protect their kid from viewing age restricted videos, he would do that with all services at once, wouldn't him? When would it be the case that the user hides only age restricted videos of some services?

opusforlife2:
Because different services (companies) would have different definitions and thresholds of what constitutes as 'not suitable for children'. A person might want to enable age restriction for one service but not find any need to do it for another.

On a conceptual level, age restriction is a content filter that is implemented on the service side, not in Newpipe. Hence it should be controllable on a per-service level.

wb9688:
Imho the strings @Stypox has now are fine.

@opusforlife2: If we even agree that we want that, not in this PR.

opusforlife2:
Of course.

app/src/main/res/values/strings.xml
opusforlife2:
```suggestion
    <string name="show_age_restricted_content_summary">Show content intended for a mature audience.</string>
```

comradekingu: @opusforlife2 "And above only" is not the same statement as "Show content only suitable __", because if it was it wouldn't be English. Insofar that it isn't, you already pointed out it means _a limitation to 18+ content_. Similarly, "Only show content __" is a valid statement that means _a limitation to 18+ content_, but now in functional English.
Still confused as to what the ambiguity is.

opusforlife2: When I say a user might get confused, I'm thinking of users for whom English is not the first language, but they still use the app in English. They might find it difficult to tell if that "only" applies to "show content" or to "ages 18 and above". Putting the "only" on either side would reduce chances of that confusion.

Anyway, I suggested another way to phrase it, above. Check it out.

comradekingu: Only→suitable for, suitable for→ ages 18 and above
It isn't the simplest sentence, and quite possibly not the simplest sentence possible.
Ruining the actual content for people that speak English, in their own language, is not really why there are translations, or what should be done. Will reply.

comradekingu: This is flawed for two reasons
1. It suggests the material will be adjusted for the mental capacities of someone deemed to be at a mature level.
2. It suggests the same could be acquired by cutting material intended for immature audiences (it isn't(?)
3. By turning it off (I imagine it is on by default), the same happens. Is now content for mature audiences cut?

Intention is not a good guide for what actually happens, regardless of what actually happens.

opusforlife2: I was going by the idea that this sort of terminology is already used when classifying content for various audiences.

ESRB has a Mature category, for example. And I'm sure you must have seen the quote "The following is intended only for mature audiences. Viewer discretion advised. " several times.

opusforlife2: Fair enough. What about:
```suggestion
    <string name="show_age_restricted_content_summary">Show content restricted to ages 18 and above.</string>
```
?

Stypox:
@opusforlife2 @comradekingu I don't think we have to aim for shortness. Descriptiveness is the most important thing here, and in my humble opinion all of the proposed alternatives miss some point :sweat_smile:

opusforlife2:
The problem is not the length. The sentence is a run-on sentence. It should be broken into two sentences. Let's try that now.

app/src/main/res/values/strings.xml
comradekingu: Good.

Stypox:
What about: `Show content with an age limit (like 18+). May be unsuitable for children.`

opusforlife2:
How about this? The main setting title says "Show age restricted content", so just make the description "May be unsuitable for children."

B0pol:
What happens if the string is translated (old version) but not updated? Will it produce an error with `%1$s` missing? We may not want to ship it two days before and update

TobiGr:
@B0pol We need to remove all translations of this string to ensure there is no crash. No matter if we put it into this release or the next one.

B0pol:
Then I would delay this change to another PR, there is no reason to delete, thus _untranslate_ this string for this update.

opusforlife2:
> Then I would delay this change to another PR

You mean another release. 🤭

B0pol:
Both

opusforlife2:
Wut. Why another PR and not this PR in another release?

Stypox:
> What happens if the string is translated (old version) but not updated? Will it produce an error with `%1$s` missing? We may not want to ship it two days before and update

I tested this and it won't

TobiGr (dismissed):

app/src/main/res/values/strings.xml
```diff
@@ -154,10 +154,11 @@
...
+    <string name="show_age_restricted_content_summary">Show content with an age limit (like 18+) and thus being probably unsafe for children.</string>
+    <string name="youtube_restricted_mode_enabled_title">Enable YouTube\'s Restricted Mode</string>
```
comradekingu, line 159:
```suggestion
    <string name="youtube_restricted_mode_enabled_title">Turn on YouTube\'s \"Restricted Mode\"</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -154,10 +154,11 @@
...
+    <string name="youtube_restricted_mode_enabled_summary">YouTube provides a Restricted Mode which hides potentially mature content when enabled.</string>
```
comradekingu, line 160:
```suggestion
    <string name="youtube_restricted_mode_enabled_summary">YouTube provides a \"Restricted Mode\" which hides potentially mature content.</string>
```

app/src/main/res/values/strings.xml
comradekingu, line 158:
```suggestion
    <string name="show_age_restricted_content_summary">Show content possibly unsuitable for children because it has an age limit (like 18+).</string>
```

Stypox:
@comradekingu I implemented your suggestions, thank you! @TobiGr please approve again ;-)

TobiGr (approved):

aand18:
🙏 thanks for this! I had enabled _restricted_ because of lacking description and videos weren't loading anymore and didn't have a clue why...

Issue #4329, opened 2020-09-27
Remove "Remember popup properties" toggle and make it the default behaviour


opusforlife2:
Opened after discussion on IRC.

There doesn't seem to be much use for a case where the user would prefer that the app _doesn't_ remember popup size and position. So the default behaviour should be for the app to remember, and the setting should be removed entirely.

panoreak:
Hello! Can I work on this?

opusforlife2:
Absolutely! Go ahead.

DI555:
why you're deleting the choice of using options!
the option should be workable!
who give you the right to choose for others ???
personally, i, and many others should have the option to not save popup positions!!!

PR #4367, opened 2020-10-02
Removed remember popup properties setting

panoreak:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

- Removed "Remember popup properties" setting in video and audio settings. The app now remembers the size and position of the video player popup by default. 

-Issue #4329 

#### Testing apk
[removeRememberPopupPropertiesApk.zip](https://github.com/TeamNewPipe/NewPipe/files/5320671/removeRememberPopupPropertiesApk.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
@@ -1656,13 +1656,10 @@ public class VideoPlayerImpl extends VideoPlayer
 
         updateScreenSize();
 
-        final boolean popupRememberSizeAndPos = PlayerHelper.isRememberingPopupDimensions(service);
         final float defaultSize = service.getResources().getDimension(R.dimen.popup_default_width);
         final SharedPreferences sharedPreferences =
                 PreferenceManager.getDefaultSharedPreferences(service);
-        popupWidth = popupRememberSizeAndPos
-                ? sharedPreferences.getFloat(POPUP_SAVED_WIDTH, defaultSize)
-                : defaultSize;
+        popupWidth = sharedPreferences.getFloat(POPUP_SAVED_WIDTH, defaultSize);
         popupHeight = getMinimumVideoHeight(popupWidth);
 
         popupLayoutParams = new WindowManager.LayoutParams(
@@ -1676,10 +1673,8 @@ public class VideoPlayerImpl extends VideoPlayer
 
         final int centerX = (int) (screenWidth / 2f - popupWidth / 2f);
         final int centerY = (int) (screenHeight / 2f - popupHeight / 2f);
-        popupLayoutParams.x = popupRememberSizeAndPos
-                ? sharedPreferences.getInt(POPUP_SAVED_X, centerX) : centerX;
-        popupLayoutParams.y = popupRememberSizeAndPos
-                ? sharedPreferences.getInt(POPUP_SAVED_Y, centerY) : centerY;
+        popupLayoutParams.x = sharedPreferences.getInt(POPUP_SAVED_X, centerX);
+        popupLayoutParams.y = sharedPreferences.getInt(POPUP_SAVED_Y, centerY);
 
         checkPopupPositionBounds();
 
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
@@ -209,10 +209,6 @@ public final class PlayerHelper {
         return isBrightnessGestureEnabled(context, true);
     }
 
-    public static boolean isRememberingPopupDimensions(@NonNull final Context context) {
-        return isRememberingPopupDimensions(context, true);
-    }
-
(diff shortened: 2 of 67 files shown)


opusforlife2:
Tested. Works! Thank you!

Stypox (approved):

Stypox:
Could you rebase? Then this is ready to merge, thank you for your contribution!

panoreak:
> Could you rebase? Then this is ready to merge, thank you for your contribution!

Yes! I'm on it 👍

Stypox (approved):

DI555:
why you're deleting the choice of using options!
the option should be workable!
who give you the right to choose for others ???
personally, i, and many others should have the option to not save popup positions!!!

opusforlife2:
Don't double post, or as in your case, triple post, because it achieves nothing except spamming people with notifications.

> why you're deleting the choice of using options!

Because it was discussed and looked like a pointless option. The setting was actually introduced because a developer added the functionality to remember popup size and position. When you install Newpipe, it is on by default. Most people don't even seem to know it exists. And since it is the default behaviour, it was considered needless to have the toggle. You can just move the popup and resize it however you want within seconds.

> who give you the right to choose for others ???

Lol, that is the definition of developing an app. You make choices for yourself and for all users of that app. If you want to gain that right, become a developer or join a development team.

> personally, i, and many others should have the option to not save popup positions!!!

Now we come to the actual point, which is the only sentence you needed to type: What benefit do you get from the app not remembering popup position and size?

DI555:
i might the one of some/many (as seems as were thumbups)... who used exactly this option.
logically i did this because the reason -
i may change popup size and position in time of using NP , but just doesn't want to get it remembered.., just was little cute option that don't want to lose.
and after all would suggest a little co-options 'popup default size' and 'popup default position' just to use with that i discribed upper!

the benefit is in that - i may variaty changing position and size during the session , but want it to be starting from one position and size what i want!
imho this concept with using optionable trigger switch to on/off is giving much more degrees of freedom!

and of course my appologize for @panoreak , huge work is done...
but things're always doing the way you think nobody use it, and suddenly appears some dudes that are using it for years((..
so i wanted to say that some thumbuped dudes and i will feel the lack of this option if it'll be cut out the app((..

opusforlife2:
So if you would like it to start from a certain position the next time, you could place it in that position and _then_ close the popup? Having 3 separate options for a niche feature is too much.

Also, when you regularly use the popup feature, eventually you find out a favourite size and position, which lets you multitask in the best manner. I don't imagine many people would deviate from it much in everyday use.

DI555:
but, lets lalk just without needness of it fornow...
it is an option yet, switchable on and off. guyes who doing this - use just switch on of this option, let us (others) using the swither in position OFF !!! Why don't you, hu ???

opusforlife2:
> Have you made a vote for/against of this option before going to throw it away ??? Why did you decide just in talk with each other in irc chat that this option won't be needed for others ???!!! 

We have several users in our IRC channel. No one objected to removing the setting. Furthermore, I opened an issue about it _on Github_ (#4329) where, again, no one objected to it. So when a contributor came forward and asked to send a PR, we gave the go ahead.

Basically, you're the first person who has raised an objection to the removal, so we had no reason to believe until now that anyone wanted the setting.

> Some thumbups give me the right to talk about that..

You don't gain special rights because of thumbs up or anything like that. Where did you get that idea? Simply by being a user of the app, you're more than welcome to talk about your problems. That's what the bug tracker is for. ;)

> but, lets lalk without needness of it fornow...

Why? An option that no one needs is additional clutter. We should definitely keep 'need' in mind when we're talking about it.

DI555:
i'm a bit in that situation, but it's a pity that i can't change  it((.
but nevertheless ;), remember what the big g doing and did with appearence of adress line??? yes, they just say the full line of adress isn't needed((  and they even didn't get the switch to bring the choice((

the other way, MoFo, if they're doing a new feature - they almost always making a switch for that in about:config!

so, if you decide eliminate an option in ui and/or symplify the options,- please try to make a switch for that in smth.analogue the mozilla's about:config page!
What do you think about that? ..there are a lots of parameters that may be tuned more gently, timings for examples, or some others...

opusforlife2:
I agree with you about an about:config-like menu in settings. See my comment here: https://github.com/TeamNewPipe/NewPipe/pull/4038#issuecomment-667524560

Stypox:
Yeah, I also proposed an advanced-mode menu with miscellaneous niche options

Stypox:
This was reverted since some users (@DI555) turned out to be using this setting. A better solution is needed (like I said above, a miscellaneous/niche settings menu), but for now this won't be included

DI555:
wow, great thanks for that!
i also have some solution about how this option should logically work rightly.
So, when 'remember popup properties' is 'on' , - all must work like it shoud, - you may choose your favorite size&position !!!
And after that if turn 'remember popup properties' to 'off' , - NP should remember your previousely chosen size&position as default size&position !!!
So could be possible always do popup start from your chosen size&position if after you broke it or played with it during the watch ;)

Issue #4380, opened 2020-10-04
Enqueue in the backgorund / popup changes the current player


TobiGr:
### Version

- 0.20.0

### Steps to reproduce the bug
1. Play a video in the "normal" MainVideoPlayer
2. Minimize by pulling down
3. long tap on a stream list item
4. you'll see two options:
    - enqueue in the backgorund
    - enqueue in a pop

### Expected behavior

taping on one of the two buttons enqueues the stream and the current video keeps playing

### Actual behaviour

Current video is played in either audio or pop-up mode.

Solution: replace the two options with a simple "enqueue" option.

PR #4425, opened 2020-10-06
Replace specific enqueue options with one

vkay94:
#### What is it?
- [x] Bug fix (user facing)
- [x] Feature (user facing)

This PR replaces the enqueue options within the StreamDialogEntry-dialogs with one.

**Dev**:
- Added static method to `PlayerHolder` to determine the current running `PlayerType` (VIDEO, AUDIO, POPUP).
- Added new entry `enqueue` which automatically checks which type is running and enqueues the stream accordingly to the right player.
- Adjustments to `showStreamDialog` (see more in User)
- Removed unused `StreamDialogEntry` items and their strings.
- Set generic Toast message for all three types (needs to be translated).

**User**:
If there is no player playing a stream, then the options for queuing won't be shown - there is no need for them, in this case it's the same as start playing in background/popup. 
But If there is a stream playing then there'll be the entry `Enqueue` which enqueues the next stream without interrupting the current flow.

I hope the behavior is what you've expected @TobiGr 

- #4380 

#### Testing apk

[app-debug-2.zip](https://github.com/TeamNewPipe/NewPipe/files/5334875/app-debug-2.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHolder.java
@@ -6,8 +6,12 @@ import android.content.Intent;
 import android.content.ServiceConnection;
 import android.os.IBinder;
 import android.util.Log;
+
+import androidx.annotation.Nullable;
+
 import com.google.android.exoplayer2.ExoPlaybackException;
 import com.google.android.exoplayer2.PlaybackParameters;
+
 import org.schabi.newpipe.App;
 import org.schabi.newpipe.MainActivity;
 import org.schabi.newpipe.extractor.stream.StreamInfo;
@@ -30,6 +34,22 @@ public final class PlayerHolder {
     public static boolean bound;
     private static MainPlayer playerService;
     private static VideoPlayerImpl player;
+
+    /**
+     * Returns the current {@link MainPlayer.PlayerType} of the {@link MainPlayer} service,
+     * otherwise `null` if no service running.
+     */
+    @Nullable
+    public static MainPlayer.PlayerType getType() {
+        if (player == null) {
+            return null;
+        }
+
+        return player.videoPlayerSelected() ? MainPlayer.PlayerType.VIDEO
+                : player.popupPlayerSelected() ? MainPlayer.PlayerType.POPUP
+                : player.audioPlayerSelected() ? MainPlayer.PlayerType.AUDIO
+                : null;
+    }
 
     public static void setListener(final PlayerServiceExtendedEventListener newListener) {
         listener = newListener;
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHolder.java
@@ -46,11 +46,7 @@ public final class PlayerHolder {
         if (player == null) {
             return null;
         }
-
-        return player.videoPlayerSelected() ? MainPlayer.PlayerType.VIDEO
-                : player.popupPlayerSelected() ? MainPlayer.PlayerType.POPUP
-                : player.audioPlayerSelected() ? MainPlayer.PlayerType.AUDIO
-                : null;
+        return player.getPlayerType();
     }
 
     public static void setListener(final PlayerServiceExtendedEventListener newListener) {
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHolder.java
@@ -38,6 +38,8 @@ public final class PlayerHolder {
     /**
      * Returns the current {@link MainPlayer.PlayerType} of the {@link MainPlayer} service,
      * otherwise `null` if no service running.
+     *
+     * @return Current PlayerType
      */
     @Nullable
     public static MainPlayer.PlayerType getType() {
--- a/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
+++ b/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
@@ -1,7 +1,6 @@
 package org.schabi.newpipe.util;
 
 import android.content.Context;
-import android.widget.Toast;
 
 import androidx.fragment.app.Fragment;
 
@@ -12,8 +11,11 @@ import org.schabi.newpipe.player.MainPlayer;
 import org.schabi.newpipe.player.helper.PlayerHolder;
 import org.schabi.newpipe.player.playqueue.SinglePlayQueue;
 
-import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
+
+import static org.schabi.newpipe.player.MainPlayer.PlayerType.AUDIO;
+import static org.schabi.newpipe.player.MainPlayer.PlayerType.POPUP;
 
 public enum StreamDialogEntry {
     //////////////////////////////////////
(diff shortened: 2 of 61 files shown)


app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -1,21 +1,61 @@
...
 import java.util.Collections;
 
 public enum StreamDialogEntry {
     //////////////////////////////////////
     // enum values with DEFAULT actions //
     //////////////////////////////////////
 
+    /**
+     * Enqueues the stream automatically to the current PlayerType.<br>
+     * <br>
```
opusforlife2, line 25: br br?

vkay94: This is  just Java Doc syntax for linebreaks.

opusforlife2: Ah.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHolder.java
```diff
@@ -31,6 +35,24 @@ private PlayerHolder() {
...
+    @Nullable
+    public static MainPlayer.PlayerType getType() {
+        if (player == null) {
+            return null;
+        }
+
+        return player.videoPlayerSelected() ? MainPlayer.PlayerType.VIDEO
+                : player.popupPlayerSelected() ? MainPlayer.PlayerType.POPUP
+                : player.audioPlayerSelected() ? MainPlayer.PlayerType.AUDIO
+                : null;
```
Stypox, line 53: Add a method to `VideoPlayerImpl` that just returns the `playerType` field

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -1,21 +1,61 @@
...
+            case VIDEO:
+                NavigationHelper.enqueueOnVideoPlayer(fragment.getContext(),
+                        new SinglePlayQueue(item), false);
+                break;
+            default:
+                // Same as above, but keep it for now for debugging.
+                Toast.makeText(fragment.getContext(),
+                        "Unreachable code executed", Toast.LENGTH_SHORT).show();
+                break;
+        }
```
Stypox, line 56: I think you can replace this part with something like this, and call it a day. As you said the `type` can never theoretically be `null` or an invalid value, so we can just include that impossible situation into the condition for opening the main video player (enqueueing when there is no player does not raise errors, but just opens that player).
```suggestion
        if (type == AUDIO) {
            NavigationHelper.enqueueOnBackgroundPlayer(fragment.getContext(),
                    new SinglePlayQueue(item), false);
        } else if (type == POPUP) {
            NavigationHelper.enqueueOnPopupPlayer(fragment.getContext(),
                    new SinglePlayQueue(item), false);
        } else /* type == VIDEO */ {
                NavigationHelper.enqueueOnVideoPlayer(fragment.getContext(),
                        new SinglePlayQueue(item), false);
        }
```

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -1,21 +1,61 @@
...
+    }),
+
     enqueue_on_background(R.string.enqueue_on_background, (fragment, item) ->
```
Stypox, line 59: Remove this and the one below, since they are now unused.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -69,6 +109,10 @@
     // non-static methods to initialize and edit entries //
     ///////////////////////////////////////////////////////
 
+    public static void setEnabledEntries(final ArrayList<StreamDialogEntry> entries) {
```
Stypox, line 112:
```suggestion
    public static void setEnabledEntries(final List<StreamDialogEntry> entries) {
```

app/src/main/res/values/strings.xml
```diff
@@ -445,6 +445,7 @@
     <string name="play_queue_stream_detail">Details</string>
     <string name="play_queue_audio_settings">Audio Settings</string>
     <string name="hold_to_append">Hold to enqueue</string>
+    <string name="enqueue_stream">Enqueue stream</string>
     <string name="enqueue_on_background">Enqueue in the background</string>
     <string name="enqueue_on_popup">Enqueue in a popup</string>
```
Stypox, line 450: I am not sure, but if these are unused, please remove them.

Stypox (changes requested):
Thank you for taking care of this! :-D

app/src/main/res/values/strings.xml
opusforlife2: We just need "Enqueue", no matter which button is long-pressed.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHolder.java
vkay94: Alright, I keep the static method in `PlayerHolder`, so it will call `return player.getPlayerType()`.

app/src/main/res/values/strings.xml
Stypox: That's true

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
vkay94: I kept it before reviews, when there might occur something unexpected, but yeah, I'll make it more compact.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
vkay94: I didn't want to remove it before review. I'll remove it alongside the strings ;)

app/src/main/res/values/strings.xml
Stypox: Probably yes, and if you do that please also rename the StreamDialogEntry item from `enqueue_stream` to just `enqueue`

vkay94: @Stypox I can rename the enum, but I can't rename the string resource since there is also `R.string,.enqueue` (which value is "Queue"), so I'd keep the string resource as `enqueue_stream` but change the value to "Enqueue". Okay?

Stypox: Mmmh, having the value "Queue" for `R.string.enqueue` does not seem right, though changing it would create problems with translations. So yeah, do as you said ;-)

vkay94: By the way, regarding strings: 
I just noticed that `NavigationHelper#enqueueOnBackgroundPlayer` and `enqueueOnPopupPlayer` show specific Toast-messages (`R.string.popup_playing_append` and `R.string.background_player_append`). They should be replaced/unified, too. For example with `R.string.player_append`. What do you think?

opusforlife2: Yup.

vkay94:
I've pushed your suggestions/requested changes and updated the OP.

app/src/main/res/values/strings.xml
```diff
@@ -445,8 +445,8 @@
     <string name="play_queue_stream_detail">Details</string>
     <string name="play_queue_audio_settings">Audio Settings</string>
     <string name="hold_to_append">Hold to enqueue</string>
-    <string name="enqueue_on_background">Enqueue in the background</string>
-    <string name="enqueue_on_popup">Enqueue in a popup</string>
+    <string name="enqueue_stream">Enqueue</string>
+    <string name="player_append">Enqueued</string>
```
Stypox, line 449: Just use `enqueued` as the id, or is it already used?

Stypox (approved):
Thank you for implementing the suggestions, almost there :-D

app/src/main/res/values/strings.xml
vkay94: No, it's not used, I've just oriented on the other resources ;)

Stypox (approved):
Thank you again!

Stypox:
And also works after some testing

avently:
@vkay94 thank you!

89z:
@vkay94 I think this PR broke creating playlists in some cases, see my linked issue

chouhanyang:
@vkay94 I looked several related open issues, however I think my daily uses cases is broken and doesn't seem your fix is working. My daily routine is:

1. Open YouTube app.
2. Click share buttons on my subscriptions.
3. Click NewPipe app (with previously set to background player always).
4. Start listening the video/talk show.
5. While listening, I would add bunch of other shows to the background player queue.

Based on your PR description, if the background player is playing, new videos should be "enqueued" instead of clearing entire queue and playing the newly added video. Is this some corner cases I miss?

Thanks for the help.

PR #4453, opened 2020-10-08
Add button in settings to clear reCAPTCHA cookies

### LINKED ISSUE #4279: Option to delete google cookies
Afaik newpipe don't use any types of cookies exept cookies for ReCaptcha. So, I think it would be great to have button to clear them

wb9688:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)

Add button in settings to clear reCAPTCHA cookies

Fixes #4279

#### Testing apk
[app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/5378041/app-debug.apk.zip)

#### Agreement

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
@@ -7,13 +7,13 @@ import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Bundle;
-import androidx.preference.PreferenceManager;
 import android.util.Log;
 import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.preference.Preference;
+import androidx.preference.PreferenceManager;
 
 import com.nononsenseapps.filepicker.Utils;
 import com.nostra13.universalimageloader.core.ImageLoader;
@@ -21,6 +21,7 @@ import com.nostra13.universalimageloader.core.ImageLoader;
 import org.schabi.newpipe.DownloaderImpl;
 import org.schabi.newpipe.NewPipeDatabase;
 import org.schabi.newpipe.R;
+import org.schabi.newpipe.ReCaptchaActivity;
 import org.schabi.newpipe.extractor.NewPipe;
 import org.schabi.newpipe.extractor.localization.ContentCountry;
 import org.schabi.newpipe.extractor.localization.Localization;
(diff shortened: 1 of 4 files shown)


TobiGr (dismissed):
Code looks good to me. Did not test, but I guess, you did :)

app/src/main/res/xml/content_settings.xml
```diff
@@ -1,120 +1,124 @@
...
-        android:title="@string/export_data_title"
         android:key="@string/export_data"
-        android:summary="@string/export_data_summary"/>
+        android:summary="@string/export_data_summary"
+        android:title="@string/export_data_title"
+        app:iconSpaceReserved="false" />
+
+    <Preference
+        android:key="@string/clear_cookie_key"
+        android:title="@string/clear_cookie_title"
```
Stypox, line 100: Maybe also add a summary, saying something like "if you were asked to solve a reCAPTCHA, permission cookies may have been stored on the device"

Stypox (changes requested):
Code looks good to me, too, I don't see why it shouldn't work, though I will test.

@wb9688 could you provide an APK?

app/src/main/res/xml/content_settings.xml
wb9688: A summary is to explain what the setting will do. That sentence doesn't explain that.

Stypox: Yeah, but it explains why that setting is there. Otherwise one would see the word "cookie" and go mad at NewPipe thinking it is tracking him, even if he never solved a recaptcha

TobiGr: Good point. There should be an explanation.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -75,6 +75,15 @@ public void onCreate(@Nullable final Bundle savedInstanceState) {
...
         initialLanguage = PreferenceManager
                 .getDefaultSharedPreferences(requireContext()).getString("app_language_key", "en");
+
+        findPreference(getString(R.string.clear_cookie_key))
+                .setOnPreferenceClickListener(preference -> {
+                    defaultPreferences.edit()
+                            .putString(getString(R.string.recaptcha_cookies_key), "").apply();
+                    Toast.makeText(getActivity(), R.string.recaptcha_cookies_cleared,
+                            Toast.LENGTH_SHORT).show();
+                    return true;
```
Stypox, line 85: Here you should also clear the downloader's cookies:
`DownloaderImpl.getInstance().setCookie(ReCaptchaActivity.RECAPTCHA_COOKIES_KEY, "");`

Stypox (changes requested):
Thank you. Shouldn't the preference be disabled or hidden when there are no cookies to clear?

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
wb9688: Oh, I forgot we kept it in memory there.

wb9688:
@Stypox: Now it's hidden when there's no cookie (and will hide when you press it). I confirmed that it works using Stetho.

opusforlife2:
Does it always check for "no cookie" before hiding or does tapping it suffice? In other words, does it assume the operation will succeed or does it wait and check that cookies have been cleared before hiding? Is hiding prevented if the operation fails for whatever reason?

Stypox:
Either the operation fails making the app crash before it gets to hide the button or the operation succeeds thus correctly hiding the button. So there is no room for inconsistency here

Stypox (approved):
Great, thank you!

PR #4493, opened 2020-10-11
Replace AsyncTask implementations with coroutines.

Isira-Seneviratne:
#### What is it?
- [ ] Bug fix (user facing)
- [ ] Feature (user facing)

- Rewrite implementations of `AsyncTask` as Kotlin suspend functions, as [`AsyncTask`](https://developer.android.com/reference/android/os/AsyncTask) has been deprecated starting with API level 30.

#### Testing apk

[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5361240/debug.zip)

#### Agreement


Stypox:
Thank you! Why coroutines and not ReactiveX? That would prevent us form adding three more libraries

Isira-Seneviratne:
> Thank you! Why coroutines and not ReactiveX? That would prevent us form adding three more libraries

I'm a bit more familiar with coroutines than ReactiveX.

Should I use ReactiveX instead?

Stypox:
I'd prefer that, if it does not create problems

wb9688:
You should definitely use ReactiveX in NewPipe, yeah.

I don't know how good it is to change anything in the GigaGet though.

PR #4534, opened 2020-10-16
Add a secondary control panel to video detail fragment

### LINKED ISSUE #4039: Regression - Share and other buttons in video details
### Version

- 0.20.0

## Problem 
Now that the unified player PR is merged, let me be the first to add a bug to the Unified Player project (https://github.com/TeamNewPipe/NewPipe/projects/17).

We need to find a way to place the Share button back somewhere on the video details page.

Edit: List of buttons reported to be missing:
- Share
- Resolution selector
- Cast to Kodi

List of other possibly useful buttons:
- Play in external player
- Playback speed
- Fit/Fill/Zoom
- Captions
- Mute
- Open in browser

### Solution
~My suggestion was that it could be placed next to the Autoplay toggle.~

From the discussion below:

- How about a left to right swipeable list where the 4 action buttons are? The default visible actions would be the most used ones, and the lesser used ones could be swiped into view, 4 at a time.
- Better would be to allow to rearrange them via long tap and drag.
- I also think it should show as many buttons as possible on that device, e.g. a tablet in landscape will definitely be able to show more than 4 buttons I think.
- Actually, now that I think about it, that right arrow indicator should be present for all form factors, not just TV. We have to let the user know there is something new there, after all. A right arrow beside the download button would be obvious enough to be noticed. Otherwise mobile users would likely not know about this feature until they accidentally swipe sideways in just the right way or read about the feature online.
- Moreover, once the user has swiped left once to access the next 4 action buttons, we should also show a left arrow indicator next to the first action button, as well.

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

<table><tr><td></td><td></td><td>This PR readds the Share, Open in browser and Play in kodi buttons to the video detail fragment interface, after many users gave negative feedback about the segregation of those buttons inside the player. In particular, I added another row of controls (below the one containing Add to, Background, Popup and Download), which is shown only when the description is expanded. In #4039 there have been many other solutions proposed, but unfortunately the scrollable-list-of-buttons one would be a nightmare to implement properly while still making it look good on every device.<br/><br/>This PR also removes the description from then dropdown menu and moves it to a tab alongside related streams and comments.<br/><br/>Since tabs were difficult to find out about in the previous version, since some small and almost invisible dots were shown, this PR changes the look of the bottom tab layout.</td></tr></table>

#### TODO
- [x] Add `TooltipText`s to the `tabView`s
- [x] Fix tabs visible when in fullscreen 
- [x] Animate secondary controls toggle dropdown button

Fixes #4039

#### APK testing 
@opusforlife2 @Younes-L @sanityormadness @WeAreJMJ @nbmrjuhneibkr @cyber199 could you test this apk? Would it suit you?
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5393541/app-debug.zip)

#### Due diligence


opusforlife2:
That looks great!

nbmrjuhneibkr:
Still not as good as the top menu bar that existed before v0.20. What was the reason for removing it, anyway?

Way too much space is occupied just by two buttons, look at all that emptiness between them. My suggestion for improving this is to put icons and text in the same row.

This doesn't fix another regression introduced in 0.20: previously top menu bar also contained video quality selection, which is now inaccessible without starting playback first. As I said here https://github.com/TeamNewPipe/NewPipe/issues/4039#issuecomment-706183593 ,

>if I know that I want to play video in a quality that is different from the default, why would I want to start playing it in the default quality first (wasting limited mobile data)?

opusforlife2:
Yes! My beloved Share button! It is so good to tap you again!

opusforlife2:
> if I know that I want to play video in a quality that is different from the default, why would I want to start playing it in the default quality first (wasting limited mobile data)?

@nbmrjuhneibkr There is a Limit quality setting for mobile data, no? (I feel like this has been discussed before.)

wb9688:
Didn't we agree on a swipeable thingy? Imho this implementation wastes to much space (based on looking at the screenshot).

opusforlife2:
I think that will be fixed by adding more buttons.

nbmrjuhneibkr:
>There is a Limit quality setting for mobile data, no?

There is, but it is a global setting that automatically applies to all videos. You can't deny that previous implementation offered more choice to the user.

avently:
First of all, thank you for trying to find a solution for this problem.
This PR makes the description in very bottom of the page which hurts usability. The solution proposed somewhere was to add a new button in the row near the download button or (my choice) to replace a download button in some kind of "More" button which after click shows second line like you did in this PR. 
It needed to prevent always shown second row which needed only one-two times in month (in my case). I often read description so it's important to make it wasting as less space as possible. Some guys have a small screen so it's important for them too. 
Don't you think it's a better choice to show the "More" button?

avently:
>Didn't we agree on a swipeable thingy?

This solution is great too. The only thing that needs to be done is a horizontal scroll view. Or similar thing that allows pages like android launcher

Younes-L:
> Still not as good as the top menu bar that existed before v0.20. What was the reason for removing it, anyway?
> 
> Way too much space is occupied just by two buttons, look at all that emptiness between them. My suggestion for improving this is to put icons and text in the same row.
> 
> This doesn't fix another regression introduced in 0.20: previously top menu bar also contained video quality selection, which is now inaccessible without starting playback first. As I said here [#4039 (comment)](https://github.com/TeamNewPipe/NewPipe/issues/4039#issuecomment-706183593) ,
> 
> > if I know that I want to play video in a quality that is different from the default, why would I want to start playing it in the default quality first (wasting limited mobile data)?

In some ways this is better, because the buttons are placed lower making them more accessible and easy to reach compared to the action bar in 0.19.8 and older.

That empty space you mentioned could be used to readd the quality selector, what do you think ?

TobiGr:
> What about shifting the 4 action buttons to the left slightly and reducing the space between them, to make space for an expandable arrow button like we have for the description? It could be of the same size as the description arrow.

Having too many collapsible UI elements looks clumsy. I'd go for the swipeable solution as discussed earlier this week.

Stypox:
> Way too much space is occupied just by two buttons, look at all that emptiness between them.

I don't think there is too much emptiness between them, at the end of the day even if some space is wasted, that happens only when the description is open, and it would not bother otherwise

> This doesn't fix another regression introduced in 0.20: previously top menu bar also contained video quality selection, which is now inaccessible without starting playback first. 

I agree this should be fixed, and I will fix it by making use of this same menu, just not in this PR ;-)

> Didn't we agree on a swipeable thingy? Imho this implementation wastes to much space (based on looking at the screenshot).

As I said that would be a nightmare to implement, due to too many nested views listening for scrolls and touches. Also, adding a quality selector to a scrollable layout would just look bad, while making space for it in the current layout would not be a problem.

> Don't you think it's a better choice to show the "More" button?

I don't think that's a viable solution, as it would require removing the very-used download button and making that part of the ui asymmetric, which would probably look bad.

> I'd go for the swipeable solution as discussed earlier this week. 

Mmmh ok, if you all agree that's the way to go then I will try implementing it, though I don't know if I will manage to yield a great result

WeAreJMJ:
@Stypox Very Nice :D 

As for suggestions - It would be nice to have the two buttons exposed all the time. 

Also I saw that you would also be adding the resolution selection in the new area :) So exposing it always would make it easier to use.

In regards to swiping gestures, you would need a clear visual indication of the feature which would mean hiding the the download button :(

WeAreJMJ:
Pushing the description down doesn't affect usability imho. Because it's hidden by default.

Stypox:
What do you think about making the description a tab, side by side with comments and related streams?

opusforlife2:
@Stypox Would have to see a mock up or test apk to decide.

Stypox:
I moved the description to a tab and I really like this solution. What do you all think? [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5470851/app-debug.zip)
@opusforlife2 @Younes-L @sanityormadness @WeAreJMJ @nbmrjuhneibkr @cyber199

Stypox:
The description tab could even contain more information, such as the meta information added by @TobiGr and the various other metadata provided by services

opusforlife2:
Initial impression: I like it.

But if we are going this route, then we **absolutely must** have tab headers, just like the Main Activity. Description should be the first tab.

daufinsyd:
Hi there,
I'm trying the linked apk and so far so good; I'm happy you added the feature back !
One thing, the popup asking what should do when sharing a link from another app doesn't show all options (cf screenshot).
Screenshot_2020-11-26-22-30-35-294_org schabi newpipe debug secondarycontrols

Is that intended ?
Thanks !

Stypox:
@daufinsyd that's unrelated from this PR and was changed recently, after unified ui merged the player with the details view. So now for videos only one of "Video player" and "Show info" will be shown based on the autoplay setting

Stypox:
Rebased, new apk: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5609255/app-debug.zip)

daufinsyd:
Using it since 3 days (mainly listening to music in background), sharing to/from other apps, using queues and other stuff. Seems pretty stable to me. I'll come back if I find a bug.

Stypox:
<table><tr><td></td><td>Improved the bottom tab layout selector: <a href="https://github.com/TeamNewPipe/NewPipe/files/5661873/app-debug.zip">app-debug.zip</a></td></tr></table>

Stypox:
<table><tr><td></td><td>Improved the bottom tab layout selector: <a href="https://github.com/TeamNewPipe/NewPipe/files/5661873/app-debug.zip">app-debug.zip</a></td></tr></table>
@TobiGr @opusforlife2 @daufinsyd @WeAreJMJ @Younes-L @avently @nbmrjuhneibkr @B0pol

avently:
@Stypox I still don't get why not to make a horizontal scroll view with all those buttons like share. It's much easier, and looks better, and more universal.
For me description in the tab isn't so convenient as the current release implementation

Younes-L:
@Stypox I can't see those bottom tabs/buttons, but it does swipe between comments/suggested/description.
How to enable them ?
It defaults to comments which is a bit inconvenient, description is 2 swipes away

The share and browser buttons are also new to me in this build, so imma give feedback on that too : pretty good, better than a swipe or 0.20.5 imo and utilises the already existing expand button. 👍

Stypox:
> I still don't get why not to make a horizontal scroll view with all those buttons like share. It's much easier, and looks better, and more universal.

Maybe yes, but as I said such a menu can't include a quality selector (which I would like to include in the future). Also, I don't know how we could make sure the user notices the menu is swipeable (while the dropdown is obvious). I was also unable to get it working.

> I can't see those bottom tabs/buttons

They show up only when all of a tab is visible, so you'd have to swipe down until comments/related_videos/description uses up all of the screen

avently:
@Stypox quality selector could be added on top of thumbnail like in video player view but for those who have autoplay off

Stypox:
@avently what if someone wants to open the video in an external player and choose the video or audio quality even though he has autoplay enabled? From all of the user's complaint lately, I think it's better to make everything available to everyone instead of requiring some action.

avently:
@Stypox I already implemented quality selector after click on thumbnail for external playback on. You can check it via enabling external playback

Stypox:
Yeah, I know, but that requires enabling the external player, and some people asked to show a selector even if those are not enabled.

avently:
Quality selector should be close to focus because of android TV. It can't be 5 clicks away, for example, from top elements

Stypox:
I am surely not going to change what happens when the user taps on the thumbnail with external players active

Stypox:
> quality selector could be added on top of thumbnail

Also, I already proposed showing buttons on top of thumbnail, but I got bad feedback

avently:
@Stypox 
>Also, I already proposed showing buttons on top of thumbnail, but I got bad feedback

Where is that discussion? It's strange for me to see a bad feedback on that easy to use ui.

Stypox:
https://github.com/TeamNewPipe/NewPipe/issues/4039#issuecomment-706708929

avently:
Don't see any posts with bad feedback. Opus thought that you're going to remove in player buttons but that's not what you wanted. Some buttons you wanted to show is ok for being on top of thumbnail. Other buttons like those we don't need often could be added as a swipeable buttons.

Stypox:
How can two selectors fit there? We would need one for audio and one for video, each along with buttons to open in browser, open in external player and start playing as the current video. Maybe they could be merged, but probably the whole screen would be filled just with that. Also, I don't think it would be a good idea to clutter the upper part of the screen (which is the most looked-at one) with many different UI elements which are only important to some users.
The thumbnail would partially be hidden by those overlay buttons, and I already see some users complaining that they can't see the full thumbnail, etc.
I think hiding almost-unused buttons under a dropdown menu is an expected behaviour (similarly to the three-dot menu), while a swipe menu is usually only used with e.g. chips or customizable tabs, not static buttons. Also, there are probably already too many swipeable things in that page.
Therefore this seems to me a good-enough solution. I agree it isn't perfect for everyone, but I'd prefer not to impact the average user (who does not need an external player and rarely needs share, open in browser and kodi outside of the player, as you also affirmed) with confusing buttons and actions.

avently:
@Stypox 
>But what would users with Autoplay turned on do

Users with autoplay will still have all those buttons in in-player ui.

So you think we need to have a three dots button in a row with download button like a fifth button? Where the quality selector will be then? Or did I understand incorrectly?
In this case all those buttons can be manually ordered like you did with notification. So users can choose what should be shown in a first place and what should be hidden.

Stypox:
> Users with autoplay will still have all those buttons in in-player ui.

Yes, but they may need more specific options for external players, and e.g. @opusforlife2 said he needs both share-at-a-specific-time and default-share.

> So you think we need to have a three dots button in a row with download button like a fifth button?

No, I mean that the current dropdown controls are similar to a three-dot menu, since they are static, non-swipeable and accessible by a tap that reveals more options. Seeing how Android toolbar menus are non-swipeable and use instead a static three-dot menu, I think the current implementation would be more consistent and expected. What I meant was that making an immutable row of buttons swipeable is not common in Android apps, the only instance of this I can think of are Chips, but those are again meant as dynamic and mutable actions.

> In this case all those buttons can be manually ordered like you did with notification.

Which brings in too much complexity for the average user. Adding that amount of new settings for something niche is not a good idea. Customizing the notification is not niche, customizing a menu that would anyway be easily accessible with just one more tap is, also because, as you said, share, open in browser and kodi can already be accessed from other places, so it is expected to make them secondary actions.

Stypox:
Let's sum up pros and cons of the various proposals. Vote on this comment using the emoji in the parenthesis (i.e. 🚀, 👀, 🎉). Feel free to point out other pros or cons.
@TobiGr @opusforlife2 @daufinsyd @WeAreJMJ @Younes-L @avently @nbmrjuhneibkr @B0pol

🚀 Buttons on top of thumbnail
   - 👍 Accessible right away
   - ❌ Partially hides thumbnail
   - ❌ Makes the UI cluttered and complicated for new users
   - ❌ Not much space available for e.g. both audio and video selectors
<br/>

👀 Swipeable row of buttons
   - 👍 Easy to access
   - ❌ Not completely intuitive for all users, also probably difficult to find out about
   - ❌ Too many nested swipeable views
   - ❌ Audio and video selectors would not fit
<br/>

🎉 Current implementation, secondary control panel with dropdown button
   - 👍 Not shown if not needed (which is expected for most users, thus not cluttering the UI)
   - 👍 Space available to add external player selectors with multiple specific and embedded actions in the future
   - 👍 Intuitive even for new users since it is normal to have buttons that appear after tapping on something (see e.g. three-dot menu)
   - ❌ Takes one more click to access, and worse on Android TVs

B0pol:
If you keep the current design, in settings / content you should add an option to hide description as there is the option for comments and related videos

And I don't really like to have to swipe for it, I prefer to click to show description as it is right now and still show share/open in browser when you expand

Stypox:
> In settings / content you should add an option to hide description

Oh, right

> And I don't really like to have to swipe for it, I prefer to click to show description as it is right now and still show share/open in browser when you expand

Yeah, I wouldn't mind that either, but @avently said he does not like swiping down so much to reach the description. So I found this solution which imo feels consistent, since description is just another variable-length video-related info, just like other tabs (title and buttons are not, though). In that fragment in the future we could also show metadata, i.e. license, tags, ...

TobiGr:
> Takes one more click to access, and worse on Android TVs

No problem. There should be enough space to display all buttons in the first row on TV (@B0pol Could you test this?)

Bottom action bar works fine on my device. we might need to show it the whole time on TVs.

B0pol:
> > Takes one more click to access, and worse on Android TVs
> 
> No problem. There should be enough space to display all buttons in the first row on TV (@B0pol Could you test this?)

I don't know what you ask exactly, so here are screenshots (well, pictures)
When you load the video | click on title | scroll down on comments and right button
--- | ---- | ----
IMG_20201209_175948 | IMG_20201209_180004 | IMG_20201209_180026

It's a bit harder because you can access description only by scrolling down on comments and right button, but you can go to related videos everytime with right button. Not impossible though.

@Stypox it would be really cool to display description on the right side when "show related and next videos" is disabled. I think there is enough space for it there.

> Bottom action bar works fine on my device. we might need to show it the whole time on TVs.

I would show it everytime on every devices. Otherwise it's not discoverable until you scroll down.

TobiGr:
I though about something like that for TVs (Please ignore that _very_ bad editing).
101661967-33bd4e00-3a49-11eb-837c-47de64b974a2

Younes-L:
@stypox thanks for the tag
I'd like to suggest some improvements for the secondary controls :
Right now the description is not shown by default. There seems to be no default and the app just remembers the previous tab (comments/suggested/description) and shows it again the next time a video is opened.
Also it would be nice to have the controls expanded by default.
Can I open new issues for these suggestions ? The new controls _"don't exist yet"_  ^^'

TobiGr:
Just a quick thought on https://github.com/TeamNewPipe/NewPipe/pull/4534#issuecomment-741028435: 
| We could add a button to open the play queue view or integrate it as additional tab. Not sure if the last one is a good idea, because we use horizontal swipe in it to remove elements from the queue. That conflicts with the swipe to change the tab. 	|  	|
|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------	|----------------------------------------------------------------------------------------------------------------------	|

vkay94:
>@vkay94 this would probably integrate well with your stream segments, wouldn't it?

@Stypox Sure, it'd would match the though above this comment from TobiGr  (added playqueue tab)- one additional tab if there are segments available. The only question is how it will be presented. Similar to current YouTube's way seems good to me on portrait. The question is how it will be presented in landscape then (in my demo I've copied the playqueue layout :'))

Let's wait how TobiGr imagines the presentation of the items in the playqueue tab ;)

Younes-L:
@TobiGr  what about a right side drawer (similar to the navigation drawer)
Swipe right from left side of screen to open it, swipe right again on any song to remove it from list, and swipe left to close it (or tap outside the drawer, normal drawer stuff)
Pros : 
* easier to use and reach compared to a 4th tab
* Available anywhere (having the queue only available inside video details seems odd)
* Less interference with other horizontal swipe actions

opusforlife2:
> Improved the bottom tab layout selector (swipe down to see it):

@Stypox That's not what I meant. I was thinking of those tab headers just below the action buttons. _Above_ the swipeable section, not below it. It will be far more intuitive and always visible to users.

Also, Android TV users would be able to focus on and select the tab headers easily, which some might prefer, like @B0pol.

> And I don't really like to have to swipe for it

Moreover, this would also make @TobiGr's suggestion easier to implement, I think.

> We could add a button to open the play queue view or integrate it as additional tab. Not sure if the last one is a good idea, because we use horizontal swipe in it to remove elements from the queue. That conflicts with the swipe to change the tab.

> Swipeable row of buttons
Would look exactly like now (in 0.20.5)

Well, we don't seem to be going that way, but if that _was_ the chosen UI, we would obviously add a small arrow indicator or something to show that there's more stuff hidden off screen.

Stypox:
> the app just remembers the previous tab

This is expected and not going to be changed, remembering the last thing is usually a good design since most of the times the user would land on the thing he wanted without having to add a niche option just for that.

> Also it would be nice to have the controls expanded by default.

After this PR is merged you could open an issue for it. A solution I propose is making the dropdown button long-clickable, with a long click making the description always expanded (I mean, for all other videos, too). A simple tap on the dropdown button again returns to normal mode.

Stypox:
> the app just remembers the previous tab

This is expected and not going to be changed, remembering the last thing is usually a good design since most of the times the user would land on the thing he wanted without having to add a niche option just for that.

> Also it would be nice to have the controls expanded by default.

After this PR is merged you could open an issue for it. A solution I propose is making the dropdown button long-clickable, with a long click making the description always expanded (I mean, for all other videos, too). A simple tap on the dropdown button again returns to normal mode.

> I thought about something like that for TVs

Yeah, good idea, I'm going to implement that

> We could add a button to open the play queue view or integrate it as additional tab.

This should be discussed separately, as it is definitely not going to be done in this PR, but could be a good idea ;-)

> That's not what I meant. I was thinking of those tab headers just below the action buttons. Above the swipeable section, not below it. It will be far more intuitive and always visible to users.

<table><tr><td>Yes, I know you didn't mean that, but I thought it would look really ugly... Anyway, here are two test apks with different configurations: <a href="https://github.com/TeamNewPipe/NewPipe/files/5683082/bottom.zip">tab layout always visible at the bottom</a> and <a href="https://github.com/TeamNewPipe/NewPipe/files/5683087/above.zip">tab layout above the pager</a>. I like the first option, it doesn't seem too intrusive even on really small screens (tested with emulator). Instead, the second one just feels wrong in my opinion.</td><td></td><td></td></tr></table>

vkay94:
I like the bottom bar ;) If you decide to stick with it a good addition would be hiding it when scrolling down and re-appearing as soon as the user scrolls up again - similar to browser with their action bar :)

Younes-L:
> > the app just remembers the previous tab
> 
> This is expected and not going to be changed, remembering the last thing is usually a good design since most of the times the user would land on the thing he wanted without having to add a niche option just for that.
> 
The description used to have its own spot before becoming just another tab, in this case I don't think remembering previous tab is a good idea (especially because there are only 2 other tabs, comments and related are 1-2 swipes away from muscle memory).
Here's the usecase : after watching a video , I check the related videos and tap one of them. The first thing I'm interested in is always the description, but because I can't access related videos without "giving up" the description tab, the app will show more related videos every time I tap on a related video.


> > Also it would be nice to have the controls expanded by default.
> 
> After this PR is merged you could open an issue for it. A solution I propose is making the dropdown button long-clickable, with a long click making the description always expanded (I mean, for all other videos, too). A simple tap on the dropdown button again returns to normal mode.
> 
In the secondary-controls test build , the description is outside the dropdown panel, so having it expanded or not doesn't change the visibility of the description... Unless I'm testing the wrong apk ?
My suggestion was an option to make all 6 control buttons visible by default without having to expand the dropdown.

opusforlife2:
@Stypox Ah. Now I realise that my main problem was with the tabs not being visible unless you scrolled down. Now that they are permanently visible, top or bottom doesn't make a difference to me, visually. Anything you go with is fine.

Although this is very minor, I would prefer the Description tab to be the leftmost tab in LTR layout (we don't have RTL layout yet, do we? I think that PR is pending). It seems more logical because you generally read some info about the video and then "swipe right for more", which could be related videos or comments.

Stypox:
Now the tab layout is at the bottom of the screen and is always shown. I also rebased. The apk can be found in GitHub actions: https://github.com/TeamNewPipe/NewPipe/suites/1739245496/artifacts/32993467

TiA4f8R:
@Stypox I think there is a small design problem when long pressing Share and Open in browser commands. See the screenshots above:

| Share long press  | Open in browser long press  |
|---|---|

Stypox:
@TiA4f8R isn't that ok? At least it is how it usually works in android

Stypox:
@TiA4f8R here is an alternative, though I prefer the current behaviour: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5751901/app-debug.zip)

TobiGr:
I guess @TiA4f8R means the size of the backdrop. It is significantly larger than in the row above

Stypox:
@TobiGr the buttons are larger, since there is more space, so selectableItemBackgroundBorderless makes a bigger ripple

TiA4f8R:
@Stypox So you can't change backdrop's size of the two last buttons, am I right?

Stypox:
@TiA4f8R yes, you understood correctly. That's not under my control. I can only decide whether the ripple goes over other views (current implementation) or stays within the button (the provided apk), and I prefer the firs option.

Stypox:
> in settings / content you should add an option to hide description as there is the option for comments and related videos

Done

> I thought about something like that for TVs

This is how it would look. Unfortunately there is not enough space for the "Open in browser" tooltip. And if I try this with the minimum screen density that still triggers the `large-land` view, it's even worse, since we're trying to fit 7 buttons in a place where only 4 would fit

image

Younes-L:
> here is an alternative, though I prefer the current behaviour: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5751901/app-debug.zip)

Installed this on Android Pie and now the permanent bottom tabs are also shown in full screen video player. I don't think that's intended ?

> Unfortunately there is not enough space for the "Open in browser" tooltip. 

It looks ok, the browser icon does spell "browser" well enough x)

opusforlife2:
Github Actions doesn't allow guest users to download artifacts, which is pretty stupid. @Stypox and other PR authors, could you post nightly.link URLs so that users don't have to sign in just to access an APK? If not, then maybe @mhmdanas's bot could be used to download and upload the latest artifact to its comment somehow?

mhmdanas:
@opusforlife2 we discussed this yesterday, and I think we agreed to post a comment using a GitHub Action or something, but we're still discussing WHAT the link in the comment will be. Here's what we have:
- Use nightly.link
- Self-host nightly.link (I can't remember why we wanted to do that)
- Since artifacts can be only downloaded by logged-in users, @Stypox suggested that we can get the link, put a token in it somehow, and then put that in the comment.

By the way, I changed the bot to a GitHub Action :). I'd like it to stay for templates only, but maybe we could create a GitHub Actions for all things NewPipe-specific, like this one.

opusforlife2:
> (I can't remember why we wanted to do that)

Because the website warns of hitting the API limit if too many people use the common token. If we're using nightly.link and we hit the API limit, we'll have to switch to self-hosting anyway.

> we can get the link, put a token in it somehow, and then put that in the comment.

Or have a (new) bot (probably a Github Action) download the artifact APK and upload it to Github normally in the second comment. Those links work for everyone, as we know. Plus, with comment edit history available to us, we'll have all the links the bot generates, in case we don't want the latest artifact, but an older one some commits behind. It could even label them "branch_name-v1.apk", "branch_name-v2.apk", "branch_name-v3.apk", etc., depending on which revision of the artifact it is.

mhmdanas:
There's one problem with your suggestion, if I understood it correctly. You can't upload stuff using the GItHub API like you can upload a file by dragging it into the comment text field.

mhmdanas:
This is a bit off-topic, so I think we should continue this discussion on the IRC channel.

TobiGr (changes requested):
Please add `TooltipText`s to the tabViews.

opusforlife2:
> Installed this on Android Pie and now the permanent bottom tabs are also shown in full screen video player. I don't think that's intended ?

Moreover, it blocks the user from interacting with the bottom elements in full screen.

Stypox:
@TobiGr @opusforlife2 I rebased and solved the issues. Ready to review again :-D

> Please add TooltipTexts to the tabViews.

Done

> Moreover, it blocks the user from interacting with the bottom elements in full screen.

Fixed, now the tab layout only shows when there is enough space. I also added a nice scrolling behaviour, what do you think about it?

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -779,6 +811,14 @@ public void onLoadingFailed(final String imageUri, final View view,
         }
     }
 
+    private void showHideKodiButton() {
+        // show kodi button if it supports the current service and it is enabled in settings
+        final boolean showKodiButton = KoreUtil.isServiceSupportedByKore(serviceId)
+                && PreferenceManager.getDefaultSharedPreferences(requireContext())
+                .getBoolean(getString(R.string.show_play_with_kodi_key), false);
```
TobiGr, line 818: You can use `shouldShowPlayWithKodi()` now:
https://github.com/TeamNewPipe/NewPipe/blob/8fd48a88be33dc09a7215ef31044922ea7c7e376/app/src/main/java/org/schabi/newpipe/util/KoreUtil.java#L19

TobiGr (commented):
I did not have time to test this PR on TVs, tablets and devices with KitKat, Lollipop etc. (not sure if the latter is necessary here).

Code looks good to me

Stypox:
@TobiGr done

opusforlife2:
Full screen tab overlay gone! Thanks.

> I also added a nice scrolling behaviour

What exactly is it? I haven't noticed anything.

Stypox:
@opusforlife2 try scrolling down while in fullscreen

opusforlife2:
You mean how the tab scrolls into view? Nice attention to detail. :)

Stypox:
I rebased again

TobiGr:
I tested everything on phones, tablets and a TV emulator. Works great! However, I did not run an Android 4.4 emulator yet and I'd like to review the code one last time.

There is a small bug on TVs which allows to select items of the minimized player when the video description tab is shown and one is switching from the description to the related videos and vice versa. However, I cannot reproduce it in reliably and thus provide no further info.

Suggestion for a later update: when the next and related videos are hidden and tablets / TV, move the "share" and "open in browser" buttons in the first (and thus only) row next to the download button.

Stypox:
@TobiGr I rebased again (it was tough, this time) and added an animation for the secondary controls toggle dropdown button.

TobiGr (approved):

PR #4548, opened 2020-10-17
Play directly in fullscreen when long-pressing on thumbnail

### LINKED ISSUE #4152: Can't go directly to fullscreen when tapping thumbnail
### Version

- 0.20.0

### Regression
The user can no longer go into fullscreen directly upon tapping the thumbnail in the video details page, like in 0.19.8. (for those checking: requires locked system rotation and landscape mode in main player)

### Additional Note

MUH FULLSCREEN!

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

This PR implements a long-press gesture on the video thumbnail to start playing the video directly in fullscreen. This gesture obviously applies only if autoplay is disabled and autorotation is disabled.

Fixes #4152

#### APK testing 
@opusforlife2 @EliasKomar @nikhilCad @sanityormadness @johanw666 @DuskyVoltage what do you think of this?
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5395812/app-debug.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -243,9 +243,15 @@ public class VideoDetailFragment
     private TabLayout tabLayout;
     private FrameLayout relatedStreamsLayout;
 
+
+    /*//////////////////////////////////////////////////////////////////////////
+    // Service
+    //////////////////////////////////////////////////////////////////////////*/
+
     private ContentObserver settingsContentObserver;
     private MainPlayer playerService;
     private VideoPlayerImpl player;
+    private boolean makeFullscreenWhenPlaybackStarts = false;
 
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -550,6 +556,10 @@ public class VideoDetailFragment
         }
 
         switch (v.getId()) {
+            case R.id.detail_thumbnail_root_layout:
+                openVideoPlayer();
+                makeFullscreenWhenPlaybackStarts = true;
+                break;
             case R.id.detail_controls_background:
                 openBackgroundPlayer(true);
                 break;
@@ -683,6 +693,7 @@ public class VideoDetailFragment
         uploaderRootLayout.setOnLongClickListener(this);
         videoTitleRoot.setOnClickListener(this);
         thumbnailBackgroundButton.setOnClickListener(this);
+        thumbnailBackgroundButton.setOnLongClickListener(this);
         detailControlsBackground.setOnClickListener(this);
         detailControlsPopup.setOnClickListener(this);
         detailControlsAddToPlaylist.setOnClickListener(this);
@@ -1771,6 +1782,16 @@ public class VideoDetailFragment
                         && player.getPlayQueue().getItem().getUrl().equals(url)) {
                     animateView(positionView, true, 100);
                     animateView(detailPositionView, true, 100);
+                }
+                break;
+            case BasePlayer.STATE_BUFFERING:
+                if (makeFullscreenWhenPlaybackStarts) {
+                    // This will only run after the user long clicked on a video to start
+                    // playback directly in fullscreen.
+                    makeFullscreenWhenPlaybackStarts = false;
+                    if (globalScreenOrientationLocked(requireContext())) {
+                        player.onScreenRotationButtonClicked();
+                    }
                 }
                 break;
         }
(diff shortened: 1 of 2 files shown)


opusforlife2:
Awesome! Finally! Thank you!

I still think it should be a separate setting, because this is a permanent behaviour that users want every time, as opposed to sometimes using full screen and sometimes not.

opusforlife2:
Hmmm. The long press duration seems too long. And the animations seem to have slowed down quite a bit. Both entering and exiting full screen with rotations.

avently:
What about Android TV users? I think they need this feature more than others. Since long press is not an option for them maybe would be better to use a setting which allows to force fullscreen on every play. What do you think?

MD77MD:
> What about Android TV users? I think they need this feature more than others. Since long press is not an option for them maybe would be better to use a setting which allows to force fullscreen on every play. What do you think?

poor android TV users, they always seems to get forgotten 🤑

doodhout:
Just an FYI, the video DOES NOT play in full screen automatically while having autorotation enabled on my phone, it just portrays the same video details view with the video playing only in the thumbnail viewport, even when rotating into landscape mode.

Does this PR add the option to just start playing in full screen regardless of any of these other settings like autorotation? Because I'd like to have 1/ autoratation enabled, and 2/ immediate fullscreen play enabled, both of them.

krushia:
Requiring a long press does not solve #4152. The only thing I hate more than mini video players is long pressing.

doodhout:
For several reasons or use-cases, it seems better to give the user the choice (via a setting) to start playing fullscreen immediately (like it was in 0.19) or not fullscreen (like it is since 0.20).

However, there hasn't been any movement on this PR from Stypox, although I can see in his activity log that he's been busy on other stuff. Maybe this can be picked up by someone else is Stypox doesn't mind, or does he still plan on picking this up?

Stypox:
I don't think I am going to keep working on this specific PR, as it seems like there is no agreement on how things should work

doodhout:
I think the solution that will satisfy everyone is keep the behaviour from 0.19 and add a user preference to switch to windowed video playback (like in 0.20) if preferred?

@opusforlife2 @avently Do you agree?

I'm wondering why the default behaviour from 0.19 was changed so radically in 0.20 and that's without the option to revert the behaviour by means of a user preference. I don't think it's bad to just reset the baseline to how it was in 0.19 and add choices from there.

avently:
@doodhout
>I'm wondering why the default behaviour from 0.19 was changed so radically in 0.20

That's because I had no idea why anyone would want to have auto fullscreen after click on thumbnail. I thought it is non-comfortable thing. Also there were problems with performance of orientation change from portrait to landscape and back (initially). I made performance better only after a couple of months from initial PR

avently:
@doodhout
>Do you agree?

I agree that permanent settings is fine in this case. So if a user chose to play in fullscreen he should get thus behavior on every click on thumbnail. However this will not work with auto play on

Stypox:
> the behaviour from 0.19

The behaviour from 0.19 was autoplay always off, and when tapping on the thumbnail fullscreen is accessed directly?

krushia:
@Stypox correct

opusforlife2:
> I think the solution that will satisfy everyone is keep the behaviour from 0.19 and add a user preference to switch to windowed video playback (like in 0.20) if preferred?

@doodhout It's bad to flip flop on defaults like this. In general, it's better if the user is made aware of a feature than not, because most users don't ever bother checking settings to see what's on offer.

Hence I would suggest the reverse. Let the embedded player stay the default, with the option to go straight to fullscreen with a toggle in settings (which would also have to be linked to the autoplay toggle).

doodhout:
@opusforlife2 I agree with not annoying the user with sudden changes to default behaviour, hence the issue behind this PR and the annoyance at the change in 0.20 to play windowed.

I do agree that changing the default again is unnecessary, so your suggestion is fine. 👍

Stypox:
So, I thought about another solution. Adding an entry to the `Autoplay` setting in addition to "Always", "Only on Wi-Fi" and "Never": "**Never, and directly enter fullscreen when tapping on the thumbnail**". What do you think about this? As @avently said, this setting makes sense only when `Autoplay` is set to "Never", so it feels just natural to add the entry in the `Autoplay` menu. Every user changing the autoplay mode comes across it and can decide what he prefers.
@opusforlife2 @doodhout @krushia @MD77MD @avently

avently:
@Stypox I couldn't imaging so clever idea

opusforlife2:
What about discoverability? This solution sounds perfect from our perspective because we just want to add the option back, but how would a new user checking out Newpipe discover that there is an option to go directly to fullscreen?

How about renaming the setting to "Autoplay and full screen"?

Stypox:
Ok then. I'm closing this and aiming for the solution discussed above ;-)

PR #4559, opened 2020-10-18
Removed "Coming features" because why??

opusforlife2:


TobiGr (approved):

wb9688 (approved):
Maybe also add new features to above?

TobiGr:
I think we should go through the features list and remove "trivial" features and only list important ones.
It might also be good to change the general design of the features list, because one cannot find anything in it. It's just messy.
One could make either a table or use an image.

PR #4562, opened 2020-10-18
Fix opening VideoDetailFragment and much more

### LINKED ISSUE #4403: Sharing to Newpipe fails if stream already playing
### Checklist

- [x] I am using the latest version. 
- [x] I checked, but didn't find any duplicates of this issue in the repo. 

1. Play a video in popup or background (or make sure main player minimises to one of them)
2. Switch to another app.
3. Share a link to Newpipe/Open a link that would open in Newpipe.

### Actual behaviour

The currently playing video opens, not the shared one.

### Expected behavior

The shared video should open. This only happens now if the notification is absent (main player is paused and you're in another app).

### LINKED ISSUE #4432: v20 background playback stops and doesn't recover
### Checklist

- [x] I am using the latest version. 
- [x] I checked, but didn't find any duplicates of this issue in the repo. 

1. Add a few long videos to background queue
2. Maybe lose connectivity after a few minutes (I'm not sure it's the cause) 

### Actual behaviour

I was playing a long video in background and it stopped after a few minutes. I think it was buffering (notification showed the `hourglass` instead of the `play` button). 
It never recovered.

 I tried jumping to next in queue, but it still stayed on buffering (I don't remember if it switched to next video or ignored the "next video" command). 

This could be a different bug - on v19 you could press `next` when the current video was stuck and it would play the next queued video. 

Finally, I exited with the `X` notification button. 

### Expected behavior

Background playback should recover from connectivity issues. 

### Device info

 - Android version/Custom ROM version:  stock Android 9  
 - Device model: Samsung S8

_Just installed v20 and this hasn't happened with v19 so I think it's a bug._

_Immediately after, I reinstalled v19 and it played back the same videos with no problems._

If you need more info, please advise on a way that I can contribute further.

### LINKED ISSUE #4541: Accessing details from `play queue` window breaks playback
Thanks to @AnotherLife I was able to reproduce the issue which was annoying me for a long time (since version 20.0). Here is his comment: https://github.com/TeamNewPipe/NewPipe/issues/4513#issuecomment-710558557

### Checklist

- [x] I am using the latest version. 
- [x] I checked, but didn't find any duplicates of this issue in the repo. 

0. Start the app with an empty play queue
1. Enqueue a youtube video in background (for example via push-and-hold menu entry in history)
2. Go to play queue by touching android player/notification
3. Press and hold on current track entry and select `details`
4. Try to seek to an arbitrary position in the player in the android notification panel (or by using controls in play queue window in the bottom).

### Actual behaviour

Playback stopped and pause button icon changed to loading icon. Seeking to other places or playing other tracks in a queue doesn't help.

### Expected behavior

Playback continues normally, as if I didn't access details from play queue.

### Device info

 - Android version/Custom ROM version: Android 10, magic UI 3.1.0
 - Device model: Honor 20 (YAl-L21)

### Important note
The main problem is that playback breaks even if I don't try to seek forward/backward (it's just a way to quickly reproduce the issue), it's enough just to access details once. And I have to completely wipe my play queue to get rid of frozen playback and lose all my history and queued tracks because of this.

### LINKED ISSUE #4712: Player: app does not show bottom bar player after exiting NewPipe with active background player
### Checklist

- [x] I am using the latest version - x.xx.x 

1. Play a song in background
2. Exit NewPipe with back button
3. Open back NewPipe

### Actual behaviour

When reopen NewPipe  while the background player is active, NewPipe does not show the bottom player bar to indicate active playing. 

### Expected behavior

When reopen NewPipe with a background player active, the NewPipe app should show the bottom player bar like how it is before the app was closed. 

Screenshot_20201030-030829

### Logs

_none_

### Device info

 - Android version/Custom ROM version: 9/Havoc 2.8
 - Device model: SM-N910G

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Unify all ways of opening the VideoDetailFragment into two methods: `openVideoDetailFragment()` which sends the new data directly to the fragment manager, `openVideoDetail()` which sends an `Intent` received by `MainActivity` which in turn calls `openVideoDetailFragment()`.
- The arguments taken by `openVideoDetailFragment()` have changed: `autoPlay` is not there anymore, and `switchingPlayers` was added to be able to always determine correctly if the function was called as a way to switch players. `autoPlay` is now calculated based on the user's preference, on `switchingPlayers` and on whether there is already a background/popup player running. For more info see [these lines](https://github.com/TeamNewPipe/NewPipe/pull/4562/files#diff-7c684dc26bc21808937b0cd7c85de9810279b15b2ba4fdd2e31a0ad2c9e94962R347-R361). This makes it far more clear when the mini player will start playing (I think some situations were handled correcly only by accident before)
- Disable player stream preloading only if the current stream is going to be replaced for sure (see [this](https://github.com/TeamNewPipe/NewPipe/pull/4562/files#diff-b2e369e39e24aea82f68bcb17acd1135173d776a871986ac8ef85ad73e767333R871-R875)). `equals()` was implemented for `PlayQueueItem`s, so that (only) the url is compared when checking them.
- Add `@Nullable` and `@NotNull` to some variables in `VideoDetailFragment`, allowing me to enforce or remove some null checks. Rename `name` to `title` for consistency.
- Remove `static` qualifier from `currentInfo` and `playQueue` in `VideoDetailFragment`: they were never accessed as static members, and having them static caused random problems while replacing the `VideoDetailFragment` with a new instance of itself.
- [Do not force reopen](https://github.com/TeamNewPipe/NewPipe/pull/4562/files#diff-7c684dc26bc21808937b0cd7c85de9810279b15b2ba4fdd2e31a0ad2c9e94962R510-R511) `MainActivity` when sharing something to NewPipe and pressing on "Show info". This was introduced in 67d2b9131e7871505454d12dd1d245528b251228 by @yausername while handling urls in comments, but at that point there was no unified player, so probably without that it didn't work. But now everything works good even by keeping the current `MainActivity` instance, and this yields many improvements: opening urls in descriptions in NewPipe now does not erase backstack, nor does clicking `Open details` on queue items, and probably something else. :-D 
- Replace "Video player" with "Show info" when sharing a video to NewPipe but the background or popup players are running, and thus `autoPlay` would be `false`.
- Handle "Video player" in the exact same way as "Show info" for streams, i.e. do not start a service just to fetch the info, since `VideoDetailFragment` can fetch data itself without issues.
- When sharing a video to NewPipe and a play button is pressed, replace the current queue instead of enqueueing. Enqueueing can still be achieved opening info and then enqueueing, but enqueueing directly was bad UX since the user would not see what happened.
- When opening NewPipe from scratch but a player is already running, show the bottom mini player #4712
- Make sure play/pause state is preserved when switching players via the specific buttons. Also replace `START_PAUSED` with `PLAY_WHEN_READY` for better consistency (i.e. not inverting twice).
- Move menu listener to switch players from `BackgroundPlayerActivity` to `ServicePlayerActivity`, thus removing a useless abstract function.
- Remove [some unused code](https://github.com/TeamNewPipe/NewPipe/pull/4562/files#diff-01c2c8e04aff70b98bd2a94267e903d313f32a2e1df712a023e56ace95c9bf62R556-R557) in `RouterActivity`
- Refactor `getPlayerIntent()` functions in `NavigationHelper` and removed unused navigation functions.
- Remove testing-only code in [PlayQueue](https://github.com/TeamNewPipe/NewPipe/pull/4562/files#diff-00e000f7dc0e535b7d97ab5d6056c88e7191a7480725fb20ccd4c49e36b4aabc) `reportingReactor` was used to capture and log broadcast events in debug builds. This can be undone if unwanted.
- Remove unused preference `autoplay_through_intent_key` (left there by accident after 2907, probably)
- Some refactoring here and there: fix warnings in `openDownloadDialog`, in `openDownloads`, in `PlayerHelper` and more

Fixes #4541 
Probably fixes #4432
Fixes #4403
Fixes (maybe) #4687
Fixes #4712

#### APK testing 
Since this PR changes the logic of many things, the more people can test, the better.
@opusforlife2 @Jean757 @shadow00 @nbmrjuhneibkr could you test if the new share behaviour suits you, i.e. #4403?
@AnotherLife @pew-pew could you test if there are no more issues with playback breaking after opening details, i.e. #4541?
@aand18 @Medard22 @UserX404 @ventilaar @floral-qua-floral @Idiotten could you test if this solves #4432, i.e. background playback stopping at random after some (constant) time?
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5476731/app-debug.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -88,6 +88,7 @@ import org.schabi.newpipe.views.FocusOverlayView;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import static org.schabi.newpipe.util.Localization.assureCorrectAppLanguage;
 
@@ -822,15 +823,23 @@ public class MainActivity extends AppCompatActivity {
     }
 
     private void openMiniPlayerUponPlayerStarted() {
+        if (getIntent().getSerializableExtra(Constants.KEY_LINK_TYPE)
+                == StreamingService.LinkType.STREAM) {
+            // handleIntent() already takes care of opening video detail fragment
+            // due to an intent containing a STREAM link
+            return;
+        }
+
         if (PlayerHolder.isPlayerOpen()) {
-            // no need for a broadcast receiver if the player is already open
+            // if the player is already open, no need for a broadcast receiver
             openMiniPlayerIfMissing();
         } else {
-            // listen for player intents being sent around
+            // listen for player start intent being sent around
             broadcastReceiver = new BroadcastReceiver() {
                 @Override
                 public void onReceive(final Context context, final Intent intent) {
-                    if (intent.getAction().equals(VideoDetailFragment.ACTION_PLAYER_STARTED)) {
+                    if (Objects.equals(intent.getAction(),
+                            VideoDetailFragment.ACTION_PLAYER_STARTED)) {
                         openMiniPlayerIfMissing();
                         // At this point the player is added 100%, we can unregister. Other actions
                         // are useless since the fragment will not be removed after that.
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -69,6 +69,7 @@ import org.schabi.newpipe.fragments.detail.VideoDetailFragment;
 import org.schabi.newpipe.fragments.list.search.SearchFragment;
 import org.schabi.newpipe.player.VideoPlayer;
 import org.schabi.newpipe.player.event.OnKeyDownListener;
+import org.schabi.newpipe.player.helper.PlayerHolder;
 import org.schabi.newpipe.player.playqueue.PlayQueue;
 import org.schabi.newpipe.report.ErrorActivity;
 import org.schabi.newpipe.util.Constants;
@@ -152,7 +153,7 @@ public class MainActivity extends AppCompatActivity {
         if (DeviceUtils.isTv(this)) {
             FocusOverlayView.setupFocusObserver(this);
         }
-        setupBroadcastReceiver();
+        openMiniPlayerUponPlayerStarted();
     }
 
     private void setupDrawer() throws Exception {
@@ -809,34 +810,39 @@ public class MainActivity extends AppCompatActivity {
         }
     }
 
-    private void setupBroadcastReceiver() {
-        broadcastReceiver = new BroadcastReceiver() {
-            @Override
-            public void onReceive(final Context context, final Intent intent) {
-                if (intent.getAction().equals(VideoDetailFragment.ACTION_PLAYER_STARTED)) {
-                    final Fragment fragmentPlayer = getSupportFragmentManager()
-                            .findFragmentById(R.id.fragment_player_holder);
-                    if (fragmentPlayer == null) {
-                        /*
-                         * We still don't have a fragment attached to the activity.
-                         * It can happen when a user started popup or background players
-                         * without opening a stream inside the fragment.
-                         * Adding it in a collapsed state (only mini player will be visible)
-                         * */
-                        NavigationHelper.showMiniPlayer(getSupportFragmentManager());
+    private void openMiniPlayerIfMissing() {
+        final Fragment fragmentPlayer = getSupportFragmentManager()
+                .findFragmentById(R.id.fragment_player_holder);
+        if (fragmentPlayer == null) {
+            // We still don't have a fragment attached to the activity. It can happen when a user
+            // started popup or background players without opening a stream inside the fragment.
+            // Adding it in a collapsed state (only mini player will be visible).
+            NavigationHelper.showMiniPlayer(getSupportFragmentManager());
+        }
+    }
+
+    private void openMiniPlayerUponPlayerStarted() {
+        if (PlayerHolder.isPlayerOpen()) {
+            // no need for a broadcast receiver if the player is already open
+            openMiniPlayerIfMissing();
+        } else {
+            // listen for player intents being sent around
+            broadcastReceiver = new BroadcastReceiver() {
+                @Override
+                public void onReceive(final Context context, final Intent intent) {
+                    if (intent.getAction().equals(VideoDetailFragment.ACTION_PLAYER_STARTED)) {
+                        openMiniPlayerIfMissing();
+                        // At this point the player is added 100%, we can unregister. Other actions
+                        // are useless since the fragment will not be removed after that.
+                        unregisterReceiver(broadcastReceiver);
+                        broadcastReceiver = null;
                     }
-                    /*
-                    * At this point the player is added 100%, we can unregister.
-                    * Other actions are useless since the fragment will not be removed after that
-                     * */
-                    unregisterReceiver(broadcastReceiver);
-                    broadcastReceiver = null;
                 }
-            }
-        };
-        final IntentFilter intentFilter = new IntentFilter();
-        intentFilter.addAction(VideoDetailFragment.ACTION_PLAYER_STARTED);
-        registerReceiver(broadcastReceiver, intentFilter);
+            };
+            final IntentFilter intentFilter = new IntentFilter();
+            intentFilter.addAction(VideoDetailFragment.ACTION_PLAYER_STARTED);
+            registerReceiver(broadcastReceiver, intentFilter);
+        }
     }
 
     private boolean bottomSheetHiddenOrCollapsed() {
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -30,9 +30,7 @@ import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
-import androidx.preference.PreferenceManager;
 import android.util.Log;
-
 import android.view.KeyEvent;
 import android.view.LayoutInflater;
 import android.view.Menu;
@@ -46,6 +44,7 @@ import android.widget.FrameLayout;
 import android.widget.ImageView;
 import android.widget.Spinner;
 import android.widget.TextView;
+
 import androidx.annotation.NonNull;
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.ActionBarDrawerToggle;
@@ -55,6 +54,7 @@ import androidx.core.view.GravityCompat;
 import androidx.drawerlayout.widget.DrawerLayout;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
+import androidx.preference.PreferenceManager;
 
 import com.google.android.material.bottomsheet.BottomSheetBehavior;
 import com.google.android.material.navigation.NavigationView;
@@ -71,8 +71,8 @@ import org.schabi.newpipe.player.VideoPlayer;
 import org.schabi.newpipe.player.event.OnKeyDownListener;
 import org.schabi.newpipe.player.playqueue.PlayQueue;
 import org.schabi.newpipe.report.ErrorActivity;
-import org.schabi.newpipe.util.DeviceUtils;
 import org.schabi.newpipe.util.Constants;
+import org.schabi.newpipe.util.DeviceUtils;
 import org.schabi.newpipe.util.KioskTranslator;
 import org.schabi.newpipe.util.Localization;
 import org.schabi.newpipe.util.NavigationHelper;
@@ -758,32 +758,36 @@ public class MainActivity extends AppCompatActivity {
             if (intent.hasExtra(Constants.KEY_LINK_TYPE)) {
                 final String url = intent.getStringExtra(Constants.KEY_URL);
                 final int serviceId = intent.getIntExtra(Constants.KEY_SERVICE_ID, 0);
-                final String title = intent.getStringExtra(Constants.KEY_TITLE);
-                switch (((StreamingService.LinkType) intent
-                        .getSerializableExtra(Constants.KEY_LINK_TYPE))) {
+                String title = intent.getStringExtra(Constants.KEY_TITLE);
+                if (title == null) {
+                    title = "";
+                }
+
+                final StreamingService.LinkType linkType = ((StreamingService.LinkType) intent
+                        .getSerializableExtra(Constants.KEY_LINK_TYPE));
+                assert linkType != null;
+                switch (linkType) {
                     case STREAM:
-                        final boolean autoPlay = intent
-                                .getBooleanExtra(VideoDetailFragment.AUTO_PLAY, false);
-                        final String intentCacheKey = intent
-                                .getStringExtra(VideoPlayer.PLAY_QUEUE_KEY);
+                        final String intentCacheKey = intent.getStringExtra(
+                                VideoPlayer.PLAY_QUEUE_KEY);
                         final PlayQueue playQueue = intentCacheKey != null
                                 ? SerializedCache.getInstance()
                                 .take(intentCacheKey, PlayQueue.class)
                                 : null;
-                        NavigationHelper.openVideoDetailFragment(getSupportFragmentManager(),
-                                serviceId, url, title, autoPlay, playQueue);
+
+                        final boolean switchingPlayers = intent.getBooleanExtra(
+                                VideoDetailFragment.KEY_SWITCHING_PLAYERS, false);
+                        NavigationHelper.openVideoDetailFragment(
+                                getApplicationContext(), getSupportFragmentManager(),
+                                serviceId, url, title, playQueue, switchingPlayers);
                         break;
                     case CHANNEL:
                         NavigationHelper.openChannelFragment(getSupportFragmentManager(),
-                                serviceId,
-                                url,
-                                title);
+                                serviceId, url, title);
                         break;
                     case PLAYLIST:
                         NavigationHelper.openPlaylistFragment(getSupportFragmentManager(),
-                                serviceId,
-                                url,
-                                title);
+                                serviceId, url, title);
                         break;
                 }
             } else if (intent.hasExtra(Constants.KEY_OPEN_SEARCH)) {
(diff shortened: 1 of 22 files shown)


Medard22:
So far seems it seems good. I didn´t encounter any problems with the playback. I will test it thoroughly tommorow.

ventilaar:
Tested for issue #4432, fixed for me. Thank you

avently:
> Disable player stream preloading only if the current stream is going to be replaced for sure (see this ) @avently is this ok?

I don't think so. 
Why did you add all those checks?
```java
newQueue != null && playQueue != null
!Objects.equals(newQueue.getItem(), playQueue.getItem())
```

What's the reason for them to exist? They will not fix the issue with preloading of background & popup because the fix would be to apply disabled preloading for `player.videoPlayerSelected()`. In your case you probably will see a situation when `Objects.equals(newQueue.getItem(), playQueue.getItem())` is always true even for the same streams because of different restore position (or maybe I'm wrong, don't remember the contents of playqueueitem). But it will not fix the problem with preloading anyway.

Why did you delete lines after https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L858 ? They were here for a reason. If you found another way to do the same thing, then ok.

You made here many good things (based on description). It's hard to see all new code on a phone but fixes and futures you introduced is very important!

avently:
Ok. Some things I found:
- the result of removing the code I mentioned and objects.equals checks:
Open any video in audio player
Tap on notification
Choose to play in main player in three dots menu
Player is not changed to main, still audio only (problem)
If you tap on notification and choose another time on timeline - endless loading - another problem.

- try to load some videos one by one (using related videos), play some of them, then try to press back and (with autorotation enabled) rotate the phone into another orientation. Try this step with backpress and orientation change multiple times and you'll see that after rotation there is another video (not the one before rotation). It's probably because if removing static modifiers or maybe it was like this even in 0.20.0, didn't check.

AnotherLife:
The "popup player stopping after viewing video details" problem does not happen with this apk from what I tested.

Idiotten:
I have not encountered #4432 after a couple hours of queuing and such. Seems fixed for me.

opusforlife2:
Yay! So many good changes! Gonna test now.

opusforlife2:
Does this not include the decryption retry code? I got that error just now.

TobiGr:
@opusforlife2 This branch is not based on the latest dev. that fix is not included

opusforlife2:
Can confirm it fixes #4403. 👍

MD77MD:
>When sharing a video to NewPipe and a play button is pressed, replace the current queue instead of enqueueing. Enqueueing can still be achieved opening info and then enqueueing, but enqueueing directly was bad UX since the user would not see what happened.

this is great if share to player, but please make it that it sould enqueue when sharing to pup-up or background player.... or a separate enqueue option

Stypox:
> This is the only way to enqueue videos from other apps to newPipe

As I explained, this is not true, since now if there is a video already playing the "Show info" button is shown, and from video details you can just long press on background or popup

Stypox:
@avently About the preloading part, I then am not sure I understand when preloading should be disabled. From what I understood so far: when the current stream is being dismissed in favour of a new one, then preloading can be disabled for that track since it is soon going to be replaced. Therefore, when switching queues, we should make sure the newly selected video is different from the current one, otherwise preloading would be disabled forever. Ok, maybe if the video is the same and the timestamp differs we could disable preloading, but that's just a special case, so I wouldn't add complexity for that for now. 

> the result of removing the code I mentioned and objects.equals checks:

Do you mean that by removing my conditions for preloading it creates the explained issues, or that my code creates those?

> Why did you delete lines after https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L858 ? They were here for a reason. If you found another way to do the same thing, then ok.

If you are talking about "Situation when user switches from players to main player. All needed data is here, we can start watching" then yeah, I am handling the switching players case directly [here](https://github.com/Stypox/NewPipe/blob/fix-detail-open/app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java#L388)

Stypox:
Another apk with the same changes but with updated extractor: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5467484/app-debug.zip)

avently:
@Stypox here I provided steps for reproduce problems. You said that you made another way for switching players so that way should be used in order to switch to main player from BackgroundActivity like in my reproducible example. Because now you're not switching players when moving to video details from backgroundactivity.

About preloading part:
I think you just need to make sure it's actually works. If it is then current code is ok. But again in my reproducible example the code is not working properly. 

Did you checked with logs, for example, that your code for preloading really getting called and not just skipped when one stream changes to another?

MD77MD:
> @MD77MD As I explained, this is not true, since now if there is a video already playing the "Show info" button is shown, and from video details you can just long press on background or popup

@Stypox that is true but please bear with me. lets say you did a Google search in browser and found a number of useful videos... now you have to open the details page of the video then long press to enqueue then go back the search page and do this all over again for each video... can you imagine how much time, effort and data is wasted going back and forth. all off this while the video pup-up is in your face, moving it up and down because its abstracting the button you want to tap.

I'm not criticizing, i really like what your doing. but  i wish if you would reconsider finding a work around for this please . perhaps a tiny window asking to show details or enqueue. 

or we could just use the regular share window but utilizing the new enqueue option from #4425. This would add an extra stab, but a light and easy one.

anyway I'm really thankful to everything you've do here.

opusforlife2:
@MD77MD Set Preferred open action to Always Ask, then long press on background or popup to enqueue.

Stypox:
@avently @TobiGr @opusforlife2 @B0pol I added more changes, and fixed the issue reported by avently:
- **also have the popup fullscreen button and the player queue switch-to-main button follow the new method**
- `equals()` was implemented for PlayQueueItems, so that (only) the url is compared when checking them.
- Make sure play/pause state is preserved when switching players via the specific buttons. Also replace `START_PAUSED` with `PLAY_WHEN_READY` for better consistency (i.e. not inverting twice).
- Move menu listener to switch players from `BackgroundPlayerActivity` to `ServicePlayerActivity`, thus removing a useless abstract function.
- Refactor `getPlayerIntent()` functions in `NavigationHelper` and removed unused navigation functions.
- Fix some warnings in `PlayerHelper`

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5469727/app-debug.zip)

avently:
Changes sound good, the code for new commit looks good too. I always like optimizations:)

MD77MD:
> @MD77MD Set Preferred open action to Always Ask, then long press on background or popup to enqueue.

@opusforlife2 I did what you said, the menu did show up but when long pressing on the pop-up option nothing happens.... I wish this would work but it's not working... could you try it yourself. i used the apk from stypox last comment.

opusforlife2:
@MD77MD Apparently, you need to tap the option once so that it is highlighted, then long press and leave it. A bit unintuitive, which should be improved.

opusforlife2:
@Stypox 2 of the 3 linked issues are reported to be fixed. What else needs to be tested, exactly?

Stypox:
I guess a last round of tests about the things I changed would be enough, just to be sure the new behaviour is not broken. Other than that, yeah, I think this is ready

opusforlife2:
> I guess a last round of tests about the things I changed would be enough

@Stypox But... but the stuff you've changed is all programming jargon. :( That's why I don't know what to test.

Stypox:
@opusforlife2 just opening the video detail fragment under many different conditions is enough

opusforlife2:
> opening urls in descriptions in NewPipe now does not erase backstack, nor does clicking Open details on queue items, and probably something else. :-D

Yay!

opusforlife2:
@Stypox With Autoplay off, tapping on the thumbnail to play is broken now. It opens a notification which buffers forever, then does nothing but stay on the screen. If you open a related video and tap the thumbnail, it says "could not play this stream".

If you go back to the previous video and tap the thumbnail, this time it opens the video, but paused at the beginning. Manually tapping the play button finally makes it play.

Stypox:
@opusforlife2 fixed, thank you for testing
@leewp14 I added a check that opens the mini player on app open if there is a player running in background/popup. Could you test if this apk solves #4712?
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5471719/app-debug.zip)

MD77MD:
> @MD77MD Apparently, you need to tap the option once so that it is highlighted, then long press and leave it. A bit unintuitive, which should be improved.

just to be clear we are at the same page, long pressing is supposed to enqueue video. if so, in my case long pressing does nothing, and when i release my finger it just replace the already playing video and doesn't enqueue. 

if it not to much to ask, could you make a small video of enqueueing two or three videos in pop-up mode. maybe im doing something wrong. 

Anyway, thank you @opusforlife2 for trying to help. i really appreciate it 🌷😊

opusforlife2:
@Stypox Confirmed fixed.

> I added a check that opens the mini player on app open if there is a player running in background/popup.

That was really bugging me! Thanks! It even works if you have nothing playing but the mini player is still visible, like when you have video details open. You shouldn't have to lose your backstack just by an accidental app switch.

opusforlife2:
@MD77MD If you could come on IRC this would go much, _much_ faster. Make a Matrix account and join #freenode_#newpipe:matrix.org, dude.

Are you trying to long press on Background or Popup? Those are the only two where this works.

Stypox:
An apk containing both VideoDetailFragment fixes TeamNewPipe/NewPipe#4562 and deobfuscation improvements TeamNewPipe/NewPipeExtractor#447, in case someone (@opusforlife2) wants to test both at the same time: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5473958/app-debug.zip)

Medard22:
> An apk containing both VideoDetailFragment fixes #4562 and deobfuscation improvements [TeamNewPipe/NewPipeExtractor#447](https://github.com/TeamNewPipe/NewPipeExtractor/pull/447), in case someone (@opusforlife2) wants to test both at the same time: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5473958/app-debug.zip)

This build is consistently crashing on my device when going from popup player to fullscreen. 

Screenshot_20201102-150441

Stypox:
@Medard22 thank you for testing! Does this apk fix the issue? @opusforlife2 is also does not contain better-deobfuscation anymore.
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5476731/app-debug.zip)

Medard22:
@Stypox Yes it is fixed now. Thanks.

Also I noticed one thing, but I am not sure if it is directly related to this issue. When I have a queue in the background player and then run any video in the foreground the queue will be emptied. Is that intended? Sorry for the possible offtopic.

opusforlife2:
@Medard22 There is an option in the settings to warn you in case that happens, letting you choose what to do.

Medard22:
@opusforlife2 Oops, I totally missed that. Thank you.

opusforlife2:
Bug: Volume and brightness gesture areas are swapped in this APK. Volume is on the left, but it should be on the right.

Stypox:
Why? I didn't touch anything about that... it is probably caused by the recently merged improvements to gestures by @vkay94 and should be a really small fix (i.e. inverting an if)

opusforlife2:
Yeah, it's probably that. I didn't test that PR.

vkay94:
@Stypox The swap lies in my PR indeed :( 
But it can be fixed easily here: https://github.com/Stypox/NewPipe/blob/97f149b0da4c02518496b46fb3f6a79f8ecaaad6/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java#L112

Sorry,

Stypox:
@vkay94 there is nothing to be sorry about, accidental problems always happen ;-)

vkay94:
Yeah, I know, but I through I checked it carefully :) @Stypox Can you add the fix with this PR? Simply replace both methods with each other ;)

Stypox:
@vkay94 is this ok?
@opusforlife2 [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5488555/app-debug.zip)

opusforlife2:
Fixed.

Stypox:
I rebased and tested the rebased version all day with multiple situations and everything worked flawlessly: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5506813/app-debug.zip)

The only issue I found, that is present even in NewPipe 0.20.2, and is thus not caused by this PR, is:
- enqueue a random video in the background
- open a local playlist, a remote playlist, a channel or the history fragment
- tap on "Play all"
- the main player does not start playing right away (as it would be expected), instead the video details of the first item of the playlist are shown without playing and the rest of the queue is not preserved

Stypox:
I rebased and tested the rebased version all day with multiple situations and everything worked flawlessly: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5506813/app-debug.zip)

The only issue I found, that is present even in NewPipe 0.20.2, and is thus not caused by this PR, is:
- enqueue a random video in the background
- open a local playlist, a remote playlist, a channel or the history fragment
- tap on "Play all"
- the main player does not start playing right away (as it would be expected), instead the video details of the first item of the playlist are shown without playing and the rest of the queue is not preserved

Since the above issue is not caused by this PR I will take care of fixing it in a future one, so I think this can be merged @TobiGr

B0pol:
> The only issue I found, that is present even in NewPipe 0.20.2, and is thus not caused by this PR

Then please open a new issue

Stypox:
@B0pol I was going to ;-)

TobiGr (approved):

jay00128:
#4678. my issue ：video names on Youtube are Chinese，but on NewPipe they became english titles，maybe it's because of the community contributions 。🤔

Stypox:
@jay00128 please open an issue in english about what you are experiencing

89z:
https://github.com/TeamNewPipe/NewPipe/issues/5530
https://github.com/TeamNewPipe/NewPipe/issues/4779

https://github.com/TeamNewPipe/NewPipe/pull/4562

89z:
https://github.com/TeamNewPipe/NewPipe/issues/4779
https://github.com/TeamNewPipe/NewPipe/issues/5541

https://github.com/TeamNewPipe/NewPipe/pull/4562

PR #4566, opened 2020-10-19
Use Objects methods in the extractor.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use `Objects` methods in the extractor.

- https://github.com/TeamNewPipe/NewPipeExtractor/pull/419

#### APK testing 

debug.zip

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -171,7 +171,7 @@ dependencies {
 
     // NewPipe dependencies
     // You can use a local version by uncommenting a few lines in settings.gradle
-    implementation 'com.github.TeamNewPipe:NewPipeExtractor:350eed6214b93255d788dfa208b1e9a5e5da91e6'
+    implementation 'com.github.Isira-Seneviratne:NewPipeExtractor:Use_Objects_methods-SNAPSHOT'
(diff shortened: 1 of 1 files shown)


Stypox:
This is unneeded if the API of the Extractor exposed to the app didn't change

Isira-Seneviratne:
> This is unneeded if the API of the Extractor exposed to the app didn't change

In that case, shall I close this? And should the create PR option in the Extractor PR be unchecked?

opusforlife2:
> I agree to create a pull request for NewPipe as soon as possible to make it compatible with the changed API.

If the API didn't change, you don't need to open a PR. But you can still leave the box checked because you're not refusing it, you just don't need it.

Isira-Seneviratne:
> > I agree to create a pull request for NewPipe as soon as possible to make it compatible with the changed API.
> 
> If the API didn't change, you don't need to open a PR. But you can still leave the box checked because you're not refusing it, you just don't need it.

Okay, thanks!

PR #4578, opened 2020-10-21
Use Dagger.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Make use of [Dagger](https://dagger.dev/)'s dependency injection functionality.

#### APK testing 

[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5611557/debug.zip)

#### Due diligence


wb9688:
What's the difference in size for the release APK?

Isira-Seneviratne:
> What's the difference in size for the release APK?

I found it to be roughly 0.1 MB larger than it would have been without the change.

wb9688:
Hmm… OK. Are there more places where we'd be able to use Dagger? Currently you seem to only use it for the database, right?

Isira-Seneviratne:
> Hmm… OK. Are there more places where we'd be able to use Dagger?

It could be used to inject `CompositeDisposable`s, but I'm sure there are other places as well.

> Currently you seem to only use it for the database, right?

Yeah.

Stypox:
Could you explain the purpose of Dagger and why you'd like NewPipe to have it?

Isira-Seneviratne:
> Could you explain the purpose of Dagger and why you'd like NewPipe to have it?

Dagger is a dependency injection framework, and it reduces coupling by automatically handling the dependency creation instead of having to instantiate them manually. This makes testing easier, e.g. a mock implementation of a DAO could be passed to a class that relies on it.

Stypox:
Oh ok, I see, thank you for the explanation. Though I don't like much the fact that this adds 500 lines of code and 0.1MB to the APK

XiangRongLin:
@Isira-Seneviratne  
> This makes testing easier, e.g. a mock implementation of a DAO could be passed to a class that relies on it

Could you add some unit test to showcase that. I currently have trouble seing how i would pass in mocks if needed. It would also help others, by having an example to go by.

And i don't think changing everything at once is a good idea with the current state of automated tests in NewPipe. The problems with the unified-player are still ongoing and will probably take very long to fix.

app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    @Test
+    fun testCreatePlaylist() {
+        localPlaylistManager.createPlaylist("Playlist 1", listOf()).test()
+            .assertNoValues()
+
+        val streamEntity1 = StreamEntity(0, 0, "", "", StreamType.NONE, 0, "")
+        val streamEntity2 = StreamEntity(1, 0, "", "", StreamType.NONE, 0, "")
+        localPlaylistManager.createPlaylist("Playlist 2", listOf(streamEntity1, streamEntity2)).test()
+            .assertNoValues()
```
XiangRongLin, line 39: I'll preface it by saying that i have no clue about flowable and how to test them, but this test seems wrong.
From reading the method one would expect values. From playing around a bit it seems seems that `test().await().assertValues(xyz)` should be the way.

app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.schabi.newpipe.TestApp
+import org.schabi.newpipe.database.AppDatabase
+import org.schabi.newpipe.database.stream.model.StreamEntity
+import org.schabi.newpipe.extractor.stream.StreamType
+import javax.inject.Inject
+
+@RunWith(AndroidJUnit4::class)
+class LocalPlaylistManagerTest {
+    @Inject
+    lateinit var appDatabase: AppDatabase
```
XiangRongLin, line 17: This is the actual implementation and not a mock/fake right?

How would i go about mocking this, so that i could convert it to an unit test, instead of instrumented. Same for when i want to run different `Mockito.verify()` or `Mockito.when()` depending on the test

app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
```diff
@@ -0,0 +1,41 @@
...
+    @Inject
+    lateinit var localPlaylistManager: LocalPlaylistManager
```
XiangRongLin, line 19: With `LocalPLaylistManager` being the class that is tested, i would not inject it. It is supposed to be tightly coupled

app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
Isira-Seneviratne: I created a `TestRoomModule` where the testing version of the database is created, so the actual database implementation can be replaced with a mock one, and Dagger will inject the mock version.

app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
Isira-Seneviratne: Okay, thanks.

app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
Isira-Seneviratne: It uses the testing version of the database.

app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
XiangRongLin: Doesn't this then mean, that every test will have to use mocks, taking away the possibility to use the real implementation in only some tests?

Also some tests only need a stub and can safe some memory, others need a full spy. With one global place to define it, one can't fine tune that.

Isira-Seneviratne:
Mock versions of the classes being tested can be assigned if needed.

Stypox:
I am sorry, but I still do not understand why this should be used inside NewPipe. I don't have any experience with this, so my opinion might not count at all.
These are the cons I see:
- increases APK size
- increases number of lines of code
- increases code complexity, since Dagger is not standard Android, leading to many more things that one needs to know and remember in order to develop NewPipe. Furthermore, PRs would be more troublesome to review, since I guess most people (including me if I don't think about it) would keep using `getDefaultSharedPreferences` instead of using `@Inject`. Also, issues would arise from e.g. forgetting to call `inject(this)`, which is not what one thinks commonly when creating a new Activity/Fragment.
The only pro I see is:
- uses modern code
What is your opinion on this? @Isira-Seneviratne @TobiGr @B0pol @vkay94 @TacoTheDank @XiangRongLin

Isira-Seneviratne:
There's a library that extends Dagger for Android called [Hilt](https://developer.android.com/training/dependency-injection/hilt-android), but it's still in alpha.

Isira-Seneviratne:
I opened [a separate PR](https://github.com/TeamNewPipe/NewPipe/pull/5182).

TacoTheDank:
I'm iffy about it; I'm not very familiar with Dagger either. Personally I'm not much a fan of it adding 500 lines lol

PR #4588, opened 2020-10-21
Enable lint

apoorv-arora:
#### What is it?
- Codebase improvement (dev facing)

Current project has lint disabled. As a result of which, any new commit which is scheduled to go live may not pass the list of checks provided by android lint. To learn the list of checks that are done by lint, navigate here.

This PR enables lint workflow integration by creating a baseline file which includes all errors (1) & warnings (89) in the current project & treat them as a baseline. Any new commit will now throw an error if they do not follow the lint standard guideline.
#### Implementation
1) ./gradlew lintDebug


TheAssassin:
I'm not sure if merging is a good idea. You're adding a file with 27k lines of XML. It's pretty much impossible to review this. You don't explain how you added created this file (I'm pretty sure you didn't hand craft it), so it remains unclear how this linting system would have to be maintained in the future by the team.

Please elaborate. Also, it would be better to open an issue first which you can then reference here. Please create one. Quoting the contribution guidelines:

> If you want to add a feature or change one, please open an issue describing your change. This gives the team and community a chance to give feedback before you spend any time on something that could be done differently or not done at all. It also prevents two contributors from working on the same thing and one being disappointed when only one user's code can be added.

apoorv-arora:
> I'm not sure if merging is a good idea. You're adding a file with 27k lines of XML. It's pretty much impossible to review this. You don't explain how you added created this file (I'm pretty sure you didn't hand craft it), so it remains unclear how this linting system would have to be maintained in the future by the team.
> 
> Please elaborate. Also, it would be better to open an issue first which you can then reference here. Please create one. Quoting the contribution guidelines:
> 
> > If you want to add a feature or change one, please open an issue describing your change. This gives the team and community a chance to give feedback before you spend any time on something that could be done differently or not done at all. It also prevents two contributors from working on the same thing and one being disappointed when only one user's code can be added.

We can take a snapshot of the project's current set of warnings & errors specified by lint, and then use the snapshot as a baseline for future inspection runs so that only new issues are reported. The baseline snapshot will let us start using lint to fail the build without having to go back and address all existing issues first. All the subsequent checks are will have to go through lint now.
By specifying the file name in build.gradle, & running (./gradlew lintDebug) we can generate the baseline file.

Stypox:
Please fix the errors and warnings instead of adding them to the blacklist: that's what pther PRs adding lint did.

litetex:
Closing this for now:
* no progress
* no GitHub actions build
* merge conflicts

Feel free to reopen it when there is progress again.

PR #4652, opened 2020-10-26
Convert notification actions to a custom preference

### LINKED ISSUE #4646: [Codebase] Non standard notification setting layout xml
Is there a specific reason why the notification setting xml is defined as a custom view in folder /res/layout not using the standard `<PreferenceScreen>` tag like the other settings in folder /res/xml.
Because this custom definition brings a lot of boilerplate when adding a simple switch setting.
Just compare my 2 PR
appearance setting: #4642
notification setting: #4643

XiangRongLin:
#### What is it?

- Convert `NotificationSettingsFragment` into a custom `Preference`
  - extract scale switch out of this
- Create "standard" preference with scale switch and custom preference above

- closes  #4646

#### APK testing 
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5443584/app-debug.zip) (updated with Stypox suggestions)

### Site notes
- `Preference.onDetached()` seems to equivalent to `Fragment.onPause()` for the purpose of saving the preference and recreating the notification.
- From the doc on the 4 constructors of `Preference`, only modifying the one that i did should be enogh (I think).

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/custom/NotificationActionsPreference.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/custom/NotificationActionsPreference.java
@@ -1,10 +1,10 @@
-package org.schabi.newpipe.settings;
+package org.schabi.newpipe.settings.custom;
 
+import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.graphics.drawable.Drawable;
-import android.os.Bundle;
-import android.preference.PreferenceManager;
+import android.util.AttributeSet;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
@@ -13,18 +13,16 @@ import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.RadioButton;
 import android.widget.RadioGroup;
-import android.widget.Switch;
 import android.widget.TextView;
 import android.widget.Toast;
-
 import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.content.res.AppCompatResources;
 import androidx.core.graphics.drawable.DrawableCompat;
 import androidx.core.widget.TextViewCompat;
-import androidx.fragment.app.Fragment;
-
+import androidx.preference.Preference;
+import androidx.preference.PreferenceViewHolder;
+import java.util.List;
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.player.MainPlayer;
 import org.schabi.newpipe.player.NotificationConstants;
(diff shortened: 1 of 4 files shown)


app/src/main/res/xml/notification_settings.xml
```diff
@@ -0,0 +1,21 @@
...
+                android:defaultValue="false"
+                android:key="@string/scale_to_square_image_in_notifications_key"
+                android:summary="@string/notification_scale_to_square_image_summary"
+                android:title="@string/notification_scale_to_square_image_title"
+                app:iconSpaceReserved="false" />
+
+    </PreferenceCategory>
+
+    <PreferenceCategory>
+        <org.schabi.newpipe.settings.custom.NotificationActionsPreference></org.schabi.newpipe.settings.custom.NotificationActionsPreference>
```
Stypox, line 17:
```suggestion
        <org.schabi.newpipe.settings.custom.NotificationActionsPreference />
```

app/src/main/java/org/schabi/newpipe/settings/custom/NotificationActionsPreference.java
```diff
@@ -1,10 +1,10 @@
-package org.schabi.newpipe.settings;
+package org.schabi.newpipe.settings.custom;
```
Stypox, line 1: Why change package?

app/src/main/res/layout/settings_notification.xml
```diff
@@ -8,71 +8,11 @@
         android:layout_width="match_parent"
```
Stypox, line 8: I think you should remove the `ScrollView` above here, since it is already integrated in the `PreferenceFragment` and it creates strange issues with scroll bars. Also be sure to use `wrap_content` on the root layout's height, otherwise the view can be scrolled more than needed.

app/src/main/res/xml/notification_settings.xml
```diff
@@ -0,0 +1,21 @@
...
+
+    <PreferenceCategory>
+        <SwitchPreferenceCompat
+                android:defaultValue="false"
+                android:key="@string/scale_to_square_image_in_notifications_key"
+                android:summary="@string/notification_scale_to_square_image_summary"
+                android:title="@string/notification_scale_to_square_image_title"
+                app:iconSpaceReserved="false" />
+
+    </PreferenceCategory>
```
Stypox, line 14: Maybe remove `PreferenceCategor`ies or put everything in a single one: now there is some strange spacing above and below the normal preference, so try to play around with this until the layout looks good

Stypox (changes requested):
Thank you! I didn't know this could be done, but it certainly simplifies code ;-)

app/src/main/java/org/schabi/newpipe/settings/custom/NotificationActionsPreference.java
XiangRongLin: There is `DurationListPreference` in that package which is a custom preference.
Since this is also a custom preference i put it in there.

app/src/main/res/xml/notification_settings.xml
XiangRongLin: The reason for the strange spacing was that i forgot to set the layout of PreferenceCategory. I just had in my mind that in the other settings the PreferenceCategory created a visual seperation, which was needed since i removed the divider.

app/src/main/java/org/schabi/newpipe/settings/custom/NotificationActionsPreference.java
Stypox: Ok, good

TobiGr (approved):

XiangRongLin:
Github does have a squash and merge feature right? Otherwise I'm going to squash it manually.

Stypox (approved):

Stypox:
@XiangRongLin thank you! @TobiGr merge, I already added it in release draft

PR #4759, opened 2020-11-01
Change YouTube subscription import instructions to Google takeout

### LINKED ISSUE #4720: Youtube import subscription instructions is not working anymore
# Description 

Following the instructions at https://newpipe.schabi.org/FAQ/tutorials/import-export-data/#import-youtube to download subscriptions no longer works as it seems like Youtube has changed its interface. Going to https://www.youtube.com/subscription_manager?action_takeout=1 no longer starts a download, but it redirects to https://www.youtube.com/feed/channels. This prevents importing of subscriptions list which could have been then loaded to Newpipe app.

# Expected behavior

It should be possible to import subscriptions from youtube account using instructions at https://newpipe.schabi.org/FAQ/tutorials/import-export-data/#import-youtube. 

# Actual Behavior

Instructions given to import subscriptions from Youtube not working anymore. The instructions ask you to download subscriptions using the URL: https://www.youtube.com/subscription_manager?action_takeout=1, but this merely redirects to https://www.youtube.com/feed/channels and does not start a download.

# Steps to reproduce

1. Read import instructions from https://newpipe.schabi.org/FAQ/tutorials/import-export-data/#import-youtube

2. Follow the link at https://www.youtube.com/subscription_manager?action_takeout=1

3. Gets redirected to https://www.youtube.com/feed/channels. No download starts.

4. Cannot import subscriptions from youtube

# Platform

All platforms

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This are the new steps:
1. Go to this URL: https://takeout.google.com/takeout/custom/youtube
2. Log in when asked
3. Click on \"All data included\", then on \"Deselect all\", then select only \"subscriptions\" and click \"OK\"
5. Click on \"Next step\" and then on \"Create export\"
6. Click on the \"Download\" button after it appears and 
7. From the downloaded takeout zip extract the .json file (usually under \"YouTube and YouTube Music/subscriptions/subscriptions.json\") and import it here

Point 7, i.e. extracting things from the archive, can't be done automatically, due to the different structure for different localizations.

Fixes #4720

TeamNewPipe/NewPipeExtractor#452

#### APK testing 
@simula67 @inigosola @Djay86 @IvanProgramming @Saecki 
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5471647/app-debug.zip)

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -176,7 +176,7 @@ dependencies {
 
     // NewPipe dependencies
     // You can use a local version by uncommenting a few lines in settings.gradle
-    implementation 'com.github.TeamNewPipe:NewPipeExtractor:62912ee8349f5d26617c039f337297628ff52ead'
+    implementation 'com.github.Stypox:NewPipeExtractor:501ec30152642ad49ce0a1825410d200942d174c'
     implementation "com.github.TeamNewPipe:nanojson:1d9e1aea9049fc9f85e68b43ba39fe7be1c1f751"
 
     implementation "org.jsoup:jsoup:1.13.1"
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -514,7 +514,7 @@
     <string name="previous_export">Previous export</string>
     <string name="subscriptions_import_unsuccessful">Could not import subscriptions</string>
     <string name="subscriptions_export_unsuccessful">Could not export subscriptions</string>
-    <string name="import_youtube_instructions">Import YouTube subscriptions by downloading the export file:\n\n1. Go to this URL: %1$s\n2. Log in when asked\n3. A download should start (that\'s the export file) </string>
+    <string name="import_youtube_instructions">Import YouTube subscriptions from Google takeout:\n\n1. Go to this URL: %1$s\n2. Log in when asked\n3. Click on \"All data included\", then on \"Deselect all\", then select only \"subscriptions\" and click \"OK\"\n4. Click on \"Next step\" and then on \"Create export\"\n5. Click on the \"Download\" button after it appears and \n6. From the downloaded takeout zip extract the .json file (usually under \"YouTube and YouTube Music/subscriptions/subscriptions.json\") and import it here.</string>
     <string name="import_soundcloud_instructions">Import a SoundCloud profile by typing either the URL or your ID:\n\n1. Enable \"desktop mode\" in a web-browser (the site is not available for mobile devices)\n2. Go to this URL: %1$s\n3. Log in when asked\n4. Copy the profile URL you were redirected to.</string>
     <string name="import_soundcloud_instructions_hint">yourID, soundcloud.com/yourid</string>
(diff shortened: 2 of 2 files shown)


TobiGr:
Instructions should also be changed on the website: https://newpipe.schabi.org/FAQ/tutorials/import-export-data/

https://github.com/TeamNewPipe/website/blob/master/_tutorials/import-export-data.html

Djay86:
Tested the .apk, import worked with no isssue.

IvanProgramming:
Tested .APK file, no issues was found

Saecki:
Works for me too.

Stypox:
I updated the url to https://takeout.google.com/takeout/custom/youtube , see the extractor PR for discussion
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5474147/app-debug.zip)

simula67:
@Stypox The import works with the debug app.

Thanks for doing this.

atar-axis:
Works, the only issue is that google sends you an archive and you will have to extract the json by hand. You should therefore maybe also note that on the instructions or let newpipe extract it if necessary

Stypox:
@atar-axis as I said I don't think we can have NewPipe extract the json automatically, sincs different localizations have different names and structures

atar-axis:
I think all we need to do is finding thw json where it says `"kind" : "youtube#subscription",` should be pretty easy I guess

Stypox:
@TobiGr this should also be merged, it is only a string change

atar-axis:
The instructions don't tell the user to extract the file, find the subscriptions.json on beforehand. At least those instructions should be added before merging

Stypox:
@atar-axis no, it does: `From the downloaded takeout zip extract the .json file (usually under "YouTube and YouTube Music/subscriptions/subscriptions.json")`

atar-axis:
So then it is an translation issue, it doesn't on the german version. Thanks for clarification!

atar-axis:
> I think all we need to do is extracting the `.zip` and searching for the json where it says `[ { ..., "kind" : "youtube#subscription", ... } ]` should be pretty easy I guess.

@Stypox what do you think about that? I could give it a try, shouldn't be too hard

Stypox:
> So then it is an translation issue, it doesn't on the german version.

Oh, yeah, translations will be there only after syncing with WebLate

Stypox:
> So then it is an translation issue, it doesn't on the german version.

Oh, yeah, translations will be there only after syncing WebLate ;-)

> @Stypox what do you think about that? I could give it a try, shouldn't be too hard

I am not sure if that would work out perfectly (maybe there are still issues in some localizations? idk). Anyway, after a second thought I think we should detect automatically if the user is passing us a json file directly (in which case we parse it as we do currently) or a zip archive (in which case we try to look for the json). This should bring us the best of both worlds, since we would both be able to pass a single file and the whole archive. Feel free to open a PR in the NewPipe repo (imo it's better not to put this part in the Extractor since it has to do with files, zips, etc.). Thank you :-D

ochen1:
The APK worked like a charm! Thanks for doing this!

PR #4859, opened 2020-11-11
Update Invidious URL list in the manifest

opusforlife2:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Update list of Invidious URLs that work in the app.

- New Invidious URLs don't work in Newpipe.

- https://github.com/TeamNewPipe/NewPipeExtractor/pull/463

#### Due diligence

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -227,20 +227,18 @@
                 <data android:host="invidio.us" />
                 <data android:host="dev.invidio.us" />
                 <data android:host="www.invidio.us" />
+                <data android:host="vid.encryptionin.space" />
                 <data android:host="invidious.snopyta.org" />
-                <data android:host="fi.invidious.snopyta.org" />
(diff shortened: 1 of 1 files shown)


Stypox:
Are you sure we should include all of them in the manifest? It will slow down the global share menu for the users installing NewPipe on their phones, and I guess many of those instances are not popular

opusforlife2:
This is reducing the number of instances by two, so I'm sure it won't make it slower than it currently is, but probably a tiny bit faster.

Stypox:
Oh, then ok, this can be merged right after 0.20.3

opusforlife2:
> It will slow down the global share menu for the users 

As for this, maybe once we're able to figure out how that PeerTube app is accepting share intents without putting URLs in the manifest, we could use that?

Stypox:
@opusforlife2 yes

TobiGr (approved):

PR #4891, opened 2020-11-15
Improve performance for some updateState calls (local lists)

vkay94:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

This PR improves the binding for local list items (in playlist details and history) which require the progress time from streams (red indicator) by adding the field `progressTime` to some classes. The field is retrieved directly from the database and doesn't need to be fetched later (which blocked the UI by calling `blockingGet`).

With this change it is also easier to compare the items for changes more efficient when the lists are calculated asynchronously sometime in the future (e.g. through `ListAdapter` or Groupie).

#### APK testing 

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5542397/app-debug.zip)

#### Due diligence


Stypox (approved):
I tested this and it works as expected. Thank you for looking into it, it should improve list performance while having no tradeoff (live-updatiing progress was not working before anyway, and would require using room's `LiveData`). Thank you :-D

vkay94:
Yeah, there shouldn't be a case where it would result in unexpected loading updating. All lists using these entries with the progress bar are loaded as whole in `handleResult`, so there isn't any specific update for single items (at least I didn't found any :') )

PR #4893, opened 2020-11-15
Pull to Refresh Feed

### LINKED ISSUE #3329: Improve "What's New" refresh experience: auto refresh, pull to refresh
###  Is your feature request related to a problem? Please describe it 

Related to Issue #3321  
Refreshing the what's new page feels slower/clunkier than it needs to.
Refreshing is a commitment, especially if you have a lot of subscriptions. I feel like I need to check all of the videos I've loaded before I take the 2 minutes needed to refresh 200 channels.

### Describe the solution you'd like

A few things:

Add #3321 (pull to refresh)

Option to refresh on app start.

When refreshing, the loading bar should be contained in the status box which currently holds the refresh button. Additionally, the old data should be kept until the new data has been loaded. 

Ideally, the new data is loaded "over" the old data, pushing it further down the list, using the current position in the list as an anchor (either near the top of the list, or the items you're viewing in the list). 

For example, if I've refreshed and stayed near the top, the old videos will be pushed down/unloaded if needed, whereas if I had scrolled down to view older content, the new data would be loaded above the old, as to not lose my place in the list. (in this case, data that should be unloaded will be unloaded once the user has navigated to another screen)

### How will you/everyone benefit from this feature?

The what's new page is amazing, and much thanks to @mauriciocolli for all their hard work, but I personally believe that refreshing is a bit clunky and could use a touch up.

I think if these proposals can be achieved, then the what's new page will have near seamless and intuitive interaction. Pull to refresh is near universal now, and if you have a lot of subscriptions, if gives you newish stuff to look through while newer stuff is loaded.

okan35:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

added pull to refresh to what's new section of app

- fixes #3329

- 

#### APK testing 

debug.zip
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5542783/app-debug.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
@@ -57,7 +57,7 @@ import org.schabi.newpipe.util.Localization
 
 class FeedFragment : BaseListFragment<FeedState, Unit>() {
     private lateinit var viewModel: FeedViewModel
-    private lateinit var  swipeRefreshLayout: SwipeRefreshLayout
+    private lateinit var swipeRefreshLayout: SwipeRefreshLayout
     @State
     @JvmField
     var listState: Parcelable? = null
@@ -80,17 +80,13 @@ class FeedFragment : BaseListFragment<FeedState, Unit>() {
     }
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
-
         return inflater.inflate(R.layout.fragment_feed, container, false)
     }
 
     override fun onViewCreated(rootView: View, savedInstanceState: Bundle?) {
         super.onViewCreated(rootView, savedInstanceState)
         swipeRefreshLayout = requireView().findViewById(R.id.swiperefresh)
-        swipeRefreshLayout.setOnRefreshListener {
-            reloadContent()
-        }
-
+        swipeRefreshLayout.setOnRefreshListener { reloadContent() }
         viewModel = ViewModelProvider(this, FeedViewModel.Factory(requireContext(), groupId)).get(FeedViewModel::class.java)
         viewModel.stateLiveData.observe(viewLifecycleOwner, Observer { it?.let(::handleResult) })
     }
@@ -196,9 +192,7 @@ class FeedFragment : BaseListFragment<FeedState, Unit>() {
 
         empty_state_view?.let { animateView(it, false, 0) }
         animateView(error_panel, false, 0)
-        if (swipeRefreshLayout.isRefreshing){
-            swipeRefreshLayout.isRefreshing = false
-        }
+        swipeRefreshLayout.isRefreshing = false
     }
 
     override fun showEmptyState() {
--- a/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
@@ -34,6 +34,7 @@ import androidx.core.view.isVisible
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProvider
 import androidx.preference.PreferenceManager
+import androidx.swiperefreshlayout.widget.SwipeRefreshLayout
 import icepick.State
 import java.util.Calendar
 import kotlinx.android.synthetic.main.error_retry.error_button_retry
@@ -56,6 +57,7 @@ import org.schabi.newpipe.util.Localization
 
 class FeedFragment : BaseListFragment<FeedState, Unit>() {
     private lateinit var viewModel: FeedViewModel
+    private lateinit var  swipeRefreshLayout: SwipeRefreshLayout
     @State
     @JvmField
     var listState: Parcelable? = null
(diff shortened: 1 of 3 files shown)


Stypox:
Thank you! Please provide a debug apk when opening a PR ;-)

okan35:
> 
> 
> Thank you! Please provide a debug apk when opening a PR ;-)

Sorry about that, just updated the request with apk.

app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
```diff
@@ -189,6 +196,9 @@ class FeedFragment : BaseListFragment<FeedState, Unit>() {
 
         empty_state_view?.let { animateView(it, false, 0) }
         animateView(error_panel, false, 0)
+        if (swipeRefreshLayout.isRefreshing){
+            swipeRefreshLayout.isRefreshing = false
+        }
```
mhmdanas, line 201:
```suggestion
        swipeRefreshLayout.isRefreshing = false
```
If I'm not mistaken, there's no difference.

okan35: But there is a difference of checking if currently there is a refreshing action and set that action false otherwise dont. I thought this is a healthier way of doing like when you dismiss a progressDialog.

mhmdanas: In your original code, `isRefreshing` will be set to `false` if it's `true`, otherwise it stays the same. This is the same as setting it to `false` regardless of its value.

opusforlife2:
Just tried it out. The spinner stutters. It seems to pause momentarily in the exact same places as the progress bar in the feed update notification.

okan35:
I don't know what to say for stutter, I just used swipe refresh layout and haven't changed much code.
Also started  and stopped the refresh in the same place with feed refresh so that might be the reason it acts same.

app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
okan35: I will try the change your recommended, if it doesn't change any behavior I will just commit it.

app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
```diff
@@ -56,6 +57,7 @@ import org.schabi.newpipe.util.Localization
 
 class FeedFragment : BaseListFragment<FeedState, Unit>() {
     private lateinit var viewModel: FeedViewModel
+    private lateinit var  swipeRefreshLayout: SwipeRefreshLayout
```
XiangRongLin, line 60: double space

app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
```diff
@@ -78,11 +80,16 @@ class FeedFragment : BaseListFragment<FeedState, Unit>() {
     }
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+
```
XiangRongLin, line 83: here

app/src/main/res/layout/fragment_feed.xml
```diff
@@ -1,10 +1,14 @@
...
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout
+
+    xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
+
```
XiangRongLin, line 10: here

app/src/main/res/layout/fragment_feed.xml
```diff
@@ -70,15 +74,23 @@
             android:background="?attr/separator_color" />
     </RelativeLayout>
 
-    <androidx.recyclerview.widget.RecyclerView
-        android:id="@+id/items_list"
+
```
XiangRongLin, line 77: here

app/src/main/res/layout/fragment_feed.xml
```diff
@@ -1,10 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout
+
```
XiangRongLin, line 3: here

XiangRongLin (commented):
There are some unnecessary whitespaces

opusforlife2 (changes requested):
Notes from a secret admirer:

> You should put the new dependency in the AndroidX dependencies block.
And use Ctrl+Alt+L to format your XML.

jaffery0605:
Please add pull down to refresh feed instead of clicking the touchable portion down the header. It will improve the UX of the app.

mhmdanas:
@jaffery0605 and that's exactly what this PR does.

XiangRongLin:
Btw how do android TV people refresh the feed afterwards?

okan35:
Made the necessary changes.

okan35:
> 
> 
> Btw how do android TV people refresh the feed afterwards?

I haven't removed existing refreshing feature just added another one so they will continue to do as they did before. Not sure pull to refresh is even done on android tv.

opusforlife2 (approved):

Stypox (approved):
I tested this on multiple devices and it works good. Thank you, we really needed this! :-D

B0pol:
> > Btw how do android TV people refresh the feed afterwards?
> 
> I haven't removed existing refreshing feature just added another one so they will continue to do as they did before. Not sure pull to refresh is even done on android tv.

I don't know how to do pull to refresh. I wish there was a button on the remote, because it's lacking in many apps and sometimes the only way is to clear app data (:eyes: @f-droid)

okan35:
Actually that could be done in a way like double bottom button click on android tv remote so when a user double clicks bottom button in less than a second we could refresh the feed. I don't use newpipe on android tv but since I use android tv what I said could give a good solution. If you like this idea you can open an issue or if you have a better suggestion you can open with that.

B0pol:
No, the current way is great. There is no reason to change to a "less than a second" thing.

PR #4908, opened 2020-11-17
Use TeamNewPipe repo instead of Stypox repo as dependency

XiangRongLin:
#### What is it?

Well there i was wondering why dependency substitution wasn't working and what setting i forgot to import from my old pc this time. 
And see there! It cant subtitute `com.github.TeamNewPipe:NewPipeExtractor`, because it's not used!

Edit: btw whats up with the security/synk check always failing?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -176,7 +176,7 @@ dependencies {
 
     // NewPipe dependencies
     // You can use a local version by uncommenting a few lines in settings.gradle
-    implementation 'com.github.Stypox:NewPipeExtractor:501ec30152642ad49ce0a1825410d200942d174c'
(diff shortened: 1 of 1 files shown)


Stypox:
@XiangRongLin this should not be merged, it will be fixed when 0.20.3 is merged and dev is rebased onto it 😉
But yeah, I forgot about that in one of my commits 😅

XiangRongLin:
Can this then be done soon (like today). This easily cost me 30+minutes yesterday and as long as it is not fixed it can happen to anybody else.

Stypox:
@XiangRongLin yes, 0.20.3 will be released shortly. I'm sorry for what I caused

PR #5002, opened 2020-11-25
Add "Screen layout" dropdown

GarbageHamburger:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

Some ROMs allow users to set their screen density. Beyond a certain density,
NewPipe treats screens with high density as "tablet". This can be undesirable.
This patch adds a dropdown to the Appearance menu to allow users to force
the screen layout to specific values.

#### APK testing 
[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5594323/debug.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
@@ -10,8 +10,10 @@ import android.view.KeyEvent;
 
 import androidx.annotation.NonNull;
 import androidx.core.content.ContextCompat;
+import androidx.preference.PreferenceManager;
 
 import org.schabi.newpipe.App;
+import org.schabi.newpipe.R;
 
 public final class DeviceUtils {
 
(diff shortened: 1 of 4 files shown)


avently:
Thanks!

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -21,11 +23,39 @@
...
     }
 
+    /**
+     * Checks if the screen layout was forced to the given layout.
+     * @param context Application context
+     * @param resId The resource ID of the wanted screen layout
+     * @return true if the layout was forced, false if the layout was forced to something else,
+     * and null if the screen layout is auto
+     */
+    private static Boolean layoutForcedTo(final Context context, final int resId) {
```
Stypox, line 33: `boolean` and `@StringRes`
```suggestion
    private static boolean layoutForcedTo(final Context context, @StringRes final int resId) {
```
I don't like the fact that this function could return `null`. It should return `true` if the layout is forced to the requested value and `false` othewise.

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -21,11 +23,39 @@
...
+        final String screenLayoutKey = context.getString(R.string.screen_layout_key);
+        final String autoKey = context.getString(R.string.screen_layout_auto_key);
+        final String wantedKey = context.getString(resId);
```
Stypox, line 36: Inline these three variables

Stypox (changes requested):
Maybe I'm understanding this wrong, but shouldn't you also override Android's auto mode somewhere (probably in `BaseActivity`), in order to force a particular layout?

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
TobiGr: but that is required below to distinguish correctly.

I suggest to annotate the method with `@Nullable`

Stypox: Mmmh ok, but I really don't like giving a contextual meaning to `null`. A better thing would be to implement a `getScreenLayoutMode()` function which returns an enum value which can be `AUTO`, `FORCE_TV`, `FORCE_TABLET`, `FORCE_PHONE`, and then let the caller decide what to do with them. What do you think?

TobiGr: Yes, that sounds like a good solution

litetex:
Closing this for now:
* no progress
* no GitHub actions build
* merge conflicts
* no feedback from the author

Feel free to reopen it when there is progress again.

PR #5059, opened 2020-12-01
Extract export database logic into own class

XiangRongLin:
#### What is it?

- Separate database/settings generation logic from the UI.
- Add happy path unit test.

With this seperation, unit tests can be introduced. I just added one as proof of concept for now.
If this style is ok i would work on extracting more stuff out.

This also allows cheap unit tests instead of comparatively expensive androidTests.

#### APK testing 

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5624885/app-debug.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
@@ -30,25 +30,23 @@ import org.schabi.newpipe.report.UserAction;
 import org.schabi.newpipe.util.FilePickerActivityHelper;
 import org.schabi.newpipe.util.ZipHelper;
 
-import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.Locale;
 import java.util.Map;
 import java.util.zip.ZipFile;
-import java.util.zip.ZipOutputStream;
 
 import static org.schabi.newpipe.util.Localization.assureCorrectAppLanguage;
 
 public class ContentSettingsFragment extends BasePreferenceFragment {
     private static final int REQUEST_IMPORT_PATH = 8945;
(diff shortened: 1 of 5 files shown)


XiangRongLin:
I added the setting `Mockito.withSettings().stubOnly()` in the test, since a stub is enough

Stypox (approved):
We definitely need more tests, so I like this. I tested on my device and it works; I also run the test locally and it works. Thank you!

app/src/test/java/org/schabi/newpipe/settings/ContentSettingsManagerTest.kt
```diff
@@ -0,0 +1,72 @@
...
+class ContentSettingsManagerTest {
+
+    @RunWith(MockitoJUnitRunner::class)
+    class ExportTest {
+
+        private lateinit var preferences: SharedPreferences
+        private lateinit var newpipeDb: File
+        private lateinit var newpipeSettings: File
+
+        @Before
```
XiangRongLin, line 29: Well from refactoring too much stuff around, i apparently created a `beforeClass` method annotated `@Before`. I'll fix this later on in my next PR.

TobiGr: Please create a separate PR for this

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -131,6 +129,8 @@ public void onCreatePreferences(final Bundle savedInstanceState, final String ro
         newpipeSettings = new File(homeDir + "/databases/newpipe.settings");
         newpipeSettings.delete();
 
+        manager = new ContentSettingsManager(homeDir);
```
TobiGr, line 132: That causes the build to fail. Can you fix that, please.

Stypox: Oh, how is this possible? I ran the tests successfully on my machine...

XiangRongLin: WOW, i see where this is coming from. PR coming

PR #5067, opened 2020-12-02
Add a workaround for a possible null offset ID.

Isira-Seneviratne:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Works around an issue reported by @opusforlife2 in which the ID of the UTC `ZoneOffset` might be null (even though its documentation explicitly says that a `ZoneId`'s ID value is not null).

#### APK testing 

[app-release.zip](https://github.com/TeamNewPipe/NewPipe/files/5671795/app-release.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/Localization.java
+++ b/app/src/main/java/org/schabi/newpipe/util/Localization.java
@@ -20,6 +20,7 @@ import org.ocpsoft.prettytime.units.Decade;
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.extractor.ListExtractor;
 import org.schabi.newpipe.extractor.localization.ContentCountry;
+import org.schabi.newpipe.ktx.OffsetDateTimeKt;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
@@ -30,7 +31,6 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.FormatStyle;
 import java.util.Arrays;
 import java.util.Calendar;
-import java.util.GregorianCalendar;
 import java.util.List;
 import java.util.Locale;
 
@@ -314,7 +314,7 @@ public final class Localization {
     }
 
     public static String relativeTime(final OffsetDateTime offsetDateTime) {
-        return relativeTime(GregorianCalendar.from(offsetDateTime.toZonedDateTime()));
+        return relativeTime(OffsetDateTimeKt.toCalendar(offsetDateTime));
     }
 
     public static String relativeTime(final Calendar calendarTime) {
--- a/app/src/main/java/org/schabi/newpipe/ktx/OffsetDateTime.kt
+++ b/app/src/main/java/org/schabi/newpipe/ktx/OffsetDateTime.kt
@@ -1,10 +1,29 @@
 package org.schabi.newpipe.ktx
 
 import java.time.OffsetDateTime
-import java.time.ZoneId
+import java.time.ZoneOffset
+import java.time.temporal.ChronoField
 import java.util.Calendar
+import java.util.Date
 import java.util.GregorianCalendar
+import java.util.TimeZone
 
-fun OffsetDateTime.toCalendar(zoneId: ZoneId = ZoneId.systemDefault()): Calendar {
-    return GregorianCalendar.from(if (zoneId != offset) atZoneSameInstant(zoneId) else toZonedDateTime())
+// This method is a modified version of GregorianCalendar.from(ZonedDateTime).
+// Math.addExact() and Math.multiplyExact() are desugared even though lint displays a warning.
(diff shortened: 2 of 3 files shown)


opusforlife2:
Debug builds always work (I tried this one as well just to be certain) so I need a release build to test, remember?

Isira-Seneviratne:
> Debug builds always work (I tried this one as well just to be certain) so I need a release build to test, remember?

Right, my bad. I'll upload one now.

opusforlife2:
Works! Fixed for real this time! (Hopefully xD )

Isira-Seneviratne:
> Works! Fixed for real this time! (Hopefully xD )

Maybe you could report the bug [here](https://issuetracker.google.com/issues/new?component=317603&template=1018721)?

opusforlife2:
It could be a bug in LineageOS 16.

XiangRongLin (commented):
I feel like a bug fix like this, which which does not contain any UI stuff, should be covered by a unit tests.
When the underlying bug is fixed and the workaround is not needed anymore, it provides an easy way to verify that it still behaves as expected.

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -328,7 +328,7 @@ public static String relativeTime(final Calendar calendarTime) {
     // This method is a modified version of GregorianCalendar.from(ZonedDateTime).
     // Math.addExact() and Math.multiplyExact() are desugared even though lint displays a warning.
     @SuppressWarnings("NewApi")
-    private static GregorianCalendar fromUTC(final OffsetDateTime offsetDateTime) {
+    public static GregorianCalendar fromUTC(final OffsetDateTime offsetDateTime) {
```
XiangRongLin, line 331: Don't make a private method public for the sake of testing.

Instead you should test the methods thats using the private method or extract the method into it's own class (not 100% sure on the second part myself yet)

In this case i would write tests for `public static String relativeTime(final OffsetDateTime offsetDateTime)`. If done this way you will provide the value of having regression test for when the workaround is removed. Because currently your tests will need to be removed, when the workaround ist removed.

XiangRongLin (changes requested):

app/src/main/java/org/schabi/newpipe/util/Localization.java
Isira-Seneviratne: I tried creating an instrumented test, but attempting to run it results in an error message saying that the test class cannot be found.

XiangRongLin: I'll open an pr to your branch tonight/tomorrow with what i mean. Maybe i missed something, that makes what i mean not possible.

Isira-Seneviratne: No, I get that you want to test `Localization.relativeTime()` directly, but I can't run the instrumented test, so I tested PrettyTime manually.

XiangRongLin: Ok, now i understand why you need an instrumented test. The dependency are direcly accessed and not injected. This would need some major refactoring, which can be done in another PR.

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -314,13 +318,33 @@ private static void initPrettyTime(final Context context) {
     }
 
     public static String relativeTime(final OffsetDateTime offsetDateTime) {
-        return relativeTime(GregorianCalendar.from(offsetDateTime.toZonedDateTime()));
+        return relativeTime(fromUTC(offsetDateTime));
```
XiangRongLin, line 321: The tests are not covering the bug that you fixed. I can change back this line and your tests still pass successfully

XiangRongLin (changes requested):

app/src/main/java/org/schabi/newpipe/util/Localization.java
Isira-Seneviratne: I created a PR with Dagger, so I could add a commit for DI with `Localization` there.

Isira-Seneviratne:
That's because it seems to be specific to @opusforlife2's device. On my
device, for instance, `GregorianCalendar.from()` works fine.

Isira-Seneviratne:
@opusforlife2 could you test the new APK? I moved the workaround to `OffsetDateTime.toCalendar()` so that this issue won't potentially affect `FeedViewModel`.

TobiGr (approved):
thanks again

opusforlife2:
@Isira-Seneviratne 0.20.6 works perfectly fine, so your modification seems to be okay.

PR #5180, opened 2020-12-15
Fix strange behavior after removing watched items in playlist via menu

### LINKED ISSUE #5169: Strange behavior after deleting all (partially) watched videos in playlists
### Checklist
- [x] I am using the latest version - 0.20.6 

1. Create a playlist with some (partially) watched videos and not-watched ones.
2. Open the options menu on the top and choose "Yes and partially watched videos" (third option)

3a. Start to play one video.
3b. Open the options menu and choose "Yes" (first option)
 
### Actual behaviour

After 3a: The video is removed from the list immediately and all items are updating very often.
After 3b: The old items are appearing again (even if the optiion says "it can't be undone xD)

### Expected behavior

3a: The video should not disappear/deleted from list.
3b: The videos shouldn't re-appear again (in this case there shouldn't be any difference after third option action)

### Device info

 - Android version/Custom ROM version: Android 10 (LineageOS 17.1)
 - Device model: S4

#### PS:
I've already looked into it a little bit and one reason could be that the remove action doesn't delete items from the database and 
which is the reason for 3b (after deleting only references instead of the actual items from the database, another loading from the table will load the old list).

You can assign me to this issue, @opusforlife2 , but I can already tell that there might be some structure changes.

vkay94:
Fixes the re-appear bug in playlists which occurred when the watched items are removed via menu.

#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This PR fixes the strange behavior in local playlists after removing all (partially) watched items. The problem is described in the referenced issue below. The changes:
- Adds method to `StreamHistoryDAO` to retrieve all stream ids (more efficient than retrieving all items and iterating over them just to get the ids)
- Moves filter logic from `LocalPlaylistFragment` to `LocalPlaylistManager`

The main flow change is that the database operations/updates are performed before emitting the updated data to the UI. This way another removal call will use the database data instead of the previous data because the database changes are pending.
Also the returning observable is a `Maybe` instead of `Flowable` to reduce multiple emittings.

- fixes #5169 

#### APK testing 

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5694857/app-debug.zip)

#### Due diligence

@mhmdanas Could you please whether you can reproduce the bug with leaving the app and re-opening it? I couldn't reproduce your description (or probably I just didn't understand :')) Thank you.


mhmdanas:
> @mhmdanas Could you please whether you can reproduce the bug with leaving the app and re-opening it? I couldn't reproduce your description (or probably I just didn't understand :')) Thank you.

I can't reproduce with your changes. To be more specific, I only the playlist is still open when I leave the app.

vkay94:
>Also, there is still reloading when I re-open the app, but I think this should have a separate issue and PR.

Does it also happen when you don't delete the watched streams (simply being on the playlist page and then re-opening)? How do you notice it (flashing items etc.)? I couldn't find such reloading myself (I tried on two physical devices running 8.0 and 10.0).

If your issue is playlist related I may try to focus it here ;)

mhmdanas:
> Does it also happen when you don't delete the watched streams (simply being on the playlist page and then re-opening)? 

I found the root cause now; it has nothing to do with this issue. It only happens when I start playing one of the videos in the playlist. All of the videos reload a few times (sometimes only once) when I leave and re-open when one of the videos is being played, and they reload once if I play or pause the video. Also, I can't reproduce this consistently (always works when video is played, but only sometimes when video is paused). Anyway, I know this description is confusing, so if I do make an issue, I'll be sure to attach a screen recording.

> How do you notice it (flashing items etc.)?

Flickering thumbnails.

> I couldn't find such reloading myself (I tried on two physical devices running 8.0 and 10.0).

This may or may not be relevant: I have "Settings > Video and audio > Minimize on app switch> set to "Minimize to background player."

> If your issue is playlist related I may try to focus it here ;)

Well, the contributing guidelines state that you should "send PRs that only cover one specific issue/solution/bug." So we can't incorporate that here. However, I'll check if this was reported before and if it wasn't, I'll open an issue myself and ping you.

vkay94:
>Flickering thumbnails.

Alright. Indeed the issue with multiple times of refreshing while starting/pausing the video is unrelated to this PR (you can see it also in the GIF of the linked issue too). The reason should lie in the added `Flowable` to the disposable. When I debugged the changes in this PR I've also seen multiple log calls when using `Flowable`. I had to transform it into `Maybe` to make it a single operation which it basically is.

>However, I'll check if this was reported before and if it wasn't, I'll open an issue myself and ping you.

Okay ;)

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistManager.java
```diff
@@ -82,6 +86,51 @@ public Completable updateJoin(final long playlistId, final List<Long> streamIds)
         })).subscribeOn(Schedulers.io());
     }
 
+    public Maybe<List<PlaylistStreamEntry>> clearWatched(final long playlistId,
+                                                         final boolean remPartiallyWatched) {
```
Stypox, line 90: Do not truncate names
```suggestion
                                                         final boolean removePartiallyWatched) {
```

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistManager.java
vkay94: Whoops, forgot to change it, I'll reset the commit now und change it correctly ;)

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistManager.java
```diff
@@ -82,6 +86,51 @@ public Completable updateJoin(final long playlistId, final List<Long> streamIds)
...
 
+    public Maybe<List<PlaylistStreamEntry>> clearWatched(final long playlistId,
+                                                         final boolean removePartiallyWatched) {
+        return Flowable.combineLatest(getPlaylistStreams(playlistId),
+                    streamHistoryDAO.getHistoryStreamIdsSortedById(),
+                    (streams, historyIds) -> {
+                        final List<PlaylistStreamEntry> notWatched;
+                        if (removePartiallyWatched) {
+                            notWatched = streams.stream()
+                                    .filter(s -> !historyIds.contains(s.getStreamId()))
```
Stypox, line 97: Even though history ids are sorted, this won't do a binary search to determine whether `s.getStreamId()` is contained, since the type of `historyIds` is just `List`. This can slow down the algorithm from `O(n*log(n))` to `O(n^2)`, which is unacceptable for big playlists/histories. So use `Collections.binarySearch()` instead

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -388,76 +383,19 @@ public void removeWatchedStreams(final boolean removePartiallyWatched) {
...
+                    final String currentThumbnail = playlistManager
+                            .getPlaylistThumbnail(playlistId);
+                    final boolean requireThumbnailChange = notWatchedItems.stream()
+                            .noneMatch(s -> Objects.equals(s.getStreamEntity().getThumbnailUrl(),
+                                    currentThumbnail));
 
-                    if (thumbnailVideoRemoved) {
+                    if (requireThumbnailChange) {
                         updateThumbnailUrl();
                     }
```
Stypox, line 400: I'd also move this to a separate function inside `LocalPlaylistManager`, I see it is being used in multiple places around this file.

Stypox (changes requested):
Looks mostly good. Thanks :-D

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
vkay94: Like this?
```java
final List<PlaylistStreamEntry> notWatched;
if (removePartiallyWatched) {
   notWatched = streams.stream()
      .filter(s -> Collections.binarySearch(historyIds, s.getStreamId()) < 0)
      .collect(Collectors.toList());
} else {
   notWatched = streams.stream()
      .filter(s -> Collections.binarySearch(historyIds,  s.getStreamId()) < 0 || s.getProgressTime() > 0)
      .collect(Collectors.toList());
}
```

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistManager.java
vkay94: Like this?
```java
final List<PlaylistStreamEntry> notWatched;
if (removePartiallyWatched) {
   notWatched = streams.stream()
      .filter(s -> Collections.binarySearch(historyIds, s.getStreamId()) < 0)
      .collect(Collectors.toList());
} else {
   notWatched = streams.stream()
      .filter(s -> Collections.binarySearch(historyIds,  s.getStreamId()) < 0 || s.getProgressTime() > 0)
      .collect(Collectors.toList());
}
```

Stypox: Yes, looking good

PR #5208, opened 2020-12-17
Remove pbj=1 parameter from YouYube urls in recaptcha activity

### LINKED ISSUE #3714: Recaptcha challenge
Whenever I click any video it shows me reCAPTCHA challange and theere is nothing(white)
 On the screen except top red bar which shows done when solved
Screenshot_20200601-132409_NewPipe
. I waited for while but nothing happens the screen remains same. But it works completely fine when I use VPN. Need a help! 
   Tried clearing data and reinstalling latest version of newpipe.

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This PR removes the `pbj=1` parameter from the URLs loaded in the reCAPTCHA webview, so that an actual HTML page is loaded instead of the ajax json requested by the extractor, which would render as a blank page.

Hopefully fixes #3714

#### APK testing 
This apk throws a `ReCaptchaException` on every request, with the first request being the "test" search, always run on app opening.
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5710472/app-debug.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
@@ -53,6 +53,16 @@ public class ReCaptchaActivity extends AppCompatActivity {
     public static final String YT_URL = "https://www.youtube.com";
     public static final String RECAPTCHA_COOKIES_KEY = "recaptcha_cookies";
 
+    public static String sanitizeRecaptchaUrl(@Nullable final String url) {
+        if (url == null || url.trim().isEmpty()) {
+            return YT_URL; // YouTube is the most likely service to have thrown a recaptcha
+        } else {
+            // remove "pbj=1" parameter from YouYube urls, as it makes the page JSON and not HTML
+            return url.replace("&pbj=1", "").replace("pbj=1&", "").replace("?pbj=1", "");
+        }
+    }
+
+
     private WebView webView;
     private String foundCookies = "";
 
@@ -64,14 +74,9 @@ public class ReCaptchaActivity extends AppCompatActivity {
         final Toolbar toolbar = findViewById(R.id.toolbar);
         setSupportActionBar(toolbar);
 
-        String url = getIntent().getStringExtra(RECAPTCHA_URL_EXTRA);
-        if (url == null || url.isEmpty()) {
-            url = YT_URL;
-        }
-
+        final String url = sanitizeRecaptchaUrl(getIntent().getStringExtra(RECAPTCHA_URL_EXTRA));
         // set return to Cancel by default
         setResult(RESULT_CANCELED);
-
 
         webView = findViewById(R.id.reCaptchaWebView);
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
```diff
@@ -155,6 +150,16 @@ public boolean onOptionsItemSelected(final MenuItem item) {
         }
     }
 
+    private String getUrlFromIntent() {
```
XiangRongLin, line 153: The method should be renamed.
For me a getter should only get a value and not modify it in between. If you only read the line above `final String url = getUrlFromIntent();` and not the implementation itself, it could cause confusion down the line

app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
```diff
@@ -155,6 +150,16 @@ public boolean onOptionsItemSelected(final MenuItem item) {
...
+        final String url = getIntent().getStringExtra(RECAPTCHA_URL_EXTRA);
+        if (url == null || url.isEmpty()) {
+            return YT_URL; // YouTube is the most likely service to have thrown a recaptcha
+        } else {
+            // remove "pbj=1" parameter from YouYube urls, as it makes the page JSON and not HTML
+            return url.replace("&pbj=1", "").replace("?pbj=1&", "").replace("?pbj=1", "");
```
XiangRongLin, line 159: This looks like it would break urls like `www.host.com/?pbj=1&otherparam=2` by turning it to `www.host.com/otherparam=2`

XiangRongLin (changes requested):
Also unit tests for the removal of pbj would be nice, but optional.
It also forces one to write the code in a way which allows tests, which makes writing more tests easier in the future.

app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
Stypox: Oh, I accidentally left an additional ? in there

app/src/main/java/org/schabi/newpipe/ReCaptchaActivity.java
Stypox: Ok, I will change the signature and add a test for it

XiangRongLin: Something like `resolveUrlFromIntent` would be ok or any other verb that fits

Stypox:
Done @XiangRongLin 

> Something like resolveUrlFromIntent would be ok or any other verb that fits

In order to make it testable, I made it static and used `sanitizeRecaptchaUrl`, which makes sense to me

XiangRongLin (approved):
Static is not ideal, but since there is no logic layer underneath the UI layer, it is fine. 

Gets an ok from me from the code standpoint.

PR #5225, opened 2020-12-19
Extract settings import

XiangRongLin:
#### What is it?

- Extract logic for settings import out of `ContentSettingsFragment` into `ContentSettingsManager` and `ZipHelper`
- Add basic unit tests for `ContentSettingsManager`
- Introduce `NewPipeFileLocator` which resolves the db and settings files. Inject it into `ContentSettingsManager`
- remove subclass `ExportTest`. No benefit with only one test
- added real zip file and newpipe.settings of export as test resources 

#### APK testing 

~~[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5719734/app-debug.zip)~~

https://github.com/TeamNewPipe/NewPipe/actions/runs/459642451

#### Due diligence

#### Notes
- The git history went wide, because while extracting it I found my initial approach for DI and testing not good. Easiest would be to just squash everything together.
- With this changes to the import/export logic should always be covered by additional unit tests


Stypox (approved):
Looks good to me. Thank you!

Stypox:
I am unable to run `ContentSettingsManagerTest` on my machine:
```
Cannot mock/spy class org.schabi.newpipe.settings.NewPipeFileLocator
Mockito cannot mock/spy because :
 - final class
 ```
 
If I make `NewPipeFileLocator` an `open` class (in kotlin classes are `final` by default), I get null pointer exceptions whenever something is being mocked, like `'when'(fileLocator.db).thenReturn(db)`.

XiangRongLin:
Did you sync the Gradle files? Because the mockito-inline dependency should fix the problem with mocking final classes

XiangRongLin:
@Stypox the `NullPointerException`occurs for the same reason. Methods are final by default, so mockito can't mock then out of the box.

Issue #4316, opened 2020-09-25
Share does not work from soundcloud


garawaa:
Sharing does not work

TobiGr:
When playing a video and sharing it, the timestamp is appended to the URL. Removing it makes SoundCloud recognize it.

TobiGr:
Ah wait. We produce an invalid URL here in the first place. We should check the cause for this, too.

wb9688:
@TobiGr: And this is why `URL` exists… we really shouldn't append `String`s and hope it becomes a valid URL.

spacemonkey1101:
Hey, I would love to work on this. Can you assign me?

opusforlife2:
> I guess people only thought about YouTube when changing the share logic.

(sideways glance at @Stypox 🤭)

opusforlife2:
@spacemonkey1101 Sure.

spacemonkey1101:
I am unable to build because of the checkstyle errors already present. Should I solve them one by one or is there a way to disable checkstyle checks.

B0pol:
> I am unable to build because of the checkstyle errors already present. Should I solve them one by one or is there a way to disable checkstyle checks.

could paste the error log?

there should be only warnings, no errors. Anyway if you want to disable checkstyle [comment or remove runCheckstyle here](https://github.com/TeamNewPipe/NewPipe/blob/e0f02d40805f121180a6ef75e24421989386b00f/app/build.gradle#L138).

opusforlife2:
@spacemonkey1101 So, could you get past the build issue?

PR #5228, opened 2020-12-19
Remove timestamp from share url for all services except youtube

B0pol:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Remove timestamp from share url unless it's youtube.
It produced not found error for PeerTube, media.ccc.de, SoundCloud.

Ideally we should implement this
https://github.com/TeamNewPipe/NewPipe/issues/4316#issuecomment-698948468
> we also shouldn't even use URL there client-side, but add an extractor-side API for getting an URL with timestamp.

but it has been three months and no one is willing to work on this, then I'd like to get this PR merged for the next release.

- fixes https://github.com/TeamNewPipe/NewPipe/issues/4316

#### APK testing 

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5720264/app-debug.zip)
#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
@@ -103,6 +103,7 @@ import org.schabi.newpipe.util.ShareUtils;
 
 import java.util.List;
 
+import static org.schabi.newpipe.extractor.ServiceList.YouTube;
 import static org.schabi.newpipe.player.MainPlayer.ACTION_CLOSE;
 import static org.schabi.newpipe.player.MainPlayer.ACTION_FAST_FORWARD;
 import static org.schabi.newpipe.player.MainPlayer.ACTION_FAST_REWIND;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
```diff
@@ -889,10 +889,15 @@ private void onMoreOptionsClicked() {
     private void onShareClicked() {
         // share video at the current time (youtube.com/watch?v=ID&t=SECONDS)
         // Timestamp doesn't make sense in a live stream so drop it
-        final String ts = isLive() ? "" : ("&t=" + (getPlaybackSeekBar().getProgress() / 1000));
+
+        String videoUrl = getVideoUrl();
+        final int ts = getPlaybackSeekBar().getProgress() / 1000;
+        if (!isLive() && videoUrl.contains("https://www.youtube.com") && ts >= 0) {
```
XiangRongLin, line 895: Can't `ServiceHelper.getSelectedService` be used? It has the advantage of not relying on the specific format of the url.

XiangRongLin (changes requested):

app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
B0pol: No because selected service isn't always video service. E.g, go to app drawer, select peertube, open a youtube link from an external lihk, and your are playing a youtube video in peertube selected service, and there won't be timestamp even if it should.

Worse, select youtube service (by default when you open the app), open a link except youtube from an external app, there will be timestamp and will break video url

B0pol: Now I check video service id, instead of www.youtube.com (or selected service id)

TobiGr: Perfect, I just wanted to suggest this :+1:

TobiGr:
PeerTube sharing works like this https://framatube.org/videos/watch/9c9de5e8-0a1e-484a-b099-e80766180a6d?start=1m32s

Stypox (approved):
Looks good in my opinion

B0pol:
@TobiGr please read the pr description

TobiGr:
media.ccc.de: https://media.ccc.de/v/mygruni-adventskalender-20-hoechstkultur#t=317

Stypox:
@TobiGr that should be handled extractor-side, this is just a (hopefully temporary) workaround, so it shouldn't include timestamp share for all services

TobiGr:
I read the PR description. But why only fix the bug for YouTube and not for other services as well?

B0pol:
@Stypox you are right, so that it doesn't break URLs*

PR #5253, opened 2020-12-23
Use view binding in VideoPlayer.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use view binding in `VideoPlayer` and `VideoPlayerImpl`.

#### APK testing 

[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5732811/debug.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
+++ b/app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
@@ -33,21 +33,18 @@ import android.graphics.PorterDuff;
 import android.graphics.PorterDuffColorFilter;
 import android.os.Build;
 import android.os.Handler;
-import androidx.preference.PreferenceManager;
 import android.util.Log;
-
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
-import android.widget.ImageView;
-import android.widget.LinearLayout;
 import android.widget.PopupMenu;
-import android.widget.ProgressBar;
 import android.widget.SeekBar;
 import android.widget.TextView;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.content.res.AppCompatResources;
+import androidx.preference.PreferenceManager;
 
 import com.google.android.exoplayer2.C;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
```diff
@@ -178,54 +153,37 @@ private static boolean containsCaseInsensitive(final List<String> list, final St
...
-        this.bottomControlsRoot = view.findViewById(R.id.bottomControls);
-        this.topControlsRoot = view.findViewById(R.id.topControls);
-        this.qualityTextView = view.findViewById(R.id.qualityTextView);
-
-        this.subtitleView = view.findViewById(R.id.subtitleView);
+    public void initViews(@NonNull final PlayerBinding binding) {
+        playerBinding = binding;
+        playbackSeekBar = (SeekBar) binding.playbackSeekBar;
+        qualityTextView = (TextView) binding.qualityTextView;
+        playbackSpeed = (TextView) binding.playbackSpeed;
```
Stypox, line 168: Why did you keep these three variables?

Stypox (changes requested):
I tested and it works, changes also look good, thank you :-D
Maybe you could rename `playerBinding` into just `binding`, so that there is less code to read. `closeOverlayBinding` is ok though, since it's a specific binding.

app/src/main/java/org/schabi/newpipe/player/VideoPlayer.java
Isira-Seneviratne: I was getting compilation errors as those three views were being stored as `View` variables for some reason.

Isira-Seneviratne:
> I tested and it works, changes also look good, thank you :-D
> Maybe you could rename `playerBinding` into just `binding`, so that there is less code to read. `closeOverlayBinding` is ok though, since it's a specific binding.

I chose the name to keep the bindings distinguishable, but I'll rename the binding.

Stypox (approved):
I rebased and tested. This works. Thank you again :-D
Now I'll have to rebase #5371, wish me luck ;-)

Issue #5157, opened 2020-12-12
Show "Stream to Kodi" in context menu


lameventanas:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

This is very easy.
There already is a function to send a video to Kodi, but you first have to open it.
I would like to have a "Send to Kodi" function in the context menu of searches, 

#### Is your feature request related to a problem? Please describe it

![2020_12_12_092430](https://user-images.githubusercontent.com/11934069/101966721-2cba5980-3c5c-11eb-9d06-05ca958a61e4.png)

#### How will you/everyone benefit from this feature?

It will make it much better for owners of Kodi devices.

lameventanas:
Now that I think about it, a "Queue to Kodi" would also be good.

khimaros:
this would be awesome. a few code pointers:

[callback for current "Play in Kodi" video action](https://github.com/TeamNewPipe/NewPipe/blob/531859ac607d55ecda0b67fcc9ce4932d27d9a82/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java#L906)

[where the list of context menu items are constructed](https://github.com/TeamNewPipe/NewPipe/blob/7c80233f2625a28b96d07108e67c62d704b542f2/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java#L21)

[implementation of playWithKore()](https://github.com/TeamNewPipe/NewPipe/blob/c4a739bef66aaed3149595b7e4f4f38fde716961/app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java#L616)

i believe queue vs. play is handled by Kore itself.

PR #5310, opened 2020-12-30
add list item to play video on kodi

khimaros:
this allows the user to long press on a video in any list (search, history, playlist, etc) and send it directly to Kodi. this change honors the existing `Show "Play with Kodi" option` setting and only displays on streams supported by Kore.

there is opportunity to reduce duplication in this part of the code base generally.

closes: #5157

#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- add list context menu item to "Play on Kodi"

- #5157

#### APK testing 

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5752846/app-debug.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
@@ -17,7 +17,6 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.content.res.AppCompatResources;
-import androidx.preference.PreferenceManager;
 
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
@@ -176,11 +175,8 @@ public class PlaylistFragment extends BaseListInfoFragment<PlaylistInfo> {
                     StreamDialogEntry.share
             ));
         }
-        final boolean enableKodiEntry = KoreUtil.isServiceSupportedByKore(item.getServiceId())
-                && PreferenceManager.getDefaultSharedPreferences(context)
-                .getBoolean(context.getString(R.string.show_play_with_kodi_key), false);
-        if (enableKodiEntry) {
-            entries.add(StreamDialogEntry.play_on_kodi);
+        if (KoreUtil.shouldShowPlayWithKodi(context, item)) {
+            entries.add(StreamDialogEntry.play_with_kodi);
         }
         StreamDialogEntry.setEnabledEntries(entries);
 
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
@@ -175,7 +175,7 @@ public class PlaylistFragment extends BaseListInfoFragment<PlaylistInfo> {
                     StreamDialogEntry.share
             ));
         }
-        if (KoreUtil.shouldShowPlayWithKodi(context, item)) {
+        if (KoreUtil.shouldShowPlayWithKodi(context, item.getServiceId())) {
             entries.add(StreamDialogEntry.play_with_kodi);
         }
         StreamDialogEntry.setEnabledEntries(entries);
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -174,6 +176,12 @@ protected void showStreamDialog(final StreamInfoItem item) {
                     StreamDialogEntry.share
             ));
         }
+        final boolean enableKodiEntry = KoreUtil.isServiceSupportedByKore(item.getServiceId())
+                && PreferenceManager.getDefaultSharedPreferences(context)
+                .getBoolean(context.getString(R.string.show_play_with_kodi_key), false);
```
XiangRongLin, line 181: You could provide this as a method in `KoreUtil` so if the logic for when to show the kodi entry is centralized and not copied over 3 places.

XiangRongLin (changes requested):
If you can think of a way to add an unit test (not instrumented) that would be ideal.
Only thing i can think of is passing in service id, context and preference manager through the method and mocking the latter 2 in tests.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -70,6 +71,15 @@
         }
     }),
 
+    play_on_kodi(R.string.play_with_kodi_title, (fragment, item) -> {
```
Stypox, line 74: Call this "play_with_kody" to stay consistent

Stypox (changes requested):

khimaros:
i've resolved all inline feedback. i'm afraid i'm not able to contribute instrumentation and test harness at this time.

XiangRongLin (dismissed):

khimaros:
thank you, @XiangRongLin! i believe i accidentally put @Stypox in the critical path as well, so it looks like i'll just need one more review from them before merging. thank you all for this amazing project.

app/src/main/java/org/schabi/newpipe/util/KoreUtil.java
```diff
@@ -16,6 +18,12 @@ public static boolean isServiceSupportedByKore(final int serviceId) {
                 || serviceId == ServiceList.SoundCloud.getServiceId());
     }
 
+    public static boolean shouldShowPlayWithKodi(final Context context, final StreamInfoItem item) {
```
Stypox, line 21: This method can be also used in other places, e.g. in the player, can't it?

Stypox (changes requested):
Changes look good, thank you :-D

app/src/main/java/org/schabi/newpipe/util/KoreUtil.java
TobiGr: Do you mean places like
https://github.com/TeamNewPipe/NewPipe/blob/8fe1a76ec6f88556ff8964df09040135c9c51e1e/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java#L934-L937

khimaros: this seems like a reasonable place to use the new method, but i couldn't find a way to get a `StreamInfoItem` from a `playQueue`. any suggestions?

XiangRongLin: @khimaros Instead of passing in a `StreamInfoItem` you can pass in the service id, since that is the thing you actually need. Otherwise those items should have a common superclass, that would work too, but just passing the id is preferred.

khimaros: that works, thanks!

khimaros:
> Changes look good, thank you :-D

are you able to approve? it seems that merge is blocked on you and i'm not able to remove you as an approver.

khimaros:
resolved remaining feedback in latest commit. please take another look.

khimaros:
@Stypox thank you, are you able to approve? it seems merge is blocked on your approval and i'm not able to remove you.

TobiGr (approved):

khimaros:
@TobiGr -- thank you for the approval. are you able to merge this or does @Stypox need to approve first? i'm not able to merge it myself.

khimaros:
@Stypox -- friendly ping

PR #5331, opened 2021-01-01
In Fullscreen playback, toggle play/pause with hardware space button

mbarashkov:
… Space button.

#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

- While in Fullscreen mode, support hardware space button to play/pause the video.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -1891,8 +1891,10 @@ public final class VideoDetailFragment
 
         if (fullscreen) {
             hideSystemUiIfNeeded();
+            viewPager.setVisibility(View.GONE);
         } else {
             showSystemUi();
+            viewPager.setVisibility(View.VISIBLE);
         }
 
         if (relatedStreamsLayout != null) {
--- a/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
@@ -505,6 +505,11 @@ public class VideoPlayerImpl extends VideoPlayer
         switch (keyCode) {
             default:
                 break;
+            case KeyEvent.KEYCODE_SPACE:
+                if (isFullscreen) {
+                    onPlayPause();
+                }
+                break;
             case KeyEvent.KEYCODE_BACK:
                 if (DeviceUtils.isTv(service) && isControlsVisible()) {
                     hideControls(0, 0);


app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1891,8 +1891,12 @@ public void onFullscreenStateChanged(final boolean fullscreen) {
 
         if (fullscreen) {
             hideSystemUiIfNeeded();
+            viewPager.setVisibility(View.GONE);
         } else {
             showSystemUi();
+            if (playerService.getView().getVisibility() != View.GONE) {
```
Redirion, line 1897: shouldn't this be != View.VISIBLE here?

what does this change intent to do here anyway?

mbarashkov: I'm checking (when fullscreen is turned off) if the player view is shown, for a sanity check. If it is, I show the viewpager again.
The whole idea of hiding it is to disable keyboard focusing to it when pressed. Also you could press "down" key on the hardware keyboard (in fullscreen view!) and scroll down to comments - which definitely looked like a bug. This is also fixed here.

mbarashkov: I've removed this check after considering it. It does not seem to be really needed. @Redirion

app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
```diff
@@ -505,6 +505,9 @@ public boolean onKeyDown(final int keyCode) {
         switch (keyCode) {
             default:
                 break;
+            case KeyEvent.KEYCODE_SPACE:
+                onPlayPause();
```
Redirion, line 509: add `if(isFullscreen) ` before onPlayPause please. Just to make sure we don't have any side effects if space bar is hit while not in fullscreen

Redirion: and then squash your commits please so that we just have one :)

mbarashkov: Done

Redirion: I like the subtle bugfix more than the actual change of this PR: I think we should rename it somehow to make it clear, that it also fixes that bug.

Redirion (approved):

app/src/main/java/org/schabi/newpipe/player/VideoPlayerImpl.java
Redirion: I just merged it and added another line to the upcoming changelog. Thank you

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1891,8 +1891,10 @@ public void onFullscreenStateChanged(final boolean fullscreen) {
 
         if (fullscreen) {
             hideSystemUiIfNeeded();
+            viewPager.setVisibility(View.GONE);
```
Stypox, line 1894: Uhm, what is this change?

Redirion: this avoids that the comment / next video page is reachable even though the video is playing in fullscreen

Stypox: I think that was intended, wasn't it @avently?

Redirion: but it doesn't make sense for an invisible/overlayed layout to be focusable. I don't think this was intended.

avently: @Stypox, thank you for letting me know.
@Redirion you just removed an ability to view comments, description, etc by long pressing on arrow down button. Now I'm unable to view anything because viewpager is hidden. 
The behavior that was before is intended. Please, remove this line or find a workaround for situation you're trying to solve that doesn't ruin the intended ability.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
avently: If I remember correctly you could use arrow down key to reach description in fullscreen on android TV. Now it's 100% impossible which is a bad thing if you ask me

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
Redirion: I am sorry then. Was too quick

avently: No problem. You probably didn't use/know that feature so it's expected

mbarashkov: @avently 
Possible solution is to only enable this feature on non-Android TV devices (as in the accepted answer here: https://stackoverflow.com/questions/27138838/how-can-i-check-if-an-app-is-running-on-an-android-tv).
What do you think? This is really *not* needed on phones or tablets.

Stypox: @mbarashkov why did you introduce this change in the first place? What issue does it solve? I mean, for me the fact that we can scroll down to comments is never in the way

mbarashkov: @Stypox 
On Android (non-TV) devices with a physical keyboard (BT keyboard or relatively rare devices like Cosmo Communicator) it would be very useful to press space to pause/unpause video.
Doing it while the comments view is visible is impossible, because it gets first focus.
Also, it's really very confusing to press down while in full screen and see comments - it really looks like a bug. I agree it can be useful on TV devices, of course, but not on tablets/phones.

Stypox: This is needed on phones or tablets, it was introduced since it could be useful to view description or comments while watching a video, not only on android tvs

mbarashkov: Well.. I was honestly confused by this. I'm in full screen, I press down and I'm in comments. 
And I **can't go back to the video at all.**
This is why I thought it's an incorrect behavior.
What do you think? @Stypox

TobiGr: In this case, you need to use this method to check whether NewPipe is running on a TV:
https://github.com/TeamNewPipe/NewPipe/blob/3c67df263c494ba437d4788f7208e41aaaabed43/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java#L24

avently: @mbarashkov
Take a look at how we use the arrow button:
[ScreenRecord-2021-01-10-15-07-37.mp4.zip](https://github.com/TeamNewPipe/NewPipe/files/5792554/ScreenRecord-2021-01-10-15-07-37.mp4.zip)

>it would be very useful to press space to pause/unpause video.
Doing it while the comments view is visible is impossible, because it gets first focus and onKeyDown event is lost

Event can not be lost. Find it and intercept. Move focus to play/pause button after that if needed. There is already working code for TVs which react on DPAD buttons without problems. Take a look on it too

mbarashkov: Well DPAD is handled in VideoPlayerImpl -> onKeyDown
Issue is: onKeyDown is never even being called (on any hardware keyboard button press) when the comments ViewPager is shown in fullscreen mode.

avently:
@mbarashkov the app I use on a phone doesn't support comments per review changes so it's not comfortable to use web version of github for this. Instead I continue our discussion via ordinal comments.

>Well DPAD is handled in VideoPlayerImpl -> onKeyDown, I understand that.
Issue is: onKeyDown is never even being called (on any hardware keyboard button press) when the comments ViewPager is shown in fullscreen mode.

I'm not sure why this happens but you probably need to set correct place of focus. So whatever you do with a player should not focus on comments section. I don't get why the same code work for dpad and don't for hardware keyboard. But anyway if you want to make it happen you need find a way:) obvious

mbarashkov:
OK. I thiought this is a bug actually. Since this is intended behavior, I'll need to find another way to make it work, and submit another pull request.

PR #5333, opened 2021-01-02
Convert AnimationUtils functions to extension functions.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Convert the static methods defined in `AnimationUtils` to Kotlin extension functions.

#### APK testing 

[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5812225/debug.zip)

#### Due diligence


app/src/main/java/org/schabi/newpipe/fragments/list/comments/CommentsFragment.java
```diff
@@ -84,16 +84,14 @@ public void showLoading() {
...
 
         if (!result.getErrors().isEmpty()) {
             showSnackBarError(result.getErrors(), UserAction.REQUESTED_COMMENTS,
                     NewPipe.getNameOfService(result.getServiceId()), result.getUrl(), 0);
         }
 
-        if (disposables != null) {
-            disposables.clear();
-        }
+        disposables.clear();
```
XiangRongLin, line 94: removing the null check won't cause problems?

XiangRongLin (commented):
Did you test everything that you changed? Since there are only very little unit tests, refactoring like this requires extensive manual testing.

app/src/main/java/org/schabi/newpipe/fragments/list/comments/CommentsFragment.java
Isira-Seneviratne: The variable is final, so there won't be any problems.

Isira-Seneviratne:
> Did you test everything that you changed? Since there are only very little unit tests, refactoring like this requires extensive manual testing.

I made sure to preserve the existing behavior.

Just to be sure, I tested it out and didn't find any issues.

Stypox (dismissed):
Looks good to me, thank you :-)

app/src/main/java/org/schabi/newpipe/ktx/TextView.kt
```diff
@@ -0,0 +1,48 @@
...
+import android.animation.ArgbEvaluator
+import android.animation.ValueAnimator
+import android.util.Log
+import android.widget.TextView
+import androidx.annotation.ColorInt
+import androidx.interpolator.view.animation.FastOutSlowInInterpolator
+import org.schabi.newpipe.BuildConfig
+
+private const val TAG = "TextViewUtils"
+private const val DEBUG = BuildConfig.BUILD_TYPE != "release"
```
TobiGr, line 16: Please import the DEBUG const from MainActivity, so we only need to change one var when there is the need to change the debug logic.

app/src/main/java/org/schabi/newpipe/ktx/View.kt
```diff
@@ -0,0 +1,325 @@
...
+import androidx.annotation.FloatRange
+import androidx.core.view.ViewCompat
+import androidx.core.view.isGone
+import androidx.core.view.isInvisible
+import androidx.core.view.isVisible
+import androidx.interpolator.view.animation.FastOutSlowInInterpolator
+import org.schabi.newpipe.BuildConfig
+
+private const val TAG = "ViewUtils"
+private const val DEBUG = BuildConfig.BUILD_TYPE != "release"
```
TobiGr, line 22: same here

TobiGr (dismissed):
Small thing

Stypox:
Oh, then also `TAG` should be changed to `= CurrentJavaClass.class.getSimpleName()`

app/src/main/java/org/schabi/newpipe/ktx/TextView.kt
Isira-Seneviratne: > Oh, then also `TAG` should be changed to `= CurrentJavaClass.class.getSimpleName()`

There doesn't seem to be a way to get a reference to the generated class for top-level declarations.

Isira-Seneviratne:
> Oh, then also `TAG` should be changed to `= CurrentJavaClass.class.getSimpleName()`

There doesn't seem to be a way to get a reference to the generated class for top-level declarations.

Stypox (approved):
Thank you. @TobiGr can this be merged?

Stypox:
I tested many places and everything seems to work well

PR #5334, opened 2021-01-02
Asking before adding duplicated videos to the playlist

ahmad-abdelbaset:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

- record videos

image

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
@@ -1,10 +1,13 @@
 package org.schabi.newpipe.local.dialog;
 
+import android.app.Dialog;
 import android.content.Context;
+import android.content.res.Resources;
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.TextView;
 import android.widget.Toast;
 
 import androidx.annotation.NonNull;
@@ -16,6 +19,7 @@ import org.schabi.newpipe.NewPipeDatabase;
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.database.LocalItem;
 import org.schabi.newpipe.database.playlist.PlaylistMetadataEntry;
+import org.schabi.newpipe.database.playlist.PlaylistStreamEntry;
 import org.schabi.newpipe.database.stream.model.StreamEntity;
 import org.schabi.newpipe.extractor.stream.StreamInfo;
 import org.schabi.newpipe.extractor.stream.StreamInfoItem;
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
```diff
@@ -179,10 +183,69 @@ private void onPlaylistSelected(@NonNull final LocalPlaylistManager manager,
...
+        final long playListId = playlist.uid;
+        int duplicate = 0;
+
+        /*get all playlist videos*/
+        final List<PlaylistStreamEntry> entries = manager.getPlaylistStreams(playListId)
+                .firstElement()
+                .blockingGet();
+
+        /*count how many duplicate videos are there*/
+        for (final PlaylistStreamEntry entry : entries) {
```
XiangRongLin, line 197: This would seem to get very slow if the playlist grows. Instead you should use a specific DB query which can make use of indexes&co

app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
```diff
@@ -179,10 +183,69 @@ private void onPlaylistSelected(@NonNull final LocalPlaylistManager manager,
...
+            if (
+                    entry.getStreamEntity().getUrl().equals(videoToAdd.getUrl())
+                            && entry.getStreamEntity().getTitle().equals(videoToAdd.getTitle())
```
XiangRongLin, line 200: I would compare the stream UID which is also indexed

XiangRongLin (changes requested):
Please don't add the logic for the duplicate check to the dialog, only UI stuff should be done here. Instead add it to the manager like `manager.appendToPlaylist(playlist.uid, streams)` is in the manager.

app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
```diff
@@ -179,10 +183,69 @@ private void onPlaylistSelected(@NonNull final LocalPlaylistManager manager,
                     .subscribe(ignored -> successToast.show()));
         }
 
-        playlistDisposables.add(manager.appendToPlaylist(playlist.uid, streams)
-                .observeOn(AndroidSchedulers.mainThread())
-                .subscribe(ignored -> successToast.show()));
+        /*Check if video had already added to this playlist*/
+        final StreamEntity videoToAdd = streams.get(0); //the first video is the target video to add
```
TobiGr, line 187: Please format all your comments like this with a space between the comments symbols and the comment to increase readability:
```suggestion
        /* Check if video had already added to this playlist */
        final StreamEntity videoToAdd = streams.get(0); // the first video is the target video to add
```

TobiGr (changes requested):
Thank you

app/src/main/res/values/strings.xml
```diff
@@ -692,4 +692,8 @@
     <string name="show_thumbnail_title">Show thumbnail</string>
     <string name="show_thumbnail_summary">Use thumbnail for both lock screen background and notifications</string>
     <string name="recent">Recent</string>
+    <string name="duplicate_video_in_playlist_title">Duplicated Video Found</string>
+    <string name="duplicate_video_in_playlist_description">This video had already added to this playlist %d times before, Do you want to add it one more time?</string>
+    <string name="yes_add_video">Yes</string>
+    <string name="dont_add_video">No</string>
```
TobiGr, line 698: please rename the strings to use `stream` instead of `video`, because services like SoundCloud do not have videos.

app/src/main/res/values/strings.xml
```diff
@@ -692,4 +692,8 @@
     <string name="show_thumbnail_title">Show thumbnail</string>
     <string name="show_thumbnail_summary">Use thumbnail for both lock screen background and notifications</string>
     <string name="recent">Recent</string>
+    <string name="duplicate_video_in_playlist_title">Duplicated Video Found</string>
+    <string name="duplicate_video_in_playlist_description">This video had already added to this playlist %d times before, Do you want to add it one more time?</string>
```
TobiGr, line 696: I suggest something like the string below. However, we need to keep in kind that `%d` can be 1+, thus some kind of `plurals` should be used to handle the different.
```suggestion
    <string name="duplicate_stream_in_playlist_description">The playlist contains this stream already %d time(s).\nDo you want to add it one more time?</string>
```

TobiGr (changes requested):
Hit the "submit review" button accidentally...

app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
Stypox: Why not just //?

app/src/main/res/values/strings.xml
Stypox: Don't we already have some Yes/No in strings?

ahmad-abdelbaset:
Thanks all, i will check your feedback

app/src/main/res/values/strings.xml
Redirion: Android has global yes and no. android.R.yes

SameenAhnaf:
@ahmad-abdelbaset It has been over a year. 🙂

opusforlife2:
Closing for now. Please leave a comment if you wish to resume working on this.

PR #5343, opened 2021-01-03
Fix typo

Razuuu:
Hey!

There's a typo in the README.md :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@
 
 ## Description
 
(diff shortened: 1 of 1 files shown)


AlexN1997:
`NewPipe does not use any Google framework libraries, nor the YouTube API`
"Nor" is the right word to choose for this sentence.
You can look it up in a dictionary.


Google: `used before the second or further of two or more alternatives (the first being introduced by a negative such as ‘neither’ or ‘not’) to indicate that they are each untrue or each do not happen.

'they were neither cheap nor convenient'`

Razuuu:
ok sry

PR #5230, opened 2020-12-20
Update prettytime.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Update prettytime to `5.0.0.Final`.

#### APK testing 

[debug.zip](https://github.com/TeamNewPipe/NewPipe/files/5720570/debug.zip)

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #5358, opened 2021-01-06
Make Localization.relativeTime testable

XiangRongLin:
#### What is it?

- Make Localization.relativeTime testable.
- Add some simple tests

#### Context

- Problem is global state in static variable `prettyTime`. But for performance reasons on Android that is preferred.
- At first i wanted to make a manager class which would then not be static. similiar how i did #5059 and #5225. But this would come with a performance loss, because every call to `relativeTime` would then create a manager object, create `prettyTime`, determine the app language, etc. https://developer.android.com/training/articles/perf-tips#ObjectCreation
- This solution allows injecting `prettyTime` for tests while only adding a method call when initializing `prettyTime` and otherwise keeping the performance fo `relativeTime`

#### APK testing 

https://github.com/TeamNewPipe/NewPipe/actions/runs/466273618

#### Due diligence

### DIFF
--- a/app/src/test/java/org/schabi/newpipe/util/LocalizationTest.kt
+++ b/app/src/test/java/org/schabi/newpipe/util/LocalizationTest.kt
@@ -18,6 +18,7 @@ class LocalizationTest {
 
         val actual = Localization.relativeTime(GregorianCalendar(2021, 1, 6))
 
+        // yes this assertion is true, even if it should be 5 days, it works as it is. Future research required.
         assertEquals("1 month from now", actual)
     }
 
--- a/app/src/test/java/org/schabi/newpipe/util/LocalizationTest.kt
+++ b/app/src/test/java/org/schabi/newpipe/util/LocalizationTest.kt
@@ -1,0 +1,39 @@
+package org.schabi.newpipe.util
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.ocpsoft.prettytime.PrettyTime
+import java.text.SimpleDateFormat
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+import java.util.GregorianCalendar
+import java.util.Locale
+
+class LocalizationTest {
+
+    @Test
+    fun `After initializing pretty time relativeTime() with a Calendar must work`() {
+        val reference = SimpleDateFormat("yyyy/MM/dd").parse("2021/1/1")
+        Localization.initPrettyTime(PrettyTime(reference, Locale.ENGLISH))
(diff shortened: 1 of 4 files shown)


app/src/test/java/org/schabi/newpipe/util/LocalizationTest.kt
```diff
@@ -0,0 +1,39 @@
...
+class LocalizationTest {
+
+    @Test
+    fun `After initializing pretty time relativeTime() with a Calendar must work`() {
+        val reference = SimpleDateFormat("yyyy/MM/dd").parse("2021/1/1")
+        Localization.initPrettyTime(PrettyTime(reference, Locale.ENGLISH))
+
+        val actual = Localization.relativeTime(GregorianCalendar(2021, 1, 6))
+
+        assertEquals("1 month from now", actual)
```
Redirion, line 21: Maybe I am just unable to understand Kotlin, in this case please enlighten me. But this assertion will never be true because both times are just 5 days apart from each other - especially when I compare it with the last test? So shouldn't it be `assertEquals("5 days from now", actual)` here as well?

XiangRongLin: You are absolutly correct, but this is actually what pretty time returns. I would guess it's either a bug in `PrettyTime` or a limitation when comparing with `GregorianCalendar`.

For these tests i went with the assumption that the current state is correct. So i just added assertions with what is returned, so regressions are covered.

Redirion: okay, can you add this explanation as an inline comment so that no one stumbles upon this later?

XiangRongLin: Do you mean the first paragraph of the second (or both)?

I would add the first with the note that further research is required

Redirion: just add anything meaningful like `// yes this assertion is true, even if it should be 5 days, it works as it is. Future research required`

Maybe it has to do something with timezone shift. Could it be that Locale.ENGLISH is defaulting to the global standard locale US in Android? This would explain why 2021, 1, 1 is in fact 2020,12,12 normalized to UTC and could explain "1 month from now"

Redirion (approved):

Redirion:
*sigh* did just see https://github.com/TeamNewPipe/NewPipe/pull/5230

mayb we should have merged this earlier

Issue #5359, opened 2021-01-06
Instrumented tests don't run


XiangRongLin:
### Checklist

- [x] I am using the latest version - dev branch 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug
1. Connect Android phone to PC/ start Android emulator
2. Execute `gradle app:connectedCheck` to run instrumented tests

### Actual behaviour

It fails at `:app:compileDebugAndroidTestJavaWithJavac`
with
```
app\src\androidTest\java\org\schabi\newpipe\report\ErrorInfoTest.java:29: error: incompatible types: Object cannot be converted to ErrorInfo
        final ErrorInfo infoFromParcel = ErrorInfo.CREATOR.createFromParcel(parcel);
```

### Expected behavior

The tests to run

### Potential solutions
1. Upgrade to kotlin 1.4.20
    - Requires (manual) testing for regressions
2. Revert https://github.com/TeamNewPipe/NewPipe/commit/340b92e32b8ae32c7779a05be25831122ff5adec
    - looses benefits of annotation

### Device info

- OS: Windows 10 

related: https://youtrack.jetbrains.com/issue/KT-25807

Stypox:
This would also be solved by #5148

PR #5360, opened 2021-01-06
Run Instrumented tests in CI pipeline

XiangRongLin:
#### What is it?

- expand ci.yml to run instrumeted test
- only run ci pipeline on push/pr against dev and master.
- uses https://github.com/marketplace/actions/android-emulator-runner
- showcase: XiangRongLin/NewPipe#8
- adds 7-8 min to CI pipeline, part of which is parallelized

- #5359 needs to be adresses which i just did by updating kotlin here.

#### Notes
- There currently are only 2 instrumentet tests, which means this doesn't really add much value yet. But i think having this already in place for when it is needed is good.
- At first i wanted it to run it on API level 19 and 29 (min and max SDK) but it fails on 19 with it not being able to find the test (https://github.com/XiangRongLin/NewPipe/runs/1657271924?check_suite_focus=true#step:5:247). I assume its because of some problem with multidex and API < 21 since when i run it normally on API 19 emulator it logs a lot of errors with `Could not find class 'xyz'`.

#### Due diligence

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,12 +5,12 @@ on:
     branches:
       - dev
   push:
-    branches: 
+    branches:
       - dev
       - master
 
 jobs:
-  build-and-test:
+  build-and-test-jvm:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -32,7 +32,7 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
 
-      - name: Build debug APK and run Tests
+      - name: Build debug APK and run jvm tests
         run: ./gradlew assembleDebug lintDebug testDebugUnitTest --stacktrace
 
       - name: Upload APK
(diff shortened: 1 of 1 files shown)


Stypox:
Everything seems ok to me, though I don't have much experience with CI. What about also testing on API 21, alongside API 29?

Stypox (approved):
Looks good to me, but let's wait for @TobiGr's opinion ;-)

XiangRongLin:
@Stypox The API level 29 was running for multiple hours for some reason. lets see if it is still like this on a rerun

XiangRongLin:
Its working now. @TobiGr after merging the required check needs to be changed since i renamed it to `build-and-test-jvm`

Stypox:
@XiangRongLin I guess it is normal that `build-and-test` is skipped, since it was replaced by `build-and-test-jvm`, wasn't it? If that's the case then this is ready to merge I think.

TobiGr (approved):

TobiGr:
> after merging the required check needs to be changed since i renamed it to `build-and-test-jvm`

Done. All workflows need to be run again now

PR #5371, opened 2021-01-08
Merge player classes into a single one

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

This PR merges the three files of the player, namely `BasePlayer`, `VideoPlayer` and `VideoPlayerImpl`, into only one. Having three files was totally useless since code randomly spanned across all three files without any clear distinction. Before unified player three different classes were needed to allow for extension, but now that's not needed anymore. Merging everything together also removes lines of code and makes the code simpler to read, since everything is in one place, as there are no overridden methods all over the place anymore.
- Create different sections inside the file with big title comments, grouping together related functions. Add `//region` and `//endregion` comments before and after each section, allowing them to be folded by the IDE.
- Use only one `enum` for player type, also for intents, and therefore use `intvalue = type.ordinal()` and `PlayerType.values()[intvalue]` 
- Remove some unused functions
- Move utility functions and functions that use preferences to`PlayerHelper`
- Change `checkPopupPositionBounds` so that it tries to clamp both x and y bounds and not only one at a time
- Fix wrong speed indicator in queue activity (finally I understood what ought to be done)
- Remove viewpager hiding from #5331
- Fix view types different between `player.xml` and `player.xml (large-land)`, causing view binding to require conversions: see https://github.com/TeamNewPipe/NewPipe/pull/5253#discussion_r554429148
- Merge `ServicePlayerActivity` and `BackgroundPlayerActivity` into a single `PlayQueueActivity` and remove useless fields `getTag` and `redraw` (which were not used)

I'd like this to be merged soon, otherwise it will be a mess to rebase.

#### Due diligence


opusforlife2:
This has been bugging you for a while, hasn't it? xD

Redirion:
as your PR also modifies the VideoDetailFragment. Can you please remove the lines 
https://github.com/TeamNewPipe/NewPipe/blob/d7265461dada460fd15eb8e58862590f202f0411/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L1892
and 
https://github.com/TeamNewPipe/NewPipe/blob/d7265461dada460fd15eb8e58862590f202f0411/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L1895

that's from PR #5331. So I don't have to do that in another PR and force you to rebase / force push.

vkay94:
I'm in favor of this :) It's always a struggle to "not breaking" stuff because of the `super`-calls, especially for methods like `showControls` and `hideControls` - requires a check to on all classes xD

app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
```diff
@@ -33,61 +33,59 @@
 public class PlayerGestureListener
         extends BasePlayerGestureListener
         implements View.OnTouchListener {
-    private static final String TAG = ".PlayerGestureListener";
-    private static final boolean DEBUG = BasePlayer.DEBUG;
+    private static final String TAG = PlayerGestureListener.class.getSimpleName();
```
XiangRongLin, line 36: Is this done like this in other places too?

app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
```diff
@@ -129,6 +128,7 @@ public boolean onCreateOptionsMenu(final Menu m) {
         getMenuInflater().inflate(R.menu.menu_play_queue, m);
         getMenuInflater().inflate(getPlayerOptionMenuResource(), m);
         onMaybeMuteChanged();
+        onPlaybackParameterChanged(player.getPlaybackParameters());
```
XiangRongLin, line 131: Where is this coming from?

XiangRongLin (commented):
I just did a quick run down, i skipped `Player` since there is no direct side-by-side comparison. But if you just pulled the methods upwards there shouldn't be much to look at.
`PlayerHelper` i also just skimmed through for the same reason.

Only thing that i would change is making seperate commits for those points. When looked at afterwards it then shows that its intended and not accidental.
> Fix wrong speed indicator in queue activity (finally I understood what ought to be done)
Remove viewpager hiding from #5331

app/src/main/java/org/schabi/newpipe/player/ServicePlayerActivity.java
Stypox: This fixes the issue with the speed on the button being wrong

avently:
Isn't it quite hard to navigate in 4k lines of code?
This seems like a logical step forward. Maybe someone will find a way to divide the file into the pieces without direct connection so they can do it's job without touching other classes

app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
Stypox: Yes, this is the way to do it

TacoTheDank:
> Isn't it quite hard to navigate in 4k lines of code?
> This seems like a logical step forward. Maybe someone will find a way to divide the file into the pieces without direct connection so they can do it's job without touching other classes

Paging @XiangRongLin :)

XiangRongLin:
@TacoTheDank Seeing how my other refactoring&tests PR are still just lying around, I'm not going to touch this. Refactoring something big like this needs many iterations, which does not work if each iteration has a very long wait time just to get feedback.

Stypox:
@XiangRongLin I separated the commits as you requested, and also merged and renamed `ServicePlayerActivity` into `PlayQueueActivity`

XiangRongLin (dismissed):
Looks ok to me. I trust you with merging those classes, since like many said 4k lines is too much to review.

opusforlife2:
> Seeing how my other refactoring&tests PR are still just lying around, I'm not going to touch this. Refactoring something big like this needs many iterations, which does not work if each iteration has a very long wait time just to get feedback.

Yes, this is a chronic problem because of general lack of free time on the team. Plus even when a dev gets some time to work on Newpipe, they might have their own priority regarding what to work on.

@XiangRongLin @vkay94 You've both been given write access, so you can approve and merge PRs. Whenever you have the inclination, please go ahead and review/merge PRs that meet your standards. :) 

Also, @XiangRongLin could you please join the newpipe IRC channel (preferably with a Matrix account)? @vkay94 is already there.

Redirion:
player.getPlayer() or player.isPlayerNull are looking a bit unfortunate. Maybe it should be renamed to exoInstance / playbackInstance or something like that?

Stypox:
I rebased onto view binding and tested again, everything seems to be working fine. @Redirion I renamed `playerIsNull` into `exoPlayerIsNull`, is that ok? I don't know, though, what you are referring to with `player.getPlayer`, there is no such method anymore. @Isira-Seneviratne I was able to solve the problem with incompatible view binding types by making sure the same view types are used in `player.xml` and `player.xml (large-land)`

Redirion (approved):
lgtm

Redirion:
I will take the bold move to merge this. I think this should go into public testing with the new release. This will give us the opportunity to collect more broad feedback.

PR #5407, opened 2021-01-14
Checkout branch in CI process

XiangRongLin:
#### What is it?

- Checkout branch in CI process for pull request. CI process for push already checks out branch correctly https://github.com/XiangRongLin/NewPipe/runs/1703962796?check_suite_focus=true#step:2:456
- This way the debug build app id contains the branch name again. Currently the app id is "NewPipe HEAD"
- `github.head_ref & github.event_name` https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context
- `git checkout -B` https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt--Bltnewbranchgt
- PR job force switches the branch: https://github.com/XiangRongLin/NewPipe/pull/9/checks?check_run_id=1703963185#step:3:4
- Push jobs stays unchaged, since it gets skipped: https://github.com/XiangRongLin/NewPipe/runs/1703962796?check_suite_focus=true

#### Due diligence

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: create and checkout branch
+        # push events already checked out the branch
+        if: github.event_name == 'pull_request'
+        run: git checkout -B ${{ github.head_ref }}
+
       - name: set up JDK 1.8
         uses: actions/setup-java@v1.4.3
         with:


XiangRongLin:
Currently does not work because it is trying to fetch the branch from the NewPipe repo. But the branch exists on the repo of the PR opener, which usually is not NewPipe

TobiGr:
What about
``` sh 
git branch ${VAR_FOR_BRANCH_NAME}
```

XiangRongLin:
that will throw an error if the branch already exists. which can occur if the PR comes from the newpipe repo itself or comes from the dev branch of another repo.

I don't know if that aborts the run, but still not ideal

Stypox:
Then what about deleting the branch if it exists and then doing git branch?

XiangRongLin:
@TobiGr @Stypox I went with `checkout -B` now. in the normal cases where the branch does not exists it works like `checkout -b`. In the cases where it does exist, it moves that branch to the current position and checks it out.

see my updated pr body
  
 ----

> Then what about deleting the branch if it exists and then doing git branch?

I didn't want to figure out how to do that in shell script

TobiGr (approved):
Thanks. Looks good to me!

PR #5418, opened 2021-01-15
Combine the two Constants files into one file.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Combine `Constants.java` and `ConstantsKt.kt` into one file.

#### APK testing 

debug.zip

#### Due diligence


app/src/main/java/org/schabi/newpipe/util/Constants.kt
```diff
@@ -0,0 +1,20 @@
+@file:JvmName("Constants")
+
+package org.schabi.newpipe.util
+
+/**
+ * Default duration when using throttle functions across the app, in milliseconds.
+ */
+const val DEFAULT_THROTTLE_TIMEOUT = 120L
```
Stypox, line 8: I see there are no other file changes. Isn't this variable used anywhere? If that's the case remove it

Isira-Seneviratne: It's used in some of the `ViewModel`s.

Stypox: Shouldn't the `ViewModel`s imports also be changed? Otherwise they would point to `ConstantsKt`, wouldn't they?

Isira-Seneviratne: The `ViewModel`s are written in Kotlin, so no change is needed.

Stypox (approved):
Ok, thank you

PR #5187, opened 2020-12-15
Share improvements + fix crash when no default browser is set on some devices

TiA4f8R:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### General description of the changes
This PR improves sharing content to other apps with the following changes:

- fix crash when no browser is set on some devices (NewPipe thinks that the system app chooser is a browser because its package ID isn't "android" due to some OEM's system customizations)
- try to use Android's system share sheet instead of OEM's share sheet (this last is used as a fallback if this first was removed) for:
  - share content in clipboard (playlists, videos, tracks, channels and bug reports)
  - open content in browser (playlists, videos, tracks, channels, report on GitHub in bug reports options and comments, install a player prompts) where there is no default app
  - open a video stream / an audio stream in a external player where there is no default app
- fix double toast `Copied to clipboard` shown when copying formatted report in crash report activity

#### Code changes
- catch ActivityNotFoundException when opening an URL or a file if it's a chooser
- add openFileInApp method in ShareUtils to be able to open streams in external app / downloads in default app or in the app chooser
- edit getDefaultAppPackageName method of ShareUtils to be able to return default app for opening downloads, streams, or URLs
- use ShareUtils.openFileInApp for downloads / playing streams in external player
- use ShareUtils.shareUrl for sharing crash report from ErrorActivity to other apps and for sharing a stream in the play queue
- use ShareUtils.openUrlInBrowser when tap "Install" button from install VLC and install Kore dialogs
- use an ACTION_CHOOSER intent and putting as an extra intent the ACTION_SEND intent to share downloads
- add a string open_with in English and French which is used for Android's system chooser title
- add / edit comments and change JavaDocs for remembering that "android" package isn't returned when no default app on some devices
- also indent checkstyle-suppressions.xml file

Here is the difference between the original code and my changes (my device is an Honor 9X, running EMUI 10):

| Before this PR | After this PR |
|----------------|---------------|
| Share panel before               | Share panel after              |
| Share panel before this PR               | Share panel after this PR              |
| Open in browser before               | Open in browser after              |
| Open in browser before this PR              | Open in browser after this PR              |
| Open a download before             | Open a download after              |
| Open a download when no default app set before this PR              | Open a download when no default app set after this PR              |

#### What I changed to do this
[...]

PR #5429, opened 2021-01-16
Fix crash when no browser is present and use an ACTION_CHOOSER intent in the app update notification

TiA4f8R:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Show a Toast when no app is present on user's device to open a content in an app and in a browser and use an ACTION_CHOOSER intent with the ACTION_VIEW intent put as an extra intent in the update notification.

https://github.com/TeamNewPipe/NewPipe/pull/5187#issuecomment-761558752

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
@@ -154,12 +154,12 @@ public final class ShareUtils {
      * If no app is set as default, it will return "android" (not on some devices because some
      * OEMs changed the app chooser).
      * <p>
-     * If no app is installed on user's device to handle the intent, it will return null.
+     * If no app is installed on user's device to handle the intent, it will return an empty string.
      *
      * @param context the context to use
      * @param intent  the intent to get default app
-     * @return the package name of the default app, an empty string if there's no app installed to
-     * handle the intent or the app chooser if there's no default
+     * @return the package name of the default app to open the intent, an empty string if there's no
+     * app installed to handle it or the app chooser if there's no default
      */
     private static String getDefaultAppPackageName(final Context context, final Intent intent) {
         final ResolveInfo resolveInfo = context.getPackageManager().resolveActivity(intent,
@@ -177,7 +177,7 @@ public final class ShareUtils {
      * If no browser is set as default, it will return "android" (not on some devices because some
      * OEMs changed the app chooser).
      * <p>
-     * If no app is installed on user's device to handle the intent, it will return null.
+     * If no browser is installed on user's device, it will return an empty string.
      * @param context the context to use
      * @return the package name of the default browser, an empty string if there's no browser
      * installed or the app chooser if there's no default
--- a/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
@@ -6,6 +6,7 @@ import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.content.pm.ResolveInfo;
 import android.net.Uri;
 import android.widget.Toast;
 
@@ -64,13 +65,18 @@ public final class ShareUtils {
             // No browser set as default (doesn't work on some devices)
             openInDefaultApp(context, intent);
         } else {
-            try {
-                intent.setPackage(defaultPackageName);
-                context.startActivity(intent);
-            } catch (final ActivityNotFoundException e) {
-                // Not a browser but an app chooser because of OEMs changes
-                intent.setPackage(null);
-                openInDefaultApp(context, intent);
+            if (defaultPackageName.isEmpty()) {
+                // No app installed to open a web url
+                Toast.makeText(context, R.string.no_app_to_open_intent, Toast.LENGTH_LONG).show();
+            } else {
+                try {
+                    intent.setPackage(defaultPackageName);
+                    context.startActivity(intent);
+                } catch (final ActivityNotFoundException e) {
+                    // Not a browser but an app chooser because of OEMs changes
+                    intent.setPackage(null);
+                    openInDefaultApp(context, intent);
+                }
             }
         }
     }
@@ -104,19 +110,24 @@ public final class ShareUtils {
      * @param intent  the intent to open
      */
     public static void openIntentInApp(final Context context, final Intent intent) {
-        final String defaultAppPackageName = getDefaultAppPackageName(context, intent);
+        final String defaultPackageName = getDefaultAppPackageName(context, intent);
 
-        if (defaultAppPackageName.equals("android")) {
+        if (defaultPackageName.equals("android")) {
             // No app set as default (doesn't work on some devices)
             openInDefaultApp(context, intent);
         } else {
-            try {
-                intent.setPackage(defaultAppPackageName);
-                context.startActivity(intent);
-            } catch (final ActivityNotFoundException e) {
-                // Not an app to open the intent but an app chooser because of OEMs changes
-                intent.setPackage(null);
-                openInDefaultApp(context, intent);
+            if (defaultPackageName.isEmpty()) {
+                // No app installed to open the intent
+                Toast.makeText(context, R.string.no_app_to_open_intent, Toast.LENGTH_LONG).show();
+            } else {
+                try {
+                    intent.setPackage(defaultPackageName);
+                    context.startActivity(intent);
+                } catch (final ActivityNotFoundException e) {
+                    // Not an app to open the intent but an app chooser because of OEMs changes
+                    intent.setPackage(null);
+                    openInDefaultApp(context, intent);
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -149,8 +158,9 @@ private static void openInDefaultApp(final Context context, final Intent viewInt
      * @return the package name of the default app, or the app chooser if there's no default
      */
     private static String getDefaultAppPackageName(final Context context, final Intent intent) {
-        return context.getPackageManager().resolveActivity(intent,
-                PackageManager.MATCH_DEFAULT_ONLY).activityInfo.packageName;
+        final ResolveInfo resolveInfo = context.getPackageManager().resolveActivity(intent,
+                PackageManager.MATCH_DEFAULT_ONLY);
+        return resolveInfo.activityInfo.packageName;
```
Stypox, line 163: This will still throw a null pointer exception

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -104,14 +109,18 @@ public static void openUrlInBrowser(final Context context, final String url) {
...
 
-        if (defaultAppPackageName.equals("android")) {
+        if (defaultPackageName.equals("android")) {
             // No app set as default (doesn't work on some devices)
             openInDefaultApp(context, intent);
         } else {
+            if (defaultPackageName == null) {
+                Toast.makeText(context, R.string.no_app_to_open_intent, Toast.LENGTH_LONG).show();
+            }
+
```
Stypox, line 121: I'd move this inside the catch below or even remove it alltogether

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
TiA4f8R: What do you think about this?
```
    private static String getDefaultBrowserPackageName(final Context context) {
        final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse("http://"))
                .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
        String packageName = "";
        try {
            packageName = context.getPackageManager().resolveActivity(intent,
                    PackageManager.MATCH_DEFAULT_ONLY).activityInfo.packageName;
        } catch (final NullPointerException e) {
            return null;
        }
        return packageName;
    }
```
```
    private static String getDefaultAppPackageName(final Context context, final Intent intent) {
        String packageName = "";
        try {
            packageName = context.getPackageManager().resolveActivity(intent,
                    PackageManager.MATCH_DEFAULT_ONLY).activityInfo.packageName;
        } catch (final NullPointerException e) {
            return null;
        }
        return packageName;
    }
```

Stypox:
```
    private static String getDefaultAppPackageName(final Context context, final Intent intent) {
        try {
            return context.getPackageManager().resolveActivity(intent,
                    PackageManager.MATCH_DEFAULT_ONLY).activityInfo.packageName;
        } catch (final NullPointerException e) {
            return null;
        }
    }
```

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -140,33 +148,45 @@ private static void openInDefaultApp(final Context context, final Intent viewInt
...
+     * intent or the app chooser if there's no default
      */
     private static String getDefaultAppPackageName(final Context context, final Intent intent) {
-        return context.getPackageManager().resolveActivity(intent,
-                PackageManager.MATCH_DEFAULT_ONLY).activityInfo.packageName;
+        try {
+            return context.getPackageManager().resolveActivity(intent,
+                    PackageManager.MATCH_DEFAULT_ONLY).activityInfo.packageName;
+        } catch (final NullPointerException e) {
+            return null;
```
TobiGr, line 166: If you return `null` here, please make sure to check the method's usages and add null checks to not create NPE there.
Also please add the `@Nullable` annotation to the method, so the IDE can warn when using this method and not checking for `null`.

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -140,33 +148,45 @@ private static void openInDefaultApp(final Context context, final Intent viewInt
...
     private static String getDefaultBrowserPackageName(final Context context) {
         final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse("http://"))
                 .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-        return context.getPackageManager().resolveActivity(intent,
-                PackageManager.MATCH_DEFAULT_ONLY).activityInfo.packageName;
+        try {
+            return context.getPackageManager().resolveActivity(intent,
+                    PackageManager.MATCH_DEFAULT_ONLY).activityInfo.packageName;
+        } catch (final NullPointerException e) {
+            return null;
```
TobiGr, line 188: Same as above

TobiGr (changes requested):
Thank you.
Since you are now returning `null`, we need to add checks whenever the methods are used. Running the current implementation just elevates the NPEs up in the call hierarchy.

TiA4f8R:
@TobiGr Is what I've done better?

TobiGr:
Yes, that looks good. Please update the JDoc to not contain the `null` info

TiA4f8R:
@TobiGr Done. What do you think of the string shown in the toast (`No app installed on your device can open this`)?

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -64,6 +65,11 @@ public static void openUrlInBrowser(final Context context, final String url,
...
             openInDefaultApp(context, intent);
         } else {
+            if (defaultPackageName.isEmpty()) {
+                // No app installed to open a web url
+                Toast.makeText(context, R.string.no_app_to_open_intent, Toast.LENGTH_LONG).show();
+            }
+
             try {
                 intent.setPackage(defaultPackageName);
                 context.startActivity(intent);
```
TobiGr, line 75: Toast is good!
But here we tell NewPipe to open the stream we just tried to open in a browser:

https://user-images.githubusercontent.com/17365767/104925439-20ce5f80-599f-11eb-96a1-8a925acc9307.mp4

The black flickering is caused by Android Studio. Seems like recording is somewhat broken

TiA4f8R:
@TobiGr This should be now fixed. Also, can you change in 0.20.9 changelog the sentence `Fixed crash when no default browser is set on some devices and improve share dialogs` to `Fixed crash when no default browser is set and improve share dialogs (on some devices)` (because the share improvements are only applied on some devices, particularly EMUI devices)?

app/src/main/res/values/strings.xml
```diff
@@ -699,4 +699,5 @@
     <string name="show_thumbnail_summary">Use thumbnail for both lock screen background and notifications</string>
     <string name="recent">Recent</string>
     <string name="chapters">Chapters</string>
+    <string name="no_app_to_open_intent">No app installed on your device can open this</string>
```
TobiGr, line 702: Works perfectly!
One last thing:
```suggestion
    <string name="no_app_to_open_intent">No app on your device can open this</string>
```

TiA4f8R:
@TobiGr String and its French translation changed.

> Also, can you change in 0.20.9 changelog the sentence `Fixed crash when no default browser is set on some devices and improve share dialogs` to `Fixed crash when no default browser is set and improve share dialogs (on some devices)` (because the share improvements are only applied on some devices, particularly EMUI devices)?
Please can you do this ?

TobiGr:
I changed that in the GitHub release notes, but I cannot change the fastlane / F-Droid changelog, because that has a size of 497 bytes. 500 are allowed, so adding more characters will cause the other parts of the release notes to be cut off.

TobiGr (dismissed):

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -140,33 +151,47 @@ private static void openInDefaultApp(final Context context, final Intent viewInt
     /**
      * Get the default app package name.
      * <p>
-     * If no app is set as default, it will return "android".
+     * If no app is set as default, it will return "android" (not on some devices because some
+     * OEMs changed the app chooser).
      * <p>
-     * Note: it doesn't return "android" on some devices because some OEMs changed the app chooser.
+     * If no app is installed on user's device to handle the intent, it will return null.
```
Stypox, line 157:
```suggestion
     * If no app is installed on user's device to handle the intent, it will return an empty string.
```

Stypox (changes requested):

TiA4f8R:
Sorry for doing all changes in separate commits, I am on my phone and I can only use Github Web Editor right now.

Stypox (approved):

TobiGr:
No problem. There is a squash & merge button ;)

PR #5457, opened 2021-01-19
Update to ExoPlayer 2.12.3

Redirion:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [X] Codebase improvement (dev facing)

Updates ExoPlayer from 2.11 to 2.12

Note that we will have to do another PR to migrate from deprecated mediasession to media2.
However this will most likely require to also rework everything to the new MediaItems instead of working with MediaSources.
For now however this can stay and if ExoPlayer stays being compatible with mediasession we aren't in a rush.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -1433,7 +1433,7 @@ public final class Player implements
     }
 
     public boolean getPlaybackSkipSilence() {
-        return simpleExoPlayer != null
+        return !exoPlayerIsNull() && simpleExoPlayer.getAudioComponent() != null
                 && simpleExoPlayer.getAudioComponent().getSkipSilenceEnabled();
     }
 
@@ -1460,7 +1460,9 @@ public final class Player implements
         savePlaybackParametersToPrefs(this, roundedSpeed, roundedPitch, skipSilence);
         simpleExoPlayer.setPlaybackParameters(
                 new PlaybackParameters(roundedSpeed, roundedPitch));
-        simpleExoPlayer.getAudioComponent().setSkipSilenceEnabled(skipSilence);
+        if (simpleExoPlayer.getAudioComponent() != null) {
+            simpleExoPlayer.getAudioComponent().setSkipSilenceEnabled(skipSilence);
+        }
     }
     //endregion
 
@@ -2336,6 +2338,7 @@ public final class Player implements
             case ExoPlaybackException.TYPE_OUT_OF_MEMORY:
             case ExoPlaybackException.TYPE_REMOTE:
             case ExoPlaybackException.TYPE_RENDERER:
+            case ExoPlaybackException.TYPE_TIMEOUT:
             default:
                 showUnrecoverableError(error);
                 onPlaybackShutdown();
@@ -3358,7 +3361,7 @@ public final class Player implements
         final List<String> availableLanguages = new ArrayList<>(textTracks.length);
         for (int i = 0; i < textTracks.length; i++) {
             final TrackGroup textTrack = textTracks.get(i);
-            if (textTrack.length > 0 && textTrack.getFormat(0) != null) {
+            if (textTrack.length > 0) {
                 availableLanguages.add(textTrack.getFormat(0).language);
             }
         }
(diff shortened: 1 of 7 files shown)


TacoTheDank:
@Redirion Also, I think media2 requires API 21 or newer, so that would require dropping KitKat support, which could perhaps cut off a small chunk of users :thinking:

Stypox:
@TacoTheDank we will be forced to drop KitKat support sooner or later anyway due to OkHttp being discontinued on [December 31, 2021](https://github.com/square/okhttp#requirements)

TacoTheDank:
@Stypox All okay with me lol, though I do feel kinda bad for those older phone users. The only option they really have left now is SkyTube or using a mobile browser. (Also, you should edit your comment to specify the 3.12.x LTS branch, because you scared me with "OkHttp being discontinued" 😰)

Redirion:
@TacoTheDank do you have a source for the API 21 requirement of media2 for me? I wasn't able to find that info.

TacoTheDank:
@Redirion Nevermind, I think I misunderstood something. I got that idea from the answer here: https://stackoverflow.com/questions/57566102/what-are-the-differences-between-mediasessioncompat-and-mediasession-in-android and thought it meant that Media2 would only support API 21 and newer

oops :sweat_smile:

Redirion:
found something. According to this website media2 in the current version that is used by ExoPlayer requires API 16: https://androidx.tech/artifacts/media2/media2-session/1.1.0

while the media2 extension of ExoPlayer declares API 19: https://github.com/google/ExoPlayer/blob/release-v2/extensions/media2/build.gradle (this could however be overwritten by our gradle config, if ExoPlayer raises this to 21, as long as media2 stays below 21).

TacoTheDank:
@Redirion I see... :thinking: 

Oh btw, Exoplayer 2.12.0 introduced two convenience methods `play()` and `pause()`, which are equivalent to `setPlayWhenReady(true)` and `setPlayWhenReady(false)`, respectively. There are two usages of the latter set in https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/helper/AudioReactor.java that could be replaced with the former. ([Here](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/helper/AudioReactor.java)'s the link)

Stypox:
> The only option they really have left now is SkyTube or using a mobile browser

@TacoTheDank they would still be able to use NewPipe legacy

opusforlife2:
@Stypox Does Newpipe Legacy not use OkHttp?

Stypox:
It does, but I think it can be kept or replaced with something else specific to old android versions

TacoTheDank:
> The only option they really have left now is SkyTube or using a mobile browser

Oh yeah, and that :)

TacoTheDank:
Is this good to merge?

Redirion:
@Stypox I am a bit hesitant to merge my own PR. From my opinion I would say this is good to merge. Or would you prefer if we skip ExoPlayer 2.12 and move directly to 2.13?

Stypox:
I added a commit with a few style improvements and fixes of warnings. I tested playing some videos and enabling skip silence and everything worked as expected

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3358,7 +3361,7 @@ private void onTextTracksChanged() {
         final List<String> availableLanguages = new ArrayList<>(textTracks.length);
         for (int i = 0; i < textTracks.length; i++) {
             final TrackGroup textTrack = textTracks.get(i);
-            if (textTrack.length > 0 && textTrack.getFormat(0) != null) {
```
Redirion, line 3358: what was the warning here? can never be null?

Stypox: Yes, probably a `@NonNull` was added by exo player

Redirion:
Now we seem to have the situation, that we require a third dev to approve this PR as we both contribute to it

opusforlife2:
You could always approve each other's changes. :P

opusforlife2:
@Redirion Were you ever invited to join the Newpipe channel on IRC? If not, could you please make a Matrix account and join it?

Redirion:
> You could always approve each other's changes. :P

I just tried but I may not approve my own pull request even though the last commit was not by me

Stypox (approved):
I tested again and everything works fine. Thanks :-)

opusforlife2:
> Were you ever invited to join the Newpipe channel on IRC? If not, could you please make a Matrix account and join it?

@Redirion

PR #5462, opened 2021-01-20
Convert the abstract class DAOs to interfaces.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Convert the Java abstract class DAOs to interfaces. 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/database/stream/dao/StreamStateDAO.java
+++ b/app/src/main/java/org/schabi/newpipe/database/stream/dao/StreamStateDAO.java
@@ -17,31 +17,31 @@ import static org.schabi.newpipe.database.stream.model.StreamStateEntity.JOIN_ST
 import static org.schabi.newpipe.database.stream.model.StreamStateEntity.STREAM_STATE_TABLE;
 
 @Dao
-public abstract class StreamStateDAO implements BasicDAO<StreamStateEntity> {
+public interface StreamStateDAO extends BasicDAO<StreamStateEntity> {
     @Override
     @Query("SELECT * FROM " + STREAM_STATE_TABLE)
-    public abstract Flowable<List<StreamStateEntity>> getAll();
+    Flowable<List<StreamStateEntity>> getAll();
 
     @Override
     @Query("DELETE FROM " + STREAM_STATE_TABLE)
-    public abstract int deleteAll();
+    int deleteAll();
 
     @Override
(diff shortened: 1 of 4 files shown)


vkay94:
Just of curiosity:  I know that since NP uses Java 8 now and therefore because of the `default` feature the conversion is possible now (without side-effects I think). Are there benefits of switching from abstract classes to interfaces (except of getting rid of `public abstract`)?

Isira-Seneviratne:
> Just of curiosity: I know that since NP uses Java 8 now and therefore because of the `default` feature the conversion is possible now (without side-effects I think). Are there benefits of switching from abstract classes to interfaces (except of getting rid of `public abstract`)?

Aside from the change you mentioned, no.

mhmdanas (approved):

app/src/main/java/org/schabi/newpipe/database/stream/dao/StreamStateDAO.java
```diff
@@ -17,31 +17,31 @@
...
-    public abstract Flowable<List<StreamStateEntity>> getState(long streamId);
+    Flowable<List<StreamStateEntity>> getState(long streamId);
 
     @Query("DELETE FROM " + STREAM_STATE_TABLE + " WHERE " + JOIN_STREAM_ID + " = :streamId")
-    public abstract int deleteState(long streamId);
+    int deleteState(long streamId);
 
     @Insert(onConflict = OnConflictStrategy.IGNORE)
-    abstract void silentInsertInternal(StreamStateEntity streamState);
+    void silentInsertInternal(StreamStateEntity streamState);
```
mhmdanas, line 41: The only issue I see with this change is that "internal" methods like this one have to be public (all of an interface's members can be _only_ public).

Isira-Seneviratne: Private interface methods were introduced in Java 9.

mhmdanas: @Isira-Seneviratne indeed, you're right; I was not aware of that. However, we're using Java 8 here, so we still can't do that.

TobiGr:
I think it is a reasonable change (in other word: yes)

Stypox (approved):

PR #5466, opened 2021-01-20
Fix some things in ShareUtils file and do little improvements

TiA4f8R:
#### What is it?
- [x] Bugfix (user facing)

This PR fixes a bug that prevents NewPipe to fall back to Google Play Store URL when using InstallApp method. I also updated some JavaDocs and did little improvements. See commit for more details.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
@@ -21,21 +21,26 @@ public final class ShareUtils {
     /**
      * Open an Intent to install an app.
      * <p>
-     * This method will first try open to Google Play Store with the market scheme and falls back to
-     * Google Play Store web url if this first cannot be found.
+     * This method tries to open the default app market with the package id passed as the
+     * second param (a system chooser will be opened if there are multiple markets and no default)
+     * and falls back to Google Play Store web URL if no app to handle the market scheme was found.
+     * <p>
+     * It uses {@link ShareUtils#openIntentInApp(Context, Intent)} to open market scheme and
+     * {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} to open Google Play Store web
+     * URL with false for the boolean param.
      *
-     * @param context the context to use
-     * @param packageName the package to be installed
+     * @param context   the context to use
+     * @param packageId the package id of the app to be installed
      */
-    public static void installApp(final Context context, final String packageName) {
-        try {
-            // Try market:// scheme
-            openIntentInApp(context, new Intent(Intent.ACTION_VIEW,
-                    Uri.parse("market://details?id=" + packageName)));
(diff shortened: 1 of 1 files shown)


TiA4f8R:
@Stypox I think this should be reviewed and merged before 0.20.9 release.

Stypox:
Yes, I agree, I will add it to the TODOs

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -21,80 +21,87 @@ private ShareUtils() {
     /**
      * Open an Intent to install an app.
      * <p>
-     * This method will first try open to Google Play Store with the market scheme and falls back to
-     * Google Play Store web url if this first cannot be found.
+     * This method will first try to open the Google Play Store with the market scheme and falls
+     * back to Google Play Store web url if this first cannot be found.
```
TobiGr, line 25: That JDoc is wrong We try to open a store with that. The Google Play Store is not the only store listening for that scheme, e.g. F-Droid and Aurora also accept that. The method allows the user to choose the store if no default store is set.

TiA4f8R: Oh yes, sorry. I will update the JavaDoc.

TiA4f8R:
@TobiGr This should be resolved now.

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -21,80 +21,92 @@ private ShareUtils() {
...
      * <p>
-     * This method will first try open to Google Play Store with the market scheme and falls back to
-     * Google Play Store web url if this first cannot be found.
+     * This method tries to open the default app market with the package id putted as the
+     * second param (a system chooser will be opened if there are multiple markets and no default)
+     * and falls back to Google Play Store web url if no app to handle the market scheme was found.
+     * <p>
+     * It uses {@link ShareUtils#openIntentInApp(Context, Intent)} to open market scheme and
+     * {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} to open Google Play Store web
+     * url with false for the boolean param.
```
TobiGr, line 30:
```suggestion
     * This method tries to open the default app market with the package id passed as the
     * second param (a system chooser will be opened if there are multiple markets and no default)
     * and falls back to Google Play Store web URL if no app to handle the market scheme was found.
     * <p>
     * It uses {@link ShareUtils#openIntentInApp(Context, Intent)} to open market scheme and
     * {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} to open Google Play Store web
     * URL with false for the boolean param.
```

TobiGr (commented):
Looks good!

Stypox (approved):

PR #5503, opened 2021-01-28
Set all default settings at the beginning

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

I noticed NewPipe does not call [`PreferenceManager.setDefaultValues()`](https://developer.android.com/reference/androidx/preference/PreferenceManager#setDefaultValues(android.content.Context,%20int,%20boolean)) for all settings screens. This PR fixes this, by adding `notification_settings` and `update_settings`.

Probably not doing this didn't cause any issue, since there are always fallbacks when getting preferences, but it surely does not hurt.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
@@ -55,12 +55,14 @@ public final class NewPipeSettings {
             isFirstRun = true;
         }
 
-        PreferenceManager.setDefaultValues(context, R.xml.appearance_settings, true);
-        PreferenceManager.setDefaultValues(context, R.xml.content_settings, true);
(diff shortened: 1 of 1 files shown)


XiangRongLin:
To help my understanding: This method just saves the default values from the xml into the PreferenceManager? The method does not set the value to the default value? Meaning for each preference 2 values are saved, one being the current value and one the default value?

Stypox:
@XiangRongLin no, that method checks for each of the `defaultValue`s defined in in the xml if their corresponding key has already a value inside the shared preferences. If it has one, nothing is done (the setting was already changed or the default was already set), otherwise the value corresponding to that key is set to the default value.

XiangRongLin (approved):

PR #5523, opened 2021-02-01
Recognize timestamps and hashtags in descriptions and do some sharing fixes and improvements

### LINKED ISSUE #5507: Clicking YouTube video TimeStamps on Video description and Comment Section shows different behaviour
### Checklist

- [x] I am using the latest version - v0.20.9_RC4 

1. Go to any YouTube video
2. Press on Time Stamp @ video description
3. Press Time Stamp @ video comment  (if available)

### Actual behaviour

clicking TimeStamp on video description --> opens Action Picker
clicking TimeStamp on video comment --> opens Popup player

### Expected behavior

For now, I think it will be good to open Popup player in both cases.
Because some videos has long description and long comment list.
(i am telling it, because you( developers ) may have future plan to open timestamp in Main Player but i prefer a popup player
for description timestamp also)

Also, I like to mention that I know new version of NewPipe now support stream segments #5274 for description based timestamps. 
But I also prefer to click timestamps available at video description.

Screenshot_2021-01-30-10-14-32

Screenshot_2021-01-30-10-15-46

### Device info

 - Android version/Custom ROM version: 
 - Device model:

### LINKED ISSUE #5695: [Bug] Can't access timestamps from description but comments
### Checklist

- [x] I am using the latest version - 0.20.10 

### Actual behaviour

I can't access my timestamps from description. However, I can do them so from comments. This needs to be fixed.

### Expected behavior

Timestamps from anywhere should be able to open the video at the certain moment. Those timestamps should stay underlined as well.

[https://drive.google.com/file/d/1sE2Yu2iXf1ouERz8Vf391CqxHOnepHiG/view?usp=sharing](https://drive.google.com/file/d/1sE2Yu2iXf1ouERz8Vf391CqxHOnepHiG/view?usp=sharing) 

### Device info

 - Android version/Custom ROM version: Android 10 QP1A.190711.020
 - Device model: Xiaomi Redmi Note 8 Pro

TiA4f8R:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)

This PR:
- removes a toast which is shown when no app is able to handle the maket scheme on user's device (a toast is only shown if no browser is installed).
- opens the popup player for timestamps in the description of contents.
- fixes action of open with Kodi button in the player
- adds title of the shared content in Android Share Sheet (only for Android 10+ users)
- does some other improvements.
- adds support for hashtags in descriptions, which open for each a search for the hastag in the current service of the content

~~**TO DO**: add a proper support of disposables (I can't do it myself)~~ done

Fixes #5507, fixes https://github.com/TeamNewPipe/NewPipe/issues/5455#issuecomment-770903090, fixes #5695

#### APK testing 

#### Due diligence

(Sorry in advance for my English)

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
@@ -213,13 +213,21 @@ public final class ShareUtils {
      * @param url     the url to share
      */
     public static void shareText(final Context context, final String subject, final String url) {
+        shareText(context, subject, url, true);
+    }
+
+
+    public static void shareText(final Context context,
+                                 final String subject,
+                                 final String url,
+                                 final boolean showPreviewText) {
         final Intent shareIntent = new Intent(Intent.ACTION_SEND);
         shareIntent.setType("text/plain");
-        if (!subject.isEmpty()) {
+        if (!subject.isEmpty() && showPreviewText) {
             shareIntent.putExtra(Intent.EXTRA_SUBJECT, subject);
+            shareIntent.putExtra(Intent.EXTRA_TITLE, subject);
         }
         shareIntent.putExtra(Intent.EXTRA_TEXT, url);
-        shareIntent.putExtra(Intent.EXTRA_TITLE, context.getString(R.string.share_dialog_title));
 
         openAppChooser(context, shareIntent, false);
     }
--- a/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
@@ -25,9 +25,9 @@ public final class ShareUtils {
      * second param (a system chooser will be opened if there are multiple markets and no default)
      * and falls back to Google Play Store web URL if no app to handle the market scheme was found.
      * <p>
-     * It uses {@link ShareUtils#openIntentInApp(Context, Intent)} to open market scheme and
-     * {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} to open Google Play Store web
-     * URL with false for the boolean param.
+     * It uses {@link ShareUtils#openIntentInApp(Context, Intent, boolean)} to open market scheme
+     * and {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} to open Google Play Store
+     * web URL with false for the boolean param.
      *
      * @param context   the context to use
      * @param packageId the package id of the app to be installed
@@ -36,7 +36,7 @@ public final class ShareUtils {
         // Try market:// scheme
         final boolean marketSchemeResult = openIntentInApp(context, new Intent(Intent.ACTION_VIEW,
                 Uri.parse("market://details?id=" + packageId))
-                .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK));
+                .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK), false);
         if (!marketSchemeResult) {
             // Fall back to Google Play Store Web URL (F-Droid can handle it)
             openUrlInBrowser(context,
@@ -48,7 +48,7 @@ public final class ShareUtils {
      * Open the url with the system default browser.
      * <p>
      * If no browser is set as default, fallbacks to
-     * {@link ShareUtils#openAppChooser(Context, Intent, String)}
+     * {@link ShareUtils#openAppChooser(Context, Intent, boolean)}
      *
      * @param context                the context to use
      * @param url                    the url to browse
@@ -71,7 +71,7 @@ public final class ShareUtils {
 
         if (defaultPackageName.equals("android")) {
             // No browser set as default (doesn't work on some devices)
-            openAppChooser(context, intent, context.getString(R.string.open_with));
+            openAppChooser(context, intent, true);
         } else {
             if (defaultPackageName.isEmpty()) {
                 // No app installed to open a web url
@@ -84,7 +84,7 @@ public final class ShareUtils {
                 } catch (final ActivityNotFoundException e) {
                     // Not a browser but an app chooser because of OEMs changes
                     intent.setPackage(null);
-                    openAppChooser(context, intent, context.getString(R.string.open_with));
+                    openAppChooser(context, intent, true);
                 }
             }
         }
@@ -96,7 +96,7 @@ public final class ShareUtils {
      * Open the url with the system default browser.
      * <p>
      * If no browser is set as default, fallbacks to
-     * {@link ShareUtils#openAppChooser(Context, Intent, String)}
+     * {@link ShareUtils#openAppChooser(Context, Intent, boolean)}
      * <p>
      * This calls {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} with true
      * for the boolean parameter
@@ -116,22 +116,29 @@ public final class ShareUtils {
      * {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} should be used.
      * <p>
      * If no app is set as default, fallbacks to
-     * {@link ShareUtils#openAppChooser(Context, Intent, String)}
+     * {@link ShareUtils#openAppChooser(Context, Intent, boolean)}.
+     * <p>
      *
-     * @param context the context to use
-     * @param intent  the intent to open
+     * @param context   the context to use
+     * @param intent    the intent to open
+     * @param showToast the boolean to set if a toast is displayed to user when no app is installed
+     *                  to open the intent (true) or not (false)
      * @return true if the intent can be opened or false if it cannot be
      */
-    public static boolean openIntentInApp(final Context context, final Intent intent) {
+    public static boolean openIntentInApp(final Context context, final Intent intent,
+                                          final boolean showToast) {
         final String defaultPackageName = getDefaultAppPackageName(context, intent);
 
         if (defaultPackageName.equals("android")) {
             // No app set as default (doesn't work on some devices)
-            openAppChooser(context, intent, context.getString(R.string.open_with));
+            openAppChooser(context, intent, true);
         } else {
             if (defaultPackageName.isEmpty()) {
                 // No app installed to open the intent
-                Toast.makeText(context, R.string.no_app_to_open_intent, Toast.LENGTH_LONG).show();
+                if (showToast) {
+                    Toast.makeText(context, R.string.no_app_to_open_intent, Toast.LENGTH_LONG)
+                            .show();
+                }
                 return false;
             } else {
                 try {
@@ -140,7 +147,7 @@ public final class ShareUtils {
                 } catch (final ActivityNotFoundException e) {
                     // Not an app to open the intent but an app chooser because of OEMs changes
                     intent.setPackage(null);
-                    openAppChooser(context, intent, context.getString(R.string.open_with));
+                    openAppChooser(context, intent, true);
                 }
             }
         }
@@ -152,18 +159,25 @@ public final class ShareUtils {
      * Open the system chooser to launch an intent.
      * <p>
      * This method opens an {@link android.content.Intent#ACTION_CHOOSER} of the intent putted
-     * as the viewIntent param. A string for the chooser's title must be passed as the last param.
+     * as the intent param. If the setTitleChooser boolean is true, the string "Open with" will be
+     * set as the title of the system chooser.
+     * For Android P and higher, title for {@link android.content.Intent#ACTION_SEND} system
+     * choosers must be set on this intent, not on the
+     * {@link android.content.Intent#ACTION_CHOOSER} intent.
      *
-     * @param context              the context to use
-     * @param intent               the intent to open
-     * @param chooserStringTitle   the string of chooser's title
+     * @param context         the context to use
+     * @param intent          the intent to open
+     * @param setTitleChooser set the title "Open with" to the chooser if true, else not
      */
-    private static void openAppChooser(final Context context, final Intent intent,
-                                       final String chooserStringTitle) {
+    private static void openAppChooser(final Context context,
+                                       final Intent intent,
+                                       final boolean setTitleChooser) {
         final Intent chooserIntent = new Intent(Intent.ACTION_CHOOSER);
         chooserIntent.putExtra(Intent.EXTRA_INTENT, intent);
-        chooserIntent.putExtra(Intent.EXTRA_TITLE, chooserStringTitle);
         chooserIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+        if (setTitleChooser) {
+            chooserIntent.putExtra(Intent.EXTRA_TITLE, context.getString(R.string.open_with));
+        }
         context.startActivity(chooserIntent);
     }
 
@@ -201,10 +215,13 @@ public final class ShareUtils {
     public static void shareText(final Context context, final String subject, final String url) {
         final Intent shareIntent = new Intent(Intent.ACTION_SEND);
         shareIntent.setType("text/plain");
-        shareIntent.putExtra(Intent.EXTRA_SUBJECT, subject);
+        if (!subject.isEmpty()) {
+            shareIntent.putExtra(Intent.EXTRA_SUBJECT, subject);
+        }
         shareIntent.putExtra(Intent.EXTRA_TEXT, url);
+        shareIntent.putExtra(Intent.EXTRA_TITLE, context.getString(R.string.share_dialog_title));
 
-        openAppChooser(context, shareIntent, context.getString(R.string.share_dialog_title));
+        openAppChooser(context, shareIntent, false);
     }
 
     /**
--- a/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
@@ -8,6 +8,7 @@ import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.content.pm.ResolveInfo;
 import android.net.Uri;
+import android.os.Build;
 import android.widget.Toast;
 
 import androidx.core.content.ContextCompat;
@@ -25,15 +26,15 @@ public final class ShareUtils {
      * second param (a system chooser will be opened if there are multiple markets and no default)
      * and falls back to Google Play Store web URL if no app to handle the market scheme was found.
      * <p>
-     * It uses {@link ShareUtils#openIntentInApp(Context, Intent, boolean)} to open market scheme
-     * and {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} to open Google Play Store
+     * It uses {@link #openIntentInApp(Context, Intent, boolean)} to open market scheme
+     * and {@link #openUrlInBrowser(Context, String, boolean)} to open Google Play Store
      * web URL with false for the boolean param.
      *
      * @param context   the context to use
      * @param packageId the package id of the app to be installed
      */
     public static void installApp(final Context context, final String packageId) {
-        // Try market:// scheme
+        // Try market scheme
         final boolean marketSchemeResult = openIntentInApp(context, new Intent(Intent.ACTION_VIEW,
                 Uri.parse("market://details?id=" + packageId))
                 .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK), false);
@@ -48,7 +49,7 @@ public final class ShareUtils {
      * Open the url with the system default browser.
      * <p>
      * If no browser is set as default, fallbacks to
-     * {@link ShareUtils#openAppChooser(Context, Intent, boolean)}
+     * {@link #openAppChooser(Context, Intent, boolean)}
      *
      * @param context                the context to use
      * @param url                    the url to browse
@@ -56,7 +57,8 @@ public final class ShareUtils {
      *                               for HTTP protocol or for the created intent
      * @return true if the URL can be opened or false if it cannot
      */
-    public static boolean openUrlInBrowser(final Context context, final String url,
+    public static boolean openUrlInBrowser(final Context context,
+                                           final String url,
                                            final boolean httpDefaultBrowserTest) {
         final String defaultPackageName;
         final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(url))
@@ -96,9 +98,9 @@ public final class ShareUtils {
      * Open the url with the system default browser.
      * <p>
      * If no browser is set as default, fallbacks to
-     * {@link ShareUtils#openAppChooser(Context, Intent, boolean)}
+     * {@link #openAppChooser(Context, Intent, boolean)}
      * <p>
-     * This calls {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} with true
+     * This calls {@link #openUrlInBrowser(Context, String, boolean)} with true
      * for the boolean parameter
      *
      * @param context the context to use
@@ -113,19 +115,20 @@ public final class ShareUtils {
      * Open an intent with the system default app.
      * <p>
      * The intent can be of every type, excepted a web intent for which
-     * {@link ShareUtils#openUrlInBrowser(Context, String, boolean)} should be used.
+     * {@link #openUrlInBrowser(Context, String, boolean)} should be used.
      * <p>
      * If no app is set as default, fallbacks to
-     * {@link ShareUtils#openAppChooser(Context, Intent, boolean)}.
+     * {@link #openAppChooser(Context, Intent, boolean)}.
      * <p>
      *
      * @param context   the context to use
      * @param intent    the intent to open
-     * @param showToast the boolean to set if a toast is displayed to user when no app is installed
+     * @param showToast a boolean to set if a toast is displayed to user when no app is installed
      *                  to open the intent (true) or not (false)
      * @return true if the intent can be opened or false if it cannot be
      */
-    public static boolean openIntentInApp(final Context context, final Intent intent,
+    public static boolean openIntentInApp(final Context context,
+                                          final Intent intent,
                                           final boolean showToast) {
         final String defaultPackageName = getDefaultAppPackageName(context, intent);
 
@@ -178,6 +181,36 @@ public final class ShareUtils {
         if (setTitleChooser) {
             chooserIntent.putExtra(Intent.EXTRA_TITLE, context.getString(R.string.open_with));
         }
+
+        // Migrate any clip data and flags from the original intent.
+        final int permFlags;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            permFlags = intent.getFlags() & (Intent.FLAG_GRANT_READ_URI_PERMISSION
+                    | Intent.FLAG_GRANT_WRITE_URI_PERMISSION
+                    | Intent.FLAG_GRANT_PERSISTABLE_URI_PERMISSION
+                    | Intent.FLAG_GRANT_PREFIX_URI_PERMISSION);
+        } else {
+            permFlags = intent.getFlags() & (Intent.FLAG_GRANT_READ_URI_PERMISSION
+                    | Intent.FLAG_GRANT_WRITE_URI_PERMISSION
+                    | Intent.FLAG_GRANT_PERSISTABLE_URI_PERMISSION);
+        }
+        if (permFlags != 0) {
+            ClipData targetClipData = intent.getClipData();
+            if (targetClipData == null && intent.getData() != null) {
+                final ClipData.Item item = new ClipData.Item(intent.getData());
+                final String[] mimeTypes;
+                if (intent.getType() != null) {
+                    mimeTypes = new String[] {intent.getType()};
+                } else {
+                    mimeTypes = new String[] {};
+                }
+                targetClipData = new ClipData(null, mimeTypes, item);
+            }
+            if (targetClipData != null) {
+                chooserIntent.setClipData(targetClipData);
+                chooserIntent.addFlags(permFlags);
+            }
+        }
         context.startActivity(chooserIntent);
     }
 
@@ -208,24 +241,45 @@ public final class ShareUtils {
     /**
      * Open the android share menu to share the current url.
      *
-     * @param context the context to use
-     * @param subject the url subject, typically the title
-     * @param url     the url to share
+     * @param context         the context to use
+     * @param subject         the url subject, typically the title
+     * @param url             the url to share
+     * @param imagePreviewUrl the image of the subject
      */
-    public static void shareText(final Context context, final String subject, final String url) {
-        shareText(context, subject, url, true);
-    }
-
-
     public static void shareText(final Context context,
                                  final String subject,
                                  final String url,
+                                 final String imagePreviewUrl) {
+        shareText(context, subject, url, imagePreviewUrl, true);
+    }
+
+    /**
+     * Open the android share sheet to share the current url.
+     *
+     * For Android 10+ users, a content preview is shown, which includes the title of the shared
+     * content.
+     * Support sharing the image of the content needs to done, if possible.
+     *
+     * @param context         the context to use
+     * @param subject         the url subject, typically the title
+     * @param url             the url to share
+     * @param imagePreviewUrl the image of the subject
+     * @param showPreviewText show the subject as an extra title of the Android share sheet if true
+     */
+    public static void shareText(final Context context,
+                                 final String subject,
+                                 final String url,
+                                 final String imagePreviewUrl,
                                  final boolean showPreviewText) {
         final Intent shareIntent = new Intent(Intent.ACTION_SEND);
         shareIntent.setType("text/plain");
-        if (!subject.isEmpty() && showPreviewText) {
-            shareIntent.putExtra(Intent.EXTRA_SUBJECT, subject);
+        if (!imagePreviewUrl.isEmpty() && !subject.isEmpty() && showPreviewText) {
             shareIntent.putExtra(Intent.EXTRA_TITLE, subject);
+            /* TODO: add the image of the content to Android share sheet with setClipData after
+                generating a content URI of this image, then use ClipData.newUri(the content
+                resolver, null, the content URI) and set the ClipData to the share intent with
+                shareIntent.setClipData(generated ClipData).*/
+            //shareIntent.setFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
         }
         shareIntent.putExtra(Intent.EXTRA_TEXT, url);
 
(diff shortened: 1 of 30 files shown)


TiA4f8R:
@skyGtm Does the APK in the CI works as you want in #5507?

skyGtm:
Yes, it worked as expected.

But I found 2 new issues:
1. Clicking timestamp shows Toast "Playing in popup mode", even if the player is already in popup mode.
(Its ok for now, but in future surely someone will open an issue)

2. Crashes when clicking http/https links at the description.(No such behaviour was observed in previous version of the app)
This bug is somewhat serious, it crashed again after reopening. (Video demo below)
"Guru Meditation" wasn't even able to capture the crash log.

Luckily, my phone is rooted and Scoop App installed which captured the crash log.
(_I don't think root and Xposed hook methods are any problem
as the previous versions of NewPipe used to work without any issues_)

(Video demo and Crash log are posted below)

https://user-images.githubusercontent.com/50160188/106504000-73a32d80-64ee-11eb-8c92-305d67d18c35.mp4

[crash 1.txt](https://github.com/TeamNewPipe/NewPipe/files/5906099/crash.1.txt)
[crash 2.txt](https://github.com/TeamNewPipe/NewPipe/files/5906102/crash.2.txt)

TiA4f8R:
Looks very strange, it perfectly works on my device, an Honor 9X running EMUI 10. In order to test, I build a debug APK: [Debug share-improvements APK build with JDK 15.zip](https://github.com/TeamNewPipe/NewPipe/files/5906219/Debug.share-improvements.APK.build.with.JDK.15.zip).
Does it works for you (may be something is wrong with your ROM)?

skyGtm:
Nope.
Again after crash if I open the app again, it goes to app not responding state.
In the title you mentioned about about Google Playstore url, so not having 
PlayStore might be the cause? Actually, I had deleted Play Store and play 
services as it was taking huge space on my phone.

TiA4f8R:
@skyGtm No because it opens a web url so if there is a browser on your device, it will open the URL, unless it will show a message which is `No app on your device can open this`. F-Droid can also handle the Google Play Store URL (`https://play.google.com/store/apps/details?id=` + `app_package_id`).

skyGtm:
My phone has Android OS 5.1 API 22.
I also tried running in android emulator with similar configuration as my phone in my laptop.
But still same issue persists.
On clicking hyperlinks, app crashes.
On reopening it, app freezes with black screen.

_Trying to run in emulator might help you to find the cause_

TiA4f8R:
@skyGtm Does this APK fix the crash: [Share-improvements debug APK.zip](https://github.com/TeamNewPipe/NewPipe/files/5911743/Share-improvements.debug.APK.zip)?

skyGtm:
Yes. now it works perfectly.
Thankyou.

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
```diff
@@ -0,0 +1,113 @@
...
+     * 0 for {@code &t=} (used for timestamps in descriptions),
+     * 1 for {@code #timestamp=} (used for timestamps in comments).
+     * Any other value of this integer will return false.
+     *
+     * @param context       the context to be used
+     * @param url           the URL to check if it can be handled
+     * @param timestampType the type of timestamp
+     * @return true if the URL can be handled by NewPipe, false if it cannot
+     */
+    public static boolean handleUrl(final Context context, final String url, final int timestampType) {
```
TobiGr: While you are at it: please rename this method to `canHandleUrl` to make it more clear what this method is supposed to do.

TiA4f8R: Yeah, this makes sense. I will change it later, when I have some time.

prateekmedia:
@TiA4f8R Can you add a option to select what player to open timestamp from Content settings, https://github.com/TeamNewPipe/NewPipe/issues/5507#issuecomment-773301379_

Also in some videos description timestamp are not clickable on newpipe orignal one + this one too, but are recognized by player with timestamps menu.

Stypox:
@prateekmedia this is not the place to ask

TiA4f8R:
@TobiGr @Stypox On some videos, timestamps are not recognized by NewPipe. Here is an example (this video is one of the Martin Garrix Radio Show episodes on YouTube):

Unrecognized timestamps

These timestamps are supported in the YouTube app and in the desktop website, but not in the mobile website.

Do you know if I can implement the support of these timestamps in this PR? If yes, can you give how to do (and maybe some code to help me; I think I should use a regex, but I don't know what its expression is)?

prateekmedia:
>Do you know if I can implement the support of these timestamps in this PR? If yes, can you give how to do (and maybe some code to help me; I think I should use a regex, but I don't know what its expression is)?

It should use the same algorithm by which the extractor extracts the timestamps and show it in the player as a menu in latest newpipe.

TiA4f8R:
@prateekmedia The code you linked is the code for YouTube chapters support. This is **not** a code to support timestamps in the description of a content.

TiA4f8R:
@TobiGr What do you think of the last feature that I tried to implement? For now I found only one bug: two toasts `Playing in popup mode` are shown when playing a content with a timestamp.

SameenAhnaf:
Faced no crash. But still facing #5507

IMG_20210304_230557

TiA4f8R:
Can you please share the URL of the content where it happens?

TiA4f8R:
@SameenAhnaf There is no timestamp in this video. I think what you said is a different issue, which is more a feature request (supported YouTube links should open in the app), maybe I will add it.

SameenAhnaf:
@TiA4f8R Oh, yes! I actually didn't notice that the issue was for timestamps. Anyway, as you can see on screenshot, the app should still let users choose other browser or choose a default browser for supported links (which is not currently available) in the app would be nice.

Plus, clicking on any timestamp should open the video in main player, not on popup. However, clicking on external link should open the video (if Newpipe is chosen) according to "Preferred Open Action".

And, yes, "Copying video URL from the timestamp" and "Copy video URL at current time" would also be nice additions. This will let us mark a certain part of the video easily for our friends.

TiA4f8R:
@SameenAhnaf This is not the place to ask. There is issues for that.

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
```diff
@@ -0,0 +1,121 @@
...
+     * 0 for {@code &t=} (used for timestamps in descriptions),
+     * 1 for {@code #timestamp=} (used for timestamps in comments).
+     * Any other value of this integer will return false.
+     *
+     * @param context       the context to be used
+     * @param url           the URL to check if it can be handled
+     * @param timestampType the type of timestamp
+     * @return true if the URL can be handled by NewPipe, false if it cannot
+     */
+    public static boolean canHandleUrl(final Context context,
```
TobiGr, line 42: I was wrong with renaming this method.
It actually returns whether the URL was handled by NewPipe (either by opening a popup or through the RouterActivity). So the method needs to be re-renamed and the JDoc altered, too. Sorry for the inconvinience. 

-------

**Note for the future:** We should not play the video in a popup by default. Instead, if astream which represents the given url is currently playing, the current timestamp should be modified. 
Otherwise we could do one of the following
1. open the stream in popup,
1. enqueue the stream and start playing it,
1. ask the user what to do with the stream (e.g. when long-pressing the link) 

But if I didn't miss anything, this needs some deep refactoring to make the player queues available in NavigationHelper or the PlayerHelper.

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
```diff
@@ -0,0 +1,121 @@
...
+                                       final String url,
+                                       final int timestampType) {
+        String matchedUrl = "";
+        int seconds = -1;
+        final Pattern timestampPattern;
+
+        if (timestampType == 0) {
+            timestampPattern = Pattern.compile("(.*)&t=(\\d+)");
+        } else if (timestampType == 1) {
+            timestampPattern = Pattern.compile("(.*)#timestamp=(\\d+)");
```
TobiGr, line 52: I am not sure how often this method is going to be called. It could be a good idea to store both compiled patterns as static class vars.

app/src/main/java/org/schabi/newpipe/util/TextLinkifier.java
```diff
@@ -30,92 +37,175 @@ private TextLinkifier() {
...
+
+        while (timestampMatches.find()) {
+            final int timestampStart = timestampMatches.start(0);
+            final int timestampEnd = timestampMatches.end(0);
+            final String parsedTimestamp = descriptionText.substring(timestampStart, timestampEnd);
+            final String[] timestampParts = parsedTimestamp.split(":");
+            final int seconds;
+            if (timestampParts.length == 3) { // timestamp format: XX:XX:XX
+                seconds = Integer.parseInt(timestampParts[0]) * 3600 + Integer.parseInt(
+                        timestampParts[1]) * 60 + Integer.parseInt(timestampParts[2]);
```
TobiGr, line 150: For better readbility:
```suggestion
                seconds = Integer.parseInt(timestampParts[0]) * 3600 // hours
                        + Integer.parseInt(timestampParts[1]) * 60   // minutes
                        + Integer.parseInt(timestampParts[2]);       // seconds
```

app/src/main/java/org/schabi/newpipe/util/TextLinkifier.java
```diff
@@ -30,92 +37,175 @@ private TextLinkifier() {
...
+                spannableDescription.setSpan(new ClickableSpan() {
+                    @Override
+                    public void onClick(@NonNull final View view) {
+                        playOnPopup(context, contentUrl, streamingService, seconds);
+                    }
+                }, timestampStart, timestampEnd, 0);
+            } else if (timestampParts.length == 2) { // timestamp format: XX:XX
+                seconds = Integer.parseInt(timestampParts[0]) * 60 + Integer.parseInt(
+                        timestampParts[1]);
```
TobiGr, line 159:
```suggestion
                seconds = Integer.parseInt(timestampParts[0]) * 60 // minutes
                        + Integer.parseInt(timestampParts[1]);     // seconds
```

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
TiA4f8R: > But if I didn't miss anything, this needs some deep refactoring to make the player queues available in NavigationHelper or the PlayerHelper.

Yes, always open popup player when clicking on any timestamp is a bad idea. However, I will not take care of changing this.

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -208,24 +241,45 @@ private static String getDefaultAppPackageName(final Context context, final Inte
...
         shareIntent.setType("text/plain");
-        if (!subject.isEmpty() && showPreviewText) {
-            shareIntent.putExtra(Intent.EXTRA_SUBJECT, subject);
+        if (!imagePreviewUrl.isEmpty() && !subject.isEmpty() && showPreviewText) {
             shareIntent.putExtra(Intent.EXTRA_TITLE, subject);
+            /* TODO: add the image of the content to Android share sheet with setClipData after
+                generating a content URI of this image, then use ClipData.newUri(the content
+                resolver, null, the content URI) and set the ClipData to the share intent with
+                shareIntent.setClipData(generated ClipData).*/
+            //shareIntent.setFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
```
TiA4f8R, line 282: Does someone have an idea how to do this, and do not download again the image of the content if possible?

Stypox: @TiA4f8R I don't know if this is doable, it would require the library we are using for images to provide ContentUris of cached images. I'd skip this for now and create a different PR later.

TiA4f8R: Ok, I will let you do this later.

Stypox: > Ok, I will let you do this later. 

Ok, I will try, though I don't promise anything ;-)

> Should I keep the comment?

Yes, keep the comment, thank you

app/src/main/java/org/schabi/newpipe/util/TextLinkifier.java
```diff
@@ -30,92 +39,173 @@ private TextLinkifier() {
...
+     *                             content description
+     * @param contentUrl           the URL of the content
+     * @param streamingService     the {@link StreamingService} of the content
+     */
+    private static void addClickListenersOnTimestamps(final Context context,
+                                                      final SpannableStringBuilder
+                                                              spannableDescription,
+                                                      final String contentUrl,
+                                                      final StreamingService streamingService) {
+        final String descriptionText = spannableDescription.toString();
```
Stypox, line 138: Are you making sure that the timestamp is not inside an url or attached to a word?

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
```diff
@@ -194,17 +241,49 @@ private static String getDefaultAppPackageName(final Context context, final Inte
...
+     * @param url             the url to share
+     * @param imagePreviewUrl the image of the subject
+     * @param showPreviewText show the subject as an extra title of the Android share sheet if true
      */
-    public static void shareText(final Context context, final String subject, final String url) {
+    public static void shareText(final Context context,
+                                 final String subject,
+                                 final String url,
+                                 final String imagePreviewUrl,
+                                 final boolean showPreviewText) {
```
Stypox, line 273: Do we really need a switch to disable preview text? Why not just enable it always?

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
```diff
@@ -0,0 +1,124 @@
...
+import org.schabi.newpipe.player.playqueue.SinglePlayQueue;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
+import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.schedulers.Schedulers;
+
+public final class URLHandler {
```
Stypox, line 21: The name of this class is not meaningful enough. Something like DescriptionUrlHandler would be better (suggest a better one if you have one).

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
```diff
@@ -0,0 +1,124 @@
+package org.schabi.newpipe.util;
```
Stypox, line 1: Maybe you could create a subpackage of `util` containing the classes related to sharing, open intents, description url handling, etc... The `util` package has grown too big.

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
TiA4f8R: I was thinking it was necessary for error reports, but we can show the translated string `Error report`.

app/src/main/java/org/schabi/newpipe/util/TextLinkifier.java
TiA4f8R: No, I asked myself the question. Do you have a better regular expression?

Stypox: Adding `(?:^|\s)` at the beginning of the regex and `(?:\s|$)` at the end should do it, but do some checks [here](https://regex101.com/) to make sure.

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
Stypox: Without introducing more complexity I'd say `subject` should just be the translated `Error report`, what do you think?

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
TiA4f8R: I don't have a better for now, DescriptionUrlsHandler.

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
TiA4f8R: That's I want to say.

app/src/main/java/org/schabi/newpipe/util/TextLinkifier.java
TiA4f8R: @Stypox Timestamps are successfully parsed, but results include spaces and linebreaks:

Screenshot

Is it normal or a better regex is needed?

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
TiA4f8R: Do you have a name for this subpackage? I can't find an appropriate/a coherent one.

app/src/main/java/org/schabi/newpipe/util/TextLinkifier.java
Stypox: That's normal, if you look at the capturing group content you will see it only contains the timestamp, but the regex also matches the spaces since it makes sure there is nothing at the beginning nor at the end.

TiA4f8R: So, I need to remove spaces when adding click listeners or not?

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
Stypox: `open`? Every action contained is related to opening some piece of information with another app (or NewPipe itself). What do you think?

app/src/main/java/org/schabi/newpipe/util/TextLinkifier.java
Stypox: Yes, you need to remove spaces, I think

Stypox:
I'd postpone this to 0.21.1, it requires deep testing

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
TiA4f8R: external_communication?

Stypox: That's better, probably too long but who cares ;-)
I'd go with this

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
TiA4f8R: This code is used for descriptions and comments, so I think URLsHandler is the best one.

app/src/main/java/org/schabi/newpipe/util/TextLinkifier.java
TiA4f8R: I found better:
```java
            final int timestampStart = timestampMatches.start(2);
            final int timestampEnd = timestampMatches.end(3);
```

app/src/main/java/org/schabi/newpipe/util/URLHandler.java
Stypox: What about InternalUrlHandler? This would help distinguishing between the extractor's url handlers

TiA4f8R: Done

app/src/main/java/org/schabi/newpipe/util/ShareUtils.java
TiA4f8R: Done, parameter removed.

app/src/main/java/org/schabi/newpipe/util/external_communication/TextLinkifier.java
```diff
@@ -0,0 +1,281 @@
...
+        final String descriptionText = spannableDescription.toString();
+        final Matcher timestampsMatches = TIMESTAMPS_PATTERN.matcher(descriptionText);
+
+        while (timestampsMatches.find()) {
+            final int timestampStart = timestampsMatches.start(2);
+            final int timestampEnd = timestampsMatches.end(3);
+            final String parsedTimestamp = descriptionText.substring(timestampStart, timestampEnd);
+            final String[] timestampParts = parsedTimestamp.split(":");
+            final int time;
+            if (timestampParts.length == 3) { // timestamp format: XX:XX:XX
```
Stypox, line 184:
```suggestion
            final int time;

            if (timestampParts.length == 3) { // timestamp format: XX:XX:XX
```

app/src/main/java/org/schabi/newpipe/util/external_communication/TextLinkifier.java
```diff
@@ -0,0 +1,281 @@
...
+        while (hashtagsMatches.find()) {
+            final int hashtagStart = hashtagsMatches.start(1);
+            final int hashtagEnd = hashtagsMatches.end(1);
+            final String parsedHashtag = descriptionText.substring(hashtagStart, hashtagEnd);
+
+            spannableDescription.setSpan(new ClickableSpan() {
+                @Override
+                public void onClick(@NonNull final View view) {
+                    NavigationHelper.openSearch(context, streamingService.getServiceId(),
+                            parsedHashtag);
```
Stypox, line 151: Shouldn't you search for the value of the hashtag without the `#` prefixed? (maybe not)

app/src/main/java/org/schabi/newpipe/util/external_communication/TextLinkifier.java
```diff
@@ -0,0 +1,281 @@
...
+            if (timestampParts.length == 3) { // timestamp format: XX:XX:XX
+                time = Integer.parseInt(timestampParts[0]) * 3600 // hours
+                        + Integer.parseInt(timestampParts[1]) * 60 // minutes
+                        + Integer.parseInt(timestampParts[2]); // seconds
+                spannableDescription.setSpan(new ClickableSpan() {
+                    @Override
+                    public void onClick(@NonNull final View view) {
+                        playOnPopup(context, contentUrl, streamingService, time);
+                    }
+                }, timestampStart, timestampEnd, 0);
```
Stypox, line 193: Move this out of the `if else`, and add an `else continue` at the end to prevent (impossible) false matches

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -0,0 +1,126 @@
...
+     * 0 for {@code &t=} (used for timestamps in descriptions),
+     * 1 for {@code #timestamp=} (used for timestamps in comments).
+     * Any other value of this integer will return false.
+     *
+     * @param context       the context to be used
+     * @param url           the URL to check if it can be handled
+     * @param timestampType the type of timestamp
+     * @return true if the URL can be handled by NewPipe, false if it cannot
+     */
+    public static boolean handleUrl(final Context context,
```
Stypox, line 47: I think this function's code can be heavily reduced, or at least made less dispersive

app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
```diff
@@ -377,19 +372,16 @@ private void shareFile(Mission mission) {
         shareIntent.setType(resolveMimeType(mission));
         shareIntent.putExtra(Intent.EXTRA_STREAM, resolveShareableUri(mission));
         shareIntent.addFlags(FLAG_GRANT_READ_URI_PERMISSION);
+
         final Intent intent = new Intent(Intent.ACTION_CHOOSER);
         intent.putExtra(Intent.EXTRA_INTENT, shareIntent);
-        intent.putExtra(Intent.EXTRA_TITLE, mContext.getString(R.string.share_dialog_title));
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.O_MR1) {
+            intent.putExtra(Intent.EXTRA_TITLE, mContext.getString(R.string.share_dialog_title));
```
Stypox, line 379: Please add a comment to explain why the `if` is needed

Stypox (changes requested):
Code looks mostly good :-)

app/src/main/java/org/schabi/newpipe/util/external_communication/TextLinkifier.java
TiA4f8R: No, I forgot this part.

TiA4f8R: Should be done now.

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -49,22 +49,20 @@ public static boolean handleUrl(final Context context,
...
             return false;
         }
 
-        final Matcher matcher = timestampPattern.matcher(url);
         if (matcher.matches()) {
             matchedUrl = matcher.group(1);
             seconds = Integer.parseInt(matcher.group(2));
         }
-
         if (matchedUrl == null || matchedUrl.isEmpty()) {
```
Stypox, line 66: Use `isNullOrEmpty` here

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -49,22 +49,20 @@ public static boolean handleUrl(final Context context,
                                     final int timestampType) {
```
Stypox, line 49: After thinking more about this, I think you should split this into two functions: `handleUrlDescriptionTimestamp(Context, String)` and `handleUrlCommentsTimestamp(Context, String)`, and those functions should both call the **private** `handleUrl(Context, String, Pattern)` with the relative pattern.

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -49,22 +49,20 @@ public static boolean handleUrl(final Context context,
...
-            timestampPattern = AMPERSAND_TIMESTAMP_PATTERN;
+            matcher = AMPERSAND_TIMESTAMP_PATTERN.matcher(url);
         } else if (timestampType == 1) {
-            timestampPattern = HASHTAG_TIMESTAMP_PATTERN;
+            matcher = HASHTAG_TIMESTAMP_PATTERN.matcher(url);
         } else {
             return false;
         }
 
-        final Matcher matcher = timestampPattern.matcher(url);
```
Stypox: Having a common matcher actually made sense, I think

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
TiA4f8R: Yes, good idea.

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
TiA4f8R: This is an outdated code, fixed.

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
TiA4f8R: Done

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+     * https://www.youtube.com/watch?v=}video_id{@code #timestamp=}time_in_seconds). If yes, the
+     * popup player will be opened when the user will click on the timestamp in the comment,
+     * at the time and for the video indicated in the timestamp.
+     *
+     * @param context the context to use
+     * @param url     the URL to check if it can be handled
+     * @return true if the URL can be handled by NewPipe, false if it cannot
+     */
+    public static boolean handleUrlCommentsTimestamp(final Context context, final String url) {
+        return handleUrl(context, url, HASHTAG_TIMESTAMP_PATTERN);
```
Stypox, line 46:
```suggestion
    public static boolean handleUrlCommentsTimestamp(final CompositeDisposable disposables, final Context context, final String url) {
        return handleUrl(disposables, context, url, HASHTAG_TIMESTAMP_PATTERN);
```

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+     * https://www.youtube.com/watch?v=}video_id{@code &t=}time_in_seconds). If yes, the popup
+     * player will be opened when the user will click on the timestamp in the video description,
+     * at the time and for the video indicated in the timestamp.
+     *
+     * @param context the context to use
+     * @param url     the URL to check if it can be handled
+     * @return true if the URL can be handled by NewPipe, false if it cannot
+     */
+    public static boolean handleUrlDescriptionTimestamp(final Context context, final String url) {
+        return handleUrl(context, url, AMPERSAND_TIMESTAMP_PATTERN);
```
Stypox, line 62:
```suggestion
    public static boolean handleUrlDescriptionTimestamp(final CompositeDisposable disposables, final Context context, final String url) {
        return handleUrl(disposables, context, url, AMPERSAND_TIMESTAMP_PATTERN);
```

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+     * else, false will be returned.
+     *
+     * @param context       the context to use
+     * @param url           the URL to check if it can be handled
+     * @param pattern       the pattern
+     * @return true if the URL can be handled by NewPipe, false if it cannot
+     */
+    private static boolean handleUrl(final Context context,
+                                     final String url,
+                                     final Pattern pattern) {
```
Stypox, line 79:
```suggestion
    private static boolean handleUrl(final CompositeDisposable disposables,
                                     final Context context,
                                     final String url,
                                     final Pattern pattern) {
```

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+            service = NewPipe.getServiceByUrl(matchedUrl);
+            linkType = service.getLinkTypeByUrl(matchedUrl);
+        } catch (final ExtractionException e) {
+            return false;
+        }
+        if (linkType == StreamingService.LinkType.NONE) {
+            return false;
+        }
+        if (linkType == StreamingService.LinkType.STREAM && seconds != -1) {
+            return playOnPopup(context, matchedUrl, service, seconds);
```
Stypox, line 105:
```suggestion
            return playOnPopup(disposables, context, matchedUrl, service, seconds);
```

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+     * @param context the context to be used
+     * @param url     the URL of the content
+     * @param service the service of the content
+     * @param seconds the position in seconds at which the floating player will start
+     * @return true if the playback of the content has successfully started or false if not
+     */
+    public static boolean playOnPopup(final Context context,
+                                      final String url,
+                                      final StreamingService service,
+                                      final int seconds) {
```
Stypox, line 124:
```suggestion
    public static boolean playOnPopup(final CompositeDisposable disposables,
                                      final Context context,
                                      final String url,
                                      final StreamingService service,
                                      final int seconds) {
```

app/src/main/java/org/schabi/newpipe/util/external_communication/InternalUrlsHandler.java
```diff
@@ -0,0 +1,145 @@
...
+
+        final Single<StreamInfo> single
+                = ExtractorHelper.getStreamInfo(service.getServiceId(), cleanUrl, false);
+        single.subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(info -> {
+                    final PlayQueue playQueue
+                            = new SinglePlayQueue(info, seconds * 1000);
+                    NavigationHelper.playOnPopupPlayer(context, playQueue, false);
+                });
```
Stypox, line 142:
```suggestion
        disposables.add(single.subscribeOn(Schedulers.io())
                .observeOn(AndroidSchedulers.mainThread())
                .subscribe(info -> {
                    final PlayQueue playQueue
                            = new SinglePlayQueue(info, seconds * 1000);
                    NavigationHelper.playOnPopupPlayer(context, playQueue, false);
                }));
```

Stypox (changes requested):

TobiGr (commented):
Looks good to me!
Just a small thing: please annotate the params with either `@NotNull` or `@Nullable`.

TobiGr (approved):

Stypox:
@TiA4f8R I rebased and added the last two commits. The first one improves the parameters being passed to the functions; the second one fixes the issues with disposables.

app/src/main/java/org/schabi/newpipe/util/external_communication/TextLinkifier.java
```diff
@@ -0,0 +1,283 @@
...
+
+                textBlockLinked.setSpan(clickableSpan, textBlockLinked.getSpanStart(span),
+                        textBlockLinked.getSpanEnd(span), textBlockLinked.getSpanFlags(span));
+                textBlockLinked.removeSpan(span);
+            }
+
+            // add click actions on plain text timestamps only for description of contents,
+            // unneeded for meta-info or other TextViews
+            if (relatedInfo != null) {
+                addClickListenersOnTimestamps(context, textBlockLinked, relatedInfo, disposables);
```
Stypox, line 261: Maybe this should be called only if `relatedInfo instanceof StreamInfo`? What do you think?

TiA4f8R:
Thank you @Stypox for your work on my PR! You are the best!

app/src/main/java/org/schabi/newpipe/util/external_communication/TextLinkifier.java
TiA4f8R: I think it's a good idea, I will commit it.

Stypox (approved):
Everything looks good in my opinion :-)
Thank you for this big chunk of work!
I tested everything I could (share, open in browser, kodi, timestamps, hashtags) and could not find any issues on API 19 emulated (with no play store), API 29 TV emulated (with no browser; with play store), API 29 phone (with Kore; with Foxy Droid and Aurora Store (Foxy Droid was correctly chosen)), API 30 emulated (with no play store).

PR #5563, opened 2021-02-12
Fix tab layout visibility with age restricted videos

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This fixes the tab layout visibility in video detail fragment when showing age restricted videos:
- never show tab layout if view pager is hidden
- always show view pager and tab layout again when a new stream info has to be shown in video detail fragment 

- https://github.com/TeamNewPipe/NewPipe/issues/5455#issuecomment-774741197

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -33,7 +33,6 @@ import android.widget.FrameLayout;
 import android.widget.RelativeLayout;
 
 import androidx.annotation.AttrRes;
-import androidx.annotation.DrawableRes;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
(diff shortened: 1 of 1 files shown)


TobiGr:
Please add comments into the code. 
That does not work on TV:
<details><summary>Screenshots</summary>





</details>

nadiration:
The layout problem is gone but could you make the play button to refresh the video since now when you go and enable restricted content on settings and come back to the video it doesn't [auto] refesh and the play button (on the minimized video) does nothing thus requiring you to close the video then you have to find the video again and play.
It would also be great if you hide tab layout for unsupported/geolocked or all other unavailable/unplayable videos. No pressure just pointing out. Thanks.

Stypox:
@TobiGr I added comments. Regarding your Android-TV issue: that's not tv-specific and is expected behaviour. The tab layout is visible because the view pager is actually visible and has its tabs, since e.g. the comments could have been loaded regardless of whether the video shows a network error or not. Should the view pager be hidden completely in case of main-video network errors?

Stypox:
@TobiGr I now pushed a commit that completely hides tabs both on age restricted content and on network errors (it can be reverted if unwanted).
One issue, which was there also in previous versions and was not caused by secondary-controls, is that the view pager layout is not scrollable when hidden, leading to strange behaviour. I already fixed this across all NewPipe in #5148 (in particular [this commit](c2841f13829b2405ba84c19aaefe8307d6beb2a1)), basically by setting the layout alpha value to `0.0f` instead of making it `GONE`, so that it keeps listening for touches. But this solution would require changes in many places (see how many files the commit changed) and proper testing, so I wouldn't include it in 0.20.10, also because it is not a regression.

TobiGr (approved):

Issue #5660, opened 2021-02-21
Snackbars queued when deleting downloads instead of flushing immediately


ix5:
### Checklist
- [x] I am using the latest version - x.xx.x 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo.
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug
1. Download multiple videos/audio files
2. Go to download page
3. Delete multiple videos quickly one after another

### Actual behaviour
Snackbars at the bottom of the UI keep queuing up. You need to wait for all of them to dismiss themselves for files to actually be deleted. If you close NewPipe before all snackbars are dismissed, your files will not be deleted and show up again next time you start NewPipe.

### Expected behavior
Deleting a second, third etc. file should flush the handler queue and immediately delete the previously marked-for-deletion file and dismiss its snackbar.

### Screenshots/Screen recordings

See [this screen recording](https://ix5.org/files/newpipe.mp4)

### Logs
(This bug is not about crashes)

### Device info

 - Android version/Custom ROM version: Android 11
 - Device model: Any, e.g. emulator

---

Now that this newbie template is filled out, let's get down to the actual issue.

Codepath:
https://github.com/TeamNewPipe/NewPipe/blob/1f77e00df41756185fdbe79e6c34e3b14736ce15/app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java#L690
https://github.com/TeamNewPipe/NewPipe/blob/1f77e00df41756185fdbe79e6c34e3b14736ce15/app/src/main/java/us/shandian/giga/ui/common/Deleter.java#L57-L62
https://github.com/TeamNewPipe/NewPipe/blob/1f77e00df41756185fdbe79e6c34e3b14736ce15/app/src/main/java/us/shandian/giga/ui/common/Deleter.java#L71-L78
This causes `next()` to be posted with a delay of 350ms (doc: [Handler.postDelayed](https://developer.android.com/reference/android/os/Handler#postDelayed(java.lang.Runnable,%20long)))

https://github.com/TeamNewPipe/NewPipe/blob/1f77e00df41756185fdbe79e6c34e3b14736ce15/app/src/main/java/us/shandian/giga/ui/common/Deleter.java#L80-L91

[...]

Issue #5419, opened 2021-01-15
Confirmation before deleting all downloads


kcpatricio:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [ x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [ ] This issue contains only one feature request. I will open one issue for every feature I want to request.

My suggestion is to make a pop up confirmation when selecting the "delete downloaded files" option or a trash can so we can recover those deleted files

Having a single touch wipe is way too dangerous for a backup
As my own experience, a single missclick cost all my backup of temporary videos I had for studying

SpinHit:
Hello we are two students who want to work on this issue, this is the first time we are contributing to a project if you could help us a little bit it would be great.

kcpatricio:
Yeah
How could I help you?

kcpatricio:
> Hello we are two students who want to work on this issue, this is the first time we are contributing to a project if you could help us a little bit it would be great.

Yeah
How could I help you?

XiangRongLin:
@kcpatricio Is this even still the case now?
You have to first click on the "trash" icon, then select "Delete downloaded files", which are 2 actions.

After confirming both, you get a snackbar at the bottom with the "undo" option.

kcpatricio:
> @kcpatricio Is this even still the case now?
> You have to first click on the "trash" icon, then select "Delete downloaded files", which are 2 actions.
> 
> After confirming both, you get a snackbar at the bottom with the "undo" option.

Yes
{
1 click (trash icon)
1 click (delete downloaded files)
*All files deleted*
3-4 seconds notification at the bottom with undo option
}
But no confirmation request before deleting
My suggestion is to prompt a warning about the wipe action, requesting a confirmation before executing it
Or to change from "delete" to "send to bin" where it would hold files till a given expiration date or manual deletion

SameenAhnaf:
@kcpatricio I am not sure why this process has to be so complicated. Personally, I think, a 'Trash' tab under "Downloads" could be added.

Let the videos under "Trash" tab completely undeletable. All information could be automatically deleted 30 days after deletion or after redownloading the video in same or any other resolution.

vhouriet:
[...]

PR #5671, opened 2021-02-22
Downloader: Deleter: Dismiss previous Snackbars

ix5:
#### What is it?
- [x] Bugfix (user facing)

When running `append()`, trigger the `commit()` action immediately and cancel all delayed callbacks for `commit()`.

Snackbars no longer pile up. For previous behaviour, see https://github.com/TeamNewPipe/NewPipe/issues/5660

(Checking if there are any tasks running via [hasCallbacks()](https://developer.android.com/reference/android/os/Handler#hasCallbacks(java.lang.Runnable)) would be nice, but that method is only available in API 29+.)

For a visual comparison:

The current behaviour in NewPipe 0.20.10
[Screencast](https://ix5.org/files/newpipe.mp4) (note: Snackbars keep piling up)

With this PR applied:
[Screencast](https://ix5.org/files/newpipe-dismiss.mp4) (note: No piling up, but users can still undo actions)

https://github.com/TeamNewPipe/NewPipe/issues/5660

#### APK testing

[My server](https://sx.ix5.org/files/misc/newpipe-debug.apk)
[Generated from GitHub CI](https://github.com/TeamNewPipe/NewPipe/suites/2097924546/artifacts/42600853)

#### Due diligence

### DIFF
--- a/app/src/main/java/us/shandian/giga/ui/common/Deleter.java
+++ b/app/src/main/java/us/shandian/giga/ui/common/Deleter.java
@@ -55,6 +55,14 @@ public class Deleter {
     }
 
     public void append(Mission item) {
+
+        /* If a mission is removed from the list while the Snackbar for a previously
+         * removed item is still showing, commit the action for the previous item
+         * immediately. This prevents Snackbars from stacking up in reverse order.
+         */
+        mHandler.removeCallbacks(rCommit);
+        commit();
+
         mIterator.hide(item);
         items.add(0, item);
 


ix5:
@kapodamy what do you think?

XiangRongLin:
The idea would be fine if there were some system similar to a trashcan. Since we do not, I think this could cause accidental deletes.

See #5419 

@kcpatricio since you representing the other opinion, what's your take on this.

---

On a sidenote, since you titled it WIP you can convert it to a draft PR

ix5:
I don't remember whether converting to draft disallows inline reviews. But I've changed it.

ix5:
As for accidental deletes: The idea is that if a user deletes one file and then proceeds to delete another file, he must be pretty sure about the first one. It takes two taps to delete the next file, which no sane person would do if they noticed they accidentally deleted the wrong file.
What is, however, very annoying is the current behaviour where deleted files do *not* get deleted if one closes the app before all snackbars are gone (which takes `<number of items deleted> x 5 seconds`)

kcpatricio:
> The idea would be fine if there were some system similar to a trashcan. Since we do not, I think this could cause accidental deletes.
> 
> See #5419 
> 
> @kcpatricio since you representing the other opinion, what's your take on this.
> 
> ---
> 
> On a sidenote, since you titled it WIP you can convert it to a draft PR

Yes, a trashcan would be the safest option

Clearing the download's history is a  way to prevent accidental deletes since the 'delete downloaded files' only removes  those that still are in history. That way you'll have to browse to NewPipe's folder with another app to access your files 
Still, you are a few pixels and 2-3 seconds away from a wipe

ix5:
You might be conflating this PR with https://github.com/TeamNewPipe/NewPipe/issues/5419.

This PR explicitly does **not** address clearing **all** downloads. If you've cleared all downloads (deliberately or by accident), there is no possibility of triggering the codepath in this PR (no more downloads to click).

ix5:
I've added two screencasts for people to better understand the issue, see the original post at the top.

kcpatricio:
Here's mine

[Screencast](https://imgflip.com/gif/4z5a31)

ix5:
Again, this has nothing to do with the `Clear Download History` button.

The codepath for that action, which hard-deletes or "forgets" (files still present in filesystem) all downloads, lies here:

https://github.com/TeamNewPipe/NewPipe/blob/d6855a6b50349c74f6d87d867470c65254645c92/app/src/main/java/us/shandian/giga/ui/fragment/MissionsFragment.java#L190-L199
https://github.com/TeamNewPipe/NewPipe/blob/d6855a6b50349c74f6d87d867470c65254645c92/app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java#L599-L623

kapodamy:
@ix5 llooks good, now it's easier to delete multiple files

ix5:
I haven't found a better or more elegant way to do this, and it seems no one else has tackled this issue so far either.
Could I get a review, please?

Stypox (approved):
Looks good, this will be merged in next release.

ix5:
@Stypox any news?

ix5:
@Stypox would you mind merging this if you have no further objections?

TobiGr (approved):

Issue #5836, opened 2021-03-15
[Feature Request] Set channel groups in context menu and/or on long press


argonfilm:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

I want to be able to add individual channels to groups without needing to use that "Select Subscriptions" menu. I think this wouls best be done by adding a "groups" item to the context menu when displaying a channel. Something that could also be done is long pressing a channel from the subscriptions page (or even a video) and having a groups item pop up there.

#### Is your feature request related to a problem? Please describe it

I want to use groups really badly, but setting up all the channels without being able to preview their content and select the group that the channel fits best in is a real hassle. Having a way to select what groups a channel belongs to while viewing the channel would alleviate this, and I think it would become a pretty powerful and convenient feature.

#### Additional context

![Screenshot_20210315-052728](https://user-images.githubusercontent.com/22607364/111133672-271d2a00-8551-11eb-91bf-f6fa553001b3.png)
The menu in question

![Screenshot_20210315-052428](https://user-images.githubusercontent.com/22607364/111133740-3a2ffa00-8551-11eb-86be-c9daf05b01ad.png)
Possible convenient alternative (long press on a channel in the subscriptions page)

#### How will you/everyone benefit from this feature?

I think this could really make the groups feature shine with possibility. Being able to conveniently manage what group a channel is a part of could help people organize their subscriptions better. I think more people would repeatedly use groups as being able to load a selection of channels is already quicker and easier than refreshing the whole of your subscriptions.

SameenAhnaf:
I think this info should be accessed by long press only. Because, you will need to be online to access channel feed. Plus, channel feed should include things solely related to or uploaded by the channel, not something edited or added by me.

MD77MD:
Describe the feature you want
The ability to add subscribed channels to feed groups directly from channel page.

Is your feature request related to a problem? Please describe it
anyone uses the group feed will soon find it hard to create new groups or maintaining them because it becomes harder and harder to keep track. also subscribing to a new channel and adding it to multiple related groups is not easy in itself.

Additional context
[...]

Issue #5783, opened 2021-03-07
Add "Group" option for Long Press on Channel and Playlist


SameenAhnaf:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

I would be really grateful if grouping or categorising available for Long Press on playlist and channel. Sometimes, we save a single channel or playlist. It's a little tricky to group or categorise it from tiles.

#### Additional context

Sort and categorise (or group) playlists #5681

PR #5851, opened 2021-03-18
Add channel details option to long-press menu

### LINKED ISSUE #2171: [feature request] Add direct download option to long tap menu of a video in the news / channel list

Hi.
I live in an area with bad mobile network availability and slow connections. So usually I download all videos before watching. Esp when the connection is slow it sometimes takes 2 - 3 s (and even more) to load and decode the info page. For me it would be very helpful if I could directly start the download of a video from the news / abo / channel page without first opening the corresponding info page. If that is possible without increasing the data traffic load on the overview pages. Maybe get the download link after tapping on the (future) menu entry. Without (optionally) even asking for a file name for the downloaded file if that would also take 2+ s to decode the info before asking for a file name. So newpipe downloader could do its job in the background and I could go on scrolling through the list. Just an idea.
Thx

### LINKED ISSUE #5447: Longpress (menu) option for direct channel videos overview access (tablet optimisation)
### Checklist

A longpress menu option that takes me directly to the channel's videos overview (YT channel content).

#### Is your feature request related to a problem? 
Currently, with the 1.mainpage for searching and the 2."pop-up bar" for currently playing video, there's no way to access **another channel's** videos overview, unless a) I'm starting a new search for the channel or b) starting a new video to get access to the channel's name with a direct link to the channel's videos overview.

Other option:
UI element in the "pop-up bar" video player to jump back again to the previous video.
Currently, NewPipe does not keep track of the previously shown video, backtracking only happens between YT channel page and Newpipe start page.  

#### Additional context
This is all in relation that I most likely don't want to close my current video or search for the newly found channel, since the longpress option is much faster than typing.

Most of the time I'm using NewPipe on a tablet, typing even with optimized keyboard spacing is rather slow in comparison to a smartphone. Therefore I think a longpress menu option for direct YT channel's videos overview would come in handy and extend the usefulness for tablets without greater changes.

Happy and healthy new year 2021!

mhmdanas:
#### What is it?
- [x] Feature (user facing)

Add option to open channel details page when long-pressing items.

- Fixes #5447.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
@@ -124,8 +124,8 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I>
     /**
      * If the default implementation of {@link StateSaver.WriteRead} should be used.
      *
-     * @see StateSaver
      * @param useDefaultStateSaving Whether the default implementation should be used
+     * @see StateSaver
      */
     public void setUseDefaultStateSaving(final boolean useDefaultStateSaving) {
         this.useDefaultStateSaving = useDefaultStateSaving;
@@ -350,7 +350,8 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I>
             return;
         }
 
-        final ArrayList<StreamDialogEntry> entries = new ArrayList<>();
+        final ArrayList<StreamDialogEntry> entries =
+                new ArrayList<>(Arrays.asList(StreamDialogEntry.show_channel_details));
 
         if (PlayerHolder.getType() != null) {
             entries.add(StreamDialogEntry.enqueue);
@@ -361,7 +362,7 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I>
                     StreamDialogEntry.append_playlist,
                     StreamDialogEntry.share
             ));
-        } else  {
+        } else {
             entries.addAll(Arrays.asList(
                     StreamDialogEntry.start_here_on_background,
                     StreamDialogEntry.start_here_on_popup,
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
@@ -350,8 +350,8 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I>
             return;
         }
 
-        final List<StreamDialogEntry> entries =
-                new ArrayList<>(Arrays.asList(StreamDialogEntry.show_channel_details));
+        final List<StreamDialogEntry> entries = new ArrayList<>();
+        entries.add(StreamDialogEntry.show_channel_details)
 
         if (PlayerHolder.getType() != null) {
             entries.add(StreamDialogEntry.enqueue);
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
@@ -351,7 +351,6 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I>
         }
 
         final List<StreamDialogEntry> entries = new ArrayList<>();
-        entries.add(StreamDialogEntry.show_channel_details)
 
         if (PlayerHolder.getType() != null) {
             entries.add(StreamDialogEntry.enqueue);
@@ -373,6 +372,9 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I>
         if (KoreUtil.shouldShowPlayWithKodi(context, item.getServiceId())) {
             entries.add(StreamDialogEntry.play_with_kodi);
         }
+
+        entries.add(StreamDialogEntry.show_channel_details)
+
         StreamDialogEntry.setEnabledEntries(entries);
 
         new InfoItemDialog(activity, item, StreamDialogEntry.getCommands(context),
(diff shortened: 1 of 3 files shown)


Stypox:
Maybe try with `getParentFragmentManager()`

mhmdanas:
Unfortunately that doesn't fix the issue.

Stypox:
With `fragment.getActivity().getSupportFragmentManager()` it works (this is strange, though)

app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
```diff
@@ -350,7 +350,8 @@ protected void showStreamDialog(final StreamInfoItem item) {
             return;
         }
 
-        final ArrayList<StreamDialogEntry> entries = new ArrayList<>();
+        final ArrayList<StreamDialogEntry> entries =
+                new ArrayList<>(Arrays.asList(StreamDialogEntry.show_channel_details));
```
TobiGr, line 354: No idea, why the implementqtion was used here. We should go with the interface.
```suggestion
        final List<StreamDialogEntry> entries =
                new ArrayList<>(Arrays.asList(StreamDialogEntry.show_channel_details));
```

Stypox: And maybe also use `Collections.singletonList` instead of `Arrays.asList`

mhmdanas: @Stypox is there any reason in particular to do that?

Stypox: Arrays.asList creates a new List of fixed size, Collections.singletonList creates a custom List of size 1 where everything is optimized for that.

Stypox: Maybe, even better, just replace it with:
```suggestion
final List<StreamDialogEntry> entries =
                new ArrayList<>();
entries.add(StreamDialogEntry.show_channel_details);
```

Stypox: Oh, also, please put the open channel option at the bottom of the menu, it's less important than others

mhmdanas: @Stypox done.

SameenAhnaf:
Could you guys please work on these related issues too? Thanks in advance.

https://github.com/TeamNewPipe/NewPipe/issues/2171 , https://github.com/TeamNewPipe/NewPipe/issues/5836 ,
#5836 , #5783

mhmdanas:
@SameenAhnaf please keep the discussion on each PR related to the PR itself.

SameenAhnaf:
@mhmdanas If I am not wrong, aren't #2171 , #5836 , #5783 related to long press menu? Or, is a PR solely supposed to solve the title problem? Please forgive my unintentional mistakes. I am still learning GitHub.

mhmdanas:
@SameenAhnaf the discussion around a PR should generally focus only on the changes proposed by the PR.

> Please forgive my unintentional mistakes. I am still learning GitHub.

No problem.

Stypox:
@mhmdanas please fix [the issues](https://sonarcloud.io/project/issues?id=TeamNewPipe_NewPipe&pullRequest=5851&resolved=false&types=CODE_SMELL) pointed out by Sonar

mhmdanas:
I fixed the braces "code smell," but kept the other two because on is a TODO and the other is because I'm following the naming convention of the rest of the file.

Stypox:
Do you think this can be merged even though there is an unsolved TODO?

mhmdanas:
I searched quite a bit and failed to find any explanation. Can I just replace the TODO comment with something like "For some reason `getParentFragmentManager()` doesn't work, but this does"?

app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
```diff
@@ -372,6 +372,9 @@ protected void showStreamDialog(final StreamInfoItem item) {
         if (KoreUtil.shouldShowPlayWithKodi(context, item.getServiceId())) {
             entries.add(StreamDialogEntry.play_with_kodi);
         }
+
+        entries.add(StreamDialogEntry.show_channel_details);
```
Stypox, line 376: I realized this should be shown only if there is a channel attached to this item (this is the case for YouTube, but it may not be the case for other services. Only add the channel option if `!isNullOrEmpty(item.getUploaderUrl())`

Stypox (changes requested):
> Can I just replace the TODO comment with something like "For some reason getParentFragmentManager() doesn't work, but this does"?

@mhmdanas ok, that's fine

Stypox (approved):
Thanks :-)

opusforlife2:
This option doesn't appear in the Feeds (What's New or any other). Known issue? Postponed?

mhmdanas:
@opusforlife2 not known, at least to me.

The only way that would happen is if calling `getUploaderUrl` on the feed items returns `null`, which may sounds unlikely, but is the only possibility here. I'll investigate later.

opusforlife2:
> what service did you reproduce that on? 

I tried it with a feed with only YouTube channels in it. I haven't tried it with any other service.

mhmdanas:
The reason why this doesn't work with feeds is because the database doesn't store the feed items' uploader URLs. I don't think going back through all the feed items in feeds like "What's New" to add the uploader URL is a good idea. Maybe we could only add the URLs for new feed items? Anyway, I'll open an issue to continue the discussion there.

opusforlife2:
@mhmdanas Please ping me in that issue so I can follow it when you open one.

opusforlife2:
> Maybe we could only add the URLs for new feed items?

Yeah, adding them when the user initiates a feed refresh seems like a practical solution.

mhmdanas:
To be clear, I did not mean to add it for all previous feed items, but only for new ones (i.e. ones added in a feed refresh).

opusforlife2:
I got what you meant. We're on the same frequency. :3

Plugdemholes-Imeanpatchdembugs:
@mhmdanas 
Thanks for the implementation.
Usage has become far more frictionless than ever before since the implementation! :D

PR #5856, opened 2021-03-19
Update to ExoPlayer 2.13.2

Redirion:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [X] Codebase improvement (dev facing)

Updates ExoPlayer and fixes related deprecations where applicable.

@TacoTheDank I have also changed the setPlayWhenReady calls to play() and pause() where appropriate

#### APK testing 
the APK can be found by going to the "Checks" tab below the title and then on "artifacts" on the right.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -489,10 +489,7 @@ public final class Player implements
         simpleExoPlayer.addTextOutput(binding.subtitleView);
 
         // Setup audio session with onboard equalizer
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-            trackSelector.setParameters(trackSelector.buildUponParameters()
-                    .setTunnelingAudioSessionId(C.generateAudioSessionIdV21(context)));
-        }
+        trackSelector.setParameters(trackSelector.buildUponParameters().setTunnelingEnabled(true));
     }
 
     private void initListeners() {
@@ -624,10 +621,10 @@ public final class Player implements
                 && newQueue.getItem().getUrl().equals(playQueue.getItem().getUrl())
                 && newQueue.getItem().getRecoveryPosition() != PlayQueueItem.RECOVERY_UNSET) {
             // Player can have state = IDLE when playback is stopped or failed
-            // and we should retry() in this case
+            // and we should retry in this case
             if (simpleExoPlayer.getPlaybackState()
                     == com.google.android.exoplayer2.Player.STATE_IDLE) {
-                simpleExoPlayer.retry();
+                simpleExoPlayer.prepare();
             }
             simpleExoPlayer.seekTo(playQueue.getIndex(), newQueue.getItem().getRecoveryPosition());
             simpleExoPlayer.setPlayWhenReady(playWhenReady);
@@ -638,10 +635,10 @@ public final class Player implements
                 && !playQueue.isDisposed()) {
             // Do not re-init the same PlayQueue. Save time
             // Player can have state = IDLE when playback is stopped or failed
-            // and we should retry() in this case
+            // and we should retry in this case
             if (simpleExoPlayer.getPlaybackState()
                     == com.google.android.exoplayer2.Player.STATE_IDLE) {
-                simpleExoPlayer.retry();
+                simpleExoPlayer.prepare();
             }
             simpleExoPlayer.setPlayWhenReady(playWhenReady);
 
@@ -1649,7 +1646,7 @@ public final class Player implements
 
         saveWasPlaying();
         if (isPlaying()) {
-            simpleExoPlayer.setPlayWhenReady(false);
+            simpleExoPlayer.pause();
         }
 
         showControls(0);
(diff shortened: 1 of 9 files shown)


Redirion:
I have tested on my Pixel 4a (Android 11): basic video playback, play, pause and subtitles tested. Works so far, so any more feedback is appreciated.

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
```diff
@@ -43,13 +44,13 @@ default MediaSource buildLiveMediaSource(@NonNull final PlayerDataSource dataSou
         switch (type) {
             case C.TYPE_SS:
                 return dataSource.getLiveSsMediaSourceFactory().setTag(metadata)
-                        .createMediaSource(uri);
+                        .createMediaSource(new MediaItem.Builder().setUri(uri).build());
```
TacoTheDank, line 47: These MediaItem builders can be simplified to just `MediaItem.fromUri(uri)`

Redirion: you are right. I was copying that over from Exoplayer sources even though I should have known better, fixed. Thanks

TacoTheDank:
Overall very nice at a quick glance. Also, thanks for this; I was having trouble figuring out how to update ExoPlayer to 2.13 in another repository, so this will help a ton!

(You can ignore this little nitpick of mine, since it's not important at all really, but I feel that the changes concerning the ExoPlayer 2.13.x upgrade should be a separate commit from the changes that concern previous ExoPlayer updates (the MediaItem builders, setPlayWhenReady, setMediaSource, etc.)

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -1987,16 +1992,12 @@ public void changeState(final int state) {
         notifyPlaybackUpdateToListeners();
     }
 
-    private void onPrepared(final boolean playWhenReady) {
-        if (DEBUG) {
-            Log.d(TAG, "onPrepared() called with: playWhenReady = [" + playWhenReady + "]");
```
Stypox, line 1992: Do not remove this but keep "onPrepared() called"

Stypox (commented):
Looks good at a firsf glance, thank you!

app/src/main/java/org/schabi/newpipe/player/Player.java
Redirion: the log is instead to be found at onPlayWhenReadyChanged. ExoPlayer 2.13 splitted up the events in more methods, so I had to rearrange here

Redirion: "Deprecate EventListener.onPlayerStateChanged, replacing it with EventListener.onPlayWhenReadyChanged and EventListener.onPlaybackStateChanged."  

This was the change from the release notes (was actually already in 2.12.0).

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -489,10 +489,7 @@ private void initPlayer(final boolean playOnReady) {
         simpleExoPlayer.addTextOutput(binding.subtitleView);
 
         // Setup audio session with onboard equalizer
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-            trackSelector.setParameters(trackSelector.buildUponParameters()
-                    .setTunnelingAudioSessionId(C.generateAudioSessionIdV21(context)));
-        }
+        trackSelector.setParameters(trackSelector.buildUponParameters().setTunnelingEnabled(true));
```
TobiGr, line 492: Is that backwards compatible for NewPipe legacy (Android 4.1) or do we need to notify friendlyanon about this change?

Redirion: Yes, it's safe. In ExoPlayer source they have basically the same SDK version check before either applying generateAudioSessrionIdV21 or setting it to UNSET on lower Api

Redirion: https://github.com/google/ExoPlayer/blob/4364b9150f236813a2070fc79d3bcb76e11969ac/library/core/src/main/java/com/google/android/exoplayer2/SimpleExoPlayer.java#L662

And in the renderers there are also V21 methods regarding tunneling. 

From the release notes:
Simplify enabling tunneling with DefaultTrackSelector. ParametersBuilder.setTunnelingAudioSessionId has been replaced with ParametersBuilder.setTunnelingEnabled. The player's audio session ID will be used, and so a tunneling specific ID is no longer needed.

Stypox:
I tested a little bit: functionality-wise everything works fine, but the play-pause buttons always show the "Pause" icon even when the video is paused @Redirion

Redirion:
I have restored the "onPlayerStateChanged" listener. The splitting up into "onPlaybackStateChanged" and "onPlayWhenReadyChanged" was causing the issues. This further simplifies this PR and makes it easy to merge in my opinion.

Stypox (approved):
I tested, everything works fine. Thanks :-)

PR #5933, opened 2021-03-28
Add note about "Fixes" prefix to auto-link issues

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

#### Due diligence

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -13,8 +13,8 @@
 - take over the world
 
 #### Fixes the following issue(s)
-<!-- Also add any other links relevant to your change. -->
-- 
+<!-- Prefix issues with "Fixes" so that GitHub closes them when the PR is merged (note that each "Fixes #" should be in its own item). Also add any other relevant links. -->
+- Fixes #
 
 #### Relies on the following changes
 <!-- Delete this if it doesn't apply to you. -->


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -13,8 +13,8 @@
 - take over the world
 
 #### Fixes the following issue(s)
-<!-- Also add any other links relevant to your change. -->
-- 
+<!-- Prefix issues by "Fixes" so that GitHub links them. Also add any other relevant links. -->
+- Fixes #
```
mhmdanas, line 17:
```suggestion
<!-- Prefix issues with "Fixes" so that GitHub closes then when the PR is merged. Also add any other relevant links.

Example:
- Fixes #1
- Fixes #2
- example.com
-->
-
```

I'm not sure if we should have an item already filled with "Fixes #" or not to be honest.

mhmdanas (changes requested):
Minor question, and add an example.

.github/PULL_REQUEST_TEMPLATE.md
Stypox: I don't like this, it's too long to skim through each time and uses too much space. I made the comment as short as possible so that people tend to read it more. I think adding "Fixes #" is enough of an example, and it is also useful because one does not have to write it manually each time.

mhmdanas: The issue is that many people don't realize that "fixes" should only be followed by one issue, not multiple. I'll edit my suggestion in a minute to account for your notes.

Stypox: Ok, that looks good

mhmdanas (approved):

Stypox:
@TobiGr could you merge this if you think it's ok?

PR #5935, opened 2021-03-28
Fix hidden icon to the left of notification actions in settings

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Even though the `ConstraintLayout` code seemed right, for some reason the icon to left of the five notification actions under "Settings -> Notification" has width `0dp` and were thus invisible. I don't know when this happened, they were working when the PR about notifications was merged, but nevermind. This fixes it by setting a constant width instead of a constant height, and for some reason only this way `layout_constraintDimensionRatio="1"` does its job forcing the image into a square.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/res/layout/settings_notification_action.xml
+++ b/app/src/main/res/layout/settings_notification_action.xml
@@ -8,13 +8,13 @@
 
     <androidx.appcompat.widget.AppCompatImageView
         android:id="@+id/notificationActionIcon"
-        android:layout_width="0dp"
-        android:layout_height="48dp"
+        android:layout_width="48dp"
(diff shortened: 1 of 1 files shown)


mhmdanas:
Sorry, I was going to download the APK to test this but re-ran the workflow by mistake.

Stypox:
@mhmdanas I just re-ran all checks again and it seems to have worked

mhmdanas (approved):
I can confirm that this works; somebody else should take a look at the code changes.

TobiGr (approved):

PR #6037, opened 2021-04-10
Allow installation on external storage

### LINKED ISSUE #251: Allow install on external memory card
Can you set the flag to allow it to be installed on an external memory card?

mhmdanas:
#### What is it?
- [x] Feature (user facing)

Allow installation on external storage.

Unfortunately, I'm not in possession of an external SD card, so I can't test this. Those who have SD cards, could you please test this?

Note that NewPipe uses [`Service`s](https://developer.android.com/reference/android/app/Service), which Android's docs state as one of [the features that if present in an app, would mean that it shouldn't be allowed to be installed on external storage](https://developer.android.com/guide/topics/data/install-location). I'm not sure if this means we shouldn't merge this PR, but I figured it would at least be relevant.

- Fixes #251.

#### APK testing 

https://github.com/TeamNewPipe/NewPipe/suites/2463929014/artifacts/52944489

#### Due diligence

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    package="org.schabi.newpipe">
+    package="org.schabi.newpipe"
+    android:installLocation="auto">
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />


AlexN1997:
It works (LG G5 android 8.0)

**I noticed a bug:**


**how to reproduce:**
play a video 
change the audio

mhmdanas:
@AlexN1997 that doesn't seem like an issue caused by my changes. Can you try out [the latest development APK at the time](https://github.com/TeamNewPipe/NewPipe/actions/runs/735610463) and see if the bug is present there too?

AlexN1997:
Its still here, so as you said, its not this PR causing the problem

Redirion (approved):
lgtm

PR #6045, opened 2021-04-10
show overall duration of videos in playlist

bg172:
earlier only overall amount of videos was shown in playlist (in both searched online playlist and local playlist). Now overall duration is shown there too - as formatted by existing Localization.concatenateStrings() and Localization.getDurationString().

#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

earlier only overall amount of videos was shown. Now overall duration is shown there too - as formatted by existing Localization.concatenateStrings() and Localization.getDurationString().

#### APK testing 

see https://github.com/TeamNewPipe/NewPipe/pull/6045/checks (artefacts) for the uptodate APK as soon as built by github - tested OK on Xiaomi Mi 9 SE

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
@@ -89,6 +89,9 @@ public class PlaylistFragment extends BaseListInfoFragment<StreamInfoItem, Playl
 
     private MenuItem playlistBookmarkButton;
 
+    private long streamCount;
+    private long playlistOverallDurationSeconds;
+
     public static PlaylistFragment getInstance(final int serviceId, final String url,
                                                final String name) {
         final PlaylistFragment instance = new PlaylistFragment();
@@ -278,6 +281,12 @@ public class PlaylistFragment extends BaseListInfoFragment<StreamInfoItem, Playl
     }
 
     @Override
+    public void handleNextItems(final ListExtractor.InfoItemsPage result) {
+        super.handleNextItems(result);
+        setStreamCountAndOverallDuration(result.getItems(), !result.hasNextPage());
+    }
+
+    @Override
     public void handleResult(@NonNull final PlaylistInfo result) {
         super.handleResult(result);
 
@@ -322,8 +331,8 @@ public class PlaylistFragment extends BaseListInfoFragment<StreamInfoItem, Playl
                     .into(headerBinding.uploaderAvatarView);
         }
 
-        headerBinding.playlistStreamCount.setText(Localization
-                .localizeStreamCount(getContext(), result.getStreamCount()));
+        streamCount = result.getStreamCount();
+        setStreamCountAndOverallDuration(result.getRelatedItems(), !result.hasNextPage());
 
         final Description description = result.getDescription();
         if (description != null && description != Description.EMPTY_DESCRIPTION
@@ -486,4 +495,20 @@ public class PlaylistFragment extends BaseListInfoFragment<StreamInfoItem, Playl
         playlistBookmarkButton.setIcon(drawable);
         playlistBookmarkButton.setTitle(titleRes);
     }
+
+    private void setStreamCountAndOverallDuration(final List<StreamInfoItem> list,
+                                                  final boolean isDurationComplete) {
+        if (activity != null && headerBinding != null) {
+            playlistOverallDurationSeconds += list.stream()
+                    .mapToLong(x -> x.getDuration())
+                    .sum();
+            headerBinding.playlistStreamCount.setText(
+                Localization.concatenateStrings(
+                    Localization.localizeStreamCount(activity, streamCount),
+                    Localization.getDurationString(playlistOverallDurationSeconds,
+                            isDurationComplete))
+            );
+        }
+    }
+
 }
--- a/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
@@ -502,7 +502,7 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
                     }
 
                     final long videoCount = itemListAdapter.getItemsList().size();
-                    setVideoCount(videoCount);
+                    setStreamCountAndOverallDuration(itemListAdapter.getItemsList());
                     if (videoCount == 0) {
                         showEmptyState();
                     }
@@ -532,7 +532,7 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
             itemsList.getLayoutManager().onRestoreInstanceState(itemsListState);
             itemsListState = null;
         }
-        setVideoCount(itemListAdapter.getItemsList().size());
+        setStreamCountAndOverallDuration(itemListAdapter.getItemsList());
 
         PlayButtonHelper.initPlaylistControlClickListener(activity, playlistControlBinding, this);
 
@@ -665,7 +665,7 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
                 .subscribe(itemsToKeep -> {
                     itemListAdapter.clearStreamItemList();
                     itemListAdapter.addItems(itemsToKeep);
-                    setVideoCount(itemListAdapter.getItemsList().size());
+                    setStreamCountAndOverallDuration(itemListAdapter.getItemsList());
                     saveChanges();
 
                     hideLoading();
@@ -684,7 +684,7 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
             updateThumbnailUrl();
         }
 
-        setVideoCount(itemListAdapter.getItemsList().size());
+        setStreamCountAndOverallDuration(itemListAdapter.getItemsList());
         saveChanges();
     }
 
@@ -855,10 +855,20 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
         this.name = !TextUtils.isEmpty(title) ? title : "";
     }
 
-    private void setVideoCount(final long count) {
+    private void setStreamCountAndOverallDuration(final ArrayList<LocalItem> itemsList) {
         if (activity != null && headerBinding != null) {
-            headerBinding.playlistStreamCount.setText(Localization
-                    .localizeStreamCount(activity, count));
+            final long streamCount = itemsList.size();
+            final long playlistOverallDurationSeconds = itemsList.stream()
+                    .filter(PlaylistStreamEntry.class::isInstance)
+                    .map(PlaylistStreamEntry.class::cast)
+                    .map(PlaylistStreamEntry::getStreamEntity)
+                    .mapToLong(StreamEntity::getDuration)
+                    .sum();
+            headerBinding.playlistStreamCount.setText(
+                    Localization.concatenateStrings(
+                            Localization.localizeStreamCount(activity, streamCount),
+                            Localization.getDurationString(playlistOverallDurationSeconds))
+            );
         }
     }
 
--- a/app/src/main/java/org/schabi/newpipe/util/Localization.java
+++ b/app/src/main/java/org/schabi/newpipe/util/Localization.java
@@ -238,7 +238,25 @@ public final class Localization {
         }
     }
 
+    /**
+     * Get a readable text for a duration in the format {@code days:hours:minutes:seconds}.
+     * Prepended zeros are removed.
+     * @param duration the duration in seconds
+     * @return a formatted duration String or {@code 0:00} if the duration is zero.
+     */
     public static String getDurationString(final long duration) {
+        return getDurationString(duration, true);
+    }
+
+    /**
+     * Get a readable text for a duration in the format {@code days:hours:minutes:seconds+}.
+     * Prepended zeros are removed. If the given duration is incomplete, a plus is appended to the
+     * duration string.
+     * @param duration the duration in seconds
+     * @param isDurationComplete whether the given duration is complete or whether info is missing
+     * @return a formatted duration String or {@code 0:00} if the duration is zero.
+     */
+    public static String getDurationString(final long duration, final boolean isDurationComplete) {
         final String output;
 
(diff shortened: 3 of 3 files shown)


XiangRongLin (dismissed):
This won't work like this. The remote playlist are paginated, meaning you only ever load a fraction of longer playlist. The playlist also can't be preloaded, because it could potentially be an infinite amount of videos (YouTube Mix)

It also looks like as if the change will break the existing counter for long playlist, for the same reason.

XiangRongLin:
This would need to be done in the extractor, in order to be done properly. And it would only work if the services provide that specific information for us to parse

bg172:
Hi @XiangRongLin, thank you very much for reviewing. Now reworked to show overall duration only for local lists - they are not so tricky I hope. Can you give an example-link of an infinite list by the way?

XiangRongLin:
> Can you give an example-link of an infinite list by the way?

https://www.youtube.com/watch?v=FfAjtZgLkPQ&list=RDFfAjtZgLkPQ

Its basically youtube.com/watch?v={videoID}&list=RD{videoID}

bg172:
so now improved enough, tried with that infinite list. Next thing would be to extend the service to deliver the duration, but this service is not in our hands, right?:)

XiangRongLin:
The service is in our hands https://github.com/TeamNewPipe/NewPipeExtractor

And like i said before, it needs to be fixed there. From a quick glance the duration would increase as more items are loaded.
Imagine you are a user and experience that. Would you not be confused?

bg172:
I am a user, and i am already satisfied with this - i did this implementation for me in the first line;)
Most of playlists i see are under 100 items and this seems to be the threshold of the service currently to provide items, i.e. in many cases the duration will be correct, and the longer it is the less important to know its value exactly;)
To not confuse users there is this "+" in the end of duration string like 12:20:23+ until exact entire duration is known, meaning it is longer than 12h20m23s. 
With your info that the service is editable, i consider my PR as kind of an incremental improvement, but not sure when i can make the next incremental step.
What do you thing?

mhmdanas:
I think this could be made as an option disabled by default, with a note that for larger playlists, it might not display the actual duration immediately.

bg172:
now made it configurable

bg172:
> This would need to be done in the extractor, in order to be done properly. And it would only work if the services provide that specific information for us to parse

If I do this via the extractor I have to return the overall duration in _PlaylistInfo_ which is called only once - when opening the view for a playlist. Because the online-service (e.g. youtube) does not provide playlist-duration it is needed to loop through all items, sum up the duration and provide it via something like _PlaylistInfo.getStreamDuration()_ of extractor - for multipage playlists (>100 videos) it is needed to load all pages which takes too long (about 12s to show a playlist with 1615 videos I had tried - all this time the list itself appears to be loading and no items of it is shown to the user) which makes user experience too bad. 
The alternative which I implemented - summing up only for the items which are currently shown and considering additional items incrementally as soon as they are added - does not fit current design of extractor usage: the callback adding next page items _handleNextItems(final ListExtractor.InfoItemsPage result)_ cannot read the original _PlaylistInfo_ which would have new getter _getStreamDuration()_ i.e. does not allow to update the duration incrementally. 

So either we implement outside of extractor showing only partial duration in case of long playlist but having good performance, or we implement inside of extractor but kill the performance and show the exact duration which nobody needs in long playlists:)

What do you think, am I missing something?

XiangRongLin:
That is unfortunate that youtube does not provide that information. Since the services (youtube) themself do not provide that information, it would then also not belong inside the extractor.

> So either we implement outside of extractor showing only partial duration in case of long playlist but having good performance

I don't personally see much use in this solution, but I also don't like to decide about which features to add and which not. These discussions are best made in an issue and not the PR itself.
@Stypox your thoughts on this, i'm slightly against this.

SameenAhnaf:
I wish to see this overall video duration even without opening the playlist. This info could be written on playlist details or under totap video number.
IMG_20210514_174554

Not a request, just a suggestion that overall video duration could be shown for channels too.

>So either we implement outside of extractor showing only partial duration in case of long playlist but having good performance.

How about if we show overall duration of first 50 or 100 videos? Maybe, "12:08:09 (∞)" info could be written. Infinite has no real unit. So, I guess it's needless to put a unit here.

bg172:
"without opening the playlist" is a good point, let me try. 

Concerning units: I made it already without units, it looks like this now: 01:35:17 in case complete or in case incomplete 12:08:09+, 
but 12:08:09 (∞) is also ok if you confirm, @SameenAhnaf.

About "first 50 or 100 videos": it works already like this out-of-the-box just because this amount of videos the extrator delivers.

SameenAhnaf:
@bg172 Extremely sorry, I skimmed through the PR. So, I didn't notice everything. Thought of 12:08:09+ didn't come to my mind. But I seem to like it.

I am not sure which unit you are referring to. Pardon me my mistakes. As a typical user, I am not sure if you made changes into commits. I am kinda mixing this up with the currently provided apk.
IMG_20210514_201813

>"without opening the playlist" is a good point, let me try.

Thank you so much for your valuable contribution. You don't know how much you helped me! If possible, please try this for channels too.
IMG_20210514_202558
IMG_20210514_202620

bg172:
hi @SameenAhnaf , 
I checked about "without opening the playlist", and unfortunately cannot do because e.g. youtube provides info/duration of only 2 first videos on the search-results page with playlists - showing overall duration of just two first videos appears meaningless to me. Only when one opens the playlist, a meaningful amount of info/durations of videos becomes available. Might be possible with a larger redesign perhaps by sending more requests than the user directly triggers, but I cannot do it, sorry.


Showing duration for a channel (when one is open on the screen) I would consider as a new PullRequest, and seems to be possible as I understand the corresponding youtube page for that. But on my phone-screen there seems to be no space where to add duration - even the number of videos is not shown, see 
Screenshot 2021-05-15 023623channel
Were your screenshots done on some larger device like a tablet? Please suggest.


And some good news: on your screenshot with infinite number of videos no duration is shown I bet because it was not enabled in Settings as requested by @mhmdanas:), I now updated the link to the latest APK, sorry for confusion, can you please check again.
Screenshot 2021-05-15 020735settings
Screenshot 2021-05-15 021000infList

Cool that you've found my feature useful, looking forward to hear from you.

SameenAhnaf:
>I checked about "without opening the playlist", and unfortunately cannot do because e.g. youtube provides info/duration of only 2 first videos on the search-results page with playlists - showing overall duration of just two first videos appears meaningless to me. Only when one opens the playlist, a meaningful amount of info/durations of videos becomes available. Might be possible with a larger redesign perhaps by sending more requests than the user directly triggers, but I cannot do it, sorry.

>Showing duration for a channel (when one is open on the screen) I would consider as a new PullRequest, and seems to be possible as I understand the corresponding youtube page for that.

That's okay. Thanks for the valuable information!

>But on my phone-screen there seems to be no space where to add duration - even the number of videos is not shown, see

My screenshots were done on a 6.5" phone. Probably, my fonts are way too small😅.

It seems like that we will have to keep the information beside the channel name as you can see for grid view.
IMG_20210515_105019

>And some good news: on your screenshot with infinite number of videos no duration is shown I bet because it was not enabled in Settings as requested by @mhmdanas:), 

Thanks once again. But I wonder why this setting should be optional. 
Maybe, a setting like "Show duration of next unwatched videos" with options "None/All (Watched+Unwatched)/5/10/25/50/100 videos" could be added. This setting is required as none of us won't always have the time to stream hundreds of videos at a sitting.

If "All" is selected for an infinite playlist and duration of first 100 videos is 19:09:08, duration could be shown as "19:09:08+". If a certain value is set (5/10/25/50/100), no '+' icon needs to be shown for infinite playlists. If certain value (5/10/25/50/100) is set for this setting, "Watch history" is disabled, duration of first (5/10/25/50/100) videos will be shown.

I kinda like now "∞ videos" info. The behaviour will remain constant for all types of playlists. Plus, applying certain value for my mentioned setting won't be affected.

>Cool that you've found my feature useful, looking forward to hear from you.

As a student, it is a necessity for me in this time of pandemic. Go ahead with your brilliant ideas ☺️.

bg172:
> It seems like that we will have to keep the information beside the channel name as you can see for grid view.
> IMG_20210515_105019
This screenshot is a showing search-results page with playlists from youtube, for which youtube does not provide durations - not sure if possible to implement without youtube support but can you create a new feature request?


> Thanks once again. But I wonder why this setting should be optional.
> Maybe, a setting like "Show duration of next unwatched videos" with options "None/All (Watched+Unwatched)/5/10/25/50/100 videos" could be added. This setting is required as none of us won't always have the time to stream hundreds of videos at a sitting.
> 
> If "All" is selected for an infinite playlist and duration of first 100 videos is 19:09:08, duration could be shown as "19:09:08+". If a certain value is set (5/10/25/50/100), no '+' icon needs to be shown for infinite playlists. If certain value (5/10/25/50/100) is set for this setting, "Watch history" is disabled, duration of first (5/10/25/50/100) videos will be shown.
Yeah, the dependency on watch-history would be a new advanced feature building upon this PR. And it would depend on the sort-by criteria too, which I was dreaming of as next implementation. Until that this setting is too much, therefore I will remove it now.

SameenAhnaf:
>This screenshot is a showing search-results page with playlists from youtube, for which youtube does not provide durations - not sure if possible to implement without youtube support but can you create a new feature request?

No problem, I will open an issue as early as possible ☺️.

Much obliged for your efforts.

Akv2021:
Any chances of merging this feature ?
I think NewPipe can provide an "experimental features" section which enables features like above which might not work for all users but are useful for the smaller audience nonetheless.

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -78,6 +78,9 @@ public class PlaylistFragment extends BaseListInfoFragment<PlaylistInfo> {
 
     private MenuItem playlistBookmarkButton;
 
+    private long videoCount;
```
Stypox, line 81: Consistency
```suggestion
    private long streamCount;
```

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -791,10 +791,18 @@ private void setInitialData(final long pid, final String title) {
         this.name = !TextUtils.isEmpty(title) ? title : "";
     }
 
-    private void setVideoCount(final long count) {
+    private void setVideoCountAndOverallDuration(final ArrayList<LocalItem> itemsList) {
         if (activity != null && headerBinding != null) {
-            headerBinding.playlistStreamCount.setText(Localization
-                    .localizeStreamCount(activity, count));
+            final long videoCount = itemsList.size();
+            final long playlistOverallDurationSeconds = itemsList.stream().mapToLong(x ->
```
Stypox, line 797: I would first filter by `PlaylistStreamEntry.class::isInstance`, then map with `PlaylistStreamEntry.class::cast`, then map to `PlaylistStreamEntry::getStreamEntity`, finally mapToLong `StreamEntryOrWhatever::getDuration` and sum.

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -243,6 +243,10 @@ private static String getQuantity(final Context context, @PluralsRes final int p
     }
 
     public static String getDurationString(final long duration) {
+        return getDurationString(duration, true);
+    }
+
+    public static String getDurationString(final long duration, final boolean isDurationComplete) {
```
Stypox, line 249: Maybe add some javadoc to these methods ;-)

Stypox (changes requested):
Thank you for this nice addition. Code looks almost good, and is also really simple. I think this PR should be merged, but we should not proceed further imo (e.g. by implementing it in the extractor), because we would soon get into niche territory.

bg172:
Hi @Stypox, @Akv2021, thank you for your comments and attention to this pull-request! 
Meanwhile I detected there is already a similar feature in NewPipe available (v0.23.1), see videos attached - e.g. in a found list or a favorite list you can see the full duration of the list and remaining list-time over the list if you switch to a next video. Let me know what do you think:)

https://user-images.githubusercontent.com/6170902/179813800-8f8f950e-1e82-4521-a7b3-759f0bf6353a.mp4


https://user-images.githubusercontent.com/6170902/179813828-aae7defd-b6d3-4cb6-8cf4-96d91a465b14.mp4

Stypox:
I think this PR would be useful anyway, even if a similar feature exists in another place

app/src/main/res/values/strings.xml
```diff
@@ -790,4 +790,7 @@
     <string name="audio_track_type_original">original</string>
     <string name="audio_track_type_dubbed">dubbed</string>
     <string name="audio_track_type_descriptive">descriptive</string>
+    <string name="show_playlist_duration_key" translatable="false">show_playlist_duration</string>
+    <string name="show_playlist_duration_summary" translatable="false">Show overall duration in playlists, for larger playlists it might not display the actual duration</string>
+    <string name="show_playlist_duration_title" translatable="false">Show playlist duration</string>
```
TobiGr, line 795: We'll need to move the key into settings_keys.xml.
```suggestion
    <string name="show_playlist_duration_title">Show playlist duration</string>
    <string name="show_playlist_duration_summary">Show combined duration of all streams in playlists. The display duration might be too small if the playlist has many items.</string>
```

Stypox: There is no setting anymore so I don't know why this is still here ;-)

Stypox (approved):
Thank you to everyone involved! I rebased and squashed, and removed some leftover strings that were used in the settings.

bg172:
> It feels like the current UI is a bit non-intuitive. Maybe there should be a prefix before it, like "Total duration: 45:03" or something.

A nice suggestion. I don't know how to update this pull-request here, but I have an idea how to fulfill this request - see https://github.com/TeamNewPipe/NewPipe/compare/dev...bg172:NewPipe:showOverallDurationInPlaylist . Maybe somebody can take it over in a meaningful way?

Stypox:
Yes, this should go in a new PR, but I guess it will be a quite simple change, so it can be merged in 0.27.0

bg172:
> Yes, this should go in a new PR, but I guess it will be a quite simple change, so it can be merged in 0.27.0

https://github.com/TeamNewPipe/NewPipe/pull/10952

PR #6051, opened 2021-04-11
[Bandcamp] Add Radio stream segments

TobiGr:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Display channel / creator / artist name in stream segment list when available.
| YouTube (no channel name) | Bandcamp Radio (artist name) |
| ------------------------------------- | ------------------------------------------ |
|  |  |

- TeamNewPipe/NewPipeExtractor#607

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/StreamSegmentItem.kt
+++ b/app/src/main/java/org/schabi/newpipe/info_list/StreamSegmentItem.kt
@@ -1,5 +1,6 @@
 package org.schabi.newpipe.info_list
 
+import android.view.View
 import android.widget.ImageView
 import android.widget.TextView
 import com.nostra13.universalimageloader.core.ImageLoader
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/item_stream_segment.xml
```diff
@@ -48,6 +48,18 @@
             app:layout_constraintTop_toTopOf="parent"
             tools:text="Lorem ipusum is widely used to create long sample text which is used here too" />
```
fynngodau, line 49: This now looks like this in the preview:

image

To fix the rendering, you could add `tools:maxLines="1"` here.

app/src/main/res/layout/item_stream_segment.xml
```diff
@@ -48,6 +48,18 @@
...
+        <TextView
+            android:id="@+id/textViewChannel"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:ellipsize="end"
+            android:maxLines="1"
+            android:textAppearance="@style/PlayQueueItemSubtitle"
+            android:textSize="@dimen/video_item_search_title_text_size"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent"
```
fynngodau, line 60: I noticed the other view have this too. However the parent view is not a `ConstraintLayout` but a `LinearLayout`.

app/src/main/java/org/schabi/newpipe/info_list/StreamSegmentItem.kt
```diff
@@ -29,6 +30,14 @@ class StreamSegmentItem(
             )
         }
         viewHolder.root.findViewById<TextView>(R.id.textViewTitle).text = item.title
+        if (item.channelName == null) {
+            viewHolder.root.findViewById<TextView>(R.id.textViewChannel).visibility = View.GONE
```
fynngodau, line 34: It could happen that only some items in the list set a channel name, therefore the visibility should be set to `View.VISIBLE` in the `else` block.

fynngodau (approved):
I tested this one, too, and it works well. Please see my comments; other than that it looks good.

app/src/main/res/layout/item_stream_segment.xml
TobiGr: Good catch. I did not notice that while copy-pasting the text view.

fynngodau (approved):
My review has been addressed :+1:

opusforlife2:
Will the heading still be "Chapters" on Bandcamp? What about something like "Streams", or "Tracks"?

TobiGr:
@opusforlife2 I don't have time for that. Maybe someone else can do that. Please open an issue for that.

Issue #5967, opened 2021-03-31
Return to portrait after video finish in fullscreen


Frivolous2:
### Checklist

- [x] I am using the latest version - 0.21.0 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug
1. Use NewPipe in portrait mode with locked screen rotate
2. Launch a video and press the fullscreen button. Let it finish.

### Actual behaviour

The video exits from fullacreen and NewPipe is stuck in landscape. You have to enable autorotate to fix it.

### Expected behavior

NewPipe returns to portrait, the mode I launched the video from.

### Device info

 - Android version/Custom ROM version: Lineage 17
 - Device model: Galaxy A3(2017)

sauravrao637:
Hi!
Thanks for reporting the issue and keep doing such good things :)
but it's no bug
I checked this out , if you have auto rotate enabled then the screen will not get stuck but if it's disabled then it's not bug it is the expected behavior.
Please check again on your device and tell if it's not the case

Frivolous2:
I know about auto rotate working. I was just shocked this is "the expected behaviour". I start fullscreen from portrait, why would I want to go landscape when I exit fullscreen?

sauravrao637:
If you click again on full screen button , it will exit the full screen . Also you can check its the same case with the official YouTube app

Frivolous2:
I'm not sure if I'm doing it right, but in official YT, when a video is over, it doesn't exit fullscreen. But pressing any button to exit it returna it to portrait. It doesn't force me to use landscape after a video is over though.

sauravrao637:
Nor does NewPipe force you :), if you tap on "full screen icon" you will successfully return to portrait.
When the rotation in locked but you choose fullscreen that means you want to use it in landscape now (in my opinion).
Maybe the team should create a poll regarding the same as all users are different and we should know opinion of the majority and work accordingly :)

lokesh-krishna:
Glad to see this has already been reported and I'm in agreement. This should definitely not be the expected behaviour. Orientation should go back to portrait upon exiting full screen. Not only that, the UI also seems to get wonky after exiting full screen. Attaching screenshots to explain.

[...]

PR #6059, opened 2021-04-12
Adds manual tablet mode setting

### LINKED ISSUE #6044: In Split Screen maximize video
### Checklist

When in split screen mode, the video should be maximized to fill the area, the way the official Youtube app does it.
Screenshot_20210410-231531

At least, when the full screen button is pressed, this should be the result. Currently, the full screen button does nothing in split screen mode.

When I use NewPipe in split screen mode, the app tries to squeeze the full view into the tiny split screen view, instead of displaying only the video. This makes it very tiny.
Screenshot_20210410-231448

This feature would allow multitasking while watching videos. Currently, I am stuck with a tiny video and half the screen wasted.

Dakkaron:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Adds a setting to manually force tablet mode to be on or off under Appearance settings
- Works as a work-around for tablet-mode related issues like #5967
- Also helps all users who increase the minimum width to fit more stuff on their screen, but prefer the better-fitting phone mode for NewPipe

- Fixes #6044 

#### APK testing 

https://github.com/TeamNewPipe/NewPipe/suites/2474591406/artifacts/53233850

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
@@ -12,8 +12,10 @@ import android.view.KeyEvent;
 import androidx.annotation.Dimension;
 import androidx.annotation.NonNull;
 import androidx.core.content.ContextCompat;
+import androidx.preference.PreferenceManager;
 
 import org.schabi.newpipe.App;
+import org.schabi.newpipe.R;
 
 public final class DeviceUtils {
 
@@ -65,10 +67,18 @@ public final class DeviceUtils {
     }
 
     public static boolean isTablet(@NonNull final Context context) {
-        return (context
-                .getResources()
-                .getConfiguration().screenLayout & Configuration.SCREENLAYOUT_SIZE_MASK)
-                >= Configuration.SCREENLAYOUT_SIZE_LARGE;
+        final String tabletModeSetting = PreferenceManager.getDefaultSharedPreferences(context)
+                .getString(context.getString(R.string.tablet_mode_key), "");
+
+        if (tabletModeSetting.equals(context.getString(R.string.tablet_mode_on_key))) {
+            return true;
+        } else if (tabletModeSetting.equals(context.getString(R.string.tablet_mode_off_key))) {
+            return false;
+        }
+
+        // else automatically determine whether we are in a tablet or not
+        return (context.getResources().getConfiguration().screenLayout
+                & Configuration.SCREENLAYOUT_SIZE_MASK) >= Configuration.SCREENLAYOUT_SIZE_LARGE;
     }
 
     public static boolean isConfirmKey(final int keyCode) {
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -742,5 +742,8 @@
     <string name="metadata_privacy_unlisted">Unlisted</string>
     <string name="metadata_privacy_private">Private</string>
     <string name="metadata_privacy_internal">Internal</string>
-    <string name="open_website_license">Open Website</string>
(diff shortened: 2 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -714,4 +714,7 @@
     <string name="night_theme_summary">Select your favorite night theme — %s</string>
     <string name="select_night_theme_toast">You can select your favorite night theme below</string>
     <string name="download_has_started">Download has started</string>
+    <string name="tablet_mode_title">Tablet Mode</string>
+    <string name="on">On</string>
+    <string name="off">Off</string>
```
mhmdanas, line 719: These are translatable, which we almost definitely don't want them to be.

mhmdanas (changes requested):

app/src/main/res/values/strings.xml
mhmdanas: Actually, ignore that. I thought you were using them in the Java code, but it turns out you're not.

Dakkaron:
Are you sure? These are user-facing strings. For the non-userfacing key IDs (which should not be translatable and are marked as non-translatable) there are `@string/tablet_mode_auto_key`, `@string/tablet_mode_on_key` and `@string/tablet_mode_off_key`.

app/src/main/res/values/strings.xml
Dakkaron: Are you sure? These are user-facing strings. For the non-userfacing key IDs (which should not be translatable and are marked as non-translatable) there are `@string/tablet_mode_auto_key`, `@string/tablet_mode_on_key` and `@string/tablet_mode_off_key`.

mhmdanas: No, I was mistaken. Ignore my comment.

Dakkaron: Sorry, Github doesn't auto-update, so I didn't see your second comment before posting. So please disregard my comment as well.

Dakkaron:
Is there anything else I need to do to get this PR merged?

mhmdanas:
@Dakkaron someone from the team who is actually familiar with Android apps would have to review it and approve it. I'll request reviews from a couple people now.

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -55,6 +57,16 @@ public static boolean isTv(final Context context) {
     }
 
     public static boolean isTablet(@NonNull final Context context) {
+        final String tabletModeKey = context.getString(R.string.tablet_mode_key);
+        final String tabletModeSetting = PreferenceManager.getDefaultSharedPreferences(context)
+                                                    .getString(tabletModeKey, "");
```
TiA4f8R, line 62: I am not an Android developer or not very familiar with Android apps, so I cannot review the code itself but I think if the indentation isn't correct on these lines. Here is what I got when I paste the code in Android Studio:
```suggestion

        final String tabletModeSetting = PreferenceManager.getDefaultSharedPreferences(context)
                .getString(tabletModeKey, "");
```

Dakkaron: This kind of indentation is pretty much preference. There are coding styles that say to indent the one way or the other. From what I could find, the contribution guidelines only said to stick with checkstyle, which hasn't flagged this.

But either way, I don't really care. So if that's the preferred way, I can change it like that.

mhmdanas (dismissed):
LGTM. Waiting on other reviewer(s).

Stypox (approved):
Great, thank you! I added a comment before merging, just so that auto mode is clearer in code.

PR #6093, opened 2021-04-16
minor code improvement in DownloadActivity

sauravrao637:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Replaced switch using resourceID to if  statement as [resourceID will be non final soon](https://stackoverflow.com/questions/64335374/how-to-resolve-resource-ids-will-be-non-final-in-android-gradle-plugin-version) 

- Fixes future issues :)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/download/DownloadActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/download/DownloadActivity.java
@@ -86,12 +86,11 @@ public class DownloadActivity extends AppCompatActivity {
 
     @Override
     public boolean onOptionsItemSelected(final MenuItem item) {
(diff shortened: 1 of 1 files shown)


XiangRongLin:
I would prefer the old variant with switch case, since its easier to add more cases to it if needed

sauravrao637:
Hi @XiangRongLin 
I totally agree with you but I had to replace it because from upcoming gradle versions Reource IDs will be non final and thus be avoided in switch case .
Please have a look at all the answers [here](https://stackoverflow.com/questions/64335374/how-to-resolve-resource-ids-will-be-non-final-in-android-gradle-plugin-version) for better understanding. :)

PR #6094, opened 2021-04-16
Code improvement (replacing switch case  using resourceID with if/else)

sauravrao637:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

Replaced switch case with if/else because [resource_id will be non final soon](https://stackoverflow.com/questions/64335374/how-to-resolve-resource-ids-will-be-non-final-in-android-gradle-plugin-version)

- Fixes future issues ;)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/download/DownloadActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/download/DownloadActivity.java
@@ -86,12 +86,11 @@ public class DownloadActivity extends AppCompatActivity {
 
     @Override
     public boolean onOptionsItemSelected(final MenuItem item) {
-        switch (item.getItemId()) {
-            case android.R.id.home:
-                onBackPressed();
-                return true;
-            default:
-                return super.onOptionsItemSelected(item);
+        if(item.getItemId() == android.R.id.home){
+            onBackPressed();
+            return true;
         }
+
+        return super.onOptionsItemSelected(item);
     }
 }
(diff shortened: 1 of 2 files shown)


Redirion:
Can you please explain what this improves?

From my understanding I would say replacing if else cascade with switch is a code improvement, not the other way around. A switch case is also internally lesser burden on CPU as it internally relies on a map, which is easier to track for schedule optimizations than conditional code branching.

sauravrao637:
Hi @Redirion 
Please have a look at the answers in the link attached in this PR ("resourceId will be non final soon").

Redirion:
http://tools.android.com/tips/non-constant-fields

This is only affecting library projects. NewPipe is not a library project.

sauravrao637:
I didn't know it will just be affecting library projects
but still shouldn't the warnings be addressed as closely as possible?

Redirion:
Fixing a false warning does not sound useful to me. But we could disable the warning: https://stackoverflow.com/a/64377108

sauravrao637:
But like in ErrorActivity (line 192) also if/else has been used instead of switch .

Redirion:
> 
> 
> But like in ErrorActivity (line 192) also if/else has been used instead of switch .

this could be improved by a switch case then. Would you like to create a PR with disabling that false warning and improving if/else to switch case?

sauravrao637:
Sure , I'll do it :) @Redirion

Issue #4148, opened 2020-08-17
Amoled support


blackCauldron7:
#### Describe the feature you want

Full Pitch black UI support for amoled devices to save battery. Currently the black theme still leaves a lot of UI elements colored like the top bar.
#### Is your feature request related to a problem? Please describe it

Not exactly but yes high battery consumption.
#### How will you/everyone benefit from this feature?

The extra battery life would be worth for every amoled user.

opusforlife2:
Test the apk in #3632 and reply if it doesn't match your expectations.

blackCauldron7:
#3532 doesn't match my expectations @opusforlife2 as it still leaves things like the top bar colored.

opusforlife2:
Thanks!

B0pol:
> The extra battery life would be worth for every amoled user.

[The "extra battery life" is **really really** small](https://www.xda-developers.com/amoled-black-vs-gray-dark-mode/)

blackCauldron7:
@B0pol it piles up as newer phones screen gets bigger and bigger and people tend to use newpipe for quite a good chunk of time so pretty sure it would be worth it.

mhmdanas:
I would expect most people to spend most of their time in NewPipe watching stuff, not navigating the app, so even if you disregard @B0pol's comment, the gains are probably very negligible.

andrewbanchich:
I'd love this just for the aesthetic.

blackCauldron7:
@mhmdanas tbh for me its just worth it for the look, all my other apps are all black. And i usually watch videos while reading the comments so theres that

mhmdanas:
I am not against adding this; I'm just saying that the battery life gains from this change are extremely negligible.

k9janer:
I would also appreciate a completely black title bar for night time use and to protect OLED and Plasma tvs.

Also I think NewPipe looks really sleek in black 😁
![nsmail-1_crop_crop](https://user-images.githubusercontent.com/26097284/121231299-8957a380-c890-11eb-82cf-9e1d5d4b4560.jpg)

SameenAhnaf:
@litetex This should be closed in favor of https://github.com/TeamNewPipe/NewPipe/issues/6272#issuecomment-856608962

litetex:
Closing this issue.

Why?
* https://github.com/TeamNewPipe/NewPipe/issues/6272#issuecomment-856608962
[...]

PR #6244, opened 2021-05-06
changed dark theme colors to darker variant

### LINKED ISSUE #6203: Whe using dark mode, change the color of red elements so they have a dark pale shade of red
When using the dark mode, make red colored elements to have a dark pale shade of red. 

The current red, that is very vivid, causes eye strain.

#### Additional context
imgonline-com-ua-twotoone-Iku5fRLxPNu

Left: purposed design¹ | Right: currently

¹In the phone, it's a lot more vibrant than it is when viewing through a PC, so this mockup may look weird on some phones '-'

To reduce eye strain, the accent have to be dark and pale. Putting a red vivid color in dark mode is not aligned with the purpose of this mode.

sauravrao637:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] UI improvement

improved dark_youtube_primary_color and dark_youtube_dark_color

- Fixes #6203 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -25,7 +25,7 @@
     <!-- Dark Theme -->
     <color name="dark_background_color">#222222</color>
     <color name="dark_dialog_background_color">#424242</color>
-    <color name="dark_settings_accent_color">#ff5252</color>
+    <color name="dark_settings_accent_color">#992722</color>
     <color name="dark_separator_color">#0affffff</color>
     <color name="dark_ripple_color">#48ffffff</color>
     <color name="dark_selected_color">#2affffff</color>
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -25,7 +25,7 @@
     <!-- Dark Theme -->
     <color name="dark_background_color">#222222</color>
     <color name="dark_dialog_background_color">#424242</color>
-    <color name="dark_settings_accent_color">#992722</color>
+    <color name="dark_settings_accent_color">#ff5252</color>
     <color name="dark_separator_color">#0affffff</color>
     <color name="dark_ripple_color">#48ffffff</color>
     <color name="dark_selected_color">#2affffff</color>
--- a/app/src/main/res/values/colors_services.xml
+++ b/app/src/main/res/values/colors_services.xml
@@ -6,8 +6,8 @@
     <color name="light_youtube_accent_color">#000000</color>
     <color name="light_youtube_statusbar_color">#e53935</color>
 
-    <color name="dark_youtube_primary_color">#CD322E</color>
-    <color name="dark_youtube_dark_color">#992722</color>
+    <color name="dark_youtube_primary_color">#992722</color>
+    <color name="dark_youtube_dark_color">#7a1717</color>
     <color name="dark_youtube_accent_color">#FFFFFF</color>
     <color name="dark_youtube_statusbar_color">#CD322E</color>
 
--- a/app/src/main/res/values/colors_services.xml
+++ b/app/src/main/res/values/colors_services.xml
@@ -9,7 +9,7 @@
     <color name="dark_youtube_primary_color">#992722</color>
     <color name="dark_youtube_dark_color">#7a1717</color>
     <color name="dark_youtube_accent_color">#FFFFFF</color>
-    <color name="dark_youtube_statusbar_color">#CD322E</color>
+    <color name="dark_youtube_statusbar_color">#992722</color>
 
     <!-- SoundCloud -->
     <color name="light_soundcloud_primary_color">#f57c00</color>
--- a/app/src/main/res/values/colors_services.xml
+++ b/app/src/main/res/values/colors_services.xml
@@ -17,10 +17,10 @@
     <color name="light_soundcloud_accent_color">#000000</color>
     <color name="light_soundcloud_statusbar_color">#f57c00</color>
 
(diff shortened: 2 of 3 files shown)


iamthesenate1:
Hi! I installed it and I think it looks good but I noticed a mismatch betwen the titlebar color and the color of the notification bar (it is the old, brighter one).
Screenshot_2021-05-06-23-06-09-243_org schabi newpipe debug

iamthesenate1:
Also, in the settings menu (almost) the same thing happens, but in reverse.
Screenshot_2021-05-06-23-11-04-901_org schabi newpipe debug

peat80:
This should probably be done with the other services colours too and then included as a separate theme to choose in settings. 🤔

iamthesenate1:
It works better now but in the settigs menu there is still a mismatch. Is this intended?
Screenshot_2021-05-07-01-45-47-387_org schabi newpipe debug

iamthesenate1:
> This should probably be done with the other services colours too and then included as a separate theme to choose in settings. 🤔

I agree

blackandyellowgecko:
Great! :D

Looks way better, congrats! The status bar still show the vivid red color (so does the splash screen):


The statusbar background color could be black, with opacity.


Another possible solution would be a transparent top panel, following Material Design 2 guidelines (for example, YouTube and Play Store ditched the red and the green background in the top panel).

sauravrao637:
I am closing this PR as #6272 will provide better solution

SameenAhnaf:
#6244 addresses customisation for every single thing including top bar colour. See castrik's comment: https://github.com/TeamNewPipe/NewPipe/issues/6272#issuecomment-841838765

That's why, this issue should stay open instead as it is expected to solve a very specific problem.

Stypox:
No @sauravrao637, this is a good idea, while #6272 is probably not doable at all.

The color of Android's top bar should be made the same as the toolbar.

sauravrao637:
@Stypox 
Okay I'll do it when I get some time these weeks busy with something else (sorry for this).😔

sauravrao637:
@Stypox I've done done the changes and you can now check them and suggest the changes if needed or merge it :)

app/src/main/res/values/colors.xml
```diff
@@ -25,7 +25,7 @@
     <!-- Dark Theme -->
     <color name="dark_background_color">#222222</color>
     <color name="dark_dialog_background_color">#424242</color>
-    <color name="dark_settings_accent_color">#ff5252</color>
+    <color name="dark_settings_accent_color">#992722</color>
```
Stypox, line 28: I think this should not be changed, so that settings buttons / bullet points / checkboxes / ..., overlayed on top of black/grey, have enough contrast. But I'm not sure this is the color causing that behaviour.

Stypox (changes requested):
Looks mostly good to me ;-)

app/src/main/res/values/colors.xml
sauravrao637: Yeah your are correct, I've reverted it now

OutshineIssue:
Can we add a full black top bar when on "black" mode theme?

Stypox (approved):
Great, thank you! :-)

sauravrao637:
@Stypox 
I think this can be merged and closed now.

Stypox:
@TeamNewPipe/all do we want to merge this?

Poolitzer:
Since I am in the team somehow and you pinged the whole team, yep, I like that change.

opusforlife2:
Can we have some screenshots for the updated code?

sauravrao637:
@opusforlife2 
Here are few screenshots, hope you like it :)

opusforlife2:
I like it! Thanks for this! The muted colours look very nice. Now I'm comparing this to the current release and the bright red kind of hurts the eyes.

sauravrao637:
So is this now finally ready to be merged?
:)

opusforlife2:
@sauravrao637 Thanks! The muted colours look great in the latest debug APKs I'm using. But now the splash screen also looks too bright.

Is it possible to use a darkened splash screen with the dark theme?

sauravrao637:
@opusforlife2 Unfortunately the splash screen cannot depend on the in app preferences because it is loaded before app is in memory so we cannot make multiple theme specific splash screens but what we can do instead is that we can use system preferences that is when the system is in night mode we can load a darker variant.
In easy and short terms we can implement it for when the system is running in night mode irrespective of the user's settings for theme in New Pipe.
**We can also do the custom splash screens for multiple themes but there's only UGLY way (making splash activity and defeating the real purpose of splash screen) to do so.**

opusforlife2:
@sauravrao637 Indeed. I was thinking about the Play Store splash screen when I suggested it. You're right that this depends on the OS theme, not app theme.

rahaaatul:
@sauravrao637 Can you add a true black theme, where the top red bar is black? Accent color can be red for YouTube and other colors in other instances.

mhmdanas:
@rahaaatul see https://github.com/TeamNewPipe/NewPipe/issues/4148. You may want to open a new issue for this, focusing on the aesthetic of such a theme.

rahaaatul:
> @rahaaatul see https://github.com/TeamNewPipe/NewPipe/issues/4148. You may want to open a new issue for this, focusing on the aesthetic of such a theme.

Can you open that, please? I think you can explain it much better than I can.

opusforlife2:
@rahaaatul If your reasoning is the same as in the linked issue, then there is no point. If it is different, then you'll need to open a new issue anyway with your new reasoning.

PR #6288, opened 2021-05-14
Add debug option to disable media tunneling

Redirion:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

Added an option for DEBUG builds only to help users to check if their issues with video (blackscreen or stuttering) are a cause of not working media tunneling on their device / custom os.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/res/values/settings_keys.xml
+++ b/app/src/main/res/values/settings_keys.xml
@@ -171,8 +171,8 @@
     <string name="allow_heap_dumping_key" translatable="false">allow_heap_dumping_key</string>
     <string name="show_memory_leaks_key" translatable="false">show_memory_leaks_key</string>
     <string name="allow_disposed_exceptions_key" translatable="false">allow_disposed_exceptions_key</string>
-    <string name="show_original_time_ago_key" translatable="false">show_original_time_ago_text_key</string>
-    <string name="disable_media_tunneling_key" translatable="false">disable_media_tunneling_text_key</string>
+    <string name="show_original_time_ago_key" translatable="false">show_original_time_ago_key</string>
+    <string name="disable_media_tunneling_key" translatable="false">disable_media_tunneling_key</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/settings_keys.xml
```diff
@@ -172,6 +172,7 @@
     <string name="show_memory_leaks_key" translatable="false">show_memory_leaks_key</string>
     <string name="allow_disposed_exceptions_key" translatable="false">allow_disposed_exceptions_key</string>
     <string name="show_original_time_ago_key" translatable="false">show_original_time_ago_text_key</string>
+    <string name="disable_media_tunneling_key" translatable="false">disable_media_tunneling_text_key</string>
```
TobiGr, line 175: What's the reason for using a different value here?
```suggestion
    <string name="disable_media_tunneling_key" translatable="false">disable_media_tunneling_key</string>
```

Redirion: was asking that myself. I was just copying from  `<string name="show_original_time_ago_key" translatable="false">show_original_time_ago_text_key</string>`

Redirion: I have removed it from both

TobiGr: if you change the key value, we might want to add a SettingsMigration. I am  not sure if this a bit too much effort for a rarely used debug setting.

Redirion: yes, I would say that a migration is not required. But out of curiosity, is there a concept for it in Android and/or do we already have a settings migration in NewPipe?

TobiGr: We already have one: [SettingsMigrations.java](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java)

TobiGr (approved):

PR #6319, opened 2021-05-17
Save backup import/export location for future import/exports

ATofighi:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

To save location of import/export, I add a new preference key for `import_export_path` and save this preference on result of file chooser activity and send it on intent of file chooser on later tries.

So as a user, If a export data on folder `foo`, when I want to import this data, file chooser show `foo` as start path.

#### Fixes the following issue(s)

- Fixes #6039 

#### APK testing 

On the website the APK can be found by going to the "Checks" tab below the title and then on "artifacts" on the right.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #5415, opened 2021-01-14
Support SAF properly

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
This is the rebased #3777, and also contains:
- Setting migration that enables SAF everywhere possible.

Other than that, it is identical to #3777 (by @wb9688):
1. Use SAF or NoNonsense-FilePicker *everywhere* based on the setting.
2. Don't have `requestLegacyExternalStorage` anymore in the manifest.
3. Allow only NoNonsense-FilePicker for versions below Android 5 and allow only SAF starting at Android 10.
4. Move `SharpInputStream` into NewPipe and add `SharpOutputStream`.
5. Show folder picker in download dialog if no download folder set yet.

#### TODO
- [x] "SAF" in the UI should be replaced with something more easily understandable for non-Android devs, like "system folder picker"
- [x] Remove "some devices are not compatible" from SAF setting description
- [x] Investigate whether renaming the zip works - it works
- [x] ~~Keeping the setting in the Download menu doesn't make sense anymore~~ implementing a miscellaneous menu is left for another PR
- [x] Investigate "NewPipe saf Leaks" - I can't understand why the name changed, but it's not important
- [x] Explain SharpOutputStream
- [x] "filename.mp4 (1)" is created instead of "filename (1).mp4" - fixed by passing mime type to file picker
- [x] Fix strange behaviour when app loses access to saf download folder
- [x] Reimplement storing import/export path

#### Fixes the following issue(s)
Fixes #3465, fixes #5198, and perhaps some others.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #6321, opened 2021-05-17
Allow rename database export file in settings

### LINKED ISSUE #5780: Allow renaming export file
### Checklist

Could you please add an option to rename the file while exporting? It's frustrating that we can't name it on our own. Every single time, it creates file with a different name.

If a new export file has the same name of any previous file in that folder, a popup should appear like "Np.xml already exists." Three options can be added as reply, "Replace file", "Save as Np1" and "Cancel exporting"

It becomes really difficult as there is no way to replace the previous file. Plus, keeping more files also turns things confusing. Allowing renaming will also let us avoid mixing up files with similar names.

ATofighi:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

In export database on settings. I show a new dialog and get file name from user. then I save export file with entered name.

- Fixes #5780 

- #6319 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
@@ -29,6 +29,7 @@ import org.schabi.newpipe.extractor.localization.ContentCountry;
 import org.schabi.newpipe.extractor.localization.Localization;
 import org.schabi.newpipe.util.FilePickerActivityHelper;
 import org.schabi.newpipe.util.FilePathUtils;
+import org.schabi.newpipe.util.FilenameUtils;
 import org.schabi.newpipe.util.ZipHelper;
 
 import java.io.File;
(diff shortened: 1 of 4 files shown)


SameenAhnaf:
Thank you so much for all the efforts. Works great on my device.

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -195,6 +197,33 @@ public void onActivityResult(final int requestCode, final int resultCode,
...
+        fileNameET.setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_URI);
+        fileNameET.setHint(R.string.export_data_file_name_hint);
+
+        final AlertDialog dialog = new AlertDialog.Builder(c)
+                .setTitle(R.string.export_data_file_name)
+                .setIcon(R.drawable.ic_import_export)
+                .setNegativeButton(R.string.cancel, null)
+                .setPositiveButton(R.string.export_button, (dialog1, which) -> {
+                    final String fileName = fileNameET.getText().toString();
+                    if (FilenameUtils.isValidFileName(fileName, "zip")) {
```
TobiGr, line 215: We could also append ".zip" when the file name does not have that extension.

ATofighi: So, should I don't ask extension of file in file name or if user enter file name without any dots, I append ".zip" at the end of it?

What should I do if user enter "foo.rar" ? make it "foo.zip"? or "foo.rar.zip"?

Stypox:
Note: don't merge this before #5415

Stypox:
There is no need for this anymore, as SAF supports renaming while creating files

PR #6322, opened 2021-05-17
issue #6217 fixed. replay button doesn't appear twice.

### LINKED ISSUE #6217: Replaying a video in any player requires tapping the replay button twice
### Checklist

- [x] I am using the latest version - 0.21.2 

A weird bug, but also weirdly consistent. 
1. Start playing a video in any player.
2. Wait for it to finish.
3. Swipe down to open up the notification tray, and tap the "replay" button, or tap the button in the main/popup player.
4. Observe the video doesn't replay.
5. Tap the replay button again, and watch it replay.

### Actual behaviour

Included above. Funnily enough, this behavior also persists after closing the notification tray or locking the phone - eg. if you tap it once, turn off the phone, and then turn it on and tap it again, it'll play.

### Expected behavior

The video should replay after one tap.

### Device info

 - Android version/Custom ROM version: GrapheneOS 11
 - Device model: Pixel 4a

khashiii97:
For solving this issue, the following changes where made in Player.java:
1. A boolean field named isOnReplay was added to reflect when the user has pressed the replay button.
2. when the video would reach it's end and the onCompleted method be called, the first time isOnreplay would be set to true. the second time it entered onCompleted (fallacious behavior) it would return since isOnReplay is true.
3.  in the onclick method, when the playpause button was clicked, if it was the replay button which triggered this event, playpause will be called twice (by checking isOnReplay for the second call).

Fixes #6217

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -375,7 +375,11 @@ public final class Player implements
     @NonNull private final SharedPreferences prefs;
     @NonNull private final HistoryRecordManager recordManager;
 
+     /*//////////////////////////////////////////////////////////////////////////
+    // this field is added to address issue #6217 regarding the replay button appearing twice.
+    //////////////////////////////////////////////////////////////////////////*/
 
+    private boolean isOnreplay = false;
 
     /*//////////////////////////////////////////////////////////////////////////
     // Constructor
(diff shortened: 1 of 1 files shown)


tsiflimagas:
This seems to have fixed the issue, in both normal and mini player, but not for the quick settings controls. You still have to press "replay" twice there.

litetex (changes requested):
Thank you for your PullRequest!

I found a few issues with this code:
1. We introduce a new variable only to check if we are in the "replay" state. That can certainly done without one as the state is already hold in the player.
2. It's not 100% safe that the variable is set to the correct when e.g. switching to another video (playlist) etc. This will certainly cause problems.

There is a way simpler solution:
grafik

I'll open a PR for the new solution.

litetex:
Succeeded by #7014

PR #6345, opened 2021-05-21
Test and clean up PlayQueue

Imericxu:
#### What is it?

- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Unit test PlayQueue class
- Clean up and refactor class as necessary (no major changes)
  - Fix edge cases in `setIndex(int): void`

#### APK testing 

On the website, the APK can be found by going to the "Checks" tab below the title and then on "artifacts" on the right.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
@@ -40,17 +40,15 @@ import io.reactivex.rxjava3.subjects.BehaviorSubject;
  */
 public abstract class PlayQueue implements Serializable {
     public static final boolean DEBUG = MainActivity.DEBUG;
-
-    private List<PlayQueueItem> backup;
-    private List<PlayQueueItem> streams;
-
     @NonNull
     private final AtomicInteger queueIndex;
     private final List<PlayQueueItem> history = new ArrayList<>();
 
+    private List<PlayQueueItem> backup;
+    private List<PlayQueueItem> streams;
+
     private transient BehaviorSubject<PlayQueueEvent> eventBroadcast;
     private transient Flowable<PlayQueueEvent> broadcastReceiver;
-
     private transient boolean disposed = false;
 
     PlayQueue(final int index, final List<PlayQueueItem> startWith) {
--- a/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
@@ -509,6 +509,11 @@ public abstract class PlayQueue implements Serializable {
         return streams.equals(other.streams);
     }
 
+    @Override
+    public int hashCode() {
+        return streams.hashCode();
+    }
+
     public boolean isDisposed() {
         return disposed;
     }
--- a/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
@@ -196,8 +196,6 @@ public abstract class PlayQueue implements Serializable {
      * @return the index of the given item
      */
     public int indexOf(@NonNull final PlayQueueItem item) {
-        // referential equality, can't think of a better way to do this
-        // todo: better than this
         return streams.indexOf(item);
     }
 
--- a/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
@@ -54,8 +54,7 @@ public abstract class PlayQueue implements Serializable {
     private transient boolean disposed;
 
     PlayQueue(final int index, final List<PlayQueueItem> startWith) {
-        streams = new ArrayList<>();
-        streams.addAll(startWith);
+        streams = new ArrayList<>(startWith);
         history = new ArrayList<>();
         if (streams.size() > index) {
             history.add(streams.get(index));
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
```diff
@@ -500,22 +519,19 @@ public synchronized boolean previous() {
      * we don't have to do anything with new queue.
      * This method also gives a chance to track history of items in a queue in
      * VideoDetailFragment without duplicating items from two identical queues
-     * */
+     */
     @Override
     public boolean equals(@Nullable final Object obj) {
```
XiangRongLin, line 524: Did you autogenerate it or manually implement it, same for hashCode

app/src/test/java/org/schabi/newpipe/player/playqueue/PlayQueueTest.java
```diff
@@ -0,0 +1,200 @@
...
+                    field.get(nonEmptyQueue)
+            );
+            assertEquals(1, history.size());
+
+            // Index 3 != 0, so the second history element should be the item at streams[3]
+            nonEmptyQueue.setIndex(3);
+            history = (List<PlayQueueItem>) Objects.requireNonNull(
+                    field.get(nonEmptyQueue)
+            );
+            assertEquals(nonEmptyQueue.getItem(3), history.get(1));
```
XiangRongLin, line 138: I have trouble understanding what this does and why reflection is necessary for it. Can't this somehow be done with mocks?

app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
Imericxu: They’re manually implemented. Since `List.equals` calls `equals` for every pair of objects in the list and `PlayQueueItem` overrides `equals` to compare `url`, I simplified it to `streams.equals(other.streams)`.

app/src/test/java/org/schabi/newpipe/player/playqueue/PlayQueueTest.java
```diff
@@ -0,0 +1,200 @@
...
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNull;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+
+@SuppressWarnings("checkstyle:HideUtilityClassConstructor")
+@RunWith(Enclosed.class)
+public class PlayQueueTest {
+    static PlayQueue mockPlayQueue(final int index, final List<PlayQueueItem> streams) {
```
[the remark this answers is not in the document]
Imericxu: Ok, I’ll change that. I just made them default so Android Studio didn’t have test gutter icons for them.

app/src/test/java/org/schabi/newpipe/player/playqueue/PlayQueueTest.java
Imericxu: Adding to history is a possible side effect of `setIndex(int): void`, so I thought I’d test that. I haven’t really worked with Mockito, so I just did this the old-fashioned way. `Whitebox.getInternalState(…)` also requires PowerMock.

XiangRongLin: I would not test this or at least not in this way. This is internal state of the class which should have some impact on the result of other methods. These other methods should be called and asserted on

app/src/test/java/org/schabi/newpipe/player/playqueue/PlayQueueTest.java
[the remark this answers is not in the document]
XiangRongLin: > Oh, wait. Actually, the inner test classes need them to be public/p-p to call them.

Ok.

> I was thinking of renaming it to makePlayQueue(…), or did you have something else in mind?

Nope, your suggestion is fine

app/src/test/java/org/schabi/newpipe/player/playqueue/PlayQueueTest.java
```diff
@@ -0,0 +1,186 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.spy;
+
+@SuppressWarnings("checkstyle:HideUtilityClassConstructor")
+@RunWith(Enclosed.class)
```
XiangRongLin, line 25: Right, there was this thing. Is it still experimental in JUnit 5?

From my understanding this only makes development inside Android Studio a bit more convenient, because you can run the class and all it's inner classes at once. But the usecase of the build system not finding the tests in not the case here with gradle

XiangRongLin:
Other than that it looks good to go for me

app/src/test/java/org/schabi/newpipe/player/playqueue/PlayQueueTest.java
Imericxu: I don’t know too much about it, I just found it to be a working way of grouping related test methods.

> this only makes development inside Android Studio a bit more convenient

This does seem to be the case; I just tried running tests from Gradle without it and they run just fine. I’ll remove the runner.

Imericxu:
I noticed that the `outOfBoundsNotComplete` test takes *much* longer than the rest of the tests; 1 whole second as compared to milliseconds. Could it be the `spy` and is this an issue?

XiangRongLin:
> 
> 
> I noticed that the `outOfBoundsNotComplete` test takes _much_ longer than the rest of the tests; 1 whole second as compared to milliseconds. Could it be the `spy` and is this an issue?

But aren't the spies used in every test? What i know of is, that the first tests always takes the longest, because it gets all the overhead of initializing stuff

XiangRongLin (approved):

XiangRongLin:
@TobiGr Should this be squashed before merging? On one hand, it's __20__ commits, on the other hand, some of them do have usefull messages that may be worth keeping

XiangRongLin:
No complaints means no problems 🤷

Issue #6145, opened 2021-04-23
Filter, order and set limit for videos


SameenAhnaf:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

Video filtering is a really useful feature. Unfortunately, Newpipe doesn't have it anywhere. Sometimes, only certain number of videos uploaded on a certain time are required. In such cases, set limit option is useful.

Following options can be added for video filtering anywhere. These options should be available as radio buttons.

1  Upload date: Last hour, today, before or after certain date
2  By video title
3. By channel name (This option might be removed from channel feed)
4. By video length
5. Features: Live, 4K, HD, Cc, creative commons, vr180, 360, 3D, HDR, Location
6. Most popular

All filtering options should come with ascending and descending order. 

Setting limits for videos will be really helpful especially when you only need a certain number of videos. "Set Limit" option should also follow "Sort by" and "Order" options.

Limit options may be like

1. Unlimited: All videos will be loaded.
2. Video number: A certain number of videos will be loaded.
3. Video length: Only videos of certain length will be loaded.
4.bTotal video length: Videos will be loading until it exceeds a certain time.
5..Keywords: Include (Single tap) or exclude (long press) videos with certain keywords in title and description. 
6. Channel: Include or exclude videos from certain channels. Not for channel  #feed.
7. Playlist: Include or exclude videos from certain playlists. Not for playlist feed.

These options could be placed under "Background, Play All and popup" row.

Default settings might be "Sort by: Date added", "Order: Descending", "Set Limit: Unlimited". However, all these settings should be seperately customisable for

1. Channel feed
2. Subscription feed
3. Playlist feed
4. Search results

#### How will you/everyone benefit from this feature?

Get your desired videos directly without wasting data.

Angelk90:
@SameenAhnaf : There are news regarding this, in the app I do not find the similar function.


![image](https://user-images.githubusercontent.com/20476002/128170650-9d107d28-847b-4623-ac41-64b98c481c01.png)

SameenAhnaf:
@Angelk90 Yes, there are only basic sorting for now. Advanced options are still unavailable.

[...]

Issue #5782, opened 2021-03-07
Add Download option for Long Press on Playlists


SameenAhnaf:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

"Download" option for playlist will be immensely helpful. It will help us download a full playlist. One more thing, it should not directly download the full playlist. Users may need to deselect a few videos.
"Deselect all" or "Select all" should appear on the basis of all selected or not all videos selected respectively.
![IMG_20210408_190944](https://user-images.githubusercontent.com/79245641/114032457-23ee3300-989e-11eb-9c59-73f49a85834c.jpg)

#### How will you/everyone benefit from this feature?

Bulk download will be much more easier.

NewNewPipe:
A similar proposal of mine has already been made. 8 days ago.

DOWNLOAD LIST - for later downloading.

https://github.com/TeamNewPipe/NewPipe/issues/5714

SameenAhnaf:
@NewNewPipe Thanks, but that's not what I actually asked. I asked to download all videos of the playlist on a single click. Bulk selection options #5657 should be added to enhance downloading experience.

You asked for a option to download later. I asked to download right here and now.

SameenAhnaf:
@mhmdanas Would you please on this issue as Nikilchad suggested in https://github.com/TeamNewPipe/NewPipe/pull/6309#issuecomment-846441294 ? Thanks in advance.

mhmdanas:
@SameenAhnaf please avoid repeatedly pinging people in issues where they are not participating.

opusforlife2:
@SameenAhnaf It is impolite to ping developers asking them to work on issues they may not be interested in. The developers are team members. They are generally aware of the goings-on in the repo, and if they're not, that means they are busy in real life. If they see an issue they are interested in working on, they will do so without needing to be prompted by anyone. For developers who contribute to the repo but aren't part of the team, it is okay to mention them in issues related to things they have already worked on, but even there it is better to exercise caution before doing such a thing.

[...]

Issue #5741, opened 2021-03-03
Replace "Play All" button with "Enqeue Method" drop-down menu


SameenAhnaf:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

"Play all" button only lets user to play all videos from the beginning. Sometimes, we play another video from the playlist. Still, we need to enqueue the whole or certain videos of the playlist. Following enqeue methods can be added to solve this issue.

1. All: All videos will be enqeued. Only, first video will be the selected video.
2. None: No video will be enqeued from the channel or playlist. However, enabling "Auto-queue for a related stream" should let the user auto-queue a related stream. Not next video from the channel or playlist 
3. Only next videos: Only videos arranged after selected videos will be enqeued.
4. Only previous videos: Only videos arranged before selected videos will be enqeued.
5. Continue to next videos: All videos will be enqeued with a slight variation. Suppose, a playlist has five videos: A, B, C, D, E. If C video is selected, 1st video will be C, 2nd will be D, 3rd be E, 4th be A and 5th be B and 5th be B.
6. Continue to previous videos: All videos will be enqeued with a slight variation. Suppose, a playlist has five videos: A, B, C, D, E. If C video is selected, 1st video will be C, 2nd will be B, 3rd will be A, 4th will be E and 5th will be D.

Reverse order for videos is not necessary as videos' sorting, order and limit can be set (mentioned in #5669). Also, only sorted videos should get enqeued if any sorting, order or limit option is applied.

An option in settings can be added for "Default Enqeue Method". However, default Enqeue Method drop-down menu should still persist in order to change any time without the hassle of changing settings every time. This option should only mark the enqeue method by default.

Plus, This should be kept customisable or each of the following options. Because, typically we play playlist at once. But that might not be the case for channels. That's why, seperate options for each feed are required. (I have already asked for "Background, Enqeue Method drop-down menu and Popup row" in #5737.for what's new, playlist category and subscription category feed)
1. Channel feed
2. Playlist feed
3. Subscription category feed
4. Playlist category feed
5. What's New feed

#### Additional context

Sort and set limit of the videos #5669
[...]

PR #6385, opened 2021-05-27
Option for download is redundant and thus removed when linkType is CH…

### LINKED ISSUE #6371: "Preferred 'Open' Action" setting should be applicable for videos only
### Checklist

- [x] I am using the latest version - 0.21.2 

It seems like that shared URL action doesn't always work the way as expected. In reality, people share videos, channels and playlists for different reasons. However, playback options are actually quite relevant for videos.

But that's not always the case with channels or playlists. Many users share directly by long press on link without opening the link. Checkback is a necessary feature and that's one of the key reasons why long press menu was not added for unsubscribed channels and playlists. That's why, info should always be shown.

One more thing, "Video Player" and "Show Info" works exactly the same for videos. So, having both options makes no real sense.

### Actual behaviour

"Preferred 'Open' Action" setting should be applicable for videos only. "Download" option ends up giving us no output for channels and playlists.

### Expected behavior

Info should always be shown for channels and playlists.

### Device info

 - Android version/Custom ROM version:
 - Device model:

sauravrao637:
…ANNEL or PLAYLIST

#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- The Download button should not be included in option action choices when linkType is CHANNEL or PLAYLIST (as it will just give unsupported url ) , thus done the needed changes

- Fixes #6371 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -466,6 +466,11 @@ public class RouterActivity extends AppCompatActivity {
             if (capabilities.contains(AUDIO)) {
                 returnList.add(backgroundPlayer);
             }
+            // download is redundant for linkType CHANNEL AND PLAYLIST (till playlist downloading is
+            // not supported )
+            returnList.add(new AdapterChoiceItem(getString(R.string.download_key),
+                    getString(R.string.download),
+                    R.drawable.ic_file_download));
 
         } else {
             returnList.add(showInfo);
@@ -477,10 +482,6 @@ public class RouterActivity extends AppCompatActivity {
                 returnList.add(backgroundPlayer);
             }
         }
-
-        returnList.add(new AdapterChoiceItem(getString(R.string.download_key),
-                getString(R.string.download),
-                R.drawable.ic_file_download));
 
         return returnList;
     }


SameenAhnaf:
Please, wait a minute. Bulk download was actually a requested feature. See: https://github.com/TeamNewPipe/NewPipe/issues/5782. It's expected that 'Download' option will work the same way.

However, https://github.com/TeamNewPipe/NewPipe/issues/6145 and https://github.com/TeamNewPipe/NewPipe/issues/5741 are requested for sorting videos within a playlist. That's why, I asked for 'Show info' option only for playlists as users may want to sort or enqueue videos according to their preferences. In fact, that's when Newpipe app will come in handy the most.

sauravrao637:
From what I can understand, for linkType CHANNEL download is completely redundant and thus should not be shown as an option and till "bulk Download" is pending we should rely on this and when it's available we can put the option back for linkType PLAYLIST (I'll myself add back as soon as the feature is available)
And by "'Show info' option only for playlists" ,I think this should be an option for CHANNELS as well as for any STREAM, as user may only want to see say "Description" of a video
If I've misinterpreted , I would like to be further assisted with the issue :)

SameenAhnaf:
>From what I can understand, for linkType CHANNEL download is completely redundant and thus should not be shown as an option 

Your points are quite valid and logical. So, this PR still makes sense.

>and till "bulk Download" is pending we should rely on this and when it's available we can put the option back for linkType PLAYLIST (I'll myself add back as soon as the feature is available)

Thank you so much for your help in advance.

>And by "'Show info' option only for playlists" ,I think this should be an option for CHANNELS as well as for any STREAM, as user may only want to see say "Description" of a video

Unfortunately, 'Show info' is still useless for videos. Enabling auto-play makes the video play. Disabling it doesn't do so as expected. Rather, shared videos should not be dependent on such a setting.

Show info: Don't auto-play videos
Video player: Play videos on main player

Could you please add this feature on some other PR? It would be really nice of you.

>If I've misinterpreted , I would like to be further assisted with the issue :)

It's me who should have apologized. Pardon me for my vague explanation at first. Also, thanks once again for your time and efforts.

TobiGr (approved):
Thank you

sauravrao637:
@SameenAhnaf 
Thanks for being patient while explaining and also for reporting the issue 
Also feel free to open an Issue for any feature you feel can be useful

Issue #6435, opened 2021-06-05
Add a button to copy URL of a video to clipboard


Abanoub8:
### Checklist

- [ ] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

I considered the issue, and I made a patch that adds a button, beside the open in browser, and the share, buttons.

#### Is your feature request related to a problem? Please describe it

No, my feature is not related to a problem.

#### Additional context

I have the patch and I want to know if you will be interested to accept it.

#### How will you/everyone benefit from this feature?

This will make it easier to take the URL, and save it in a file, or open it in another app, or sending it to a friend, or sync it.

SameenAhnaf:
You can install "Copy to clipboard" app as an alternative. This app will copy link if you select the app from share menu. There are many apps just like Newpipe which don't have 'Copy link' option.

https://f-droid.org/packages/com.tengu.sharetoclipboard/

gshirodkar:
In Android, you can also copy to clipboard from within the share menu.

SameenAhnaf:
@gshirodkar As far as I know, only YouTube app can do so. The key reason why this issue was opened is the fact that 'Copy link' option doesn't appear in case of all apps.

opusforlife2:
@Abanoub8 Rather than add a new button entirely, why not overload the Share button? You could make it so that long-pressing it copies the URL to the clipboard. (If you do this, make sure to also add a tooltip for taps which explain the user could long-press, like some of the other buttons.)

infinitewaveparticle:
I like these ideas because clicking Share opens a bunch of apps that I then have to force close (this is an issue on older devices with limited CPU). Once so many apps are open simply browsing the internet will cause Android UI itself to crash and all my apps restart (which means my firewall and ad blocker are also killed).

rey0n:
Long press on Share button to copy link is best variant.!

gshirodkar:
@SameenAhnaf not sure if it's a feature introduced in a new version of Android, but I get the copy link for sharing from any app. 
![Screenshot_20210607-003042](https://user-images.githubusercontent.com/11176259/120959474-d704cc80-c727-11eb-8873-0b6a0adcdac4.png)

[...]

PR #6436, opened 2021-06-05
Add copy url function to share button

Abanoub8:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

This pull request adds a button beside the share button and open in browser button, that copies the URL of the video to clipboard. This is very useful in many situations.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -532,6 +532,7 @@ public final class Player implements
         binding.moreOptionsButton.setOnClickListener(this);
         binding.moreOptionsButton.setOnLongClickListener(this);
         binding.share.setOnClickListener(this);
+        binding.share.setOnLongClickListener(this);
         binding.fullScreenButton.setOnClickListener(this);
         binding.screenRotationButton.setOnClickListener(this);
         binding.playWithKodi.setOnClickListener(this);
@@ -3629,6 +3630,10 @@ public final class Player implements
             fragmentListener.onMoreOptionsLongClicked();
             hideControls(0, 0);
             hideSystemUIIfNeeded();
+        } else if (v.getId() == binding.share.getId()) {
+            if (currentMetadata != null) {
+                ShareUtils.copyToClipboard(context, currentMetadata.getMetadata().getOriginalUrl());
+            }
         }
         return true;
     }
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -2919,6 +2919,18 @@ public final class Player implements
     }
 
     @NonNull
+    private String getVideoUrlAtCurrentTime() {
+        final int timeSeconds = binding.playbackSeekBar.getProgress() / 1000;
+        String videoUrl = getVideoUrl();
+        if (!isLive() && timeSeconds >= 0 && currentMetadata != null
+                && currentMetadata.getMetadata().getServiceId() == YouTube.getServiceId()) {
+            // Timestamp doesn't make sense in a live stream so drop it
+            videoUrl += ("&t=" + timeSeconds);
+        }
+        return videoUrl;
+    }
+
+    @NonNull
     public String getVideoTitle() {
         return currentMetadata == null
                 ? context.getString(R.string.unknown_content)
@@ -3581,7 +3593,7 @@ public final class Player implements
         } else if (v.getId() == binding.moreOptionsButton.getId()) {
             onMoreOptionsClicked();
         } else if (v.getId() == binding.share.getId()) {
-            onShareClicked();
+            ShareUtils.shareText(context, getVideoTitle(), getVideoUrlAtCurrentTime());
         } else if (v.getId() == binding.playWithKodi.getId()) {
             onPlayWithKodiClicked();
         } else if (v.getId() == binding.openInBrowser.getId()) {
@@ -3631,9 +3643,7 @@ public final class Player implements
             hideControls(0, 0);
             hideSystemUIIfNeeded();
         } else if (v.getId() == binding.share.getId()) {
-            if (currentMetadata != null) {
-                ShareUtils.copyToClipboard(context, currentMetadata.getMetadata().getOriginalUrl());
-            }
+            ShareUtils.copyToClipboard(context, getVideoUrlAtCurrentTime());
         }
         return true;
     }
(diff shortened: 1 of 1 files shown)


app/src/main/res/layout/player.xml
```diff
@@ -305,6 +305,21 @@
...
+                        android:layout_height="35dp"
+                        android:layout_marginEnd="8dp"
+                        android:background="?attr/selectableItemBackground"
+                        android:clickable="true"
+                        android:contentDescription="@string/get_url"
+                        android:focusable="true"
+                        android:padding="@dimen/player_main_buttons_padding"
+                        android:scaleType="fitXY"
+                        app:tint="@color/white"
+                        app:srcCompat="@drawable/ic_edit"
```
Stypox, line 320: Why ic_edit?

app/src/main/res/values/strings.xml
```diff
@@ -9,6 +9,7 @@
     <string name="cancel">Cancel</string>
     <string name="fdroid_vlc_url" translatable="false">https://f-droid.org/repository/browse/?fdfilter=vlc&amp;fdid=org.videolan.vlc</string>
     <string name="open_in_browser">Open in browser</string>
+    <string name="get_url">Get URL</string>
```
Stypox, line 12: I'd use "Copy URL" instead of "get". Also change this in code

Stypox (changes requested):

app/src/main/res/layout/player.xml
Abanoub8: Just because I needed an icon, and edit seems accurate enough, as a place holder at least. It also looks like copy to edit or something.

app/src/main/res/values/strings.xml
Abanoub8: Ok, I was thinking about copy or get, but I choose get, since get is more like: get the url of this video. But it makes sense to be copy as well.

app/src/main/res/layout/player.xml
Stypox: Add the `content_copy` icon: https://fonts.google.com/icons?selected=Material%20Icons%3Acontent_copy
You can add it seamlessly in Android Studio by right-clicking on `res` -> New -> Vector asset
Then follow the dialog. Add two icons: a white one to the `drawable` folder and a black one to the `drawable-night` folder. Name them `ic_content_copy.xml`.

Abanoub8:
I added the icons, and changed the string to Copy URL, do I need to change anything else?

Stypox:
Wait, now that I think of it, maybe this solution is better: https://github.com/TeamNewPipe/NewPipe/issues/6435#issuecomment-855354504
Adding many buttons is not a good idea; just let the share button be long-pressed. Also, add this long-press action to the button in video details, too. Sorry for not telling you this before.

Abanoub8:
To be honest I was thinking about doing that do, but I thought that something basic like this would have its own button. Anyways, I would have to figure out how to do this now.

Abanoub8:
I found out that you already had an OnLongClickListener, so it was only a matter of rewiring, and now it is done. I can do one thing more, which is remove the ic_content_copy.xml file if you don't want it?

opusforlife2:
@Abanoub8 Please update your PR title and description accordingly, and link it to your issue so it gets closed automatically upon merging this.

Also, please add a tooltip like the other long-press buttons.

In fact, @Stypox shouldn't there also be a tooltip for the download button? Long-pressing it opens Downloads, which is almost like an easter egg every user has to discover on their own or read about somewhere.

Abanoub8:
I already changed the title, and for some reason I am unable to link the issue, so I can close it manually, or delete it.
As for the tooltip, I will figure out how to do that, and I can also do it for the download button if you want.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3628,6 +3629,8 @@ public boolean onLongClick(final View v) {
             fragmentListener.onMoreOptionsLongClicked();
             hideControls(0, 0);
             hideSystemUIIfNeeded();
+        } else if (v.getId() == binding.share.getId()) {
+            getUrlClicked();
```
Stypox, line 3633: Just move the function inline, as it is small
```suggestion
        if (currentMetadata != null) {
            ShareUtils.copyToClipboard(context, currentMetadata.getMetadata().getOriginalUrl());
        }
```

app/src/main/res/drawable-night/ic_content_copy.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
Stypox, line 1: Remove these, they are not needed anymore

Stypox (changes requested):
Thank you!

> Edit: I noticed that some buttons are showing the wrong tooltip, like the share button and mute button showing "Bookmarked Playlists".
Also how should the share button have two tooltips? for the short and long press.

If it requires these many changes, maybe let's keep this PR just for the player and do that in another PR. What do you think?

app/src/main/res/drawable-night/ic_content_copy.xml
Abanoub8: Done

app/src/main/java/org/schabi/newpipe/player/Player.java
Abanoub8: Done

Abanoub8:
I fixed the problems with the code, and the unneeded icons, as for the tooltips, I would say maybe starting a new PR would be better.

opusforlife2:
Agreed on the separate PR for the other tooltips. They aren't related to your feature or code, anyway. If you're willing to make a PR for all the tooltips, that's fine. But if not, then I think you should add at least the Share button tooltip in this PR.

> Also how should the share button have two tooltips? for the short and long press.

There is only one type of tooltip. There is a toggle for this in the Appearance menu that is on by default. When you tap on the Play in Background or Popup buttons, it shows a tooltip telling you you can long-press the same buttons to enqueue instead.

So what I'm getting at is that tapping on the Share button should show a tooltip saying something like "You can long-press to copy the URL directly".

Abanoub8:
Ok, I will add the tooltip, for the share button, then we can make a new PR to fix the others.

Abanoub8:
@opusforlife2 Is the tooltip the same as the android:contentDescription?
and can I set the tooltip using android:tooltipText="for bar" in the xml?

I would be happy to work on the tooltip for other stuff, but need to know how do you set it.

opusforlife2:
@Abanoub8 No idea about the code. You could follow the trail from the "Show 'Hold to append' tip" toggle in the Appearance menu? Maybe that helps?

Abanoub8:
I can add cad tooltips in the xml like so: android:tooltipText="foo bar", but then you would need to add it in all layouts and views, and need to edit them one by one in the future if you need so.

opusforlife2:
Well, I'm out of my depth here. ¯\\\_(ツ)\_/¯ Only a developer can help you.

Abanoub8:
So to add a tooltip to the share button, what do I need to do?

app/src/main/res/values/strings.xml
```diff
@@ -9,6 +9,7 @@
     <string name="cancel">Cancel</string>
     <string name="fdroid_vlc_url" translatable="false">https://f-droid.org/repository/browse/?fdfilter=vlc&amp;fdid=org.videolan.vlc</string>
     <string name="open_in_browser">Open in browser</string>
+    <string name="copy_url">Copy URL</string>
```
Stypox, line 12: Oh, sorry, forgot to say you should also remove this string, then the PR can be merged.

@opusforlife2 this PR only has to do with a button in the player dropdown menu, so no tooltips here

Stypox (changes requested):

Stypox (dismissed):
I did that last thing myself. Thank you for the contribution :-)

Stypox (approved):

PR #6480, opened 2021-06-11
Added itemsListHeaderDuration to large-hand player

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
- The ``temsListHeaderDuration`` element was missing in the ``large-hand`` player which caused crashes on larger devices like TVs or tablets to crash

Problem was caused by #6441 / https://github.com/TeamNewPipe/NewPipe/commit/d921e2e61b3577469b26e361bd05483343f7b864

#### Fixes the following issue(s)
- Fixes #6467

#### APK testing 
On the website the APK can be found by going to the "Checks" tab below the title and then on "artifacts" on the right.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #6444, opened 2021-06-07
Increase buffer for playback after rebuffer

Redirion:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

NewPipe used a very small/short buffer of just 500ms to fill after a buffer depletion. Some devices seem to have issues with that and get in a loop of draining this buffer and having to refill it again leading to a short buffering every 500ms.
From ExoPlayer doc:
`The default duration of media that must be buffered for playback to resume after a rebuffer, in milliseconds. A rebuffer is defined to be caused by buffer depletion rather than a user action`
We are using ExoPlayer's default value now which is 5000ms.

#### Fixes the following issue(s)

- Fixes #6425 

#### APK testing 

On the website the APK can be found by going to the "Checks" tab below the title and then on "artifacts" on the right.

#### Due diligence
- [X] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #6493, opened 2021-06-15
Hotfix release 0.21.5

B0pol:
# Changelog
### Hotfix

- Increase buffer for playback after rebuffer #6444 
- Fixed crash on tablets and TVs when clicking on the play-queue icon in the player #6480 
- disable desugaring
- 
# APK for testing

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/6658963/app-debug.zip)

Please report new bugs in #6494

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -17,8 +17,8 @@ android {
         resValue "string", "app_name", "NewPipe"
         minSdkVersion 19
         targetSdkVersion 29
-        versionCode 970
-        versionName "0.21.4"
+        versionCode 971
+        versionName "0.21.5"
 
         multiDexEnabled true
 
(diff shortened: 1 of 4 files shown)


app/build.gradle
```diff
@@ -73,7 +73,7 @@ android {
 
     compileOptions {
         // Flag to enable support for the new language APIs
-        coreLibraryDesugaringEnabled true
+        coreLibraryDesugaringEnabled false
```
Stypox, line 76: @Isira-Seneviratne do you have any idea why the build works without issues even if this is set to `false`? How can Java methods be desugared if this is missing?

TobiGr (approved):

app/build.gradle
TobiGr: Build works, but running the app on API 19 does not. We need to revert this

TobiGr (approved):

PR #6504, opened 2021-06-17
dismiss choice dialog in onStop() to avoid a leaked window Exception:

evermind-zz:
E/WindowManager: android.view.WindowLeaked: Activity org.schabi.newpipe.RouterActivity has leaked window DecorView@d99fe3b[] that was originally added here
        at android.view.ViewRootImpl.<init>(ViewRootImpl.java:418)

#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- dismiss the dialog correctly so no leaks might occur

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -107,6 +107,7 @@ public class RouterActivity extends AppCompatActivity {
     protected String currentUrl;
     private StreamingService currentService;
     private boolean selectionIsDownload = false;
+    private AlertDialog alertDialogChoice = null;
 
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -126,6 +127,15 @@ protected void onCreate(final Bundle savedInstanceState) {
                 ? R.style.RouterActivityThemeLight : R.style.RouterActivityThemeDark);
     }
 
+    @Override
+    protected void onStop() {
```
Redirion, line 131: is there any reason why you decided to use onStop in favor of onPause? I think it could also be put there to be cleared up as early as possible

Stypox: I don't know, I don't think it matters much

Stypox (approved):
It looks good in my opinion, since `onStop` is called once when the activity is being shut down completely, while `onPause` may be also called for other reasons (e.g. user turned off the screen, but that shouldn't dismiss dialogs). Thank you @evermind-zz :-)

app/src/main/java/org/schabi/newpipe/RouterActivity.java
evermind-zz: the reason is as @Stypox stated.

PR #6512, opened 2021-06-18
Player.java Move listeners to internal classes

evermind-zz:
For me it was at first not so easy  to understand which method belongs to which Interface in Player.java.

So I decided to move methods implementing interfaces into their own innner classes. In my opinion this will increase the readability for new developers. 

The patches are from May but I rebased them today!

**

#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- create innner classes for each Listener
- increase readabilty

#### APK testing 

#### Due diligence


app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -401,6 +401,17 @@ public Player(@NonNull final MainPlayer service) {
...
+        playbackStateListener = new PlaybackStateListener();
+        playerPlaybackListener = new PlayerPlaybackListener();
+        playerImageLoaderListener = new PlayerImageLoadingListener();
+        playerVideoListener = new PlayerVideoListener();
+
+        playerOnSeekBarChangeListener = new PlayerOnSeekBarChangeListener();
+        playerViewOnClickListener = new PlayerViewOnClickListener();
+        playerViewOnLongClickListener = new PlayerViewOnLongClickListener();
+        playerPopupMenuOnMenuItemClickListener = new PlayerPopupMenuOnMenuItemClickListener();
+        playerPopupMenuOnDismissListener = new PlayerPopupMenuOnDismissListener();
```
Stypox, line 414: Just initialize them directly in the `class` body.

Stypox (commented):
I don't know if I want this merged. When I ordered the functions I used some semantically chosen categories. What you are proposing here is, instead, choosing categories just based on what listeners the functions belong to, not what they do and what they mean. So here the click listeners are all over the place, and they are more difficult to navigate in my opinion.

evermind-zz:
> I don't know if I want this merged. When I ordered the functions I used some semantically chosen categories. What you are proposing here is, instead, choosing categories just based on what listeners the functions belong to, not what they do and what they mean. So here the click listeners are all over the place, and they are more difficult to navigate in my opinion.

Yes, I saw the categories and I understand the idea, But I still felt lost inside this big class. Yes all related methods should also be moved inside the corresponding inner listener class. I'll update the pull request.

app/src/main/java/org/schabi/newpipe/player/Player.java
evermind-zz: will change that

evermind-zz:
Maybe its not worth at all. Maybe we just could fix the region displaying correctly in android studio. For example changing

`--- a/app/src/main/java/org/schabi/newpipe/player/Player.java`
`+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java`
`@@ -4038,10 +4038,9 @@ public final class Player implements`
`     //endregion`
` `
` `
`-    /*//////////////////////////////////////////////////////////////////////////`
`-    // Getters`
`-    //////////////////////////////////////////////////////////////////////////*/`
`-    //region`
`+    //////////////////////////////////////////////////////////////////////////`
`+    //region Getters`
`+    //////////////////////////////////////////////////////////////////////////`

Nevertheless I've updated the pull request with moving the initialization  to the body

Stypox:
@evermind-zz Oh, I didn't know you could use labels with `//region`. I like your solution, that looks readable both in IDEs with region code folding (Android Studio) and others (e.g. GitHub).

Stypox:
Any update?

evermind-zz:
> Any update?

I'll did not yet had the time to do it! I will try to find some time

Issue #6510, opened 2021-06-17
Buffering video unrelated to connection


n0rder:
[Screen recording](https://linksharing.samsungcloud.com/5PFO7lkTlSVj) 

As you can see in the screen record videos randomly buffer when in video mode, however in background mode the audio plays fine.

This issue is limited to newpipe and other video apps work fine.

Samsung galaxy s10+ android 11

Reinstalling newpipe doesn't work
Changing Internet connection doesn't work

matheusAMDS:
I can't visualize the link you shared, but i'm having problems with my videos infinitely loading. Maybe its something related?

LauriKat:
I am having same issue as yoy.

Redmi Note 8 pro
Android 10

botistreamer:
I am also having same issue in 0.21.5 on some videos randomly. 
Poco X3 - Android 11

B0pol:
Same

djuarezr:
Same here, even though with version 0.21.5 the problem has been mitigated a lot. If logs are needed just let me know.

Xiaomi Mi 9T - Miui 12.0.5 - Android 10

Redirion:
please try this apk: https://github.com/TeamNewPipe/NewPipeExtractor/pull/604

if the problem also happens with that apk, please collect the adb logcat logs from the issue

mhmdanas:
For all those facing this issue, using external players will probably not have this issue.

djuarezr:
It is the first time I try to get a logcat so I do not know whether it is what you need, but I got it. The logcat is from TeamNewPipe/NewPipeExtractor#604. Video starts at line 12, and after 30 seconds it starts the continuous buffering.

Hope this helps.

[NewPipe_YoutubeI_Api.log](https://github.com/TeamNewPipe/NewPipe/files/6676465/NewPipe_YoutubeI_Api.log)

Redirion:
thanks. A quick glance shows these log lines demand further inspection as they might be the cause (but not the root cause) of the issue:

AudioTrack: getTimestamp_l(417): retrograde timestamp time corrected
AudioTrack: getTimestamp_l(417): device stall time corrected using current time

Redirion:
@djuarezr could you try if the issue also happens with Media Tunneling disabled? You find the option to disable that in the debug options.

djuarezr:
[...]

PR #6515, opened 2021-06-18
Use ExoPlayer default values for buffers

Redirion:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)

ExoPlayer 2.12 noted this in the changelog:
_Set the default minimum buffer duration in DefaultLoadControl to 50 seconds (equal to the default maximum buffer), and treat audio and video the same._

Maybe there was a reason they decided to have both minimum and maximum buffer identical since 2.12. NewPipe used years old constants which are now removed in favor of ExoPlayer constants so that future versions will keep up with what changes at ExoPlayer.

Hopefully fixes some of the buffer issues we have lately like #6510 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/helper/LoadController.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/LoadController.java
@@ -20,18 +20,16 @@ public class LoadController implements LoadControl {
     //////////////////////////////////////////////////////////////////////////*/
 
     public LoadController() {
-        this(PlayerHelper.getPlaybackStartBufferMs(),
-                PlayerHelper.getPlaybackMinimumBufferMs(),
-                PlayerHelper.getPlaybackOptimalBufferMs());
+        this(PlayerHelper.getPlaybackStartBufferMs());
     }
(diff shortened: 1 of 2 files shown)


opusforlife2:
It does seem like this solves the issue.

In fact, it seems to have taken Newpipe back to the old behaviour, where it keeps buffering the video endlessly, unless you rewind or fast forward (may not work), or change the resolution (usually works).

tsiflimagas:
@opusforlife2 I got the same error too, without this commit, so maybe it's not related to it.

litetex:
Is this pull request anymore required?

The exact problem was found and we got https://github.com/TeamNewPipe/NewPipe/pull/6673 that hopefully fixes it 😄

XiangRongLin:
Unless there are good reasons to deviate from the default values, I would still merge this. The guys from exoplayer probably put more thought into those values than we.

TobiGr (approved):
Looks good to me. I did not ran into any issues while testing.

PR #6516, opened 2021-06-18
Don't enable SAF on Fire TV

### LINKED ISSUE #6455: SAF file open/close dialogues can't be confirmed on FireTV (FireOS bug)
### Checklist

- [x] I am using the latest version - v0.21.4 

So this isn't really a bug by NewPipe. In fact NewPipe doesn't do anything wrong but the new SAF dialogue triggers a long standing bug/limitation in FireOS used on some Amazon FireTV devices.

The issue is that Amazon forgot to modify the SAF file open/save dialogues to make them work with a remote. While you can navigate them, it's not possible to confirm any selection made in these menus unless you are using a separate input device like a mouse.

My Fire TV Stick 4K running the latest FireOS 6 (based on Android 7.1) is affected by this issue. I don't know if newer FireTV devices running FireOS 7 (based on Android 9) are still affected by this.

If possible, one solution would be to detect FireOS TV devices and disable SAF by default for the time being.

Do you think it's worth spending time working around an OS bug like this?

Edit: Of course this bug only applies to Fire TVs and TV boxes. The SAF dialogues can be navigated perfectly fine on my Fire HD 10 tablet with its touch screen.

1. Install NewPipe on an up to date FireTV 4K Stick
2. Leave the settings on default (SAF is turned on by default)
3. Use any option that causes the SAF file open/save dialogue to appear - for example try downloading a video
4. The SAF file save dialogue will appear.
5. You will notice that it's not possible to confirm this dialogue. For example in the save dialogue you won't be able to choose a folder for the default download location as you can't reach "select".

### Device info

 - Android version/Custom ROM version: FireOS 6 based on Android 7.1
 - Device model: Fire TV Stick 4K

mhmdanas:
#### What is it?
- [x] Bugfix (user facing)

Don't enable SAF when on Fire TV.

- Fixes #6455

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
@@ -111,6 +111,8 @@ public final class NewPipeSettings {
     }
 
     public static boolean useStorageAccessFramework(final Context context) {
+        // There's a FireOS bug which prevents SAF open/close dialogs from being confirmed with a
+        // remote (see #6455).
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP || DeviceUtils.isFireTv()) {
             return false;
         } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
--- a/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
@@ -111,7 +111,9 @@ public final class NewPipeSettings {
     }
 
     public static boolean useStorageAccessFramework(final Context context) {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q && !DeviceUtils.isFireTv()) {
+        if (DeviceUtils.isFireTv()) {
+            return false;
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
             return true;
         } else if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
             return false;
--- a/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
@@ -111,12 +111,10 @@ public final class NewPipeSettings {
     }
 
     public static boolean useStorageAccessFramework(final Context context) {
-        if (DeviceUtils.isFireTv()) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP || DeviceUtils.isFireTv()) {
             return false;
         } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
             return true;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -35,6 +36,18 @@
     private DeviceUtils() {
     }
 
+    public static boolean isFireTv() {
+        if (isFireTV != null) {
+            return isFireTV;
+        }
+
+        final boolean isFireTv =
```
TobiGr, line 44: why do you create a new variable?

mhmdanas: I was just following the pattern of `isTv()` below. If you think it's not needed I can remove it.

app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
```diff
@@ -110,7 +111,9 @@ private static String getNewPipeChildFolderPathForDir(final File dir) {
     }
 
     public static boolean useStorageAccessFramework(final Context context) {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+        if (DeviceUtils.isFireTv()) {
+            return false;
+        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
             return true;
         } else if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
             return false;
```
XiangRongLin, line 119: Can the 2 `return false` cases be combined with `||`?

mhmdanas: Done.

TobiGr: There should be a comment for each condition. I forgot to request that when reviewing the SAF PR.
```suggestion
        if (DeviceUtils.isFireTv()
            || Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
             // SAF file open/close dialogues can't be confirmed on FireTV (see #6455)
             // SAF is not supported on Android versions prior Lollipop
            return false;
        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
            return true;
```
@Stypox I forgot the reason for using SAF on Android 10 and above only.

TobiGr: Edit: delete me, I am stupid and late to the game :laughing:

mhmdanas: So is this fine as-is, or do I add a comment for Lollipop? I think the Lollipop part is fairly explanatory, while I added a comment for the FireTV part because it's not immediately obvious.

XiangRongLin (approved):
Codewise looks fine to me.

Please use squash merge

app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
TobiGr: All good.

PR #6537, opened 2021-06-20
Add support for delivery methods other than progressive HTTP (in the player only)

### LINKED ISSUE #2415: No playback of certain videos, infinite loading for HLS/DASH streams

I am having issues playing certain videos, they all exhibit the same problem by sitting with the spinning wheel and nothing happening. If i select an external player, there is a pause followed by an error notification, which I have screencapped and says 'the location.......full screen url....cannot be played'. The fault occurs only on some resolutions, and it has been observed that in most cases the qualities 720p or 360p always work (e.g. https://youtu.be/zCGouRiflIw).

For example videos;
https://www.youtube.com/watch?v=2NT-MRkYeSo
All the recent videos on his channel exhibit the same issue.

https://www.youtube.com/watch?v=ZHeqOwV5bow
Some of the video on this channel play, some do not.

### LINKED ISSUE #6860: Crash when playing PeerTube videos
navigated newpipe to the video at
https://watch.breadtube.tv/videos/watch/2a7d96e2-3f4e-432c-97cc-f3774659324f

Newpipe threw the following exception:

## Exception
* __User Action:__ ui error
* __Request:__ ACRA report
* __Content Country:__ US
* __Content Language:__ en-US
* __App Language:__ en_US
* __Service:__ none
* __Version:__ 0.21.8
* __OS:__ Linux samsung/crownqltesq/crownqltesq:9/PPR1.180610.011/N960USQU2CSGA:user/release-keys 9 - 28

```
java.lang.NumberFormatException: For input string: ""
	at java.lang.Integer.parseInt(Integer.java:627)
	at java.lang.Integer.parseInt(Integer.java:650)
	at org.schabi.newpipe.util.ListHelper.compareVideoStreamResolution(ListHelper.java:488)
	at org.schabi.newpipe.util.ListHelper.compareVideoStreamResolution(ListHelper.java:503)
	at org.schabi.newpipe.util.ListHelper.lambda$UGNhfSv5KuYwsqrcThyeG189sI0(Unknown Source:0)
	at org.schabi.newpipe.util.-$$Lambda$ListHelper$UGNhfSv5KuYwsqrcThyeG189sI0.compare(Unknown Source:4)
	at java.util.Collections$ReverseComparator2.compare(Collections.java:5226)
	at java.util.TimSort.countRunAndMakeAscending(TimSort.java:360)
	at java.util.TimSort.sort(TimSort.java:220)
	at java.util.Arrays.sort(Arrays.java:1498)
	at java.util.ArrayList.sort(ArrayList.java:1470)
	at java.util.Collections.sort(Collections.java:201)
	at org.schabi.newpipe.util.ListHelper.sortStreamList(ListHelper.java:269)
	at org.schabi.newpipe.util.ListHelper.getSortedStreamVideosList(ListHelper.java:243)
	at org.schabi.newpipe.util.ListHelper.getSortedStreamVideosList(ListHelper.java:130)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.handleResult(VideoDetailFragment.java:1540)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.lambda$runWorker$6(VideoDetailFragment.java:893)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.lambda$runWorker$6$VideoDetailFragment(Unknown Source:0)
	at org.schabi.newpipe.fragments.detail.-$$Lambda$VideoDetailFragment$X-UcNln23jPsop2TvHe2VPTu0NQ.accept(Unknown Source:8)
	at io.reactivex.rxjava3.internal.observers.ConsumerSingleObserver.onSuccess(ConsumerSingleObserver.java:62)
	at io.reactivex.rxjava3.internal.operators.single.SingleObserveOn$ObserveOnSingleObserver.run(SingleObserveOn.java:81)
	at android.os.Handler.handleCallback(Handler.java:873)
	at android.app.ActivityThread.main(ActivityThread.java:6990)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1445)

```
</details>
<hr>

TiA4f8R:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)

This PR adds support of YouTube OTF streams and the full support of ended livestreams in NewPipe (these streams cannot be downloaded or played in external players right now).
It also fixes the seek of PeerTube streams which have HLS-only variants of streams.
The support for downloading these streams **will be not added in this PR**.

Some code is inspired from #3803. 

For more detailed changes, check the commits.

**TO DO:**
- [x] Fix the crash when pressing the Download button (`IndexOutOfBoundsException`);
- [x] Add a method in `ListHelper` to return only streams which have a delivery method (for e.g. return only HLS streams for PeerTube videos in the player)
- [ ] Use the `PROGRESSIVE_HTTP` delivery method for external players when possible and add ability to play other variants of streams in these players
- [x] Support HLS "raw" manifests (the manifests which are returned by the extractor and not the URLs of HLS manifests) (if ExoPlayer supports them) (thanks to @Redirion for this).

- Fixes #2415
- Fixes #6860 (some audio streams were recognized as video streams causing a crash in `ListHelper.compareVideoStreamResolution(String res1, String res2)` which is caught in this PR. The original error cause was also fixed in the extractor.)

- TeamNewPipe/NewPipeExtractor#663

#### APK testing 

#### Due diligence


TiA4f8R:
I don't know how to fix the crash when pressing the Download button and how to add support of HLS "raw" manifests. Can you help me please? Thank you in advance.

wb9688:
For HLS there is [`HlsPlaylistParser`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.html) in ExoPlayer, which is similar to `DashManifestParser`.

TiA4f8R:
> For HLS there is [`HlsPlaylistParser`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistParser.html) in ExoPlayer, which is similar to `DashManifestParser`.

I saw this, but I am not able to build an HLS media source with an HlsPlaylist object because no constructor of createMediaSource with an HlsPlaylist object exists. Here is my code:

```java
            if (stream.isUrl()) {
                return dataSource.getHlsMediaSourceFactory().setTag(metadata)
                        .createMediaSource(MediaItem.fromUri(Uri.parse(stream.getContent())));
            } else {
                final HlsPlaylist hlsPlaylist;
                try {
                    final ByteArrayInputStream hlsManifestInput = new ByteArrayInputStream(
                            stream.getContent().getBytes(StandardCharsets.UTF_8));
                    hlsPlaylist = new HlsPlaylistParser().parse(Uri.parse(stream.getBaseUrl()),
                            hlsManifestInput);
                } catch (final IOException e) {
                    throw new IOException("Error when parsing manual HLS manifest", e);
                }

                return dataSource.getHlsMediaSourceFactory().setTag(metadata)
                        .createMediaSource(hlsPlaylist);
            }
```
Does anyone has an idea?

TiA4f8R:
I don't know how to fix the crash when pressing the Download button on videos which have OTF streams. Here is the crash report:

## Exception
* __User Action:__ ui error
* __Request:__ ACRA report
* __Content Country:__ FR
* __Content Language:__ fr-FR
* __App Language:__ fr_FR
* __Service:__ none
* __Version:__ 0.21.5

```
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
	at java.util.ArrayList.get(ArrayList.java:437)
	at org.schabi.newpipe.util.StreamItemAdapter.getItem(StreamItemAdapter.java:76)
	at org.schabi.newpipe.util.StreamItemAdapter.getCustomView(StreamItemAdapter.java:109)
	at org.schabi.newpipe.util.StreamItemAdapter.getView(StreamItemAdapter.java:92)
	at android.widget.Spinner.makeView(Spinner.java:729)
	at android.widget.Spinner.getBaseline(Spinner.java:567)
	at android.widget.RelativeLayout.onMeasure(RelativeLayout.java:498)
	at android.view.View.measure(View.java:24742)
	at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6903)
	at android.widget.FrameLayout.onMeasure(FrameLayout.java:194)
	at android.view.View.measure(View.java:24742)
	at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6903)
	at android.widget.LinearLayout.measureChildBeforeLayout(LinearLayout.java:1562)
	at android.widget.LinearLayout.measureVertical(LinearLayout.java:849)
	at android.widget.LinearLayout.onMeasure(LinearLayout.java:728)
	at android.view.View.measure(View.java:24742)
	at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6903)
	at android.widget.FrameLayout.onMeasure(FrameLayout.java:194)
	at com.android.internal.policy.DecorView.onMeasure(DecorView.java:896)
	at android.view.View.measure(View.java:24742)
	at android.view.ViewRootImpl.performMeasure(ViewRootImpl.java:3208)
	at android.view.ViewRootImpl.measureHierarchy(ViewRootImpl.java:1944)
	at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:2292)
	at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1854)
	at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:8086)
	at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1057)
	at android.view.Choreographer.doCallbacks(Choreographer.java:875)
	at android.view.Choreographer.doFrame(Choreographer.java:776)
	at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:1042)
	at android.os.Handler.handleCallback(Handler.java:888)
	at android.os.Looper.loop(Looper.java:213)
	at android.app.ActivityThread.main(ActivityThread.java:8178)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:513)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1101)

```
</details>
<hr>

Can someone help me?

opusforlife2:
This breaks PeerTube playback completely. All videos show the "Could not play this stream" toast.

opusforlife2:
How is this implemented? Does it look for progressive streams first? Then DASH if those aren't available?

For videos where only one or two progressive streams are available, like 720p and 360p MP4, does it replace those resolutions with their DASH versions as well?

TiA4f8R:
> This breaks PeerTube playback completely. All videos show the "Could not play this stream" toast.

The app tries to play torrent streams. I will investigate why.

> How is this implemented? Does it look for progressive streams first? Then DASH if those aren't available?

A DASH manifest is only returned by YouTube when OTF streams are present.

> For videos where only one or two progressive streams are available, like 720p and 360p MP4, does it replace those resolutions with their DASH versions as well?

Only for the 1080p30/60 stream (and the 720p30 stream for cipher protected URLs). Other streams will use DASH versions.

opusforlife2:
> A DASH manifest is only returned by YouTube when OTF streams are present.

Still, does this PR prefer DASH or progressive? Or are progressive streams absent when OTF ones are present for all Youtube videos?

> Only for the 1080p30/60 stream (and the 720p30 stream for cipher protected URLs). Other streams will use DASH versions.

So do you try to replace all streams with DASH versions and only show progressive ones where DASH streams aren't available?

TiA4f8R:
> Still, does this PR prefer DASH or progressive? Or are progressive streams absent when OTF ones are present for all Youtube videos?
>
> So do you try to replace all streams with DASH versions and only show progressive ones where DASH streams aren't available?

I didn't change anything about the current implementation in the app but I think I need to, because NewPipe wants to play torrent streams first instead of progressive streams.

opusforlife2:
Oh, I'm asking specifically in the context of Youtube.

TiA4f8R:
Yes, I know but I don't know right now how the app manages which streams are chosen among the streams returned by the extractor.

TiA4f8R:
On long livestream replays, I got this crash after leaving the app or closing the player or going into settings (my device has 4GB of RAM):

## Exception
* __User Action:__ ui error
* __Request:__ ACRA report
* __Content Country:__ FR
* __Content Language:__ fr-FR
* __App Language:__ fr_FR
* __Service:__ none
* __Version:__ 0.21.6

```
java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 1383248 bytes
	at android.app.servertransaction.PendingTransactionActions$StopInfo.run(PendingTransactionActions.java:161)
	at android.os.Handler.handleCallback(Handler.java:888)
	at android.os.Looper.loop(Looper.java:213)
	at android.app.ActivityThread.main(ActivityThread.java:8178)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:513)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1101)
Caused by: android.os.TransactionTooLargeException: data parcel size 1383248 bytes
	at android.os.BinderProxy.transactNative(Native Method)
	at android.os.BinderProxy.transact(BinderProxy.java:526)
	at android.app.IActivityTaskManager$Stub$Proxy.activityStopped(IActivityTaskManager.java:4561)
	at android.app.servertransaction.PendingTransactionActions$StopInfo.run(PendingTransactionActions.java:145)
	... 7 more

```
</details>
<hr>

How to fix this bug?

TiA4f8R:
@opusforlife2 Can you confirm that PeerTube streams are playable for you and that for the streams that have only HLS variants, you can download these streams and seek properly within them?

opusforlife2:
@TiA4f8R Finally! Yes, PeerTube streams are playable, and seeking works perfectly well within them.

Downloading them works fine (apart from the "some streams hidden" toast), but seeking within the downloaded streams doesn't. It varies from app to app, actually:
- VLC can't seek at all.
- FX File Explorer's media player shows a seekbar with a really tiny duration, like 4-5 seconds, but you can drag on the thumb to seek blindly.
- mpv-android is the only one where the correct duration is shown, and seeking works properly. I assume it is making the calculations itself, then, and not relying on Android's Media Store.

opusforlife2:
> seeking within the downloaded streams doesn't

@kapodamy! We need your help!

kapodamy:
somenoe link the apk or the downloaded stream

TiA4f8R:
@opusforlife2 Please answer with a link of an HLS-only stream on which you have this issue.

opusforlife2:
@kapodamy @TiA4f8R Most videos on the Tilvids instance are HLS-only. For example: https://tilvids.com/videos/watch/cc6137e8-284c-4ab8-a0c1-eb10da13a440

kapodamy:
@opusforlife2  oh the HLS format... thats gonna be a nigthmare download the segments at later concat them. sorry but ill pass

opusforlife2:
Oh no. :(

What will we do now?

Stypox:
I can confirm it seems to work, I tried a few videos and resolutions both on PeerTube and YouTube.

> I think it will be better to generate manifests when the quality is selectioned instead of generate them in the video loading (after fetching the "page" of the content).

@TiA4f8R good idea, it's bad to generate everything even when unneeded

TiA4f8R:
> @TiA4f8R good idea, it's bad to generate everything even when unneeded

@Stypox I made the corresponding changes in the app but the manifest is generated at each time the quality is requested (also after switching to another app and then going back to NewPipe or changing player type) (and so the first segment and its redirections if there are any are fetched).
So is there a way to cache the result, until the user closes the app (and maybe until when they change their network connection, for redirections purposes), in order to reduce the energy consumption?

I don't have the knowledge to do this.

Stypox:
> I don't have the knowledge to do the part of the network state check.

A method should already be available in NewPipe's utils; if not just look it up online, it should be straightforward

TiA4f8R:
>  oh the HLS format... thats gonna be a nigthmare download the segments at later concat them. sorry but ill pass

@kapodamy The stream used in HLS playlists is a single video file. Only range requests of these streams are requested, not different video files. For the video which @opusforlife2 posted, here are the direct video links extracted by NewPipe Extractor:

`1080p`: https://tilvids.com/download/streaming-playlists/hls/videos/cc6137e8-284c-4ab8-a0c1-eb10da13a440-1080-fragmented.mp4

`360p`: https://tilvids.com/download/streaming-playlists/hls/videos/cc6137e8-284c-4ab8-a0c1-eb10da13a440-360-fragmented.mp4

So maybe you can do something.

kapodamy:
@TiA4f8R multiple "Segment Indexing Box" (`sidx` box) is not supported, this will require some modifications in the muxer. I'll see later how I do it
aaa

TiA4f8R:
@kapodamy Thank you for taking care of this!

Redirion:
@TiA4f8R 

in PlayerdataSource.java create this method:

  `public HlsMediaSource.Factory getHlsMediaSourceFactory(HlsPlaylistParserFactory playlistParserFactory) {
          return new HlsMediaSource.Factory(cacheDataSourceFactory).setPlaylistParserFactory(playlistParserFactory);
      }`

add this class to the same package as the PlaybackResolver.java

`package org.schabi.newpipe.player.resolver;

import android.net.Uri;

import androidx.annotation.Nullable;

import com.google.android.exoplayer2.source.hls.playlist.HlsMasterPlaylist;
import com.google.android.exoplayer2.source.hls.playlist.HlsMediaPlaylist;
import com.google.android.exoplayer2.source.hls.playlist.HlsPlaylist;
import com.google.android.exoplayer2.source.hls.playlist.HlsPlaylistParserFactory;
import com.google.android.exoplayer2.upstream.ParsingLoadable;

import java.io.IOException;
import java.io.InputStream;

public class NewPipeHlsPlaylistParserFactory implements HlsPlaylistParserFactory {

    private final HlsPlaylist hlsPlaylist;

    public NewPipeHlsPlaylistParserFactory(HlsPlaylist hlsPlaylist) {
        this.hlsPlaylist = hlsPlaylist;
    }

    private class NewPipeHlsPlayListParser implements ParsingLoadable.Parser<HlsPlaylist> {

        @Override
        public HlsPlaylist parse(Uri uri, InputStream inputStream) throws IOException {
            return hlsPlaylist;
        }
    }

    @Override
    public ParsingLoadable.Parser<HlsPlaylist> createPlaylistParser() {
        return new NewPipeHlsPlayListParser();
    }

    @Override
    public ParsingLoadable.Parser<HlsPlaylist> createPlaylistParser(HlsMasterPlaylist masterPlaylist, @Nullable HlsMediaPlaylist previousMediaPlaylist) {
        return new NewPipeHlsPlayListParser();
    }
}
`


and finally, instead of calling ` return dataSource.getHlsMediaSourceFactory().setTag(metadata)
                        .createMediaSource(hlsPlaylist);`

use ` return dataSource.getHlsMediaSourceFactory(new NewPipeHlsPlaylistParserFactory(hlsPlayList)).setTag(metadata)
                        .createMediaSource(MediaItem.fromUri(uri));`

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -2078,25 +2092,36 @@ private void showClearingQueueConfirmation(final Runnable onAllow) {
...
+        final List<VideoStream> videoOnlyStreams = removeNonUrlAndTorrentStreams(
+                currentInfo.getVideoOnlyStreams());
+        final List<VideoStream> sortedVideoStreams = ListHelper.getSortedStreamVideosList(
+                activity, videoStreams, videoOnlyStreams, false);
+        final AlertDialog.Builder builder = new AlertDialog.Builder(activity);
+        builder.setTitle(R.string.select_quality_external_player);
+        builder.setNegativeButton(android.R.string.cancel, null);
+        builder.setNeutralButton(R.string.open_in_browser, (dialog, i) ->
+                ShareUtils.openUrlInBrowser(requireActivity(), url));
+        if (!sortedVideoStreams.isEmpty()) {
```
TobiGr, line 2109: Inverting this statement makes it easier to read

TobiGr:
This PR also closes #6860. I added a little information to the PR description.

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -578,16 +583,51 @@ private void openDownloadDialog() {
...
+                            audioStreams,
+                            subtitlesStreams,
+                            progressiveHttpVideoStreams,
+                            progressiveHttpVideoOnlyStreams,
+                            progressiveHttpAudioStreams,
+                            progressiveHttpSubtitlesStreams)) {
+                        downloadDialog.setNonProgressiveStreamsRemoved();
+                    }
+
                     downloadDialog.setSelectedVideoStream(selectedVideoStreamIndex);
```
TobiGr, line 631: please use `newInstance(Context context, StreamInfo info)`

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
```diff
@@ -51,36 +68,164 @@ default MediaSource buildLiveMediaSource(@NonNull final PlayerDataSource dataSou
...
+                    throw new IOException("Error when parsing manual DASH manifest", e);
+                }
+
+                return dataSource.getDashMediaSourceFactory().setTag(metadata)
+                        .createMediaSource(dashManifest);
+            }
+        } else {
+            throw new IllegalArgumentException("Unsupported delivery type" + deliveryMethod);
+        }
+    }
```
TobiGr, line 193: This method is long and deeply nested. Please split it up into multiple methods.

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
```diff
@@ -51,36 +68,164 @@ default MediaSource buildLiveMediaSource(@NonNull final PlayerDataSource dataSou
...
+    default void removeTorrentAndNonUrlStreams(@NonNull final List<? extends Stream> streamList) {
+        if (streamList.isEmpty()) {
+            return;
+        }
+        final Iterator<? extends Stream> streamIterator = streamList.iterator();
+        while (streamIterator.hasNext()) {
+            final Stream stream = streamIterator.next();
+            if (stream.getDeliveryMethod() == DeliveryMethod.TORRENT || !stream.isUrl()) {
+                streamIterator.remove();
+            }
```
TobiGr, line 228: These methods are similar to the ones in `ListHelper`. Why are both versions needed? I think we can keep those in `ListHelper`.

TobiGr (commented):
Looks mostly good!

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
TiA4f8R: I will split this method into three methods: one for DASH, one for progressive HTTP and one for HLS.

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
TiA4f8R: I removed this version.

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
TiA4f8R: The `buildMediaSource` method is now split into three methods, and the DASH method itself is split into three methods.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
TiA4f8R: Done

TobiGr (commented):
The instace creation without the context in DownloadDialog can be removed. It is only used in the other `newInstance` method.

app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
```diff
@@ -172,40 +205,66 @@ private void setInfo(final StreamInfo info) {
...
 
+    private void showIfStreamsWereRemovedMessage() {
+        dialogBinding.streamsHidden.setVisibility(this.nonProgressiveStreamsRemoved ? View.VISIBLE
+                : View.GONE);
+    }
+
+    public void setNonProgressiveStreamsRemoved() {
+        if (!this.nonProgressiveStreamsRemoved) {
+            this.nonProgressiveStreamsRemoved = true;
+        }
```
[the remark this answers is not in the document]
TiA4f8R: I think this also apply for energy consumption. I removed this check.

Stypox:
Shall I review? What is still left to do so that this PR exits the draft stage? @TiA4f8R

TiA4f8R:
@Stypox There is currently a big issue that needs to be fixed: the manifests generated with `YoutubeDashManifestCreator` are never cleared when the app is running so the cache will grow infinitely.
This must be fixed before this PR can be merged (it will be great if we can removed the cache of manifests when a network change happen to reduce some redirects that may happen next time the same source is loaded), otherwise, it will cause some issues.

The text `Note that some streams might have been hidden [...]` in the DownloadDialog needs to be shown only when there are streams which are really hidden (on Peertube, the message is shown because the app relies on the number of streams and not the resolution and the media format) and this text is not shown anymore after rotating your device.

However, I don't have the time to do this right now so if you have some time, free feel to do it yourself :)

I have also a bit of work to do in the manifest creator of the extractor side of the changes (and I need also update mocks and the test class I created).

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
```diff
@@ -62,28 +56,26 @@ public PlayerDataSource(@NonNull final Context context, @NonNull final String us
...
-            return new DefaultDashChunkSource.Factory(
-                    MediaParserChunkExtractor.FACTORY,
-                    dataSourceFactory,
-                    1
-            );
+            return new DefaultDashChunkSource.Factory(MediaParserChunkExtractor.FACTORY,
+                    dataSourceFactory, 1);
         }
 
         return new DefaultDashChunkSource.Factory(dataSourceFactory);
```
Stypox, line 70:
```suggestion
        } else {
            new DefaultDashChunkSource.Factory(dataSourceFactory);
        }
```

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
```diff
@@ -62,28 +56,26 @@ public PlayerDataSource(@NonNull final Context context, @NonNull final String us
...
-    public HlsMediaSource.Factory getHlsMediaSourceFactory() {
+    public HlsMediaSource.Factory getHlsMediaSourceFactory(
+            @Nullable final HlsPlaylistParserFactory hlsPlaylistParserFactory) {
         final HlsMediaSource.Factory factory = new HlsMediaSource.Factory(cacheDataSourceFactory);
+        if (hlsPlaylistParserFactory != null) {
+            factory.setPlaylistParserFactory(hlsPlaylistParserFactory);
+        }
 
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
             return factory;
```
Stypox, line 81: Same below here, create the else branch

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -163,13 +169,39 @@ public static String resizeTypeOf(@NonNull final Context context,
...
     public static String cacheKeyOf(@NonNull final StreamInfo info,
                                     @NonNull final VideoStream video) {
-        return info.getUrl() + video.getResolution() + video.getFormat().getName();
+        String cacheKey = info.getUrl();
+        final String resolution = video.getResolution();
+        if (!resolution.equals(RESOLUTION_UNKNOWN)) {
+            cacheKey += resolution;
+        } else {
+            cacheKey += String.valueOf(cachedVideoStreamsWithResolutionUnknown);
+            cachedVideoStreamsWithResolutionUnknown += 1;
```
Stypox, line 178: It doesn't seem like a good idea to me to use a static counter to differentiate between different cached video/audio streams. Isn't there any other information that the extractor could provide that allows differentiating between streams? Otherwise it's just better to disable cache for these streams, which is probably what you are trying to achieve by making a different key each time, but this way the cache is not really disabled for these streams.

Stypox (changes requested):
(partial review, will finish later)

app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
```diff
@@ -172,40 +199,88 @@ private void setInfo(final StreamInfo info) {
     }
 
     public void setAudioStreams(final List<AudioStream> audioStreams) {
+        removeDeliveryDistinctStreams(audioStreams, DeliveryMethod.PROGRESSIVE_HTTP);
         setAudioStreams(new StreamSizeWrapper<>(audioStreams, getContext()));
     }
 
     public void setAudioStreams(final StreamSizeWrapper<AudioStream> was) {
         this.wrappedAudioStreams = was;
```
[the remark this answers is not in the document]
TiA4f8R: Actually, this method returns only streams of the delivery method passed as the second parameter.

Here this is intended, as the downloader only supports progressive streams.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1709,8 +1720,11 @@ private void updateProgressInfo(@NonNull final StreamInfo info) {
                 binding.detailPositionView.setVisibility(View.GONE);
                 // TODO: Remove this check when separation of concerns is done.
                 //  (live streams weren't getting updated because they are mixed)
-                if (!info.getStreamType().equals(StreamType.LIVE_STREAM)
-                        && !info.getStreamType().equals(StreamType.AUDIO_LIVE_STREAM)) {
+                final StreamType streamType = info.getStreamType();
+                if (!streamType.equals(StreamType.LIVE_STREAM)
+                        && !streamType.equals(StreamType.AUDIO_LIVE_STREAM)
+                        && !streamType.equals(StreamType.POST_LIVE_STREAM)
+                        && !streamType.equals(StreamType.POST_LIVE_AUDIO_STREAM)) {
```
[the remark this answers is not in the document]
TiA4f8R: This was probably a rebase issue (rebasing this PR is really annoying for me so I may have kept old changes instead of new ones).

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1631,8 +1642,11 @@ public void handleResult(@NonNull final StreamInfo info) {
             }
         }
 
-        binding.detailControlsDownload.setVisibility(info.getStreamType() == StreamType.LIVE_STREAM
-                || info.getStreamType() == StreamType.AUDIO_LIVE_STREAM ? View.GONE : View.VISIBLE);
+        binding.detailControlsDownload.setVisibility(streamType.equals(StreamType.LIVE_STREAM)
+                || streamType.equals(StreamType.AUDIO_LIVE_STREAM)
+                || streamType.equals(StreamType.POST_LIVE_STREAM)
+                || streamType.equals(StreamType.POST_LIVE_AUDIO_STREAM)
```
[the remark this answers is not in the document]
TiA4f8R: Yes, you are right. Same explanation as above.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
TiA4f8R: This is probably outdated / non-applicable because of #7555. I will check after I rebase the branch when I get the conflict part.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
TiA4f8R: This is probably outdated / non-applicable because of #7555. I will check after I rebase the branch when I get the conflict part.

app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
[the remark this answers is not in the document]
Stypox: Yeah, sorry, I meant:

Why in this case the **non-**`PROGRESSIVE_HTTP` streams are not removed?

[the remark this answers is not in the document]
TiA4f8R: Because this method is only called by `setAudioStreams(List<AudioStream> audioStreams)`. That's the same behavior with the corresponding video and subtitle methods.

By finding this, I also found that the `removeDeliveryDistinctStreams` method is called twice, for audio, video and subtitle streams. I will change this.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
TiA4f8R: Yes, I am agree with you.

What do you think of this code instead?

- for video streams:
  ```java
  @NonNull
  public static String cacheKeyOf(@NonNull final StreamInfo info,
                                  @NonNull final VideoStream videoStream) {
      String cacheKey = info.getUrl() + " " + videoStream.getId();

      final String resolution = videoStream.getResolution();
      final MediaFormat mediaFormat = videoStream.getFormat();
      if (resolution.equals(RESOLUTION_UNKNOWN) && mediaFormat == null) {
          final String content = videoStream.getContent();
          cacheKey += " " + content.substring(20 * content.length() / 100);
      } else {
          if (mediaFormat != null) {
              cacheKey += " " + videoStream.getFormat().getName();
          }
          if (!resolution.equals(RESOLUTION_UNKNOWN)) {
              cacheKey += " " + resolution;
          }
      }

      return cacheKey;
  }
  ```

- for audio streams:
  ```java
  @NonNull
  public static String cacheKeyOf(@NonNull final StreamInfo info,
                                  @NonNull final AudioStream audioStream) {
      String cacheKey = info.getUrl() + " " + audioStream.getId();
  
      final int averageBitrate = audioStream.getAverageBitrate();
      final MediaFormat mediaFormat = audioStream.getFormat();
      if (averageBitrate == UNKNOWN_BITRATE && mediaFormat == null) {
          final String content = audioStream.getContent();
          cacheKey += " " + content.substring(20 * content.length() / 100);
      } else {
          if (mediaFormat != null) {
              cacheKey += " " + audioStream.getFormat().getName();
          }
          if (averageBitrate != UNKNOWN_BITRATE) {
              cacheKey += " " + averageBitrate;
          }
      }
  
      return cacheKey;
  }
  ```

app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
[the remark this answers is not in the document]
TiA4f8R: Done

app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
```diff
@@ -172,40 +199,88 @@ private void setInfo(final StreamInfo info) {
     }
 
     public void setAudioStreams(final List<AudioStream> audioStreams) {
+        removeDeliveryDistinctStreams(audioStreams, DeliveryMethod.PROGRESSIVE_HTTP);
         setAudioStreams(new StreamSizeWrapper<>(audioStreams, getContext()));
```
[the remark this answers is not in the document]
TiA4f8R: With my changes, this review comment is outdated.

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -104,6 +108,86 @@ public static int getDefaultAudioFormat(final Context context,
...
+     * Return a {@link Stream} list which uses the given delivery method from a {@link Stream}
+     * list.
+     *
+     * @param streamList     the original stream list
+     * @param deliveryMethod the delivery method
+     * @param <S>            the item type's class that extends {@link Stream}
+     * @return a stream list which uses the given delivery method
+     */
+    @NonNull
+    public static <S extends Stream> List<S> removeDeliveryDistinctStreams(
```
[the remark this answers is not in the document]
TiA4f8R: > Also, maybe I'd call this `keepStreamsWithDelivery`

Yes, agree with you, this name was given by Redirion.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -162,14 +169,46 @@ public static String resizeTypeOf(@NonNull final Context context,
...
-                                    @NonNull final VideoStream video) {
-        return info.getUrl() + video.getResolution() + video.getFormat().getName();
+                                    @NonNull final VideoStream videoStream) {
+        String cacheKey = info.getUrl() + " " + videoStream.getId();
+
+        final String resolution = videoStream.getResolution();
+        final MediaFormat mediaFormat = videoStream.getFormat();
+        if (resolution.equals(RESOLUTION_UNKNOWN) && mediaFormat == null) {
+            final String content = videoStream.getContent();
+            cacheKey += " " + content.substring(20 * content.length() / 100);
```
Stypox, line 179: I'm not so sure about this way of creating a key: the first 20% of the content might be equal for some streams, wouldn't it? E.g. for videos with normal urls, the first 20% characters might always be the same. What about using the hash of the string? It wouldn't ensure uniqueness, but the probability of clashes should be pretty low.

app/src/main/java/org/schabi/newpipe/player/resolver/AudioPlaybackResolver.java
```diff
@@ -1,20 +1,26 @@
...
 import org.schabi.newpipe.player.helper.PlayerDataSource;
 import org.schabi.newpipe.player.helper.PlayerHelper;
 import org.schabi.newpipe.util.ListHelper;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
 public class AudioPlaybackResolver implements PlaybackResolver {
+    private static final String TAG = "AudioPlaybackResolver";
```
Stypox, line 22:
```suggestion
    private static final String TAG = AudioPlaybackResolver.class.getSimpleName();
```

app/src/main/java/org/schabi/newpipe/player/resolver/AudioPlaybackResolver.java
```diff
@@ -34,14 +40,23 @@ public MediaSource resolve(@NonNull final StreamInfo info) {
...
-        return buildMediaSource(dataSource, audio.getUrl(), PlayerHelper.cacheKeyOf(info, audio),
-                MediaFormat.getSuffixById(audio.getFormatId()), tag);
+        MediaSource mediaSource = null;
+        try {
+            mediaSource = buildMediaSource(dataSource, audio, info, PlayerHelper.cacheKeyOf(info,
+                    audio), tag);
+        } catch (final IOException e) {
+            Log.e(TAG, "Unable to create audio source:", e);
+        }
+        return mediaSource;
```
Stypox, line 60:
```suggestion
        try {
            return buildMediaSource(dataSource, audio, info, PlayerHelper.cacheKeyOf(info, audio), tag);
        } catch (final IOException e) {
            Log.e(TAG, "Unable to create audio source:", e);
            return null;
        }
```

app/src/main/java/org/schabi/newpipe/player/resolver/NewPipeHlsPlaylistParserFactory.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public ParsingLoadable.Parser<HlsPlaylist> createPlaylistParser() {
+        return new NewPipeHlsPlayListParser();
+    }
+
+    @NonNull
+    @Override
+    public ParsingLoadable.Parser<HlsPlaylist> createPlaylistParser(
+            @NonNull final HlsMasterPlaylist masterPlaylist,
+            @Nullable final HlsMediaPlaylist previousMediaPlaylist) {
```
Stypox, line 44: (not really a problem) Why does exoplayer pass in these parameters? May they be useful?

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
```diff
@@ -68,38 +89,217 @@ default MediaSource buildLiveMediaSource(@NonNull final PlayerDataSource dataSou
...
-            default:
-                throw new IllegalStateException("Unsupported type: " + type);
+    default <T extends Stream> ProgressiveMediaSource buildProgressiveMediaSource(
+            @NonNull final PlayerDataSource dataSource,
+            @NonNull final T stream,
+            @NonNull final String cacheKey,
+            @NonNull final MediaSourceTag metadata) throws IOException {
+        final String url = stream.getContent();
+
+        if (!isNullOrEmpty(url)) {
```
Stypox, line 116: Swap the brances and remove `!`

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
```diff
@@ -68,38 +89,217 @@ default MediaSource buildLiveMediaSource(@NonNull final PlayerDataSource dataSou
...
+        }
+    }
+
+    /**
+     * Remove streams which are using the {@link DeliveryMethod#TORRENT torrent delivery method}.
+     *
+     * @param streamList the list of {@link Stream streams} for which you want to delete the
+     *                   torrent streams.
+     */
+    default void removeTorrentStreams(@NonNull final List<? extends Stream> streamList) {
```
Stypox, line 293: Why is this here and not in `ListHelper` like the others? And why is it not a `static` method?

app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java
```diff
@@ -19,12 +20,16 @@
...
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
 import static com.google.android.exoplayer2.C.TIME_UNSET;
+import static org.schabi.newpipe.util.ListHelper.removeNonUrlAndTorrentStreams;
 
 public class VideoPlaybackResolver implements PlaybackResolver {
+    private static final String TAG = "VideoPlaybackResolver";
```
Stypox, line 31: Also here

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -482,12 +541,20 @@ private static int compareAudioStreamBitrate(final AudioStream streamA,
...
+    private static int compareVideoStreamResolution(@NonNull final String r1,
+                                                    @NonNull final String r2) {
+        try {
+            final int res1 = Integer.parseInt(r1.replaceAll("0p\\d+$", "1")
+                    .replaceAll("[^\\d.]", ""));
+            final int res2 = Integer.parseInt(r2.replaceAll("0p\\d+$", "1")
+                    .replaceAll("[^\\d.]", ""));
+            return res1 - res2;
+        } catch (final NumberFormatException e) {
+            // Return 1 because we don't know if the two streams are different or not (a
```
Stypox, line 553:
```suggestion
            // Consider the first one greater because we don't know if the two streams are different or not (a
```

app/src/main/java/org/schabi/newpipe/util/SecondaryStreamHelper.java
```diff
@@ -29,33 +31,37 @@ public SecondaryStreamHelper(final StreamSizeWrapper<T> streams, final T selecte
...
-            case MPEG_4:// ¿is mpeg-4 DASH?
-                break;
-            default:
-                return null;
-        }
+        final MediaFormat mediaFormat = videoStream.getFormat();
+        if (mediaFormat != null) {
+            switch (mediaFormat) {
+                case WEBM:
+                case MPEG_4:// ¿is mpeg-4 DASH?
```
Stypox, line 41: I guess this should be resolved, shouldn't it?

app/src/main/java/org/schabi/newpipe/util/SecondaryStreamHelper.java
```diff
@@ -29,33 +31,37 @@ public SecondaryStreamHelper(final StreamSizeWrapper<T> streams, final T selecte
...
+                    break;
+                default:
+                    return null;
+            }
 
-        final boolean m4v = videoStream.getFormat() == MediaFormat.MPEG_4;
+            final boolean m4v = mediaFormat == MediaFormat.MPEG_4;
```
Stypox, line 47:
```suggestion
            final boolean m4v = (mediaFormat == MediaFormat.MPEG_4);
```

app/src/main/java/org/schabi/newpipe/util/SecondaryStreamHelper.java
```diff
@@ -29,33 +31,37 @@ public SecondaryStreamHelper(final StreamSizeWrapper<T> streams, final T selecte
...
+            if (m4v) {
+                return null;
+            }
 
-        // retry, but this time in reverse order
-        for (int i = audioStreams.size() - 1; i >= 0; i--) {
-            final AudioStream audio = audioStreams.get(i);
-            if (audio.getFormat() == MediaFormat.WEBMA_OPUS) {
-                return audio;
+            // retry, but this time in reverse order
```
Stypox, line 59: What? Why retry and why in reverse order?

app/src/main/java/org/schabi/newpipe/util/StreamItemAdapter.java
```diff
@@ -154,14 +167,15 @@ private View getCustomView(final int position, final View view, final ViewGroup
...
-                    break;
-                default:
-                    formatNameView.setText(stream.getFormat().getName());
-                    break;
+                } else {
+                    formatNameView.setText(mediaFormat.getName());
+                }
+            } else {
+                formatNameView.setText(context.getString(R.string.unknown_format));
             }
```
Stypox, line 179:
```suggestion
if (mediaFormat == null) {
    formatNameView.setText(context.getString(R.string.unknown_format));
} else if (mediaFormat == MediaFormat.WEBMA_OPUS) {
    // noinspection AndroidLintSetTextI18n
    formatNameView.setText("opus");
} else {
    formatNameView.setText(mediaFormat.getName());
}
```

app/src/main/java/org/schabi/newpipe/util/StreamItemAdapter.java
```diff
@@ -154,14 +167,15 @@ private View getCustomView(final int position, final View view, final ViewGroup
         if (stream instanceof SubtitlesStream) {
             formatNameView.setText(((SubtitlesStream) stream).getLanguageTag());
         } else {
-            switch (stream.getFormat()) {
-                case WEBMA_OPUS:
+            if (mediaFormat != null) {
+                if (mediaFormat == MediaFormat.WEBMA_OPUS) {
                     // noinspection AndroidLintSetTextI18n
                     formatNameView.setText("opus");
```
Stypox, line 173: Well, maybe it would make sense to create a string

app/src/main/java/us/shandian/giga/get/DownloadMissionRecover.java
```diff
@@ -131,31 +132,38 @@ private void resolveStream() throws IOException, ExtractionException, HttpError
 
         switch (mRecovery.getKind()) {
```
Stypox, line 133: Have you checked if it's ok for this method to return `null` (either an exception is manually thrown or the case is handled, by the caller)?

app/src/main/res/layout/download_dialog.xml
```diff
@@ -106,4 +107,16 @@
             android:max="31"
             android:progress="3" />
     </LinearLayout>
+
+    <org.schabi.newpipe.views.NewPipeTextView
```
Stypox, line 111: You're not using this anymore if I understand correctly, then remove it (and also the string)

Stypox (changes requested):
General structure looks fine, though I'm not an expert of the player. I didn't test.
Thank you :-)

Please refrain from reformatting things when doing such big PRs, it makes reviewing more cumbersome

app/src/main/res/layout/download_dialog.xml
TiA4f8R: > You're not using this anymore if I understand correctly

No, you didn't understand correctly :)

This message is always shown in the download dialog, even when there are no streams removed (the previous message was shown even when some streams were removed but at least one stream for each original resolution was still available for download).

app/src/main/java/org/schabi/newpipe/util/SecondaryStreamHelper.java
TiA4f8R: This isn't my code, I just reformatted it.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
TiA4f8R: I didn't have an idea to how use the content, but your idea of using the hash code is really great! Why not using the hash code of the stream object itself instead of using the hashcode of the content?

app/src/main/java/org/schabi/newpipe/util/SecondaryStreamHelper.java
TiA4f8R: This isn't my code, I just reformatted it.

app/src/main/java/org/schabi/newpipe/util/StreamItemAdapter.java
TiA4f8R: Probably, don't forget that it's not my code, I just reformatted it.

app/src/main/java/org/schabi/newpipe/util/SecondaryStreamHelper.java
kapodamy: thats my comment

app/src/main/java/us/shandian/giga/get/DownloadMissionRecover.java
TiA4f8R: That's not my code, I don't think we should change it in this PR.

app/src/main/java/org/schabi/newpipe/util/SecondaryStreamHelper.java
opusforlife2: Yo, kapodamy. Any luck with the multiple sidx boxes?

TiA4f8R: (Please this move this discussion outside this review comment and note the relevant code will not embed to this PR.)

opusforlife2:
@kapodamy Any luck with the multiple sidx boxes?

app/src/main/res/layout/download_dialog.xml
Stypox: Oh, ok, makes sense

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
Stypox: Yeah, that makes sense, though the `Stream` class does not seem to have `hashCode` overridden, and the default implementation obviously [does not return something useful](https://stackoverflow.com/questions/2237720/what-is-an-objects-hash-code-if-hashcode-is-not-overridden). So either you make changes in the extractor and override the `hashCode()` method in the various `*Stream` classes, or you just use `getContent().hashCode()` and call it a day. (obviously only use hashes if `resolution.equals(RESOLUTION_UNKNOWN) && mediaFormat == null`)

TiA4f8R: > Override the hashCode() method in the various *Stream classes

What do you suggest me to use to generate the hash code?

Stypox: See https://github.com/TeamNewPipe/NewPipeExtractor/pull/663#discussion_r790264754

app/src/main/java/org/schabi/newpipe/player/resolver/NewPipeHlsPlaylistParserFactory.java
TiA4f8R: I don't really know myself but here is what ExoPlayer documentation says for this method (in `HlsPlaylistParserFactory` interface):

```java
  /**
   * Returns a playlist parser for playlists that were referenced by the given {@link
   * HlsMasterPlaylist}. Returned {@link HlsMediaPlaylist} instances may inherit attributes from
   * {@code masterPlaylist}.
   *
   * @param masterPlaylist The master playlist that referenced any parsed media playlists.
   * @param previousMediaPlaylist The previous media playlist or null if there is no previous media
   *     playlist.
   * @return A parser for HLS playlists.
   */
```

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -162,14 +169,45 @@ public static String resizeTypeOf(@NonNull final Context context,
...
```
Stypox, line 179: You forgot to change this I think ;-)
```suggestion
            cacheKey += " " + videoStream.hashCode();
```

I'd also add a comment in both `cacheKeyOf` functions explaining that the hash code is only used if the other information is not usable, since the other information is better at creating a cache key.

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
```diff
@@ -68,38 +87,198 @@ default MediaSource buildLiveMediaSource(@NonNull final PlayerDataSource dataSou
 
     @NonNull
     default MediaSource buildMediaSource(@NonNull final PlayerDataSource dataSource,
```
Stypox, line 89: Remove `default` from all of these methods and make them `static` instead:
- they are methods of **an interface**, so they can't (and don't) do anything with the object on which they are called (since an interface has no fields / data)
- `default` methods are meant to be overridden, but none of these methods are overridden anywhere
- these methods do not do anything with `this`, i.e. they are already "static"

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -482,12 +566,20 @@ private static int compareAudioStreamBitrate(final AudioStream streamA,
...
+            final int res1 = Integer.parseInt(r1.replaceAll("0p\\d+$", "1")
+                    .replaceAll("[^\\d.]", ""));
+            final int res2 = Integer.parseInt(r2.replaceAll("0p\\d+$", "1")
+                    .replaceAll("[^\\d.]", ""));
+            return res1 - res2;
+        } catch (final NumberFormatException e) {
+            // Consider the first one greater because we don't know if the two streams are
+            // different or not (a NumberFormatException was thrown so we don't know the resolution
+            // of one stream or of all streams)
+            return 1;
```
Stypox, line 581: Note: this fixes #6860 :confetti_ball:

Stypox (changes requested):
Looks almost good to me, can this be converted to non-draft?
I still haven't tested

app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
Stypox: This is the diff to apply, an it works:
```diff
diff --git a/app/src/main/java/org/schabi/newpipe/player/resolver/AudioPlaybackResolver.java b/app/src/main/java/org/schabi/newpipe/player/resolver/AudioPlaybackResolver.java
index e92a2722f..3be6503b9 100644
--- a/app/src/main/java/org/schabi/newpipe/player/resolver/AudioPlaybackResolver.java
+++ b/app/src/main/java/org/schabi/newpipe/player/resolver/AudioPlaybackResolver.java
@@ -37,7 +37,7 @@ public class AudioPlaybackResolver implements PlaybackResolver {
     @Override
     @Nullable
     public MediaSource resolve(@NonNull final StreamInfo info) {
-        final MediaSource liveSource = maybeBuildLiveMediaSource(dataSource, info);
+        final MediaSource liveSource = PlaybackResolver.maybeBuildLiveMediaSource(dataSource, info);
         if (liveSource != null) {
             return liveSource;
         }
@@ -54,8 +54,8 @@ public class AudioPlaybackResolver implements PlaybackResolver {
         final MediaSourceTag tag = new MediaSourceTag(info);
 
         try {
-            return buildMediaSource(dataSource, audio, info, PlayerHelper.cacheKeyOf(info, audio),
-                    tag);
+            return PlaybackResolver.buildMediaSource(
+                    dataSource, audio, info, PlayerHelper.cacheKeyOf(info, audio), tag);
         } catch (final IOException e) {
             Log.e(TAG, "Unable to create audio source:", e);
             return null;
diff --git a/app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java b/app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
index 17cb2bbee..29af4585a 100644
--- a/app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
+++ b/app/src/main/java/org/schabi/newpipe/player/resolver/PlaybackResolver.java
@@ -39,7 +39,7 @@ import java.util.Objects;
 public interface PlaybackResolver extends Resolver<StreamInfo, MediaSource> {
 
     @Nullable
-    default MediaSource maybeBuildLiveMediaSource(@NonNull final PlayerDataSource dataSource,
+    static MediaSource maybeBuildLiveMediaSource(@NonNull final PlayerDataSource dataSource,
                                                   @NonNull final StreamInfo info) {
         final StreamType streamType = info.getStreamType();
         if (!StreamTypeUtil.isLiveStream(streamType)) {
@@ -57,7 +57,7 @@ public interface PlaybackResolver extends Resolver<StreamInfo, MediaSource> {
     }
 
     @NonNull
-    default MediaSource buildLiveMediaSource(@NonNull final PlayerDataSource dataSource,
+    static MediaSource buildLiveMediaSource(@NonNull final PlayerDataSource dataSource,
                                              @NonNull final String sourceUrl,
                                              @C.ContentType final int type,
                                              @NonNull final MediaSourceTag metadata) {
@@ -86,7 +86,7 @@ public interface PlaybackResolver extends Resolver<StreamInfo, MediaSource> {
     }
 
     @NonNull
-    default MediaSource buildMediaSource(@NonNull final PlayerDataSource dataSource,
+    static MediaSource buildMediaSource(@NonNull final PlayerDataSource dataSource,
                                          @NonNull final Stream stream,
                                          @NonNull final StreamInfo streamInfo,
                                          @NonNull final String cacheKey,
@@ -104,7 +104,7 @@ public interface PlaybackResolver extends Resolver<StreamInfo, MediaSource> {
         }
     }
 
-    default <T extends Stream> ProgressiveMediaSource buildProgressiveMediaSource(
+    static <T extends Stream> ProgressiveMediaSource buildProgressiveMediaSource(
             @NonNull final PlayerDataSource dataSource,
             @NonNull final T stream,
             @NonNull final String cacheKey,
@@ -125,7 +125,7 @@ public interface PlaybackResolver extends Resolver<StreamInfo, MediaSource> {
         }
     }
 
-    default <T extends Stream> DashMediaSource buildDashMediaSource(
+    static <T extends Stream> DashMediaSource buildDashMediaSource(
             @NonNull final PlayerDataSource dataSource,
             @NonNull final T stream,
             @NonNull final StreamInfo streamInfo,
@@ -156,7 +156,7 @@ public interface PlaybackResolver extends Resolver<StreamInfo, MediaSource> {
         }
     }
 
-    default <T extends Stream> String generateDashManifestOfYoutubeDashStream(
+    static <T extends Stream> String generateDashManifestOfYoutubeDashStream(
             @NonNull final StreamInfo streamInfo,
             @NonNull final T stream) throws IOException {
         final String content;
@@ -209,7 +209,7 @@ public interface PlaybackResolver extends Resolver<StreamInfo, MediaSource> {
         return content;
     }
 
-    default <T extends Stream> DashMediaSource createDashMediaSource(
+    static <T extends Stream> DashMediaSource createDashMediaSource(
             @NonNull final PlayerDataSource dataSource,
             @NonNull final String content,
             @NonNull final T stream,
@@ -238,7 +238,7 @@ public interface PlaybackResolver extends Resolver<StreamInfo, MediaSource> {
                         .build());
     }
 
-    default <T extends Stream> HlsMediaSource buildHlsMediaSource(
+    static <T extends Stream> HlsMediaSource buildHlsMediaSource(
             @NonNull final PlayerDataSource dataSource,
             @NonNull final T stream,
             @NonNull final String cacheKey,
diff --git a/app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java b/app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java
index 598548242..ec1cbd9b0 100644
--- a/app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java
+++ b/app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java
@@ -52,7 +52,7 @@ public class VideoPlaybackResolver implements PlaybackResolver {
     @Override
     @Nullable
     public MediaSource resolve(@NonNull final StreamInfo info) {
-        final MediaSource liveSource = maybeBuildLiveMediaSource(dataSource, info);
+        final MediaSource liveSource = PlaybackResolver.maybeBuildLiveMediaSource(dataSource, info);
         if (liveSource != null) {
             return liveSource;
         }
@@ -80,8 +80,8 @@ public class VideoPlaybackResolver implements PlaybackResolver {
 
         if (video != null) {
             try {
-                final MediaSource streamSource = buildMediaSource(dataSource, video, info,
-                        PlayerHelper.cacheKeyOf(info, video), tag);
+                final MediaSource streamSource = PlaybackResolver.buildMediaSource(
+                        dataSource, video, info, PlayerHelper.cacheKeyOf(info, video), tag);
                 mediaSources.add(streamSource);
             } catch (final IOException e) {
                 Log.e(TAG, "Unable to create video source:", e);
@@ -99,8 +99,8 @@ public class VideoPlaybackResolver implements PlaybackResolver {
         // merge with audio stream in case if video does not contain audio
         if (audio != null && (video == null || video.isVideoOnly())) {
             try {
-                final MediaSource audioSource = buildMediaSource(dataSource, audio, info,
-                        PlayerHelper.cacheKeyOf(info, audio), tag);
+                final MediaSource audioSource = PlaybackResolver.buildMediaSource(
+                        dataSource, audio, info, PlayerHelper.cacheKeyOf(info, audio), tag);
                 mediaSources.add(audioSource);
             } catch (final IOException e) {
                 Log.e(TAG, "Unable to create audio source:", e);
```

TiA4f8R: Thank you! You are sometimes crazy, Stypox!

Issue #6233, opened 2021-05-03
Crash when using pause/play button on a bluetooth device


ChunkyProgrammer:
### Checklist

- [x] I am using the latest version - x.xx.x 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

1) Open NewPipe (don't play a video)
2) Click Pause/Play button a few times in a row

### Actual behaviour
App crashes

### Expected behavior
App ignores play/pause button when nothing is in the queue

## Exception
* __User Action:__ ui error
* __Request:__ ACRA report
* __Content Country:__ US
* __Content Language:__ en-US
* __App Language:__ en_US
* __Service:__ none
* __Version:__ 0.21.2
* __OS:__ Linux Android 11 - 30
<details><summary><b>Crash log </b></summary><p>

```
java.lang.NullPointerException: Attempt to invoke virtual method 'int org.schabi.newpipe.player.playqueue.PlayQueue.getIndex()' on a null object reference
	at org.schabi.newpipe.player.Player.onCompleted(Player.java:2154)
	at org.schabi.newpipe.player.Player.changeState(Player.java:1993)
	at org.schabi.newpipe.player.Player.onPlayerStateChanged(Player.java:1913)
	at com.google.android.exoplayer2.ExoPlayerImpl.lambda$updatePlaybackInfo$13(ExoPlayerImpl.java:1052)
	at com.google.android.exoplayer2.-$$Lambda$ExoPlayerImpl$eAKzukwdd6P91AoYEF8JC1HDnJ4.invoke(Unknown Source:4)
	at com.google.android.exoplayer2.util.ListenerSet$ListenerHolder.invoke(ListenerSet.java:297)
	at com.google.android.exoplayer2.util.ListenerSet.lambda$queueEvent$0(ListenerSet.java:184)
	at com.google.android.exoplayer2.util.-$$Lambda$ListenerSet$NbKDn9xtItiyMgYZmjIx_Sv1FFQ.run(Unknown Source:6)
	at com.google.android.exoplayer2.util.ListenerSet.flushEvents(ListenerSet.java:205)
[...]

PR #6570, opened 2021-06-25
Fix NPE when connecting via BT

Redirion:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)

The last weeks and NewPipe versions I had the issue that as soon as my phone connected via BT with my car NewPipe crashed as for some reason "onCompleted" was called.

`
```
java.lang.NullPointerException: Attempt to invoke virtual method 'int org.schabi.newpipe.player.playqueue.PlayQueue.getIndex()' on a null object reference
org.schabi.newpipe.player.Player.onCompleted
org.schabi.newpipe.player.Player.changeState
org.schabi.newpipe.player.Player.onPlayerStateChanged
```

Maybe I am the only one and this is a specific issue with my current car / car software, but as we have the same null checks in other listener methods, I felt it doesn't hurt to add it here as well.

#6233

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -2160,7 +2160,10 @@ public final class Player implements
 
     private void onCompleted() {
         if (DEBUG) {
-            Log.d(TAG, "onCompleted() called");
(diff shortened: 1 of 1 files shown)


mhmdanas (approved):
I can't seem to reproduce the NPE with this. Code also LGTM.

PR #6638, opened 2021-07-10
Use DisplayCompat.getMode().

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use [`DisplayCompat.getMode()`](https://developer.android.com/reference/androidx/core/view/DisplayCompat#getMode(android.content.Context,%20android.view.Display)) to get the current display size.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -10,7 +10,6 @@ import android.content.SharedPreferences;
 import android.content.pm.ActivityInfo;
 import android.database.ContentObserver;
 import android.graphics.Color;
-import android.graphics.Point;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
@@ -41,6 +40,7 @@ import androidx.appcompat.content.res.AppCompatResources;
 import androidx.appcompat.widget.Toolbar;
 import androidx.coordinatorlayout.widget.CoordinatorLayout;
 import androidx.core.content.ContextCompat;
+import androidx.core.view.DisplayCompat;
 import androidx.preference.PreferenceManager;
 
 import com.google.android.exoplayer2.ExoPlaybackException;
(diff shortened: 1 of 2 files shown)


TacoTheDank:
Dev notes: Be sure to read the changes between Core 1.3.2 and 1.6.0: https://developer.android.com/jetpack/androidx/releases/core

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1022,15 +1022,17 @@ public void updateTabLayoutVisibility() {
...
-                    final Point displaySize = new Point();
-                    Objects.requireNonNull(ContextCompat.getSystemService(getContext(),
-                            WindowManager.class)).getDefaultDisplay().getSize(displaySize);
+                    final WindowManager windowManager = Objects.requireNonNull(ContextCompat
+                            .getSystemService(context, WindowManager.class));
+                    final DisplayCompat.ModeCompat mode = DisplayCompat.getMode(context,
+                            windowManager.getDefaultDisplay());
 
-                    final int viewPagerVisibleHeight = displaySize.y - pagerHitRect.top;
+                    final int viewPagerVisibleHeight = mode.getPhysicalHeight() - pagerHitRect.top;
```
Stypox, line 1035: This is not correct. `viewPagerVisibleHeight` should represent the height of the view pager that is visible on the screen, but here the size of Android's bottom bar is also summed, since `getPhysicalHeight()` is used. But now that I see it, it seems like it was also wrong before, since `getDefaultDisplay().getSize()` returned the height of the app and the above difference was also wrong. What `viewPagerVisibleHeight` should be is "y_of_the_bottom_of_the_activity (i.e. just above Android's bottom bar) - y_of_the_top_of_viewPager". Is there a way to do this with the compat library? I saw that `getDefaultDisplay().getSize()` was deprecated in API 30 in favour of `DisplayMetrics`, and `WindowMetrics#getBounds()` would provide us with "y_of_the_bottom_of_the_activity", but that method has not yet been backported in the compat library.

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
Isira-Seneviratne: There's a [Window Manager](https://developer.android.com/jetpack/androidx/releases/window) library, but it's still in alpha.

Stypox:
> There's a Window Manager library which backports WindowMetrics, but it's still in alpha.

Mmmh, then I'd rather not merge this and keep the old behaviour, that has proven to work well (I haven't seen any issue about it and for me it works fine). If you or someone from the team disagrees, feel free to reopen ;-)
I'd also **not** switch to the Window Manager library even after it comes out of alpha, for the same reason.

PR #6686, opened 2021-07-16
Fixed scrolling on refresh in the feed

### LINKED ISSUE #6650: After Feed Update, Requires Manual Scroll Up to show Latest Videos
### Checklist

- [x] I am using the latest version - 0.21.6 (pre-release) 

### Steps to Reproduce:
_Bug Introduced in v0.21.6 (pre-release)_
_Note: This issue is created as per https://github.com/TeamNewPipe/NewPipe/issues/6577#issuecomment-877654793 on TobiGr's request https://github.com/TeamNewPipe/NewPipe/issues/6577#issuecomment-878303609_

1. Go to what's new section.
2. Pull down to refresh.
3. (If new videos are available) You have to manually scroll up to see the latest videos.

### Actual behavior

https://user-images.githubusercontent.com/50160188/125167877-60de0480-e1c2-11eb-88e4-115c6c0b2598.mp4

_**Ignore** the video pattern change. I don't know why that happened_

**Problem of Interest:**
_Before refresh:_
First video: 'Dear Apple'

_After refresh:_
First Video: (is still) 'Dear Apple'
Until you manually scroll up. You'll see new video above it.

### Expected behavior

After feed update, app must show latest videos at first. (without having to manually scroll)

### Device info

 - Android version/Custom ROM version: 9 Pie
 - Device model: Huawei Nova 3i

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- When the feed is refreshed, the grid will now always scroll to the top
- The problem only seems to occur when more items are added than removed

<details><summary>Click to open</summary>

Before:

https://user-images.githubusercontent.com/40789489/125986894-f4e7c510-f11a-4a08-bfce-a1153db2ebc6.mp4

After:

https://user-images.githubusercontent.com/40789489/125986902-fb4759f1-2188-47ef-a58d-e9c4d87f6173.mp4

</details>

- Fixes #6650

#### APK testing 

#### Due diligence

#### How to test
* Install the apk / Checkout the branch
* Set your phone / emulator system time to 2021-07-15 at 23:55 (you might need to reset it as the clock counts)
* Import this: [NewPipeData-20210716_190308.zip](https://github.com/TeamNewPipe/NewPipe/files/6832340/NewPipeData-20210716_190308.zip)
* Go to Feed/"What's New" and refresh it

→ It should now scroll to the top 🥳

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
@@ -41,6 +41,7 @@ import androidx.core.view.isVisible
 import androidx.lifecycle.ViewModelProvider
 import androidx.preference.PreferenceManager
 import androidx.recyclerview.widget.GridLayoutManager
+import androidx.recyclerview.widget.RecyclerView
 import com.xwray.groupie.GroupAdapter
 import com.xwray.groupie.GroupieViewHolder
 import com.xwray.groupie.Item
@@ -135,6 +136,19 @@ class FeedFragment : BaseStateFragment<FeedState>() {
             setOnItemClickListener(listenerStreamItem)
             setOnItemLongClickListener(listenerStreamItem)
         }
+
+        // Scrolls to the top when a new item is inserted e.g. when refreshing
+        // Using a workaround from
+        // https://dev.to/aldok/how-to-scroll-recyclerview-to-a-certain-position-5ck4
+        // here - the underlying dataset must be ready for scrolling or else scrollToPosition fails
+        groupAdapter.registerAdapterDataObserver(object : RecyclerView.AdapterDataObserver() {
+            override fun onItemRangeInserted(
+                positionStart: Int,
+                itemCount: Int
+            ) {
+                feedBinding.itemsList.scrollToPosition(0)
+            }
+        })
 
         feedBinding.itemsList.adapter = groupAdapter
         setupListViewMode()


TobiGr:
Thank you! That looks good.
I found a little bug which should be easy to solve by introducing a variable which checks whether items were inserted due to refreshing/updating the feed or making hidden items visible. Steps to reproduce:
- Watch a few videos from the feed to the end (or skip to the end).
- Go the the feed
- Press the "eye" button to hide already watched streams
- scroll down
- Press the "eye" button to make the watched streams visible again
- Notice the scroll to the top

timjefferies:
To me this was a feature not a bug. If you fix it I won't be able to see which new videos where added to the feed.

Could there be a line separator to show which new videos are added since the last update?

TobiGr:
> To me this was a feature not a bug. If you fix it I won't be able to tell which new videos where added to the feed.

That's true. Maybe we can add a chip at the top saying ":arrow_up:  X new streams" with an onclick listener which triggers the scroll to top.

litetex:
I will rethink the concept when I have time again 😄

litetex:
FYI:

I'm currently experimenting using new approaches e.g.:

Code is [here](https://github.com/litetex/NewPipe/tree/unstab-exp/rework-feed-new-items).

However there seem to be some problems:
1. When a feed is refreshed there are (randomly) multiple SuccessEvents:
  
  https://user-images.githubusercontent.com/40789489/131260463-80e30638-dd2a-45d1-b768-e3cd8cd3a19a.mp4

  This causes problems as it can't be computed what items are new ones. 

  @Stypox 
  I think you are the only person that  a) is currently active and b) has experience with the [FeedLoadService according to git blame](https://github.com/TeamNewPipe/NewPipe/blame/f63a4ee2aec91157927a29538d50b633762599ea/app/src/main/java/org/schabi/newpipe/local/feed/service/FeedLoadService.kt). 
  Do you have any ideas?

2. I'm a bit unsure how to implement the "X new items" popup/baloon whatever as seems to be dynamically inflated. 
    However that should be doable - at least somehow.

Stypox:
@litetex I can't reproduce the double success event. Maybe you can try debugging in the only place where a success event is generated, i.e. [here](https://github.com/TeamNewPipe/NewPipe/blob/f18a7c91ca247d8dc7f674044b9a79f8c6565a9f/app/src/main/java/org/schabi/newpipe/local/feed/service/FeedLoadService.kt#L297)? If the problem is not there climb up the call stack until you find where the events are duplicated.

Below I try to climb up the call stack that led to the code you are analyzing in the video above:
- In [FeedLoadService](https://github.com/TeamNewPipe/NewPipe/blob/f18a7c91ca247d8dc7f674044b9a79f8c6565a9f/app/src/main/java/org/schabi/newpipe/local/feed/service/FeedLoadService.kt#L297), i.e. the only other place where Android Studio says `SuccessResultEvent` is used, it seems like the success event is generated only once when `onComplete()` is called.
- `onComplete()` is in turn called by the huge ReactiveX chain [here](https://github.com/TeamNewPipe/NewPipe/blob/f18a7c91ca247d8dc7f674044b9a79f8c6565a9f/app/src/main/java/org/schabi/newpipe/local/feed/service/FeedLoadService.kt#L184-L252) which seems to be a `Flowable<MutableList<Notification<Pair<Long, ListInfo<StreamInfoItem>>>>>`, before `subscribe(resultSubscriber)` is called (try to add `;` at the end of the last `.observeOn(AndroidSchedulers.mainThread())`, then Alt+Enter and "introduce local variable", then on the created variable Alt+Enter and "Specify type explicitly"). So a flowable should theoretically only call `onComplete()` once, i.e. when it has finished processing all items.
- Maybe then the reason is that `startLoading()` turns out to be called multiple times? But Android Studio says it's only called in [FeedLoadService.onStartCommand](https://github.com/TeamNewPipe/NewPipe/blob/f18a7c91ca247d8dc7f674044b9a79f8c6565a9f/app/src/main/java/org/schabi/newpipe/local/feed/service/FeedLoadService.kt#L133).
- The documentation about `onStartCommand` says "Called by the system every time a client explicitly starts the service by calling `Context.startService`", which only seems to be happening in [FeedFragment.reloadContent](https://github.com/TeamNewPipe/NewPipe/blob/2027b743b4346f568d32d3653ab787894b2bc425/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt#L532)
- `reloadContent()` is overridden from `BaseStateFragment` and is called in multiple places, maybe the problem lies here?

> I'm a bit unsure how to implement the "X new items" popup/baloon whatever as seems to be dynamically inflated.

I think you will need to add a new item holder to work with groupie

litetex:
@Stypox 
Thank you for the detailed feedback.

*Disclaimer: The following text contains 2.5 hours of research so it got a bit longer...*

However none of the above code seems to be the cause of the error.
I'm to 95% sure that it's some kind of problem with the database-events from [``streamItems``](https://github.com/TeamNewPipe/NewPipe/blob/be676ad93cc94e8534fa5ce7cbedb218de4fa990/app/src/main/java/org/schabi/newpipe/local/feed/FeedViewModel.kt#L33).

I think to trigger the problem at least some new feed-items have to be fetched/added and some old ones to be removed.

Because when I remove https://github.com/TeamNewPipe/NewPipe/blob/be676ad93cc94e8534fa5ce7cbedb218de4fa990/app/src/main/java/org/schabi/newpipe/local/feed/FeedViewModel.kt#L55 I get event more events. 
Most of them (however sometimes not all) have different ``List<StreamWithState>``-sizes.

Usually it's 77->114->67 (last 2 differ according to current time and subscriptions) on my setup.

That looks like something on the internal database is raising these events.
And I'm pretty sure that's the case because:
* I got 37 inserts when monitoring the ``upsert``-function: https://github.com/TeamNewPipe/NewPipe/blob/be676ad93cc94e8534fa5ce7cbedb218de4fa990/app/src/main/java/org/schabi/newpipe/local/feed/FeedDatabaseManager.kt#L98 which matches exactly the 77->114 gap
* I got 47 deletions when monitoring the ``removeOrphansOrOlderStreams`` https://github.com/TeamNewPipe/NewPipe/blob/be676ad93cc94e8534fa5ce7cbedb218de4fa990/app/src/main/java/org/schabi/newpipe/local/feed/FeedDatabaseManager.kt#L107-L108 which matches exactly the 114->67 gap

So I guess the database events come in too late.
That would also explain why the problems are only randomly: Sometimes these 114, 67-events occur with the ``ProgressEvent``("Processing feedback")-phase and sometimes with the "SuccessResultEvent`` phase.

Source-Code I used: https://github.com/litetex/NewPipe/tree/e7e48aba35e18531800b004a9d82da5448de2d61
Full protocol: [log.txt](https://github.com/TeamNewPipe/NewPipe/files/7086441/log.txt)

Maybe it's also just a race condition because I'm debugging / using an emulator...

But I have at least one question regarding:
https://github.com/TeamNewPipe/NewPipe/blob/be676ad93cc94e8534fa5ce7cbedb218de4fa990/app/src/main/java/org/schabi/newpipe/local/feed/FeedViewModel.kt#L33-L38
Why is there a ``Flowable`` in the first place? 
Should a change in the ``streamItems`` really cause further processing?

Stypox:
> Maybe it's also just a race condition because I'm debugging / using an emulator...

This is really strange. It didn't seem to happen to me, and I was debugging on an emulator, too. It could as well be a race condition that should be solved.

> 
Why is there a Flowable in the first place?

Stypox:
<sub>Sorry I closed by accident</sub>

> Maybe it's also just a race condition because I'm debugging / using an emulator...

This is really strange. It didn't seem to happen to me, and I was debugging on an emulator, too. It could as well be a race condition that should be solved.

> Why is there a Flowable in the first place?

It should in theory be there so that updates to `showPlayedItems` can be sent one after the other, but I'm not so sure

> Should a change in the streamItems really cause further processing?

I am not sure I understand correctly, but I don't think changing something in those four lines would affect much performance.

litetex:
Succeeded by #7050

PR #6688, opened 2021-07-17
Fix some build warnings

litetex:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

This issue fixes 2 build warnings/deprecations:
grafik

The code on these parts was relatively old (>2 years).

Description of the fixes:
* ``FeedFragment``: Removed updateRelativeTimeViews when the activity is paused
We don't need to call ``updateRelativeTimeViews`` when the activity is paused, because the user likely won't notice it (because it is paused).
Despite that ``onResume`` already calls ``updateRelativeTimeViews`` so there is no need to do that twice.
* ``SubscriptionFragment``: Simplified code and adjusted the style so that it's similar to FeedFragment
The title there got set twice (no idea why). Simply copied use the same code (2 lines) from FeedFragment

not applicable

not applicable

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
@@ -167,14 +167,6 @@ class FeedFragment : BaseStateFragment<FeedState>() {
         }
     }
 
-    override fun setUserVisibleHint(isVisibleToUser: Boolean) {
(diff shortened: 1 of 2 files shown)


Stypox (approved):
Code looks good. I tested with a video in the feed published 19 minutes ago and after waiting 1 minute and switching to another app and then back to newpipe it became 20 minutes. Thanks :-)

PR #6712, opened 2021-07-20
Fix duplicate items in queue causing endless buffering

### LINKED ISSUE #4940: Endless buffering on duplicate videos in queue
### Checklist
- [x] I am using the latest version - 0.21.1 (But this bug happens at least since the v0.20.4)

1. Play a video in background that is duplicated in a playlist (the duplicate has to come right after the video) and the bug will occur when the duplicate plays.

### Actual behaviour
The video basically gets on an endless buffering and never starts playing or moves on to the next playlist video (unless you do it manually).

### Expected behavior
The video, whether duplicate or not should play normally.

### Device info

 - Android version/Custom ROM version: MIUI 11 (Android 7.0)
 - Device model: Redmi Note 4

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This is a simple bugfix that allows having play queues with duplicate items. For some reason an `equals()` method was implemented in `PlayQueueItem` that only compared the url, and since duplicate items have the same url they were considered the exact same `PlayQueueItem`s. Using Java's default `equals()` method, that just checks if two objects are the same with `==`, solves the issue, since duplicate items were anyway created one by one and thus are never identical. Removing `equals()` should be completely fine since `PlayQueueItem`s are never `clone()`d, and therefore we never end up with two objects that should be considered the same but down in memory aren't because they were cloned.

- Fixes #4940

#### APK testing 

I am not completely sure that this doesn't introduce regressions, so it needs some testing. @ekiniko12 @RickyM7 @sqtvrnqliq could you check if the endless buffering on duplicate videos in queue is solved?

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueueItem.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueueItem.java
@@ -64,20 +64,6 @@ public class PlayQueueItem implements Serializable {
         this.recoveryPosition = RECOVERY_UNSET;
     }
 
-    @Override
-    public boolean equals(final Object o) {
-        if (o instanceof PlayQueueItem) {
-            return url.equals(((PlayQueueItem) o).url);
-        } else {
-            return false;
-        }
-    }
-
-    @Override
-    public int hashCode() {
-        return url.hashCode();
-    }
-
     @NonNull
     public String getTitle() {
         return title;
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -813,7 +813,7 @@ public final class VideoDetailFragment
                                    @NonNull final String newTitle,
                                    @Nullable final PlayQueue newQueue) {
         if (isPlayerAvailable() && newQueue != null && playQueue != null
-                && !Objects.equals(newQueue.getItem(), playQueue.getItem())) {
+                && playQueue.getItem() != null && !playQueue.getItem().getUrl().equals(newUrl)) {
             // Preloading can be disabled since playback is surely being replaced.
             player.disablePreloadingOfCurrentTrack();
         }
--- a/app/src/test/java/org/schabi/newpipe/player/playqueue/PlayQueueItemTest.java
+++ b/app/src/test/java/org/schabi/newpipe/player/playqueue/PlayQueueItemTest.java
@@ -1,0 +1,19 @@
+package org.schabi.newpipe.player.playqueue;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+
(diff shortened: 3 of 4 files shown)


RickyM7:
I just tested it and the bug was fixed, thank you very much!

- Device: Redmi Note 4 (ArrowOS v11 - Android 11)

app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueueItem.java
```diff
@@ -64,20 +64,6 @@ private PlayQueueItem(@Nullable final String name, @Nullable final String url,
         this.recoveryPosition = RECOVERY_UNSET;
     }
 
-    @Override
-    public boolean equals(final Object o) {
```
TobiGr, line 68: Oh my! what a bug xD
Did you check the usages of the method? Maybe someone implemented this on purpose.

ekiniko12:
It's finally working, thank you!

PinkFriday:
Hey, I'm sqtvrnqliq, I was locked out of my other account so I'm now on this one! Is there any other way for me to download the debug apk because my phone isn't letting me download it when I go to artifacts :(

app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueueItem.java
Stypox: Android Studio's "Show usages" yields all of the usages of the `equals()` method on any object of any class, so that's not much useful. I tried to manually look for usages by checking out the methods in `PlayQueue` and in `Player` that use `PlayQueueItem` comparison and could not find any issue.
The only downside I can think of is that if the same video is queued multiple times in a row, the loading strategy may start to load the current and the next stream contemporarily, but since those streams point to the same url but do not compare equal, they will be loaded twice. But this only happens if two play queue items with the same url start loading contemporarily, because otherwise the caching method prevents reloading something that already was loaded. I don't think this is really a downside.

Redirion:
"for some reason" :D you have added this yourself Stypox:
https://github.com/TeamNewPipe/NewPipe/commit/2a2c82e73bb2380c07ab848b185952ab183a5502#diff-6dcb0e241835a3f32ba00835e94c64b48b6faf84c0252769617507c1c2fd8371

Stypox:
@Redirion 🙈🙈
I don't remember what the rationale for adding those methods was

Stypox:
@Redirion take a look at the last 2 commits, and thank you for taking a look ;-)
I also added some tests

Redirion (approved):
lgtm!

RickyM7:
@Stypox Was this included in the last update (v0.21.8)? I just tested this version but the bug still exists.

Stypox:
@RickyM7 no, 0.21.8 was a hotfix, just read the changelog

RickyM7:
> no, 0.21.8 was a hotfix, just read the changelog

Got it, thank you!

randacco:
Many thanks for this so needed fix!

PR #6720, opened 2021-07-21
Consolidate edittext alert dialogs into one common layout

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Consolidate alert dialogs with edittexts for views into one common layout, and we can add attributes through the code
- Differences:
  - Peertube adding dialog in the Content Settings looks a little bit different (and, dare I say, nicer :)) due to padding.
  - No visual differences in the add-video-to-playlist dialog (apart from the line now being red when the EditText box is selected).
  - I have no idea how to trigger the bookmark fragment dialog with the edittext, but I imagine it doesn't look very different.

(Ignore the comments disabled thing in the background of the second comparison)

By the way, is the AlertDialog buttons being white something we can fix? It looks bad imo, and them being red only seems to take effect in the settings screens.

| Before  | After |
| Screenshot_20210720-233023_NewPipe | Screenshot_20210720-233230_NewPipe alertdialog |

| Before  | After |
| Screenshot_20210720-233718_NewPipe | Screenshot_20210720-233727_NewPipe alertdialog |

#### APK testing 

#### Due diligence


Stypox (changes requested):
I think `setMaxLines(1);` and `setSaveEnabled(true);` can be specified in the `dialog_edit_text.xml`, since:
- we want a single line dialog everywhere
- we want the dialog content to not be lost on orientation changes everywhere: https://developer.android.com/reference/android/view/View.html#attr_android:saveEnabled

In particular, only `android:maxLines="1"` needs to be put in the xml, as `android:saveEnabled` is true by default.

---



> By the way, is the AlertDialog buttons being white something we can fix? It looks bad imo, and them being red only seems to take effect in the settings screens.

I think we might have to customize the `accentColor` for dialog buttons. But let's leave this to another PR.

TacoTheDank:
@Stypox Thanks! Fixed.

Stypox (approved):
LGTM :-)

PR #6741, opened 2021-07-23
Show hearts in comments

### LINKED ISSUE #6458: Show hearts in comments
### Checklist

Some comments have hearts given, newpipe should show them

N/A

#### Additional context

N/A

better comments, see when creator likes certain comments, etc.

KalleStruik:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Show a heart next to the like count on comments that have `heartedByUploader` set to true.

- Before:
image

- After:
image

- Fixes #6458 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
@@ -82,7 +82,7 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
         itemDislikesCountView = itemView.findViewById(R.id.detail_thumbs_down_count_view);
         itemPublishedTime = itemView.findViewById(R.id.itemPublishedTime);
         itemContentView = itemView.findViewById(R.id.itemCommentContentView);
-        itemHeartView = itemView.findViewById(R.id.detail_heart_img_view);
+        itemHeartView = itemView.findViewById(R.id.detail_heart_image_view);
 
         downloadThumbnailKey = infoItemBuilder.getContext().
                 getString(R.string.download_thumbnail_key);
@@ -157,6 +157,8 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
 
         if (item.isHeartedByUploader()) {
             itemHeartView.setVisibility(View.VISIBLE);
+        } else {
+            itemHeartView.setVisibility(View.GONE);
         }
 
         itemView.setOnClickListener(view -> {
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
@@ -7,6 +7,7 @@ import android.text.style.URLSpan;
 import android.text.util.Linkify;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.ImageView;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
 
@@ -45,6 +46,7 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
     private final TextView itemContentView;
     private final TextView itemLikesCountView;
     private final TextView itemDislikesCountView;
+    private final ImageView itemHeartView;
     private final TextView itemPublishedTime;
 
     private String commentText;
@@ -80,6 +82,7 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
         itemDislikesCountView = itemView.findViewById(R.id.detail_thumbs_down_count_view);
         itemPublishedTime = itemView.findViewById(R.id.itemPublishedTime);
         itemContentView = itemView.findViewById(R.id.itemCommentContentView);
+        itemHeartView = itemView.findViewById(R.id.detail_heart_img_view);
 
         downloadThumbnailKey = infoItemBuilder.getContext().
                 getString(R.string.download_thumbnail_key);
@@ -150,6 +153,10 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
                     .offsetDateTime()));
         } else {
             itemPublishedTime.setText(item.getTextualUploadDate());
+        }
+
+        if (item.isHeartedByUploader()) {
+            itemHeartView.setVisibility(View.VISIBLE);
         }
 
         itemView.setOnClickListener(view -> {
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
@@ -7,7 +7,6 @@ import android.text.style.URLSpan;
 import android.text.util.Linkify;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.ImageView;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
 
@@ -46,7 +45,6 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
     private final TextView itemContentView;
     private final TextView itemLikesCountView;
     private final TextView itemDislikesCountView;
-    private final ImageView itemHeartView;
     private final TextView itemPublishedTime;
 
     private String commentText;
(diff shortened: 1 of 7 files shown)


app/src/main/res/values/dimens.xml
```diff
@@ -81,11 +81,14 @@
     <dimen name="video_item_detail_sub_channel_image_size">16dp</dimen>
     <dimen name="video_item_detail_like_image_height">18sp</dimen>
     <dimen name="video_item_detail_like_image_width">18sp</dimen>
+    <dimen name="video_item_detail_heart_image_width">18sp</dimen>
+    <dimen name="video_item_detail_heart_image_height">18sp</dimen>
```
Stypox, line 85: Just use one constant:
```suggestion
    <dimen name="video_item_detail_heart_image_size">18sp</dimen>
```

app/src/main/res/layout/list_comments_item.xml
```diff
@@ -71,6 +71,17 @@
         tools:ignore="RtlHardcoded"
         tools:text="12M" />
 
+    <ImageView
+        android:id="@+id/detail_heart_img_view"
```
Stypox, line 75:
```suggestion
        android:id="@+id/detail_heart_image_view"
```

app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
```diff
@@ -152,6 +155,10 @@ public void updateFromItem(final InfoItem infoItem,
             itemPublishedTime.setText(item.getTextualUploadDate());
         }
 
+        if (item.isHeartedByUploader()) {
+            itemHeartView.setVisibility(View.VISIBLE);
+        }
```
Stypox, line 160: You should also add an "else" here, since item holders are reused

skyGtm:
I think Heart color should be RED.

Stypox (changes requested):
Thank you? :-) 
Please test on Android 4.4 and check if the heart shows up

KalleStruik:
I applied the changes requested. Also changed the heart color to be red as suggested by skyGtm. The below screenshot shows the red hearts working on android 4.4
image

opusforlife2:
Do we need a hover text on the icon saying "Liked by uploader"?

app/src/main/res/layout/list_comments_item.xml
```diff
@@ -71,6 +71,17 @@
...
         tools:text="12M" />
 
+    <ImageView
+        android:id="@+id/detail_heart_image_view"
+        android:layout_width="@dimen/video_item_detail_heart_image_size"
+        android:layout_height="@dimen/video_item_detail_heart_image_size"
+        android:layout_below="@id/itemCommentContentView"
+        android:layout_marginLeft="@dimen/video_item_detail_heart_margin"
+        android:layout_toRightOf="@+id/detail_thumbs_up_count_view"
+        android:contentDescription="@string/detail_heart_img_view_description"
```
Stypox, line 81: Android Studio gives a warning if a contentDescription is missing, but I guess you @opusforlife2 are right, it's not needed, so remove it

Stypox (changes requested):

app/src/main/res/layout/list_comments_item.xml
opusforlife2: xD I was actually advocating for it to be added since I didn't see it in the screenshots.

SameenAhnaf:
@opusforlife2 Personally, I don't think so. We have two alternatives instead.

1. Show uploader icon on the right bottom corner of the heart (just like in YouTube app)
2. A toast like 'Liked by uploader' may appear when short pressed on the 'Heart' icon. (I like this as the UI will remain clean. Also, the user might be curious about the function of heart.)

Stypox:
@SameenAhnaf the second one is bad ui; the first one could be a good idea but idk if it would be simple to implement

KalleStruik:
I added the avatar to see how it would look. What do you all think about something like this?
Currently using the comment author's avatar because I have not yet found a way to get the video uploader avatar. Any help with how to do that would be appreciated.
image

skyGtm:
Instead of showing hearts in channel avatar,
what about this?

Capture

_I prefer first and last picture_

Pick any number 1, 2, 3  or 4.

SameenAhnaf:
@skyGtm Like and Love are completely two different actions. We should not just focus on aesthetics rather on logic as well. That way, other users won't get confused.

@KalleStruik I'd personally prefer layout in https://github.com/TeamNewPipe/NewPipe/pull/6741#issuecomment-885798814 over the latest layout. I'm not sure if the latest layout will pave the way for other devs to replace uploader icon there. If that's the case, I'm completely fine with it.

KalleStruik:
@SameenAhnaf Currently I have not pushed the changes made in https://github.com/TeamNewPipe/NewPipe/pull/6741#issuecomment-886199187 onto this branch yet, because I have not found a way to use the video uploader avatar instead of the commenter avatar. I was hoping someone else reading this PR would know of a way to get it, since I do not see an obvious way to do so.

SameenAhnaf:
@KalleStruik Thanks for the reply, that's good to know. In my opinion, https://github.com/TeamNewPipe/NewPipe/pull/6741#issuecomment-885798814 should get the job done at least for now.

Stypox:
> Any help with how to do that would be appreciated.

@KalleStruik I think this is not doable unfortunately, after looking into it. The comments themselves know nothing about the uploader avatar, and obtaining that info from the video detail fragment is not doable either, since comments could finish loading before video info does so the uploader avatar url would not be available. So let's stick to just the heart with the tooltip; ignore my review above.

app/src/main/res/layout/list_comments_item.xml
```diff
@@ -71,6 +71,17 @@
...
+        android:visibility="gone"
+        app:srcCompat="@drawable/ic_heart" />
```
Stypox, line 83: Add `tools:visibility="visible"` here, so that the heart is visible in the Android Studio preview

app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
```diff
@@ -152,6 +155,12 @@ public void updateFromItem(final InfoItem infoItem,
             itemPublishedTime.setText(item.getTextualUploadDate());
         }
 
+        if (item.isHeartedByUploader()) {
+            itemHeartView.setVisibility(View.VISIBLE);
+        } else {
+            itemHeartView.setVisibility(View.GONE);
+        }
+
```
Stypox, line 163: I just realized this should be moved to `CommentsInfoItemHolder`, a subclass of `CommentsMiniInfoItemHolder`, since the `list_comments_mini_item.xml` layout does not contain the heart, only `list_comments_item.xml` does

Stypox (changes requested):
Almost ready ;-)

app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
```diff
@@ -152,6 +155,12 @@ public void updateFromItem(final InfoItem infoItem,
             itemPublishedTime.setText(item.getTextualUploadDate());
         }
 
+        if (item.isHeartedByUploader()) {
```
litetex, line 158: Tip to turn this into a one liner and make it more readable:
``itemHeartView.setVisibility(item.isHeartedByUploader() ? View.VISIBLE : View.GONE);``

KalleStruik:
I made the changes you requested, but I seem to be getting hit by #6744 now, so I can't test it.

Stypox (approved):
Code looks good. I couldn't test on YouTube either (though at least in SoundCloud and PeerTube it doesn't crash), let's wait until that problem is solved. Thank you!

Stypox:
It works, I finally tested by using by opening [this fork](https://github.com/FireMasterK/NewPipeExtractor/commits/youtubei-api) by @FireMasterK with fixed comments, except that they are not completely fixed so the "heartedByUploader" part was wrong, but I was able to test anyway by debugging and manually setting `CommentsInfoItem.heartedByUploader = true` for some items. Thank you!

castrik:
I personally prefer how the youtube app does this, they show the avatar of the uploader as the icon with a small heart beside it
Screenshot_20210802-054258

tsiflimagas:
@castrik as it was discussed in the above messages, that's not feasible in newpipe.

castrik:
I think stypox said comments don't know about the commentor's avatar as the ss posted by kalle struik does that, youtube app has the video uploader's avatar in the heart
is that not possible either? or was he saying this exact thing?

Stypox:
> I think stypox said comments don't know about the commentor's avatar 

No, obviously comments do know about the commentor's avatar, otherwise how would they be able to display it?

> youtube app has the video uploader's avatar in the heart

Yes, but doing this would require many YouTube-only strange changes in the extractor, and I don't think it's worth it for such a little design improvement.

FireMasterK:
> Yes, but doing this would require many YouTube-only strange changes in the extractor, and I don't think it's worth it for such a little design improvement.

I think this can be avoided by getting the channel avatar from the `StreamInfo` object itself.

Stypox:
@FireMasterK that can't be done since comments can load before the stream info. I would not pospone loading comments just for a style improvement

PR #6773, opened 2021-07-27
Add ability to mark an item as played

### LINKED ISSUE #2642: [Feature request] manually mark videos as watched

-----

I welcomed addition of 'watched videos' feature and it's really great. However just today I figured out that quite often I watch videos on the internet (invidious) and it would be helpful to be able to manually mark video as watched (so, in the future, I would be able to filter them out of channel views for example)

nschulzke:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

- Add a method to HistoryRecordManager that adds a record to the history and sets a video to 100% watched
- Add an item to the stream dialog labeled "Mark as played" that calls the above method

- Before:
Screenshot_1627354761
- After:
Screenshot_1627354961

- Fixes #2642

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/history/HistoryRecordManager.java
+++ b/app/src/main/java/org/schabi/newpipe/local/history/HistoryRecordManager.java
@@ -44,6 +44,7 @@ import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.stream.StreamInfo;
 import org.schabi.newpipe.extractor.stream.StreamInfoItem;
 import org.schabi.newpipe.player.playqueue.PlayQueueItem;
+import org.schabi.newpipe.util.ExtractorHelper;
 
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
@@ -91,16 +92,27 @@ public class HistoryRecordManager {
         return Maybe.fromCallable(() -> database.runInTransaction(() -> {
             final long streamId = streamTable.upsert(new StreamEntity(info));
 
+            long duration = info.getDuration();
+            if (duration < 0) {
+                duration = ExtractorHelper.getStreamInfo(
+                        info.getServiceId(),
+                        info.getUrl(),
+                        false)
+                        .subscribeOn(Schedulers.io())
+                        .blockingGet()
+                        .getDuration();
+            }
+
             final List<StreamStateEntity> states = streamStateTable.getState(streamId)
                     .blockingFirst();
             if (!states.isEmpty()) {
                 final StreamStateEntity entity = states.get(0);
-                entity.setProgressMillis(info.getDuration() * 1000);
+                entity.setProgressMillis(duration * 1000);
                 streamStateTable.update(entity);
             } else {
                 final StreamStateEntity entity = new StreamStateEntity(
                         streamId,
-                        info.getDuration() * 1000
+                        duration * 1000
                 );
                 streamStateTable.insert(entity);
             }
--- a/app/src/main/java/org/schabi/newpipe/local/history/HistoryRecordManager.java
+++ b/app/src/main/java/org/schabi/newpipe/local/history/HistoryRecordManager.java
@@ -28,6 +28,7 @@ import org.schabi.newpipe.NewPipeDatabase;
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.database.AppDatabase;
 import org.schabi.newpipe.database.LocalItem;
+import org.schabi.newpipe.database.feed.dao.FeedDAO;
 import org.schabi.newpipe.database.history.dao.SearchHistoryDAO;
 import org.schabi.newpipe.database.history.dao.StreamHistoryDAO;
 import org.schabi.newpipe.database.history.model.SearchHistoryEntry;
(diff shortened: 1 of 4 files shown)


Stypox:
Thank you! Have you tested this with the fast-loading feed (press on the (?))?

Fs00:
That seems a nice addition! I was thinking that maybe "Mark as _watched"_ (or _viewed_) would sound a bit clearer than _played_?

nschulzke:
> Thank you! Have you tested this with the fast-loading feed (press on the (?))?

No, I didn't know about that feature! I probably won't get to it today, but I'll test it tomorrow.

> That seems a nice addition! I was thinking that maybe "Mark as watched" (or viewed) would sound a bit clearer than played?

Yes, I'd agree. I chose "played" because that's the word that's already used in the "Show played items" button, which this is intended to synergize with. @Stypox, would it be okay if I changed both locations to "watched", or should we leave the wording as is?

Stypox:
Yes, change the wording

nschulzke:
Okay, I've changed the wording for both the Show Watched Items button text and the new dialog item.

I also tested it with the fast-loading feed for both YouTube and FramaTube, and it works the same as with the regular feed.

Stypox:
@nschulzke actually, the "Mark as watched" entry does nothing on videos loaded by the fast-feed mode that have never been opened. That happens since items loaded with the fast-feed mode do not come with a duration.

nschulzke:
> @nschulzke actually, the "Mark as watched" entry does nothing on videos loaded by the fast-feed mode that have never been opened. That happens since items loaded with the fast-feed mode do not come with a duration.

I expected that behavior based on the help text, but that's not what I'm seeing in running it. I'm probably somehow accidentally caching the durations at some other point in the process.

I've also tried refreshing the feed while fast mode is on and then testing, it still works. It also works after wiping all the caches and history via the settings. Is there a way to force the app to drop cached duration data for testing?

---

Regardless, I pushed up a change that, when the duration is missing, loads it via `ExtractorHelper`. I tested it by first making it use `ExtractorHelper` all the time, verified that it works, and then I made it only happen if it doesn't have a duration.

Stypox:
> Is there a way to force the app to drop cached duration data for testing?

Try to import an old database, then turn on fast mode and finally reload the feed. That's what I did. Otherwise turn on fast mode, subscribe to one of those channels putting out a new video every minute, wait a minute and reload the feed ;-)

nschulzke:
> Code looks good to me, thank you! Before merging could you add some comments to the markAsWatched method?

Done.

Stypox (changes requested):
After testing a little bit it seems like your code does the correct thing when saving the stream state for items w/o duration, but the watched bar below the item never shows up (not even after reloading the app). All of this is solved as soon as the video is opened by tapping on it: the bar shows up correctly. Is the stream info loaded in `markAsWatched` saved into the known videos of the database?

nschulzke:
Ah, yep, you're correct. I was able to replicate your findings using the method you suggested (loading an old database). I pushed up a commit that switches the order of steps to ensure that the fetched data gets saved.

I've tested it now and confirmed that it works as expected in both fast and slow mode. If no duration has yet been loaded, it takes a second or two to fetch the duration. It then saves both the duration and the watched state to the database.

app/src/main/java/org/schabi/newpipe/local/history/HistoryRecordManager.java
```diff
@@ -101,31 +101,31 @@ public HistoryRecordManager(final Context context) {
...
                         info.getServiceId(),
                         info.getUrl(),
                         false)
                         .subscribeOn(Schedulers.io())
                         .blockingGet()
                         .getDuration();
+                info.setDuration(duration);
             }
+            // Upsert to get a stream ID and update durations if we fetched one
+            final long streamId = streamTable.upsert(new StreamEntity(info));
```
Stypox, line 116: I'd not do it this way. Instead of `upsert`ing the previous info + the newly fetched duration, you should `upsert` the more complete info you got as a result of `ExtractorHelper.getStreamInfo()`. This would be the code:
```java
StreamInfo completeInfo = info;
// Duration will not exist if the item was loaded with fast mode, so fetch it if empty
if (completeInfo.getDuration() < 0) {
    completeInfo = ExtractorHelper.getStreamInfo(info.getServiceId(), info.getUrl(), false)
            .subscribeOn(Schedulers.io())
            .blockingGet();
}
// Upsert to get a stream ID and update durations if we fetched one
final long streamId = streamTable.upsert(new StreamEntity(completeInfo));
```
And then in the code below use `completeInfo` instead of `info`.

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/local/history/HistoryRecordManager.java
nschulzke: I made it so that if we fetch the complete info, we save it, otherwise we save the incomplete info. I couldn't do it exactly that way, because the types don't match (`StreamInfo` versus `StreamInfoItem`), but the effect should be the same.

Stypox (approved):
Thank you! I tested again and it works perfectly, and code looks good :-)

Stypox:
Ok, I see. It should be simple, if you want to try to implement it @nschulzke

nschulzke:
Yep, I'll do that.

PR #6789, opened 2021-07-28
Ignore paths unrelated to builds in CI

mhmdanas:
#### What is it?

Avoid running CI when only the only change is a README file.

#### Due diligence

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,12 +7,16 @@ on:
       - master
     paths-ignore:
       - 'README*.md'
+      - 'fastlane/**'
+      - 'assets/**'
   push:
     branches:
       - dev
       - master
     paths-ignore:
       - 'README*.md'
+      - 'fastlane/**'
+      - 'assets/**'
 
 jobs:
   build-and-test-jvm:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,7 @@ on:
       - 'README*.md'
       - 'fastlane/**'
       - 'assets/**'
-      - .github/PULL_REQUEST_TEMPLATE.md
-      - '.github/ISSUE_TEMPLATE/**'
+      - '.github/**/*.md'
   push:
     branches:
       - dev
@@ -19,8 +18,7 @@ on:
       - 'README*.md'
       - 'fastlane/**'
       - 'assets/**'
-      - .github/PULL_REQUEST_TEMPLATE.md
-      - '.github/ISSUE_TEMPLATE/**'
+      - '.github/**/*.md'
 
 jobs:
   build-and-test-jvm:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,10 +5,14 @@ on:
     branches:
       - dev
       - master
+    paths-ignore:
+      - 'README*.md'
   push:
     branches:
       - dev
       - master
+    paths-ignore:
+      - 'README*.md'
 
 jobs:
   build-and-test-jvm:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,8 @@ on:
       - 'README*.md'
       - 'fastlane/**'
       - 'assets/**'
+      - .github/PULL_REQUEST_TEMPLATE.md
+      - '.github/ISSUE_TEMPLATE/**'
   push:
     branches:
       - dev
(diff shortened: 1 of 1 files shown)


TobiGr:
Can these PRs be merged if CI is required?

mhmdanas:
Probably, but I need to test this first. Give me a bit and I'll come back with results.

mhmdanas:
So I made a pull request with changes only to READMEs (Korean and English), and no checks were run. Also, I added a branch protection rule that requires all status checks to pass before merging, and I'm able to merge the PR, so this should be good to go.

https://github.com/mhmdanas/NewPipe/pull/1

mhmdanas:
Any others paths you think should be ignored?

TobiGr:
the `/fastlane` and `/assets` directories

XiangRongLin:
Is there no option for the inverse functionality. As in "only run CI when changes in path are made", instead of "skip CI when changes are only in path"

.github/workflows/ci.yml
```diff
@@ -5,10 +5,22 @@ on:
     branches:
       - dev
       - master
+    paths-ignore:
+      - 'README*.md'
+      - 'fastlane/**'
+      - 'assets/**'
+      - .github/PULL_REQUEST_TEMPLATE.md
```
XiangRongLin, line 12: why do the other lines have single ticks '' but this one doesn't?

XiangRongLin (commented):
Does it still trigger correctly if changes too the source code AND the readme are made? Or do you have a link to the documentation?

.github/workflows/ci.yml
```diff
@@ -5,10 +5,22 @@ on:
     branches:
       - dev
       - master
+    paths-ignore:
+      - 'README*.md'
```
litetex, line 9: Would it be a good idea to ignore all *.md files?

litetex:
@XiangRongLin 
> Does it still trigger correctly if changes too the source code AND the readme are made? Or do you have a link to the documentation?

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths

> When all the path names match patterns in paths-ignore, the workflow will not run.

mhmdanas:
> Is there no option for the inverse functionality. As in "only run CI when changes in path are made", instead of "skip CI when changes are only in path"

There is, it's `paths` instead of `paths-ignore`.

.github/workflows/ci.yml
mhmdanas: Maybe, but I tried to be conservative here so so that there are no false negatives.

.github/workflows/ci.yml
mhmdanas: The other ones have stars in them, which are reserved characters in YAML.

.github/workflows/ci.yml
litetex: Hmm, I still don't feel glad about picking so many paths explicitly.

I don't think there can be any false positives (at least currently):

All ``.md`` files are either ``Readme.md``'s or stuff in the ``.github`` folder:
https://github.com/TeamNewPipe/NewPipe/search?l=Markdown&p=1

So maybe just go with the simpler ``**.md``?
Or what about simply ignoring all ``.md``'s in the ``.github`` folder?

mhmdanas: My concern is that somebody might change something in the code to depend on one of these files (or adds one), and then they forgot to change the paths listed here. However, I'm not strongly opposed to being less conservative and ignoring more things.

litetex: Okay.

However consider also the opposite situation could occur. 
e.g. someone adds a new ``.md`` to the ``.github`` folder and thinks it's ignore by the CI.

Anyway, I'm also fine with the current method.
But can you at least also add ``.github/CONTRIBUTING.md``?

mhmdanas: > someone adds a new .md to the .github folder and thinks it's ignore by the CI.

Not everybody is necessarily aware that files that the CI ignore actually exist, so it's more likely for a mistake to happen in this case I'd say. Anyway, I'm going to add `.github/**/*.md` then.

mhmdanas:
Anything stopping this from getting merged?

<sub>sorry for the nagging</sub>

Stypox (approved):

XiangRongLin:
Fine by me. I would not spend too much time worrying about a performance problem, that only hits GitHub. They will come if it becomes a problem.

XiangRongLin (approved):

PR #6820, opened 2021-08-01
Provide mime type to file picker to gray out unselectable files

### LINKED ISSUE #1493: Requesting changes to exporting database
* On back the database the default content country is not saved 

* Automatically backup the database within a limited period, say every week

* Backup the data to NewPipe downloads folder 

* And on importing automatically open the NewPipe downloads folder

And the overriding of existing database will be helpful because the too many files will be confused which one is the latest 
And if you can change the format of database file from zip to JSON

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

This PR just adds a mime type to all `getPicker` calls. When using SAF now, each time the app will want to open a file (to import the database or subscriptions), all files with the incorrect file format will be grayed out. So the database import picker will only let `.zip` be selected and the subscription import picker will only let `.json` be selected. The service-specific subscription imports, though, still have `*/*` as mime type as the extension of those files could be subject to change and is different between services.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

- Partially fixes #1493
- Fixes problems with people mixing up the database zip and the subscriptions json

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. @opusforlife2 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionsImportFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionsImportFragment.java
@@ -177,7 +177,8 @@ public class SubscriptionsImportFragment extends BaseFragment {
     }
 
     public void onImportFile() {
-        requestImportFileLauncher.launch(StoredFileHelper.getPicker(activity));
+        // leave */* mime type to support all services with different mime types and file extensions
+        requestImportFileLauncher.launch(StoredFileHelper.getPicker(activity, "*/*"));
     }
 
     private void requestImportFileResult(final ActivityResult result) {
--- a/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.kt
@@ -179,7 +179,7 @@ class SubscriptionFragment : BaseStateFragment<SubscriptionState>() {
     }
 
     private fun onImportPreviousSelected() {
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionsImportFragment.java
```diff
@@ -177,7 +177,8 @@ public void onImportUrl(final String value) {
     }
 
     public void onImportFile() {
-        requestImportFileLauncher.launch(StoredFileHelper.getPicker(activity));
+        // leave */* mime type to support all services with different mime types and file extensions
+        requestImportFileLauncher.launch(StoredFileHelper.getPicker(activity, "*/*"));
```
mhmdanas, line 181: What is the difference between `"*/*"` and not passing the argument at all?

opusforlife2:
Werks. 🎉 

Damn that is so much more convenient.

app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionsImportFragment.java
Stypox: Yes, there was no parameter at all before, mow I've added it

PR #6827, opened 2021-08-02
Optimize assets.

RDKRACZ:
#### What is it?
- [x] Meta improvement to the project (dev facing) **I THINK**
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

Lossless compression. I used oxipng.

No visual differences

#### APK testing 

#### Due diligence


Stypox:
What is the APK size difference?

mhmdanas:
@Stypox I'll check and come back with the results.

mhmdanas:
I built a release APK for each of this PR's commit and the previous commit. This PR's APK is 8904298 bytes long, while the previous APK is 8906273 bytes long, a difference of around 2 kilobytes.

mhmdanas (changes requested):
I believe it may not be worth it to minify anything not included in the APK, as I believe it would increase the Git repo's size, as Git does not store diffs of binary files (correct me if I'm wrong though).

TobiGr:
Minimizing the screenshots makes sense to me, because that results in faster loading of the readme

Stypox:
> Minimizing the screenshots makes sense to me, because that results in faster loading of the readme

Also to me

mhmdanas (approved):

mhmdanas:
@RDKRACZ thanks!

PR #6851, opened 2021-08-04
Catch errors while processing timestamp-links

### LINKED ISSUE #6842: Clicking the 1st comment of this youtube video to expand exists & crashes newpipe to homescreen
### Checklist

- [x] I am using the latest version - x.xx.x 

### Actual behavior

Newpipe exists to the homescreen which I believe is a form of crash. This does not invoke Guru Meditation or any other bug report pop-up

### Expected behavior

Expand the comment with ellipsis "..."

The comment which crashes the app in question

### Device info

 - Android version/Custom ROM version: Android 10, miui 12

 - Device model: Redmi 7a

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Catches errors while processing timestamp-links.

- Fixes #6842 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
@@ -72,7 +72,7 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
                 }
                 return streamUrl + url.replace(match.group(0), "#timestamp=" + timestamp);
             } catch (final Exception ex) {
-                Log.d(TAG, "Unable to process url='" + url + "' as timestampLink", ex);
+                Log.e(TAG, "Unable to process url='" + url + "' as timestampLink", ex);
                 return url;
             }
         }
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
@@ -47,7 +47,6 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
     public final CircleImageView itemThumbnailView;
     private final TextView itemContentView;
     private final TextView itemLikesCountView;
-    private final TextView itemDislikesCountView;
     private final TextView itemPublishedTime;
 
     private String commentText;
@@ -81,7 +80,6 @@ public class CommentsMiniInfoItemHolder extends InfoItemHolder {
         itemRoot = itemView.findViewById(R.id.itemRoot);
         itemThumbnailView = itemView.findViewById(R.id.itemThumbnailView);
         itemLikesCountView = itemView.findViewById(R.id.detail_thumbs_up_count_view);
-        itemDislikesCountView = itemView.findViewById(R.id.detail_thumbs_down_count_view);
         itemPublishedTime = itemView.findViewById(R.id.itemPublishedTime);
         itemContentView = itemView.findViewById(R.id.itemCommentContentView);
 
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
@@ -5,6 +5,7 @@ import android.text.TextUtils;
 import android.text.method.LinkMovementMethod;
 import android.text.style.URLSpan;
 import android.text.util.Linkify;
+import android.util.Log;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.RelativeLayout;
@@ -32,6 +33,8 @@ import java.util.regex.Pattern;
 import de.hdodenhof.circleimageview.CircleImageView;
 
 public class CommentsMiniInfoItemHolder extends InfoItemHolder {
+    private static final String TAG = "CommentsMiniIIHolder";
+
     private static final int COMMENT_DEFAULT_LINES = 2;
     private static final int COMMENT_EXPANDED_LINES = 1000;
     private static final Pattern PATTERN = Pattern.compile("(\\d+:)?(\\d+)?:(\\d+)");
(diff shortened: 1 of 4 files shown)


TobiGr:
Why is that method even called? This fixes the symtoms of parsing an unexpactedly structured URL. Shouldn't we take a closer look at the `PATTERN`?
```
(\\d+:)?(\\d+)?:(\\d+)
```
This regex looks like it is accepting more than it should.

Stypox:
Maybe we should use:
```regex
(\\d{1,2}:)?(\\d{1,2})?:(\\d{1,2})
```

app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
```diff
@@ -53,20 +56,25 @@
...
+                }
+                if (minutes != null) {
+                    timestamp += (Integer.parseInt(minutes.replace(":", "")) * 60);
+                }
+                if (seconds != null) {
+                    timestamp += (Integer.parseInt(seconds));
+                }
+                return streamUrl + url.replace(match.group(0), "#timestamp=" + timestamp);
+            } catch (final Exception ex) {
+                Log.d(TAG, "Unable to process url='" + url + "' as timestampLink", ex);
```
Stypox, line 75: I think this would also be good as an error
```suggestion
                Log.e(TAG, "Unable to process url='" + url + "' as timestampLink", ex);
```

Stypox (changes requested):

litetex:
I think a problem with using
```
(\\d{1,2}:)?(\\d{1,2})?:(\\d{1,2})
```

Is that you get cut-off results like:
grafik

Should we stick with the old implementation?

Btw: Is it correct that even results like ``:12`` are consider valid?

Stypox:
> Is that you get cut-off results like:

Oh ok, then nevermind. Just keep the current one, but check that minutes and seconds are <= 60

> Btw: Is it correct that even results like :12 are consider valid?

I think so, but as @TiA4f8R

Stypox:
Wait, but why isn't `TextLinkifier` used? There the correct timestamps patter is used (we worked on it together a couple hours on Element):
```java
(?:^|(?!:)\W)(?:([0-5]?[0-9]):)?([0-5]?[0-9]):([0-5][0-9])(?=$|(?!:)\W)
```

Stypox:
Using `TextLinkifier.TIMESTAMPS_PATTERN` and this code for `timestampLink` seems to work for me. Using `TextLinkifier` wouldn't work since it would disable clicks on the text view.
```java
    private final Linkify.TransformFilter timestampLink = new Linkify.TransformFilter() {
        @Override
        public String transformUrl(final Matcher match, final String url) {
            final int timestampStart = match.start(2);
            final int timestampEnd = match.end(3);
            final String parsedTimestamp = commentText.substring(timestampStart, timestampEnd);
            final String[] timestampParts = parsedTimestamp.split(":");

            final int seconds;
            if (timestampParts.length == 3) { // timestamp format: XX:XX:XX
                seconds = Integer.parseInt(timestampParts[0]) * 3600 // hours
                        + Integer.parseInt(timestampParts[1]) * 60 // minutes
                        + Integer.parseInt(timestampParts[2]); // seconds
            } else if (timestampParts.length == 2) { // timestamp format: XX:XX
                seconds = Integer.parseInt(timestampParts[0]) * 60 // minutes
                        + Integer.parseInt(timestampParts[1]); // seconds
            } else {
                return url;
            }
            return streamUrl + url.replace(match.group(0), "#timestamp=" + seconds);
        }
    };
```

litetex:
I'm also currently writing some unit-tests and fun fact:
"Timestamp format: XX:XX:XX" never works because the wrong group (it should be group 1 but it's group 2) is matched 😆 

So e.g. 1:23:45 becomes 1425s (23:45) when it should be 5025s

Fixing it on the fly

app/src/main/java/org/schabi/newpipe/util/external_communication/TimestampExtractor.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Extracts timestamps.
+ */
+public final class TimestampExtractor {
+    public static final Pattern TIMESTAMPS_PATTERN = Pattern.compile(
+            "(?:^|(?!:)\\W)(?:([0-5]?[0-9]):)?([0-5]?[0-9]):([0-5][0-9])(?=$|(?!:)\\W)");
```
Stypox, line 11: Why stop with hours at `60`? I'd just make that illimitate: on a video 100 hours long, how could one target the 90th hour (e.g. 90:32:11)
```suggestion
            "(?:^|(?!:)\\W)(?:([0-9]+):)?([0-5]?[0-9]):([0-5][0-9])(?=$|(?!:)\\W)");
```

Stypox (changes requested):

app/src/test/java/org/schabi/newpipe/util/external_communication/TimestampExtractorTest.java
```diff
@@ -0,0 +1,101 @@
...
+@RunWith(Parameterized.class)
+public class TimestampExtractorTest {
+
+    @Parameterized.Parameter(0)
+    public Duration expected;
+
+    @Parameterized.Parameter(1)
+    public String stringToProcess;
+
+    @Parameterized.Parameters(name = "Expecting {0} for \"{1}\"")
```
Stypox, line 25: Oh, does this look great :-D

Stypox (changes requested):
Thank you, I like this implementation :-)

app/src/main/java/org/schabi/newpipe/util/external_communication/TimestampExtractor.java
opusforlife2: @Stypox ~illimitate~ limitless

litetex: > Why stop with hours at `60`? I'd just make that ~illimitate~ limitless: on a video 100 hours long, how could one target the 90th hour (e.g. 90:32:11)

Some things to consider:
* Don't ask me I just copied it 😆, ask the inventor → https://github.com/TeamNewPipe/NewPipe/pull/6851#issuecomment-894060228
* Such a solution will throw an error if it exceeded the int limit
* Video lengths are capped on most platforms:
  * YouTube: max 12h
  Source: https://support.google.com/youtube/answer/71673
  * SonarCloud: max 6h 
  Source: https://help.soundcloud.com/hc/en-us/articles/115003452847-Uploading-requirements
  * Media CCC: ?
  * PeerTube: depends on the instance; clicked through a few instances and most of the have a contingent of ~20GB which are ~31h in average resolution video (I'm posting no links regarding peer tube instances, most of whom I visited host hoax videos - even though I searched with "Sensitive videos: blur")
  * Bandcamp: 600MB; mp3 min bitrate =96Kbps (according to [adobe](https://www.adobe.com/creativecloud/video/discover/audio-bitrate.html)) → 600MB / 96Kbps / 8 = ~781s = ~13h
  Source: https://get.bandcamp.help/hc/en-us/articles/360007901993-What-s-the-maximum-upload-size-

According to https://www.statista.com/statistics/1026923/youtube-video-category-average-length/ the average YT video has a length of ~12minutes. I myself rarely see/noticed videos which are longer than 1-2h.

According to the above the case of navigating to a video timestamp which is >60h is nearly zero. 
If there is such a case, it currently [simply doesn't detect the timestamp](https://regex101.com/r/EWQG8D/1) and the user has to move there by himself.

So I wouldn't change anything for now. 
If someone will ever have that problem, the person can open an issue.

Stypox: Ok, you are right

Stypox (approved):

opusforlife2:
@litetex Remember to always add a PR to the draft release whenever you merge it.

PR #6855, opened 2021-08-05
Fix readme link

TobiGr:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Fixed two warnings which I got while running the code inspection in Android Studio
- Fixed link to Brazil README in Japanese README
- ~~Fixed redundant character class declaration in `EMAIL_ADDRESS` regex~~

#### APK testing 

#### Due diligence

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -17,7 +17,7 @@
 <p align="center"><a href="https://newpipe.net">ウェブサイト</a> &bull; <a href="https://newpipe.net/blog/">ブログ</a> &bull; <a href="https://newpipe.net/FAQ/">FAQ</a> &bull; <a href="https://newpipe.net/press/">ニュース</a></p>
 <hr>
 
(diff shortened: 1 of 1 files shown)


Stypox (dismissed):

app/src/main/java/org/schabi/newpipe/util/urlfinder/PatternsCompat.java
```diff
@@ -361,7 +361,7 @@
 
     public static final Pattern EMAIL_ADDRESS
             = Pattern.compile(
-            "[a-zA-Z0-9\\+\\.\\_\\%\\-\\+]{1,256}"
+            "[a-zA-Z0-9\\.\\_\\%\\-\\+]{1,256}"
                     + "\\@"
```
XiangRongLin, line 365: Android Studio say that there are a lot more redundant escapes and that the field is unused. So we could even delete it.
``` java
 "[a-zA-Z0-9+._%\\-]{1,256}"
                    + "@"
```

TobiGr: This class is copied from the AOSP. That's why I'd say we should not modify it at all. I removed the commit.

XiangRongLin (approved):

PR #6872, opened 2021-08-07
Added a play next button to the long press menu

### LINKED ISSUE #4136: Feature Request: Add "Play next" button
#### An option to put a video on the top of the background queue.
I think it would be great to have such an option, as it saves a lot of time scrolling through the background queue, and you can play a song instantly without having to skip the current video or deleting the queue.

### LINKED ISSUE #6026: add „pre-queue“ option
### Checklist

A „pre-queue“ option would appear in the context-menu of a video only when there is a active stream already. Pressing it would play the video immediately and shift the media played before toching the button (and all enqueued) below it in the queue.

(Maybe it would be possible to remember the progress of the before consumed stream and skip there, after a video is finished?)

It is a easy and undistructive partial replacement for multiple queues.

It is for watching a video when you have listened to something else before and want to continue afterwards

z3r0r4:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Adds a play next button in the long press menu of a video, as asked for in  #4136.

- add getPlayerNextIntent in NavigationHelper
- handle not Append Only Intent in Player class
- add nextOn*Player functions in NavigationHelper
- create new StreamDialogEntry: next
- add new entry to the BaseListFragment
- add Strings for names

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

- Fixes #4136
- Fixes #6026

#### APK testing 

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/ONTRIBUTING.md).

#### Critique / Todo
- remove my comments
- reduce code duplication
- add translation strings


app/src/main/res/values/strings.xml
```diff
@@ -471,6 +471,8 @@
     <string name="hold_to_append">Hold to enqueue</string>
     <string name="show_channel_details">Show channel details</string>
     <string name="enqueue_stream">Enqueue</string>
+    <string name="play_next_stream">Next</string>
+    <string name="playing_next">Playing Next</string>
```
z3r0r4, line 475: maybe move them down a bit, as not to split enqueue_stream and enqueued

z3r0r4: moved them

app/src/main/res/values/strings.xml
```diff
@@ -472,6 +472,8 @@
     <string name="show_channel_details">Show channel details</string>
     <string name="enqueue_stream">Enqueue</string>
     <string name="enqueued">Enqueued</string>
+    <string name="play_next_stream">Next</string>
```
opusforlife2, line 475:
```suggestion
    <string name="play_next_stream">Play next</string>
```

z3r0r4: reasonable change for descriptiveness

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -209,6 +220,75 @@ public static void enqueueOnBackgroundPlayer(final Context context,
...
+                                              final PlayQueue queue,
+                                              final boolean selectOnAppend,
+                                              final boolean resumePlayback) {
+        Toast.makeText(context, R.string.playing_next, Toast.LENGTH_SHORT).show();
+        final Intent intent = getPlayerNextIntent(
+                context, MainPlayer.class, queue, selectOnAppend, resumePlayback);
+
+        intent.putExtra(Player.PLAYER_TYPE, MainPlayer.PlayerType.AUDIO.ordinal());
+        ContextCompat.startForegroundService(context, intent);
+    }
```
Stypox, line 291: Just do not add any of this and add only a `enqueueNextOnPlayer` with this code:
```java
    public static void enqueueNextOnPlayer(final Context context,
                                              final PlayQueue queue,
                                              final boolean selectOnAppend,
                                              final boolean resumePlayback) {
        Toast.makeText(context, R.string.playing_next, Toast.LENGTH_SHORT).show();
        final Intent intent = getPlayerNextIntent(
                context, MainPlayer.class, queue, selectOnAppend, resumePlayback);

        intent.putExtra(Player.PLAYER_TYPE, PlayerHolder.getInstance().getType().ordinal());
        ContextCompat.startForegroundService(context, intent);
    }
```
Assuming it is really needed to pass `PLAYER_TYPE`. Otherwise remove that, too.

---

Edit: just noticed you did this since you took inspiration from `enqueueOn*Player` methods. While you are at it, you could remove them all and create a single `enqueueOnPlayer`, similarly to what I suggested above.

---

Edit 2: I also noticed the `selectOnAppend` parameter is always `false`. Also remove that completely, from `Player`, from every method in `NavigationHelper` and anywhere else.

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -111,6 +111,17 @@ private NavigationHelper() {
                 .putExtra(Player.SELECT_ON_APPEND, selectOnAppend);
     }
 
+    @NonNull
+    public static <T> Intent getPlayerNextIntent(@NonNull final Context context,
```
Stypox, line 115: I'd call this `getPlayerEnqueueNextIntent`

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -55,6 +55,21 @@
...
+        if (type == AUDIO) {
+            NavigationHelper.nextOnVideoPlayer(fragment.getContext(),
+                    new SinglePlayQueue(item), false);
+        } else if (type == POPUP) {
+            NavigationHelper.nextOnPopupPlayer(fragment.getContext(),
+                    new SinglePlayQueue(item), false);
+        } else /* type == VIDEO */ {
+            NavigationHelper.nextOnVideoPlayer(fragment.getContext(),
+                    new SinglePlayQueue(item), false);
+        }
```
Stypox, line 71: So this would become:
```suggestion
        NavigationHelper.nextOnVideoPlayer(fragment.getContext(),
                new SinglePlayQueue(item), false);
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -606,15 +606,29 @@ public void handleIntent(@NonNull final Intent intent) {
         if (intent.hasExtra(PLAYBACK_QUALITY)) {
             setPlaybackQuality(intent.getStringExtra(PLAYBACK_QUALITY));
         }
-
+//TODO reduce code duplication
         // Resolve append intents
-        if (intent.getBooleanExtra(APPEND_ONLY, false) && playQueue != null) {
+        if (intent.getBooleanExtra(APPEND_ONLY, false)
```
Stypox, line 611: I don't think `APPEND_ONLY` was created for this reason. The result of `intent.getBooleanExtra(APPEND_ONLY, false)` is always either `true` or `false`, so when `playQueue != null` one of the two if branches will surely be executed and `return;` will be reached. `APPEND_ONLY=false` doesn't mean "enqueue next", though I get the confusion (the player code is a mess :sob:). I'd do this instead:
- Rename `APPEND_ONLY` to `ENQUEUE` and use it for this `if`, where previously `APPEND_ONLY` was used
- Add `ENQUEUE_NEXT` and use it for the `if` below (the new one you added)

Stypox (changes requested):
Thank you for adding this requested features :-)

Sorry for the many things I asked you to do, feel free to ask for help or delegate something to me if you can't figure it out ;-)

app/src/main/res/values/strings.xml
```diff
@@ -472,6 +472,8 @@
     <string name="show_channel_details">Show channel details</string>
     <string name="enqueue_stream">Enqueue</string>
     <string name="enqueued">Enqueued</string>
+    <string name="play_next_stream">Play next</string>
+    <string name="playing_next">Playing Next</string>
```
Stypox, line 476:
```suggestion
    <string name="playing_next">Playing next</string>
```

Stypox (changes requested):

z3r0r4:
> Thank you for adding this requested features :-)
> 
> Sorry for the many things I asked you to do, feel free to ask for help or delegate something to me if you can't figure it out ;-)

My pleasure! 
Thank you for the thorough review. Some of the requested changes should have been quite obvious optimizations when the goal is more than a mvp. 
As long as it isnt urgent i'll get it done by myself (hopefully this week).
I might be back for some clarification on how to integrate your code though.

SameenAhnaf:
A few bugs found.

1. Play queue become inaccessible once 'Play next' is selected. User needs to click on Background or selected player button. Also, some videos get multiple times although selected just once.
https://drive.google.com/file/d/1BJnia6hJJasVy_LY-90n9eK4UQH7Czp2/view?usp=drivesdk

2. There are multiple bugs here. Watch the video for more details.
https://drive.google.com/file/d/1BHzXe_Wf1YdcACAjEmAo5RmmLComv4ni/view?usp=drivesdk

MD77MD:
just s suggestion... could we call it "enqueue next"... also more consistent. 

this is what VLC uses if I remember correctly.

Stypox:
@MD77MD mmmh yeah, I agree with you

opusforlife2:
Hmmm. If there is only the current video in the queue, Enqueue and Enqueue Next will have the same function. Should the button check for >1 items in the queue before showing?

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
z3r0r4: i've removed and merged the redundant `enqueueOn*Player` methods and replaced them with ones without `selectOnAppend` and `RESUME_PLAYBACK`.
`PLAYER_TYPE` is still passed as in your code.

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
z3r0r4: done

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
z3r0r4: enqueueOn*Player methods are needed in other places
will try to resolve this

z3r0r4: I want to note that `resumePlayback `seems to be always true except for in `VideoDetailFragment.enqueueOnPopupPlayer` where it is false.
Since according to you it is unused and i cant make out any uses;
 I'll replace any method calls if the `enqueueOn*Player` unrelated to if they specify `resumePlayback` as false

z3r0r4: f7f98518c15ea2a67deab14a73ab574c3807905e removes the remaining calls to the `enqueueOn*Player` methods with the `enqueueOnPlayer` method, independent of the `resumePlayback` parameter.
Thus it can be build again.

app/src/main/java/org/schabi/newpipe/player/Player.java
z3r0r4: 9bab471ff07cde5cf9ca73b762221adfa27788ed does this.
I have not only renamed `APPEND_ONLY` but also corrected the String to be `enqueue`.
No String named `APPEND_ONLY` remains in the Player Class.

z3r0r4:
> A few bugs found.
> 
>     1. Play queue become inaccessible once 'Play next' is selected. User needs to click on Background or selected player button. Also, some videos get multiple times although selected just once.
>        https://drive.google.com/file/d/1BJnia6hJJasVy_LY-90n9eK4UQH7Czp2/view?usp=drivesdk
> 
>     2. There are multiple bugs here. Watch the video for more details.
>        https://drive.google.com/file/d/1BHzXe_Wf1YdcACAjEmAo5RmmLComv4ni/view?usp=drivesdk

I cant seem to replicate these anymore, would it be possible for you to check again @SameenAhnaf ?

z3r0r4:
I do have to note a new bug tho:
The play next option is inaccessible when selecting videos inside of a playlist.

z3r0r4:
> I do have to note a new bug tho:
> The play next option is inaccessible when selecting videos inside of a playlist.
> 
> I think it might not be referenced in the code which generates the dialog inside of playlist, dont know yet where to check though

b8ebda79d896913b91eb83265bae8c1fcb9d4d1a fixes this hopefully

z3r0r4:
> just s suggestion... could we call it "enqueue next"... also more consistent.
> 
> this is what VLC uses if I remember correctly.

renamed with c6dbad1e81a482f88628ce46a7191d34defecdf3

z3r0r4:
> 
> 
> Hmmm. If there is only the current video in the queue, Enqueue and Enqueue Next will have the same function. Should the button check for >1 items in the queue before showing?

I've added a method to the PlayerHolder that returns the size of the PlayQueue.
The dialog entry is now only being added when the player is open and the queue contains >1 item.
See c9d5e211550688685ceda76cb3d95f23bd01eb99
Checking the size of the queue when the player isnt open crashes the app. Although this can never happen in the current code I could add catches to prevent this.

z3r0r4:
Bug (maybe from the code snipped recommended by @Stypox?):
 `PlayerHolder.getInstance().getType().ordinal()` is Null if no video is playing anywhere, this crashes the app when long pressing the background or popup button on the video page.

z3r0r4:
Bug:
Long press of the "Background Button" doesnt `enqueue`, but results in `enqueue next` ?

Bug:
If a video is playing and the "Background button" is long pressed the playqueue becomes inaccessible. This doesnt happen if the video is playing in the background.

z3r0r4:
> Bug:
> Long press of the "Background Button" doesnt `enqueue`, but seems to result in `enqueue next` ?

83df7b098be90809d664c2a89da97e70957f0878 fixes this
the wrong string was being toasted

z3r0r4:
> Bug:
> If a video is playing and the "Background button" is long pressed the playqueue becomes inaccessible. This doesnt happen if the video is playing in the background.

5b3487de8db52672ad8eab0291b3eaf55b860fbc attempts to fixes this by checking if the playertype is defined and defaulting to background audio
BUT: this removes the feature to long press on the pop up button to start pop up play, if nothing is playing. (it plays in background instead).
A single press still starts pop up play.
long pressing on the pop up button if something is playing in popup still continues playing in pop up.

I'd recommend removing the long press action on the pop up button or checking if anything is playing and only enabling the long press if there is a pop up.
I'd be able implement this as well, if i get any hints to where the long press methods are.

Stypox:
> Checking the size of the queue when the player isnt open crashes the app.

Just return 0 from the player holder method in that case, since the queue is really empty

> I'd recommend removing the long press action on the pop up button or checking if anything is playing and only enabling the long press if there is a pop up (In the VideoDetailFragment class).

Yes, remove the long press actions for _both_ the background and popup buttons when the queue is empty

z3r0r4:
> > Checking the size of the queue when the player isnt open crashes the app.
> 
> Just return 0 from the player holder method in that case, since the queue is really empty
> 
> > I'd recommend removing the long press action on the pop up button or checking if anything is playing and only enabling the long press if there is a pop up (In the VideoDetailFragment class).
> 
> Yes, remove the long press actions for _both_ the background and popup buttons when the queue is empty

Should i still make it default to Audio playback as a failsafe?

Stypox:
> Should i still make it default to Audio playback as a failsafe?

No, if you remove long-press for the popup, then remove it also for background, thanks ;-) <sub>(for everyone reading out-of-context, this refers to when the queue is empty and hence enqueueing would be the same as playing)</sub>

Stypox:
@z3r0r4 Wait, I noticed something. People now use long-pressing popup as a way to enqueue the stream and then switch to popup (or viceversa for background). I don't know if we want to keep this feature, since I guess removing, even though it would make the app more consistent, would create problems for someone. What do you think? @SameenAhnaf @opusforlife2 @mhmdanas @avently @TobiGr

MD77MD:
> @z3r0r4 Wait, I noticed something. People now use long-pressing popup as a way to enqueue the stream and then switch to popup (or viceversa for background). I don't know if we want to keep this feature, since I guess removing, even though it would make the app more consistent, would create problems for someone. What do you think? @SameenAhnaf @opusforlife2 @mhmdanas @avently @TobiGr

that's true

SameenAhnaf:
>could we call it "enqueue next"... also more consistent.

According to me, 'Queue next' sounds better. After all, 'Enqueue' is a separate feature and should not be mixed up.

>I don't know if we want to keep this feature, since I guess removing, even though it would make the app more consistent, would create problems for someone.

I absolutely agree with you.

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -155,57 +158,29 @@ public static void playOnBackgroundPlayer(final Context context,
...
-    public static void enqueueOnVideoPlayer(final Context context, final PlayQueue queue,
-                                            final boolean resumePlayback) {
-        enqueueOnVideoPlayer(context, queue, false, resumePlayback);
-    }
-
-    public static void enqueueOnVideoPlayer(final Context context, final PlayQueue queue,
-                                            final boolean selectOnAppend,
-                                            final boolean resumePlayback) {
-
+    /* ENQUEUE */
```
Stypox, line 161: Add spaces between the functions you added and above here

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -155,57 +158,29 @@ public static void playOnBackgroundPlayer(final Context context,
...
+    public static void enqueueOnPlayer(final Context context, final PlayQueue queue) {
```
Stypox, line 162: As discussed above, we apparently still need to be able to enqueue and switch to a different player at the same time. Therefore I would just have this function take `playerType` as a parameter...
```suggestion
    public static void enqueueOnPlayer(final Context context, final PlayQueue queue, final PlayerType playerType) {
```

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -155,57 +158,29 @@ public static void playOnBackgroundPlayer(final Context context,
...
+        int playerType = MainPlayer.PlayerType.AUDIO.ordinal();
+        if (PlayerHolder.getInstance().getType() != null) {
+            playerType = PlayerHolder.getInstance().getType().ordinal();
         }
-
-        Toast.makeText(context, R.string.enqueued, Toast.LENGTH_SHORT).show();
-        final Intent intent = getPlayerEnqueueIntent(
-                context, MainPlayer.class, queue, selectOnAppend, resumePlayback);
-        intent.putExtra(Player.PLAYER_TYPE, MainPlayer.PlayerType.POPUP.ordinal());
+        intent.putExtra(Player.PLAYER_TYPE, playerType);
```
Stypox, line 170: ...change this to just `intent.putExtra(Player.PLAYER_TYPE, playerType.ordinal())`...

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -155,57 +158,29 @@ public static void playOnBackgroundPlayer(final Context context,
...
         ContextCompat.startForegroundService(context, intent);
     }
+    /* NEXT */
```
Stypox, line 173: ... and add a helper function here
```suggestion

    /* ENQUEUE */
    public static void enqueueOnPlayer(final Context context, final PlayQueue queue) {
        PlayerType playerType = PlayerHolder.getInstance().getType();
        if (playerType == null) {
            Log.e(TAG, "Enqueueing when no player is open, defaulting to background");
            playerType = MainPlayer.PlayerType.AUDIO;
        }
        
        enqueueOnPlayer(context, queue, playerType);
    }

    /* ENQUEUE NEXT */
```
This function should be used anywhere except for the background and popup long-press actions (also on playlists, history, ...), where the other one should be used instead.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -606,16 +606,19 @@ public void handleIntent(@NonNull final Intent intent) {
         if (intent.hasExtra(PLAYBACK_QUALITY)) {
             setPlaybackQuality(intent.getStringExtra(PLAYBACK_QUALITY));
         }
-
-        // Resolve append intents
-        if (intent.getBooleanExtra(APPEND_ONLY, false) && playQueue != null) {
+        // Resolve enqueue intents
```
Stypox, line 609:
```suggestion

        // Resolve enqueue intents
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -606,16 +606,19 @@ public void handleIntent(@NonNull final Intent intent) {
...
             playQueue.append(newQueue.getStreams());
 
-            if ((intent.getBooleanExtra(SELECT_ON_APPEND, false)
-                    || currentState == STATE_COMPLETED) && newQueue.getStreams().size() > 0) {
-                playQueue.setIndex(sizeBeforeAppend);
-            }
+            return;
+        // Resolve enqueue next intents
+        } else if (intent.getBooleanExtra(ENQUEUE_NEXT, false)
+                && playQueue != null) { /* Enqueue next = true && playQueue not empty*/
```
Stypox, line 618: `/* Enqueue next = true && playQueue not empty*/`
Mmmh, this comment seems redundant to me. Also, this is not checking whether the playqueue is not empty, but just whether is it is non null, i.e. whether a queue was provided at all.

Stypox (changes requested):
Thank you for doing all of the refactoring! This looks so much better ;-)

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
z3r0r4: This implies that the helper function will only be used in the StreamDialogEntry Class.
see commit ca17b02e8642af85f5043b738b37b12915d3580d

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
z3r0r4: 8cbfcc4fa02f4668c427d48d42f7978f5b9a2a80

z3r0r4:
> Thank you for doing all of the refactoring! This looks so much better ;-)

thanks :D

app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
```diff
@@ -350,12 +350,18 @@ protected void showStreamDialog(final StreamInfoItem item) {
...
             return;
         }
-
         final List<StreamDialogEntry> entries = new ArrayList<>();
 
         if (PlayerHolder.getInstance().getType() != null) {
-            entries.add(StreamDialogEntry.enqueue);
+            entries.addAll(Arrays.asList(
+                    StreamDialogEntry.enqueue
+            ));
```
Stypox, line 358:
```suggestion
            entries.add(StreamDialogEntry.enqueue);
```

app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
```diff
@@ -350,12 +350,18 @@ protected void showStreamDialog(final StreamInfoItem item) {
...
         }
+
+        if (PlayerHolder.getInstance().getQueueSize() > 1) {
```
Stypox, line 361: Move this if above inside the other if just above, as if there is no player there surely isn't there a queue

Stypox (changes requested):

z3r0r4:
> > could we call it "enqueue next"... also more consistent.
> 
> According to me, 'Queue next' sounds better. After all, 'Enqueue' is a separate feature and should not be mixed up.

I think the parallelism between both entries is very intuitive UX. The features are very similar and only differ by one position in the queue, which is indicated by the "next" keyword. 
Only with the current name their similarities become really obvious on first view.
Thus I'd prefer not changing it.

Kr1sh1:
I agree with either "Queue next" or "Play next" rather than just "next". When I first saw that I felt it was a bit ambiguous.

z3r0r4:
> I agree with either "Queue next" or "Play next" rather than just "next". When I first saw that I felt it was a bit ambiguous.

"next" Isnt the used string anymore since quite some commits ago.
see [this commit](https://github.com/TeamNewPipe/NewPipe/pull/6872/commits/83df7b098be90809d664c2a89da97e70957f0878#diff-5e01f7d37a66e4ca03deefc205d8e7008661cdd0284a05aaba1858e6b7bf9103)

Kr1sh1:
I see, sorry I just glossed over that. Though I still feel "Play next" is more suitable, however it's your call

SameenAhnaf:
I think, a poll is more appropriate over scattered comments. @z3r0r4 @Kr1sh1 @MD77MD @Stypox @mhmdanas Please vote. Sorry for the pinge though.

👍 Play Next
♥️ Queue next
👀 Enqueue next

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -544,10 +545,14 @@ public boolean onLongClick(final View v) {
 
         switch (v.getId()) {
             case R.id.detail_controls_background:
-                openBackgroundPlayer(true);
+                if (PlayerHolder.getInstance().getQueueSize() > 0) {
```
Stypox, line 548:
```suggestion
                if (PlayerHolder.isPlayerOpen()) {
```

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -544,10 +545,14 @@ public boolean onLongClick(final View v) {
...
+                    openBackgroundPlayer(true);
+                }
                 break;
             case R.id.detail_controls_popup:
-                openPopupPlayer(true);
+                if (PlayerHolder.getInstance().getQueueSize() > 0) {
```
Stypox, line 553:
```suggestion
                if (PlayerHolder.isPlayerOpen() > 0) {
```

app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
```diff
@@ -350,12 +350,16 @@ protected void showStreamDialog(final StreamInfoItem item) {
         if (context == null || context.getResources() == null || activity == null) {
             return;
         }
-
         final List<StreamDialogEntry> entries = new ArrayList<>();
 
         if (PlayerHolder.getInstance().getType() != null) {
```
Stypox, line 355:
```suggestion
        if (PlayerHolder.isPlayerOpen()) {
```

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -147,6 +148,11 @@ protected void showStreamDialog(final StreamInfoItem item) {
         if (PlayerHolder.getInstance().getType() != null) {
```
Stypox, line 148:
```suggestion
        if (PlayerHolder.isPlayerOpen() != null) {
```

Stypox (changes requested):
Let's go with "Enqueue next" and call it a day, here is why:
- it is clear that the video is being added to the queue
- it is inconsistent with the enqueue button
- the first word aligns with "Enqueue" and will be more easily recognizeable in the blink of an eye

@z3r0r4 my suggestions just ask you to replace the places where it is checked if the player is open with the actual `PlayerHolder.isPlayerOpen()` method instead of other outdated workarounds involving other methods from `PlayerHolder`. Could you check if I missed anything, and correct those as well? You should be able to find them by having Android Studio look for all of the usages of e.g. the `PlayerHolder.getType()` and see if it is e.g. used just to check for `null`ity.

z3r0r4:
replaced workarounds using `getQueueSize` and `getType`. 
I had to change the suggestions from `PlayerHolder.isPlayerOpen()` to `PlayerHolder.getInstance().isPlayerOpen()` inside of all static methods.

z3r0r4:
unsure if i should merge the new commits of the dev branch into this pull request.
Also it seems that there are some conflicting files according to the automatic checks.

mhmdanas:
Since there are conflicts, yes, you should merge `dev` into your branch or rebase.

z3r0r4:
resolved conflicts
had to change one method call as it wasnt implemented yet?
although it seems to implemented in the dev branch...

app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
```diff
@@ -326,7 +326,7 @@ class FeedFragment : BaseStateFragment<FeedState>() {
         if (context == null || context.resources == null || activity == null) return
 
         val entries = ArrayList<StreamDialogEntry>()
-        if (PlayerHolder.getInstance().getType() != null) {
+        if (PlayerHolder.getInstance().isPlayerOpen) {
```
Stypox, line 329: This doesn't build. Please test on your machine before pushing here next time ;-)

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
z3r0r4: weird
I've build it containing that line multiple times on my machine, including just now.

Going to change it anyways if you had problems with it.
Just something to note.

app/src/main/java/org/schabi/newpipe/local/history/StatisticsPlaylistFragment.java
```diff
@@ -338,7 +338,7 @@ private void showStreamDialog(final StreamStatisticsEntry item) {
 
         final ArrayList<StreamDialogEntry> entries = new ArrayList<>();
 
-        if (PlayerHolder.getInstance().getType() != null) {
+        if (PlayerHolder.getInstance().isPlayerOpen()) {
             entries.add(StreamDialogEntry.enqueue);
```
Stypox, line 342: Also show it in the statistics fragment
```suggestion
            entries.add(StreamDialogEntry.enqueue);

            if (PlayerHolder.getInstance().getQueueSize() > 1) {
                entries.add(StreamDialogEntry.next);
            }
```

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -752,7 +753,7 @@ protected void showStreamItemDialog(final PlaylistStreamEntry item) {
 
         final ArrayList<StreamDialogEntry> entries = new ArrayList<>();
 
-        if (PlayerHolder.getInstance().getType() != null) {
+        if (PlayerHolder.getInstance().isPlayerOpen()) {
             entries.add(StreamDialogEntry.enqueue);
```
Stypox, line 757: Also show it in the local playlist fragment
```suggestion
            entries.add(StreamDialogEntry.enqueue);

            if (PlayerHolder.getInstance().getQueueSize() > 1) {
                entries.add(StreamDialogEntry.next);
            }
```

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -346,7 +334,7 @@ public static void openVideoDetailFragment(@NonNull final Context context,
 
         final boolean autoPlay;
         @Nullable final MainPlayer.PlayerType playerType = PlayerHolder.getInstance().getType();
-        if (playerType == null) {
+        if (PlayerHolder.getInstance().isPlayerOpen()) {
```
Stypox, line 337: You changed the meaning!
```suggestion
        if (!PlayerHolder.getInstance().isPlayerOpen()) {
```

app/src/main/res/values/strings.xml
```diff
@@ -418,6 +418,9 @@
     <string name="show_channel_details">Show channel details</string>
     <string name="enqueue_stream">Enqueue</string>
     <string name="enqueued">Enqueued</string>
+    <string name="enqueue_next_stream">Enqueue next</string>
+    <string name="enqueued_next">Enqueued next</string>
+    <string name="start_here_on_main">Start playing here</string>
```
Stypox, line 423: Why was this string added?
```suggestion
```

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -92,23 +93,28 @@ private NavigationHelper() {
...
-                .putExtra(Player.SELECT_ON_APPEND, selectOnAppend);
+                                                    @Nullable final PlayQueue playQueue) {
+        return getPlayerIntent(context, targetClazz, playQueue)
+                .putExtra(Player.ENQUEUE, true);
+    }
+
+    @NonNull
+    public static <T> Intent getPlayerEnqueueNextIntent(@NonNull final Context context,
+                                                 @NonNull final Class<T> targetClazz,
+                                                 @Nullable final PlayQueue playQueue) {
```
Stypox, line 112:
```suggestion
                                                        @NonNull final Class<T> targetClazz,
                                                        @Nullable final PlayQueue playQueue) {
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -608,15 +608,19 @@ public void handleIntent(@NonNull final Intent intent) {
             setPlaybackQuality(intent.getStringExtra(PLAYBACK_QUALITY));
         }
 
-        // Resolve append intents
-        if (intent.getBooleanExtra(APPEND_ONLY, false) && playQueue != null) {
-            final int sizeBeforeAppend = playQueue.size();
+        // Resolve enqueue intents
+        if (intent.getBooleanExtra(ENQUEUE, false)
+                && playQueue != null) {
```
Stypox, line 613:
```suggestion
        if (intent.getBooleanExtra(ENQUEUE, false) && playQueue != null) {
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -608,15 +608,19 @@ public void handleIntent(@NonNull final Intent intent) {
...
 
-            if ((intent.getBooleanExtra(SELECT_ON_APPEND, false)
-                    || currentState == STATE_COMPLETED) && newQueue.getStreams().size() > 0) {
-                playQueue.setIndex(sizeBeforeAppend);
-            }
+            return;
+
+        // Resolve enqueue next intents
+        } else if (intent.getBooleanExtra(ENQUEUE_NEXT, false)
+                && playQueue != null) {
```
Stypox, line 620:
```suggestion
        } else if (intent.getBooleanExtra(ENQUEUE_NEXT, false) && playQueue != null) {
```

Stypox (changes requested):

app/src/main/res/values/strings.xml
ShareASmile: @Stypox you decided to leave it as it is so that if it ever becomes useful again it's ready to be used  more than 2 years ago here [https://github.com/TeamNewPipe/NewPipe/pull/2368#issuecomment-497330767](url)
I still remember

mhmdanas: That link seems to give back a 404 error?

ShareASmile: Thank you for pointing out the error. I have edited it.

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
z3r0r4: no good, sorry about that
hopefully obvious that it was not intentional.

app/src/main/res/values/strings.xml
z3r0r4: > I still remember that.. (in a lighter way..)

Impressive.
Gone and forgotten now.

Stypox (changes requested):
Sorry for keeping asking for changes, I hope this is the last one. Just to be sure, please make sure that the `if (player is open)` **before enqueueing** is removed everywhere, and that the play next action is available in every long press menu of each fragment. Thank you :-)

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -544,10 +545,14 @@ public boolean onLongClick(final View v) {
 
         switch (v.getId()) {
             case R.id.detail_controls_background:
-                openBackgroundPlayer(true);
+                if (PlayerHolder.getInstance().isPlayerOpen()) {
+                    openBackgroundPlayer(true);
+                }
```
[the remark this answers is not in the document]
z3r0r4: I'd go with removing it as also suggested in [this comment of yours](https://github.com/TeamNewPipe/NewPipe/pull/6872#issuecomment-899794880) (for future reference).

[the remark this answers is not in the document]
z3r0r4: note for myself: the openBackgroundPlayer(boolean append) method works within the current videoDetailfragment and appends the current video if true.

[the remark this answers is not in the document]
z3r0r4: > Just to be sure, please make sure that the `if (player is open)` **before enqueueing** is removed everywhere,

I am assuming that this refers to those `if (player is open)` **before enqueueing** that are called by long pressing a button.
Which would imply changes to the LocalPlaylistFragment, PlaylistFragment and VideoDetailFragment (handleResult, onLongClick).

> and that the play next action is available in every long press menu of each fragment.

assuming that this refers to the FeedFragment, StatisticsPlaylistFragment, LocalPlaylistFragment, PlaylistFragment and their Dialogs; mostly called showStreamDialog and showStreamItemDialog

I will be applying the requested changes to the listed Fragments that I have discovered. If i have missed some please just notify me @Stypox.

P.S.; I intend on opening a new issue (or directly a PR) which merges and unifies all these into one extendable base (Fragment?, List?). Not quiet sure of the implementation, just pretty sure that it needs refactoring, since its a huge pain to work with. Indicated by how many times I've missed that code needs to be duplicated and added to another fragment as well.

z3r0r4:
I've checked the fragments mentioned in [this comment](https://github.com/TeamNewPipe/NewPipe/pull/6872#discussion_r707357075) and I could not find a dialog which doesnt include a play next action.

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -154,56 +160,38 @@ public static void playOnBackgroundPlayer(final Context context,
...
-    public static void enqueueOnPopupPlayer(final Context context, final PlayQueue queue,
-                                            final boolean selectOnAppend,
-                                            final boolean resumePlayback) {
-        if (!PermissionHelper.isPopupEnabled(context)) {
-            PermissionHelper.showPopupEnablementToast(context);
-            return;
+    public static void enqueueOnPlayer(final Context context, final PlayQueue queue) {
+        PlayerType playerType = PlayerHolder.getInstance().getType();
+        if (!PlayerHolder.getInstance().isPlayerOpen()) {
+            Log.e(TAG, "Enqueueing when no player is open; defaulting to background player");
```
mhmdanas, line 177: Maybe "but" would be clearer than "when"? Same below.

z3r0r4: indeed

Stypox (approved):
Thank you!

z3r0r4:
my pleasure.
Thank you for the helpful feedback!

Issue #4569, opened 2020-10-20
Back button behaviour with no backstack


opusforlife2:
### Checklist

- [x] I checked, but didn't find any duplicates of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want
The OS Back button should close Video Details instead of minimizing to the mini player if:
1. It is the last video in the backstack (or the only video the user opened).
2. The video is not open in the embedded player.

#### Is your feature request related to a problem? Please describe it

Coming from 0.19.8, a user would expect the Back button to close Video Details instead of minimizing it. This behaviour has naturally changed so that a video playing in the embedded player isn't accidentally stopped. But if the user has just opened Video Details, then there is nothing to preserve. So a Back button tap should close Video Details instead.

#### How will you/everyone benefit from this feature?

More familiar navigation with the Back button.

#### Problem

If you play in popup or background players, closing the Video Details (by swiping down and closing the mini player) currently closes them too. However, if you back out of the app, then open it again, you can see that the mini player is gone but playback continues. This implies that background/popup playback isn't tied to the mini player, so that behaviour could be changed.

TheAssassin:
This issue is the result of a discussion on IRC. I mentioned that it makes no sense to create a mini player in case a video isn't playing. This behavior is unexpected and counterintuitive. Instead, the mini player should only be created when leaving a video details page while the video is playing. Example cases in which this would make sense:

- returning to search results from video details (-> mini player in search results)
- clicking a suggested video in the video details (-> mini player in new video details, stopped as soon as the new video is playing)

The latter would be a new addition, but I think it would be nice to have. Right now, the switch from video details page 1 to video details page 2 is quite broken; the old video keeps playing until the new details are loaded, if the latter fails you get a broken page with "Network error", the old video is still playing and the back button does *not* bring you to the old video but the search result.

Also, something that makes absolutely no sense to me is this swiping up and down. I would suggest to get rid of this stuff. It's anything but intuitive. I was told that since the video details swipe up when you open them, it should be "obvious" that you can also swipe down to "minimize", but the opposite is true. To me, a simple history of activities switches would be more intuitive.

[...]

Issue #5590, opened 2021-02-15
[feature request] settings option to disable minimized video 


AwesomelyHumble:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### How will you/everyone benefit from this feature?

I Iike the fairly new addition of the minimize feature but I find a lot of times I back out of a video but do not want it to continue playing. Can you add a setting to toggle this feature on/off? It's essentially the feature that minimizes and keeps a video when you hit the back button. Thanks!

JCU0:
I agree, it's a very annoying feature because it pops up for every video. With "back" button. With a click on channel icon from a video info screen. Every time. Basically, I am watching non-musical videos, so minimized and background players aren't useful for me. I know, two fingers gesture is a thing but it is not ideal solution to close every video with two fingers.

**What I suggest** is the setting to invert this logic: two fingers gesture should start background player and everything else (back button, channel icon, etc) should just stop&close a video.

P.S. For official YT site and app, an automatic minimized player is the very annoying anti-feature too. It feels like you are loosing control over your devices because they are trying to play when you want them to stop.

mhmdanas:
@AwesomelyHumble if I understand your explanation correctly, you can disable the popup player being opened when leaving NewPipe by changing "Settings > Video and audio > Minimize on app switch" to None. I'll be closing this issue, but if I misunderstood your explanation please make a comment and I'll re-open the issue.

AwesomelyHumble:
@mhmdanas Thank you for looking into this issue. I believe you have it misunderstood as my settings currently reflect your suggestion. What I'm referring to is when a video is playing and you press the back button to return to the previous window (i.e. search results), and the playing video is continuing to play in a smaller window at the bottom of the screen (see attached photo). I don't know what this specific feature is called but it was added a few revisions ago, fairly recently. What I'm requesting is there be an option in the settings menu toggle this from happening. Sometimes I want the video to continue playing while I browse (for example music), but many times I do not. Thank you. 

![Screenshot_20210215-111758~2](https://user-images.githubusercontent.com/28993697/107986003-0a0c3000-6f80-11eb-81ef-0099feb70b24.jpg)

[...]

PR #6895, opened 2021-08-10
Selectable behavior of Back button

### LINKED ISSUE #4479: An option to disable the video history/queue control with the "back" button
### Checklist

- [x] I checked, but didn't find any duplicates of this issue in the repo. 

v0.20 appears to be keeping the history/queue of previously opened videos separately from the actual browsing history.
As discussed in https://github.com/TeamNewPipe/NewPipe/issues/4462 , this interferes with the normal behavior of the Back button, and may be inconvenient and counterintuitive.  
The solution would be to add an option to always treat the Back button the same way as it's treated in any web browsing application, and don't control the video queue with it.

Unless the minimized player is closed manually (which also stops any playback), Back button will always open previous video (even if autoplay is disabled, and user never actually watched the video) instead of going to the previous page (playlist/channel/video/search results).

#### Additional context
https://github.com/TeamNewPipe/NewPipe/issues/4462#issuecomment-706205009

This will allow to keep traditional navigation with Back button always leading to the previously opened page.

avently:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)

Made an option to change behavior of back button on video fragment. It's something that should be included into initial unified player PR. Now NewPipe users can choose what to do when back button pressed. There are six choices:
- open previous video from history (like now, stays default choice)
- open previous video only in case it was playing from the same playlist as current one
- hide to mini player regardless of what was in the history
- hide into the mini player if the stream was played, close otherwise
- close the player fragment completely but only if nothing is playing or only video is playing. Just hides the player to mini player in case audio or popup are playing
- close even in fullscreen.

Also I noticed that for some reason `equals` check is not working correctly currently. So I reimplemented it based on urls compare of inner streams (see the first commit).

Fixes #4479

#### APK testing 

#### Due diligence


SameenAhnaf:
Thanks for the amazing PR. If I am not wrong, #4569 should be linked with the PR.

"open previous video only in case it was playing from the same playlist as current one" option doesn't work properly on my device. Could you please fix it?

https://user-images.githubusercontent.com/79245641/128966314-e56cfa49-543a-45a7-9058-9bba0d7e07ae.mp4

opusforlife2:
@avently Is this intended to close #4569? If option 4 is meant for that, then I'll address it on that basis (I found some problems). If not, then no issues.

avently:
@SameenAhnaf 
I'll add

@opusforlife2
> @avently Is this intended to close #4569? If option 4 is meant for that, then I'll address it on that basis (I found some problems). If not, then no issues.

Address it, if it is needed for you

opusforlife2:
Okay, so for option 4, we ideally want the Back button to close the player directly only if backstack size is 0, as stated in #4569. Right now, this PR closes using Back button even if backstack size is non-zero. This happens even if a video is playing.

opusforlife2:
> I added two new options: hide and close

@avently Bug: Back button closes the player even if the backstack size is non-zero. 
- Open a video, but don't play it.
- Open a related video, but don't play that either.
- Tap the Back button.
- Player closed, losing the entire backstack.

What it should do in this case is rewind through the stack, just like the current release does.

To be clear, closing should only happen if both conditions are satisfied: `size of backstack == 0 && player is open == false`.

avently:
It was intentional. So what's the point of your idea then? It acts as the default behavior with exception to non-played first item in backstack?

SameenAhnaf:
I think, choosing "Close the player even in full screen" should put the app back in portrait on smartphone after clicking 'Back'. Could you please do something about it?

avently:
Check again, made these two changes you both asked

SameenAhnaf:
On my device, 'Close the main player even in full screen' is chosen and auto-rotation is disabled. 2 seconds of lag for changing rotation can be seen when I am trying to exit the video from full screen.

Such a lag is not found when getting back into main player both through 'Minimize' and 'Back' buttons. So, this lag is unexpected.

Could you please fix it?

opusforlife2:
@avently Now the opposite happens. I open a video but don't tap the thumbnail. I tap Back. The page goes into the mini player instead of closing. Since I opened only one video, it's the only one in the backstack, it should have closed directly.

> It acts as the default behavior with exception to non-played first item in backstack?

Yes, that's the idea.

avently:
> Such a lag is not found when getting back into main player both through 'Minimize' and 'Back' buttons. So, this lag is unexpected.

"The lag" is equal to 500ms, not two seconds. Player closes first, after that rotation happens. Without 500ms delay rotation mechanism skipping player closing. So nothing special here. If you see the 2 sec delay is because of slow rotation on debug apk. On release apk it will be much faster.


@opusforlife2 could not reproduce. I have disabled autoplay and after I open a video without clicking on thumbnail and then click back button, player closes. Make sure you selected Hide into the mini player if the stream was played, close otherwise

opusforlife2:
> Make sure you selected Hide into the mini player if the stream was played, close otherwise

Ah, indeed. I am a dum-dum. After deleting and installing the latest APK and importing my database, I forgot to set the option again. After setting it to this, the app works exactly as expected.

Thank you, @avently!

SameenAhnaf:
@avently As #5459 is merged, it's expected that users would like to have a complete 'Full screen mode'. A dedicated option is not necessary right now.

What I actually want to have is the ability to have a media player alike experience. Many users like me just need videos. As you can see in the video, there's no lag in closing the player of MX Player. I have no idea about coding side. But, if possible, please remove the rotation lag.

https://user-images.githubusercontent.com/79245641/131782506-a41d79ee-5f7b-4e77-a704-9c5c98110d74.mp4

opusforlife2:
> If I am not wrong, #4569 and #5590 should also be linked with the PR.

Indeed. Done. 👍

avently:
Guys, let me know when you'll be ready to merge this. I'll fix merge conflicts then

opusforlife2:
@avently Could you add the two new options into the PR description? Thanks.

avently:
@opusforlife2 done

skyGtm:
This PR also addresses the issue I had.

Issue: I am using NP v0.21.10
When autoplay is disabled
1. play some video
2. scroll down the comments
3. Press back button 
Comments jumps to the top, rather than minimizing the player.

But this PR fixes this. Thankyou. :tada: :tada:

app/src/main/res/values/strings.xml
```diff
@@ -579,6 +579,15 @@
     <string name="minimize_on_exit_none_description">None</string>
     <string name="minimize_on_exit_background_description">Minimize to background player</string>
     <string name="minimize_on_exit_popup_description">Minimize to popup player</string>
+    <!-- Back button behaviour in a fragment -->
+    <string name="back_behavior_title">Back press in a player</string>
+    <string name="back_behavior_summary">Action when clicking on a back button on a player screen — %s</string>
+    <string name="back_behavior_prev_any_description">Open previous video from a history</string>
+    <string name="back_behavior_prev_playlist_description">Open previous video from a history (playlist only)</string>
```
Stypox, line 586:
```suggestion
    <string name="back_behavior_prev_playlist_description">Open previous video only if belonging to a playlist</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -579,6 +579,15 @@
...
+    <string name="back_behavior_hide_description">Hide into the mini player</string>
```
Stypox, line 587:
```suggestion
    <string name="back_behavior_hide_description">Minimize into the mini player</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -579,6 +579,15 @@
...
+    <string name="back_behavior_hide_or_close_description">Hide into the mini player if the stream was played, close otherwise</string>
```
Stypox, line 588:
```suggestion
    <string name="back_behavior_hide_or_close_description">Minimize into the mini player if the stream is completed, close otherwise</string>
```
I am not sure I understood what this does exactly. If I understood correctly maybe this would be a better phrasing:
```suggestion
    <string name="back_behavior_hide_or_close_description">Close the video player if it is playing something, minimize into the mini player otherwise</string>
```

app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
```diff
@@ -526,7 +526,16 @@ public boolean equals(@Nullable final Object obj) {
...
+        if (size() != other.size()) {
+            return false;
+        }
+        for (int i = 0; i < size(); i++) {
+            // Check is based on URL
+            if (!streams.get(i).getUrl().equals(other.streams.get(i).getUrl())) {
+                return false;
+            }
+        }
+        return true;
```
Stypox, line 538: > Also I noticed that for some reason equals check is not working correctly currently. So I reimplemented it based on urls compare of inner streams (see the first commit).

Why this change? Is it essential to do it in this PR? What do you mean by "not working" and in which cases is this working better?
I guess it is correct to compare play queues based just on stream urls. But [a similar change](https://github.com/TeamNewPipe/NewPipe/commit/2a2c82e73bb2380c07ab848b185952ab183a5502#diff-6dcb0e241835a3f32ba00835e94c64b48b6faf84c0252769617507c1c2fd8371) on `PlayQueueItem`s caused [buffering problems](https://github.com/TeamNewPipe/NewPipe/issues/4940) on duplicate items in queue, finally fixed by #6712. That's why I'd prefer to move this to a different PR, where we could better analyze which differences are to be found in the app if the play queue comparation is changed.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -2115,6 +2166,7 @@ private void showExternalPlaybackDialog() {
      * Remove unneeded information while waiting for a next task
      * */
     private void cleanUp() {
+        toggleFullscreenIfInFullscreenMode();
```
Stypox, line 2169: This was added so that when the `VideoDetailFragment` is disposed/reused it always starts with the device orientation and not in fullscreen? Please add a comment

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -736,6 +736,18 @@ public boolean onBackPressed() {
             Log.d(TAG, "onBackPressed() called");
         }
 
+        final int behavior = PlayerHelper.getBackBehavior(activity);
+        if (behavior == PlayerHelper.BackBehavior.BACK_BEHAVIOR_CLOSE_FULLSCREEN) {
+            if (!isPlayerAvailable() || player.videoPlayerSelected()) {
+                bottomSheetBehavior.setState(BottomSheetBehavior.STATE_HIDDEN);
+                requireView().postDelayed(this::rotateScreenOnPhoneIfInLandscape, 500);
```
Stypox, line 743: This looks suspicious, how did you calculate the 500ms delay to make sure the rotation happens correctly?

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -746,18 +758,48 @@ public boolean onBackPressed() {
...
                 && player.getPlayQueue() != null
                 && player.videoPlayerSelected()
                 && player.getPlayQueue().previous()) {
             return true;
         }
+
+        if (behavior == PlayerHelper.BackBehavior.BACK_BEHAVIOR_PREV_PLAYLIST) {
+            bottomSheetBehavior.setState(BottomSheetBehavior.STATE_COLLAPSED);
+            return true;
+        }
```
Stypox, line 787: Could you explain how this code achieves the behavior you described? I see nothing checking whether the two videos belong to the same playlist
> open previous video only in case it was playing from the same playlist as current one

Stypox (changes requested):
The code looks neat! I'd add more comments though, e.g. above every `if (behavior == ...)`, explaining what the behavior does and why the code works the way it is. Thank you for this much appreciated change and sorry for not reviewing earlier :-)

app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
avently: I should write more miningfull description of commits, just forgot details about this one. 
If I remember correctly wrong equals checks was the reason of duplicated queues in backstack. It means that the same queue was added to backstack and when you click back you'll see the same stream. 
It's probably something that gives problems others didn't find the cause of.

I don't think the fix could give other problems because the idea of equals check is to be sure that queues are the same. In case of queue we care only about the URLs of all streams contained in a queue.

Bugginme:
Hey the default back button works great on touch screen but can the default behavior be changed on TVs to close the player fragment? Thanks for the work @avently

avently:
@Bugginme as far as I remember there is such option in this PR

Bugginme:
@avently oh thanks, this pull request fixes a few issues on TV

Stypox:
@avently do you need some more time or reviews or anything before proceeding with this PR?

avently:
@Stypox nope, just a couple of free hours. Will try to do it today/yesterday maybe

app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
avently: @Stypox see https://github.com/TeamNewPipe/NewPipe/pull/7910

avently:
Force-pushed my changes and packed them into one commit.

Here I removed the commit related to equals check because @Stypox asked about that. However this build (as well as `dev` branch) without this commit works bad. Looks like someone is ruined previous behavior and whenever I choose any video it gets added to the backstack twice which shouldn't happen. The fix is here: #7910 

About strings. Thanks to @Stypox for correcting me, I changed some of the strings to match his version of them.

Added more comments and explanations in some places.

Other parts of code are remain the same as before. Let me know if some unexpeced things happen (except duplicated videos in a queue).

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
avently: Oh, forgot to made something about that. This delay was just brute-forced. If it will not work for someone I'll try to find a better way.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
avently: Explained now

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
avently: Added a comment

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -765,6 +787,21 @@ public boolean onBackPressed() {
             return true;
         }
 
+        if (behavior == PlayerHelper.BackBehavior.BACK_BEHAVIOR_CLOSE) {
```
Stypox, line 790:
```suggestion
        // just like BACK_BEHAVIOR_CLOSE_FULLSCREEN, but the fullscreen case is handled just above here
        if (behavior == PlayerHelper.BackBehavior.BACK_BEHAVIOR_CLOSE) {
```

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -765,6 +787,21 @@ public boolean onBackPressed() {
...
+            if (!isPlayerAvailable() || player.videoPlayerSelected()) {
+                bottomSheetBehavior.setState(BottomSheetBehavior.STATE_HIDDEN);
+            } else {
+                // Since it's not a video playing we can't close the player
+                bottomSheetBehavior.setState(BottomSheetBehavior.STATE_COLLAPSED);
+            }
+            return true;
+        }
+
+        if (behavior == PlayerHelper.BackBehavior.BACK_BEHAVIOR_HIDE) {
```
Stypox, line 800:
```suggestion
        // just collapse the video detail fragment, whatever the player's state (but the fullscreen case is handled above)
        if (behavior == PlayerHelper.BackBehavior.BACK_BEHAVIOR_HIDE) {
```

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -773,10 +810,30 @@ public boolean onBackPressed() {
             return true; // no code here, as previous() was used in the if
         }
 
-        // That means that we are on the start of the stack,
+        // If we passed previous check at the top `player.getPlayQueue().previous()`
```
Stypox, line 813:
```suggestion
        // If we passed the previous check at the top, that is, `player.getPlayQueue().previous()` returned true,
```

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -773,10 +810,30 @@ public boolean onBackPressed() {
...
+        // then we can be sure that queue have no more streams in the history.
+        // If the user chose BACK_BEHAVIOR_PREV_PLAYLIST behavior
+        // we have to minimize the player into the mini player here
+        if (behavior == PlayerHelper.BackBehavior.BACK_BEHAVIOR_PREV_PLAYLIST) {
+            bottomSheetBehavior.setState(BottomSheetBehavior.STATE_COLLAPSED);
+            return true;
```
Stypox, line 819: I still do not understand why this code would cause the previous item from the playlist to be played. Also, how do you check if a play queue item belongs to a playlist, and if its previous item also belongs to a playlist? And where is the code that queues the previous item from the playlist?

Or maybe I didn't understand what you mean by playlist, do you mean play queue (though still the code would not explain that, either)?

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -773,10 +810,30 @@ public boolean onBackPressed() {
...
+        }
+
+        // That means that we are on the start of the stack
         if (stack.size() <= 1) {
             restoreDefaultOrientation();
-            return false; // let MainActivity handle the onBack (e.g. to minimize the mini player)
+
+            if (behavior == PlayerHelper.BackBehavior.BACK_BEHAVIOR_HIDE_OR_CLOSE) {
```
Stypox, line 826: Why is "BACK_BEHAVIOR_HIDE_OR_CLOSE" here and not just after "BACK_BEHAVIOR_HIDE"? Shouldn't the two act the same way, with the only difference that "BACK_BEHAVIOR_HIDE_OR_CLOSE" should close the player if nothing is playing?

Stypox (changes requested):
Added some comments and have a couple of questions (you can let me add the comments myself, just answer the questions if you don't have time)

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
avently: Yes, I mean PlayQueue that has a history of played items. 
See https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java#L45

It's a List that holds previously played items from this queue. 

Here https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L772
I call this method https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java#L506
which returns true in case it started to play previously played stream. If it isn't then it returns false and the body of `if` at https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L772 will be skipped.

Maybe would be better to rename `previous()` method to something like `playPreviousFromHistoryIfAny()` or something like that. Or maybe make two methods insteaf of `previous()` like `hasPrevious()` and `playPrevious()`. But I think having one method is better because calling `playPrevious` without checking `hasPrevious` (when no previous) would raise an exception which is not needed at all.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
avently: The idea is to have a backstack working (like in BACK_BEHAVIOR_PREV_ANY) but to allow to close mini player completely. With BACK_BEHAVIOR_PREV_ANY the mini player will be at the bottom even when video is paused. With this option you're able to hide it in this case.

That's why it is the last check and it is inside `f (stack.size() <= 1) {`, which guarantees that backstack is empty and nothing can be played.

avently:
@Stypox 
> Added some comments and have a couple of questions (you can let me add the comments myself, just answer the questions if you don't have time)

Answered on the questions. Agree with your comments, they are more detailed and still correct. So you can commit them if you like.

Stypox:
Here is a table explaining what each setting does. To me this table looks really really complex and many of the behaviors have nothing to do with the description. This is a possibly a source of many complaints from users: "I expected setting X to do YZ but instead it does AYBZC, I think this is a bug as it's not explained anywhere it should also do ABC!". So either we make the descriptions far more *descriptive*, or we simplify the conditions. I would go with the latter, since many of these conditions do not make sense to me and would seem completely arbitrary to a user reading the descriptive descriptions. What do you think @opusforlife2 @SameenAhnaf @litetex ?
<table>
  <tr>
    <th>Key</th>
    <th>Description</th>
    <th>Behavior (condition -> result; conditions coming first have priority)</th>
  </tr>
  <tr>
    <td>BACK_BEHAVIOR_PREV_ANY</td>
    <td>Open previous video from backstack of opened videos</td>
    <td><table>
      <tr><td>(video) player open and in fullscreen</td><td>Exit from fullscreen</td></tr>
      <tr><td>Video player open and there is a previous stream</td><td>Play previous stream</td></tr>
      <tr><td>There is a previous video in the backstack</td><td>Go to previous video in the backstack</td></tr>
      <tr><td>This is the last video in the backstack</td><td>Restore default orientation and let main activity handle the onBack</td></tr>
    </table></td>
  </tr>
  <tr>
    <td>BACK_BEHAVIOR_PREV_PLAYLIST</td>
    <td>Open previous video only if belonging to a playlist <i>(playlist is to be interpreted as play queue)</i></td>
    <td><table>
      <tr><td>(video) player open and in fullscreen</td><td>Exit from fullscreen</td></tr>
      <tr><td>Video player open and there is a previous stream</td><td>Play previous stream</td></tr>
      <tr><td>Otherwise</td><td>Collapse the bottom sheet</td></tr>
    </table></td>
  </tr>
  <tr>
    <td>BACK_BEHAVIOR_HIDE</td>
    <td>Minimize into the mini player</td>
    <td><table>
      <tr><td>(video) player open and in fullscreen</td><td>Exit from fullscreen</td></tr>
      <tr><td>Otherwise</td><td>Collapse the bottom sheet</td></tr>
    </table></td>
  </tr>
  <tr>
    <td>BACK_BEHAVIOR_HIDE_OR_CLOSE</td>
    <td>Minimize into the mini player if something is playing, close otherwise</td>
    <td><table>
      <tr><td>(video) player open and in fullscreen</td><td>Exit from fullscreen</td></tr>
      <tr><td>Video player open and there is a previous stream</td><td>Play previous stream</td></tr>
      <tr><td>There is a previous video in the backstack</td><td>Go to previous video in the backstack</td></tr>
      <tr><td>This is the last video in the backstack and (player not open or video player open, but with different play queue)</td><td>Restore default orientation and hide the bottom sheet</td></tr>
      <tr><td>This is the last video in the backstack and not (^)</td><td>Restore default orientation and let main activity handle the onBack</td></tr>
    </table></td>
  </tr>
  <tr>
    <td>BACK_BEHAVIOR_CLOSE</td>
    <td>Close the video player</td>
    <td><table>
      <tr><td>(video) player open and in fullscreen</td><td>Exit from fullscreen</td></tr>
      <tr><td>Player not open or video player selected (not fullscreen)</td><td>Close the player if needed and hide the bottom bar</td></tr>
      <tr><td>Background or popup player selected</td><td>Collapse the bottom sheet</td></tr>
    </table></td>
  </tr>
  <tr>
    <td>BACK_BEHAVIOR_CLOSE_FULLSCREEN</td>
    <td>Close the video player even in fullscreen</td>
    <td><table>
      <tr><td>Player not open or video player selected</td><td>Exit from fullscreen if needed, close the player if needed and hide the bottom bar</td></tr>
      <tr><td>Background or popup player selected</td><td>Collapse the bottom sheet</td></tr>
    </table></td>
  </tr>
</table>

SameenAhnaf:
@Stypox https://github.com/TeamNewPipe/NewPipe/issues/7973 is a better solution, I think. https://github.com/TeamNewPipe/NewPipe/issues/7974 for `BACK_BEHAVIOR_CLOSE_FULLSCREEN`

opusforlife2:
I think it is pertinent to ask: who asked for each of these different behaviours? For example, `BACK_BEHAVIOR_HIDE_OR_CLOSE` came from a discussion between @TheAssassin and I. I'm assuming some user or other has asked for each of the other options. Is that the case, @avently?

If so, we probably do need to have a discussion on simplification. Assassin and I came up with the `BACK_BEHAVIOR_HIDE_OR_CLOSE` solution because that looked like the major grievance for most users who complained about the Back button behaviour in the 0.19.x -> 0.20.x transition.

So we need such users to chime in about what issue they had, (or avently could link to such comments, if you remember them?) so that we can see if each option is actually needed. See next part.

-------------

It is possible that there are other solutions than modifying the Back button behaviour for some of the grievances users put out.

For example, the full screen one: `BACK_BEHAVIOR_CLOSE_FULLSCREEN`. Currently, it closes the video player completely, causing you to lose the backstack.

The 0.19.x behaviour was: open the video page -> open the video in fullscreen (use the current fullscreen button only for rotating the video between portrait/landscape) -> close the video, but stay on the video page. -- This would preserve the backstack if the option is modified this way.

So from my POV, this `BACK_BEHAVIOR_CLOSE_FULLSCREEN` setting, with the 0.19.x behaviour, could be combined with the "Start main player in full screen" option, into a single "Use main player in full screen only". Tapping a thumbnail on the video page would start the video in fullscreen, and tapping Back would close the video entirely, and take you back to the same state, where the video page is opened with the thumbnail, not the embedded portrait player.

This would mean the "Use main player in full screen only" would satisfy the users who preferred the 0.19.x behaviour. And at least one option would be removed from this setting.

avently:
@Stypox 
> To me this table looks really really complex and many of the behaviors have nothing to do with the description.

In this situation all descriptions need to be as precise as possible. I still not the best in English, so I rely on your comments. For me they look good. If you don't think so, then they need to be edited by someone who understands English well.

> So either we make the descriptions far more descriptive, or we simplify the conditions. I would go with the latter, since many of these conditions do not make sense to me and would seem completely arbitrary to a user reading the descriptive descriptions.

I would choose the first option. These conditions can meet someone's requirements. For example, someone wants the app acts like Mxplayer when you can view the video and close right after click on back button from fullscreen or from non-fullscreen. Someone wants to have a backstack for playlist only. Someone don't want to have the backstack at all even in playlists. 

Since it's quite easy to make and maintain all those choices (all the code is inside on function, and it's just `if` conditions), I prefer to have all of them. Obviously in my case I WILL NOT USE any of these settings but default behavior, since I love it so much. I made this PR just for those who have other requirements.

@opusforlife2


> who asked for each of these different behaviours? 

You can read this PR from the start and you'll find some queries. Others I found in other issues that people use for complaining about the new backstack behavior. 

Actually it doesn't matter who asked and asked or not at all. We have a possibility to make these all behaviors possible which means bigger part of people will find what they need the most.

As I said, the code is pretty easy to maintain. The only questionable thing is settings descriptions.

avently:
@Stypox are we waiting something?

opusforlife2:
Regardless of its ease of maintenance, I think it's a very bad idea to have such a complex behaviour table for a very basic and frequently used navigation button. I can already imagine the tonnes of bug reports where we keep having trouble understanding users' problems because they have some custom navigation flow. It will become a nightmare where we have to act as tech support more than an app development team. We have _already_ had troubles with the `Preferred open action` and `Minimize on app switch` options. I think those could be simplified too, but that's a separate issue.

Most apps which allow modifying Back button behaviour only allow 2, or max 3, options, and that too for very simple cases, like closing the app on double tap or single tap of the Back button.

We must be very conservative and only allow very minimal changes from the current navigation flow. My current stance on each option:
- `BACK_BEHAVIOR_PREV_PLAYLIST` is too specific a use-case. ❎ 
- `BACK_BEHAVIOR_PREV_ANY` is the current behaviour. ✅ 
- `BACK_BEHAVIOR_HIDE` is redundant, since swiping down does the same thing. ❎ 
- `BACK_BEHAVIOR_HIDE_OR_CLOSE`, in my (and Assassin's) opinion, should become the default behaviour, not a setting. As written in the related issue, this was a major complaint in the 0.19.x -> 0.20.0 transition. We arrived at this behaviour because it's a minor change from the current behaviour, AND because it takes us closer to 0.19.x behaviour, WITHOUT the risk of the user losing their backstack. ❎ 
- The above point would make `BACK_BEHAVIOR_CLOSE` redundant. It's also quite a specific use case which should be handled by swiping down only, since we've had this as default behaviour for over a year. ❎ 
- `BACK_BEHAVIOR_CLOSE_FULLSCREEN` should be handled like https://github.com/TeamNewPipe/NewPipe/pull/6895#issuecomment-1055482016 instead, which, now I realise, is what Sameen wrote in #7974. 😅 The people who asked for direct-to-full-screen were the ones unhappy with the new 0.20.0 behaviour and wanted the 0.19.x behaviour back. ❎

SameenAhnaf:
I think, I couldn't specify how #7973 will solve the related issues. There will be two lists of videos in the same video tab. *Auto-queue drop down menu* will be shown in the middle to differentiate the lists.

1. Manually queued: This section will show all the videos enqueued by the user. If any user selects any video from related videos or any other feed, that video will be kept in this section. Even if a user selects *Start playing* for such videos, previously selected videos won't be cleared. Previously selected videos will be just kept on the top. They will be only cleared if the user closes the mini player or selects `Clear queue` option.
2. Related videos: It'll suggest videos according to *Auto-queue drop down menu*. I gave a detailed explanation in that issue how this menu should work.

Now, you may ask how main queue and background queue should work as they queue videos differently. Well, there are already player buttons (Background, Play All, Popup) on the top of playlists and channels. If the user requires to queue all the videos, he will use these options. #5741 lets the user choose which videos to enqueue. But we may skip solving that issue for now as there's a temporary alternative at least.

avently:
@SameenAhnaf why do you post unrelated things here? Please, use your issue for these dilemmas

SameenAhnaf:
@avently Wasn't this PR supposed to hold video info for every single video? I suggest such videos' info to be accessible from play queue, not via `Back` button.

opusforlife2:
> Wasn't this PR supposed to choose where to apply history queue control?

@SameenAhnaf Ah, no. It's about allowing Back button behaviour modification to appease the users who were angry about 0.20.0's breaking change in navigation from 0.19.x. The backstack (history queue) control talk is a side effect, but not relevant to the main point of the PR.

Please go ahead with the off-topic marking, since you won't have to read through them slowly like I would.

Stypox:
@opusforlife2 so in your opinion we should keep only `BACK_BEHAVIOR_PREV_ANY` and `BACK_BEHAVIOR_HIDE_OR_CLOSE`?

SameenAhnaf:
@opusforlife2 Dude, I wasn't the commenter this time. 😅

Anyway, I am not actually in favor of `BACK_BEHAVIOR_HIDE_OR_CLOSE` as default behavior. If you look at the chart, you'll find that it's the most inconsistent behavior. It neither specifies what hiding or closing means. I personally suggest `BACK_BEHAVIOR_HIDE` as default option instead. If someone requires to save videos on backstack or `BACK_BEHAVIOR_PREV_ANY`, he can do so by scrolling up in play queue as I stated earlier for #7973.

Considering the scope of this PR, we can go on with *Close main player on `Back` button* setting for now. It'll have three options: 1. No, 2. Yes, 3. Also in full screen.

Stypox:
@opusforlife2 I don't understand... We can't replace the current behavior with the old one or people used to the current behavior will get angry

opusforlife2:
@SameenAhnaf Oops. I just typed "S" and entered the first username without looking. :P

> it's the most inconsistent behavior. It neither specifies what hiding or closing means.

Why does it have to? There would be no _options_ to read the descriptions _of_. The change will be made permanent. Only the blog post will mention it.

---

@Stypox Have you read #4569?

avently:
@opusforlife2 

> No. I'm saying we should not have a setting for the Back button at all.
BACK_BEHAVIOR_HIDE_OR_CLOSE should become the default behaviour. It is very slightly different from BACK_BEHAVIOR_PREV_ANY, but in a way that, as I've mentioned, we think will improve user experience greatly.

Am I understand correctly that idea? So you basically saying that me (who created this PR for helping others with their use-cases for back button) should NOT receive what I like the most (`BACK_BEHAVIOR_PREV_ANY`)? I just opened Google Translate for looking at how one word is translated to English, because I never needed to prononce it. The word is `impudence`. Or maybe `insolence` also. 

> Regardless of its ease of maintenance, I think it's a very bad idea to have such a complex behaviour table for a very basic and frequently used navigation button

Every this behavior cab be useful in some cases. The cases is more than 1.
When I say it's easy to maintain I mean really easy. You don't need to know anything about how it work, all this methods inside one place, no need to worry about that options.

> I can already imagine the tonnes of bug reports where we keep having trouble understanding users' problems because they have some custom navigation flow.

No need to imaging. Just open issues. There are already tons of stupid reports who knows why. Nothing will change.

> Most apps which allow modifying Back button behaviour only allow 2, or max 3, options, and that too for very simple cases, like closing the app on double tap or single tap of the Back button.

Most apps don't have so much functions like NewPipe. You can't just compare one feature and close eyes on other features in comparison. 

@Stypox you made a good chart! About @opusforlife2 point. I often agree with him but not in this case.
I agree with you that the conditions for all options can not be understood just by reading description in one line. But who cares? User can choose one option and check how it's going. Didn't like? Ok, he will choose another option. At least one option will be good for someone. So I don't see a problem. I often use software and don't know what dev wanted to say in options. But it don't stop me from checking every option when I interested in it.

If all these options would be a headache for developers, I would say that we need to focus only on the most useful for everyone. But it's not the case. Pretty easy to maintain.

opusforlife2:
> So you basically saying that me (who created this PR for helping others with their use-cases for back button) should NOT receive what I like the most

🤣 I'm not telling _you_ to implement anything, @avently. This discussion just happens to be taking place in your PR because you didn't open a new corresponding issue like the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md#before-starting-development) say. If you opened an issue before opening this PR, I would have been commenting there, not here.

> When I say it's easy to maintain I mean really easy.

It doesn't _matter_. Ease of maintenance is completely irrelevant to my point. I'm not looking at the developer's perspective _at all_.

> There are already tons of stupid reports

We don't need more, thank you.

> You can't just compare one feature and close eyes on other features in comparison.

Apples and oranges. We implement multiple services and provide integrated functionality for lots of service-specific things, like Youtube Mixes, Bandcamp Radio, etc. This is good and a USP of Newpipe.

This PR is about modifying a _basic, universal functionality_ and adding several complex navigation flows, which we will have to support _forever_.

avently:
> It doesn't matter. Ease of maintenance is completely irrelevant to my point. I'm not looking at the developer's perspective at all.

> This PR is about modifying a basic, universal functionality and adding several complex navigation flows, which we will have to support forever.

> we will have to support forever

Man, you probably need to rest:) Working hard this days?

mhmdanas:
@avently I think the point @opusforlife2 is trying to communicate is that maintenance costs in this case arise not from maintaining the code itself, but from maintaining the _complex set of behaviors_ that the code produces.

(To be clear, I'm not stating anything about my position, just clarifying @opusforlife2's)

peat80:
If the team is so concerned about too much options and arising support problems maybe there should be some sort of settings-level introduced to newpipe in the future. For example like KODI does. 🤔

avently:
@Stypox what is the desision? what's next?

Stypox:
@opusforlife2 why did you mark `BACK_BEHAVIOR_HIDE_OR_CLOSE` as :negative_squared_cross_mark:, if you think it should be added? Also, could you please provide the full conversation you had with TheAssassin? I don't understand why `BACK_BEHAVIOR_HIDE_OR_CLOSE` would be the best option for people coming from pre-0.19, since it's basically the same thing as `BACK_BEHAVIOR_PREV_ANY` except for when there is no previous stream in the backstack (in which case the bottom sheet is hidden instead of collapsed).

@avently after thinking more about this (sorry for the late reply):
- `BACK_BEHAVIOR_PREV_ANY` (the current behavior) and `BACK_BEHAVIOR_HIDE_OR_CLOSE` are really similar, I'd keep only one of the two. :x: :heavy_check_mark: 
- `BACK_BEHAVIOR_PREV_PLAYLIST` is too similar to `BACK_BEHAVIOR_PREV_ANY` and a too-specific option :x: 
- `BACK_BEHAVIOR_HIDE` makes sense, in my opinion :heavy_check_mark: 
- `BACK_BEHAVIOR_CLOSE` and `BACK_BEHAVIOR_CLOSE_FULLSCREEN` are really similar, I would keep only one of the two. If I remember correctly these two behaviors are the ones who best resemble pre-0.19. :x: :heavy_check_mark:

opusforlife2:
@Stypox Have you read #4569? I can try searching for the conversation on IRC, but the issue description and Assassin's comments there might be enough.

I don't really remember the logic in my head from when I was typing that out. Basically, I think that having an option list for the Back button is bad. Turning that cross into a tick doesn't really change that.

avently:
@Stypox 
So, what is the final list of options to keep?

BACK_BEHAVIOR_PREV_ANY
BACK_BEHAVIOR_HIDE
BACK_BEHAVIOR_CLOSE

This one?

I would say that both BACK_BEHAVIOR_CLOSE and BACK_BEHAVIOR_CLOSE_FULLSCREEN are useful. First is useful for pre 20.x lovers and second for Android TVs. Since there is no way currently to control mini player on TV, it would be useful to press Back and get a closed player view completelly, then returning back to channel feed or wherever. Without auto-close from fullscreen user would need to press back button twice (in case of BACK_BEHAVIOR_CLOSE is enabled).

But if we drop BACK_BEHAVIOR_CLOSE users on phone will not have an easy access to comments and video descriptions (in case of BACK_BEHAVIOR_CLOSE_FULLSCREEN is enabled) because on phones it is matter. On TVs it's not important at all.

So I prefer:

BACK_BEHAVIOR_PREV_ANY (default on phone)
BACK_BEHAVIOR_HIDE
BACK_BEHAVIOR_CLOSE
BACK_BEHAVIOR_CLOSE_FULLSCREEN (default on TV).

Make your choice with the team, I'll make another commit and we will merge this. Finally.)

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -95,6 +101,19 @@ public final class PlayerHelper {
...
+            BACK_BEHAVIOR_HIDE, BACK_BEHAVIOR_HIDE_OR_CLOSE,
+            BACK_BEHAVIOR_CLOSE, BACK_BEHAVIOR_CLOSE_FULLSCREEN})
+    public @interface BackBehavior {
+        int BACK_BEHAVIOR_PREV_ANY = 0;
+        int BACK_BEHAVIOR_PREV_PLAYLIST = 1;
+        int BACK_BEHAVIOR_HIDE = 2;
+        int BACK_BEHAVIOR_HIDE_OR_CLOSE = 3;
+        int BACK_BEHAVIOR_CLOSE = 4;
+        int BACK_BEHAVIOR_CLOSE_FULLSCREEN = 5;
+    }
```
litetex, line 115: I wouldn't name all values inside "BackBehavior" with the prefix "BACK_BEHAVIOR_" because it's redundant.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -267,6 +286,26 @@ public static int getMinimizeOnExitAction(@NonNull final Context context) {
...
+            return BACK_BEHAVIOR_PREV_ANY; // default
+        } else if (action.equals(context.getString(R.string.back_behavior_prev_playlist_key))) {
+            return BACK_BEHAVIOR_PREV_PLAYLIST;
+        } else if (action.equals(context.getString(R.string.back_behavior_hide_key))) {
+            return BACK_BEHAVIOR_HIDE;
+        } else if (action.equals(context.getString(R.string.back_behavior_hide_or_close_key))) {
+            return BACK_BEHAVIOR_HIDE_OR_CLOSE;
+        } else if (action.equals(context.getString(R.string.back_behavior_close_key))) {
+            return BACK_BEHAVIOR_CLOSE;
+        } else {
```
litetex, line 304: This else is useless and can be removed

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
avently: ALL interfaces in PlayerHelper are prefixed with their name. It's just made the same as the rest of the code

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
avently: With or without `else` branch I have to write `return BACK_BEHAVIOR_CLOSE_FULLSCREEN` anyway. By using else the code looks better than:
```
...
else if(....)

return BACK_BEHAVIOR_CLOSE_FULLSCREEN
```

avently:
After 5 days I'll celebrate 1 yeah of this PR

newhinton:
@avently Maybe you can remove some of the options to slim it down and easier review? There was a suggestion of what to keep and what not, and i guess most users would be fine if we can get rid of the backstack.

Maybe that would speed up things?

On the other hand, maybe the maintainers should just decide what do keep, since to me it seems the implementation is already thoroughly done and we just need a decision.

Stypox:
Yes, we just need a decision

avently:
@newhinton 

> Maybe you can remove some of the options to slim it down and easier review?

The problem is that there is no decision what to remove and what to keep.

Stypox:
Sorry @avently, we forgot about it once again. Our final decision is the following. Do you have some time to finally finish off this PR? Thanks ;-)

### The solution

The solution we agreed upon is to force #4569 (see below) on phones/tablets and force `BACK_BEHAVIOR_CLOSE` on Android TVs. Yeah, you read it right, there will be no "selectable behavior", but rather the current behavior will be improved. Here is why we think this would make more people happy without impacting too much on people who want different behaviors:
- People on Android TVs will get a behavior consistent with the platform they are on, since the backstack is almost unusable there, so I think they would be happy with only having `BACK_BEHAVIOR_CLOSE`.
- Some people may want the video detail fragment to be **minimized and not closed** when they press back after they just opened a (single) video. #4569 would not suit them, but a workaround would be to minimize the player by swiping it down, instead of pressing back. Some people will still be unhappy, though maybe not many, as most would just have to get used to the change.
- Some people may want the video detail fragment to **always be minimized** on back presses. I am not sure who really wants this, because at this point the backstack is useless, and, again, minimizing the video detail fragment by swiping down would yield the same result.
- Some people may want the video detail fragment to **always be closed** on back presses (pre 0.19 behavior): #4569 would partially suit them, at least when there is only one video in the backstack.

### Behavior on phones/tablets (#4569)

When a back press event occurs in the video detail fragment:
- IF video player open
  - IF fullscreen
    - -> *exit from fullscreen*
  - ELSE IF there is a previous video in the **backstack**
    - -> *play previous* video from backstack
  - ELSE (i.e. not fullscreen and no previous video in backstack)
    - -> stop playing the video and show the thumbnail on the video detail fragment
- ELSE IF background/popup player open
  - IF there is a previous video in the **backstack**
    - -> *open previous* video from backstack, without playing it
  - ELSE (i.e. no previous video in backstack)  
    - -> *minimize* the video detail fragment into the bottom sheet
- ELSE (i.e. no player open)
  - IF there is a previous video in the **backstack**
    - -> *open previous* video from backstack, without playing it
  - ELSE (i.e. no previous video in backstack)  
    - -> *completely close* the video detail fragment

### Behavior on Android TVs (`BACK_BEHAVIOR_CLOSE`)

When a back press event occurs in the video detail fragment:
- IF video player open
  - IF fullscreen
    - -> *exit from fullscreen*
  - ELSE (i.e. not fullscreen)
    - -> stop playing the video
- ELSE (i.e. no player open, or background/popup player open)
  - -> *completely close* the video detail fragment

newhinton:
Could you explain how such a complicated decision-tree is an improvement?

And what part exactly changes from the current behaviour?



My usecase for this pr is to actually get rid of the backstack, because 99% of the time my backstack is completely the opposite of what i want, and while i cant prove it i feel that i am not alone with that sentiment

Stypox:
> Could you explain how such a complicated decision-tree is an improvement?

It looks complicated, but it isn't really. It is just two layers deep: the first layer depends on which player is currently open, while the second layer depends on the backstack and has similar behaviors across all three player situations.

According to what @opusforlife2 said a while ago:
> I tried to remember more about the discussion I had with Assassin. The reason we proposed https://github.com/TeamNewPipe/NewPipe/issues/4569 was because a majority of complaints we were seeing at the time (I think it was multiple places, including, e.g. Reddit) were about breaking a simple navigation flow. Users who were just browsing the app without opening any video were forced to deal with "this stupid bottom bar thingy" each and every time they tried to go back. #4569 would make it so that unless there is a queue or active playback, the minimised player wouldn't bother the user anymore.

---

> And what part exactly changes from the current behaviour?

The fact that the video detail fragment is completely closed (and its backstack thrown away) if you press back when there is only one video in the backstack and it is not currently playing. @opusforlife2 correct me if I'm wrong ;-)

---

> My usecase for this pr is to actually get rid of the backstack, because 99% of the time my backstack is completely the opposite of what i want, and while i cant prove it i feel that i am not alone with that sentiment

Yes, you are not alone. The backstack is almost useless to me, too. https://github.com/TeamNewPipe/NewPipe/issues/4569 would suit you (and me) since it would prevent the backstack from building up. Say you play a video by tapping on it from e.g. trending: if you then want to stop playing that video and watch another one, you can just press back one time to exit fullscreen, one other time to stop the video, and one last time to completely close the video detail fragment. Next time you open a video you will be able to do the same thing without having the previous video in the backstack.

newhinton:
>Yes, you are not alone. The backstack is almost useless to me, too. https://github.com/TeamNewPipe/NewPipe/issues/4569 would suit you (and me) since it would prevent the backstack from building up. Say you play a video by tapping on it from e.g. trending: if you then want to stop playing that video and watch another one, you can just press back one time to exit fullscreen, one other time to stop the video, and one last time to completely close the video detail fragment. Next time you open a video you will be able to do the same thing without having the previous video in the backstack.


And if minimize the video A, and start a new one B, i would have a backstack of [A, B], right? Afaik, that should be the current behaviour. This is not my preferred solution but i guess its a good compromise. (I sometimes browse and switch while watching, or stop a video and come back waaaay later to open a different one, both will create a backstack that to me, logically, is "wrong")

opusforlife2:
> you can just press back one time to exit fullscreen, one other time to stop the video, and one last time to completely close the video detail fragment

No, that's 2 presses, not 3. If you tap Back to exit full screen, the video automatically pauses. Then it's just one more tap to close the backstack.

avently:
@Stypox 

>ELSE (i.e. not fullscreen and no previous video in backstack)
    -> stop playing the video and show the thumbnail on the video detail fragment


Why do we need this step? Why not to make this instead?

ELSE (i.e. not fullscreen and no previous video in backstack)
    -> minimize the video detail fragment into the bottom sheet

?

From my point of view it's a useless action and looks like a problem with the app as from the user's perspective. While hiiding to bottom sheet will show that it's THE LAST video in backstack, no more back pressing can happen andat the same time, the video is still active (no interrupt).


> play previous video from backstack

Does this action includes playing previous videos from playlistQueueItem?

Consider the following backstack:
```
- one video
- another video
- playlist queue item
-- first video in playlist
-- second video in playlist    <<<<--------  WE ARE HERE
```

Once user presses back button, what video should be played: `first video in playlist` or `another video`?

avently:
@Stypox

I reading your post again and again and still can't understand why can't we just integrate selectable behavior in the settings?
Having different default options for TV and for phone/tablet is a good thing and can easily be implemented on this PR's basis.
If for the user these options will be good, he will not even go to settings. If not good, the user will just try one-by-one every option until he finds what he like more (i'm talking about problem you mention earlier that it's hard to distinguish one option from another by reading it's name in the settings).

By integrating this PR you don't need a separate option for disabling/enabling backstack. It's just already included here for those who needs it.

We can remove one or two options from the list. For example, I can remove:
- BACK_BEHAVIOR_PREV_PLAYLIST.

It is useless. If the user likes the idea of backstack he will like the idea to play non-playlist too.

I can remove BACK_BEHAVIOR_CLOSE_FULLSCREEN too but it's useful for TV. Good to have the ability to start playing something in the feed and close it right after you want it without double pressing back button.

opusforlife2:
> Why not to make this instead?

@avently The point is to get rid of the bottom sheet ASAP when using Back button navigation. If the user wants to retain the bottom sheet for the last video in the backstack, they can always swipe down instead.

> Does this action includes playing previous videos from playlistQueueItem?

What is the `- playlist queue item` in your sample backstack? What separates it from the 1st and 2nd playlist videos? I thought the backstack would have been
```
- 1st video
- 2nd video
- 1st playlist video
- 2nd playlist video
```

> If for the user these options will be good, he will not even go to settings. If not good, the user will just try one-by-one every option until he finds what he like more

If after implementing the suggested behaviour, we still see user complaints about wanting to do navigation differently, then we can think about changing/adding navigation options in the future. For now, we're aiming to minimise disruption by going for the most needed navigation flows.

avently:
>What is the - playlist queue item in your sample backstack?

There is a technical difference in the implementation. It's not important in the context of discussion since the option with playlist will be removed anyway for the reason I said above.

>If after implementing the suggested behaviour, we still see user complaints about wanting to do navigation differently, then we can think about changing/adding navigation options in the future

I'm complain about wanting to do navigation differently right now. Your ideas leave me with nothing. As well as all people I did this PR for. What do you think about this?

opusforlife2:
> I'm complain about wanting to do navigation differently right now.

Isn't the current implementation already designed as per your idea of the ideal navigation flow? After all, you wrote the code for it... 😕

avently:
@opusforlife2 i'm talking about idea you and @Stypox  are discussing. Current flow in release version is brilliant for me. '

opusforlife2:
Okay, so the only difference between the current implementation and our suggestion is that on the very last backstack item, under very specific conditions, tapping Back should do something different than swiping down.

Stypox:
> can't we just integrate selectable behavior in the settings?

I have already said this multiple times. The back behaviour is a standard thing, not something that should be customizable. Also, it is really difficult to explain what happens exactly with a specific option, leading to maintenance burden as opus explained.

> Once user presses back button, what video should be played: first video in playlist or another video?

`another video`, since moving through the playlist should be done with the prev/next buttons in the player

avently:
@Stypox
>    ...
        ELSE (i.e. not fullscreen and no previous video in backstack)
            -> stop playing the video and show the thumbnail on the video detail fragment
    ...
    ELSE (i.e. no previous video in backstack)
            -> minimize the video detail fragment into the bottom sheet
      ... 
      ELSE (i.e. no previous video in backstack)
            -> completely close the video detail fragment


>  Also, it is really difficult to explain what happens exactly with a specific option, leading to maintenance burden as opus explained


Hm, I think that it's really difficult to explain what exactly happens here .^. even without any options. But current behaviour `PREV_ANY`) is pretty consistent in this situation: it always minimize the player into mini player in all three cases above. So whenever you do some movement you 100% sure what happens next. Same for other options in this PR. 
More options in this specific PR != more troubles for devs. Because it's so easy to follow in code.

The thought I noticed by talking with all here is that the backstack behaviour is not as easy as we want to. I mean, really, we can't have one behaviour that feels good to everyone or at least to many. And we can make it comfortable for them by giving more options. 

As from the maintaining complexity... Did you notice that this PR have only one file conflict after one year? And it's even in non-important file (not in fragment)! No complexity, no problems. As from the user's perspective, more options is what people find useful in NewPipe in comparison to official Youtube. It's how things work here

opusforlife2:
> More options in this specific PR != more troubles for devs. Because it's so easy to follow in code.

> Did you notice that this PR have only one file conflict after one year?

https://github.com/TeamNewPipe/NewPipe/pull/6895#issuecomment-1072665003

newhinton:
@opusforlife2 @avently What about this compromise:

We keep the current behavior. We dont change anything, we dont introduce a different behavior


Instead, we provide a simple toggle switch somewhere in the settings, that disables the backstack. (In technical terms: limit the size to 1).


From a maintenance standpoint: It adds next to nothing
From a behavioural standpoint: It only adds a little complexity, but not (really) to the back button.
It also should be more easy to communicate that than a full list of items.

Would that help/solve the issue? I am not exactly sure anymore what the desired goal is :D

opusforlife2:
@newhinton Well, you'll have to go back and read the issues that were created regarding navigation after the unified player was released. We are intending for the backstack/bottom sheet to be disabled for Android TV, where it is very difficult to use, but we want to keep it for handheld devices, where the backstack is very useful.

newhinton:
Yes and i understand that quite a few users are disagreeing with the usefulness of the backstack. 

Wouldn't my suggestion fix the original issue (#4479) without introducing a problematic amount of options?

opusforlife2:
We're not intending to introduce any options. The navigation flow should be uniform for all TV users, and separately for all phone/tablet users.

newhinton:
So, this pr is completely useless because there wont be a change to this anyway? To be honest, i do think introducing 3 or 5 different behaviours is over the top and untenable, but the backstack does annoy quite a few users as can be seen by the interest those issues&pr's gather. Even @Stypox agreed that for a part of the userbase the backstack is exactly the wrong behaviour. Why not expand on that in a reasonable manner?

opusforlife2:
We're trying to do exactly that.

avently:
Looks like we will not come to any decision here that will fit our needs. You don't want selectable behavior so this PR will not help anyway.
Closing it after one year.

newhinton:
This is very sad to see. I hope the maintainers find a way to integrate parts of this PR into the app. There is no reason to change behaviour that does not work properly once in a while, even breaking with long established behaviour.

fevers:
I think the choice not to add a settings screen, and instead to force users into your preferred workflow, will always be a mistaken and anti-user one.

burgrr:
Forcing users to accept settings you decided for them ahead of time is why everyone uses Revanced instead 🤣

burgrr:
> We're trying to do exactly that.

Lmao. More lies.

PR #6919, opened 2021-08-14
Add Show Channel Details where it's missing

### LINKED ISSUE #6155: No "show channel details" in long-press menu in feed or local playlist items
### Checklist

- [x] I am using the latest version - 0.21.2 RC/latest dev build 

1. Open any feed (e.g. What's New) or local playlist
2. Long press on an item

### Actual behaviour

There's no "show channel details" shown in the menu.

### Expected behavior

There should be a "show channel details" option in the meu.

### Additional context
https://github.com/TeamNewPipe/NewPipe/pull/5851#issuecomment-823188251

### Device info
This is platform-independent, but I filled this out for completeness.

 - Android version/Custom ROM version: Android 11
 - Device model: Galaxy A51

@opusforlife2

ktprograms:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)

- Add `Show Channel Details` to `Local Playlists`, `Watch History` and `Subscription Feeds`
- Had to update [StreamEntity](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/database/stream/model/StreamEntity.kt) to add a `uploader_url` column. Also added a database `Migration` and changed the version number to `4`.
- If you click `Show Channel Details` and the `uploader_url` is empty in the database, it fetches it using the `Newpipe Extractor` and saves it to the database.
- Had to add `kapt "org.xerial:sqlite-jdbc:3.34.0"` to `build.gradle` as a [workaround](https://issuetracker.google.com/issues/174695268#comment9) on M1 Macs. Doesn't seem to hurt the build on other machines so I left it in.

- Before:
Before
- After:
After

- Fixes #6155

#### TODO:
- Show some kind of loading indicator while loading the `uploader_url` from the `Newpipe Extractor`.
- Fix [AppDatabaseTest](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/androidTest/java/org/schabi/newpipe/database/AppDatabaseTest.kt) ([My SO question](https://stackoverflow.com/questions/68757574/android-room-partial-migration-testing))
<details open>
<summary>Options</summary>
<b>Option 1:</b> Just test migrating directly from `2` to `4`.
<br>
<b>Option 2:</b> Use <a href='https://developer.android.com/reference/androidx/sqlite/db/SupportSQLiteDatabase'>SupportSQLiteDatabase</a> <code><a href='https://developer.android.com/reference/androidx/sqlite/db/SupportSQLiteDatabase#query(java.lang.String)'>query</a></code> and <code><a href='https://developer.android.com/reference/android/database/DatabaseUtils#dumpCursorToString(android.database.Cursor)'>DatabaseUtils.dumpCusrorToString</a></code>.
<br>
<b>Option 3:</b> Hopefully someone knows the correct way to do it...
</details>

#### APK testing 

#### Due diligence


ktprograms:
#### Need to test:
Install the APK from GitHub Actions, then test `Show Channel Details` in:
1. Local Playlists
2. Watched History
3. Subscription Feed

Open Android Studio and connect to the app process, open `newpipe.db` and double click the `streams` table. Find the `uploader_url` column and delete the channel link for some videos. Then confirm that `Show Channel Details` still works (will take longer since it needs to call the `Newpipe Extractor`).

app/src/main/java/org/schabi/newpipe/database/Migrations.java
```diff
@@ -160,5 +161,14 @@ public void migrate(@NonNull final SupportSQLiteDatabase database) {
         }
     };
 
+    public static final Migration MIGRATION_3_4 = new Migration(DB_VER_3, DB_VER_4) {
+        @Override
+        public void migrate(@NonNull final SupportSQLiteDatabase database) {
+            database.execSQL(
+                    "ALTER TABLE streams ADD COLUMN uploader_url TEXT NOT NULL default ''"
```
XiangRongLin, line 168: 1. Does this design not duplicate the `uploader_url` over all the streams? Should it instead not be a it's own table and referenced here? One problem would be that currently `uploader` is also duplicated all over, so that would need to be migrated later on too or together here. 

---

2. I'm personally not a fan of empty default strings. Instead it should just be nullable and handled properly. For me an empty stream means that the url is an empty string, a null means that it does not exist

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -29,12 +33,35 @@
...
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(result -> {
+                        NewPipeDatabase.getInstance(fragment.getContext()).streamDAO()
+                                .setUploaderUrl(serviceId, url, result.getUploaderUrl())
+                                .subscribeOn(Schedulers.io()).subscribe();
+                        // For some reason `getParentFragmentManager()` doesn't work, but this does.
+                        NavigationHelper.openChannelFragment(
+                                fragment.requireActivity().getSupportFragmentManager(),
+                                item.getServiceId(), result.getUploaderUrl(),
+                                item.getUploaderName());
```
XiangRongLin, line 52: This block is duplicated. see line 60-64

app/build.gradle
```diff
@@ -214,6 +214,7 @@ dependencies {
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation "androidx.room:room-runtime:${androidxRoomVersion}"
     implementation "androidx.room:room-rxjava3:${androidxRoomVersion}"
+    kapt "org.xerial:sqlite-jdbc:3.34.0"
```
XiangRongLin, line 217: Does this have an impact on apk size?

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -29,12 +33,35 @@
...
+                    }, throwable -> Toast.makeText(
+                            fragment.getContext(),
+                            "Error at Show Channel Details",
```
XiangRongLin, line 55: No hardcoded strings, instead they should be declared in the strings.xml and reference here, so that it can be translated from weblate.

app/build.gradle
ktprograms: Can you please help me check? It won't build on a M1 Mac without it: https://issuetracker.google.com/issues/174695268#comment9

app/src/main/java/org/schabi/newpipe/database/Migrations.java
ktprograms: On point 2, that shouldn't be too hard to check for null instead of empty.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
ktprograms: Fixed in https://github.com/TeamNewPipe/NewPipe/pull/6919/commits/241054fd26a0a04433c3550420daa536e7b0093e

app/build.gradle
ktprograms: Nevermind, I used GitHub Actions.

```
15831094 14 Aug 08:39 app-debug-xerial.apk
15830642 14 Aug 08:36 app-debug.apk
```

So almost the same size.

https://github.com/ktprograms/NewPipe/actions

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
ktprograms: Fixed in https://github.com/TeamNewPipe/NewPipe/pull/6919/commits/bd0427c79f4474bc2c3194672d11d8223f7d4520

app/src/main/java/org/schabi/newpipe/database/Migrations.java
ktprograms: Point 2 fixed in https://github.com/TeamNewPipe/NewPipe/pull/6919/commits/21bcadeecbfa90e2bc66e77d3eb17163ae42ad55

ktprograms: I think Point 1 should be a different pr? Will probably need another database migration.

app/src/main/java/org/schabi/newpipe/database/playlist/PlaylistStreamEntry.kt
```diff
@@ -27,6 +27,7 @@ data class PlaylistStreamEntry(
         val item = StreamInfoItem(streamEntity.serviceId, streamEntity.url, streamEntity.title, streamEntity.streamType)
         item.duration = streamEntity.duration
         item.uploaderName = streamEntity.uploader
+        item.uploaderUrl = streamEntity.uploaderUrl ?: ""
```
XiangRongLin, line 30: This should not have a default value of an empty String assigned, because of the same issue as in the database.

app/build.gradle
XiangRongLin: Can you add a comment that the dependency is only as a workaround for M1 Max and link the thread of the underlying issue. So that when the underlying issue is solved, the dependecy can be removed.

app/src/main/java/org/schabi/newpipe/database/playlist/PlaylistStreamEntry.kt
ktprograms: How should it be done?

XiangRongLin:
> Fix AppDatabaseTest (My SO question)

For me that test has served it's purpose and can be deleted. An existing migration should never be changed afterwards =>code never changes => regression can never occur => tests are useless

Opinions? @Stypox

app/src/main/java/org/schabi/newpipe/database/Migrations.java
XiangRongLin: https://github.com/TeamNewPipe/NewPipe/issues/6924

app/src/main/java/org/schabi/newpipe/database/playlist/PlaylistStreamEntry.kt
XiangRongLin: Just simply remove the `?: ""`

ktprograms: The `StreamInfoItem` is defined in the extractor and it isn't `Nullable`

app/build.gradle
ktprograms: Fixed in https://github.com/TeamNewPipe/NewPipe/pull/6919/commits/65f2730261ae1c0cc544223caee666c4060d4f71

app/src/main/java/org/schabi/newpipe/database/playlist/PlaylistStreamEntry.kt
ktprograms: Fixed in https://github.com/TeamNewPipe/NewPipe/pull/6919/commits/6f1958d398f4a3a45d7f896bb03275fbaf497530

ktprograms:
When creating a `StreamEntity` from a `PlayQueueItem`, previously it would just save an empty string. I've edited `PlayQueueItem` to have the uploader url and save it to the `StreamEntity`.

Stypox:
> For me that test has served it's purpose and can be deleted. An existing migration should never be changed afterwards =>code never changes => regression can never occur => tests are useless

Mmmh, you are right. Though, we need to put a comment in the database migrations file explaining that a test must be created when creating a PR with a migration, and then be removed just before the PR is merged. Another option instead would be to always keep a migration test from the current version - 1 to the current version, so that is can be used as reference, can be easily edited and does not need to be created and then removed just afterwards (which may cause problems if e.g. something is changed after it is removed and reviewers do not notice).

ktprograms:
Seems like it would be a good idea to have the most recent migration test kept as a reminder / make sure there haven't been any database changes. I think any change in the database schema would make the test fail.

ktprograms:
@Stypox so shall I remove the v2 -> v3 migration test and add a v3 -> v4 test?

ktprograms:
Bare minimum passing test in https://github.com/TeamNewPipe/NewPipe/pull/6919/commits/02aa6fcab07d31566ed6f9d0d18ed9ba19366113, feel free to request changes.

XiangRongLin:
>Though, we need to put a comment in the database migrations file explaining that a test must be created when creating a PR with a migration, and then be removed just before the PR is merged

In a better world, you would not test the migrations, but test if the app is still working after applying the migrations.

In this case I don't see any benefit to add a test for the migration. If the feature works by trying it out manually, then the migration already succeeded. And it is simple enough, that having that manual try out is good enough.

Stypox:
Ok, I get your point @XiangRongLin. Let's remove the database test completely and test migrations manually from now on then. @ktprograms sorry for having you write a bare minimum test if it is going to be removed ;-). Could you add a comment just after `public final class Migrations {` that goes something along the lines of:
```
// Test new migrations manually by importing a database from daily usage and checking if the migration works
// If you add a migration point it out in the pull request, so that others remember to test it themselves
```
Thank you :-)

ktprograms:
> @ktprograms sorry for having you write a bare minimum test if it is going to be removed ;-)
No problem. Wasn't much trouble.

Changes made in https://github.com/TeamNewPipe/NewPipe/pull/6919/commits/967bdf8f08ef22c53789403835b9a5e2bb7ccf55

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -29,12 +33,28 @@
...
-        // For some reason `getParentFragmentManager()` doesn't work, but this does.
-        NavigationHelper.openChannelFragment(
-                fragment.requireActivity().getSupportFragmentManager(),
-                item.getServiceId(), item.getUploaderUrl(), item.getUploaderName())
-    ),
+    show_channel_details(R.string.show_channel_details, (fragment, item) -> {
+        if (isNullOrEmpty(item.getUploaderUrl())) {
+            final int serviceId = item.getServiceId();
+            final String url = item.getUrl();
+            // TODO: Some visual loading indicator
```
Stypox, line 40: I don't know what to suggest; maybe it can do without a loading indicator (similar to the "Mark as watched" action present in 0.21.9).

Stypox (changes requested):

Stypox (dismissed):
Sorry, didn't mean to request changes. This looks good to me, thanks :-)

ktprograms:
Merge?

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
ktprograms: I think there should be _some_ kind of indicator (even just a `Toast`)

Stypox:
Wait, I get this when importing my database:

## Exception
* __User Action:__ ui error
* __Request:__ ACRA report
* __Content Language:__ en
* __Service:__ none
* __Version:__ 0.21.8


```
java.lang.IllegalStateException: Migration didn't properly handle: feed_group(org.schabi.newpipe.database.feed.model.FeedGroupEntity).
 Expected:
TableInfo{name='feed_group', columns={name=Column{name='name', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=0, defaultValue='null'}, uid=Column{name='uid', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=1, defaultValue='null'}, icon_id=Column{name='icon_id', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='null'}, sort_order=Column{name='sort_order', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='null'}}, foreignKeys=[], indices=[Index{name='index_feed_group_sort_order', unique=false, columns=[sort_order]}]}
 Found:
TableInfo{name='feed_group', columns={uid=Column{name='uid', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=1, defaultValue='null'}, name=Column{name='name', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=0, defaultValue='null'}, icon_id=Column{name='icon_id', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='null'}, sort_order=Column{name='sort_order', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='0'}}, foreignKeys=[], indices=[]}
	at androidx.room.RoomOpenHelper.onUpgrade(RoomOpenHelper.java:103)
	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.onUpgrade(FrameworkSQLiteOpenHelper.java:177)
	at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:417)
	at android.database.sqlite.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:317)
	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.getWritableSupportDatabase(FrameworkSQLiteOpenHelper.java:145)
	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper.getWritableDatabase(FrameworkSQLiteOpenHelper.java:106)
	at androidx.room.RoomDatabase.inTransaction(RoomDatabase.java:622)
	at androidx.room.RoomDatabase.assertNotSuspendingTransaction(RoomDatabase.java:399)
	at androidx.room.RoomDatabase.query(RoomDatabase.java:442)
	at androidx.room.util.DBUtil.query(DBUtil.java:83)
	at org.schabi.newpipe.database.feed.dao.FeedDAO_Impl$13.call(FeedDAO_Impl.java:940)
	at org.schabi.newpipe.database.feed.dao.FeedDAO_Impl$13.call(FeedDAO_Impl.java:937)
	at io.reactivex.rxjava3.internal.operators.maybe.MaybeFromCallable.subscribeActual(MaybeFromCallable.java:47)
	at io.reactivex.rxjava3.core.Maybe.subscribe(Maybe.java:5330)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableFlatMapMaybe$FlatMapMaybeSubscriber.onNext(FlowableFlatMapMaybe.java:131)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableObserveOn$ObserveOnSubscriber.runAsync(FlowableObserveOn.java:402)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableObserveOn$BaseObserveOnSubscriber.run(FlowableObserveOn.java:176)
	at io.reactivex.rxjava3.internal.schedulers.ExecutorScheduler$ExecutorWorker$BooleanRunnable.run(ExecutorScheduler.java:322)
	at io.reactivex.rxjava3.internal.schedulers.ExecutorScheduler$ExecutorWorker.runEager(ExecutorScheduler.java:287)
	at io.reactivex.rxjava3.internal.schedulers.ExecutorScheduler$ExecutorWorker.run(ExecutorScheduler.java:248)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.lang.Thread.run(Thread.java:919)

```
</p></details>
<hr>

ktprograms:
> Wait, I get this when importing my database:

Weird. I didn't change anything in that file. How did you import your database? The `import` button in the NewPipe settings?

ktprograms:
@Stypox which version of NewPipe were you on when you exported your database?

ktprograms:
The import works fine for me (exported from latest v0.21.8, also has subscriptions).

I noticed there were several revisions to the v2 -> v3 migration and `FeedGroupEntity`. I think it's possible your database from one of those earlier versions. Could you try importing into for example the current `dev` branch apk?

Stypox:
That export was from 2020-12-25, so probably v0.20.6 or v0.20.7, and it imports correctly on `dev`.

ktprograms:
> That export was from 2020-12-25

Hmm, that's past all the changes to `Migrations.java`

I've got no idea why that error is happening.

ktprograms:
Does the database import code run migrations? It seems like it should since it restarts the app (so Room should notice).

ktprograms:
I can't reproduce the issue by exporting from v0.20.6 and importing to the GitHub actions apk.

XiangRongLin:
@Stypox I can't reproduce your issue.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -29,12 +33,29 @@
...
-        // For some reason `getParentFragmentManager()` doesn't work, but this does.
-        NavigationHelper.openChannelFragment(
-                fragment.requireActivity().getSupportFragmentManager(),
-                item.getServiceId(), item.getUploaderUrl(), item.getUploaderName())
-    ),
+    show_channel_details(R.string.show_channel_details, (fragment, item) -> {
+        if (isNullOrEmpty(item.getUploaderUrl())) {
+            final int serviceId = item.getServiceId();
+            final String url = item.getUrl();
+            Toast.makeText(fragment.getContext(), R.string.loading_channel_details,
```
ktprograms, line 40: Is this ok for a loading indicator?

ktprograms:
How can I comment on lines of code without doing a review?

opusforlife2:
@ktprograms There is a review type "Comment" meant just for this. You can select that, or choose to "Add a single comment" directly.

ktprograms:
> There is a review type "Comment" meant just for this. You can select that, or choose to "Add a single comment" directly.

Thanks, I clicked the "Add single comment" button so why does it show as `ktprograms reviewed xx hours ago`?

ktprograms:
@Stypox do you mind trying to import again (maybe from a different export) since this issue can't seem to be reproduced?

opusforlife2:
@ktprograms Because when you're commenting on the code you're also reviewing it, in essence? Does it matter, though? It's just semantics. The functionality is there for you to comment on code. How Github shows it or what it's called doesn't seem important to me.

Stypox (dismissed):
My database was missing a ``CREATE INDEX `index_feed_group_sort_order` ON `feed_group` (`sort_order`);}``, probably since I used that database in a temporary version of #2309 and then exported and re-imported it in a normal NewPipe release, and the correct migrations never ran since the database version was already up to date. So yes, ignore my crash report. I can confirm this works, thank you!

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -29,12 +33,29 @@
...
+                    Toast.LENGTH_SHORT).show();
```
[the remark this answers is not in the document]
Stypox: Yes, ok

Stypox:
Uhm, the build is failing @ktprograms

ktprograms:
Not sure why, seems to be a checkstyle error but it doesn't happen on my local repo.

app/build.gradle
```diff
@@ -214,6 +214,10 @@ dependencies {
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation "androidx.room:room-runtime:${androidxRoomVersion}"
     implementation "androidx.room:room-rxjava3:${androidxRoomVersion}"
+
+    // Apple Silicon Mac workaround (https://issuetracker.google.com/issues/174695268#comment9)
+    kapt "org.xerial:sqlite-jdbc:3.34.0"
+
```
Stypox, line 220: As in the others PRs, please remove this and only keep it locally
```suggestion
```

Stypox (changes requested):
Thank you!

app/build.gradle
ktprograms: Done in https://github.com/TeamNewPipe/NewPipe/pull/6919/commits/7e27e73532e3253a18a323421accccd4c662bd17

Stypox (approved):
Thank you!

PR #6930, opened 2021-08-15
Added plural forms for download related strings

### LINKED ISSUE #6850: Missing plural forms in download-related strings
### Checklist

Steps not available, it's just a problem with translation (code related).

In downloads, there are 2 strings that should have plural forms: `download_finished_more` and `deleted_downloads`. Currently, they are hard-coded to one form which is correct in English, but that's not the case in other languages (e.g. Polish). From what I've seen, these are only strings with that problem, others have proper plural forms.

This issue was already raised in comments in Weblate two months ago, but there was no response, and I haven't found an issue for this on GitHub.

### Actual behavior

`download_finished_more`, `deleted_downloads` have the same text no matter if there are 2, 4, 8 or any other number of elements.

### Expected behavior

These two strings should have an ability to provide plural forms like "few" and "other", so it would be possible to translate them properly to other languages.
For example, in Polish for "%s downloads finished" it should be: "2 pobrania ukończone" and "5 pobrań ukończonych" (two separate forms depending on items count).

N/A

### Logs

N/A

### Device info

 - Android version/Custom ROM version: -
 - Device model: -

Aga-C:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- added plural form to string for downloaded files notification
- added plural form to string for deleted files toast

There are no changes visible in English language, so I haven't done screenshots. However, changes will be visible for other languages, when translators will create new translations.

- Fixes #6850 

#### APK testing 

#### Due diligence


app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -226,6 +226,17 @@ public static String shortSubscriberCount(final Context context, final long subs
                 shortCount(context, subscriberCount));
     }
 
+    public static String downloadCount(final Context context, final int downloadCount) {
+        final int properCount = downloadCount + 1;
```
TobiGr, line 230: please explain, why you increase the count

TobiGr (commented):
Thank you!

app/src/main/java/org/schabi/newpipe/util/Localization.java
Aga-C: In the original code, text for one download was trigger when downloadCount was 0, and for higher amounts the count was incremented. Please check this line that I've removed from original code: https://github.com/TeamNewPipe/NewPipe/pull/6930/files#diff-22be96acbd15eb74cf0cd0f1addb47989eec8f95ae278582b5fe110629d3b6ebL489. I just moved it to this new location to have only one place with increasing the count.

TobiGr: That should be corrected when passing the parameter and not in the method, because you wrote a general utility method which can be used within other contexts, too, and thus potentially causing future usages of that method to display wrong download numbers.
On the other hand, I do not understand, why the `downloadDoneCount` is incremented at the end of the method. Incrementing it at the beginning would circumvent all of these problems. I think you can safely move [this last line](https://github.com/TeamNewPipe/NewPipe/pull/6930/files#diff-22be96acbd15eb74cf0cd0f1addb47989eec8f95ae278582b5fe110629d3b6ebL494) up to [here](https://github.com/TeamNewPipe/NewPipe/pull/6930/files#diff-22be96acbd15eb74cf0cd0f1addb47989eec8f95ae278582b5fe110629d3b6ebL469) and change the condition, That would be the result:

``` Java
        downloadDoneCount++;
        if (downloadDoneCount == 1) {
            downloadDoneList.append(name);
            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
                downloadDoneNotification.setContentTitle(getString(R.string.app_name));
            } else {
                downloadDoneNotification.setContentTitle(null);
            }

            downloadDoneNotification.setContentText(Localization.downloadCount(this, downloadDoneCount));
            downloadDoneNotification.setStyle(new NotificationCompat.BigTextStyle()
                    .setBigContentTitle(Localization.downloadCount(this, downloadDoneCount))
                    .bigText(name)
            );
        } else {
            downloadDoneList.append('\n');
            downloadDoneList.append(name);

            downloadDoneNotification.setStyle(new NotificationCompat.BigTextStyle().bigText(downloadDoneList));
            downloadDoneNotification.setContentTitle(Localization.downloadCount(this, downloadDoneCount));
            downloadDoneNotification.setContentText(downloadDoneList);
        }

        mNotificationManager.notify(DOWNLOADS_NOTIFICATION_ID, downloadDoneNotification.build());
    }
```

Aga-C:
@TobiGr Have you seen my latest changes? Is there anything more to fix?

TobiGr (approved):
Thank you

PR #6942, opened 2021-08-18
Add no-response workflow

mhmdanas:
#### What is it?

Add a workflow to help manage issues where the author hasn't responded to requests for more information. Here's an example:
1. I open an issue, but it doesn't give enough information
2. `@opusforlife` creates a comment asking me for the missing enough and labels the issue as `waiting-for-author`
3. If I reply before 14 days, the label is automatically removed. If I reply after 14 days, the label is automatically removed and the issue is re-opened.

I [opened an issue on the action's repo asking whether this works with PRs](https://github.com/lee-dohm/no-response/issues/171), but there's no reply yet. In any case, usage with PRs right now should be avoided, as 14 days is a far too short span of time. I plan to modify the action so we can have a separate `daysUntilClose` for PRs, but for now, this should be used with issues only.

I [created a repo with an identical workflow to the one added in this PR](https://github.com/mhmdanas/test-no-response-action). If anyone is interested in testing, feel free to ask and I'll send you an invite.

#### Due diligence

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -1,0 +1,20 @@
+name: No Response
+
+# Both `issue_comment` and `scheduled` event types are required for this Action
+# to work properly.
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # Schedule for five minutes after the hour, every hour
+    - cron: '5 * * * *'
+
+jobs:
+  noResponse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lee-dohm/no-response@v0.5.0
+        with:
+          token: ${{ github.token }}
+          daysUntilClose: 14
+          responseRequiredLabel: waiting-for-author
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -6,8 +6,8 @@ on:
   issue_comment:
     types: [created]
   schedule:
-    # Schedule for five minutes after the hour, every hour
-    - cron: '5 * * * *'
+    # Run daily at midnight.
+    - cron: '0 0 * * *'
 
 jobs:
   noResponse:


.github/workflows/no-response.yml
```diff
@@ -0,0 +1,20 @@
+name: No Response
+
+# Both `issue_comment` and `scheduled` event types are required for this Action
+# to work properly.
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # Schedule for five minutes after the hour, every hour
+    - cron: '5 * * * *'
```
mhmdanas, line 10: Maybe make it run every 12 hours (or even only once a day)? An interval of one hour between each run sounds unnecessary and wasteful to me.

Stypox:
I'd add `closeComment: false` so that the issue is closed without any comment that would ping anyone involved in the thread.

mhmdanas:
> that would ping anyone involved in the thread.

It doesn't though?

If you're referring to notifications, you get a notification even if the issue is only closed, so it doesn't matter much. If you're referring to actual pinging (á la `@`), then the comment doesn't contain any pings by default. I think the comment is useful for clarification on why the issue was closed, but I wouldn't entirely oppose removing it if that's the consensus.

Stypox:
> you get a notification even if the issue is only closed

Oh, then nevermind

.github/workflows/no-response.yml
TobiGr: once a day is good enough. We just want to remind a few people to answer

opusforlife2:
The run frequency should depend on how frequently the issue maintainers (mostly you) triage issues. If you do it every day, then the workflow should run every day or once in 2 days. This way, the max gap between you applying the label and the issue being closed would be 14-15 days. That is easily acceptable.

Compare this to a frequency of 5 days. The issue might get closed 19-20 days later.

mhmdanas:
Regardless of my triaging frequency, I think once per day is good enough.

Stypox (approved):

TobiGr (approved):
Looks good to me now.
Let's give it a try!

PR #6955, opened 2021-08-22
Show popup menu when long pressing in play queue (Full screen player)

### LINKED ISSUE #4778: UX: consistent behavior for long press in play queue
### Checklist

When I long press on a list item in the play queue that you get to by clicking on the little list icon in the top right of a playing video, the item gets deleted from the queue. When I long press on a list item in the play queue that you get to by clicking on the notification, while playing a video in pop-up mode, I get a context menu with options to delete, view details, add to playlist, or share. I would like to have both behaviors be consistent (with my personal preference being, that both show the context menu).

I sometimes accidentally delete a video from the queue when long pressing in the view that you get to by clicking the list icon from a playing video, because I confuse it with the list view that you get to by clicking on the notification while a video is playing in pop-up mode.

Less confusion. No accidentally deleted videos from play queue.

ktprograms:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Added `menu_play_queue_item.xml` as the resource to inflate the popup menu from.
- Refactored the existing popup menu code in the background/popup player queue to use the menu xml.
- Added the popup menu code to the main player queue.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td><video src="https://user-images.githubusercontent.com/71804605/130356012-cfe1daf4-ddf2-4e8e-9709-92756883e72c.mp4"/></td>
    <td><video src="https://user-images.githubusercontent.com/71804605/130356044-31f1da0d-6527-4b84-95ce-7f12f5a11f57.mp4"/></td>
  </tr>
</table>

- Fixes #4778 

#### APK testing 

https://github.com/TeamNewPipe/NewPipe/suites/3562943169/artifacts/85514049 (Link copied from the Checks tab)

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -214,10 +214,6 @@ dependencies {
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation "androidx.room:room-runtime:${androidxRoomVersion}"
     implementation "androidx.room:room-rxjava3:${androidxRoomVersion}"
-
-    // Apple Sillicon Mac workaround (https://issuetracker.google.com/issues/174695268#comment9)
-    kapt "org.xerial:sqlite-jdbc:3.34.0"
-
     kapt "androidx.room:room-compiler:${androidxRoomVersion}"
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
     implementation 'androidx.webkit:webkit:1.4.0'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -214,6 +214,10 @@ dependencies {
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation "androidx.room:room-runtime:${androidxRoomVersion}"
     implementation "androidx.room:room-rxjava3:${androidxRoomVersion}"
+
+    // Apple Sillicon Mac workaround (https://issuetracker.google.com/issues/174695268#comment9)
+    kapt "org.xerial:sqlite-jdbc:3.34.0"
+
     kapt "androidx.room:room-compiler:${androidxRoomVersion}"
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
     implementation 'androidx.webkit:webkit:1.4.0'
--- a/app/src/main/java/org/schabi/newpipe/QueueItemMenuUtil.java
+++ b/app/src/main/java/org/schabi/newpipe/QueueItemMenuUtil.java
@@ -23,8 +23,7 @@ public final class QueueItemMenuUtil {
                                      final View view,
                                      final boolean hideDetails,
                                      final FragmentManager fragmentManager,
-                                     final Context context,
-                                     final String tag) {
+                                     final Context context) {
         final ContextThemeWrapper themeWrapper =
                 new ContextThemeWrapper(context, R.style.DarkPopupMenu);
 
@@ -52,9 +51,9 @@ public final class QueueItemMenuUtil {
                             Collections.singletonList(item)
                     );
                     PlaylistAppendDialog.onPlaylistFound(context,
-                            () -> d.show(fragmentManager, tag),
+                            () -> d.show(fragmentManager, "QueueItemMenuUtil@append_playlist"),
                             () -> PlaylistCreationDialog.newInstance(d)
-                                    .show(fragmentManager, tag));
+                                    .show(fragmentManager, "QueueItemMenuUtil@append_playlist"));
                     return true;
                 case R.id.menu_item_share:
                     shareText(context, item.getTitle(), item.getUrl(),
--- a/app/src/main/java/org/schabi/newpipe/QueueItemMenuUtil.java
+++ b/app/src/main/java/org/schabi/newpipe/QueueItemMenuUtil.java
@@ -1,0 +1,71 @@
+package org.schabi.newpipe;
+
+import static org.schabi.newpipe.util.external_communication.ShareUtils.shareText;
+
+import android.content.Context;
+import android.view.ContextThemeWrapper;
+import android.view.View;
+import android.widget.PopupMenu;
+
+import androidx.fragment.app.FragmentManager;
+
+import org.schabi.newpipe.local.dialog.PlaylistAppendDialog;
+import org.schabi.newpipe.local.dialog.PlaylistCreationDialog;
+import org.schabi.newpipe.player.playqueue.PlayQueue;
+import org.schabi.newpipe.player.playqueue.PlayQueueItem;
+import org.schabi.newpipe.util.NavigationHelper;
+
+import java.util.Collections;
+
+public final class QueueItemMenuUtil {
+    public static void openPopupMenu(final PlayQueue playQueue,
+                                     final PlayQueueItem item,
+                                     final View view,
+                                     final boolean hideDetails,
+                                     final FragmentManager fragmentManager,
+                                     final Context context,
+                                     final String tag) {
+        final ContextThemeWrapper themeWrapper =
+                new ContextThemeWrapper(context, R.style.DarkPopupMenu);
+
+        final PopupMenu popupMenu = new PopupMenu(themeWrapper, view);
+        popupMenu.inflate(R.menu.menu_play_queue_item);
+
+        if (hideDetails) {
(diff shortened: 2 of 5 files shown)


ktprograms:
Just wondering, how does the Playback Speed Menu change it's color based on device dark mode?

Stypox:
Does the "Details" button work as expected, i.e. loading the details of another video without changing the playing stream

ktprograms:
I think It does change the playing stream. How can I open the details without replacing the stream?

Stypox:
> How can I open the details without replacing the stream?

I have no idea, probably it can't be done. I'd remove the "Details" action.

ktprograms:
Good idea. Come to think of it, Details doesn't make much sense when in the main player.

ktprograms:
@Stypox I added a `setVisible(false);` on the Details menu item.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3056,6 +3060,42 @@ public void selectQueueItem(final PlayQueueItem item) {
         playQueue.setIndex(index);
     }
 
+    private void heldQueueItem(final PlayQueueItem item, final View view) {
```
Stypox, line 3063: Create a helper class `org.schabi.newpipe.player.QueueItemMenuUtil` with a method `openPopupMenu(final PlayQueueItem item, final View view, final boolean hideDetails)` and use it in both `Player` and `PlayQueueActivity`.

Stypox (changes requested):

app/build.gradle
```diff
@@ -214,6 +214,10 @@ dependencies {
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation "androidx.room:room-runtime:${androidxRoomVersion}"
     implementation "androidx.room:room-rxjava3:${androidxRoomVersion}"
+
+    // Apple Silicon Mac workaround (https://issuetracker.google.com/issues/174695268#comment9)
+    kapt "org.xerial:sqlite-jdbc:3.34.0"
```
Stypox, line 219: Mmmh, what about not adding this? Would it be a problem for you to add this to your local workspace but to not commit it, so that you are able to build it on Silicon Mac but it does not impact the app's size and dependencies?

ktprograms: Done in https://github.com/TeamNewPipe/NewPipe/pull/6955/commits/ad99b54a39611a6968f74d2e2276f92105fd2c20.

Sorry for adding this previously.

app/src/main/java/org/schabi/newpipe/player/Player.java
ktprograms: If I'm not mistaken it'll need the fragment manager and context as parameters too?

Stypox: Yes, you are right

ktprograms: Done in https://github.com/TeamNewPipe/NewPipe/pull/6955/commits/a1cb3e59d6c5babcb79a6bedfc043494330b4cac.

Turns out I had to have 7 parameters:
- `playQueue`
- `item`
- `view`
- `hideDetails`
- `fragmentManager`
- `context`
- `tag` <- The PlaylistAppendDialog and PlaylistCreationDialog needs the class's `TAG`.

app/src/main/java/org/schabi/newpipe/QueueItemMenuUtil.java
```diff
@@ -0,0 +1,71 @@
...
+                            false);
+                    return true;
+                case R.id.menu_item_append_playlist:
+                    final PlaylistAppendDialog d = PlaylistAppendDialog.fromPlayQueueItems(
+                            Collections.singletonList(item)
+                    );
+                    PlaylistAppendDialog.onPlaylistFound(context,
+                            () -> d.show(fragmentManager, tag),
+                            () -> PlaylistCreationDialog.newInstance(d)
+                                    .show(fragmentManager, tag));
```
Stypox, line 57: You can just use these as tags. According to `FragmentTransaction#add(int, Fragment, String)`'s documentation, the tag param is an "Optional tag name for the fragment, to later retrieve the fragment with `FragmentManager#findFragmentByTag(String)`". I don't think we are ever doing this, but even if we were (or a library did) we would be safe since the playlist creation dialog would not be opened multiple times.
```suggestion
                            () -> d.show(fragmentManager, "QueueItemMenuUtil@append_playlist"),
                            () -> PlaylistCreationDialog.newInstance(d)
                                    .show(fragmentManager, "QueueItemMenuUtil@create_playlist"));
```
I'm taking inspiration from e.g. this: https://github.com/TeamNewPipe/NewPipe/blob/2928df0cc91897967ec2e1b00e7091f6b49f312e/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java#L99-L103

app/src/main/java/org/schabi/newpipe/QueueItemMenuUtil.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Collections;
+
+public final class QueueItemMenuUtil {
+    public static void openPopupMenu(final PlayQueue playQueue,
+                                     final PlayQueueItem item,
+                                     final View view,
+                                     final boolean hideDetails,
+                                     final FragmentManager fragmentManager,
+                                     final Context context,
+                                     final String tag) {
```
Stypox, line 27: So remove it from here

app/src/main/java/org/schabi/newpipe/player/PlayQueueActivity.java
```diff
@@ -280,46 +277,8 @@ private void buildControls() {
     }
 
     private void buildItemPopupMenu(final PlayQueueItem item, final View view) {
```
Stypox, line 279: I'd remove these two methods (`buildItemPopupMenu` and `heldQueueItem`) that do nothing but call another method. Just put `openPopupMenu` directly.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3198,9 +3204,8 @@ public void selected(final PlayQueueItem item, final View view) {
 
             @Override
             public void held(final PlayQueueItem item, final View view) {
-                final int index = playQueue.indexOf(item);
-                if (index != -1) {
-                    playQueue.remove(index);
+                if (playQueue.indexOf(item) != -1) {
+                    heldQueueItem(item, view);
```
Stypox, line 3208: Just as you did here, please also change
https://github.com/TeamNewPipe/NewPipe/blob/a2297fb5b8b7908ff3c2522dd5395b76fdfee763/app/src/main/java/org/schabi/newpipe/player/PlayQueueActivity.java#L372-L379
to
```java
if (player != null && player.getPlayQueue().indexOf(item) != -1) {
    openPopupMenu(/* ... */);
}
```

Stypox (changes requested):
Thanks :-)

ktprograms:
@Stypox all changes done in https://github.com/TeamNewPipe/NewPipe/pull/6955/commits/a4503eb6092d76c54c66edfb4a41719561e9ca9a

ktprograms:
@Stypox I just realised, if `PlayQueueActivity` checks if `getPlayQueue()` is null, shouldn't the popup menu in `Player.java` also check for null since they both are returned from the same property?

Stypox:
@ktprograms you are right, maybe it's better to make sure the play queue is not `null` before checking `.indexOf(item) != -1`, in both places

ktprograms:
@Stypox sorry, I think I misread the code previously. The code in `PlayQueueActivity` checks if `player` is not null, not if `getPlayQueue()` is null:
https://github.com/TeamNewPipe/NewPipe/blob/a4503eb6092d76c54c66edfb4a41719561e9ca9a/app/src/main/java/org/schabi/newpipe/player/PlayQueueActivity.java#L326

And since I haven't had any NPE yet, I think it can just be left as it is.

Stypox (approved):
I tested and it works well, thank you! :-)
Btw, you can resolve conversations when you are finished implementing them, it makes navigating the thread simpler ;-)

PR #6956, opened 2021-08-22
Only show "mark as watched" when watch history is enabled

TobiGr:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Only display the "mark as watched" item in the feed context menu when the watch history is enabled.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
@@ -349,7 +349,15 @@ class FeedFragment : BaseStateFragment<FeedState>() {
                 )
             )
         }
-        if (item.streamType != StreamType.AUDIO_LIVE_STREAM && item.streamType != StreamType.LIVE_STREAM) {
+
+        // show "mark as watched" only when watch history is enabled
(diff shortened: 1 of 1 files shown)


SameenAhnaf:
Thanks for the PR! Not sure if it's worth noting here. But there's a related bug.

'Show/Hide played items' icon works neither for now. Maybe, 'Mark as Watched' should work as 'Temporary watch history' instead if 'Watch history' is disabled.

TobiGr:
> 'Show/Hide played items' icon works neither for now.

@SameenAhnaf  Please tell us what exactly does not work and the used history settings and stream states (not watched, stream position somewhere in the middle, watched completely). I cannot reproduce that.

SameenAhnaf:
Sorry if I was not clear. I'm talking about this button. This button works neither when 'Watch history' is disabled. So, it should be hidden as well.

Should I open a separate issue?

IMG_20210823_012640

TobiGr:
It works for me when the history is disabled. One can have watched items in the history when the history is paused.
Note that when disabling the watch history it is _not_ cleared. We need to make this more obvious.

PR #6965, opened 2021-08-24
Add how to remove tab from main page text

### LINKED ISSUE #3400: add indication as to how "content of main page" is edited
Hi, I think it would be nice to add a small description at the bottom of the page "content of main page".

To be clear : I'm talking about the page in the `settings -> contents -> contents of the main page`

It is indeed not at all obvious that you can remove content via swiping laterally. It's made obvious you can add more via the "+" button and the "bar icon" indicate that we can reorder them. But the fact that one can swipe on the side to remove page is not obvious and I just realized today I could do it.

I think a small sentence in gray at the bottom of the page would serve this purpoes well : 
> swipe toward the side to hide content

This quick fix will make NewPipe's features more obvious to newcomers.

ktprograms:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Added a message on how to remove an item from the content of main page.

##### Before
Before

##### After
After

- Fixes #3400

#### APK testing 

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -214,10 +214,6 @@ dependencies {
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation "androidx.room:room-runtime:${androidxRoomVersion}"
     implementation "androidx.room:room-rxjava3:${androidxRoomVersion}"
-
-    // Apple Silicon Mac workaround (https://issuetracker.google.com/issues/174695268#comment9)
-    kapt "org.xerial:sqlite-jdbc:3.34.0"
-
     kapt "androidx.room:room-compiler:${androidxRoomVersion}"
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
     implementation 'androidx.webkit:webkit:1.4.0'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -214,6 +214,10 @@ dependencies {
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation "androidx.room:room-runtime:${androidxRoomVersion}"
     implementation "androidx.room:room-rxjava3:${androidxRoomVersion}"
+
+    // Apple Silicon Mac workaround (https://issuetracker.google.com/issues/174695268#comment9)
+    kapt "org.xerial:sqlite-jdbc:3.34.0"
+
     kapt "androidx.room:room-compiler:${androidxRoomVersion}"
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
     implementation 'androidx.webkit:webkit:1.4.0'
--- a/app/src/main/java/org/schabi/newpipe/settings/tabs/ChooseTabsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/tabs/ChooseTabsFragment.java
@@ -17,6 +17,7 @@ import android.widget.TextView;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
+import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.content.res.AppCompatResources;
 import androidx.appcompat.widget.AppCompatImageView;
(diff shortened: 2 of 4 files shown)


Stypox:
That text in that place looks odd. I would add a subtitle to the toolbar (this should be easily doable without adding new views, but just with `.setSubtitle();`) saying "Swipe items to remove them"

ktprograms:
I've moved it to the subtitle, but it seems if the text is too long it just cuts it instead of wrapping?

Text Cropped

Stypox:
@ktprograms what about "Swipe to remove"?

ktprograms:
> @ktprograms what about "Swipe to remove"?

@Stypox The cropping is only really a problem if the text is too long in different translations.

app/build.gradle
```diff
@@ -214,6 +214,10 @@ dependencies {
     implementation 'androidx.recyclerview:recyclerview:1.1.0'
     implementation "androidx.room:room-runtime:${androidxRoomVersion}"
     implementation "androidx.room:room-rxjava3:${androidxRoomVersion}"
+
+    // Apple Silicon Mac workaround (https://issuetracker.google.com/issues/174695268#comment9)
+    kapt "org.xerial:sqlite-jdbc:3.34.0"
+
```
mhmdanas, line 220: Please remove those changes; they are unrelated to this PR.

mhmdanas (dismissed):

app/build.gradle
ktprograms: Sorry, that's needed for me to build the code. Is there any way I can revert them in this PR? I'll avoid committing this next time.

mhmdanas: Just remove the lines manually and make a commit. We squash PRs here when merging them anyway.

ktprograms: Oh, I see. Will do.

ktprograms: Done in https://github.com/TeamNewPipe/NewPipe/pull/6965/commits/4f5e7721576faebf0b338152f1600cfb307d02da

Stypox: > We squash PRs here when merging them anyway.

Well, usually not, if (almost) every commit is meaningful. I only squash PRs in cases like this. But it's ok ;-)

Stypox:
> The cropping is only really a problem if the text is too long in different translations.

Mmmh, yes... I don't know what we should do. The fragment title would also be cropped in some locales, but we never cared about that as that's just the default android behaviour.

Maybe you could add a centered description text at the top of the activity with appropriate padding above and below, similarly to the "Edit each notification..." text at the top of the action button section in "Notification" settings.

ktprograms:
> The fragment title would also be cropped in some locales

Is there any plan to solve this in future?

Stypox:
> Is there any plan to solve this in future?

I don't think so, that's just the way Android works... But titles are usually small (i.e. single words) even when translated.

ktprograms:
> I don't think so, that's just the way Android works... But titles are usually small (i.e. single words) even when translated.

Ok, in that case I think I'll just follow your suggestion to have the text at the top and centered.

Stypox:
> Ok, in that case I think I'll just follow your suggestion to have the text at the top and centered.

Sorry, I don't get how this is related to "I don't [...] translated.". I was talking generally about the title in the toolbar that you can see in almost every activity and fragment throughout the app.

ktprograms:
I meant that the current approach works for titles since they're quite short, but for this help text I'll put it below the ActionBar.

Stypox:
Yes, and make sure the text can wrap

ktprograms:
@Stypox why does `fragment_choose_tabs.xml` use `RelativeLayout`? If there's no problem, I'd prefer to convert it to `ConstraintLayout` which I'm more familiar with.

ktprograms:
New Screenshot

opusforlife2:
Thanks! That looks great. I was thinking about the centre of the screen, but that would cause the text to be hidden if the user adds more tabs. This position is much better.

TobiGr (approved):
Thank you!

Stypox (approved):
Perfect, thank you! :-)

PR #6974, opened 2021-08-25
Update Android Gradle Plugin to version 7.0.2

litetex:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Updated ``com.android.tools.build:gradle`` to 7.0.1 (recommended through Android Studio)
- Release notes are available at: https://developer.android.com/studio/releases/gradle-plugin#7-0-0

#### APK testing 

#### Due diligence


litetex:
Update: The build requires now Java 11 so the GH-action workflows have to be adjusted

build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.2.2'
+        classpath 'com.android.tools.build:gradle:7.0.1'
```
TobiGr, line 11: I am asked to updated to `7.0.2`

TobiGr (commented):
Looks good to me.

build.gradle
litetex: Okay weird sentence....
I updated to 7.0.2 (although I don't find any changelogs for that version at https://developer.android.com/studio/releases/gradle-plugin#7-0-0).

Should be fine now 😄

TobiGr: asked by Android Studio :sweat_smile:

Stypox (approved):
I think we can switch to building with Java 11 without problems. I already build with JDK 11

TobiGr (approved):

opusforlife2:
> I did not find a changelog, too.

https://androidstudio.googleblog.com/2021/09/android-studio-arctic-fox-202031-patch.html

@litetex @TobiGr This, probably. Two issues under AGP.

Atemu:
This breaks compat with pure IDEA without google's ~~telemetry~~ android flavour.

See also https://youtrack.jetbrains.com/issue/IDEA-281657

Idea users need to revert this when they want to do anything.  
The AGP 4.2.2 version works just fine FYI, would it be acceptable to revert this until Idea supports AGP 7.0?

litetex:
> This breaks compat with pure IDEA without google's ~telemetry~ android flavour.
> 
> See also https://youtrack.jetbrains.com/issue/IDEA-281657
> 
> Idea users need to revert this when they want to do anything. The AGP 4.2.2 version works just fine FYI, would it be acceptable to revert this until Idea supports AGP 7.0?

This is fixed/resolved: https://youtrack.jetbrains.com/issue/IDEA-275441#focus=Comments-27-5223246.0-0

Atemu:
Only in the early access builds though.

It'd just be nice to be a bit more conservative with major revision bumps like these in the future then.

PR #6993, opened 2021-08-27
Close audio effect control session properly

### LINKED ISSUE #2373: NewPipe doesn't properly shut down audio streams when playback is finished/interrupted

NewPipe ends up leaving open the streams when it's done using them. Even worse, NewPipe doesn't recycle old streams, causing a memory leak to form in Android's internal mediaserver over time, as it requires explicit commands to be given in order to free memory used by audio it plays. On low-RAM devices, this can cause problems if the user starts playing multiple videos.

Basically, the way I found this out is via an equalizer app that I use due to my sensitivity with certain frequencies:
https://play.google.com/store/apps/details?id=com.devdnua.equalizer.free

In this app, I was exploring the settings to make sure everything was just the way I want them, and I found the "View Active Audio Sessions" area. NewPipe had a ton in this list, despite no actively-playing videos at the time, so I began to investigate. The summary of my findings with NewPipe is as follows:

- If nothing is currently playing, and the user starts a new video, NewPipe will open a new stream. This includes having a video/audio stream currently open, but stopped/paused.
- If there's currently something playing and another of the same type is enqueued, NewPipe will use the existing stream when the new one starts.
- If there's currently something playing and the user replaces it with another of the same type, a new stream is started.
- If audio is playing and the user starts up a video, NewPipe starts a new stream. Same goes for if video is playing and the user starts an audio stream.
- Switching between pop-up and full screen recycles the old stream.
- At no point does NewPipe actually close a given stream.
  - This includes when the user manually closes a stream, regardless of whether the stream is still playing or not.

I should note a few other things, as well:
- The equalizer app can't actually close the streams itself, as they are owned by NewPipe and the system's equalizer abilities only allow for a passthrough adjustment for apps like this, not a complete change of ownership. Using the "X" button in the active streams window only tells the equalizer to stop filtering. As a result, the memory leak will still occur even after I remove a stream from the list.
- Force Stopping NewPipe also does not get rid of the memory leak. I assume this is to allow for apps to resume using a given stream even after such a close. I've seen other apps such as BubbleUPNP and Vanilla Music do this.
- The only two ways available to users to get rid of the memory leak here are as follows:
  - Reboot the device.
  - Rooted users can alternatively kill the mediaserver by running `killall -9 mediaserver` as root in a terminal or something. This will kill all streams, including currently-playing ones, so if NewPipe is currently playing something and the terminal command is used, the audio will cut out and NewPipe will never be the wiser. Video will still play. However, it will be silent until NewPipe starts a new stream via one of the methods outlined above.

The recommended solution is to have NewPipe close out each stream when it's done playing, as well as have it recycle streams when shutting down currently-playing items in favor of new ones. Also, audio and video currently use separate streams, so replacing them so that only one notification/stream or the other can exist at one time would also be helpful.

Redirion:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

NewPipe would announce the audioSessionId once it is created for the ExoPlayer instance or when Audio gets enabled again after being disabled. However it did never announce closure of an audioSessionId. This is now done properly.

Apparently the missing closure of the audio effect control session could lead to memory leaks and cause problems with DSP / AudioFX / Equalizer apps. See also discussion in #2373 

- Fixes  #2373

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/helper/AudioReactor.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/AudioReactor.java
@@ -16,6 +16,7 @@ import androidx.media.AudioManagerCompat;
 
 import com.google.android.exoplayer2.SimpleExoPlayer;
 import com.google.android.exoplayer2.analytics.AnalyticsListener;
+import com.google.android.exoplayer2.decoder.DecoderCounters;
 
 public class AudioReactor implements AudioManager.OnAudioFocusChangeListener, AnalyticsListener {
 
@@ -50,6 +51,7 @@ public class AudioReactor implements AudioManager.OnAudioFocusChangeListener, An
     public void dispose() {
         abandonAudioFocus();
         player.removeAnalyticsListener(this);
+        notifyAudioSessionUpdate(false, player.getAudioSessionId());
     }
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -149,11 +151,21 @@ public class AudioReactor implements AudioManager.OnAudioFocusChangeListener, An
 
     @Override
     public void onAudioSessionId(final EventTime eventTime, final int audioSessionId) {
+        notifyAudioSessionUpdate(true, audioSessionId);
+    }
+
+    @Override
+    public void onAudioDisabled(final EventTime eventTime, final DecoderCounters counters) {
+        notifyAudioSessionUpdate(false, player.getAudioSessionId());
+    }
+
+    private void notifyAudioSessionUpdate(final boolean active, final int audioSessionId) {
         if (!PlayerHelper.isUsingDSP()) {
             return;
         }
-
-        final Intent intent = new Intent(AudioEffect.ACTION_OPEN_AUDIO_EFFECT_CONTROL_SESSION);
+        final Intent intent = new Intent(active
+                ? AudioEffect.ACTION_OPEN_AUDIO_EFFECT_CONTROL_SESSION
+                : AudioEffect.ACTION_CLOSE_AUDIO_EFFECT_CONTROL_SESSION);
         intent.putExtra(AudioEffect.EXTRA_AUDIO_SESSION, audioSessionId);
         intent.putExtra(AudioEffect.EXTRA_PACKAGE_NAME, context.getPackageName());
         context.sendBroadcast(intent);


app/src/main/java/org/schabi/newpipe/player/helper/AudioReactor.java
```diff
@@ -149,11 +151,21 @@ public void onAnimationEnd(final Animator animation) {
 
     @Override
     public void onAudioSessionId(final EventTime eventTime, final int audioSessionId) {
+        notifyAudioSessionUpdate(true, audioSessionId);
+    }
+
+    @Override
+    public void onAudioDisabled(final EventTime eventTime, final DecoderCounters counters) {
```
Stypox, line 158: Isn't there also a `onAudioEnabled`?

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/player/helper/AudioReactor.java
Redirion: there is, but it doesn't matter. When a new audioSessionId is generated, we have the method "onAudioSessionId" or "onAudioSessionIdChanged" in newer ExoPlayer versions. 

The only two reasons the audioSessionId gets reset is in onAudioDisabled and on destroy of the Player. 

So all cases are already handled with  onAudioSessionId, dispose and onAudioDisabled.

Btw. that the audiosessionId gets reset in onAudioDisabled is a bug in ExoPlayer. This commit here states what they wanted to do since 2.13.0: https://github.com/google/ExoPlayer/commit/a95b2ebb1d3125a4899c0dfe4a830077271ebca0

in 2.15.0 they found their mistake as in onAudioDisabled the audioSessionId doesn't get reset.

So we need to remember to remove "onAudioDisabled" when we update to 2.15. or later.

Redirion: oh no, they fixed it already in a newer version. So onAudioDisabled is not required actually.

https://github.com/google/ExoPlayer/commit/f52e4bed3d9bcde9fb380855c54c437b22b98b6f#diff-5078a6f4b57d380b958c26c0850edd29043d7a10f48b17bb172e55a840a52bb8

Redirion: wait: at this state of the repository it IS required. We need to remove it once #7005 is merged.

Redirion: made some edits to my comments so that they are clearer

Stypox: Ok, I see

Stypox (approved):
I also tested a little, and everything seems fine. The mute button also still works as expected, among other things. Thank you for taking some time to implement this @Redirion  :-)

PR #7036, opened 2021-09-01
Load full stream info when enqueuing a stream

### LINKED ISSUE #7035: Queued videos metadata not properly updating
### Checklist

- [x] I am using the latest version - d9086300f310dd5a508824244dd9b84b9398bcff (debug build) >= 0.21.9 

1. Play any video
2. Queue any video loaded in fast mode where the duration/metadata of the video has not been loaded (no duration on the thumbnail)
3. Skip to the queued video

### Actual behavior

The metadata while loaded for the video is not updated in the player: the progress bar ends at 00:-1, you are unable to seek and the queue time report does not include the videos duration. However if you minimize the video the duration of the video is displayed on the thumbnail after the queued video starts playing.

### Expected behavior

Like when you play a video without metadata directly from the feed the loaded metadata should be shown in the player: duration should be correct and seeking should be available.

### Device info

 - Android version/Custom ROM version: 11
 - Device model: N/A

Douile:
Please feel free to criticize the code here, I am not sure if I am using workers correctly or if there is a better way to do this. I essentially copied the code from https://github.com/TeamNewPipe/NewPipe/blob/faa7a91764cf633e668d9621eb6884b5db58bfa1/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L878-L906

I decided loading the information when adding to the queue is better than loading when the video starts playing as it allows the queue progress text to be correct as well. 

#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This PR calls getStreamInfo causing a full network fetch of stream info (I believe only if required) when adding a stream item to the queue. This should prevent UI issues of missing metadata when queuing videos that have been fast-loaded and are missing metadata.

- Fixes #7035

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
@@ -159,11 +159,6 @@ class FeedFragment : BaseStateFragment<FeedState>() {
         }
     }
 
-    // TODO: Move
-    fun redrawContent() {
-        groupAdapter.notifyItemRangeChanged(0, Int.MAX_VALUE)
-    }
-
     fun setupListViewMode() {
         // does everything needed to setup the layouts for grid or list modes
         groupAdapter.spanCount = if (shouldUseGridLayout(context)) getGridSpanCountStreams(context) else 1
--- a/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
@@ -159,6 +159,11 @@ class FeedFragment : BaseStateFragment<FeedState>() {
         }
     }
 
+    // TODO: Move
+    fun redrawContent() {
+        groupAdapter.notifyItemRangeChanged(0, Int.MAX_VALUE)
+    }
+
     fun setupListViewMode() {
         // does everything needed to setup the layouts for grid or list modes
         groupAdapter.spanCount = if (shouldUseGridLayout(context)) getGridSpanCountStreams(context) else 1
--- a/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
+++ b/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
@@ -237,7 +237,6 @@ public enum StreamDialogEntry {
                     item.getUrl(),
                     false
             )
-
                     .subscribeOn(Schedulers.io())
                     .observeOn(AndroidSchedulers.mainThread())
                     .subscribe(result -> {
--- a/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
+++ b/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
@@ -223,7 +223,7 @@ public enum StreamDialogEntry {
             final InfoCallback callback) {
         if ((item.getStreamType() == StreamType.LIVE_STREAM || item.getStreamType() == StreamType.AUDIO_LIVE_STREAM) && item.getDuration() < 0) {
             // Sparse item: fetched by fast fetch
-            final Disposable currentWorker = ExtractorHelper.getStreamInfo(
+            ExtractorHelper.getStreamInfo(
                     item.getServiceId(),
                     item.getUrl(),
                     false
--- a/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
+++ b/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
@@ -10,6 +10,7 @@ import androidx.fragment.app.Fragment;
 import org.schabi.newpipe.NewPipeDatabase;
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.extractor.stream.StreamInfoItem;
+import org.schabi.newpipe.extractor.stream.StreamType;
 import org.schabi.newpipe.local.dialog.PlaylistAppendDialog;
 import org.schabi.newpipe.local.dialog.PlaylistCreationDialog;
 import org.schabi.newpipe.local.history.HistoryRecordManager;
@@ -21,7 +22,6 @@ import java.util.Collections;
 import java.util.List;
 
 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
-import io.reactivex.rxjava3.disposables.Disposable;
 import io.reactivex.rxjava3.schedulers.Schedulers;
 
 import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
@@ -221,7 +221,9 @@ public enum StreamDialogEntry {
     private static void fetchItemInfoIfSparse(final Fragment fragment,
             final StreamInfoItem item,
             final InfoCallback callback) {
-        if ((item.getStreamType() == StreamType.LIVE_STREAM || item.getStreamType() == StreamType.AUDIO_LIVE_STREAM) && item.getDuration() < 0) {
+        if (!(item.getStreamType() == StreamType.LIVE_STREAM
+                || item.getStreamType() == StreamType.AUDIO_LIVE_STREAM)
+                && item.getDuration() < 0) {
             // Sparse item: fetched by fast fetch
             ExtractorHelper.getStreamInfo(
                     item.getServiceId(),
--- a/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
+++ b/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
@@ -62,13 +62,13 @@ public enum StreamDialogEntry {
      * Info: Add this entry within showStreamDialog.
      */
     enqueue(R.string.enqueue_stream, (fragment, item) -> {
-        fetchItemInfoIfSparse(item,
+        fetchItemInfoIfSparse(fragment, item,
                 fullItem -> NavigationHelper.enqueueOnPlayer(fragment.getContext(), fullItem)
         );
     }),
 
     enqueue_next(R.string.enqueue_next_stream, (fragment, item) -> {
-        fetchItemInfoIfSparse(item,
+        fetchItemInfoIfSparse(fragment, item,
                 fullItem -> NavigationHelper.enqueueNextOnPlayer(fragment.getContext(), fullItem)
         );
     }),
@@ -218,8 +218,9 @@ public enum StreamDialogEntry {
         void onInfo(SinglePlayQueue item);
     }
 
-    private static void fetchItemInfoIfSparse(final StreamInfoItem item,
-                                              final InfoCallback callback) {
+    private static void fetchItemInfoIfSparse(final Fragment fragment,
+            final StreamInfoItem item,
+            final InfoCallback callback) {
         if (item.getDuration() < 0) {
             // Sparse item: fetched by fast fetch
             final Disposable currentWorker = ExtractorHelper.getStreamInfo(
@@ -227,9 +228,19 @@ public enum StreamDialogEntry {
                     item.getUrl(),
                     false
             )
+
                     .subscribeOn(Schedulers.io())
                     .observeOn(AndroidSchedulers.mainThread())
                     .subscribe(result -> {
+                        final HistoryRecordManager recordManager =
+                                new HistoryRecordManager(fragment.getContext());
+                        recordManager.saveStreamState(result, 0)
+                                .subscribeOn(Schedulers.io())
+                                .observeOn(AndroidSchedulers.mainThread())
+                                .doOnError(throwable -> Log.e("StreamDialogEntry",
+                                        throwable.toString()))
+                                .subscribe();
+
                         callback.onInfo(new SinglePlayQueue(result));
                     }, throwable -> Log.e("StreamDialogEntry", throwable.toString()));
         } else {
--- a/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
+++ b/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
@@ -62,30 +62,23 @@ public enum StreamDialogEntry {
      * Info: Add this entry within showStreamDialog.
      */
     enqueue(R.string.enqueue_stream, (fragment, item) -> {
-        fetchItemInfoIfSparse(fragment, item,
-                fullItem -> NavigationHelper.enqueueOnPlayer(fragment.getContext(), fullItem)
-        );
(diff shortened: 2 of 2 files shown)


ktprograms:
I think how you're doing it now will always cause a full network fetch. The best way would probably be to check if (for example) the `item.getDuration()` is `null`, and only then doing the fetch. If it isn't `null`, you can just keep the old code (so have an `if` / `else` statement).

Here's the before and after of a similar case of adding the network fetch I recently implemented:
#### Before:
https://github.com/TeamNewPipe/NewPipe/blob/2b41f710a8cfa2ce8237de4a36bcba4b13dca78c/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java#L32-L37
#### After:
https://github.com/TeamNewPipe/NewPipe/blob/faa7a91764cf633e668d9621eb6884b5db58bfa1/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java#L36-L59

<hr/>

(In the after, `openChannelFragment` is a helper method that calls `NavigationHelper.openChannelFragment`):
https://github.com/TeamNewPipe/NewPipe/blob/faa7a91764cf633e668d9621eb6884b5db58bfa1/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java#L209-L216

GneHHHc233:
Hey, brother, thank you for solving the problem that my device YouTube can't get

Stypox:
> The best way would probably be to check if (for example) the item.getDuration() is null, and only then doing the fetch. If it isn't null, you can just keep the old code (so have an if / else statement).

Yes, good idea. The stream info would anyway be fetched when the stream is loaded and played, but not fetching it beforehand is a good idea nonetheless.

Douile:
Changes made in #6872 will conflict with this PR so I will wait for it to be merged before continuing here.

Stypox:
> getStreamInfo does check the cache

There are "three" "stages" of stream loading:
1. stream info *item* loaded with fast feed, with some metadata fields, thus missing duration field
2. stream info *item* loaded with any other method (search, trending, ... including slow feed), with all metadata fields
3. stream info with all of the data needed to play

Stream info *item*s are those loaded alongside other stream info items (i.e. when fetching a list, e.g. search results, trending, feed, ...), while stream info is loaded e.g. when a video is opened or played.

If when calling `getStreamInfo` the 3rd type of stream loading was already done before, then the cache kicks in, otherwise it doesn't, since `getStreamInfo` aims to collect all of the data needed to play a video. So preventing an early call to `getStreamInfo` would be nice, as the cache would not kick in for most stream info items, unless they have already been opened.

Stypox:
> Changes made in #6872 will conflict with this PR so I will wait for it to be merged before continuing here.

Douile:
When new info is fetched the feed item still doesn't update, I'm not sure where to trigger this update.

Douile:
Ayyy found it, https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L975-997

Douile:
Nvm that was the wrong thing...

Douile:
lmk if this needs a refactor because I'm not sure if it the best way of doing.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -203,4 +209,42 @@ private static void openChannelFragment(final Fragment fragment,
...
+    /////////////////////////////////////////////
+
+    private interface InfoCallback {
+        void onInfo(SinglePlayQueue item);
+    }
+
+    private static void fetchItemInfoIfSparse(final Fragment fragment,
+            final StreamInfoItem item,
+            final InfoCallback callback) {
+        if (item.getDuration() < 0) {
```
Stypox, line 224: The duration would be `< 0` even for live stream, but in that case fetching the info before enqueueing would not be any useful since the duration would still be `< 0`. So I'd do it like this:
```suggestion
        if ((item.getStreamType() == StreamType.LIVE_STREAM || item.getStreamType() == StreamType.AUDIO_LIVE_STREAM) && item.getDuration() < 0) {
```

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -60,11 +62,15 @@
...
+                fullItem -> NavigationHelper.enqueueOnPlayer(fragment.getContext(), fullItem)
+        );
     }),
 
     enqueue_next(R.string.enqueue_next_stream, (fragment, item) -> {
-        NavigationHelper.enqueueNextOnPlayer(fragment.getContext(), new SinglePlayQueue(item));
+        fetchItemInfoIfSparse(fragment, item,
+                fullItem -> NavigationHelper.enqueueNextOnPlayer(fragment.getContext(), fullItem)
+        );
     }),
```
Stypox, line 74: Why did you only do it for enqueue? I also get the problem of the missing duration field in the play queue when I press on "Start playing on ..."

Stypox (changes requested):
I can confirm it works when enqueueing a stream whose duration is unknown, thank you!

I get this crash when I long press an item in the fast feed, press on "Start playing in the background" and just after that I long press on another item. (both items are without duration before long pressing).

## Exception
* __User Action:__ ui error
* __Request:__ ACRA report
* __Content Language:__ en
* __Service:__ none
* __Version:__ 0.21.13


```
java.lang.NullPointerException: Attempt to invoke virtual method 'int org.schabi.newpipe.player.playqueue.PlayQueue.size()' on a null object reference
	at org.schabi.newpipe.player.helper.PlayerHolder.getQueueSize(PlayerHolder.java:74)
	at org.schabi.newpipe.local.feed.FeedFragment.showStreamDialog(FeedFragment.kt:336)
	at org.schabi.newpipe.local.feed.FeedFragment.access$showStreamDialog(FeedFragment.kt:80)
	at org.schabi.newpipe.local.feed.FeedFragment$listenerStreamItem$1.onItemLongClick(FeedFragment.kt:395)
	at com.xwray.groupie.GroupieViewHolder$2.onLongClick(GroupieViewHolder.java:32)
	at android.view.View.performLongClickInternal(View.java:7339)
	at android.view.View.performLongClick(View.java:7297)
	at android.view.View.performLongClick(View.java:7315)
	at android.view.View$CheckForLongPress.run(View.java:27850)

```
</p></details>
<hr>

app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
```diff
@@ -159,6 +159,11 @@ class FeedFragment : BaseStateFragment<FeedState>() {
         }
     }
 
+    // TODO: Move
+    fun redrawContent() {
+        groupAdapter.notifyItemRangeChanged(0, Int.MAX_VALUE)
+    }
```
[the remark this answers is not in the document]
Douile: Ah forgot to remove this

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
Douile: Ah, for some reason I neglected to check those options :)

Stypox:
In order to fix the exception I posted above I think you just need to alter the `PlayerHolder.getQueueSize()` method to something along these lines
```java
if (!isPlayerOpen()) return 0;
final PlayQueue playQueue = player.getPlayQueue();
return playQueue == null ? 0 : playQueue.size();
```

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -60,20 +62,31 @@
      * Info: Add this entry within showStreamDialog.
      */
     enqueue(R.string.enqueue_stream, (fragment, item) -> {
-        NavigationHelper.enqueueOnPlayer(fragment.getContext(), new SinglePlayQueue(item));
+        fetchItemInfoIfSparse(fragment, item,
+                fullItem -> NavigationHelper.enqueueOnPlayer(fragment.getContext(), fullItem)
```
Stypox, line 66: I think all lambda parameters should lie on the same line (also below) ;-)
```suggestion
        fetchItemInfoIfSparse(fragment, item, fullItem ->
                NavigationHelper.enqueueOnPlayer(fragment.getContext(), fullItem)
```

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -60,20 +62,31 @@
...
+    }),
 
-    start_here_on_popup(R.string.start_here_on_popup, (fragment, item) ->
-            NavigationHelper.playOnPopupPlayer(fragment.getContext(),
-                    new SinglePlayQueue(item), true)),
+    start_here_on_popup(R.string.start_here_on_popup, (fragment, item) -> {
+        fetchItemInfoIfSparse(fragment, item, fullItem -> {
+                    NavigationHelper.playOnPopupPlayer(fragment.getContext(),
+                            fullItem, true);
+                });
```
Stypox, line 88: This is the correct formatting, though the braces are unneeded
```suggestion
        fetchItemInfoIfSparse(fragment, item, fullItem ->
                NavigationHelper.playOnPopupPlayer(fragment.getContext(), fullItem, true));
```

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -203,4 +216,44 @@ private static void openChannelFragment(final Fragment fragment,
...
+        if (!(item.getStreamType() == StreamType.LIVE_STREAM
+                || item.getStreamType() == StreamType.AUDIO_LIVE_STREAM)
+                && item.getDuration() < 0) {
+            // Sparse item: fetched by fast fetch
+            ExtractorHelper.getStreamInfo(
+                    item.getServiceId(),
+                    item.getUrl(),
+                    false
+            )
+
```
Stypox, line 240:
```suggestion
```

Stypox (changes requested):
Code now looks good, the formatting just needs to be fixed

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -203,4 +209,43 @@ private static void openChannelFragment(final Fragment fragment,
...
     }
+
+    /////////////////////////////////////////////
+    // helper functions                        //
+    /////////////////////////////////////////////
+
+    private interface InfoCallback {
+        void onInfo(SinglePlayQueue item);
+    }
+
```
Stypox, line 220: Sorry for not noticing this before... instead of creating a new interface just for that one method you can use [a Java `Consumer<SinglePlayQueue>`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html). Just replace `InfoCallback` with `Consumer<SinglePlayQueue>` and replace `callback.onInfo` with `callback.accept`

Stypox (changes requested):

Stypox (approved):
Works perfectly, thank you!

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -203,4 +210,39 @@ private static void openChannelFragment(final Fragment fragment,
                 fragment.requireActivity().getSupportFragmentManager(),
                 item.getServiceId(), uploaderUrl, item.getUploaderName());
     }
+
+    /////////////////////////////////////////////
+    // helper functions                        //
+    /////////////////////////////////////////////
+
+    private static void fetchItemInfoIfSparse(final Fragment fragment,
```
TobiGr, line 218: Why is the fragment used as param here? The only usage for fragment is `fragment.getContext()` in this method.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -203,4 +210,39 @@ private static void openChannelFragment(final Fragment fragment,
...
+            final StreamInfoItem item,
+            final Consumer<SinglePlayQueue> callback) {
```
TobiGr, line 220: What is the advantage of using a consumer for a `SinglePlayQueue`? The method is more generic if the consumer accepts a `StreamInfoItem`.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -203,4 +210,39 @@ private static void openChannelFragment(final Fragment fragment,
...
+                                new HistoryRecordManager(fragment.getContext());
+                        recordManager.saveStreamState(result, 0)
+                                .subscribeOn(Schedulers.io())
+                                .observeOn(AndroidSchedulers.mainThread())
+                                .doOnError(throwable -> Log.e("StreamDialogEntry",
+                                        throwable.toString()))
+                                .subscribe();
+
+                        callback.accept(new SinglePlayQueue(result));
+                    }, throwable -> Log.e("StreamDialogEntry", throwable.toString()));
```
TobiGr, line 243: I think we should report these error via the UI. Otherwise the user gets no feedback on his action if something fails. However, I am not sure how to handle network related errors here. @Stypox Any idea?

TobiGr (commented):
While rebasing #7570 I ran into a few smaller things which I would like to get an answer / opinion on.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
Douile: This was because getStreamInfo returns a `StreamInfo` not a `StreamInfoItem`, forgive me if this exists but I didn't find a way to easily cast/convert between these types.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
Douile: Maybe we toast the user and fallback to returning the sparse info item then, although that still has its issues.

Stypox: You could use an error notification, just like https://github.com/TeamNewPipe/NewPipe/pull/7538/files#diff-212f79b4111281cdfbffed5f79bca8922f99be14642ef90d524e6c9de9a436f7R66

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
Stypox: Yes, this is ok

PR #7042, opened 2021-09-02
Make all buttons in player have selectable item background borderless

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

I replaced all `selectableItemBackground`s in the player with `selectableItemBackgroundBorderless`, so that the press animation looks better. I also did this to text buttons (fit-fill-zoom, captions, resolution, time), and I think it is better than non-borderless even though there are some small issues: the text changing when tapping fit-fill-zoom and when opening the resolution menu (MPEG-4 is added to the string) causes some artifacts in the animation, and the captions button, since it contains a long text, causes a big circle. Please provide your feedback :-)
I also ran `CTRL+ALT+L` to auto format and fixed some RTL warnings.

#### APK testing 

#### Due diligence


litetex:
Code looks good 👍 

However the background when tapping the button gets a bit too big on some buttons in my opinion:
grafik
Personally I was fine with the rectangular background-buttons 😄

opusforlife2:
Before/after screenshot, pls.

TiA4f8R:
I think we should keep the rectangular background for ratio, subtitles and quality selector commands.

Stypox:
> Before/after screenshot, pls.

I would have to upload something like twenty screenshots, one for each button, so I don't think it would be much useful in this case.

> I think we should keep the rectangular background for ratio, subtitles and quality selector commands.

I did this @TiA4f8R @litetex

litetex (approved):
LGTM

I'm personally okay with the changes 😄

opusforlife2:
> one for each button

Just one example would be enough, I think? It's only needed to get an idea of what the change looks like. I can't figure that out from your description.

Stypox:
<table>
  <tr>
    <th>Before</th>
<th>After</th>
  </tr>
  <tr>
    <td>  </td>
<td> </td>
  </tr>
</table>

opusforlife2:
Thanks, @Stypox. After seeing the screenshot, I agree with TiA4f8R. The icons certainly look better with a circular background, but text menus should still have a rectangular one.

> I think we should keep the rectangular background for ratio, subtitles and quality selector commands.

Stypox:
Ok, that's done (only the speed selector still has a circular background), thanks for the feedback ;-)

Stypox:
(just tested on API 19 emulated and it works)

PR #7043, opened 2021-09-02
Remove unused resources

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

This PR removes all Android resources that were unused. I used the automatic tool given by android studio (Analyze -> Run inspection by name... -> Unused resources) and then manually checked for things that should not have been removed, i.e. the following ones:
- the services themes, referenced in code [only through `getIdentifier()`](https://stackoverflow.com/a/47906018/9481500)
- `xml/provider_paths.xml`, I didn't know what it was, so better not to touch it
- the `.png` files for backwards compatibility with API 19 notifications
- a couple of `@+id/` on views that were removed by Android Studio but were actually still used

Other small changes:
- Remove uselessly nested layout in local_playlist_header.xml 
- Remove `tools:text` on a view where `android:text` was already defined in feed_import_export_group.xml 
- Add `tools:keep` to the resources that shouldn't be removed (service themes)

#### APK testing 

#### Due diligence


mhmdanas:
I'm unable to build an APK right now, but the debug APK produced by the CI for this PR is 0.3 MBs smaller than the one produced by latest `dev`. If you don't mind, can you compare between the release APK size with and without this PR? (This sounds like it should definitely decrease the APK size, but I'm curious how _much_ it decreases it :) )

litetex (approved):
LGTM

Works like a charm

mhmdanas (approved):
LGTM

I built the release APKs and compared them: the new APK is 8948885 bytes, while the old one is 8949784, a difference of around 1 KB. Not that big, but that's not really the point of this PR anyway, and it shouldn't be. The cleanup is more than worth it.

PR #7098, opened 2021-09-12
Allow passing in package suffix for release builds through CLI param

### LINKED ISSUE #6420: Allow passing in package name suffix through CLI
### Checklist

- Allow specifying an optional package name suffix when calling `./gradlew assembleRelease --stacktrace`
- Similiar to how debug build have different package names

- My nightly builds have the same package name, so you can't install both versions
- https://github.com/XiangRongLin/NewPipe-nightly/issues/1

- Installation of both versions is possible

XiangRongLin:
#### What is it?

- Allow passing in package suffix for release builds through CLI param
- Add documentation for the custom gradle parameters

- Fixes #6420

#### Testing Description
- Run `gradle clean assembleRelease -DpackageSuffix=nightly`
- Sign it with `apksigner` https://developer.android.com/studio/command-line/apksigner
- Install the apk and see that it has the `nightly` suffix

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -54,6 +54,11 @@ android {
         // debug build. This seems to be a Gradle bug, therefore
         // TODO: update Gradle version
         release {
+            if (System.properties.containsKey('packageSuffix')) {
+                applicationIdSuffix System.getProperty('packageSuffix')
+                resValue "string", "app_name", "NewPipe " + System.getProperty('packageSuffix')
+                archivesBaseName = 'NewPipe_' + System.getProperty('packageSuffix')
+            }
             minifyEnabled true
             shrinkResources false // disabled to fix F-Droid's reproducible build
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
--- a/doc/gradle.md
+++ b/doc/gradle.md
@@ -1,0 +1,14 @@
+# Custom gradle parameters
+
+You can use these parameters by specifying them inside the `gradle.properties` file as 
+`systemProp.<name>=<value>` or passing them through the CLI with `-D<name>=<value>`.
+
+## packageSuffix
+This allows you to specify a suffix, which will be added on release builds to the application id, 
+the `app_name` string and the apk file.  
+No validation is made, so make sure to pass in a valid value.
+
+## skipFormatKtlint
+This allows you to skip the `formatKtLint` task.  
+No value is needed.   
+It is used for CI in order to check for badly formatted files.


app/build.gradle
```diff
@@ -54,6 +54,11 @@ android {
         // debug build. This seems to be a Gradle bug, therefore
         // TODO: update Gradle version
         release {
+            if (System.properties.containsKey('packageSuffix')) {
+                applicationIdSuffix System.getProperty('packageSuffix')
```
Stypox, line 58: I would pass the `packageSuffix` parameter through the same sanitization as the git branch name

doc/gradle.md
```diff
@@ -0,0 +1,14 @@
+# Custom gradle parameters
+
+You can use these parameters by specifying them inside the `gradle.properties` file as 
+`systemProp.<name>=<value>` or passing them through the CLI with `-D<name>=<value>`.
+
+## packageSuffix
+This allows you to specify a suffix, which will be added on release builds to the application id, 
+the `app_name` string and the apk file.  
+No validation is made, so make sure to pass in a valid value.
```
Stypox, line 9: I would add that by default the git branch name is used, but this can be used to override it

Stypox (changes requested):

app/build.gradle
XiangRongLin: I would say that this is not needed.  
In the git branch case the name of something unrelated (branch name) is added to the package id, app_name and archive name. And because the branch name allows for a wider range of values, it needs to be sanitized.

In this case the provided parameter is exclusivly used for those 3 things. Additionally I would rather like to know that my `packageSuffix` is not valid than get an app with the wrong name

doc/gradle.md
XiangRongLin: This option only overrides the name of release builds, not debug build

> This allows you to specify a suffix, which will be added on >>release<< builds to the application id

Stypox:
Sorry, I think I missed the point of this PR at first. You need to allow specifying the suffix for release builds so that you can provide release testing APKs in PRs without conflicting with the default NewPipe, don't you?

doc/gradle.md
Stypox: Ok, then this looks good, sorry for the misunderstanding

app/build.gradle
Stypox: Yes, it's not needed then

XiangRongLin:
My usecase is for nightly builds that im doing over at https://github.com/XiangRongLin/NewPipe-nightly/issues/1 
It's fully automated and pulls changes from here and pushed then to that repo. Because of that I can't manually make a commit to change the package id, because otherwise when pushing new commits from here it may run into merge-conflicts.

For PR I don't need it, since I can just manaully change id, since the branch will be deleted afterwards.

Stypox (approved):
Ok, this looks good then, thanks :-)

Stypox:
I was going to merge it but then didn't find any draft release, have you added it somewhere?

XiangRongLin:
yes added it. Funny when you go to releases, it doesn't show the releases. After you refresh the page, it show them 🤔

PR #7132, opened 2021-09-21
Simple Codebase Improvement  

Jhooomn:
feat: it was added a string variable to avoid replication one.

### DIFF
--- a/app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
+++ b/app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
@@ -43,10 +43,11 @@ class LocalPlaylistManagerTest {
 
     @Test
     fun createPlaylist() {
+        val NEW_PIPE_LINK = "https://newpipe.net/"
         val stream = StreamEntity(
-            serviceId = 1, url = "https://newpipe.net/", title = "title",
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
```diff
@@ -43,10 +43,11 @@ class LocalPlaylistManagerTest {
 
     @Test
     fun createPlaylist() {
+        val NEW_PIPE_LINK = "https://newpipe.net/"
```
Stypox, line 46: Call it `NEWPIPE_URL`

Stypox (changes requested):
Thank you!

app/src/androidTest/java/org/schabi/newpipe/local/playlist/LocalPlaylistManagerTest.kt
Jhooomn: It was updated with recommendations.

mhmdanas (dismissed):
Looks good to me, but a couple things:
- please use more descriptive commit messages. I know that the message you used is the one provided by GitHub by default, but you can change it.
- please don't remove the template that was there next time you open a PR (or issue for that matter).

Thanks!

Stypox (approved):

mhmdanas (approved):

PR #7149, opened 2021-09-22
Update pager workaround code to Fragment 1.3.6

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- I ported over changes from the newer versions of AndroidX Fragment's `FragmentStatePagerAdapter` to the `FragmentStatePagerAdapterMenuWorkaround`, since the current one is from a much older version.

Of course, this will eventually be removed anyway for ViewPager2, but I figured I'd do this anyway until ViewPager2 can actually be implemented.

#### APK testing 

#### Due diligence


app/src/main/java/androidx/fragment/app/FragmentStatePagerAdapterMenuWorkaround.java
```diff
@@ -127,8 +132,7 @@ public FragmentStatePagerAdapterMenuWorkaround(@NonNull final FragmentManager fm
     }
 
     /**
-     * @param position the position of the item you want
-     * @return the {@link Fragment} associated with a specified position
+     * Return the Fragment associated with a specified position.
```
Stypox, line 135: Why this change? It is causing the build to fail

Stypox (changes requested):
Thank you!

app/src/main/java/androidx/fragment/app/FragmentStatePagerAdapterMenuWorkaround.java
TacoTheDank: Simply figured that I'd move over to the newer changes as much as possible, though I guess the documentation doesn't really matter that much. If it's causing checkstyle errors, I'll just change it back.

Issue #7087, opened 2021-09-10
Potential performance problem with MediaSessionManager#get/setMetadata


litetex:
### Checklist
- [x] I am using the latest version - 0.21.9 / develop
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo.
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug
Run the app / play a video (in the background).

What I did:
Opened a video and ran it in the background.

### Details
NewPipe performs a progress update when the player is active every 500ms/1s (was increased to 1s in #7071).

However when this progress update is executed the ``setMetadata``-method seems to be very costly (takes 40-80% of the time of the complete progress update) as seen when profiling the app:

<details><summary>Emulator Pixel 3a</summary>

[cpu-emulator-Pixel3a.zip](https://github.com/TeamNewPipe/NewPipe/files/7146749/cpu-emulator-Pixel3a.zip)

<img src="https://user-images.githubusercontent.com/40789489/132923638-8e16eea3-44e2-4077-84a4-5c1683f2c9f4.png">

</details>

<details><summary>Emulator Pixel C</summary>

[cpu-emulator-PixelC.zip](https://github.com/TeamNewPipe/NewPipe/files/7146750/cpu-emulator-PixelC.zip)

<img src="https://user-images.githubusercontent.com/40789489/132923471-e51a6810-2cdb-4fd9-a410-369fde396eec.png">
<img src="https://user-images.githubusercontent.com/40789489/132923587-1d8ad1ee-b3cc-4db6-9860-2ffd0ba086bc.png">

</details>

Note: I'm unable to determine if this is just a android emulator problem as I can't profile my real phone (crashes with SIGSEV in HeapTaskDaemon/libc when running the profiler)

### Device info
 - Android version/Custom ROM version: Android 11 - API 30
 - Device model: Android Emulator Pixel 3a/Pixel C

Redirion:
while looking at setMetaData I noticed another issue:

`// setMetadata only updates the metadata when any of the metadata keys are null
        mediaSessionManager.setMetadata(getVideoTitle(), getUploaderName(),
                showThumbnail ? getThumbnail() : null, duration);`

in other words, if showThumbnail is false, metaData updates do never happen!

[...]

PR #7161, opened 2021-09-24
Fix two issues with setMetadata

Redirion:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)

Issue 1 is when the option "Show thumbnail" is turned off. It caused setMetadata to never be called.
`// setMetadata only updates the metadata when any of the metadata keys are null mediaSessionManager.setMetadata(getVideoTitle(), getUploaderName(), showThumbnail ? getThumbnail() : null, duration);`

basically the first line in setMetadata checks, if the provided thumbnail/albumArt is not null.

Issue 2: if no thumbnail could be loaded, the dummy thumbnail resource would be used to be created through a Bitmap factory. Firstly it will be created each time (expensive) and secondly hashCode() will always yield different results. This means that setMetadata was basically executed in its whole every single time (each second since last updates).

Don't know if there is an issue that would be closed by this. I found it while looking at issue #7087

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -1606,8 +1606,12 @@ public final class Player implements
         final boolean showThumbnail = prefs.getBoolean(
                 context.getString(R.string.show_thumbnail_key), true);
         // setMetadata only updates the metadata when any of the metadata keys are null
-        mediaSessionManager.setMetadata(getVideoTitle(), getUploaderName(),
-                showThumbnail ? getThumbnail() : null, duration);
+        if (showThumbnail) {
+            mediaSessionManager.setMetadata(getVideoTitle(), getUploaderName(), getThumbnail(),
+                    duration);
+        } else {
+            mediaSessionManager.setMetadata(getVideoTitle(), getUploaderName(), duration);
+        }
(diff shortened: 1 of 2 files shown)


litetex:
@Redirion 
Thank you for the PullRequest 😄 

I quickly read the code (not in depth) and large parts of the code seem to be duplicated. 
Is it possible to de-duplicate those?

PS: Will test it out later with the Android profiler 😄

Redirion:
apart from the debug logs nothing is really duplicated. It just looks similar ;)

the albumArt is completely dropped in the new method and the check for updating does not focus around the hashCode of the bitmap  but uses equals on the title instead

However this is not a fix for the general performance of setMetadata. It fixes two specific issues I found when checking the method. I did not find a way to increase the performance. hashCode should already be the most cost effective way to compare bitmaps.
Other than that we could only drop the comparison of bitmaps altogether and just use equals for title, artist and compare the duration.

litetex:
@Redirion 
I create a update version of this pull request with additional fixes: #7166 

You might close this PR (Update: It got closed 5 minutes ago 😆)
Also note that your commit-mail seems to be wrong (no profile picture showing up).
Anyway thank you for the PR 😄

PR #7166, opened 2021-09-24
Various fixes related to MediasessionManager

### LINKED ISSUE #7087: Potential performance problem with MediaSessionManager#get/setMetadata

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

The following things are fixed
 
from #7161 (credits to @Redirion)

> Issue 1 is when the option "Show thumbnail" is turned off. It caused setMetadata to never be called.
`// setMetadata only updates the metadata when any of the metadata keys are null mediaSessionManager.setMetadata(getVideoTitle(), getUploaderName(), showThumbnail ? getThumbnail() : null, duration);`
>
> basically the first line in setMetadata checks, if the provided thumbnail/albumArt is not null.
>
> Issue 2: if no thumbnail could be loaded, the dummy thumbnail resource would be used to be created through a Bitmap factory. Firstly it will be created each time (expensive) and secondly hashCode() will always yield different results. This means that setMetadata was basically executed in its whole every single time (each second since last updates).

I also reworked/de-duplicated the code a bit 😉

Issue 3: #7087 
https://github.com/TeamNewPipe/NewPipe/issues/7087#issuecomment-926934423
> Okay I tested the stuff a bit and found out that the metadata updates - that are done every second - are completely useless.
> They should only be done when the metadata changes e.g. when opening a new video.

Issue 4:
Comparisons e.g. if the current title/artist/duration is also the requested one have been missing. 
Added those by caching the last requested data using hashcodes.
This also avoids resource-intensive ``mediaSession.getController().getMetadata()`` calls.

- Fixes #7087 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -140,6 +140,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
 import io.reactivex.rxjava3.core.Observable;
@@ -1622,16 +1623,6 @@ public final class Player implements
         if (isQueueVisible) {
             updateQueueTime(currentProgress);
         }
-
-        final boolean showThumbnail = prefs.getBoolean(
-                context.getString(R.string.show_thumbnail_key), true);
-        // setMetadata only updates the metadata when any of the metadata keys are null
-        if (showThumbnail) {
-            mediaSessionManager.setMetadata(getVideoTitle(), getUploaderName(), getThumbnail(),
-                    duration);
-        } else {
-            mediaSessionManager.setMetadata(getVideoTitle(), getUploaderName(), duration);
-        }
     }
 
     private void startProgressLoop() {
@@ -2950,6 +2941,16 @@ public final class Player implements
                 tag.getMetadata().getPreviewFrames());
 
         NotificationUtil.getInstance().createNotificationIfNeededAndUpdate(this, false);
+
+        final boolean showThumbnail = prefs.getBoolean(
+                context.getString(R.string.show_thumbnail_key), true);
+        mediaSessionManager.setMetadata(
+                getVideoTitle(),
+                getUploaderName(),
+                showThumbnail ? Optional.ofNullable(getThumbnail()) : Optional.empty(),
+                tag.getMetadata().getDuration()
+        );
+
         notifyMetadataUpdateToListeners();
 
         if (areSegmentsVisible) {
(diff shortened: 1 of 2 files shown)


Redirion:
you could add the result of isLiveEdge to decide whether to check the duration

litetex:
@Redirion
> you could add the result of isLiveEdge to decide whether to check the duration

I can't follow. Where do you mean should I add this?

Redirion:
sorry isLiveEdge was wrong. But you can use isLive(): https://github.com/TeamNewPipe/NewPipe/blob/ac371e6fb4f5cb79387e0650f2d0ad69bf8abfb0/app/src/main/java/org/schabi/newpipe/player/Player.java#L4158

you could use it as an additional parameter:
```
mediaSessionManager.setMetadata(
                getVideoTitle(),
                getUploaderName(),
                showThumbnail ? Optional.ofNullable(getThumbnail()) : Optional.empty(),
                tag.getMetadata().getDuration(),
                isLive()
        );
```

this way you can use duration for comparisons if it isn't a livestream

litetex:
@Redirion 
> this way you can use duration for comparisons if it isn't a livestream

Duration comparison is already always done: 
https://github.com/TeamNewPipe/NewPipe/blob/bbc13756f3653a5624b47f55f4654320ebee5a57/app/src/main/java/org/schabi/newpipe/player/helper/MediaSessionManager.java#L143

So I don't understand the problem in the first place...

Anyway the check for ``isLive`` is done by asking the player which is risky as it may not be fully initalized when ``onMetadataChanged`` is called. It's also quite resource expensive when compared to simply reading if it's a livestream from the metadata.

I change the code so that it now considers livestream and always uses duration -1 for them 😄

Redirion:
Yep sorry, all good

Redirion (approved):
lgtm

Two personal preferences as a comment:
- I would always prefer equals over hashCode for String comparisons
- I would rather have a bit duplicated code by using more than one method than having a method with many return statements

litetex:
> * I would always prefer equals over hashCode for String comparisons

I also considered that. However the deal with hashCode is that storing an int requires usually way less RAM than a String by - what I think are accaptable odds - of having an collision: Which is technically 1 in 2^32.
Someone tried it with 58k words and got one collision: https://stackoverflow.com/a/9407032

PR #7181, opened 2021-09-28
Add YouTube shorts path prefix to intent filter

### LINKED ISSUE #7130: Intent filters for YT shorts
### Checklist

Parsing URLs for YT shorts works fine, as per teamNewPipe/NewPipeExtractor#418. However NP is not registered in the system to be able to handle shorts URLs. Can this be added to intent filters in AndroidManifest.xml? If I tap on a link in another app and it's a "normal" YT link it happily opens in NP but if it's a "shorts" link it will open in the browser. I'd like it if shorts opened directly in NP as well. 🙂

mhmdanas:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)

Modify the intent filter in `AndroidManifest.xml` to handle URLs for YouTube shorts.

- Fixes #7130

#### APK testing 

#### Due diligence

@ravilov @shambu2k does this PR work for you?

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -146,7 +146,7 @@
                 <data android:pathPrefix="/embed/" />
                 <data android:pathPrefix="/watch" />
                 <data android:pathPrefix="/attribution_link" />
-                <data android:pathPrefix="/shorts" />
+                <data android:pathPrefix="/shorts/" />
                 <!-- channel prefix -->
                 <data android:pathPrefix="/channel/" />
                 <data android:pathPrefix="/user/" />
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -146,6 +146,7 @@
                 <data android:pathPrefix="/embed/" />
                 <data android:pathPrefix="/watch" />
                 <data android:pathPrefix="/attribution_link" />
+                <data android:pathPrefix="/shorts" />
                 <!-- channel prefix -->
                 <data android:pathPrefix="/channel/" />
                 <data android:pathPrefix="/user/" />


app/src/main/AndroidManifest.xml
```diff
@@ -146,6 +146,7 @@
                 <data android:pathPrefix="/embed/" />
                 <data android:pathPrefix="/watch" />
                 <data android:pathPrefix="/attribution_link" />
+                <data android:pathPrefix="/shorts" />
```
mhmdanas, line 149: Some items here have a trailing slash and some don't. Is there a meaningful difference between the two?

ravilov:
@mhmdanas Just tested it, seems to work as expected, thanks 👍

app/src/main/AndroidManifest.xml
litetex: Okay so the links seem to usually look like this ``https://youtube.com/shorts/xxxxxxxxxxx``.

It think there should be a ``/`` at the end as the actual video is identified by an url parameter.
The ``/watch`` endpoint e.g. uses query parameters ``?v=xyz`` that's why there isn't a ``/`` at the end.

```suggestion
                <data android:pathPrefix="/shorts/" />
```


<i><sub>This comment cost me at least 3 brain-cells whilst watching some "shorts". Is there some kind of warranty so I can get them back?</sub></i>

mhmdanas: Aha, that makes sense. A comment explaining this would be helpful; I'll open a separate PR for this.

> This comment cost me at least 3 brain-cells whilst watching some "shorts". Is there some kind of warranty so I can get them back?

<sub>Be thankful this is not TikTok we're talking about :P</sub>

Redirion (approved):
lgtm

PR #5850, opened 2021-03-17
Add enqueue background choice for share service.

chouhanyang:
Add a new menu item in share menu to enqueue into background player.

#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
- Add string resources for the new menu item.
- Add logic in menu/dialog creation.
- Add handler to enqueue stream into background player.
- Screenshot:
device-2021-03-11-171859

#### Fixes the following issue(s)
- https://github.com/TeamNewPipe/NewPipe/issues/4779

#### APK testing 
https://github.com/SavantOne/NewPipe/releases/download/v0.20.11-patch/NewPipe-v0.20.11-patch.apk

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #7194, opened 2021-10-02
Add a "add to playlist" option in the share menu

### LINKED ISSUE #5858: Add "Add to playlist" option in "Always ask" share menu
### Checklist

It's really sad that Newpipe doesn't have the ability to access "Watch Later" playlist for shared videos. Having an ability to "Add to playlist" will make the experience much better.

Watching later for any video will be possible. Plus, binge watching similar videos will be a breeze.

Sometimes, we need to browse several websites only to find different videos on the same topic. Definitely, having several queues on is not a good idea as the notification panel will get bombarded with too many media player notification bars which is not a good idea (suggested in #4460). Instead, after adding to playlist, we can watch similar videos later. This will also avoid accidental removal of videos from enqueued list. Having too many enqueued list is even more dangerous.

If a user still wants to watch popup right after background and then they can add all videos and use "Always-on button" mentioned in #5821. It will be a 2 in 1, users will be able to switch to background or popup whenever within a single tap.

KalleStruik:
[!] Before this can be merged, there is a small theming issue that needs to be fixed. When creating a new playlist from the share dialog it uses the wrong color for the buttons (See screenshot 4). I have been unable to find the cause of this bug, so any help in fixing it is appreciated.

#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Add an "Add to playlist" option in the incoming share dialog.

- Before:
image
image

- After:
image
When no playlists exist:
image
When a playlist exists:
image
image

- Fixes #5858

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -433,6 +433,7 @@
     <string name="background_player">Background player</string>
     <string name="popup_player">Popup player</string>
     <string name="always_ask_open_action">Always ask</string>
+    <string name="add_to_playlist">Add to playlist</string>
     <string name="preferred_player_fetcher_notification_title">Getting info…</string>
     <string name="preferred_player_fetcher_notification_message">"Loading requested content"</string>
     <!-- Local Playlist -->
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -1,5 +1,8 @@
 package org.schabi.newpipe;
 
+import static org.schabi.newpipe.extractor.StreamingService.ServiceInfo.MediaCapability.AUDIO;
+import static org.schabi.newpipe.extractor.StreamingService.ServiceInfo.MediaCapability.VIDEO;
+
 import android.annotation.SuppressLint;
 import android.app.IntentService;
 import android.content.Context;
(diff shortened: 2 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -433,6 +433,7 @@
     <string name="background_player">Background player</string>
     <string name="popup_player">Popup player</string>
     <string name="always_ask_open_action">Always ask</string>
+    <string name="add_to_playlist">Add to playlist</string>
```
TobiGr, line 436: We already have that string, but its name is wrong. (I think the string was changed, but its name wasn't).
Please change `append_playlist` -> `add_to_playlist`

KalleStruik:
The text of the dialog buttons is now visable
image

litetex:
> image

In my opinion "always ask" should be at the end and not in between.

KalleStruik:
Completely agree
image

litetex:
@KalleStruik 
Seems to work very good as far as I can see.

However I applied some fixes regarding the underlying code. 
There are many issues who somehow got into the code and now affect this PR.

I also added a little hint/toast-notification that adding to the playlist may take a moment:
grafik

ThisIsPaulDaily:
Hey, I understand, and appreciate this PR. 

I really would like this one to be pulled first. 
https://github.com/TeamNewPipe/NewPipe/pull/5850
#5850  is restoring an original feature that was broken several versions ago.

ThisIsPaulDaily:
You highlighted four things I said and I can explain. 
This PR and 5850 touch similar sections of code and will conflict. 

Binge watching similar videos is a breeze is what the old enqueue functionality allowed. 


The last part was a gripe to fix what was broken before trying to add new features. 5850 is the consolidation of several tickets which all had the same root cause of removing the queue. 

Is there something inherently wrong with #5850 that prevents it from being reviewed and pushes this PR ahead?

PR #7195, opened 2021-10-03
Prevent automatic replay after returning from background

sryze:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)

- If the video has completed, when the app returns from background, prevent the player from staring playback automatically by setting setPlayWhenReady to false

I'm not 100% sure if this is the good way to fix it. If not, please let me know how I can improve it.

Steps to reproduce the bug:
1. Play a video
2. Lock the device while the video is still playing
3. Wait until the video ends
4. Unlock the device
5. The video starts playing again automatically

This is kind of annoying when you listen to some podcast in audio-only mode, and then a few hours later you want to do something else on the phone, so your unlock it, and all of a sudden NewPipe starts playing the video again.

- Before:

https://user-images.githubusercontent.com/4525736/140655574-733a3749-ea7c-4979-b993-3847ed2a20dc.mp4

- After:

https://user-images.githubusercontent.com/4525736/140655773-6869c510-6657-428e-a97e-d9f86ff7d139.mp4

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -2031,6 +2031,9 @@ public final class Player implements
         currentItem = null;
         currentMetadata = null;
         simpleExoPlayer.stop();
+        if (currentState == STATE_COMPLETED) {
+            simpleExoPlayer.setPlayWhenReady(false);
+        }
         isPrepared = false;
 
         changeState(STATE_BLOCKED);


avently:
Nice catch, thank you. Didn't test it but if it works it's good

sryze:
I found that found that when I close NewPipe it does not restart the player in this build, i.e. no delay in playback like in the dev branch. I'll need to look into this. Please don't merge this PR!

opusforlife2:
> However this one doesn't fix the cause of the problem and it's just a workaround that may break other things.

Marking it as draft for now since it's being investigated further.

avently:
@sryze you probably needs to add your check here:
https://github.com/TeamNewPipe/NewPipe/blob/1593723879bb57be220f7a82107970786d07738f/app/src/main/java/org/schabi/newpipe/player/Player.java#L1167

I'm not tested but it should be a right place for your check on finished playing state since this piece of code called when the app returns from background

litetex:
@avently 
→ https://github.com/TeamNewPipe/NewPipe/pull/7195#issuecomment-949857436

sryze:
@avently This approach does not work, because reloadPlayQueueManager must be called to switch from audio to video mode. If we don't call it, the video will not show any picture when replayed.

```diff
diff --git a/app/src/main/java/org/schabi/newpipe/player/Player.java b/app/src/main/java/org/schabi/newpipe/player/Player.java
index 22e66e793..abef1d40a 100644
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -1164,7 +1164,9 @@ public final class Player implements
                 break;
             case VideoDetailFragment.ACTION_VIDEO_FRAGMENT_RESUMED:
                 fragmentIsVisible = true;
-                useVideoSource(true);
+                if (currentState != STATE_COMPLETED) {
+                    useVideoSource(true);
+                }
                 break;
             case VideoDetailFragment.ACTION_VIDEO_FRAGMENT_STOPPED:
                 fragmentIsVisible = false;
```

litetex:
Okay so I dug around a bit more and I think I finally found the root-problem.

The playqueue-manager works as expected. A reload has to be done to get the correct streaming data (e.g. with video + audio)
If I apply my above changes I also end up with no video.

To the problem: in ``useVideoSoure`` ``setRecovery`` is called:
https://github.com/TeamNewPipe/NewPipe/blob/b5ad24eb475c1d0280fc8094438643acc77bbb74/app/src/main/java/org/schabi/newpipe/player/Player.java#L4145

This works fine when you lock your phone and the video is still not completed. However it seems to fail if the playback has ended.
The current position of Exoplayer is a few ms after the duration of the video:
(Added the variable ``duration`` below for debugging purposes)
grafik

This causes the recovery-position not be set and through that the player starts to replay the video from the beginning.

litetex:
Code to fix the above problem:
```diff
index 22e66e793..1a76e5022 100644
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -858,9 +858,15 @@ public final class Player implements

         final int queuePos = playQueue.getIndex();
         final long windowPos = simpleExoPlayer.getCurrentPosition();
+        final long duration = simpleExoPlayer.getDuration();

-        if (windowPos > 0 && windowPos <= simpleExoPlayer.getDuration()) {
-            setRecovery(queuePos, windowPos);
+        if (windowPos > 0
+                // Sometimes (e.g. when the playback ended) the windowPos is a few milliseconds
+                // higher than the duration. Due to this a little buffer (100ms) was introduced.
+                // See also https://github.com/TeamNewPipe/NewPipe/pull/7195#issuecomment-962624380
+                && windowPos <= duration + 100
+        ) {
+            setRecovery(queuePos, Math.min(windowPos, duration));
         }
     }
```

sryze:
@litetex your fix is working great in regards to the replaying issue, but there is one little problem that I've noticed - when you return to NewPipe and hit the Replay button, it keeps showing the play/pause overlay, e.g. please see this video:

https://user-images.githubusercontent.com/4525736/140655671-53f10945-2987-4843-a71a-95b09d5e1216.mp4

avently:
@litetex nice finding!

@sryze same thing happens without his fix, so it's unrelated to the code

sryze:
OK, closing this PR then

litetex:
Succeeded by #7381

PR #7244, opened 2021-10-14
Show comment replies

### LINKED ISSUE #2277: Display comment replies
It's great that NewPipe has the feature to view comments since the version 0.16.0! But I miss a function to view the answers of the comments.

golfinq:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Adds a "Show Replies" for the comments with replies for Youtube; when clicked, the first page of replies is displayed below the comment.

(Ignore the different themes)
- Before: 

- After: 

Before Clicking:

After Clicking:

- Fixes #2277

- https://github.com/TeamNewPipe/NewPipeExtractor/pull/739

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsInfoItemHolder.java
@@ -5,11 +5,18 @@ import android.view.ViewGroup;
 import android.widget.ImageView;
 import android.widget.TextView;
 
+import androidx.recyclerview.widget.LinearLayoutManager;
+import androidx.recyclerview.widget.RecyclerView;
+
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.comments.CommentsInfoItem;
 import org.schabi.newpipe.info_list.InfoItemBuilder;
+import org.schabi.newpipe.info_list.InfoListAdapter;
 import org.schabi.newpipe.local.history.HistoryRecordManager;
+
+import java.util.List;
+import java.util.Objects;
 
 /*
  * Created by Christian Schabesberger on 12.02.17.
@@ -35,11 +42,34 @@ public class CommentsInfoItemHolder extends CommentsMiniInfoItemHolder {
     public final TextView itemTitleView;
     private final ImageView itemHeartView;
 
+    private final TextView showReplies;
+    private final RecyclerView repliesView;
+
     public CommentsInfoItemHolder(final InfoItemBuilder infoItemBuilder, final ViewGroup parent) {
         super(infoItemBuilder, R.layout.list_comments_item, parent);
 
         itemTitleView = itemView.findViewById(R.id.itemTitleView);
         itemHeartView = itemView.findViewById(R.id.detail_heart_image_view);
+
+        showReplies = itemView.findViewById(R.id.showReplies);
+        repliesView = itemView.findViewById(R.id.replyRecycleView);
+        repliesView.setAdapter(new InfoListAdapter(repliesView.getContext()));
+        repliesView.setLayoutManager(new LinearLayoutManager(repliesView.getContext()));
+    }
+
+    public void addReplies(final View buttonView, final CommentsInfoItem comment) {
+
+        final List<CommentsInfoItem> replies = comment.getRepliesInfoList();
+        ((InfoListAdapter) Objects.requireNonNull(repliesView.getAdapter()))
+                .setInfoItemList(replies);
+
+        final ViewGroup.MarginLayoutParams params =
+                (ViewGroup.MarginLayoutParams) repliesView.getLayoutParams();
+        params.topMargin = 45;
+
+        repliesView.setMinimumHeight(100);
+        repliesView.setHasFixedSize(true);
+        buttonView.setVisibility(View.GONE);
     }
 
     @Override
@@ -55,5 +85,13 @@ public class CommentsInfoItemHolder extends CommentsMiniInfoItemHolder {
         itemTitleView.setText(item.getUploaderName());
 
         itemHeartView.setVisibility(item.isHeartedByUploader() ? View.VISIBLE : View.GONE);
+
+        if (item.getReplies() == null) {
+            showReplies.setVisibility(View.GONE);
+        } else {
+            showReplies.setVisibility(View.VISIBLE);
+            showReplies.setOnClickListener(v -> addReplies(v, item));
+        }
+
     }
 }
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsInfoItemHolder.java
@@ -5,18 +5,13 @@ import android.view.ViewGroup;
 import android.widget.ImageView;
 import android.widget.TextView;
 
-import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.comments.CommentsInfoItem;
 import org.schabi.newpipe.info_list.InfoItemBuilder;
-import org.schabi.newpipe.info_list.InfoListAdapter;
 import org.schabi.newpipe.local.history.HistoryRecordManager;
-
-import java.util.List;
-import java.util.Objects;
 
 /*
  * Created by Christian Schabesberger on 12.02.17.
@@ -41,9 +36,7 @@ import java.util.Objects;
 public class CommentsInfoItemHolder extends CommentsMiniInfoItemHolder {
     public final TextView itemTitleView;
     private final ImageView itemHeartView;
-
-    private final TextView showReplies;
-    private final RecyclerView repliesView;
+    private final RepliesHandler repliesHandler;
 
     public CommentsInfoItemHolder(final InfoItemBuilder infoItemBuilder, final ViewGroup parent) {
         super(infoItemBuilder, R.layout.list_comments_item, parent);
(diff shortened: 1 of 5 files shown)


BratishkaErik:
Can't wait to see this in app!

opusforlife2:
@BratishkaErik if you have nothing to add to the discussion, please just react with a thumbs up to the PR.

golfinq:
Is there anything that reviewers/maintainers want from me/this pull request in order to get merged? I don't have a free time until next weekend to work on anything, but if there is anything to help this PR, it would be appreciated to know.

opusforlife2:
@golfinq We just need a team dev to get enough free time to review this PR.

app/src/main/res/layout/list_comments_item.xml
```diff
@@ -123,4 +123,34 @@
...
+        android:layout_marginStart="21dp"
+        android:layout_marginTop="0sp"
+        android:layout_marginEnd="0sp"
+        android:layout_marginBottom="0sp"
+        android:layout_toEndOf="@+id/itemPublishedTime"
+        android:text="@string/show_replies"
+        android:textAppearance="@style/TextAppearance.AppCompat.Small"
+        android:textSize="@dimen/video_item_search_upload_date_text_size" />
+
+    <androidx.recyclerview.widget.RecyclerView
```
Stypox, line 142: If I understood correctly https://github.com/TeamNewPipe/NewPipeExtractor/pull/703, a reply-comment could have replies itself, but this PR introduces support only for 1 level of reply depth. This would work fine for YouTube, but it would be problematic for other services, so the `replyRecyclerView` should be added dynamically. Doing this might not be best practice, since the layouts recycled in recycler views should not be built dynamically, but since the amount of comments we expect users to open is much smaller than the total number of comments, I don't think creating a few more views every time an expanded comment is shown would hurt performance whatsoever. Note: if you do this remember to cleanup the recycler view previously created, if there is one, before creating a new one.

app/src/main/java/org/schabi/newpipe/info_list/holder/RepliesHandler.java
```diff
@@ -0,0 +1,192 @@
...
+        } else {
+            addRepliesToUI(parentInfoItem);
+        }
+    }
+
+    public void checkForReplies(final CommentsInfoItem item) {
+        if (item.getReplies() == null) {
+            repliesView.setVisibility(View.GONE);
+            showReplies.setVisibility(View.GONE);
+        } else if (item.getRepliesOpen()) {
```
Stypox, line 184: As I said in the extractor, this method should be removed from there and implemented here in the frontend. The usual approach would be to create an item to use in the `RecyclerView` that wraps around `CommentsInfoItem` but also has a field `isExpanded`. This option is not viable, though, since in NewPipe lists use a common pattern to load, reload, etc., so changing the type of data fed to the recycler view without changing the type of data returned by the extractor is not doable without huge code changes (and we don't want those). A solution that would also be suitable for the multiple-depth replies implementation (see my other comment), would be to just keep a `static Set<CommentsInfoItem>` somewhere storing a common list of expanded `CommentsInfoItem`s (common meaning that it will be accessed by *all* (even nested) `RecyclerView`s' `CommentsInfoItem.updateFromItem()`) . When the user expands a comment view the related `CommentsInfoItem` should be added to the `Set`, when the user unexpands the same comment it should be removed and when the `RecyclerView` tries to load a comment using `CommentsInfoItemHolder.updateFromItem(item)` it is expanded based on whether `item` belongs to the set.

app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsInfoItemHolder.java
```diff
@@ -5,6 +5,8 @@
 import android.widget.ImageView;
```
Stypox, line 5: Is there a particular reason you added this code in `CommentsInfoItemHolder` and not `CommentsMiniInfoItemHolder` like all other comments code?

Stypox (changes requested):
As I explain in the comments the structure needs some work, but it's nothing too difficult after all. Thank you for the PR :-)

app/src/main/java/org/schabi/newpipe/info_list/holder/RepliesHandler.java
Stypox: @litetex do you have any idea whether this can be achieved in a better way, without storing any separate data in `RecyclerView`s (since that data would be lost when the comment view is recycled)?

litetex: @Stypox At first: Welcome back.
Whoa that's a lot of info. 

> do you have any idea whethe ...

No I don't - currently.

I think I would have to look at this in more detail to get an answer to that, however some ideas from my side to make nested commets working:
* ViewModel which holds the data?
* Write a custom Custom component for this?

Stypox: @golfinq ping ;-)

golfinq:
Thanks for the suggestions! I ended up getting swamped before the thanksgiving holiday, but it should now be alright. I'll be taking a look at this later today

app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsInfoItemHolder.java
golfinq: It was probably left over from something else I was trying to do, should be straight forward to move it to where it is used

app/src/main/java/org/schabi/newpipe/info_list/holder/RepliesHandler.java
golfinq: @Stypox @litetex 
>     * ViewModel which holds the data?
>     * Write a custom Custom component for this?

I am more inclined to use the first option as I am trying to keep the amount of new code minimal for future maintenance reasons. I was hoping that by reusing the 'InfoListAdapter` the comment views which were created would automatically have the reply handling code inside of it. How are the comment views which are being created here different than the ones which are created in the main comment holding view?

app/src/main/res/layout/list_comments_item.xml
golfinq: I am more inclined to use the first option as I am trying to keep the amount of new code minimal for future maintenance reasons. I was hoping that by reusing the 'InfoListAdapter` the comment views which were created would automatically have the reply handling code inside of it. How are the comment views which are being created here different than the ones which are created in the main comment holding view?

Stypox:
Oh, btw @golfinq, could you fix the build so that I can take a look at the apk? I think you forgot to update the extractor commit

app/src/main/java/org/schabi/newpipe/info_list/holder/RepliesHandler.java
```diff
@@ -0,0 +1,192 @@
...
+                                          commentsInfoItemInfoItemsPage) {
+
+                final List<CommentsInfoItem> actualList
+                        = commentsInfoItemInfoItemsPage.getItems();
+
+                cachedReplies.addAll(actualList);
+                addRepliesToUI(parentInfoItem);
+            }
+
+            @SuppressLint("SetTextI18n")
```
Stypox, line 93: There is a reason why Android Studio gives you this warning: you shouldn't hardcode strings, put them in the `strings.xml` resource file instead.

app/src/main/java/org/schabi/newpipe/info_list/holder/RepliesHandler.java
Stypox: > as it currently stands this code can't handle recursive levels of replies

This is what I could guess by looking at the code, but I might be wrong since I couldn't test. But the problem is also the fact that you shouldn't store data about whether the replies menu is open or not in an object coming from the extractor. The extractor shouldn't have anything to do with the UI.

golfinq:
I don't really have time at the moment to work on this, so I am going to convert it to a draft until I can properly work on everyone's suggestions.

pc00per:
Undraft when 😂

golfinq:
Honestly, I am not sure. This will require time I don't currently have IRL.
I would also have to re-familiarize myself with java and the code to get
back up to speed, which is a non-trivial task.

opusforlife2:
@golfinq Then closing this PR for now. You can reopen it if/when you resume working on it.

mhmdanas:
You don't have to do anything to get the CI to build APKs for you; pushing a branch to your fork should be sufficient.

DUOLabs333:
So, I just push, and it builds? Cool. So, what are the problems?

mhmdanas:
I'm not too sure myself, but reading the comments and reviews here should give you an idea.

DUOLabs333:
I'm going to start out with a rebase, just to get myself situated.

DUOLabs333:
Will the actions be counted towards my limit or the team's limit?

mhmdanas:
I believe there's no limit on public repositories.

DUOLabs333:
I had to manually run it, it doesn't work on push.

mhmdanas:
Hmm, actually, I think it runs automatically only on push to a pull request branch. If you create a pull request from your branch to this repo, I think it will be run automatically.

DUOLabs333:
Oh, ok. I'm getting some style errors, so I'm trying to fix that.

DUOLabs333:
I somehow got `public void addReplies` in `CommentsInfoItemHolder.java` when it doesn't exist in either `dev` or `comment-replies`.

DUOLabs333:
Huh, it's not here: https://github.com/golfinq/NewPipe/blob/comment-replies

DUOLabs333:
I'm going to redo the rebase.

DUOLabs333:
Oh, it got moved, but the rebase didn't pick up on it.

DUOLabs333:
I'll open up a new PR.

DUOLabs333:
I just need approval to start running CI tests.

Mhowser:
@golfinq can this pull request be resurrected? Or are you still too busy?

golfinq:
There was some good work done on cleaning it up, but it _really_  _really_ needs someone who is familiar with the newpipe infrastructure (or someone who is willing to dig through it and learn how) to find out how to extract and implement the  extraction for the 2nd page of replies. There was discussuon that the ui should be changed to a page which pulls up from the bottom which I like a lot more but I wouldnt know how to implement. As it stands the core stuff in this PR is okay, but async/java stuff isnt my strength right now and the people who would know the core code stuff are difficult to get responses from. I've kinda moved on from this and while I still do use newpipe; its kind of a mess to dig through the source code and figure out "okay to do this I use this class and then input this into this and need to call all of these functions and this data goes here and can be accessed with this class". If I was to work on this again I would want through documentation of everything, which is a lot of work. Dont blame NewPipe devs though as supporting youtube and everything else makes it very very difficult to do what NewPipe does and that requires complicated code.

Mhowser:
I appreciate the update on the current situation, thank you! There are some people at #2277 who might be able to finish what was started here.

golfinq:
I am glad to see people still trying to implement this! It's why my NewPipe Repo will never be deleted - as I think this PR was the furthest anyone got.

Just to add on my previous post: If anyone reading this has any android dev knowledge (or is willing to learn it) please write some API documentation for NewPipe. Even if you are a complete beginner, if you can read code and are willing to follow the "open source" trail for something in android studio - you can write documentation. Even basic, but organized, documentation help people become developers which means more active development overall.

PR #7259, opened 2021-10-16
Add better network message

laksh-21:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- **What the changes achieve:** for actions that require an Internet connection like, `Play here on background` and `Play here on Popup`, earlier a vague `Toast` message saying `Could not play this stream` would appear but a network connection is a common issue and it would've been helpful to know if that's what was wrong. So, these changes implement a `Snackbar` message for the same with an action to go to the `Downloads` section for offline viewing.
- The changes add a `boolean requireNetwork` parameter to `StreamDialogEntry` that is `false` by default and can be changed via the `setRequireNetwork` method. If `true`, when the option is clicked on, before performing the action, network availability is checked, and if a network is unavailable, an indefinite `Snackbar` is shown.
- I went with this approach since custom actions can be defined for a  `StreamDialogEntry` and maybe more options might be added later that require network connectivity. They will be easier to integrate into these checks.

- Before: 
before

- After:
before

- Fixes ##6381

#### APK testing 

[debug.apk](https://github.com/TeamNewPipe/NewPipe/suites/4071163970/artifacts/103515630)

Also, apologies for taking so long :( I had exams.
And sorry about the second commit :( I left an unused import in there somehow.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
+++ b/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
@@ -6,7 +6,6 @@ import android.content.Context;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.net.Uri;
-import android.util.Log;
 import android.widget.Toast;
 
 import androidx.fragment.app.Fragment;
--- a/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
+++ b/app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
@@ -1,10 +1,17 @@
 package org.schabi.newpipe.util;
 
+import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
+
 import android.content.Context;
+import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
 import android.net.Uri;
+import android.util.Log;
 import android.widget.Toast;
 
 import androidx.fragment.app.Fragment;
+
+import com.google.android.material.snackbar.Snackbar;
 
 import org.schabi.newpipe.NewPipeDatabase;
 import org.schabi.newpipe.R;
@@ -22,8 +29,6 @@ import java.util.List;
 
 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
 import io.reactivex.rxjava3.schedulers.Schedulers;
-
-import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
 
 public enum StreamDialogEntry {
     //////////////////////////////////////
@@ -68,13 +73,15 @@ public enum StreamDialogEntry {
         NavigationHelper.enqueueNextOnPlayer(fragment.getContext(), new SinglePlayQueue(item));
     }),
 
-    start_here_on_background(R.string.start_here_on_background, (fragment, item) ->
-            NavigationHelper.playOnBackgroundPlayer(fragment.getContext(),
-                    new SinglePlayQueue(item), true)),
+    start_here_on_background(R.string.start_here_on_background, (fragment, item) -> {
+        NavigationHelper.playOnBackgroundPlayer(fragment.getContext(),
+                new SinglePlayQueue(item), true);
+    }),
 
-    start_here_on_popup(R.string.start_here_on_popup, (fragment, item) ->
-            NavigationHelper.playOnPopupPlayer(fragment.getContext(),
-                    new SinglePlayQueue(item), true)),
+    start_here_on_popup(R.string.start_here_on_popup, (fragment, item) -> {
+        NavigationHelper.playOnPopupPlayer(fragment.getContext(),
+                new SinglePlayQueue(item), true);
+    }),
 
     set_as_playlist_thumbnail(R.string.set_as_playlist_thumbnail, (fragment, item) -> {
     }), // has to be set manually
@@ -128,11 +135,13 @@ public enum StreamDialogEntry {
     private final int resource;
     private final StreamDialogEntryAction defaultAction;
     private StreamDialogEntryAction customAction;
+    private boolean requireNetwork;
 
     StreamDialogEntry(final int resource, final StreamDialogEntryAction defaultAction) {
         this.resource = resource;
         this.defaultAction = defaultAction;
         this.customAction = null;
+        this.requireNetwork = false;
     }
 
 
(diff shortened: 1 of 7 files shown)


Stypox:
> Maybe link to the error log instead? That could be helpful in providing bug reports related to network issues.

This should be done only if the error really comes from the player, but if I understand correctly this PR does not do that. If there is just an `if` before executing `StreamDialogEntry`'s code that checks for internet access, then there is obviously no real error to report, hence we should show no button at all.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -128,11 +134,13 @@
...
     private final StreamDialogEntryAction defaultAction;
     private StreamDialogEntryAction customAction;
+    private boolean requireNetwork;
 
     StreamDialogEntry(final int resource, final StreamDialogEntryAction defaultAction) {
         this.resource = resource;
         this.defaultAction = defaultAction;
         this.customAction = null;
+        this.requireNetwork = false;
     }
```
Stypox, line 144: Instead of manually using `.setRequireNetwork(true);` in every place the actions are used, I'd just change the constructor like this, since the actions that require network access are always the same.
```suggestion
    StreamDialogEntry(final int resource, final boolean requireNetwork, final StreamDialogEntryAction defaultAction) {
        this.resource = resource;
        this.requireNetwork = requireNetwork;
        this.defaultAction = defaultAction;
        this.customAction = null;
    }
```
If you do this, obviously remove the `setRequireNetwork` method.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -206,4 +229,30 @@ private static void openChannelFragment(final Fragment fragment,
...
+
+    private static boolean isNetworkAvailable(final Context context) {
+        final ConnectivityManager cm =
+                (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+
+        final NetworkInfo activeNetwork = cm.getActiveNetworkInfo();
+        return activeNetwork != null
+                && activeNetwork.isConnectedOrConnecting();
+    }
```
Stypox, line 240: Note for a future PR: this method and all of the other methods related to `ConnectivityManager` distributed across the app should be moved into a separate `util.` class.

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -206,4 +229,30 @@ private static void openChannelFragment(final Fragment fragment,
...
                 item.getServiceId(), uploaderUrl, item.getUploaderName());
     }
+
+    private static boolean isNetworkAvailable(final Context context) {
+        final ConnectivityManager cm =
+                (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+
+        final NetworkInfo activeNetwork = cm.getActiveNetworkInfo();
+        return activeNetwork != null
+                && activeNetwork.isConnectedOrConnecting();
```
Stypox, line 239:
```suggestion
        final ConnectivityManager manager
                = ContextCompat.getSystemService(context, ConnectivityManager.class);
        if (manager == null) {
            return true; // the system service is not available, so we can't know
        }

        final NetworkInfo activeNetwork = manager.getActiveNetworkInfo();
        return activeNetwork != null && activeNetwork.isConnectedOrConnecting();
```

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -206,4 +229,30 @@ private static void openChannelFragment(final Fragment fragment,
...
+
+    /////////////////////////////////////////////////////////
+    // private method to show unavailable network Snackbar //
+    /////////////////////////////////////////////////////////
+
+    private static void showNetworkUnavailableSnackbar(final Fragment fragment) {
+        final Snackbar snackbar = Snackbar.make(
+                fragment.requireView(),
+                R.string.error_no_network,
+                Snackbar.LENGTH_INDEFINITE);
```
Stypox, line 251: Uhm, indefinite length is not a good idea, I'd just use `LENGTH_SHORT`, it doesn't take much time to read a snackbar that says "Network not available"

app/src/main/res/values/strings.xml
```diff
@@ -704,4 +704,6 @@
     <!-- Show Channel Details -->
     <string name="error_show_channel_details">Error at Show Channel Details</string>
     <string name="loading_channel_details">Loading Channel Details…</string>
+    <string name="error_no_network">No Network Available</string>
+    <string name="error_no_network_action">Goto Downloads</string>
```
Stypox, line 708:
```suggestion
    <string name="error_no_network">Network not available</string>
```

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -206,4 +229,30 @@ private static void openChannelFragment(final Fragment fragment,
...
+        snackbar.setAction(R.string.error_no_network_action, view ->
+                NavigationHelper.openDownloads(fragment.requireActivity()));
```
Stypox, line 253: Yeah, remove this, like said

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
```diff
@@ -206,4 +229,30 @@ private static void openChannelFragment(final Fragment fragment,
...
+        // Snackbar needs to adjust its position if the system has gesture navigation
+        snackbar.setGestureInsetBottomIgnored(true);
```
Stypox, line 255: Mmmh, we aren't doing this with all other `Snackbar`s across the app, why do it here?

Stypox (changes requested):
Maybe the action of the `Snackbar` could be something like "Try anyway", so that the user can decide to try to do the action anyway even if network is not available, what do you think?

app/src/main/java/org/schabi/newpipe/util/StreamDialogEntry.java
opusforlife2: How short is `LENGTH_SHORT`? 5 seconds?

Stypox: 2 seconds (LENGTH_LONG is 3.5 seconds): https://stackoverflow.com/questions/7965135/what-is-the-duration-of-a-toast-length-long-and-length-short

Stypox:
I'm closing this since we went down a different path and added error notifications. Now the player correctly reports error in a notification, and the enqueueing process also does, so this PR is not relevant anymore. Thank you anyway :-)

PR #7260, opened 2021-10-16
Add the quality selector for livestreams

### LINKED ISSUE #1239: Quality selector in lives
I know somebody told that the quality selector is useless for lives because ot set automaticaly the 'best' quality, but i dont agree with that :  when i start a live on Newpipe I have a shitty quality most of the time, whereas I can watch 720p 60fps lives on twitch... So we should have an access to set manually the quality, or maybe choose in settings a default quality (like for videos)

AudricV:
#### What is it?
- [x] Feature (user facing)

This PR adds the ability to control the quality played for livestreams in NewPipe, both on main and popup players. The livestream quality selector uses the same popup menu as the quality selector for video streams.

The resolution limit on mobile data on livestreams also now works with this PR (even if you need to reload the livestream if you started to play it on Wifi and then you switched to mobile data) with the `Auto` option. You can still choose an higher resolution than the limit you set in `Video and audio` settings, like for videos.

Without setting a limit on mobile data or when playing a livestream on Wifi, the `Auto` option does the same thing as the current behavior (the quality played is managed by ExoPlayer and is, most of times, the highest available).

However, the default video resolution on popup player and main player settings are not working. I don't really know how to do this and I don't really have the time to do it.

- Before:

  Player controls livestreams release

- After:

  Player controls livestreams with quality selector

- Fixes #1239

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -12,6 +12,7 @@ import static com.google.android.exoplayer2.Player.REPEAT_MODE_ALL;
 import static com.google.android.exoplayer2.Player.REPEAT_MODE_OFF;
 import static com.google.android.exoplayer2.Player.REPEAT_MODE_ONE;
 import static com.google.android.exoplayer2.Player.RepeatMode;
+import static com.google.android.exoplayer2.util.MimeTypes.getVideoMediaMimeType;
 import static org.schabi.newpipe.QueueItemMenuUtil.openPopupMenu;
 import static org.schabi.newpipe.extractor.ServiceList.YouTube;
 import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
@@ -3417,21 +3418,23 @@ public final class Player implements
 
             for (int i = 0; i < videoTrackGroupArray.get(0).length; i++) {
                 final Format format = videoTrackGroupArray.get(0).getFormat(i);
-                final StringBuilder stringBuilder = new StringBuilder();
+                final StringBuilder resolutionStringBuilder = new StringBuilder();
                 final int formatHeight = format.height;
 
                 if (formatHeight == Format.NO_VALUE) {
-                    stringBuilder.append(format.bitrate);
+                    resolutionStringBuilder.append(format.bitrate);
+                    resolutionStringBuilder.append("bps");
                 } else {
-                    stringBuilder.append(formatHeight);
-                    stringBuilder.append("p");
+                    resolutionStringBuilder.append(formatHeight);
+                    resolutionStringBuilder.append("p");
                     final int formatFrameRate = (int) format.frameRate;
                     if (formatFrameRate > 30) {
-                        stringBuilder.append(formatFrameRate);
+                        resolutionStringBuilder.append(formatFrameRate);
                     }
                 }
 
-                menu.add(POPUP_MENU_ID_QUALITY, i + 1, Menu.NONE, stringBuilder.toString());
+                menu.add(POPUP_MENU_ID_QUALITY, i + 1, Menu.NONE,
+                        resolutionStringBuilder.toString());
             }
 
             final TrackSelectionListener trackSelectionListener = new TrackSelectionListener(
@@ -3451,14 +3454,40 @@ public final class Player implements
             public void onDownstreamFormatChanged(@NonNull final EventTime eventTime,
                                                   @NonNull final MediaLoadData mediaLoadData) {
                 final Format currentPlayingFormat = mediaLoadData.trackFormat;
-                if (currentPlayingFormat != null) {
-                    final int resolution = currentPlayingFormat.height;
-                    final int frameRate = (int) currentPlayingFormat.frameRate;
+                if (currentPlayingFormat != null
+                        && getVideoMediaMimeType(currentPlayingFormat.codecs) != null) {
+                    final StringBuilder resolutionStringBuilder = new StringBuilder();
+                    final int formatHeight = currentPlayingFormat.height;
 
-                    if (frameRate > 30) {
-                        binding.qualityTextView.setText(resolution + "p" + frameRate);
+                    if (formatHeight == Format.NO_VALUE) {
+                        resolutionStringBuilder.append(currentPlayingFormat.bitrate);
+                        resolutionStringBuilder.append("bps");
                     } else {
-                        binding.qualityTextView.setText(resolution + "p");
+                        resolutionStringBuilder.append(formatHeight);
+                        resolutionStringBuilder.append("p");
+                        final int formatFrameRate = (int) currentPlayingFormat.frameRate;
+                        if (formatFrameRate > 30) {
+                            resolutionStringBuilder.append(formatFrameRate);
+                        }
+                    }
+
+                    final MappingTrackSelector.MappedTrackInfo currentMappedTrackInfo =
+                            trackSelector.getCurrentMappedTrackInfo();
+                    final int videoTrackGroupIndex = getVideoRendererIndex();
+
+                    // videoTrackGroupArray shouldn't be null because this method is called only if
+                    // a media source is available
+                    final TrackGroupArray videoTrackGroupArray = Objects.requireNonNull(
+                            currentMappedTrackInfo).getTrackGroups(videoTrackGroupIndex);
+
+                    if (trackSelector.getParameters().hasSelectionOverride(videoTrackGroupIndex,
+                            videoTrackGroupArray)) {
+                        binding.qualityTextView.setText(resolutionStringBuilder.toString());
+                    } else {
+                        binding.qualityTextView.setText(context.getString(
+                                R.string.auto_quality_selected,
+                                context.getString(R.string.auto_quality),
+                                resolutionStringBuilder.toString()));
                     }
                 }
             }
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -12,8 +12,6 @@ import static com.google.android.exoplayer2.Player.REPEAT_MODE_ALL;
 import static com.google.android.exoplayer2.Player.REPEAT_MODE_OFF;
 import static com.google.android.exoplayer2.Player.REPEAT_MODE_ONE;
 import static com.google.android.exoplayer2.Player.RepeatMode;
-import static com.google.android.exoplayer2.util.MimeTypes.isText;
-import static com.google.android.exoplayer2.util.MimeTypes.isVideo;
 import static org.schabi.newpipe.QueueItemMenuUtil.openPopupMenu;
 import static org.schabi.newpipe.extractor.ServiceList.YouTube;
 import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
@@ -124,15 +122,15 @@ import com.google.android.exoplayer2.Player.PositionInfo;
 import com.google.android.exoplayer2.RenderersFactory;
 import com.google.android.exoplayer2.SimpleExoPlayer;
 import com.google.android.exoplayer2.Timeline;
-import com.google.android.exoplayer2.analytics.AnalyticsListener;
 import com.google.android.exoplayer2.source.BehindLiveWindowException;
-import com.google.android.exoplayer2.source.MediaLoadData;
 import com.google.android.exoplayer2.source.MediaSource;
 import com.google.android.exoplayer2.source.TrackGroup;
 import com.google.android.exoplayer2.source.TrackGroupArray;
 import com.google.android.exoplayer2.text.Cue;
+import com.google.android.exoplayer2.trackselection.BaseTrackSelection;
 import com.google.android.exoplayer2.trackselection.DefaultTrackSelector;
 import com.google.android.exoplayer2.trackselection.MappingTrackSelector;
+import com.google.android.exoplayer2.trackselection.TrackSelection;
 import com.google.android.exoplayer2.trackselection.TrackSelectionArray;
 import com.google.android.exoplayer2.ui.AspectRatioFrameLayout;
 import com.google.android.exoplayer2.ui.CaptionStyleCompat;
@@ -343,6 +341,9 @@ public final class Player implements
     private PopupMenu playbackSpeedPopupMenu;
     private PopupMenu captionPopupMenu;
 
+    private boolean isDefaultResolutionOverrideOrMaxVideoSizeAndFrameRateSet = false;
+    private boolean qualityPopupMenuBuiltForLivestreams = false;
+
     /*//////////////////////////////////////////////////////////////////////////
     // Popup player
     //////////////////////////////////////////////////////////////////////////*/
@@ -430,7 +431,7 @@ public final class Player implements
         return new VideoPlaybackResolver.QualityResolver() {
             @Override
             public int getDefaultResolutionIndex(final StreamInfo streamInfo,
-                    final List<VideoStream> sortedVideos) {
+                                                 final List<VideoStream> sortedVideos) {
                 if (StreamTypeUtil.isLiveStream(streamInfo.getStreamType())) {
                     return -1;
                 }
@@ -2442,6 +2443,7 @@ public final class Player implements
         }
         maybeUpdateCurrentMetadata();
         onTextTracksChanged();
+        updateQualityPlayedStringAndQualityElementsForLivestreamsIfNeeded(trackSelections);
     }
 
     @Override
@@ -3273,21 +3275,33 @@ public final class Player implements
     @Override // own playback listener
     @Nullable
     public MediaSource sourceOf(final PlayQueueItem item, final StreamInfo info) {
-        final MediaSource mediaSource = (isAudioOnly ? audioResolver : videoResolver)
-                .resolve(info);
-        if ((isAudioOnly ? audioResolver : videoResolver).isLiveSource()) {
-            setDefaultResolutionLimit();
-        }
-        return mediaSource;
+        return (isAudioOnly ? audioResolver : videoResolver).resolve(info);
     }
 
     public void disablePreloadingOfCurrentTrack() {
         loadController.disablePreloadingOfCurrentTrack();
     }
 
-    private void setDefaultResolutionLimit() {
-        final String defaultResolution;
+    /**
+     * Set the default resolution for livestreams according to user choice in "Video and audio"
+     * settings.
+     *
+     * <p>
+     * If no stream match the minimum criteria required (video format height <= default height),
+     * then the maximum resolution limit and the maximum frame rate will be set.
+     * </p>
+     *
+     * @param videoTrackGroupArray the video {@link TrackGroupArray}, which cannot be null
+     * @param videoRendererIndex   the video renderer index
+     */
+    private void setDefaultResolutionForLivestreamsIfNeeded(
+            @NonNull final TrackGroupArray videoTrackGroupArray,
+            final int videoRendererIndex) {
+        if (isDefaultResolutionOverrideOrMaxVideoSizeAndFrameRateSet) {
+            return;
+        }
 
+        final String defaultResolution;
         if (popupPlayerSelected()) {
             defaultResolution = computeDefaultResolution(context,
                     R.string.default_popup_resolution_key,
@@ -3296,23 +3310,43 @@ public final class Player implements
             defaultResolution = computeDefaultResolution(context,
                     R.string.default_resolution_key, R.string.default_resolution_value);
         }
+        final String[] defaultResolutionArray = defaultResolution.split("p");
+        final int defaultHeight = !defaultResolution.equals(context.getString(
+                R.string.best_resolution_key))
+                ? Integer.parseInt(defaultResolutionArray[0]) : -1;
+        final int defaultFrameRate = defaultResolutionArray.length == 2
+                ? Integer.parseInt(defaultResolutionArray[1]) : -1;
 
-        // Best resolution set: not needed to set a maximum resolution
-        if (!defaultResolution.equals(context.getString(R.string.best_resolution_key))) {
-            final String[] defaultResolutionArray = defaultResolution.split("p");
-            final int defaultHeight = Integer.parseInt(defaultResolutionArray[0]);
+        final DefaultTrackSelector.Parameters selectionParameters = trackSelector
+                .getParameters();
+        final DefaultTrackSelector.ParametersBuilder builder = selectionParameters
+                .buildUpon().clearSelectionOverrides();
 
-            final DefaultTrackSelector.Parameters selectionParameters = trackSelector
-                    .getParameters();
-            final DefaultTrackSelector.ParametersBuilder builder = selectionParameters.buildUpon()
-                    .clearSelectionOverrides();
+        for (int i = 0; i < videoTrackGroupArray.length; i++) {
+            final TrackGroup trackGroup = videoTrackGroupArray.get(i);
+            for (int j = trackGroup.length - 1; j >= 0; j--) {
+                final Format videoFormat = trackGroup.getFormat(j);
+                if ((defaultHeight == -1 || videoFormat.height <= defaultHeight)
+                        && (defaultFrameRate <= -1 || videoFormat.frameRate == defaultFrameRate)) {
+                    final DefaultTrackSelector.SelectionOverride selectionOverride =
+                            new DefaultTrackSelector.SelectionOverride(
+                                    videoRendererIndex, j);
 
-            builder.setMaxVideoSize(Integer.MAX_VALUE, defaultHeight);
-            builder.setMaxVideoFrameRate(defaultResolutionArray.length == 2
-                    ? Integer.parseInt(defaultResolutionArray[1]) : 30);
-
-            trackSelector.setParameters(builder);
+                    builder.setSelectionOverride(videoRendererIndex, videoTrackGroupArray,
+                            selectionOverride);
+                    trackSelector.setParameters(builder);
+                    isDefaultResolutionOverrideOrMaxVideoSizeAndFrameRateSet = true;
+                    return;
+                }
+            }
         }
+
+        builder.setMaxVideoSize(Integer.MAX_VALUE, defaultHeight);
+        builder.setMaxVideoFrameRate(defaultResolutionArray.length == 2
+                ? Integer.parseInt(defaultResolutionArray[1]) : 30);
+
+        trackSelector.setParameters(builder);
+        isDefaultResolutionOverrideOrMaxVideoSizeAndFrameRateSet = true;
     }
 
     @Nullable
@@ -3349,8 +3383,6 @@ public final class Player implements
 
             case LIVE_STREAM:
                 if (!isNullOrEmpty(info.getHlsUrl()) || !isNullOrEmpty(info.getDashMpdUrl())) {
-                    buildLiveStreamQualityMenu();
-                    addQualityTextViewQualityChangeEvent();
                     binding.qualityTextView.setVisibility(View.VISIBLE);
                 } else if (info.getVideoStreams().size()
                         + info.getVideoOnlyStreams().size() != 0) {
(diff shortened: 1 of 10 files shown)


ResuUman:
Checked usage _only_ on mobile data and regular player (no popup) with auto resolution:
Live streams:
- do not uses "default resolution" option
- changing "limit resolution (...)" to "no limit" has no effect on _current_ playing live stream (nor pausing nor auto->240p->auto did not helped)
- ~~works good enough.~~
- ~~starts with 144p/240p/360p on 2G/3G/4G (or with last played resolution when second video is played)~~
- ~~uses "limit resolution when using mobile data" option as (max) target resolution.~~
- ~~forcing change in network connection 2G<->3G<->4G during play give us change in resolution as expected~~
- ~~auto resolution caped at 720p60 (max) for my connection (1080p only after manual pick)~~
- ~~changing value "limit resolution when using mobile data" had proper effect on next live stream played (and is mostly OK for current stream)~~
- ~~couldn't find live stream with missing resolution (ie. not having 240p) (it would be good to test such stream if they exist)~~

~~Non live streams:
This build handles resolution picker better than previous release versions.
Missing resolutions in streams (ie. stream having only 144p, 360p; 480p; 720p) are not handled perfectly.~~

TiA4f8R:
@ResuUman

-  > - do not uses "default resolution" option

   This is currently expected:
   > However, the default video resolution on popup player and main player settings are not working.

<br>

-  > changing "limit resolution (...)" to "no limit" has no effect on current playing live stream (nor pausing nor auto->240p->auto did not helped)

   Unfortunately, I don't how to fix this and I do not have the time to explore this right now. Sorry.

Anyway, thank you for your tests! With what you said, it looks like the biggest part of the feature's components I implemented is working.

litetex (dismissed):
LGTM (very good javadoc 👍)

Smoke tested it and seems to work as expected.

@TiA4f8R 
Could you create follow up issues when this is merged?

Would be great if someone else could maybe also "longtime" test that.

Nfnfjfjjf:
> LGTM (very good javadoc 👍)
> 
> Smoke tested it and seems to work as expected.
> 
> @TiA4f8R Could you create follow up issues when this is merged?
> 
> Would be great if someone else could maybe also "longtime" test that.

Long time Tested and works as it supposed to

app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java
```diff
@@ -128,8 +136,10 @@ public void setPlaybackQuality(@Nullable final String playbackQuality) {
     }
 
     public interface QualityResolver {
-        int getDefaultResolutionIndex(List<VideoStream> sortedVideos);
+        int getDefaultResolutionIndex(StreamInfo streamInfo, List<VideoStream> sortedVideos);
```
Stypox, line 139: Mmmh, by looking at the code above it seems like `StreamInfo` will be of livestream type only when `maybeBuildLiveMediaSource` fails (which should not happen for livestreams, unless `getHlsUrl` or `getDashMpdUrl` return empty urls) (`VideoPlaybackResolver` is the only place where `getDefaultResolutionIndex` and `getOverrideResolutionIndex` are *used*). So having `if (StreamTypeUtil.isLiveStream(streamInfo.getStreamType())) { return -1; }` in the player above is useless, I think (that's the only place where `getDefaultResolutionIndex` and `getOverrideResolutionIndex` are *overridden*).

app/src/main/java/org/schabi/newpipe/player/playback/TrackSelectionListener.java
```diff
@@ -0,0 +1,106 @@
...
+    public boolean onMenuItemClick(@NonNull final MenuItem item) {
+        final DefaultTrackSelector.Parameters selectionParameters = trackSelector.getParameters();
+        final boolean isDisabled = selectionParameters.getRendererDisabled(rendererIndex);
+        final DefaultTrackSelector.ParametersBuilder builder = selectionParameters.buildUpon()
+                .clearSelectionOverrides(rendererIndex)
+                .setRendererDisabled(rendererIndex, isDisabled);
+        // The auto string
+        final String qualityTextViewText = qualityTextView.getText().toString();
+
+        if (item.getItemId() != 0) {
```
Stypox, line 68: Invert the two branches of this `if`, and explain that `0` is always the index of `auto`

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3273,13 +3289,49 @@ public void onStartDrag(final PlayQueueItemHolder viewHolder) {
...
+            final int defaultFrameRate = defaultResolutionArray.length == 2
+                    ? Integer.parseInt(defaultResolutionArray[1]) : -1;
+
+            final DefaultTrackSelector.Parameters selectionParameters = trackSelector
+                    .getParameters();
+            final DefaultTrackSelector.ParametersBuilder builder = selectionParameters.buildUpon()
+                    .clearSelectionOverrides();
+
+            builder.setMaxVideoSize(Integer.MAX_VALUE, defaultHeight);
+            builder.setMaxVideoFrameRate(defaultFrameRate != -1 ? defaultFrameRate : 30);
```
Stypox, line 3329: And remove the variable above
```suggestion
            builder.setMaxVideoFrameRate(defaultResolutionArray.length == 2
                    ? Integer.parseInt(defaultResolutionArray[1]) : 30);
```

Stypox (changes requested):
Code looks mostly good, but I am not an expert of the player. @Redirion could you also review?

app/src/main/java/org/schabi/newpipe/player/playback/TrackSelectionListener.java
```diff
@@ -0,0 +1,106 @@
...
+            // player)
+            if (qualityTextViewText.contains(context.getString(R.string.auto_quality))) {
+                final Format currentPlayingFormat = videoTrackGroup.getFormat(qualityId);
+                qualityTextView.setText(getResolutionStringFromFormat(context,
+                        currentPlayingFormat));
+            }
+
+            trackSelector.setParameters(builder);
+        } else {
+            trackSelector.setParameters(builder);
```
litetex, line 91: Looks similar, maybe deduplicateable?

TiA4f8R: I think we can't, otherwise the previous quality played will be set as the quality currently played (because the event listener which update the quality will be not called) so I think this will lead to an issue in a specific case.

TiA4f8R:
When I worked on this PR first, I tried to set the quality played based on user settings and not to limit the maximum resolution which can be played.

But it wasn't working because the playback wasn't intialized when the setting of my selection override was called (so the currentMappedTrackInfo was null) so I was not able to set the selection override.

Is there any way to do what I said instead of the current behavior (which is setting the maximum resolution allowed)?

@Redirion I think you are the best to know how to do this, that's why I am asking for your help.

app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java
TiA4f8R: Not sure about that, I'd like to hear advice of other team members.

Stypox: Well, let's keep it and just add a comment explaining that `return -1` should be unreachable

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -416,18 +426,27 @@ public Player(@NonNull final MainPlayer service) {
...
+    @NonNull
     private VideoPlaybackResolver.QualityResolver getQualityResolver() {
         return new VideoPlaybackResolver.QualityResolver() {
             @Override
-            public int getDefaultResolutionIndex(final List<VideoStream> sortedVideos) {
+            public int getDefaultResolutionIndex(final StreamInfo streamInfo,
+                                                 final List<VideoStream> sortedVideos) {
+                if (StreamTypeUtil.isLiveStream(streamInfo.getStreamType())) {
+                    return -1;
+                }
```
Stypox, line 437: If it wasn't clear I meant here ...

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -416,18 +426,27 @@ public Player(@NonNull final MainPlayer service) {
...
             }
 
             @Override
-            public int getOverrideResolutionIndex(final List<VideoStream> sortedVideos,
+            public int getOverrideResolutionIndex(final StreamInfo streamInfo,
+                                                  final List<VideoStream> sortedVideos,
                                                   final String playbackQuality) {
+                if (StreamTypeUtil.isLiveStream(streamInfo.getStreamType())) {
+                    return -1;
+                }
```
Stypox, line 449: ... and here in my comment in the other review

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3364,13 +3470,229 @@ private void buildQualityMenu() {
...
+     * <p>
+     * This method will also update, for the first time it's called, the default resolution,
+     * according to user settings, with
+     * {@link #setDefaultResolutionForLivestreamsIfNeeded(TrackGroupArray, int)} and will build
+     * the popup quality menu {@link #buildLiveStreamQualityMenuIfNeeded(TrackGroupArray, int)}.
+     * </p>
+     * @param trackSelections the {@link TrackSelectionArray} of ExoPlayer's track selections,
+     *                        which cannot be null
+     */
+    private void updateQualityPlayedStringAndQualityElementsForLivestreamsIfNeeded(
```
Stypox, line 3571: Please use a shorter name (e.g. `updateQualityForLivestreamsIfNeeded`), the Javadoc explains the rest

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3260,6 +3280,73 @@ public void disablePreloadingOfCurrentTrack() {
...
+                if ((defaultHeight == -1 || videoFormat.height <= defaultHeight)
+                        && (defaultFrameRate <= -1 || videoFormat.frameRate == defaultFrameRate)) {
+                    final DefaultTrackSelector.SelectionOverride selectionOverride =
+                            new DefaultTrackSelector.SelectionOverride(
+                                    videoRendererIndex, j);
+
+                    builder.setSelectionOverride(videoRendererIndex, videoTrackGroupArray,
+                            selectionOverride);
+                    trackSelector.setParameters(builder);
+                    isDefaultResolutionOverrideOrMaxVideoSizeAndFrameRateSet = true;
```
Stypox, line 3336: This is being done independently of from where the function returns, so it should not be duplicated, but should just be put below `if (isDefaultResolutionOverrideOrMaxVideoSizeAndFrameRateSet) { return; }`

app/src/main/java/org/schabi/newpipe/player/playback/TrackSelectionListener.java
```diff
@@ -0,0 +1,122 @@
...
+            if (qualityTextViewText.contains(context.getString(R.string.auto_quality))) {
+                final Format currentPlayingFormat = videoTrackGroup != null
+                        ? videoTrackGroup.getFormat(qualityId)
+                        // Fallback to the first track group
+                        : rendererTrackGroups.get(0).getFormat(qualityId);
+                qualityTextView.setText(getResolutionStringFromFormat(context,
+                        currentPlayingFormat));
+            }
+
+            trackSelector.setParameters(builder);
```
Stypox, line 117: @litetex is right, this can be put at the bottom of the function just before `return true` since it is called always.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -334,6 +341,9 @@
     private PopupMenu playbackSpeedPopupMenu;
     private PopupMenu captionPopupMenu;
 
+    private boolean isDefaultResolutionOverrideOrMaxVideoSizeAndFrameRateSet = false;
```
Stypox, line 344: Also here, maybe use `isUsingDefaultResolutionForLivestreams` which is shorter and more meaningful. I can't understand what `isDefaultResolutionOverrideOrMaxVideoSizeAndFrameRateSet` means, I just see that it is only used inside `setDefaultResolutionForLivestreams` and it prevents that function from being executed multiple times

Stypox (changes requested):

TiA4f8R:
Marking as draft because it seems the quality selected automatically is not updated in the quality selected textview.

app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java
TiA4f8R: > So having `if (StreamTypeUtil.isLiveStream(streamInfo.getStreamType())) { return -1; }` in the player above is useless

Indeed, I just look better at the code and you are right, that's useless to set `-1` and this can even create bugs because no video source would be available for livestreams (because `MediaSourceTag.getSelectedVideoStream()` will return `null`).

So I will rollback my changes here and keep code as it was before.

app/src/main/java/org/schabi/newpipe/player/Player.java
TiA4f8R: Outdated

app/src/main/java/org/schabi/newpipe/player/Player.java
TiA4f8R: Outdated

app/src/main/java/org/schabi/newpipe/player/Player.java
TiA4f8R: Is `updateQualityElementsForLivestreams` OK?

app/src/main/java/org/schabi/newpipe/player/Player.java
TiA4f8R: Changed to `isDefaultResolutionSetForLivestreams`.

Redirion (approved):
tested it, seems to work just fine (default quality selection, manual quality selection)

litetex (changes requested):
<details><summary>Optimizations I found so far (quick check - reviewed just one class)</summary> 

```diff
diff --git a/app/src/main/java/org/schabi/newpipe/player/Player.java b/app/src/main/java/org/schabi/newpipe/player/Player.java
index 76709a001..24c816357 100644
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -197,10 +197,12 @@ import org.schabi.newpipe.util.external_communication.ShareUtils;
 import org.schabi.newpipe.views.ExpandableSurfaceView;

 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;

 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
 import io.reactivex.rxjava3.core.Observable;
@@ -3323,15 +3325,16 @@ public final class Player implements
             return;
         }

-        final String defaultResolution;
-        if (popupPlayerSelected()) {
-            defaultResolution = computeDefaultResolution(context,
-                    R.string.default_popup_resolution_key,
-                    R.string.default_popup_resolution_value);
-        } else {
-            defaultResolution = computeDefaultResolution(context,
-                    R.string.default_resolution_key, R.string.default_resolution_value);
-        }
+        final String defaultResolution = computeDefaultResolution(
+                context,
+                popupPlayerSelected()
+                        ? R.string.default_popup_resolution_key
+                        : R.string.default_resolution_key,
+                popupPlayerSelected()
+                        ? R.string.default_popup_resolution_value
+                        : R.string.default_resolution_value
+        );
+
         final String[] defaultResolutionArray = defaultResolution.split("p");
         final int defaultHeight = !defaultResolution.equals(context.getString(
                 R.string.best_resolution_key))
@@ -3339,10 +3342,10 @@ public final class Player implements
         final int defaultFrameRate = defaultResolutionArray.length == 2
                 ? Integer.parseInt(defaultResolutionArray[1]) : -1;

-        final DefaultTrackSelector.Parameters selectionParameters = trackSelector
-                .getParameters();
-        final DefaultTrackSelector.ParametersBuilder builder = selectionParameters
-                .buildUpon().clearSelectionOverrides();
+        final DefaultTrackSelector.Parameters selectionParameters =
+                trackSelector.getParameters();
+        final DefaultTrackSelector.ParametersBuilder builder =
+                selectionParameters.buildUpon().clearSelectionOverrides();

         for (int i = 0; i < videoTrackGroupArray.length; i++) {
             final TrackGroup trackGroup = videoTrackGroupArray.get(i);
@@ -3350,12 +3353,9 @@ public final class Player implements
                 final Format videoFormat = trackGroup.getFormat(j);
                 if ((defaultHeight == -1 || videoFormat.height <= defaultHeight)
                         && (defaultFrameRate <= -1 || videoFormat.frameRate <= defaultFrameRate)) {
-                    final DefaultTrackSelector.SelectionOverride selectionOverride =
-                            new DefaultTrackSelector.SelectionOverride(
-                                    videoRendererIndex, j);

                     builder.setSelectionOverride(videoRendererIndex, videoTrackGroupArray,
-                            selectionOverride);
+                            new DefaultTrackSelector.SelectionOverride(videoRendererIndex, j));
                     isDefaultResolutionSetForLivestreams = true;
                     break;
                 }
@@ -3367,19 +3367,20 @@ public final class Player implements
             // frame rate according to the default quality chose by the user in Video and Audio
             // settings
             builder.setMaxVideoSize(Integer.MAX_VALUE, defaultHeight);
-            builder.setMaxVideoFrameRate(defaultResolutionArray.length == 2
-                    ? Integer.parseInt(defaultResolutionArray[1]) : 30);
+            builder.setMaxVideoFrameRate(defaultFrameRate != -1 ? defaultFrameRate : 30);
             isDefaultResolutionSetForLivestreams = true;
         }

         trackSelector.setParameters(builder);
     }

-    @Nullable
-    public VideoStream getSelectedVideoStream() {
-        return (selectedVideoStreamIndex >= 0 && availableVideoStreams != null
-                && availableVideoStreams.size() > selectedVideoStreamIndex)
-                ? availableVideoStreams.get(selectedVideoStreamIndex) : null;
+    @NonNull
+    public Optional<VideoStream> getOptSelectedVideoStream() {
+        return (selectedVideoStreamIndex >= 0
+                && availableVideoStreams != null
+                && selectedVideoStreamIndex < availableVideoStreams.size())
+                ? Optional.ofNullable(availableVideoStreams.get(selectedVideoStreamIndex))
+                : Optional.empty();
     }

     private void updateStreamRelatedViews() {
@@ -3423,7 +3424,7 @@ public final class Player implements

                     availableVideoStreams = sortedAvailableStreams;
                     selectedVideoStreamIndex = currentMetadata.getSelectedVideoStreamIndex();
-                    updateQualityPlayedForVideos();
+                    updateQualityTextViewFromSelectedVideoStream();
                     buildVideoQualityMenuIfNeeded();
                     binding.qualityTextView.setVisibility(View.VISIBLE);
                 }
@@ -3446,7 +3447,7 @@ public final class Player implements
                 }
                 availableVideoStreams = sortedAvailableStreams;
                 selectedVideoStreamIndex = currentMetadata.getSelectedVideoStreamIndex();
-                updateQualityPlayedForVideos();
+                updateQualityTextViewFromSelectedVideoStream();
                 buildVideoQualityMenuIfNeeded();
                 binding.qualityTextView.setVisibility(View.VISIBLE);
                 binding.surfaceView.setVisibility(View.VISIBLE);
@@ -3545,10 +3546,9 @@ public final class Player implements
             return;
         }

-        int videoFormatsCount = 0;
-        for (int i = 0; i < videoTrackGroupArray.length; i++) {
-            videoFormatsCount += videoTrackGroupArray.get(i).length;
-        }
+        int videoFormatsCount = IntStream.range(0, videoTrackGroupArray.length)
+                .map(i -> videoTrackGroupArray.get(i).length)
+                .sum();

         final Menu menu = qualityPopupMenu.getMenu();
         if (videoFormatsCount + 1 == menu.size()) {
@@ -3562,7 +3562,7 @@ public final class Player implements
         int menuItemId = 1;
         for (int i = 0; i < videoTrackGroupArray.length; i++) {
             final TrackGroup videoTrackGroup = videoTrackGroupArray.get(i);
-                final int videoTrackGroupLength = videoTrackGroup.length;
+            final int videoTrackGroupLength = videoTrackGroup.length;
             for (int j = 0; j < videoTrackGroupLength; j++) {
                 final Format format = videoTrackGroup.getFormat(videoTrackGroupLength - 1 - j);
                 menu.add(POPUP_MENU_ID_QUALITY, menuItemId, Menu.NONE,
@@ -3583,68 +3583,73 @@ public final class Player implements
      * the automatic mode.
      */
     private void addAnalyticsListenerForLivestreamsIfNeeded() {
-        if (analyticsListener == null) {
-            analyticsListener = new AnalyticsListener() {
-                @Override
-                public void onVideoInputFormatChanged(
-                        @NonNull final EventTime eventTime,
-                        @NonNull final Format format,
-                        @Nullable final DecoderReuseEvaluation decoderReuseEvaluation) {
-                    final int videoRendererIndex = getVideoRendererIndex();
-                    if (videoRendererIndex != RENDERER_UNAVAILABLE) {
-                        final TrackGroupArray videoTrackGroupArray = Objects
-                                .requireNonNull(Objects.requireNonNull(trackSelector
-                                        .getCurrentMappedTrackInfo())).getTrackGroups(
-                                        videoRendererIndex);
-
-                        if (trackSelector.getParameters().hasSelectionOverride(
-                                videoRendererIndex, videoTrackGroupArray)) {
-                            binding.qualityTextView.setText(
-                                    getResolutionStringFromFormat(context, format));
-                        } else {
-                            binding.qualityTextView.setText(context.getString(
-                                    R.string.auto_quality_selected,
-                                    context.getString(R.string.auto_quality),
-                                    getResolutionStringFromFormat(context, format)));
-                        }
-                    }
-                }
-            };
-            simpleExoPlayer.addAnalyticsListener(analyticsListener);
+        if (analyticsListener != null) {
+            return;
         }
+
+        analyticsListener = new AnalyticsListener() {
+            @Override
+            public void onVideoInputFormatChanged(
+                    @NonNull final EventTime eventTime,
+                    @NonNull final Format format,
+                    @Nullable final DecoderReuseEvaluation decoderReuseEvaluation) {
+                final int videoRendererIndex = getVideoRendererIndex();
+                if (videoRendererIndex == RENDERER_UNAVAILABLE) {
+                    return;
+                }
+                final TrackGroupArray videoTrackGroupArray = trackSelector
+                        .getCurrentMappedTrackInfo()
+                        .getTrackGroups(videoRendererIndex);
+
+                final String resolution = getResolutionStringFromFormat(context, format);
+                if (trackSelector.getParameters()
+                        .hasSelectionOverride(videoRendererIndex, videoTrackGroupArray)) {
+                    binding.qualityTextView.setText(resolution);
+                } else {
+                    binding.qualityTextView.setText(context.getString(
+                            R.string.auto_quality_selected,
+                            context.getString(R.string.auto_quality),
+                            resolution));
+                }
+            }
+        };
+        simpleExoPlayer.addAnalyticsListener(analyticsListener);
     }

     /**
      * Update the quality played string, the quality popup menu and build the quality popup menu
      * for livestreams, if needed.
      *
-     * This method will call
-     * {@link #setDefaultResolutionForLivestreamsIfNeeded(TrackGroupArray, int)},
-     * {@link #buildLiveStreamQualityMenuIfNeeded(TrackGroupArray, int)}
-     * and {@link #addAnalyticsListenerForLivestreamsIfNeeded()} to complete this.
      */
     private void updateQualityElementsForLivestreams() {
-        if ((isAudioOnly ? audioResolver : videoResolver).isLiveSource()) {
-            final int videoRendererIndex = getVideoRendererIndex();
-            if (videoRendererIndex != RENDERER_UNAVAILABLE) {
-                final MappingTrackSelector.MappedTrackInfo currentMappedTrackInfo = trackSelector
-                        .getCurrentMappedTrackInfo();
-                final TrackGroupArray videoTrackGroupArray = Objects.requireNonNull(
-                        currentMappedTrackInfo).getTrackGroups(videoRendererIndex);
-
-                setDefaultResolutionForLivestreamsIfNeeded(videoTrackGroupArray,
-                        videoRendererIndex);
-                buildLiveStreamQualityMenuIfNeeded(videoTrackGroupArray, videoRendererIndex);
-                addAnalyticsListenerForLivestreamsIfNeeded();
-            }
+        if (!(isAudioOnly ? audioResolver : videoResolver).isLiveSource()) {
+            return;
         }
-    }

-    private void updateQualityPlayedForVideos() {
-        final VideoStream selectedVideoStream = getSelectedVideoStream();
-        if (selectedVideoStream != null) {
-            binding.qualityTextView.setText(selectedVideoStream.resolution);
+        final int videoRendererIndex = getVideoRendererIndex();
+        if (videoRendererIndex == RENDERER_UNAVAILABLE) {
+            return;
         }
+
+        final MappingTrackSelector.MappedTrackInfo currentMappedTrackInfo =
+                trackSelector.getCurrentMappedTrackInfo();
+        final TrackGroupArray videoTrackGroupArray =
+                Objects.requireNonNull(currentMappedTrackInfo).getTrackGroups(videoRendererIndex);
+
+        setDefaultResolutionForLivestreamsIfNeeded(videoTrackGroupArray, videoRendererIndex);
+        buildLiveStreamQualityMenuIfNeeded(videoTrackGroupArray, videoRendererIndex);
+        addAnalyticsListenerForLivestreamsIfNeeded();
+    }
+
+    private void updateQualityTextViewFromSelectedVideoStream() {
+        updateQualityTextViewFromSelectedVideoStream(VideoStream::getResolution);
+    }
+
+    private void updateQualityTextViewFromSelectedVideoStream(
+            @NonNull final Function<VideoStream,String> toTextConverter) {
+        getOptSelectedVideoStream()
+                .map(toTextConverter)
+                .ifPresent(binding.qualityTextView::setText);
     }

     /**
@@ -3675,27 +3680,13 @@ public final class Player implements
     public static String getResolutionStringFromFormat(
             @NonNull final Context context,
             @NonNull final Format format) {
-        final StringBuilder resolutionStringBuilder = new StringBuilder();
-        final int formatHeight = format.height;
-
-        if (formatHeight != Format.NO_VALUE) {
-            resolutionStringBuilder.append(formatHeight);
-            resolutionStringBuilder.append("p");
-            final int formatFrameRate = (int) format.frameRate;
-            if (formatFrameRate > 30) {
-                resolutionStringBuilder.append(formatFrameRate);
-            }
-        } else {
-            final int formatBitrate = format.bitrate;
-            if (formatBitrate != Format.NO_VALUE) {
-                resolutionStringBuilder.append(formatBitrate);
-                resolutionStringBuilder.append("bps");
-            } else {
-                resolutionStringBuilder.append(context.getString(R.string.unknown_quality));
-            }
+        if (format.height != Format.NO_VALUE) {
+            return format.height + "p" + ((int)format.frameRate > 30 ? format.frameRate : "");
         }
-
-        return resolutionStringBuilder.toString();
+        if (format.bitrate != Format.NO_VALUE) {
+            return format.bitrate + "bps";
+        }
+        return context.getString(R.string.unknown_quality));
     }

     /**
@@ -3705,16 +3696,18 @@ public final class Player implements
      * index available or if the current mapped track info is null.
      */
     private int getVideoRendererIndex() {
-        final MappingTrackSelector.MappedTrackInfo mappedTrackInfo = trackSelector
-                .getCurrentMappedTrackInfo();
-
-        if (mappedTrackInfo != null) {
-            for (int i = 0; i < mappedTrackInfo.getRendererCount(); i++) {
-                final TrackGroupArray trackGroups = mappedTrackInfo.getTrackGroups(i);
-                if (!trackGroups.isEmpty()
-                        && simpleExoPlayer.getRendererType(i) == C.TRACK_TYPE_VIDEO) {
-                    return i;
-                }
+        final MappingTrackSelector.MappedTrackInfo mappedTrackInfo =
+                trackSelector.getCurrentMappedTrackInfo();
+
+        if (mappedTrackInfo == null) {
+            return RENDERER_UNAVAILABLE;
+        }
+
+        for (int i = 0; i < mappedTrackInfo.getRendererCount(); i++) {
+            final TrackGroupArray trackGroups = mappedTrackInfo.getTrackGroups(i);
+            if (!trackGroups.isEmpty()
+                    && simpleExoPlayer.getRendererType(i) == C.TRACK_TYPE_VIDEO) {
+                return i;
             }
         }

@@ -3848,10 +3841,7 @@ public final class Player implements
         isSomePopupMenuVisible = false; //TODO check if this works

         if (!(isAudioOnly ? audioResolver : videoResolver).isLiveSource()) {
-            final VideoStream videoStream = getSelectedVideoStream();
-            if (videoStream != null) {
-                binding.qualityTextView.setText(videoStream.resolution);
-            }
+            updateQualityTextViewFromSelectedVideoStream();
         }

         if (isPlaying()) {
@@ -3868,12 +3858,9 @@ public final class Player implements
         isSomePopupMenuVisible = true;

         if (!(isAudioOnly ? audioResolver : videoResolver).isLiveSource()) {
-            final VideoStream videoStream = getSelectedVideoStream();
-            if (videoStream != null) {
-                final String qualityText = MediaFormat.getNameById(videoStream.getFormatId()) + " "
-                        + videoStream.resolution;
-                binding.qualityTextView.setText(qualityText);
-            }
+            updateQualityTextViewFromSelectedVideoStream(videoStream ->
+                    MediaFormat.getNameById(videoStream.getFormatId()) + " "
+                    + videoStream.resolution);
         }

         saveWasPlaying();
@@ -3936,8 +3923,9 @@ public final class Player implements
             return;
         }

-        final int textRenderer = getCaptionRendererIndex();
         captionPopupMenu.getMenu().clear();
+
+        final int textRenderer = getCaptionRendererIndex();
         if (trackSelector.getCurrentMappedTrackInfo() == null
                 || textRenderer == RENDERER_UNAVAILABLE) {
             binding.captionTextView.setVisibility(View.GONE);
@@ -3948,19 +3936,15 @@ public final class Player implements
                 .getTrackGroups(textRenderer);

         // Extract all loaded languages
-        final List<String> availableLanguages = new ArrayList<>(textTracks.length);
-        for (int i = 0; i < textTracks.length; i++) {
-            final TrackGroup textTrack = textTracks.get(i);
-            if (textTrack.length > 0) {
-                final String language = textTrack.getFormat(0).language;
+        final List<String> availableLanguages = IntStream.range(0, textTracks.length)
+                .mapToObj(textTracks::get)
+                .filter(textTrack -> textTrack.length > 0)
+                .map(textTrack -> textTrack.getFormat(0).language)
                 // We are not able to select the preferred caption language when it's null
                 // TODO: use selection override for captions instead of the preferred text language
                 //  so we can enable the selection of unknown languages
-                if (language != null) {
-                    availableLanguages.add(language);
-                }
-            }
-        }
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());

         if (availableLanguages.isEmpty()) {
             binding.captionTextView.setVisibility(View.GONE);
```

</details>

I also asked myself if the Exoplayer developers are really familiar with Java...
There are so many indices and arrays where objects or lists could be...

TiA4f8R:
To be honest, some of your "optimizations" are not really optimizations and some of them complexify the code reading: I could not understand most of what you wrote, especially the list changes part (but that's something I didn't learn, at least for now).

So I don't think these changes (at least the biggest part) should be applied.

opusforlife2:
If they're optimisations, they can be deferred to another PR.

------------

@TiA4f8R Have you seen this exception with this PR? : `com.google.android.exoplayer2.ExoPlaybackException: Source error`

TiA4f8R:
Could you provide the detailed error message, @opusforlife2? The base error doesn't help me at all.

opusforlife2:
Sure. Just didn't want to spam if the error was known.

----

## Exception
* __User Action:__ play stream
* __Request:__ Player error[type=0] occurred while playing https://www.youtube.com/watch?v=Lgqh9aKl2mg
* __Content Language:__ en-IN
* __Service:__ YouTube
* __Version:__ 0.21.14


```
com.google.android.exoplayer2.ExoPlaybackException: Source error
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:580)
	at android.os.Handler.dispatchMessage(Handler.java:103)
	at android.os.HandlerThread.run(HandlerThread.java:67)
Caused by: com.google.android.exoplayer2.source.BehindLiveWindowException
	at com.google.android.exoplayer2.source.hls.HlsChunkSource.getNextChunk(HlsChunkSource.java:393)
	at com.google.android.exoplayer2.source.hls.HlsSampleStreamWrapper.continueLoading(HlsSampleStreamWrapper.java:708)
	at com.google.android.exoplayer2.source.CompositeSequenceableLoader.continueLoading(CompositeSequenceableLoader.java:78)
	at com.google.android.exoplayer2.source.hls.HlsMediaPeriod.continueLoading(HlsMediaPeriod.java:371)
	at com.google.android.exoplayer2.source.MaskingMediaPeriod.continueLoading(MaskingMediaPeriod.java:226)
	at com.google.android.exoplayer2.source.MaskingMediaPeriod.continueLoading(MaskingMediaPeriod.java:226)
	at com.google.android.exoplayer2.MediaPeriodHolder.continueLoading(MediaPeriodHolder.java:221)
	at com.google.android.exoplayer2.ExoPlayerImplInternal.maybeContinueLoading(ExoPlayerImplInternal.java:2222)
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleContinueLoadingRequested(ExoPlayerImplInternal.java:2185)
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:500)
	... 3 more

```
</p></details>
<hr>

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3262,11 +3306,83 @@ public void disablePreloadingOfCurrentTrack() {
...
+                final Format videoFormat = trackGroup.getFormat(j);
+                if ((defaultHeight == -1 || videoFormat.height <= defaultHeight)
+                        && (defaultFrameRate <= -1 || videoFormat.frameRate <= defaultFrameRate)) {
+                    builder.setSelectionOverride(videoRendererIndex, videoTrackGroupArray,
+                            new DefaultTrackSelector.SelectionOverride(videoRendererIndex, j));
+                    isDefaultResolutionSetForLivestreams = true;
+                    break;
+                }
+            }
+        }
```
Stypox, line 3367: From what I can understand, this `for` overrides the selection for each possible `trackGroup` on the first possible `videoFormat` that satisfies `(defaultHeight == -1 || videoFormat.height <= defaultHeight) && (defaultFrameRate <= -1 || videoFormat.frameRate <= defaultFrameRate)`. A couple of questions:
- why `defaultFrameRate <= -1` and not `defaultFrameRate == -1` for consistency?
- why stop at the first suitable track? Maybe there are more similar tracks next. Wouldn't this always choose 144p?
- (question about exoplayer) why are there multiple formats for each track? Shouldn't each track have exactly one format? Otherwise it would be split into multiple tracks... I don't understand exoplayer...

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3262,11 +3306,83 @@ public void disablePreloadingOfCurrentTrack() {
...
+
+        if (!isDefaultResolutionSetForLivestreams) {
+            // No resolution for the override could not be found, set the maximum video height and
+            // frame rate according to the default quality chose by the user in Video and Audio
+            // settings
+            builder.setMaxVideoSize(Integer.MAX_VALUE, defaultHeight);
+            builder.setMaxVideoFrameRate(defaultFrameRate != -1 ? defaultFrameRate : 30);
```
Stypox, line 3374:
```suggestion
            builder.setMaxVideoFrameRate(defaultFrameRate == -1 ? 30 : defaultFrameRate);
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3355,22 +3497,232 @@ private void updateQueueTime(final int currentTime) {
...
+        }
+
+        menu.removeGroup(POPUP_MENU_ID_QUALITY);
+        menu.add(POPUP_MENU_ID_QUALITY, 0, Menu.NONE,
+                context.getString(R.string.auto_quality));
+
+        int menuItemId = 1;
+        for (int i = 0; i < videoTrackGroupArray.length; i++) {
+            final TrackGroup videoTrackGroup = videoTrackGroupArray.get(i);
+                final int videoTrackGroupLength = videoTrackGroup.length;
```
Stypox, line 3568: Do not create this variable, just use `videoTrackGroup.length` each time
```suggestion
```

Stypox (changes requested):
Almost ready to merge in my opinion, just a couple of questions

app/src/main/java/org/schabi/newpipe/player/Player.java
TiA4f8R: Why? Is not better to store the length instead of getting it at each loop tour?

(I fixed the indentation by the way.)

Stypox: You are not "getting" it. You are reading it directly. If you were calling a `getLength()` method then yes, it would be slightly faster to store it (though probably an unneeded optimization). But since there is no method call but just reading from a class field, the performance should be exactly the same.

app/src/main/java/org/schabi/newpipe/player/Player.java
Stypox: Any news @TiA4f8R ?

app/src/main/java/org/schabi/newpipe/player/Player.java
TiA4f8R: Ok, no problem. I will change this.

TiA4f8R:
@karyogamy It seems there are better ways to do quality selection when initializing the playback, by using a CustomTrackSelector.

Do you have some time to take a look at this feature and if so, to implement it? It seems to be quite complex and I don't have the time to take a deeper look on this type of things.

Also, the current way used in this PR to select/override tracks is now deprecated (and maybe we should use setMaxVideoSize and setMinVideoSize but if no video track matches the constraints, no video will be played at all).

karyogamy:
> It seems there are better ways to do quality selection when initializing the playback, by using a CustomTrackSelector.
> It seems to be quite complex 

Could you elaborate on what you mean by better ways? I'm not too familiar with TrackSelector but complexity is exactly the reason why, during latest ExoPlayer upgrade, I chose to go back and use `DefaultTrackSelector` instead of whipping out another hacky implementation of `CustomTrackSelector`. Don't really have the time to work on this now unfortunately, but I'd avoid extending the default track selector if possible, unless there is a real need to do so.

> the current way used in this PR to select/override tracks is now deprecated 

Exo devs will eventually have to update [their demo track selector](https://github.com/google/ExoPlayer/blob/release-v2/demos/main/src/main/java/com/google/android/exoplayer2/demo/TrackSelectionDialog.java) that we can use as reference.

Stypox:
@TiA4f8R ping for https://github.com/TeamNewPipe/NewPipe/pull/7260#issuecomment-1107659048 and https://github.com/TeamNewPipe/NewPipe/pull/7260#discussion_r808523567 (but if I understand this PR will have to wait for changes on exoplayer's end)

pc00per:
What's the issue for this PR?

AudricV:
Closing this PR, as it is quite old and its code is now using deprecated features in ExoPlayer. It also seems very hard to rebase it on top of the `dev` branch.

I found easier ways to do parts of the feature, starting again almost from scratch. I will open a new PR including the addition of the feature and some fixes/improvements in the player (mostly related to the `SourceType` of sources, which are needed to do properly this feature) soon.

PR #7341, opened 2021-11-01
Fix player controls not hiding if resumed from media button

### LINKED ISSUE #3403: The player controls don't hide when using external controls

### LINKED ISSUE #7340: Play Button does not hide if playback is resumed externally (keyboard or bluetooth earpiece)
### Checklist

1. Start playing a video
2. Pause the video (doesn't matter how it's done, can be with the pause button or externally)
3. Use either a keyboard's space button or a bluetooth earpiece to resume the playback.

### Actual behavior

The play button stays visible.

### Expected behavior

The play button should hide, like when you tap the button manually.

### Device info
Not applicable, happens on all the devices I've tested.

ktprograms:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Added `hideControls(0, 0)` to the Space key pressed and the `PlayerMediaSession` `play` method (only if the player is currently playing)

| Before | After |
| -- | -- |
| <video src="https://user-images.githubusercontent.com/71804605/139620899-b62f3eb7-1d07-44d3-959e-104d46e9fbea.mp4"/> | <video src="https://user-images.githubusercontent.com/71804605/139621120-d73b284d-de60-4eb7-af34-eaae7df747d6.mp4"/> |

- ~~Fixes #7340~~ (I made that issue then only looked at how to fix it)
- Fixes #3403

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -3752,9 +3752,12 @@ public final class Player implements
             default:
                 break;
             case KeyEvent.KEYCODE_SPACE:
-                playPause();
-                if (isPlaying()) {
-                    hideControls(0, 0);
+                if (isFullscreen) {
+                    playPause();
+                    if (isPlaying()) {
+                        hideControls(0, 0);
+                    }
+                    return true;
                 }
                 break;
             case KeyEvent.KEYCODE_BACK:
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -3754,6 +3754,9 @@ public final class Player implements
             case KeyEvent.KEYCODE_SPACE:
                 if (isFullscreen) {
                     playPause();
+                    if (isPlaying()) {
+                        hideControls(0, 0);
+                    }
                 }
                 break;
             case KeyEvent.KEYCODE_BACK:
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -3752,11 +3752,9 @@ public final class Player implements
             default:
                 break;
             case KeyEvent.KEYCODE_SPACE:
-                if (isFullscreen) {
-                    playPause();
-                    if (isPlaying()) {
-                        hideControls(0, 0);
-                    }
+                playPause();
+                if (isPlaying()) {
+                    hideControls(0, 0);
                 }
                 break;
             case KeyEvent.KEYCODE_BACK:
--- a/app/src/main/java/org/schabi/newpipe/player/playback/PlayerMediaSession.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playback/PlayerMediaSession.java
@@ -88,6 +88,7 @@ public class PlayerMediaSession implements MediaSessionCallback {
     @Override
     public void play() {
         player.play();
+        player.hideControls(0, 0);
     }
 
     @Override
--- a/app/src/main/java/org/schabi/newpipe/player/playback/PlayerMediaSession.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playback/PlayerMediaSession.java
@@ -88,6 +88,7 @@ public class PlayerMediaSession implements MediaSessionCallback {
     @Override
     public void play() {
         player.play();
+        // hide the player controls even if the play command came from the media session
         player.hideControls(0, 0);
     }
 


app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3754,6 +3754,9 @@ public boolean onKeyDown(final int keyCode) {
             case KeyEvent.KEYCODE_SPACE:
                 if (isFullscreen) {
                     playPause();
+                    if (isPlaying()) {
```
litetex, line 3757: Why is this needed?

ktprograms: So that the behaviour is consistent with manually pausing (keeps the controls visible).

TobiGr (approved):

app/src/main/java/org/schabi/newpipe/player/Player.java
litetex: Okay but why is this only the case for the full-screen player?

ktprograms: The `if (isFullscreen)` was already there so I just left it. But I've tested and it works fine even when not in fullscreen so I've removed the `if (isFullscreen)` in https://github.com/TeamNewPipe/NewPipe/pull/7341/commits/29348411520f56f07865c6deef722ae107b9a205

ktprograms:
@litetex Can you please review?

litetex:
@ktprograms 

I'm currently trying to understand if this is the correct place for ``hideControls``.
<details><summary>But the code who is calling this is <i>a bit</i> complicated:</summary>

grafik
(It also pretty annoying that there are zero comments and the code got rewritten in like 10 commits)

</details>

Because when e.g. ``Player#onPaused`` is called then there is always a call for ``showControls(400)`` executed.
However this is not the case in ``Player#onPlaying`` but I don't know why.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3752,8 +3752,9 @@ public boolean onKeyDown(final int keyCode) {
             default:
                 break;
             case KeyEvent.KEYCODE_SPACE:
-                if (isFullscreen) {
```
ImDeadGuy, line 3755: Not sure why this check existed here in the first place, what can/should happen when pressing Space outside of full screen? (typing in the search or comment box?)
If in doubt, could just leave the original if() check, and just add the if (isPlaying) part.

ImDeadGuy (approved):
Honestly, this bug has been _bugging_ me for months (almost always use BT headphones with the tablet slightly out of reach).
Just want this fix in sooner than later, and I've tested the custom APK in Checks and it works as expected. 👍

Bugginme:
@ImDeadGuy same :/

app/src/main/java/org/schabi/newpipe/player/Player.java
Stypox: See #5331. In that PR there doesn't seem to be any explanation about why the `if (isFullscreen)`. So I think we can safely remove it, what do you think @mbarashkov?

mbarashkov: @Stypox In that PR, I had to do this because not in full screen space key was used on TV's to navigate the screen vertically, as I recall.

Stypox: Ok, that makes sense, thank you. Then I'll restore the previous `if (isFullscreen)`.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3752,8 +3752,9 @@ public boolean onKeyDown(final int keyCode) {
...
-                    playPause();
+                playPause();
+                if (isPlaying()) {
+                    hideControls(0, 0);
                 }
```
Stypox, line 3758:
```suggestion
                if (isFullscreen) {
                    playPause();
                    if (isPlaying()) {
                        hideControls(0, 0);
                    }
                }
```

app/src/main/java/org/schabi/newpipe/player/playback/PlayerMediaSession.java
```diff
@@ -88,6 +88,7 @@ public MediaDescriptionCompat getQueueMetadata(final int index) {
     @Override
     public void play() {
         player.play();
+        player.hideControls(0, 0);
```
Stypox, line 91:
```suggestion
        // hide the player controls even if the play command came from the media session
        player.hideControls(0, 0);
```

Stypox (changes requested):

Stypox:
@litetex what do you think about my suggestions? If they look good, commit them and then merge and squash (after 0.22.1)

litetex:
@Stypox 
I think you can try to add them 😄

ktprograms:
@litetex What do you mean?

Should I commit/squash the suggestions?

Stypox (approved):

ImDeadGuy:
Hey guys, is this meant to be fixed now in release 23?

Because it's definitely not. Resuming from the BT headset play/pause button still keeps the UI on screen forever.

TiA4f8R:
It was meant to be fixed, yes, but the latest ExoPlayer update broke the original fix, and a new fix has been provided in the PR linked above: #8315.

Issue #6770, opened 2021-07-26
Can't interact with app/screen below popup player on Android 12


Joristdh:
### Checklist

- [x] I am using the latest version - 0.21.7 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

0. This is a bug on Android 12 beta 3
1. Open a video in the popup player (it doesn't matter which video or how you open it)
2. Make sure the popup uses the full width of the screen (when in portrait)
3. Open another app below the popup player, go to the home screen or open recent apps
4. Try to interact with the screen below the popup

### Actual behavior

All touches to the app/screen below the popup are ignored, because of the following behavior change in Android 12: https://developer.android.com/about/versions/12/behavior-changes-all#untrusted-touch-events 

### Expected behavior

The app below the popup responds to the touches.

### Workaround
For now this issues can be solved per device by running the following ADB command:
`adb shell settings put global block_untrusted_touches 0`

### Device info

 - Android version/Custom ROM version: Android 12 beta 3
 - Device model: Pixel 3

chrstfer:
Im seeing this sporadically on android 11 (samsung gs20 ultra with up to date stock firmware) with most recent release (0.21.7).  Often the app beneath will respond to drags and swipes but no controls will send their events even though they animate.

chrstfer:
What needs to be done to fix this? Its real annoying to need to clear my queue (because of an unrelated issue) just to send a tweet.

Joristdh:
@chrstfer I don't think it is the same bug. The behavior change is introduced in Android 12, as you can see in the link I provided in the description. If you encounter a similar bug in Android 11 it's probably caused by something else.
Maybe Samsung implemented a similar restriction in One UI or those apps implemented some restrictions themselves (for while another app overlays them). 

If your bug does have the same cause, the ADB shell command in the description should fix it. If it doesn't, or you can't execute the command at all (because you aren't on Android 12), you should probably file a new bug report detailing your specific issue.

hackel:
I also (still) have this issue on Android 12 beta 4.1 and the width of the pop-up makes no difference.  I believe I read that the actual PIP more will need to be used going forward for security.

[...]

Issue #4497, opened 2020-10-11
Not the same playback position when changing players


B0pol:
### Checklist

- [x] I am using the latest version. 
- [x] I checked, but didn't find any duplicates of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

1. Play the video in video details / full screen for some time (few tens of seconds)
2. Switch to background by turning off your screen or clicking on popup button

It does not _always_ happen, I don't know why

### Actual behaviour

The playback position is current minus about 20s

### Expected behavior

Same playback position when changing players

### Device info

 - Android version/Custom ROM version: Android 9
 - Device model: pocophone F1

opusforlife2:
For me, Newpipe has always resumed a little behind where I pause it when I switch to other apps. I thought that was a feature, trying to maintain continuity. But only something like 5-10 seconds.

If that is indeed how the code works, maybe there is a bug in it?

MD77MD:
also but not so related.... if you play video A, then play video B, then go back to video A... video A will start from the beginning

I can open a new issue if my comment here is not sufficient.

opusforlife2:
@MD77MD Do you have Resume turned off?

MD77MD:
@opusforlife2 

😲......Yub 🤣, but i have no memory off turning it off 🤑 lol... thanks bro

opusforlife2:
Wow, you're right. It doesn't work. @avently some sort of regression has happened here. We've tested this. It used to work properly.

@MD77MD Please open a new issue.

MD77MD:
#4506 playback position from video B to video A bug

avently:
@ildar I never had this problem. So nothing to fix until someone provides 100% reproducible procedure

SameenAhnaf:
This issue should be closed as https://github.com/TeamNewPipe/NewPipe/issues/7427 has a more updated discussion.

PR #7349, opened 2021-11-02
Add seamless transition between background and video players when putting the app in background (for video-only streams and audio-only streams only)

### LINKED ISSUE #5016: Unexpected pauses when switching player type
### Checklist

- [x] I am using the latest version - 0.20.4 

1. Start some video in NewPipe
2. Minimize NewPipe by pressing home or opening Android recent applications drawer
3. Enter NewPipe again

### Actual behaviour

When I minimize the app, the audio stops playing for like 100-600ms and then resumes. It's pretty annoying.
When I reenter the app, the same thing happens.

### Expected behavior

No pauses in the audio are expected.

[Recording on youtube](https://youtu.be/Y0ihnKFnIRk)

### Device info

 - Android version/Custom ROM version: latest Android 10
 - Device model: Asus ZenFone 5Z

TiA4f8R:
#### What is it?
- [x] Feature (user facing)

This PR makes the app do a seamless transition between main or popup player and background player when putting the app in background (for e.g. when turning off your device or switching to recent apps).

This will only happen **when you are playing video-only streams or audio-only streams** (when there are audio streams available and only video streams with audio, the pause when backgrounding the app will still happen, in order to reduce data usage; in the case when there are only video streams with audio and no audio streams, no pause will happen).

This **doesn't affect the pause which occur when pressing the `Background` or the `Popup` button** (the current player type is destroyed and a new player type is created).

A black video is shown for a few seconds after you come back to the player. It is **intended**, because the video is not fetched when you are in background, to prevent you to spend more data than needed to play the audio. This is the behavior of the YouTube app on Android (for Premium clients only of course).

Note that **livestreams will still fetch the video stream** (because streams we are getting right now are not demuxed (this will probably change soon) and there is an ExoPlayer bug which still fetches video even if it has been disabled with the track selector for HLS contents (the delivery method we can only use for livestreams right now) (a low priority has been given by the ExoPlayer team to this bug)).

This PR makes also the player use video-only streams first (instead of video streams with audio), when possible.

- Fixes #5016

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -178,6 +178,7 @@ import org.schabi.newpipe.player.playqueue.PlayQueueItemTouchCallback;
 import org.schabi.newpipe.player.resolver.AudioPlaybackResolver;
 import org.schabi.newpipe.player.resolver.MediaSourceTag;
 import org.schabi.newpipe.player.resolver.VideoPlaybackResolver;
+import org.schabi.newpipe.player.resolver.VideoPlaybackResolver.SourceType;
 import org.schabi.newpipe.player.seekbarpreview.SeekbarPreviewThumbnailHelper;
 import org.schabi.newpipe.player.seekbarpreview.SeekbarPreviewThumbnailHolder;
 import org.schabi.newpipe.util.DeviceUtils;
@@ -3293,8 +3294,9 @@ public final class Player implements
         if (audioPlayerSelected()) {
             return audioResolver.resolve(info);
         } else {
-            if (isAudioOnly
-                    && !videoResolver.wasLastResolvedVideoAndAudioSeparated().orElse(false)) {
+            if (isAudioOnly && videoResolver.getStreamSourceType().orElse(
+                            SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY)
+                    == SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY) {
                 // If the current info has only video streams with audio and if the stream is
                 // played as audio, we need to use the audio resolver, otherwise the video stream
                 // will be played in background.
@@ -4196,18 +4198,30 @@ public final class Player implements
         stream will be fetched and the video stream will be fetched again when the user return to a
         video player.
 
-        For audio streams: nothing is done, it's not needed to reload the player with the same
-        audio stream.
+        For audio streams and audio live streams: nothing is done, it's not needed to reload the
+        player with the same audio stream.
+
+        For video live streams: the play queue manager is not reloaded if the stream source is a
+        live source (see VideoPlaybackResolver#resolve()) and if that's not the case, the
+        requirements for video streams is applied.
 
         In the case where we don't know the index of the video renderer, the play queue manager
         is also reloaded. */
 
         final StreamType streamType = info.getStreamType();
+        final SourceType sourceType = videoResolver.getStreamSourceType()
+                .orElse(SourceType.VIDEO_WITH_SEPARATED_AUDIO);
 
+        final boolean isVideoWithSeparatedAudioOrVideoWithNoSeparatedAudioStreams =
+                sourceType == SourceType.VIDEO_WITH_SEPARATED_AUDIO
+                || (sourceType == SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY
+                        && isNullOrEmpty(info.getAudioStreams()));
         final boolean isVideoStreamTypeAndIsVideoOnlyStreamOrNoAudioStreamsAvailable =
-                (streamType == StreamType.VIDEO_STREAM || streamType == StreamType.LIVE_STREAM)
-                        && (videoResolver.wasLastResolvedVideoAndAudioSeparated().orElse(false)
-                            || isNullOrEmpty(info.getAudioStreams()));
+                streamType == StreamType.VIDEO_STREAM
+                        && isVideoWithSeparatedAudioOrVideoWithNoSeparatedAudioStreams
+                        || (streamType == StreamType.LIVE_STREAM
+                            && (sourceType == SourceType.LIVE_STREAM
+                                || isVideoWithSeparatedAudioOrVideoWithNoSeparatedAudioStreams));
 
         if (videoRenderIndex != RENDERER_UNAVAILABLE
                 && isVideoStreamTypeAndIsVideoOnlyStreamOrNoAudioStreamsAvailable) {
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -3294,24 +3294,25 @@ public final class Player implements
     public MediaSource sourceOf(final PlayQueueItem item, final StreamInfo info) {
         if (audioPlayerSelected()) {
             return audioResolver.resolve(info);
-        } else {
-            if (isAudioOnly && videoResolver.getStreamSourceType().orElse(
-                            SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY)
-                    == SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY) {
-                // If the current info has only video streams with audio and if the stream is
-                // played as audio, we need to use the audio resolver, otherwise the video stream
-                // will be played in background.
-                return audioResolver.resolve(info);
-            }
-
-            // Even if the stream is played in background, we need to use the video resolver if the
-            // info played is separated video-only and audio-only streams; otherwise, if the audio
-            // resolver was called when the app was in background, the app will only stream audio
-            // when the user come back to the app and will never fetch the video stream.
-            // Note that the video is not fetched when the app is in background because the video
-            // renderer is fully disabled (see useVideoSource method).
-            return videoResolver.resolve(info);
         }
+
+        if (isAudioOnly && videoResolver.getStreamSourceType().orElse(
+                SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY)
+                == SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY) {
+            // If the current info has only video streams with audio and if the stream is played as
+            // audio, we need to use the audio resolver, otherwise the video stream will be played
+            // in background.
+            return audioResolver.resolve(info);
+        }
+
+        // Even if the stream is played in background, we need to use the video resolver if the
+        // info played is separated video-only and audio-only streams; otherwise, if the audio
+        // resolver was called when the app was in background, the app will only stream audio when
+        // the user come back to the app and will never fetch the video stream.
+        // Note that the video is not fetched when the app is in background because the video
+        // renderer is fully disabled (see useVideoSource method), except for HLS streams
+        // (see https://github.com/google/ExoPlayer/issues/9282).
+        return videoResolver.resolve(info);
     }
 
     public void disablePreloadingOfCurrentTrack() {
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -4177,9 +4177,15 @@ public final class Player implements
 
         final int videoRenderIndex = getVideoRendererIndex();
 
-        // We can safely assume that currentMetadata is not null (otherwise this method isn't
-        // called) so we can use the requireNonNull method of the Objects class.
-        final StreamInfo info = Objects.requireNonNull(currentMetadata).getMetadata();
+        // The current metadata may be null sometimes so we will be not able to execute the
+        // block above. Reload the play queue manager in this case.
+        if (currentMetadata == null) {
+            reloadPlayQueueManager();
+            setRecovery();
+            return;
+        }
+
+        final StreamInfo info = currentMetadata.getMetadata();
 
         /* For video streams: we don't want to stream in background the video stream so if the
         video stream played is not a video-only stream and if there is an audio stream available,
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -4175,16 +4175,17 @@ public final class Player implements
             hideSystemUIIfNeeded();
         }
 
-        final int videoRenderIndex = getVideoRendererIndex();
-
-        // The current metadata may be null sometimes so we will be not able to execute the
-        // block above. Reload the play queue manager in this case.
+        // The current metadata may be null sometimes (for e.g. when using an unstable connection
+        // in livestreams) so we will be not able to execute the block below.
+        // Reload the play queue manager in this case, which is the behavior when the video stream
+        // played is not video only or when we don't know the index of the video renderer.
         if (currentMetadata == null) {
             reloadPlayQueueManager();
             setRecovery();
             return;
         }
 
+        final int videoRenderIndex = getVideoRendererIndex();
         final StreamInfo info = currentMetadata.getMetadata();
 
         /* For video streams: we don't want to stream in background the video stream so if the
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -3003,25 +3003,18 @@ public final class Player implements
         final MediaSourceTag metadata;
         try {
             final MediaItem currentMediaItem = simpleExoPlayer.getCurrentMediaItem();
-            if (currentMediaItem != null) {
-                final MediaItem.PlaybackProperties playbackProperties =
-                        currentMediaItem.playbackProperties;
-                metadata = (MediaSourceTag) (playbackProperties != null ? playbackProperties.tag
-                        : null);
-            } else {
-                metadata = null;
+            if (currentMediaItem == null || currentMediaItem.playbackProperties == null
+                    || currentMediaItem.playbackProperties.tag == null) {
+                return;
             }
-        } catch (final IndexOutOfBoundsException | ClassCastException error) {
+            metadata = (MediaSourceTag) currentMediaItem.playbackProperties.tag;
+        } catch (final IndexOutOfBoundsException | ClassCastException ex) {
             if (DEBUG) {
-                Log.d(TAG, "Could not update metadata: " + error.getMessage());
-                error.printStackTrace();
+                Log.d(TAG, "Could not update metadata", ex);
             }
             return;
         }
 
-        if (metadata == null) {
-            return;
-        }
         maybeAutoQueueNextStream(metadata);
 
         if (currentMetadata == metadata) {
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -3293,7 +3293,8 @@ public final class Player implements
         if (audioPlayerSelected()) {
             return audioResolver.resolve(info);
         } else {
-            if (isAudioOnly && !videoResolver.isVideoStreamVideoOnly()) {
+            if (isAudioOnly
+                    && !videoResolver.wasLastResolvedVideoAndAudioSeparated().orElse(false)) {
                 return audioResolver.resolve(info);
             }
 
@@ -4189,8 +4190,9 @@ public final class Player implements
 
         final boolean isVideoStreamTypeAndIsVideoOnlyStreamOrNoAudioStreamsAvailable =
                 (streamType == StreamType.VIDEO_STREAM || streamType == StreamType.LIVE_STREAM)
-                        && (videoResolver.isVideoStreamVideoOnly()
+                        && (videoResolver.wasLastResolvedVideoAndAudioSeparated().orElse(false)
                             || isNullOrEmpty(info.getAudioStreams()));
+
         if (videoRenderIndex != RENDERER_UNAVAILABLE
                 && isVideoStreamTypeAndIsVideoOnlyStreamOrNoAudioStreamsAvailable) {
             final TrackGroupArray videoTrackGroupArray = Objects.requireNonNull(
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -3295,9 +3295,18 @@ public final class Player implements
         } else {
             if (isAudioOnly
                     && !videoResolver.wasLastResolvedVideoAndAudioSeparated().orElse(false)) {
+                // If the current info has only video streams with audio and if the stream is
+                // played as audio, we need to use the audio resolver, otherwise the video stream
+                // will be played in background.
                 return audioResolver.resolve(info);
             }
 
+            // Even if the stream is played in background, we need to use the video resolver if the
+            // info played is separated video-only and audio-only streams; otherwise, if the audio
+            // resolver was called when the app was in background, the app will only stream audio
+            // when the user come back to the app and will never fetch the video stream.
+            // Note that the video is not fetched when the app is in background because the video
+            // renderer is fully disabled (see useVideoSource method).
             return videoResolver.resolve(info);
         }
     }
@@ -4160,8 +4169,8 @@ public final class Player implements
         }
 
         isAudioOnly = !videoEnabled;
-        // When a user returns from background controls could be hidden
-        // but systemUI will be shown 100%. Hide it
+        // When a user returns from background, controls could be hidden but SystemUI will be shown
+        // 100%. Hide it.
         if (!isAudioOnly && !isControlsVisible()) {
             hideSystemUIIfNeeded();
         }
@@ -4198,9 +4207,13 @@ public final class Player implements
             final TrackGroupArray videoTrackGroupArray = Objects.requireNonNull(
                     trackSelector.getCurrentMappedTrackInfo()).getTrackGroups(videoRenderIndex);
             if (videoEnabled) {
+                // Clearing the null selection override enable again the video stream (and its
+                // fetching).
                 trackSelector.setParameters(trackSelector.buildUponParameters()
                         .clearSelectionOverride(videoRenderIndex, videoTrackGroupArray));
             } else {
+                // Using setRendererDisabled still fetch the video stream in background, contrary
+                // to setSelectionOverride with a null override.
                 trackSelector.setParameters(trackSelector.buildUponParameters()
                         .setSelectionOverride(videoRenderIndex, videoTrackGroupArray, null));
             }
@@ -4436,6 +4449,15 @@ public final class Player implements
     }
     //endregion
 
+    /**
+     * Get the video renderer index of the current playing stream.
+     *
+     * This method returns the video renderer index of the current
+     * {@link MappingTrackSelector.MappedTrackInfo} or {@link #RENDERER_UNAVAILABLE} if the current
+     * {@link MappingTrackSelector.MappedTrackInfo} is null or if there is no video renderer index.
+     *
+     * @return the video renderer index or {@link #RENDERER_UNAVAILABLE} if it cannot be get
+     */
     private int getVideoRendererIndex() {
         final MappingTrackSelector.MappedTrackInfo mappedTrackInfo = trackSelector
                 .getCurrentMappedTrackInfo();
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -112,6 +112,7 @@ import androidx.recyclerview.widget.RecyclerView;
 import com.google.android.exoplayer2.C;
 import com.google.android.exoplayer2.DefaultRenderersFactory;
 import com.google.android.exoplayer2.ExoPlaybackException;
+import com.google.android.exoplayer2.MediaItem;
 import com.google.android.exoplayer2.PlaybackParameters;
 import com.google.android.exoplayer2.Player.PositionInfo;
 import com.google.android.exoplayer2.RenderersFactory;
@@ -122,6 +123,7 @@ import com.google.android.exoplayer2.source.MediaSource;
 import com.google.android.exoplayer2.source.TrackGroup;
 import com.google.android.exoplayer2.source.TrackGroupArray;
 import com.google.android.exoplayer2.text.Cue;
+import com.google.android.exoplayer2.trackselection.MappingTrackSelector;
 import com.google.android.exoplayer2.trackselection.TrackSelectionArray;
 import com.google.android.exoplayer2.ui.AspectRatioFrameLayout;
 import com.google.android.exoplayer2.ui.CaptionStyleCompat;
@@ -144,6 +146,7 @@ import org.schabi.newpipe.error.UserAction;
 import org.schabi.newpipe.extractor.MediaFormat;
 import org.schabi.newpipe.extractor.stream.StreamInfo;
 import org.schabi.newpipe.extractor.stream.StreamSegment;
+import org.schabi.newpipe.extractor.stream.StreamType;
 import org.schabi.newpipe.extractor.stream.VideoStream;
 import org.schabi.newpipe.fragments.OnScrollBelowItemsListener;
 import org.schabi.newpipe.fragments.detail.VideoDetailFragment;
@@ -2443,9 +2446,9 @@ public final class Player implements
     }
 
     @Override
-    public void onPositionDiscontinuity(
-            final PositionInfo oldPosition, final PositionInfo newPosition,
-            @DiscontinuityReason final int discontinuityReason) {
+    public void onPositionDiscontinuity(@NonNull final PositionInfo oldPosition,
+                                        @NonNull final PositionInfo newPosition,
+                                        @DiscontinuityReason final int discontinuityReason) {
         if (DEBUG) {
             Log.d(TAG, "ExoPlayer - onPositionDiscontinuity() called with "
                     + "discontinuityReason = [" + discontinuityReason + "]");
@@ -2493,7 +2496,7 @@ public final class Player implements
     }
 
     @Override
-    public void onCues(final List<Cue> cues) {
+    public void onCues(@NonNull final List<Cue> cues) {
         binding.subtitleView.onCues(cues);
     }
     //endregion
@@ -2999,8 +3002,16 @@ public final class Player implements
 
         final MediaSourceTag metadata;
         try {
-            metadata = (MediaSourceTag) simpleExoPlayer.getCurrentTag();
-        } catch (IndexOutOfBoundsException | ClassCastException error) {
+            final MediaItem currentMediaItem = simpleExoPlayer.getCurrentMediaItem();
+            if (currentMediaItem != null) {
+                final MediaItem.PlaybackProperties playbackProperties =
+                        currentMediaItem.playbackProperties;
+                metadata = (MediaSourceTag) (playbackProperties != null ? playbackProperties.tag
+                        : null);
+            } else {
+                metadata = null;
+            }
+        } catch (final IndexOutOfBoundsException | ClassCastException error) {
             if (DEBUG) {
                 Log.d(TAG, "Could not update metadata: " + error.getMessage());
                 error.printStackTrace();
@@ -3286,7 +3297,15 @@ public final class Player implements
     @Override // own playback listener
     @Nullable
     public MediaSource sourceOf(final PlayQueueItem item, final StreamInfo info) {
-        return (isAudioOnly ? audioResolver : videoResolver).resolve(info);
+        if (audioPlayerSelected()) {
+            return audioResolver.resolve(info);
+        } else {
+            if (isAudioOnly && !videoResolver.isVideoStreamVideoOnly()) {
+                return audioResolver.resolve(info);
+            }
+
+            return videoResolver.resolve(info);
+        }
     }
 
     public void disablePreloadingOfCurrentTrack() {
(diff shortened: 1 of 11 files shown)


TiA4f8R:
@alexlyzhov @Coderdude112 @Logarithmus @Darkmaster006 @IHateTheProgrammingClass @peepopoggers @avently @Stypox @opusforlife2 (pinging all participants of the corresponding issue) Can you please test this and see if it works for the cases where my changes should work (I described them on the PR body)? Thank you in advance.

app/src/main/java/org/schabi/newpipe/player/resolver/Resolver.java
```diff
@@ -6,4 +6,6 @@
 public interface Resolver<Source, Product> {
     @Nullable
     Product resolve(@NonNull Source source);
+
+    boolean isVideoStreamVideoOnly();
```
litetex, line 10: Could you add a short description as javadoc here?

app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java
```diff
@@ -118,6 +122,11 @@ public MediaSource resolve(@NonNull final StreamInfo info) {
         }
     }
 
+    @Override
+    public boolean isVideoStreamVideoOnly() {
```
litetex, line 126: I think that it should be noted that this is only set after ``resolve`` was called.

Maybe return an ``Optional`` instead? So that it's save that ``resolve`` was called.
Or maybe expand the returned data in ``resolve`` so that it contains this information?

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -192,37 +197,63 @@ static int getDefaultResolutionIndex(final String defaultResolution,
...
-                        && HIGH_RESOLUTION_LIST.contains(stream.getResolution())) {
-                    continue;
+        if (preferVideoOnlyStreams) {
+            if (videoStreams != null) {
+                for (final VideoStream stream : videoStreams) {
+                    if (!showHigherResolutions && HIGH_RESOLUTION_LIST.contains(
+                            stream.getResolution())) {
+                        continue;
+                    }
+                    retList.add(stream);
```
litetex, line 227: I see the same code here 4 times. Can we maybe de-duplicate it into a method?

litetex (commented):
Did a quick code review (no testing, just on GH's web-interface):

avently:
100% of videos I found are working without reload. Even ten years old. Is it real? Did google separate audio with video for all videos or at least for a big part of all videos?

It's amazing. When I tested this feature a couple of years ago it weren't so many videos with separated audio source.

Anyway, works cool at least for initial testing without hard scenarios (just background/foreground, screen on/off).

Good job!

Darkmaster006:
On my end it also seems to be working correctly. Tried clicking everywhere, and it doesn't ever seem to interrupt, great work, thank you so much!

Coderdude112:
I'm on a Pixel 3 running Android 12, and now I'm having the same issue as #6770 but also when the video is set to background. Other than that, the video is seamless when in popup and turning on / off the screen. Other than that I noticed the pause when clicking on the background / popup buttons is astronomically smaller, though it does reset the playback position.

opusforlife2:
@TiA4f8R You are a Super Saiyan.

opusforlife2:
> EDIT: Correction, this video (https://www.youtube.com/watch?v=p4w85vVSev4) has not worked. The interruption is there still. 

@TiA4f8R This is an OTF stream. No WebM streams available, only 3 mp4s and 1 3gpp stream (on this PR).

TiA4f8R:
> EDIT: Correction, this video (https://www.youtube.com/watch?v=p4w85vVSev4) has not worked. The interruption is there still. Other videos I've tried have worked, but this one hasn't.

It's intended, this video has OTF streams which are not supported by NewPipe right now (the support of these streams will be added in #6537). So only the highest quality (1080p) and 360p are video-only streams which can be played by NewPipe. So the 720p stream available is a video stream with audio (itag 22). So to prevent spending more data than needed, the previous process still happen (reload the player to play an audio stream).

> I'm on a Pixel 3 running Android 12, and now I'm having the same issue as #6770 but also when the video is set to background. Other than that, the video is seamless when in popup and turning on / off the screen. Other than that I noticed the pause when clicking on the background / popup buttons is astronomically smaller, though it does reset the playback position.

Does this happen when you remove the permission to display over other apps or when using the workaround with adb described in the issue?

Anyway, I will be not able to fix both of the issues (the first needs to implement Android PIP support as described in the issue, the second is #4497).

Coderdude112:
> Does this happen when you remove the permission to display over other apps or when using the workaround with adb described in the issue?

If I revoke the permission to draw over apps and then set the playback to background it works as expected, if I set the playback to popup I get an error from NewPipe  saying permission to draw over other apps is required.

If I am understanding correctly this PR solves the issue of the pause occurring when the playback is set to popup and the screen turns off or on. Is this correct?

Also, I will try the adb fix later today when I get home

MD77MD:
works great

peepopoggers:
waiting for responses to review :O

4censord:
I have tested this on a `Samsung Tab A7` and it worked perfectly.

peepopoggers:
@TiA4f8R have you resolved @litetex's review comments?

TiA4f8R:
@peepopoggers No, and I don't have the time to resolve them.

avently:
@TiA4f8R what's the reason for the last commit (progress loop releated)? 

It will be a reason for some bugs in the future. Right now when hide the app to background and stopped a played video or paused it, when you return to the app, you'll see a wrong timestamp under the video. Also as I far as I remember the progress loop saves currently played timestamp to db which then will be used before opening the same video. So an optimization you tried to make doesn't cost upcoming bugs

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2993,8 +2997,16 @@ private void maybeUpdateCurrentMetadata() {
...
+            final MediaItem currentMediaItem = simpleExoPlayer.getCurrentMediaItem();
+            if (currentMediaItem != null) {
+                final MediaItem.PlaybackProperties playbackProperties =
+                        currentMediaItem.playbackProperties;
+                metadata = (MediaSourceTag) (playbackProperties != null ? playbackProperties.tag
+                        : null);
+            } else {
+                metadata = null;
+            }
+        } catch (final IndexOutOfBoundsException | ClassCastException error) {
```
litetex, line 3009: This code can be refactored to something like this.
Notes in the comments
```java
...
        final MediaSourceTag metadata;
        try {
            final MediaItem currentMediaItem = simpleExoPlayer.getCurrentMediaItem();
            // One if block for fast exit
            if (currentMediaItem == null || currentMediaItem.playbackProperties == null
                || currentMediaItem.playbackProperties.tag == null) {
                return; // Null -> Return
            }
            metadata = (MediaSourceTag) currentMediaItem.playbackProperties.tag;
        } catch (final IndexOutOfBoundsException | ClassCastException ex) {
            if (DEBUG) {
                // Log exception nicely
                // Also error != Exception -> Fixed naming
                Log.d(TAG, "Could not update metadata", ex);
            }
            return;
        }

// The if here got removed
        maybeAutoQueueNextStream(metadata);
...
```

litetex (commented):
Another few comments from my side (the old ones above are still valid)

I also don't think https://github.com/TeamNewPipe/NewPipe/pull/7349/commits/5f8d49d89d4952030720ce7c5bfe35cbf69d165f is going to work. Because even when the video is disabled (``videoEnabled``) we need to update the PlayerNotfication.

I would also suggest doing stuff like optimizing the progress loop in another PR, as this seems to be unrelated to the current one.

litetex (approved):
Applied suggested changes.

Did a quick test and it seems to work as expected.

9Orlaf2:
Please merge this code for next version. ❤

9Orlaf2:
@TiA4f8R @litetex I'm trying the test apk now with a live video, and seems there's a sort of buffering problem (which does not occur on normal videos).

Can you solve it? This is the link to the video:

https://www.youtube.com/watch?v=DPXQG8S9Xwc

TiA4f8R:
@9Orlaf2 This will be not fixed with this PR, as the bug is not related to my changes (I can reproduce it too on video [5lMmnfVylEE](https://www.youtube.com/watch?v=5lMmnfVylEE) for example).

However, the stream can be rewinded a bit (30 seconds), so you can rewind or pause a bit the stream to prevent the issue (segments have more time to be fetched and the buffering should not happen or happen less frequently).

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -192,56 +202,55 @@ static int getDefaultResolutionIndex(final String defaultResolution,
...
             hashMap.put(videoStream.getResolution(), videoStream);
         }
 
         // Override the values when the key == resolution, with the defaultFormat
-        for (final VideoStream videoStream : retList) {
+        for (final VideoStream videoStream : allInitialStreams) {
             if (videoStream.getFormat() == defaultFormat) {
                 hashMap.put(videoStream.getResolution(), videoStream);
             }
         }
```
Stypox, line 251: What's the purpose of this `for`? Doesn't it put in the map again some of the same items from the for above? Also, can all of this be done with streams maybe?

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -309,28 +304,47 @@ static int getHighestQualityAudioIndex(@Nullable MediaFormat format,
...
+    }
+
+    /**
+     * Get the audio-stream from the list with the highest rank, depending on the comparator.
+     * Format will be ignored if it yields no results.
+     *
+     * @param targetedFormat The target format type or null if it doesn't matter
+     * @param audioStreams   List of audio streams
+     * @param comparator     The comparator used for determining the max/best/highest ranked value
+     * @return Index of audio stream that produces the most compact results or -1 if not found
```
Stypox, line 323: I think you forgot to update the `@return` description when copy-pasting the javadoc ;-)

app/src/main/java/org/schabi/newpipe/player/resolver/VideoPlaybackResolver.java
```diff
@@ -46,78 +49,95 @@ public VideoPlaybackResolver(@NonNull final Context context,
...
+            final AudioStream audio = audioStreams.isEmpty() ? null : audioStreams.get(
+                    ListHelper.getDefaultAudioFormat(context, audioStreams));
+            // Use the audio stream if there is no video stream, or
+            // Merge with audio stream in case if video does not contain audio
+            if (audio != null && (video == null || video.isVideoOnly)) {
+                final MediaSource audioSource = buildMediaSource(dataSource, audio.getUrl(),
+                        PlayerHelper.cacheKeyOf(info, audio),
+                        MediaFormat.getSuffixById(audio.getFormatId()), tag);
+                mediaSources.add(audioSource);
+                isVideoAndAudioSeparated = true;
```
Stypox, line 93: ~~Shouldn't this be `isVideoAndAudioSeparated = (video != null);`? Ok, the video is null only in corner cases, so the difference would never happen, but technically~~ I changed my mind, this is right as it is, since when there is no video audio is still separated from video

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4358,15 +4414,31 @@ private void setupVideoSurface() {
...
+            if (binding != null) {
+                binding.surfaceView.getHolder().removeCallback(surfaceHolderCallback);
             }
+            surfaceHolderCallback.release();
+            surfaceHolderCallback = null;
         }
     }
     //endregion
+
+    private int getVideoRendererIndex() {
```
Stypox, line 4428: Please add a little bit of javadoc to this method. After reading the code I get what it does, but a javadoc is always better

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3280,7 +3284,16 @@ public void onStartDrag(final PlayQueueItemHolder viewHolder) {
...
-        return (isAudioOnly ? audioResolver : videoResolver).resolve(info);
+        if (audioPlayerSelected()) {
+            return audioResolver.resolve(info);
+        } else {
+            if (isAudioOnly
+                    && !videoResolver.wasLastResolvedVideoAndAudioSeparated().orElse(false)) {
+                return audioResolver.resolve(info);
+            }
+
+            return videoResolver.resolve(info);
```
Stypox, line 3295: Add a comment here explaining why this is being done (and maybe point to this PR)

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4134,19 +4147,62 @@ public AppCompatActivity getParentActivity() {
...
+
+        In the case where we don't know the index of the video renderer, the play queue manager
+        is also reloaded. */
+
+        final StreamType streamType = info.getStreamType();
+
+        final boolean isVideoStreamTypeAndIsVideoOnlyStreamOrNoAudioStreamsAvailable =
+                (streamType == StreamType.VIDEO_STREAM || streamType == StreamType.LIVE_STREAM)
+                        && (videoResolver.wasLastResolvedVideoAndAudioSeparated().orElse(false)
+                            || isNullOrEmpty(info.getAudioStreams()));
```
Stypox, line 4187: Could you explain why `|| isNullOrEmpty(info.getAudioStreams()`?

Stypox (changes requested):
Looks mostly good. Thank you for such a great PR :-D

app/src/main/java/org/schabi/newpipe/player/Player.java
TiA4f8R: `isNullOrEmpty(info.getAudioStreams()` refers to the `NoAudioStreamsAvailable` part of the boolean name.

Why I did this? Because it's not needed to reload the play queue manager to try to get audio streams if there are no audio streams (like on most of old PeerTube videos): it will reload the same video stream (which is not video-only and contains audio) in this case, which is not needed.

That's the reason why I said `Otherwise the video renderer will be just disabled (because there is no other stream for it to play the audio)` in my comment above this line.

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
TiA4f8R: @litetex ^

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
TiA4f8R: I didn't change this, @litetex made these changes.

app/src/main/java/org/schabi/newpipe/player/Player.java
Stypox: Oh ok, that makes sense

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
litetex: > What's the purpose of this for?

I assume this is there to ensure that the default format is always there (by overriding potential non default ones in the hashmap).

Think of the following use case:

default format=mp4
res, format
480p, webm
480p, mp4
720p, webm
720p, mp4
1080p, webm

This should result in a list that looks like this:
480p, mp4
720p, mp4
1080p, webm


Maybe this could be improved by using java streams grouping, however I didn't touch it as it doesn't belong to this PR.

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
litetex, line 323:
```suggestion
     * @return Index of audio stream that produces the highest ranked result or -1 if not found
```

litetex: Done

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
Stypox: Ok, but if you take a look at the code you see that the second `for` always adds some items that were already added by the first one. This is because the first adds them all, the second just adds a subset of the ones added by the first one since there is an additional `if`.

> Maybe this could be improved by using java streams grouping, however I didn't touch it as it doesn't belong to this PR.

Well, didn't you convert to streams the code just above?

litetex: > Ok, but if you take a look at the code you see that the second for always adds some items that were already added by the first one. This is because the first adds them all, the second just adds a subset of the ones added by the first one since there is an additional if.

Note that the above code works with a ``HashMap``. The same keys (or in this case resolutions) will never be there twice. See also https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#put-K-V-

Okay so I let's go through this:
Assuming that default format=mp4 and ``allInitialStreams`` contains:
res, format
480p, mp4
480p, webm
720p, mp4
720p, webm
1080p, webm

After the first ``for`` the HashMap looks like this:
480p, webm
720p, webm
1080p, webm

However the user wanted default format=mp4 and that's why there is a second ``for`` which overrides the webm values with mp4. After this the map looks like this:
480p, mp4
720p, mp4
1080p, webm

> Well, didn't you convert to streams the code just above?

Yes because I couldn't build the code (thx to checkstyle-suppression file).

Stypox: > However the user wanted default format=mp4 and that's why there is a second for which overrides the webm values with mp4. After this the map looks like this:

You are right, sorry for the confusion ;-)

> Yes because I couldn't build the code (thx to checkstyle-suppression file).

I see, then these other lines shouldn't be touched in this PR

9Orlaf2:
> @9Orlaf2 This will be not fixed with this PR, as the bug is not related to my changes (I can reproduce it too on video [5lMmnfVylEE](https://www.youtube.com/watch?v=5lMmnfVylEE) for example).
> 
> However, the stream can be rewinded a bit (30 seconds), so you can rewind or pause a bit the stream to prevent the issue (segments have more time to be fetched and the buffering should not happen or happen less frequently).

Ok. I ran some tests, and actually the problem appears to be completely solved if the video is rewinded more than 30 seconds. Can you eventually update the PR with a new code, to set the starting playback 2 minutes earlier on live videos?

TiA4f8R:
No, like I already said, this will be not fixed with this PR because it's not related to my changes.

Also, I am not an ExoPlayer expert so I don't really know how to do this in the good way.

9Orlaf2:
> No, like I already said, this will be not fixed with this PR because it's not related to my changes.
> 
> Also, I am not an ExoPlayer expert so I don't really know how to do this in the good way.

Yeah... I actually wanted to know if it was also possible to create a different PR, but you claimed you don't know how to fix it regardless. No problem and thanks for the replies. 👍

9Orlaf2:
@Stypox Will you merge this soon?

app/src/main/java/org/schabi/newpipe/player/Player.java
litetex: @TiA4f8R Ping

Stypox (commented):
I tested a handful of combinations of actions and situations and it always worked as expected. Just waiting for the junit test to be extended and this is ready to go :-)

9Orlaf2:
> I tested a handful of combinations of actions and situations and it always worked as expected. Just waiting for the junit test to be extended and this is ready to go :-)

Wonderful. I'm waiting for it to be merged to download the apk file from dev branch. 👍

TiA4f8R:
> I tested a handful of combinations of actions and situations and it always worked as expected. Just waiting for the junit test to be extended and this is ready to go :-)

What should I test in the other test?

Stypox:
In `ListHelper` you changed a couple of things and added new methods, please test the new things that were introduced. If nothing new was introduced but the code was just converted to using streams, then there is no need for new tests, since the old tests already cover the various cases. So if that's the case then this is good to merge in my opinion

app/src/test/java/org/schabi/newpipe/util/ListHelperTest.java
```diff
@@ -47,15 +47,10 @@
     @Test
```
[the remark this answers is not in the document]
litetex: @Stypox 
Added some. Please review.

Stypox (approved):
Great, thank you! :-D
I tested thoroughly on my device (Android 10) and on emulators (API 19 and API 31).

litetex (approved):

mhmdanas:
An issue with this PR has been reported: https://github.com/TeamNewPipe/NewPipe/issues/7513

litetex:
@mhmdanas @9Orlaf2 
I can't locate the source of the problem as the exception doesn't fit the current source code:
```
...
Caused by: java.lang.NullPointerException
	at java.util.Objects.requireNonNull(Objects.java:220)
	at org.schabi.newpipe.player.Player.useVideoSource(Player.java:4149)
...
```
However on 
https://github.com/TeamNewPipe/NewPipe/pull/7349/files#diff-68cd2f051cf461f1cee8f97ab4a793ab90276943eed26f8f1baf5105948b2b5eR4194 there is no ``Objects.requireNonNull``

Seems like an outdated version was used...

mhmdanas:
@9Orlaf2 can you try to reproduce on this PR's latest build?

9Orlaf2:
> @mhmdanas @9Orlaf2 I can't locate the source of the problem as the exception doesn't fit the current source code:
> 
> ```
> ...
> Caused by: java.lang.NullPointerException
> 	at java.util.Objects.requireNonNull(Objects.java:220)
> 	at org.schabi.newpipe.player.Player.useVideoSource(Player.java:4149)
> ...
> ```
> 
> However on https://github.com/TeamNewPipe/NewPipe/pull/7349/files#diff-68cd2f051cf461f1cee8f97ab4a793ab90276943eed26f8f1baf5105948b2b5eR4149 there is no `Objects.requireNonNull`
> 
> Seems like an outdated version was used...

How it can be outdated? I just take apk from actions artifacts. 🤷‍♂️

mhmdanas:
That's completely expected; PRs can change quite often, so that's not unusual at all. Thus, please check that you're using the latest build before reporting an issue next time.

9Orlaf2:
> That's completely expected; PRs can change quite often, so that's not unusual at all. Thus, please check that you're using the latest build before reporting an issue next time.

Are you kidding me? 🤣. The last PR here is "Extended Test" made by litex, and is more older than the time I downloaded the test apk.

9Orlaf2:
Anyway I got many buffering problems with this feature on live videos. Probably are related to cache or something like that.

TiA4f8R:
@9Orlaf2 Are you still getting an app crash with my latest changes or not?

9Orlaf2:
> @9Orlaf2 Are you still getting an app crash with my latest changes or not?

I'll try to repro the bug again.

9Orlaf2:
> @9Orlaf2 Are you still getting an app crash with my latest changes or not?

Seems to be solved now. I'll try it for more time to be 100% sure about that.

9Orlaf2:
> @9Orlaf2 Are you still getting an app crash with my latest changes or not?

Ok it definitely works. Now, after got the same error, the playback start itself again. The unique problem is that after returning to the app, the video player appears to be disabled (with a bug play icon at the center) despite the audio still playing. Can you manage its status from the same exception code?

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4144,9 +4144,15 @@ private void useVideoSource(final boolean videoEnabled) {
 
         final int videoRenderIndex = getVideoRendererIndex();
```
litetex, line 4145: Please move this under the ``if (currentMetadata == null) { ... }`` check.
It will be useless when ``currentMeta`` is ``null``

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4144,9 +4144,15 @@ private void useVideoSource(final boolean videoEnabled) {
...
 
-        // We can safely assume that currentMetadata is not null (otherwise this method isn't
-        // called) so we can use the requireNonNull method of the Objects class.
-        final StreamInfo info = Objects.requireNonNull(currentMetadata).getMetadata();
+        // The current metadata may be null sometimes so we will be not able to execute the
+        // block above. Reload the play queue manager in this case.
```
litetex, line 4148: I think you mean below or?
Could you also note that this happens e.g. on unstable network connections.
You may also note that you reload the play queue manager because it's the default behavior without the optimization.

TiA4f8R: > I think you mean below or?

Yes, I mismatched the words in my mind 😅. Fixed with https://github.com/TeamNewPipe/NewPipe/pull/7349/commits/6341c14c870885a48512479a00e65dac78d0ef7c.

9Orlaf2:
@TiA4f8R This last push fix player status after recover? Otherwise the audio will be heard and the player will remain disabled (with the big play button in its middle).

TiA4f8R:
@9Orlaf2 No. Does the behavior you are describing happen on the current version (release or dev branch)? If yes, this will be not fixed with this PR; otherwise yes.

9Orlaf2:
> @9Orlaf2 No. Does the behavior you are describing happen on the current version (release or dev branch)? If yes, this will be not fixed with this PR; otherwise yes.

Happens with the apk of this PR.

TiA4f8R:
> Does the behavior you are describing happen on the current version (release or dev branch)?

I am talking about the 0.21.13 version or the latest APK of the repo's dev branch, which you can get [here](https://github.com/TeamNewPipe/NewPipe/suites/4583975183/artifacts/123522207), not about the APK from this PR.

9Orlaf2:
> > Does the behavior you are describing happen on the current version (release or dev branch)?
> 
> I am talking about the 0.21.13 version or the latest APK of the repo's dev branch, which you can get [here](https://github.com/TeamNewPipe/NewPipe/suites/4583975183/artifacts/123522207), not about the APK from this PR.

Ok, but what if it happens only with the apk of this PR? 🤷‍♂️

TiA4f8R:
So can you test if you can reproduce this bug with the 0.21.13 version or the latest APK of the repo's dev branch, which you can get [here](https://github.com/TeamNewPipe/NewPipe/suites/4583975183/artifacts/123522207) (and not with the APK from this PR)? If yes, this will be not fixed with this PR; otherwise it should be fixed in this PR.

9Orlaf2:
> So can you test if you can reproduce this bug with the 0.21.13 version or the latest APK of the repo's dev branch, which you can get [here](https://github.com/TeamNewPipe/NewPipe/suites/4583975183/artifacts/123522207) (and not with the APK from this PR)? If yes, this will be not fixed with this PR; otherwise it should be fixed in this PR.

I'll try again with both apk's.

9Orlaf2:
@TiA4f8R Ok i've tried with stable version, and the stream just stop without recover (and without generate crashes on the app). So, in theory, you should add a way to recover both playback and player together, otherwise after stream recover you'll find the player in this status (despite the audio can still be heard).

Screenshot_20211211-175758_NewPipe

Darkmaster006:
`org.schabi.newpipe.extractor.exceptions.ParsingException: Could not get dislike count
 at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.getDislikeCount(YoutubeStreamExtractor.java:388)
 at org.schabi.newpipe.extractor.stream.StreamInfo.extractOptionalData(StreamInfo.java:286)
 at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:73)
 at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:64)
 at org.schabi.newpipe.util.ExtractorHelper.lambda$getStreamInfo$3(ExtractorHelper.java:116)
 at org.schabi.newpipe.util.ExtractorHelper$$ExternalSyntheticLambda12.call(Unknown Source:4)
 at io.reactivex.rxjava3.internal.operators.single.SingleFromCallable.subscribeActual(SingleFromCallable.java:43)
 at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
 at io.reactivex.rxjava3.internal.operators.single.SingleDoOnSuccess.subscribeActual(SingleDoOnSuccess.java:35)
 at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
 at io.reactivex.rxjava3.internal.operators.maybe.MaybeFromSingle.subscribeActual(MaybeFromSingle.java:41)
 at io.reactivex.rxjava3.core.Maybe.subscribe(Maybe.java:5330)
 at io.reactivex.rxjava3.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.drain(MaybeConcatArray.java:153)
 at io.reactivex.rxjava3.internal.operators.maybe.MaybeConcatArray$ConcatMaybeObserver.request(MaybeConcatArray.java:78)
 at io.reactivex.rxjava3.internal.operators.flowable.FlowableElementAtMaybe$ElementAtSubscriber.onSubscribe(FlowableElementAtMaybe.java:66)
 at io.reactivex.rxjava3.internal.operators.maybe.MaybeConcatArray.subscribeActual(MaybeConcatArray.java:42)
 at io.reactivex.rxjava3.core.Flowable.subscribe(Flowable.java:15753)
 at io.reactivex.rxjava3.internal.operators.flowable.FlowableElementAtMaybe.subscribeActual(FlowableElementAtMaybe.java:36)
 at io.reactivex.rxjava3.core.Maybe.subscribe(Maybe.java:5330)
 at io.reactivex.rxjava3.internal.operators.maybe.MaybeToSingle.subscribeActual(MaybeToSingle.java:46)
 at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4813)
 at io.reactivex.rxjava3.internal.operators.single.SingleSubscribeOn$SubscribeOnObserver.run(SingleSubscribeOn.java:89)
 at io.reactivex.rxjava3.core.Scheduler$DisposeTask.run(Scheduler.java:614)
 at io.reactivex.rxjava3.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:65)
 at io.reactivex.rxjava3.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:56)
 at java.util.concurrent.FutureTask.run(FutureTask.java:266)
 at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:301)
 at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
 at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
 at java.lang.Thread.run(Thread.java:764)
Caused by: org.schabi.newpipe.extractor.exceptions.ParsingException: Ratings are enabled even though the dislike button is missing
 at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.getDislikeCount(YoutubeStreamExtractor.java:377)
 ... 29 more
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String[] java.lang.String.split(java.lang.String)' on a null object reference
 at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.getDislikeCount(YoutubeStreamExtractor.java:373)
 ... 29 more

</details>
</p></details>
<hr>`

This is what I get when I 'report' what seems to be an 'unexpected error'. What happens is... apparently nothing, but the 'error' pops up anytime you open a video, even though it doesn't really break anything. Thought I'd leave this here in any case.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4110,19 +4134,85 @@ public AppCompatActivity getParentActivity() {
...
+        final boolean isVideoWithSeparatedAudioOrVideoWithNoSeparatedAudioStreams =
+                sourceType == SourceType.VIDEO_WITH_SEPARATED_AUDIO
+                || (sourceType == SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY
+                        && isNullOrEmpty(info.getAudioStreams()));
+        final boolean isVideoStreamTypeAndIsVideoOnlyStreamOrNoAudioStreamsAvailable =
+                streamType == StreamType.VIDEO_STREAM
+                        && isVideoWithSeparatedAudioOrVideoWithNoSeparatedAudioStreams
+                        || (streamType == StreamType.LIVE_STREAM
+                            && (sourceType == SourceType.LIVE_STREAM
+                                || isVideoWithSeparatedAudioOrVideoWithNoSeparatedAudioStreams));
```
litetex, line 4193: Idea: This code block is very hard to read (at least for me) and the variable is very long.
Maybe we could extract it into a Utility method and call that in L4196?

I think a method would also be nicer because we could do something like this:
```java
// If it's a video livestream -> do trackselector stuff
if (streamType == StreamType.LIVE_STREAM && sourceType == SourceType.LIVE_STREAM) {
   return true;
}

// If it's
// * a video with separated audio
// * OR a video with audio only THAT has no audio streams
// -> go on
if (sourceType == SourceType.VIDEO_WITH_SEPARATED_AUDIO
                || (sourceType == SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY
                        && isNullOrEmpty(info.getAudioStreams())) {
   // If it's a video stream or a live stream -> do trackselector stuff
   if (streamType == StreamType.VIDEO_STREAM || streamType == StreamType.LIVE_STREAM) {
      return true;
   }
}

// default
return false
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4110,19 +4134,85 @@ public AppCompatActivity getParentActivity() {
...
+
+        For audio streams and audio live streams: nothing is done, it's not needed to reload the
+        player with the same audio stream.
+
+        For video live streams: the play queue manager is not reloaded if the stream source is a
+        live source (see VideoPlaybackResolver#resolve()) and if that's not the case, the
+        requirements for video streams is applied.
+
+        In the case where we don't know the index of the video renderer, the play queue manager
+        is also reloaded. */
```
litetex, line 4178: No it's not, only if it's is not a video AUDIO_STREAM or AUDIO_LIVE_STREAM

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4333,15 +4423,40 @@ private void setupVideoSurface() {
...
+     * {@link MappingTrackSelector.MappedTrackInfo} or {@link #RENDERER_UNAVAILABLE} if the current
+     * {@link MappingTrackSelector.MappedTrackInfo} is null or if there is no video renderer index.
+     *
+     * @return the video renderer index or {@link #RENDERER_UNAVAILABLE} if it cannot be get
+     */
+    private int getVideoRendererIndex() {
+        final MappingTrackSelector.MappedTrackInfo mappedTrackInfo = trackSelector
+                .getCurrentMappedTrackInfo();
+
+        if (mappedTrackInfo != null) {
```
litetex, line 4450: I hate nested ``{ }`` - in my opinion they make the code more complex and you can end up with a lot of indentations.

So maybe do a (quick exit style)
```
if (mappedTrackInfo == null) {
  return RENDERER_UNAVAILABLE; 
}
```
here instead.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4333,15 +4423,40 @@ private void setupVideoSurface() {
...
+            for (int i = 0; i < mappedTrackInfo.getRendererCount(); i++) {
+                final TrackGroupArray trackGroups = mappedTrackInfo.getTrackGroups(i);
+                if (!trackGroups.isEmpty()
+                        && simpleExoPlayer.getRendererType(i) == C.TRACK_TYPE_VIDEO) {
+                    return i;
+                }
+            }
```
litetex, line 4457: I think the Java Streaming API would be great here - less code, more understabale, no additioanl variables and for loops.
```java
return Instream.range(0, mappedTrackInfo.getRendererCount()) // Check every renderer
  .findFirst(i ->!mappedTrackInfo.getTrackGroups(i).isEmpty() // IDK Check where the trackgroup has something 
                  && simpleExoPlayer.getRendererType(i) == C.TRACK_TYPE_VIDEO) // Check where the renderer is a video renderer
 .orElse(RENDERER_UNAVAILABLE); // Otherwise return default
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4110,19 +4134,85 @@ public AppCompatActivity getParentActivity() {
         return (AppCompatActivity) ((ViewGroup) binding.getRoot().getParent()).getContext();
     }
 
-    private void useVideoSource(final boolean video) {
-        if (playQueue == null || isAudioOnly == !video || audioPlayerSelected()) {
+    private void useVideoSource(final boolean videoEnabled) {
```
litetex, line 4137: As pointed out by SonarQube the code here is very complex:
https://sonarcloud.io/project/issues?id=TeamNewPipe_NewPipe&open=AX5de2YtRtLRntykaKph&pullRequest=7349&resolved=false&types=CODE_SMELL

It would be great if we could it make simpler / break it down. I did a proposal down below.

app/src/main/java/org/schabi/newpipe/player/Player.java
litetex, line 4178: Idea: Binding this comment to actual code and not having a 10 line flowing text would be great and easier to read in the following code.

litetex (commented):
Did a review found some code quality things with proposals down below.

app/src/main/java/org/schabi/newpipe/player/Player.java
TiA4f8R: It should have been changed with my last commit.

litetex (commented):
LGTM, code looks nearly perfect.

Didn't test it so far.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3279,7 +3285,26 @@ public void onStartDrag(final PlayQueueItemHolder viewHolder) {
     @Override // own playback listener
     @Nullable
     public MediaSource sourceOf(final PlayQueueItem item, final StreamInfo info) {
-        return (isAudioOnly ? audioResolver : videoResolver).resolve(info);
+        if (audioPlayerSelected()) {
+            return audioResolver.resolve(info);
+        } else {
```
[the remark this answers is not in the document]
TiA4f8R: Done

litetex (dismissed):
Code LGTM

Didn't test it so far.
Please wait a bot so @Stypox has a chance to review (as he was requested).

litetex (commented):
Tested it.

@TiA4f8R
When I switch back from background to "foreground" player the video is black for a few seconds. (Also no loading animation or anything else)

TiA4f8R:
@litetex Can't you read the PR description, please?

> A black video is shown for a few seconds after you come back to the player. It is **intended**, because the video is not fetched when you are in background, to prevent you to spend more data than needed to play the audio. This is the behavior of the YouTube app on Android (for Premium clients only of course).

There is nothing I can do to change this I'm afraid.

The video stream is only fetched again when you come back to the app and if the video is still not fetched after some seconds after the video track is renabled, the player (so ExoPlayer) will switch in the loading state and not play until the video stream is enough fetched.

Stypox (changes requested):
Looks good to me! Just this little thing and it's good to go :-D
I tested a little bit more and found no issues

litetex:
> There is nothing I can do to change this I'm afraid.
> 
> The video stream is only fetched again when you come back to the app and if the video is still not fetched after some seconds after the video track is renabled, the player (so ExoPlayer) will switch in the loading state and not play until the video stream is enough fetched.

Sorry I overlooked that in the pretty long conversation here
grafik

Anyway...
Hmm I see. Is there really no way?
I think we should open an issue for that after this is released.

litetex:
@TiA4f8R 
Okay, then i think I can approve this after you rebased the PR and fixed Stypox's comment.

litetex (approved):

Stypox (approved):
Thanks, will merge after 0.22.1 :-)

avently:
@Stypox, why not to merge it in 0.22.1?

Stypox:
0.22.1 is a small bugfix release mostly fixing regressions, and we want to publish it fast without dedicating too much testing to it

PR #7175, opened 2021-09-26
Use NetCipher for handling TLS compatibility

mqus:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
This (somewhat) fixes #2898 by using netcipher(https://guardianproject.info/code/netcipher/) for handling TLS connections. It has the nice benefit that we can remove a lot of code on our side for handling Android Kitkat. 

A small downside is that Netcipher keeps TLS1.1 and 1.0 enabled for Android 19-23 at the moment while 1.0 was disabled for kitkat before, but I don't think this is a showstopper.

You may ask why I did not use the okhttp integration of netcipher: Would be far more complex to implement (we need a `Context` and an `Intent` which is only used for Tor integration inside netcipher)

#### APK testing 

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/suites/3876947801/artifacts/96495239)

These changes should be tested on all possible devices and versions as it impacts the whole connectivity on all platforms.
How to test? Check if the following things still work:

- Browsing/searching youtube, media.ccc.de, framatube, soundcloud (there should be no connection error)
- Viewing/hearing media from youtube, media.ccc.de, framatube, soundcloud (should load and display as before)
- Downloading media/subtitles (as applicable) from all providers. (should download as before)

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #7350, opened 2021-11-03
Enable tls related protocols for picasso on <=KitKat devices

evermind-zz:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

 Could not load images with Picasso. This problem only surfaced as I maintain a fork of NewPipe that supports the video platform Rumble and the thumbnails stopped to load. The reason is that NewPipe moved from UniversalImageLoader to Picasso. This PR fixes that. It might fix other platforms on those devices too.
  - enable TLSv1.1/1.2 for OkHttpClient.Builder for Picasso on <=KitKat devices that support those protocols but have not been enabled.
#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
@@ -1,11 +1,6 @@
 package org.schabi.newpipe;
 
 import android.content.Context;
-import android.os.Build;
-
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.preference.PreferenceManager;
 
 import org.schabi.newpipe.error.ReCaptchaActivity;
 import org.schabi.newpipe.extractor.downloader.Downloader;
@@ -14,32 +9,21 @@ import org.schabi.newpipe.extractor.downloader.Response;
 import org.schabi.newpipe.extractor.exceptions.ReCaptchaException;
 import org.schabi.newpipe.util.CookieUtils;
 import org.schabi.newpipe.util.InfoCache;
-import org.schabi.newpipe.util.TLSSocketFactoryCompat;
+import org.schabi.newpipe.util.OkHttpTlsHelper;
 
 import java.io.IOException;
-import java.security.KeyManagementException;
-import java.security.KeyStore;
-import java.security.KeyStoreException;
-import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
-import javax.net.ssl.SSLSocketFactory;
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.TrustManagerFactory;
-import javax.net.ssl.X509TrustManager;
-
-import okhttp3.CipherSuite;
-import okhttp3.ConnectionSpec;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.preference.PreferenceManager;
 import okhttp3.OkHttpClient;
 import okhttp3.RequestBody;
 import okhttp3.ResponseBody;
-
-import static org.schabi.newpipe.MainActivity.DEBUG;
 
 public final class DownloaderImpl extends Downloader {
     public static final String USER_AGENT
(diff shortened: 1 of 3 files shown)


litetex (commented):
I think this should be solved/done by #7175.

Btw: [TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA](https://ciphersuite.info/cs/TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA/) and [TLS_RSA_WITH_AES_256_CBC_SHA](https://ciphersuite.info/cs/TLS_RSA_WITH_AES_256_CBC_SHA/) are considered weak.

evermind-zz:
> I think this should be solved/done by #7175.
> 
> Btw: [TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA](https://ciphersuite.info/cs/TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA/) and [TLS_RSA_WITH_AES_256_CBC_SHA](https://ciphersuite.info/cs/TLS_RSA_WITH_AES_256_CBC_SHA/) are considered weak.

thanks for pointing that out. I will try that the NetCipher route #7175

evermind-zz:
I've checked NetCipher and get it up and running -- but they do NOT enable TLSv1.1 and TLSv1.2. So NetCipher is no solution for me as the server sp.rmbl.ws only supports TLSv1.2. Maybe I miss something to enable it there. Any hints appreciated. Thank you

litetex:
Okay, I reviewed the PR again and I don't understand the problems at all:
As stated in the javadoc:
https://www.javadoc.io/doc/com.squareup.okhttp3/okhttp/3.3.1/okhttp3/OkHttpClient.Builder.html#sslSocketFactory-javax.net.ssl.SSLSocketFactory-javax.net.ssl.X509TrustManager-

> Sets the socket factory and trust manager used to secure HTTPS connections. If unset, the system defaults will be used. 

We already overwrite the system default on devices that run Android 4.4 (API 19). 
So in theory there should be no problems as TLS 1.1. and 1.2 are already enabled for Picasso.

I think the problem here is that [PicassoHelper gets initialized](https://github.com/TeamNewPipe/NewPipe/blob/b5ad24eb475c1d0280fc8094438643acc77bbb74/app/src/main/java/org/schabi/newpipe/App.java#L107) before the [TLS is configured](https://github.com/TeamNewPipe/NewPipe/blob/b5ad24eb475c1d0280fc8094438643acc77bbb74/app/src/main/java/org/schabi/newpipe/MainActivity.java#L134-L137).
What happens if you move https://github.com/TeamNewPipe/NewPipe/blob/b5ad24eb475c1d0280fc8094438643acc77bbb74/app/src/main/java/org/schabi/newpipe/MainActivity.java#L134-L137 to https://github.com/TeamNewPipe/NewPipe/blob/b5ad24eb475c1d0280fc8094438643acc77bbb74/app/src/main/java/org/schabi/newpipe/App.java#L91?

Does this fix your problems?

evermind-zz:
This was my first try and it did not work. Does OkHttp3 build on top of [HttpsURLConnection](https://developer.android.com/reference/javax/net/ssl/HttpsURLConnection#setDefaultSSLSocketFactory(javax.net.ssl.SSLSocketFactory))? As the documentation states:

> Sets the default SSLSocketFactory inherited by new instances of this class.
>
> The socket factories are used when creating sockets for secure https URL connections.
>
I verified the OkHttpClient object is not using the TLSSocketFactoryCompat class in PicassoHelper but OpenSSLSocketFactoryImpl. Only the protocols ["TLSv1", "SSLv3"] are enabled

So I am not sure if this setting helps at all:
https://github.com/TeamNewPipe/NewPipe/blob/b5ad24eb475c1d0280fc8094438643acc77bbb74/app/src/main/java/org/schabi/newpipe/MainActivity.java#L134-L137

For me also media.ccc.de is not showing pictures on Android 4 emulator without this PR

In DownloaderImpl the TLSSocketFactoryCompat and additional Ciphers are explicitly set for its OkHttpClient.Builder().

I changed the PR now to not duplicate code and moved the DownloaderImpl enableModernTLS() method to a helper class and reuse that code to configure the PicassoHelper OkHttpClient.Builder() instance in the same way.

evermind-zz:
I created a new pull request using NetCipher which solves my problems

PR #7355, opened 2021-11-03
Add a "Add to playlist" item to the remote playlist menu

### LINKED ISSUE #7294: Add a Youtube playlist content inside a user playlist
### Checklist

#### What feature do you want?

Ability to **add a playlist** (or all of its content) from Youtube (Music) search results **inside another custom playlist**.

#### Why do you want this feature?

As of now, a user can add an already made playlist only generally in the _bookmarked playlist_, but I would like to be able to add all of the playlist music inside another playlist. For example, if I want to create a playlist which groups all of an artist's songs going album by album in the Youtube Music search results. 

Thanks for your important work!

petlyh:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

This PR adds an option to the menu for remote/non-local playlists
that appends the playlist items to a local playlist.

- Before:

- After:

https://user-images.githubusercontent.com/88139840/140081467-e98bbecd-a052-4cae-a3bb-fa0c2df97a94.mp4

- Fixes #7294

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
@@ -28,6 +28,7 @@ import org.reactivestreams.Subscription;
 import org.schabi.newpipe.NewPipeDatabase;
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.database.playlist.model.PlaylistRemoteEntity;
+import org.schabi.newpipe.database.stream.model.StreamEntity;
 import org.schabi.newpipe.databinding.PlaylistControlBinding;
 import org.schabi.newpipe.databinding.PlaylistHeaderBinding;
 import org.schabi.newpipe.error.ErrorInfo;
@@ -41,6 +42,7 @@ import org.schabi.newpipe.extractor.services.youtube.YoutubeParsingHelper;
 import org.schabi.newpipe.extractor.stream.StreamInfoItem;
 import org.schabi.newpipe.fragments.list.BaseListInfoFragment;
 import org.schabi.newpipe.info_list.dialog.InfoItemDialog;
+import org.schabi.newpipe.local.dialog.PlaylistDialog;
 import org.schabi.newpipe.local.playlist.RemotePlaylistManager;
 import org.schabi.newpipe.player.MainPlayer.PlayerType;
 import org.schabi.newpipe.player.playqueue.PlayQueue;
@@ -56,6 +58,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
 import io.reactivex.rxjava3.core.Flowable;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -267,6 +270,17 @@ public boolean onOptionsItemSelected(final MenuItem item) {
             case R.id.menu_item_bookmark:
                 onBookmarkClicked();
                 break;
+            case R.id.menu_item_append_playlist:
+                disposables.add(PlaylistDialog.createCorrespondingDialog(
+                        getContext(),
+                        getPlayQueue()
+                                .getStreams()
```
Stypox, line 277: The issue with this is that if the playlist was not completely loaded, only the first few items are added to the other playlist, not all of them. E.g. my YouTube music playlist has ~200 videos, but YouTube loads 100 items at once, so in order to add the whole playlist to another playlist with this option I'd first have to scroll down to the bottom so that I'm sure it was loaded completely.

Stypox (commented):
Code looks good (I didn't test). The only problem is the one I pointed out below. Maybe this could be merged anyway? The problem of not loading the whole playlist when we should is not only a problem of this PR but of some parts of the app in general, so I think merging this PR anyway is not really introducing a new problem.

SameenAhnaf:
The new option within three dots menu is slightly confusing. Rather, include currently available "Bookmark playlist" option within the playlist selection menu. 

IMG_20211117_001804

Stypox:
@SameenAhnaf sorry but no, that's a strange solution. Bookmarking should be easily available through the toolbar button like it is now (not hidden under two clicks). And putting "Add to playlist" to the toolbar hidden items is a perfectly normal and good practice thing to do in Android. Also, this "Add to playlist" button does not need to be that fast-accessible, hiding it under the three dot menu suffices.

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
petlyh: I had tried having it load the entire playlist automatically when selecting the menu option. However, the lack of a loading indicator when loading large playlists may be a problem (see video). What are your thoughts?

https://user-images.githubusercontent.com/88139840/144722928-d10e717a-18e3-4897-bf7f-7f24ecba7053.mp4

Stypox: I'm not sure. Somebody will definitely find this useful, but I wouldn't introduce full playlist loading in this PR since it introduces some things that need to be resolved, and I would address them in a separate PR. You'd be welcome to create such a PR :-)
- a loading indicator should be shown somewhere, as you said
- what happens if the playlist length is infinite (and we could not know that it is infinite)? Infinite playlists are e.g. YouTube mixes. We have to discuss this.

What do you think @petlyh @litetex @TobiGr @TiA4f8R ?

SameenAhnaf:
@Stypox Is this PR required to stay open? https://github.com/TeamNewPipe/NewPipe/pull/8008 is already merged.

Stypox:
I think this PR does a different thing, so I'd keep it. And I don't understand your sencond paragraph 😅

opusforlife2:
> This also allows saving a remote playlist locally.

@Stypox This is the line from the OP of #8008 that is being referred. The question is: since what this PR wishes to achieve can be done through #8008 (although in a slightly roundabout way), is this PR needed anymore?

MD77MD:
> > This also allows saving a remote playlist locally.
> 
> @Stypox This is the line from the OP of #8008 that is being referred. The question is: since what this PR wishes to achieve can be done through #8008 (although in a slightly roundabout way), is this PR needed anymore 

Yes... because it's not the same

Stypox:
@opusforlife2 yes, the roundabout way is strange, this PR is the proper way (and it is also so simple). Also, the roundabout way existed also before #8008, as I already said when reviewing the blogpost ;-)

@petlyh I would merge this, sorry for the delay. Could you rebase? Thanks :-)

petlyh:
> Could you rebase? Thanks :-)

@Stypox Done :-)

Stypox (approved):
Looks good to me, thank you! :-)
I tested a couple of situations (youtube normal, youtube learning, soundcloud, ...) on API 31 emulated and it works as expected

If you want to introduce infinite loading, a new PR would be well accepted ;-)

PR #7381, opened 2021-11-08
Prevent automatic replay after returning from background

litetex:
See also https://github.com/TeamNewPipe/NewPipe/pull/7195#issuecomment-962624380

#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Successor to #7195

* If the video has completed, when the app returns from background, prevent the player from staring playback automatically 

Analysis of the problem: https://github.com/TeamNewPipe/NewPipe/pull/7195#issuecomment-962624380

Steps to reproduce the bug:
1. Play a video
2. Lock the device while the video is still playing
3. Wait until the video ends
4. Unlock the device
5. The video starts playing again automatically

- Before:

https://user-images.githubusercontent.com/4525736/140655574-733a3749-ea7c-4979-b993-3847ed2a20dc.mp4

- After:

https://user-images.githubusercontent.com/4525736/140655671-53f10945-2987-4843-a71a-95b09d5e1216.mp4

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -859,14 +859,8 @@ public final class Player implements
         final long windowPos = simpleExoPlayer.getCurrentPosition();
         final long duration = simpleExoPlayer.getDuration();
 
-        if (windowPos > 0
-                // Sometimes (e.g. when the playback ended) the windowPos is a few milliseconds
-                // higher than the duration. Due to this a little buffer (100ms) was introduced.
-                // See also https://github.com/TeamNewPipe/NewPipe/pull/7195#issuecomment-962624380
-                && windowPos <= duration + 100
-        ) {
-            setRecovery(queuePos, Math.min(windowPos, duration));
-        }
+        // No checks due to https://github.com/TeamNewPipe/NewPipe/pull/7195#issuecomment-962624380
+        setRecovery(queuePos, Math.max(0, Math.min(windowPos, duration)));
     }
 
     private void setRecovery(final int queuePos, final long windowPos) {
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -857,9 +857,15 @@ public final class Player implements
 
         final int queuePos = playQueue.getIndex();
         final long windowPos = simpleExoPlayer.getCurrentPosition();
+        final long duration = simpleExoPlayer.getDuration();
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -860,9 +860,15 @@ public void setRecovery() {
...
 
-        if (windowPos > 0 && windowPos <= simpleExoPlayer.getDuration()) {
-            setRecovery(queuePos, windowPos);
+        if (windowPos > 0
+                // Sometimes (e.g. when the playback ended) the windowPos is a few milliseconds
+                // higher than the duration. Due to this a little buffer (100ms) was introduced.
+                // See also https://github.com/TeamNewPipe/NewPipe/pull/7195#issuecomment-962624380
+                && windowPos <= duration + 100
+        ) {
+            setRecovery(queuePos, Math.min(windowPos, duration));
```
XiangRongLin, line 871: Is the check for `windowPos <= duration` even needed? Due to the `Math.min` no value bigger than `duration` can be passed onto `setRecovery`
In theory `windowPos` should only go from 0 to `duration`. Since it apparently does not do it, is there a difference in meaning from a `windowPos` that is  100 over `duration` and one that is 200 over `duration`

```suggestion
        if (windowPos > 0) {
            // Sometimes (e.g. when the playback ended) the windowPos is a few milliseconds
            // higher than the duration.
            // See also https://github.com/TeamNewPipe/NewPipe/pull/7195#issuecomment-962624380
            setRecovery(queuePos, Math.min(windowPos, duration));
```

litetex: > Is the check for windowPos <= duration even needed?

I didn't implement it and left no comment and context (git blame also didn't really help since the code was copied more than 2 times) - like the author before - so I assume it does something useful.

However if we do this we can also do ``setRecovery(queuePos, Math.max(0, Math.min(windowPos, duration)));`` so that there are no checks at all 😄

XiangRongLin: > However if we do this we can also do setRecovery(queuePos, Math.max(0, Math.min(windowPos, duration))); so that there are no checks at all 😄

This seems to also fix the bug. But if that breaks something else is something is another question which I can't answer.

litetex: I removed the check completely.
Everything seems to be working as expected.

XiangRongLin (approved):

PR #7412, opened 2021-11-16
Some code cleanup(s)

litetex:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

Code cleanup
See the commits for more detailed description.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
+++ b/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
@@ -210,11 +210,8 @@ public class PlayerGestureListener
             Log.d(TAG, "onScrollEnd called with playerType = ["
                 + player.getPlayerType() + "]");
         }
+        
         if (playerType == MainPlayer.PlayerType.VIDEO) {
-            if (DEBUG) {
-                Log.d(TAG, "onScrollEnd() called");
-            }
-
             if (player.getVolumeRelativeLayout().getVisibility() == View.VISIBLE) {
                 animate(player.getVolumeRelativeLayout(), false, 200, SCALE_AND_ALPHA,
                         200);
--- a/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
+++ b/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
@@ -72,7 +72,6 @@ public class PlayerGestureListener
 
         if (player.isControlsVisible()) {
 
-            // -- Controls visible --
 
             player.hideControls(150, 0);
             return;
--- a/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
+++ b/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
@@ -71,8 +71,6 @@ public class PlayerGestureListener
         }
 
         if (player.isControlsVisible()) {
-
-
             player.hideControls(150, 0);
             return;
         }
--- a/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
+++ b/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
@@ -99,6 +99,8 @@ public class PlayerGestureListener
                 + player.getPlayerType() + "], portion = [" + portion + "]");
         }
         if (playerType == MainPlayer.PlayerType.VIDEO) {
+
+            // -- Brightness and Volume control --
             final boolean isBrightnessGestureEnabled =
                 PlayerHelper.isBrightnessGestureEnabled(service);
             final boolean isVolumeGestureEnabled = PlayerHelper.isVolumeGestureEnabled(service);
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
```diff
@@ -69,26 +69,22 @@ public void onSingleTap(@NotNull final MainPlayer.PlayerType playerType) {
         if (DEBUG) {
             Log.d(TAG, "onSingleTap called with playerType = [" + player.getPlayerType() + "]");
         }
-        if (playerType == MainPlayer.PlayerType.POPUP) {
 
-            if (player.isControlsVisible()) {
-                player.hideControls(100, 100);
-            } else {
-                player.getPlayPauseButton().requestFocus();
```
TobiGr, line 77: I can only guess (or run git blame), but I think requesting focus here makes sense when you have a device with a physical keyboard which allows you to pause / play the video by pressing the spacebar. Putting the focus on the play/pause button is only reasonable when using the popup player (the main player has more buttons, so switching the focus might be weird for the user, especially when using NewPipe on an Android TV with a DPad). I'd keep that line and add a little hint.

TobiGr: Uhm, sorry if I wrote that in a way leaving room for misinterpretation. The pop-up player does not work on TVs, because it cannot be moved (and that makes the whole feature useless...). The focus request is for phones with physical keyboards.

litetex: > The focus request is for phones with physical keyboards.

This is such a niche that I don't think it's even worth implementing.
I also don't know if it's expected that the playPause-Button gains focus when e.g. the player is paused via headphone.
Particularly if the popup player will be replace by a PiP.

More insight:
[There are currently like 3 phones sold that have a physical keyboard](https://www.androidauthority.com/keyboard-phones-845839/) and all also have a touchscreen...

litetex: Removed commit.

app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
```diff
@@ -69,26 +69,22 @@ public void onSingleTap(@NotNull final MainPlayer.PlayerType playerType) {
...
-                player.showControlsThenHide();
-            }
+        if (player.isControlsVisible()) {
 
-        } else /* playerType == MainPlayer.PlayerType.VIDEO */ {
+            // -- Controls visible --
```
Stypox, line 75: This comment and the newlines sorrounding it are unneeded, the argument of the `if` says the exact same thing and the comment below also clears any possible leftover doubt

app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
```diff
@@ -69,26 +69,22 @@ public void onSingleTap(@NotNull final MainPlayer.PlayerType playerType) {
...
-                if (player.getCurrentState() == Player.STATE_COMPLETED) {
-                    player.showControls(0);
-                } else {
-                    player.showControlsThenHide();
-                }
-            }
+            player.hideControls(150, 0);
+            return;
+        }
+
```
Stypox, line 80: Remove this newline, too

Stypox (changes requested):
I agree with you that removing the undocumented focus request is the right thing to do, then this is good to go (except for the small formatting issue). Btw the requestFocus line of code was added in a merge commit of #2907: https://github.com/TeamNewPipe/NewPipe/blame/3871d5aed77f63c74867b6076b31a61b69a8fda7/app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java#L370

app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
litetex, line 75: Let's see if we can fix this via GH

```suggestion
```

app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
litetex, line 79:
```suggestion
```

litetex:
@Stypox 

Please review again.

app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
```diff
@@ -69,26 +69,20 @@ public void onSingleTap(@NotNull final MainPlayer.PlayerType playerType) {
...
 
-            if (player.isControlsVisible()) {
-                player.hideControls(100, 100);
-            } else {
-                player.getPlayPauseButton().requestFocus();
-                player.showControlsThenHide();
-            }
+        if (player.isControlsVisible()) {
 
-        } else /* playerType == MainPlayer.PlayerType.VIDEO */ {
```
Stypox, line 75: There are still two misleading newlines separating the `if` from its body

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/player/event/PlayerGestureListener.java
litetex: Done

Stypox (approved):
LGTM

PR #7435, opened 2021-11-23
Confirmation before delete download

### LINKED ISSUE #5419: Confirmation before deleting all downloads

Marc2Serre:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

We have a confirmation dialog when we delete a single file from the download menu like in the screenshot below. We didn't translate the text. It is only in english. How the translation works? Is done automatically or manually?

- After:
image

- Fixes #5419 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
+++ b/app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
@@ -678,9 +678,19 @@ public class MissionAdapter extends Adapter<ViewHolder> implements Handler.Callb
                 shareFile(h.item.mission);
                 return true;
             case R.id.delete:
-                mDeleter.append(h.item.mission);
-                applyChanges();
-                checkMasterButtonsVisibility();
+                AlertDialog.Builder builder = new AlertDialog.Builder(mContext);
+                builder.setMessage(R.string.confirmation_before_deleting_download)
+                        .setPositiveButton(R.string.yes, (dialog, which) -> {
+                            mDeleter.append(h.item.mission);
+                            applyChanges();
+                            checkMasterButtonsVisibility();
+                        })
+                        .setNegativeButton(R.string.cancel, (dialog, which) -> {
+                            //nothing happens
+                        });
+                // Create the AlertDialog object and run it
+                builder.create();
+                builder.show();
                 return true;
             case R.id.md5:
             case R.id.sha1:
--- a/app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
+++ b/app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
@@ -679,7 +679,7 @@ public class MissionAdapter extends Adapter<ViewHolder> implements Handler.Callb
                 return true;
             case R.id.delete:
                 AlertDialog.Builder builder = new AlertDialog.Builder(mContext);
-                builder.setMessage(R.string.confirmation_before_deleting_download)
+                builder.setMessage(R.string.delete_downloaded_files)
                         .setPositiveButton(R.string.yes, (dialog, which) -> {
                             mDeleter.append(h.item.mission);
                             applyChanges();
--- a/app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
+++ b/app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
@@ -678,19 +678,15 @@ public class MissionAdapter extends Adapter<ViewHolder> implements Handler.Callb
                 shareFile(h.item.mission);
                 return true;
             case R.id.delete:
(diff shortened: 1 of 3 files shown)


app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
```diff
@@ -678,9 +678,19 @@ private boolean handlePopupItem(@NonNull ViewHolderItem h, @NonNull MenuItem opt
...
+                            mDeleter.append(h.item.mission);
+                            applyChanges();
+                            checkMasterButtonsVisibility();
+                        })
+                        .setNegativeButton(R.string.cancel, (dialog, which) -> {
+                            //nothing happens
+                        });
+                // Create the AlertDialog object and run it
+                builder.create();
+                builder.show();
```
Stypox, line 693: This improves the formatting, removes an unneeded variable, shows the name of the file being deleted in the dialog.
```suggestion
                new AlertDialog.Builder(mContext)
                        .setMessage(mContext.getString(R.string.delete_selected_file, h.item.mission.storage.getName()))
                        .setNegativeButton(R.string.cancel, null)
                        .setPositiveButton(R.string.yes, (dialog, which) -> {
                            mDeleter.append(h.item.mission);
                            applyChanges();
                            checkMasterButtonsVisibility();
                        })
                        .show();
```

app/src/main/res/values/strings.xml
```diff
@@ -583,6 +583,7 @@
     <string name="clear_download_history">Clear download history</string>
     <string name="confirm_prompt">Do you want to clear your download history or delete all downloaded files?</string>
     <string name="delete_downloaded_files">Delete downloaded files</string>
+    <string name="delete_selected_file">Do you want to delete the selected file ?</string>
```
Stypox, line 586: This shows the file being deleted in the string
```suggestion
    <string name="delete_selected_file">Do you want to delete %1$s?</string>
```

Stypox (changes requested):
Thank you :-)

> We didn't translate the text. It is only in english.

That's ok. It will be translated on Weblate separately from this PR, see the explanation in the README and in the contribution guidelines.

TiA4f8R (changes requested):
Thank you for your pull request!

> We didn't translate the text. It is only in english. How the translation works? Is done automatically or manually?

The translation is made by translators on [our Weblate](https://hosted.weblate.org/projects/newpipe/), but you can add translations of the languages you are speaking in your PR if you want (in the `strings.xml` resource file, which corresponds to language set in the folder: for example, the `strings.xml` file of the `values-fr` folder contains the French translations of the app).
If a translation is not available, string resources of the original language will be displayed (English in our case).

You need to do some changes first and then your changes should be good to merged.

Question for other team members: should we remove the toast displayed to cancel the deletion after the AlertDialog is dismissed in the case the deletion is confirmed by the user or not?

Stypox:
@TiA4f8R I think @Marc2Serre would be confused to see two reviews asking him to do two different things ;-)

> Question for other team members: should we remove the toast displayed to cancel the deletion after the AlertDialog is dismissed in the case the deletion is confirmed by the user or not?

What? ;-)

Hugov8:
Thanks for the feedback, we change the code with your suggestions and we add the french translation

litetex:
Question to the devs: Do we really need this in the first place?

There are already multiple fail-safes so I don't think we need another one:
* When deleting a single file a snackbar shows up (at the bottom) which allows undoing the deletion
* When clicking at the trash icon in the top right corner a dialog shows up which asks you if you really want to delete all downloaded files

So what's the point of having this? I see no advantage.

litetex:
This PR will also not solve #5419 "Confirmation before deleting >**all**< downloads".
The issue is about the trash can in the top right corner, not about the deletion of a single file.
I think we should unlink the PR and the issue.

Stypox:
> So what's the point of having this? I see no advantage.

Mmmh, didn't think about this, I think you are right

litetex:
Killing this PR as it doesn't introduces a (noteable) benefit and doesn't fix the underlying issue:
* https://github.com/TeamNewPipe/NewPipe/pull/7435#issuecomment-979464947
* https://github.com/TeamNewPipe/NewPipe/pull/7435#issuecomment-979467969

PR #7452, opened 2021-11-27
Show an alert-dialog when no appropriate file-manager was found

### LINKED ISSUE #7113: Catch ``ActivityNotFoundException`` (file-manager) and display a meaningful error
### Checklist

There are currently many issues that say that an file-manager - that should be shipped by a "standard" android - is missing:
https://github.com/TeamNewPipe/NewPipe/search?o=desc&q=No+Activity+found+to+handle+Intent&s=created&type=issues

This is usually due to incomplete custom ROMs or because the user uninstalled it's file-manager.

Background: It reached my tipping-point when I got 2 new comment-notifications in https://github.com/TeamNewPipe/NewPipe/issues/7076 within 12h as some people seem too be unable to read properly what I wrote a few comments above or search for the error.

Also sometimes the ACRA report seems to fail: https://github.com/TeamNewPipe/NewPipe/issues/7076#issuecomment-920322738

Anyway the missing file-manager usually results in a ``ActivityNotFoundException``, e.g:
* ``
android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.OPEN_DOCUMENT cat=[android.intent.category.OPENABLE] typ=*/* flg=0x43 (has extras) }
``
* ``
android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.CREATE_DOCUMENT cat=[android.intent.category.OPENABLE] typ=application/zip flg=0x43 (has extras) }
``

It would be great if we could catch exceptions like those and show a dialog/toast that says something like "Unable to open file-manager... Is one installed?".
A simple try catch on certain code blocks should be enough 😄

https://github.com/TeamNewPipe/NewPipe/search?o=desc&q=No+Activity+found+to+handle+Intent&s=created&type=issues

Devs/Mods: Less issues that can have to investigated and closed as duplicates
Users: Faster knowledge of what causes the error → can fix it themselves

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- See the linked issue for more details about the problem
- If no file-manager (or SAF compatible) file manager is installed on a device now an alert shows up

- Before: ``<App crashed>``
- After: 
grafik
grafik ← Android 10+

- Fixes #7113 and more than 10 other issues where users managed to have no file-manager.

#### APK testing 

#### Due diligence

### How to test this?
To test this you have to get rid of all your file-managers (may led to data loss or other side effects). This may require disabling system apps.

I strongly recommend doing this on an emulator only!

How to do it on an emulated Android 10 device:
* Wipe all data from that device
* Open a shell/commandline
* Connect to the emulator using ``adb shell`` (ensure that you are connected to the correct device!)
* Get into the sudo mode: ``su``
* Disable Androids default file-manager: ``pm disable com.android.documentsui``
→ Now you can deploy NewPipe to the emulator and check if the changes work as expected


mhmdanas:
I think all usages of "file-picker" and "directory-picker" should be unhyphenated (as in "file picker" and "directory picker").

<sub>sorry for nitpicking</sub>

Stypox (changes requested):
Isn't there too much code duplication? Why not just add a `launchAndCheck(ActivityResultLauncher<Intent>, Intent, Context)` function to `NoFileManagerHelper`, that does the try-catch?

app/src/main/java/org/schabi/newpipe/streams/io/NoFileManagerSafeGuard.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Shows an alert dialog when no file-manager is found.
+     * @param context Context
+     */
+    private static void showActivityNotFoundAlert(@NonNull final Context context) {
+        if (context == null) {
+            throw new IllegalArgumentException(
+                    "Unable to open no file manager alert dialog: Context is null");
+        }
```
XiangRongLin, line 29: Can you even pass in null, if you annotate it as `@NonNull`?

litetex: https://stackoverflow.com/a/35604877

If you _send_ ``null`` it throws an exception.

mhmdanas: Doesn't it warn in Android Studio though?

TacoTheDank:
Kinda makes you wonder how they managed to install the app in the first place, huh? 😉

app/src/main/java/org/schabi/newpipe/streams/io/NoFileManagerSafeGuard.java
XiangRongLin: I don't quite like it coming from the Kotlin world, but I guess thats just how Java is 😕

XiangRongLin (approved):

litetex:
@Stypox 
Please review again :)

app/src/main/res/values/strings.xml
```diff
@@ -671,6 +671,7 @@
     <string name="recent">Recent</string>
     <string name="chapters">Chapters</string>
     <string name="no_app_to_open_intent">No app on your device can open this</string>
+    <string name="no_appropriate_file_manager_message">No appropriate file manager was found for this action.\nPlease install a file manager or try to enable/disable \'%s\' in the download settings.</string>
```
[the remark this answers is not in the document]
Stypox: @litetex have you seen this?

> Also, the string should be different for users on Android 10+, since there SAF is not disableable, so it is misleading to show a message that suggests to disable it.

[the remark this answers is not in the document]
litetex: Added a switch for Android 10+ to use a different - better understandable - string.

litetex:
@Stypox 
grafik

> Kinda makes you wonder how they managed to install the app in the first place, huh? 😉

Black magic I guess 😆 (they likely installed it with F-Droid which requires no file-manager for installation)

Stypox (approved):
Great :-)

Issue #7302, opened 2021-10-23
"Start in full screen" bug: Switching audio -> main player causes "unrecoverable error" on rotation


opusforlife2:
### Checklist

- [x] I am using the latest version - 0.21.13 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

1. Turn on the "Start main player in fullscreen" toggle.
2. Set "Autoplay" to Never.
3. Open any video details in portrait mode.
4. Start background playback.
5. Tap on the thumbnail to switch to the main player.

### Actual behavior

Newpipe rotates to landscape. The thumbnail keeps showing. After a short gap, the "Unrecoverable player error occurred" toast shows up. Nothing else happens.

If you tap on the thumbnail again, then the main player opens normally, resuming playback at the correct timestamp.

### Expected behavior

The main player should start properly, without any error.

### Device info

 - Android version/Custom ROM version: LOS 17.1
 - Device model: Mi 5

AudricV:
Is it still an issue on the `dev` branch, especially since the player refactor PR?

opusforlife2:
It's not even occurring on 0.23.1, actually! 🎉

PR #7482, opened 2021-12-02
Unify error reporting and add error notification

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Followup to #7142
- Created a `ErrorUtil` class that contains the three ways to report an error:
  - `openActivity`: just opens the error activity as always
  - `showSnackbar`: just opens the error snackbar as always
  - `createNotification`: creates a notification with a description containing the usual error message, and a title explaining that by tapping on the notification you can report the error. The notification disappears upon tapping, and is set to `IMPORTANCE_LOW` like other NewPipe notifications so that it does not make sounds.
- Correctly prevent exception from being serialized in `ErrorInfo`, eliminating the need for `EnsureExceptionSerializable` (see commit message for more info)
- Remove `PlayerErrorHandler` and instead always show error notification when an error occurs in the player. Also remove the related "report player error" setting.
- Add two new options in debug settings to show an error snackbar or create an error notification

#### APK testing 

To test I suggest using the "Crash the player" button from #7142 or triggering #7302 @litetex 

#### Due diligence


app/src/main/java/org/schabi/newpipe/App.java
```diff
@@ -246,9 +246,16 @@ private void initNotificationChannels() {
                 .setDescription(getString(R.string.hash_channel_description))
                 .build();
 
+        final NotificationChannelCompat errorReportChannel = new NotificationChannelCompat
+                .Builder(getString(R.string.error_report_channel_id),
+                NotificationManagerCompat.IMPORTANCE_LOW)
```
litetex, line 251: Shouldn't there be an indentation

litetex (commented):
Code LGTM only minor things.

Tested it a bit, seems to work as expected.

I'm glad that you managed to get rid of the ugly exception workaround 😄

Only one thing is a bit confusing:
When e.g. a player error happens, the player just stops and I don't notice the notification at the top (in particularly when the notification bar is already full due to my phones notch).
So I think showing a toast that says "Unexpected error occurred, tap notification for details" would be great. I would show the toast whenever a new error notification is shown.

app/src/main/java/org/schabi/newpipe/App.java
Stypox: It also looks strange to me, but I just copied the formatting from the other notification channels above. Should I change them all?

Stypox:
> When e.g. a player error happens, the player just stops and I don't notice the notification at the top (in particularly when the notification bar is already full due to my phones notch).
So I think showing a toast that says "Unexpected error occurred, tap notification for details" would be great. I would show the toast whenever a new error notification is shown.

I agree, I'll improve this

app/src/main/java/org/schabi/newpipe/App.java
litetex: I don't know.
Do it if you want :)

Stypox:
You can also test the player error by trying to play in background this live: https://www.youtube.com/watch?v=Q3uwXbkheDo

Stypox:
@litetex done :-)

app/src/main/java/org/schabi/newpipe/error/ErrorUtil.kt
```diff
@@ -0,0 +1,117 @@
...
+import android.os.Build
+import android.view.View
+import android.widget.Toast
+import androidx.core.app.NotificationCompat
+import androidx.core.content.ContextCompat
+import androidx.fragment.app.Fragment
+import com.google.android.material.snackbar.Snackbar
+import org.schabi.newpipe.R
+
+class ErrorUtil {
```
litetex, line 18: I would appreciate some javadoc here, otherwise I think we are fine :)

Stypox: Yeah, I also wanted to add some javadoc but at the end I forgot. Done ;-)

litetex: Very nice. I love it 👍

litetex (approved):

Mhowser:
I can't find the artifacts section you were describing earlier
Screenshot_20211229-164835

litetex:
> The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

Click on CI and then scroll down ;)
Note: You have to be logged in otherwise GH won't show you the artifacts.

Mhowser:
Ah, I was clicking the dropdown arrow, oops :P thanks for the help!

PR #7491, opened 2021-12-04
Fix order of local search results

### LINKED ISSUE #7133: NewPipe v0.21.10 broke 'Search History' functionality
### Checklist
- [x] I am using the latest version - v0.21.10

1. Press the search icon
2. Search YouTube for anything, say 'Taylor Swift' and click any of the search suggestions

### Actual behavior
The search itself works, but my searched text 'Taylor Swift' doesn't get added to my search history

### Expected behavior
The searched text 'Taylor Swift' should get added to my search history

### Device info
 - Android version/Custom ROM version: MIUI v12.0.3 running Android 10 32-bit
 - Device model: Xiaomi Redmi 7A (2GB RAM, 32GB storage, 17GB free space)

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Results from the local database were being fetched from the database in the correct order, but then the `DISTINCT` part of the query was not done by SQL but by Java using a `Set`, which then caused all items to be ordered alphabetically and not by creation date. This PR fixes this by having the related queries just return a list of strings that are already the correct ones and in the correct order. I just added `DISTINCT` to the queries.

Fixes #7133

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
+++ b/app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
@@ -19,6 +19,7 @@ import static org.schabi.newpipe.database.history.model.SearchHistoryEntry.TABLE
 @Dao
 public interface SearchHistoryDAO extends HistoryDAO<SearchHistoryEntry> {
     String ORDER_BY_CREATION_DATE = " ORDER BY " + CREATION_DATE + " DESC";
+    String ORDER_BY_MAX_CREATION_DATE = " ORDER BY MAX(" + CREATION_DATE + ") DESC";
 
     @Query("SELECT * FROM " + TABLE_NAME
             + " WHERE " + ID + " = (SELECT MAX(" + ID + ") FROM " + TABLE_NAME + ")")
@@ -36,16 +37,16 @@ public interface SearchHistoryDAO extends HistoryDAO<SearchHistoryEntry> {
     @Override
     Flowable<List<SearchHistoryEntry>> getAll();
 
-    @Query("SELECT * FROM " + TABLE_NAME + " GROUP BY " + SEARCH + ORDER_BY_CREATION_DATE
-            + " LIMIT :limit")
-    Flowable<List<SearchHistoryEntry>> getUniqueEntries(int limit);
+    @Query("SELECT " + SEARCH + " FROM " + TABLE_NAME + " GROUP BY " + SEARCH
+            + ORDER_BY_MAX_CREATION_DATE + " LIMIT :limit")
+    Flowable<List<String>> getUniqueEntries(int limit);
 
     @Query("SELECT * FROM " + TABLE_NAME
             + " WHERE " + SERVICE_ID + " = :serviceId" + ORDER_BY_CREATION_DATE)
     @Override
     Flowable<List<SearchHistoryEntry>> listByService(int serviceId);
 
-    @Query("SELECT * FROM " + TABLE_NAME + " WHERE " + SEARCH + " LIKE :query || '%'"
-            + " GROUP BY " + SEARCH + " LIMIT :limit")
-    Flowable<List<SearchHistoryEntry>> getSimilarEntries(String query, int limit);
+    @Query("SELECT " + SEARCH + " FROM " + TABLE_NAME + " WHERE " + SEARCH + " LIKE :query || '%'"
+            + " GROUP BY " + SEARCH + ORDER_BY_MAX_CREATION_DATE + " LIMIT :limit")
+    Flowable<List<String>> getSimilarEntries(String query, int limit);
 }
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
@@ -1,5 +1,10 @@
 package org.schabi.newpipe.fragments.list.search;
 
+import static androidx.recyclerview.widget.ItemTouchHelper.Callback.makeMovementFlags;
+import static org.schabi.newpipe.ktx.ViewUtils.animate;
+import static org.schabi.newpipe.util.ExtractorHelper.showMetaInfoInTextView;
+import static java.util.Arrays.asList;
+
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
```diff
@@ -36,16 +36,16 @@
...
     @Query("SELECT * FROM " + TABLE_NAME
             + " WHERE " + SERVICE_ID + " = :serviceId" + ORDER_BY_CREATION_DATE)
     @Override
     Flowable<List<SearchHistoryEntry>> listByService(int serviceId);
 
-    @Query("SELECT * FROM " + TABLE_NAME + " WHERE " + SEARCH + " LIKE :query || '%'"
-            + " GROUP BY " + SEARCH + " LIMIT :limit")
-    Flowable<List<SearchHistoryEntry>> getSimilarEntries(String query, int limit);
+    @Query("SELECT DISTINCT " + SEARCH + " FROM " + TABLE_NAME
+            + " WHERE " + SEARCH + " LIKE :query || '%'" + ORDER_BY_CREATION_DATE + " LIMIT :limit")
```
litetex, line 49: `` || ``

Shouldn't that be an "OR" - we are in SQL not Java :D

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -744,11 +741,11 @@ public boolean onBackPressed() {
                 .getRelatedSearches(query, similarQueryLimit, 25)
                 .toObservable()
                 .map(searchHistoryEntries -> {
-                    final Set<SuggestionItem> result = new HashSet<>(); // remove duplicates
-                    for (final SearchHistoryEntry entry : searchHistoryEntries) {
-                        result.add(new SuggestionItem(true, entry.getSearch()));
+                    final ArrayList<SuggestionItem> suggestionItems = new ArrayList<>();
```
litetex, line 744: Maybe use java lambda(s) here.

Example:
```java
return searchHistoryEntries.stream()
    .map(entry -> new SuggestionItem(true,entry))
    .collect(Collectors.toList());
```

app/src/main/java/org/schabi/newpipe/database/history/dao/SearchHistoryDAO.java
Stypox: That's string concatenation

litetex: Oh wow I didn't know that.
Normal SQL-languages use "CONCAT" or "+" for that but sqlite seems to be special in that case...

app/src/main/java/org/schabi/newpipe/local/history/HistoryRecordManager.java
```diff
@@ -244,9 +244,9 @@ private boolean isStreamHistoryEnabled() {
                 .subscribeOn(Schedulers.io());
     }
 
-    public Flowable<List<SearchHistoryEntry>> getRelatedSearches(final String query,
-                                                                 final int similarQueryLimit,
-                                                                 final int uniqueQueryLimit) {
+    public Flowable<List<String>> getRelatedSearches(final String query,
+                                                     final int similarQueryLimit,
+                                                     final int uniqueQueryLimit) {
```
XiangRongLin, line 249: Do you want to try adding a test for this? Similiar to `LocalPlaylistManagerTest.kt`

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
Stypox: Done

app/src/main/java/org/schabi/newpipe/local/history/HistoryRecordManager.java
Stypox: Done, and I also found a bug, thank you for reminding me to write tests :-)

litetex (dismissed):
LGTM;

Tested it - seems to work as expected.

👍 for the UnitTests

app/src/androidTest/java/org/schabi/newpipe/testUtil/DatabaseMocker.kt
```diff
@@ -0,0 +1,32 @@
+package org.schabi.newpipe.testUtil
+
+import androidx.room.Room
+import androidx.test.core.app.ApplicationProvider
+import org.junit.Assert.assertSame
+import org.schabi.newpipe.NewPipeDatabase
+import org.schabi.newpipe.database.AppDatabase
+
+class DatabaseMocker {
```
XiangRongLin, line 9: I find the naming misleading since it's not mocking the database but creating an actual in-memory database.

For me a mock is something where by default the methods return "nothing" or the closest thing to nothing (empty objects) unless one specifically defines what should be returned. This does not apply to the in-memory database.

XiangRongLin (commented):
I haven't looked at the tests themselves.

app/src/androidTest/java/org/schabi/newpipe/testUtil/DatabaseMocker.kt
Stypox: Well, it is mocking NewPipeDatabase.databaseInstance with the in-memory database... How would you call it?

XiangRongLin: I think you have a different understanding from me if what a mock is.
I find this explanation decent https://stackoverflow.com/a/17810004

---

Going with that thread it would be a fake. But calling it `TestDatabase` would be fine too

XiangRongLin: A mock would be like in `ContentSettingsManagerTest`
https://github.com/TeamNewPipe/NewPipe/blob/ad405d9e0b63ac96949953f9fca0c125ee35f3c4/app/src/test/java/org/schabi/newpipe/settings/ContentSettingsManagerTest.kt#L56

Stypox: Done

XiangRongLin (approved):
Tests look good 👍

PR #7538, opened 2021-12-13
Load uploaderUrl when showing Channel Details from Play Queue

ktprograms:
This checks if the uploaderUrl is in the database, if not it gets the
uploaderUrl and puts it in the database. This is similar to the fetching
of uploaderUrl when it doesn't exist done in #6919.

#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Move the code that conditionally fetches the `uploaderUrl` and saves it to the database into a helper class + method.
- Use the `saveUploaderUrlIfNeeded` method in `QueueItemMenuUtil` when selecting `Show Channel Details` in order to load the `uploaderUrl` if it isn't available in the database.

Android Emulator screen recording seems to be broken, I'll do my best to describe the before and after in text.
- Before: When selecting `Show Channel Details` on a video that doesn't have `uploaderUrl` in the database, it will go to the channel fragment, but the `uploaderUrl` will be `null`, so it will cause a network error.
- After: When selecting `Show Channel Details` on a video that doesn't have `uploaderUrl` in the database, it will fetch the `uploaderUrl` and display a loading toast, then open the channel fragment.

Did not find any issues that will be fixed by this PR.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
@@ -1,0 +1,94 @@
+package org.schabi.newpipe.util;
+
+import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
+
+import android.content.Context;
+import android.widget.Toast;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
+
+import org.schabi.newpipe.NewPipeDatabase;
+import org.schabi.newpipe.R;
+import org.schabi.newpipe.error.ErrorInfo;
+import org.schabi.newpipe.error.ErrorUtil;
+import org.schabi.newpipe.error.UserAction;
+import org.schabi.newpipe.extractor.stream.StreamInfoItem;
+import org.schabi.newpipe.player.playqueue.PlayQueueItem;
+
+import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
+import io.reactivex.rxjava3.schedulers.Schedulers;
+
+/**
+ * Utility class for putting the uploader url into the database - when required.
+ */
+public final class SaveUploaderUrlHelper {
+    private SaveUploaderUrlHelper() {
+    }
+
+    // Public functions which call the function that does
+    // the actual work with the correct parameters
+    public static void saveUploaderUrlIfNeeded(@NonNull final Fragment fragment,
+                                               @NonNull final StreamInfoItem infoItem,
+                                               @NonNull final SaveUploaderUrlCallback callback) {
+        saveUploaderUrlIfNeeded(fragment.requireContext(),
+                infoItem.getServiceId(),
+                infoItem.getUrl(),
+                infoItem.getUploaderUrl(),
+                callback);
+    }
+    public static void saveUploaderUrlIfNeeded(@NonNull final Context context,
+                                               @NonNull final PlayQueueItem queueItem,
+                                               @NonNull final SaveUploaderUrlCallback callback) {
+        saveUploaderUrlIfNeeded(context,
+                queueItem.getServiceId(),
+                queueItem.getUrl(),
+                queueItem.getUploaderUrl(),
+                callback);
+    }
+
+    /**
+     * Fetches and saves the uploaderUrl if it is empty (meaning that it does
+     * not exist in the video item). The callback is called with either the
+     * fetched uploaderUrl, or the already saved uploaderUrl, but it is always
+     * called with a valid uploaderUrl that can be used to show channel details.
+     *
+     * @param context       Context
+     * @param serviceId     The serviceId of the item
+     * @param url           The item url
+     * @param uploaderUrl   The uploaderUrl of the item, if null or empty, it
+     *                      will be fetched using the item url.
+     * @param callback      The callback that returns the fetched or existing
+     *                      uploaderUrl
+     */
+    private static void saveUploaderUrlIfNeeded(@NonNull final Context context,
+                                         final int serviceId,
+                                         @NonNull final String url,
+                                         // Only used if not null or empty
+                                         @Nullable final String uploaderUrl,
+                                         @NonNull final SaveUploaderUrlCallback callback) {
+        if (isNullOrEmpty(uploaderUrl)) {
+            Toast.makeText(context, R.string.loading_channel_details,
+                    Toast.LENGTH_SHORT).show();
+            ExtractorHelper.getStreamInfo(serviceId, url, false)
+                    .subscribeOn(Schedulers.io())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(result -> {
+                        NewPipeDatabase.getInstance(context).streamDAO()
+                                .setUploaderUrl(serviceId, url, result.getUploaderUrl())
+                                .subscribeOn(Schedulers.io()).subscribe();
+                        callback.onCallback(result.getUploaderUrl());
+                    }, throwable -> ErrorUtil.createNotification(context,
+                            new ErrorInfo(throwable, UserAction.REQUESTED_CHANNEL,
+                                    "Could not load channel details")
+                    ));
+        } else {
+            callback.onCallback(uploaderUrl);
+        }
+    }
+
+    public interface SaveUploaderUrlCallback {
+        void onCallback(@NonNull String uploaderUrl);
+    }
+}
--- a/app/src/main/java/org/schabi/newpipe/QueueItemMenuUtil.java
+++ b/app/src/main/java/org/schabi/newpipe/QueueItemMenuUtil.java
@@ -14,6 +14,7 @@ import org.schabi.newpipe.local.dialog.PlaylistDialog;
 import org.schabi.newpipe.player.playqueue.PlayQueue;
 import org.schabi.newpipe.player.playqueue.PlayQueueItem;
 import org.schabi.newpipe.util.NavigationHelper;
+import org.schabi.newpipe.util.SaveUploaderUrlHelper;
 
 import java.util.Collections;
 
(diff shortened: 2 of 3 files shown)


litetex:
Detected a ``TODO``: https://github.com/TeamNewPipe/NewPipe/pull/7538/files#diff-212f79b4111281cdfbffed5f79bca8922f99be14642ef90d524e6c9de9a436f7R61

Assuming this PR is not yet in a reviewable state and converted it to a draft.

ktprograms:
@litetex This TODO was from the previous PR (#6919), and it isn't critical to operation of this functionality.

I can't figure out how to open the Error Activity, can you either give an example code or link to some code that does this?

litetex:
@ktprograms 
> I  can't figure out how to open the Error Activity, can you either give an example code or link to some code that does this?

Basically take a look at the ``ErrorUtil`` class 
Stypox refactored that recently, you may also checkout #7482; There are some usage examples 😄

ktprograms:
@litetex Unfortunately, I'm not sure there's a good way to show the error.

Using `ErrorUtil` opens the full blown error reporting activity, which isn't necessary since most of the time it would be a network error which can be retried.

I tried using ErrorPanelHelper, but that doesn't work because this part of the code doesn't have access to a Fragment.

I also can't use `ErrorUtil.showSnackbar` because it can't seem to find the correct rootView.

litetex:
@Stypox 
> I also can't use `ErrorUtil.showSnackbar` because it can't seem to find the correct rootView.

Do you have an idea how to fix this?

Stypox:
@ktprograms I don't know if it would be possible to open a snackbar in that case. In order to open a snackbar you need a `Context` which is `instanceof Activity`, or a `Fragment`. Only in those cases (and if the activity/fragment has a valid `rootView`, which it usually has) can the snackbar be shown. But since you call `showSnackbar` with a "random" `Context` that might not be `instanceof Activity`, it won't work. Note that `Fragment.getContext()` does not necessarily return a `Context` which is `instanceof Activity`, only `Fragment.getActivity()` does, if `!= null`. So if you might need to change the methods you created a bit, in order to make sure not to pass `fragment.getContext()` to `showSnackbar` but pass the original `fragment` instead. Also, you should ensure that the `context` in `QueueItemMenuUtil` is ` instanceof Activity`; if is not always such, also change `QueueItemMenuUtil.openPopupMenu` accordingly.

ktprograms:
@Stypox I'll look more into this tomorrow, but what do you mean by `just use an error notification`? Do you mean just standard Android notifications or something else?

ktprograms:
Would it be better to merge this first as it fixes the bug that the channel details aren't properly loaded, and move the error handling to another PR? (BTW the TODO was from my first PR about `Show Channel Details`)

Stypox:
@ktprograms just use `ErrorUtil.createNotification(context, errorInfo)`

ktprograms:
> replace `ErrorUtil.showSnackbar(context, errorInfo)` with `ErrorUtil.createNotification(context, errorInfo)`

Done. Ready for review, please.

opusforlife2:
> Done. Ready for review, please.

Then this doesn't need to be a draft anymore.

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
```diff
@@ -0,0 +1,75 @@
...
+import org.schabi.newpipe.error.ErrorInfo;
+import org.schabi.newpipe.error.ErrorUtil;
+import org.schabi.newpipe.error.UserAction;
+import org.schabi.newpipe.extractor.stream.StreamInfoItem;
+import org.schabi.newpipe.player.playqueue.PlayQueueItem;
+
+import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
+import io.reactivex.rxjava3.schedulers.Schedulers;
+
+public final class SaveUploaderUrlHelper {
```
litetex, line 21: Some documentation here would be appreciated.

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
```diff
@@ -0,0 +1,81 @@
...
+
+import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
+import io.reactivex.rxjava3.schedulers.Schedulers;
+
+/**
+ * saveUploaderUrlIfNeeded will check if the uploaderUrl is present in the StreamInfoItem
+ * or PlayQueueItem, and if so will immediately call the callback with the uploaderUrl.
+ * Otherwise, it will use the ExtractorHelper to get the uploaderUrl and save it to the
+ * database for future use, then call the callback with the uploaderUrl.
+ */
```
ktprograms, line 26: I'm not very good at writing documentation, can someone please help to reword this if needed?

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
ktprograms: Draft done, probably needs rewording, see my comment.

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
litetex: I would have gone with something simple like:
``Utility class for putting the uploader url into the database - when required.``

ktprograms: Fixed in https://github.com/TeamNewPipe/NewPipe/pull/7538/commits/607d9959faaa440ceb83e4e57a133d577223da3f

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
```diff
@@ -0,0 +1,78 @@
...
+                                               final PlayQueueItem queueItem,
+                                               final SaveUploaderUrlCallback callback) {
+        saveUploaderUrlIfNeeded(context,
+                queueItem.getServiceId(),
+                queueItem.getUrl(),
+                queueItem.getUploaderUrl(),
+                callback);
+    }
+
+    private static void saveUploaderUrlIfNeeded(final Context context,
```
Stypox, line 49: Also add a javadoc here, explaining that this function loads fetches the uploaderUrl if the item does not already have it, and in that case saves it to the db. Anyway the callback is always called with a valid channel url.

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
ktprograms: Ok, I'll do that.

> Anyway the callback is always called with a valid channel url.

What does that mean?

Stypox: I mean, in the javadoc explain that the callback is always called with a valid uploader url (unless an error has occurred). From the name of the function it is unclear that the function fetches the uploader url if it is not available.

ktprograms: Done in https://github.com/TeamNewPipe/NewPipe/commit/973fd8f6baf1beff4dcf6525205de2bd933138b7

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
```diff
@@ -0,0 +1,84 @@
...
+                                               final PlayQueueItem queueItem,
+                                               final SaveUploaderUrlCallback callback) {
+        saveUploaderUrlIfNeeded(context,
+                queueItem.getServiceId(),
+                queueItem.getUrl(),
+                queueItem.getUploaderUrl(),
+                callback);
+    }
+
+    /**
```
ktprograms, line 49: @Stypox @litetex How's this documentation?

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
```diff
@@ -0,0 +1,84 @@
...
+     * Fetches and saves the uploaderUrl if it is empty (meaning that it does
+     * not exist in the video item). The callback is called with either the
+     * fetched uploaderUrl, or the already saved uploaderUrl, but it is always
+     * called with a valid uploaderUrl that can be used to show channel details.
+     */
```
ktprograms, line 54: @Stypox @litetex How's this documentation?

ktprograms: Superseded by https://github.com/TeamNewPipe/NewPipe/commit/1e652b159ea8618acbd101fbc75983d08588efe9

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
```diff
@@ -0,0 +1,94 @@
...
+     * called with a valid uploaderUrl that can be used to show channel details.
+     *
+     * @param context       Context
+     * @param serviceId     The serviceId of the item
+     * @param url           The item url
+     * @param uploaderUrl   The uploaderUrl of the item, if null or empty, it
+     *                      will be fetched using the item url.
+     * @param callback      The callback that returns the fetched or existing
+     *                      uploaderUrl
+     */
```
ktprograms, line 64: @Stypox @litetex How's this documentation?

ktprograms:
How would I fix this warning: https://sonarcloud.io/project/issues?id=TeamNewPipe_NewPipe&issues=AX6PN5ztwvdSy9w-T0sr&open=AX6PN5ztwvdSy9w-T0sr&pullRequest=7538 ? It is in the else branch of `isNullOrEmpty` so it can't be null, but how do I indicate that to the analysis?

XiangRongLin:
> How would I fix this warning: https://sonarcloud.io/project/issues?id=TeamNewPipe_NewPipe&issues=AX6PN5ztwvdSy9w-T0sr&open=AX6PN5ztwvdSy9w-T0sr&pullRequest=7538 ? It is in the else branch of `isNullOrEmpty` so it can't be null, but how do I indicate that to the analysis?

That does not seem to be possible for you: https://community.sonarsource.com/t/sonarqube-does-not-detect-null-check-by-a-method-and-showing-possible-nullpointerexception/28809
I marked it as false positive for now

ktprograms:
On Tue Jan 25, 2022 at 5:18 PM +08, XiangRongLin wrote:
> I marked it as false positive for now

Ok, thanks!

app/src/main/java/org/schabi/newpipe/util/SaveUploaderUrlHelper.java
litetex: LGTM

Stypox (approved):
Code looks good, thank you! I tested and it works well (even though now the problem with `QueueItemMenuUtil` is already disappeared thanks to #7036 which makes sure items with full stream info to be enqueued).

ktprograms:
Thanks!

PR #7050, opened 2021-09-03
Rework feed new items handling

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
Follow up of #6686

Changes:
* Fixed the weird duplicate FeedSuccessEvents as seen in https://github.com/TeamNewPipe/NewPipe/pull/6686#issuecomment-907840843 and https://github.com/TeamNewPipe/NewPipe/pull/6686#issuecomment-909575283 by using a ``Maybe`` instead of ``Flowable``
* New items are now highlighted with a border in the background
grafik
* If new items got fetched a "New feed items" button pops up
grafik
  * Clicking the button scrolls to the top of the list and hides it
  * The button automatically vanishes after 10s (when device animations are on)
  * The button automatically vanishes when scrolling to the top
  * The button automatically vanishes if the feed is reloaded

#### Before/After Screenshots/Screen Record
<details><summary>Before (click to expand)</summary>

https://user-images.githubusercontent.com/40789489/125986894-f4e7c510-f11a-4a08-bfce-a1153db2ebc6.mp4

</details>
<details><summary>After (click to expand)</summary>

https://user-images.githubusercontent.com/40789489/132061184-fed0c633-911f-491d-90db-276728e6b91b.mp4

</details>

#### Fixes the following issue(s)
- Fixes #6650 indirectly

#### APK testing 
The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

#### How to test
* Install the apk / Checkout the branch
* **optional** Set your phone / emulator system time to 2021-07-15 at 23:55 (you might need to reset it as the clock counts)
* Import an old state of your feed, you can e.g. use this: [NewPipeData-20210716_190308.zip](https://github.com/TeamNewPipe/NewPipe/files/6832340/NewPipeData-20210716_190308.zip)
* Go to Feed/"What's New" and refresh it

PR #7570, opened 2021-12-22
Refactor generating InfoItemDialog's

TobiGr:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

I stumbled about this while doing some other code improvements and did not like the old structure for a few reasons. Most importantly, separation of concerns was neglected and the code was not easy to understand.

This PR refactors the way `InfoItemDialog`s are generated. This is necessary, because the old way used  the `StreamDialogEntry` enum for most of the dialogs' content generation process. This required static variables and methods to store the entries which are used for the dialog to be build (See e.g.`enabledEntries` and methods like `generateCommands()`). In other words, `StreamDialogEntry` wasn't an enumeration anymore.

To address this issue, a `Builder` is introduced for the `InfoItemDialog`'s genration. 

A second problem that introduced a structure which was atypical for an enumeration was the usage of non-final attributes within `StreamDialogEntry` instances. These were needed, because the default actions needed to overriden in some cases.

To address this problem, the `StreamDialogEntry` enumeration was renamed to `StreamDialogDefaultEntry` and a new `StreamDialogEntry` class is used instead.

#### ToDo
- [ ] Add a few more comments
- [ ] Use consistent names for methods that create the dialogs within the Fragments.

#### APK testing 

#### Due diligence


TobiGr:
Why is the `StreamDialogDefaultEntry.OPEN_IN_BROWSER` entry only used in some dialogs and not in all?

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
```diff
@@ -1,54 +1,296 @@
...
 
-    public InfoItemDialog(@NonNull final Activity activity,
-                          @NonNull final StreamInfoItem info,
-                          @NonNull final String[] commands,
-                          @NonNull final DialogInterface.OnClickListener actions) {
-        this(activity, commands, actions, info.getName(), info.getUploaderName());
-    }
+/**
+ * Dialog for a {@link StreamInfoItem}.
+ * The dialog'S content are actions that can be performed on the {@link StreamInfoItem}.
```
litetex, line 27:
```suggestion
 * The dialog's content are actions that can be performed on the {@link StreamInfoItem}.
```
Minor typo

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
```diff
@@ -1,54 +1,296 @@
...
         } else {
             detailsView.setVisibility(View.GONE);
         }
 
+        final String[] items = entries.stream()
+                .map(entry -> entry.getString(activity)).toArray(String[]::new);
+
+        final DialogInterface.OnClickListener action = (d, index) ->
+            entries.get(index).action.onClick(fragment, info);
+
```
litetex, line 64: optional: I think those could also be done inline.

litetex (commented):
LGTM, some minor things I found while doing a quick review:

litetex:
Did a quick test (searched something and clicked through the long-press dropdown) and almost everything seems to work as expected.
However when I try to mark a video as watched it isn't marked as watched immediately and you have to search the same term again. Not sure if this was so before but I think this should be fixed (maybe in a new issue/PR if it was so before).

Stypox:
> However when I try to mark a video as watched it isn't marked as watched immediately and you have to search the same term again. Not sure if this was so before but I think this should be fixed (maybe in a new issue/PR if it was so before).

This never worked for search lists, and after (I think) #7050 it stopped working in the feed fragment, too. Could you look into it?

app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
```diff
@@ -344,55 +337,14 @@ protected void onScrollToBottom() {
         }
     }
 
-    protected void showStreamDialog(final StreamInfoItem item) {
+    protected void showInfoItemDialog(final StreamInfoItem item) {
         final Context context = getContext();
         final Activity activity = getActivity();
         if (context == null || context.getResources() == null || activity == null) {
             return;
         }
```
Stypox, line 345: This whole function could be removed completely, also in other places where opening the info item dialog uses a single line and is used in only one place. The `null` checks should be moved into the `create()` method of `InfoItemDialog.Builder`, and an error notification should be shown in case anything is `null`.

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
```diff
@@ -1,54 +1,296 @@
...
+    /**
+     * <p>Builder to generate a {@link InfoItemDialog}.</p>
+     * Use {@link #addEntry(StreamDialogDefaultEntry)}
+     * and {@link #addAllEntries(StreamDialogDefaultEntry...)} to add options to the dialog.
+     * <br>
+     * Custom actions for entries can be set using
+     * {@link #setAction(StreamDialogDefaultEntry, StreamDialogEntry.StreamDialogEntryAction)}.
+     */
+    public static class Builder {
```
Stypox, line 84: In the Javadoc I'd explain that it can only be used to show a dialog for `StreamInfoItem`s, not just any `InfoItem`

app/src/main/java/org/schabi/newpipe/local/feed/FeedFragment.kt
```diff
@@ -271,7 +267,7 @@ class FeedFragment : BaseStateFragment<FeedState>() {
 
     override fun onDestroyView() {
         // Ensure that all animations are canceled
-        feedBinding.newItemsLoadedButton?.clearAnimation()
+        feedBinding.newItemsLoadedButton.clearAnimation()
```
Stypox, line 270: Is this what was fixed in #7648?
Remember to use what is now on `dev` when rebasing

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
```diff
@@ -1,54 +1,296 @@
...
+            this.context = context;
+            this.fragment = fragment;
+            this.infoItem = infoItem;
+            this.addDefaultEntriesAutomatically = addDefaultEntriesAutomatically;
+            if (addDefaultEntriesAutomatically) {
+                addDefaultBeginningEntries();
+            }
+        }
+
+        public void addEntry(@NonNull final StreamDialogDefaultEntry entry) {
```
Stypox, line 183: This is a `Builder`, so I think all of its function should be made so that they `return this`, so that method calls can be chained

app/src/main/java/org/schabi/newpipe/util/StreamDialogDefaultEntry.java
```diff
@@ -0,0 +1,165 @@
...
+ *     when the entry is selected (via <code>onClick()</code>).
+ *     <br/>
+ *     They action can be overridden by using the Builder's
+ *     {@link org.schabi.newpipe.info_list.InfoItemDialog.Builder#setAction(
+ *     StreamDialogDefaultEntry, StreamDialogEntry.StreamDialogEntryAction)}
+ *     method.
+ * </p>
+ */
+public enum StreamDialogDefaultEntry {
+    SHOW_CHANNEL_DETAILS(R.string.show_channel_details, (fragment, item) -> {
```
Stypox, line 48: Also remember to rebase upon #7538

app/src/main/java/org/schabi/newpipe/util/StreamDialogDefaultEntry.java
```diff
@@ -0,0 +1,165 @@
...
+
+    START_HERE_ON_POPUP(R.string.start_here_on_popup, (fragment, item) ->
+            NavigationHelper.playOnPopupPlayer(fragment.getContext(),
+                    new SinglePlayQueue(item), true)),
+
+    SET_AS_PLAYLIST_THUMBNAIL(R.string.set_as_playlist_thumbnail, (fragment, item) -> {
+    }), // has to be set manually
+
+    DELETE(R.string.delete, (fragment, item) -> {
+    }), // has to be set manually
```
Stypox, line 100: These should throw `UnsupportedOperationException`

Stypox (changes requested):
Looks mostly good, thank you for this improvement :-)

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
TobiGr: I did not do this, because I wanted to increase the code's readability. If that is done inline, one needs way more time to understand what's going on.

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
TobiGr: I thought about this as well. Returning `this` would be the "proper" implementation of a Builder.
I came to the conclusion that using chained calls makes the code hard to understand while refactoring the code (e.g. see the parts where actions are set `LocalPlaylistFragment` and `StatisticsPlaylistFragment`). For that reason, I did not return `this`. I'd refrain from using chained method calls, but I can still return `this` if you want.

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
TobiGr: I thought about renaming the whole class to `StreamInfoItemDialog` to  make it clear that the dialog only works with `StreamInfoItems` currently.
 On the other hand, we could create more builders to match the missing InfoItems: channels and playlists. Their `held()` methods are not overridden in many cases.

app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
TobiGr: I was also annoyed by the checks, but wasn't sure how to handle them correctly. Showing a snackbar should be good. However, the checks need to be performed in the constructor because the context is used when deciding whether some entries need to be added.

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
Stypox: Ok, no, no need

app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
Stypox: Right

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
litetex: I don't think so.
If you use line breaks and/or add a comment I see no problem there.

For me find one's way inside bloated java-files that has useless variable assignments (*cough* Player.java *cough*) is a way bigger problem.

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
litetex: I find Stypox point still valid:

I think this:
```java
        // set custom actions; all entries modified here have already been added within the builder
        dialogBuilder
                .setAction(
                        StreamDialogDefaultEntry.START_HERE_ON_BACKGROUND,
                        (f, i) -> NavigationHelper.playOnBackgroundPlayer(
                                context, getPlayQueueStartingAt(item), true));
                .setAction(
                        StreamDialogDefaultEntry.SET_AS_PLAYLIST_THUMBNAIL,
                        (f, i) ->
                                changeThumbnailUrl(item.getStreamEntity().getThumbnailUrl()));
                .setAction(
                        StreamDialogDefaultEntry.DELETE,
                        (f, i) -> deleteItem(item));
```
is way more readable than
https://github.com/TeamNewPipe/NewPipe/blob/dbc7218501ea5acaf4f1436b2174f7cc5d1e11c6/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java#L759-L767

because less code to read.

PS: Maybe it should also be called ``registerAction``

app/src/main/java/org/schabi/newpipe/util/StreamDialogDefaultEntry.java
```diff
@@ -0,0 +1,165 @@
+package org.schabi.newpipe.util;
```
litetex, line 1: I'm not sure if this class should belong into util.

app/src/main/java/org/schabi/newpipe/info_list/InfoItemDialog.java
```diff
@@ -1,54 +1,296 @@
...
+            entries.add(entry.toStreamDialogEntry());
+        }
+
+        public void addAllEntries(@NonNull final StreamDialogDefaultEntry... newEntries) {
+            for (final StreamDialogDefaultEntry entry: newEntries) {
+                this.entries.add(entry.toStreamDialogEntry());
+            }
```
litetex, line 190:
```suggestion
            Stream.of(newEntries).forEach(this::addEntry);
```

Java 8 streams API

app/src/main/java/org/schabi/newpipe/util/StreamDialogDefaultEntry.java
TobiGr: I moved all classes related to the InfoItemDialog into an own package.

app/src/main/java/org/schabi/newpipe/info_list/dialog/StreamDialogEntry.java
```diff
@@ -0,0 +1,101 @@
...
+                && item.getDuration() < 0) {
+            // Sparse item: fetched by fast fetch
+            ExtractorHelper.getStreamInfo(
+                    item.getServiceId(),
+                    item.getUrl(),
+                    false
+            )
+                    .subscribeOn(Schedulers.io())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(result -> {
```
Stypox, line 80: Note: the result of this `subscribe` is never added to a `Disposable` which can then be cleared on exit by the fragment. But don't to anything in this PR, as this problem has always been there.

Maybe we should create an app-wide static disposable (handled by the `App` class) for things like this? It should obviously not be used for things that do not make sense after the fragment related to them is terminated (e.g. loading data into a view, but if the view is destroyed before the loading could finish, the loading should stop, i.e. the disposable generated by the loading process should be disposed). It would instead be used for things that are not related to any fragment/activity but just to the app, such as adding streams to the play queue, opening the channel fragment from the stream info item, ...

Stypox (changes requested):
Almost ready

app/src/main/java/org/schabi/newpipe/info_list/dialog/StreamDialogEntry.java
```diff
@@ -0,0 +1,101 @@
...
+     * This method is required if the info has been fetched
+     * via a {@link org.schabi.newpipe.extractor.feed.FeedExtractor}.
+     * FeedExtractors provide a fast and lightweight method to fetch info,
+     * but the info might be incomplete
+     * (see {@link org.schabi.newpipe.local.feed.service.FeedLoadService} for more details).
+     * @param context
+     * @param item the item which is checked and eventually loaded completely
+     * @param callback
+     */
+    public static void fetchItemInfoIfSparse(@NonNull final Context context,
```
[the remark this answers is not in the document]
TobiGr: I'd leave this for a future PR.

app/src/main/java/org/schabi/newpipe/info_list/dialog/StreamDialogEntry.java
TobiGr: This sounds like a good idea. We need to open an issue for this.

app/src/main/java/org/schabi/newpipe/info_list/dialog/StreamDialogEntry.java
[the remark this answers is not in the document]
Stypox: Ok

app/src/main/java/org/schabi/newpipe/info_list/dialog/StreamDialogEntry.java
Stypox: #7942

app/src/main/java/org/schabi/newpipe/info_list/dialog/StreamDialogEntry.java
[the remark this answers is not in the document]
Stypox: #7941

PR #7592, opened 2021-12-28
Fixed search not accepting key input after closing player overlay

dkramer95:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Fixes issue where Keyboard input was not reflecting in Search bar after closing Player overlay.
    - Changed `moveFocusToMainFragment` to `false` when the bottom sheet is `STATE_HIDDEN`. This should prevent focus loss for `toolbar_search_edit_text` view.

#### Before/After Screenshots/Screen Record

- Before:

https://user-images.githubusercontent.com/6166095/147518824-464a4f22-ca80-43a1-8c1e-779a6d62e129.mp4

- After:

https://user-images.githubusercontent.com/6166095/147518911-b18710d0-4b6b-4608-ab1b-58a2b4ebced3.mp4

#### Fixes the following issue(s)

- Fixes #7490 

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [X] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #7607, opened 2022-01-01
Fixed search not accepting key input after closing player overlay

### LINKED ISSUE #7490: Keyboard doesn't appear on search sometimes
Not original author but same issue, here's my report:

### Checklist
- [x] I checked, did find a duplicate, #7089, but was told to open a new issue instead.
- [x] I have read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

1. Open app after it's been closed
2. Press the search button

### Actual behavior
The first time pressing the button, as well as the emerging text field does not call the software keyboard, closing the search and opening it again works as expected, unless the app has been unfocused, in which case the behavior is not consistent.

### Expected behavior
Keyboard being called every time the search bar is focused.

This bug is not present in Legacy 0.20.8

### Device info

 - Android version/Custom ROM version: Stock Rom 6 & 7, as well as LineageOS 17.1 (Android 11)
 - Device model: YT3-X50F, XT1771 and XT1604 respectively

I have a rooted device and can navigate ADB if that helps.

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Supersedes #7592

Fixes the problem by checking if the corresponding views have focus and only focuses them if that's not the case.

If this should still cause problems we can also limit the method call to Android TV devices.

- Before: 
See #7592
- After:

https://user-images.githubusercontent.com/40789489/147854501-00ebefb4-bc99-4ab4-82ee-239e2dd162be.mp4

- Fixes #7490

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -2208,6 +2208,7 @@ public final class VideoDetailFragment
             ((ViewGroup) requireView()).setDescendantFocusability(blockDescendants);
             // Only focus the mainFragment if the mainFragment (e.g. search-results)
             // or the toolbar (e.g. Textfield for search) don't have focus.
+            // This was done to fix problems with the keyboard input, see also #7490
             if (!mainFragment.hasFocus() && !toolbar.hasFocus()) {
                 mainFragment.requestFocus();
             }
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -2206,12 +2206,19 @@ public final class VideoDetailFragment
             mainFragment.setDescendantFocusability(afterDescendants);
             toolbar.setDescendantFocusability(afterDescendants);
             ((ViewGroup) requireView()).setDescendantFocusability(blockDescendants);
-            mainFragment.requestFocus();
+            // Check if the mainFragment (search-results) or the toolbar (Textfield for search)
+            // have focus. If so then don't focus them again.
+            if (!mainFragment.hasFocus() && !toolbar.hasFocus()) {
+                mainFragment.requestFocus();
+            }
         } else {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -2206,12 +2206,19 @@ private void moveFocusToMainFragment(final boolean toMain) {
             mainFragment.setDescendantFocusability(afterDescendants);
             toolbar.setDescendantFocusability(afterDescendants);
             ((ViewGroup) requireView()).setDescendantFocusability(blockDescendants);
-            mainFragment.requestFocus();
+            // Check if the mainFragment (search-results) or the toolbar (Textfield for search)
+            // have focus. If so then don't focus them again.
+            if (!mainFragment.hasFocus() && !toolbar.hasFocus()) {
+                mainFragment.requestFocus();
+            }
```
XiangRongLin, line 2213: I find that these types of comments are more confusing than helpful, since they are simply describing what the following line of code does. In this case it's just a simple if expression.
Here I was confused because the comment is basically an inversion of the whole expression.
``` java
if (mainFragment.hasFocus() || toolbar.hasFocus()) {
    // do nothing
} else {
    mainFragment.requestFocus();
}
```
Same for the one below

---

Adding the reason for that line (this specific bug) on the other hand would help a lot in the case that it ever can be removed.

XiangRongLin (commented):
Otherwise looks good

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
litetex: Thank you for the hint.
Should be fixed now: Made the comments easier to understand.

---

> Adding the reason for that line (this specific bug) on the other hand would help a lot in the case that it ever can be removed.

I hope this should also be fixed now with the new comments.
In case of emergency people can also use ``git blame`` :)

XiangRongLin (approved):
I did not test it out myself, but from the gif it works.

PR #7568, opened 2021-12-22
Add Already playing in background toast

vhouriet:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Added "Already playing in background" toast when clicking "Backgound" instead of not showing anything at all.

#### Before/After Screenshots/Screen Record

- Before:

https://user-images.githubusercontent.com/89147152/147089684-4e7cfe6f-d071-4ede-ad33-1f384ff5b6e6.mp4

- After:

https://user-images.githubusercontent.com/89147152/147089402-ce3fec30-dda6-4a28-9f1a-3e211f1dedab.mp4

#### Fixes the following issue(s)

- This was suggested here : https://github.com/TeamNewPipe/NewPipe/pull/7224#issuecomment-991843605

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #7224, opened 2021-10-11
Check if background player already active before displaying player toast

vhouriet:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

"Playing in background" toast no longer appear when clicking background if the player is already playing in the background.

#### Before/After Screenshots/Screen Record

- Before:

https://user-images.githubusercontent.com/89147152/136834112-bba872d8-3486-447b-a65e-c40e4255e4d3.mp4

- After:

https://user-images.githubusercontent.com/89147152/136834117-456e5277-dcf2-474b-8afa-7a725264c734.mp4

#### Fixes the following issue(s)

This was suggested here : https://github.com/TeamNewPipe/NewPipe/pull/7202#pullrequestreview-775578365 

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #7614, opened 2022-01-03
Revert #7568

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

@TiA4f8R noticed that #7568 contains a bug:
"the app shows Already playing in background if we switch contents by pressing the background button on the new content when it should show Playing in background"

For now I reverted the PR because better show no information instead of showing a false one 😄

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
@@ -157,9 +157,8 @@ public final class NavigationHelper {
             return;
         }
 
-        if (PlayerHolder.getInstance().getType() != PlayerType.POPUP) {
-            Toast.makeText(context, R.string.popup_playing_toast, Toast.LENGTH_SHORT).show();
-        }
+        Toast.makeText(context, R.string.popup_playing_toast, Toast.LENGTH_SHORT).show();
+
         final Intent intent = getPlayerIntent(context, MainPlayer.class, queue, resumePlayback);
         intent.putExtra(Player.PLAYER_TYPE, MainPlayer.PlayerType.POPUP.ordinal());
         ContextCompat.startForegroundService(context, intent);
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
```diff
@@ -168,14 +168,10 @@ public static void playOnPopupPlayer(final Context context,
...
-                PlayerHolder.getInstance().getType() == PlayerType.AUDIO
-                    ? R.string.background_player_already_playing_toast
-                    : R.string.background_player_playing_toast,
-                Toast.LENGTH_SHORT)
-                .show();
-
+        if (PlayerHolder.getInstance().getType() != MainPlayer.PlayerType.AUDIO) {
+            Toast.makeText(context, R.string.background_player_playing_toast, Toast.LENGTH_SHORT)
+                    .show();
+        }
```
Stypox, line 174: I'd just remove the `if` and call it a day. This way when starting to play in the background (which could or could not replace an already playing queue, but we don't care) a toast is always shown.
```suggestion
        Toast.makeText(context, R.string.background_player_playing_toast, Toast.LENGTH_SHORT)
                .show();
```

Stypox (changes requested):
Actually, I just noticed my suggestion is equivalent to reverting both #7568 and #7224, i.e. go back to the previous way of showing the toast, i.e. always ;-)

app/src/main/java/org/schabi/newpipe/util/NavigationHelper.java
litetex: Sounds good.

but than it would differ from the popup toast. 
@Stypox Should we fix this also here?

Stypox: I'd say yes

litetex: Done

Stypox (approved):

PR #4833, opened 2020-11-07
YouTube's Fast Forward/Rewind behavior

vkay94:
<p align="center">
<br>
<a href="https://streamable.com/e/b3eeym">Video Preview</a>
</p>

#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
This PR adds multi-double-tap seeking with YouTube's progress indicator style and a small change to alpha animation handling (dynamic duration depending on the view's alpha value).

I've decided to open this PR to discuss it here even if it isn't merge-ready instead of in the issue since the code base is basically done when #4587 was merged and I've reduced the added code to the minimum.

#### Features
- more polished design with fluid transitions
- indicator which shows the current progress
- multi double-tap-to-seek design that allows seeking faster in one flow

#### To-do
- Apply specific handling for popup-videos due to its variable sizes (triangle icon size and text size)

#### Fixes the following issue(s)
- closes #4264 
- closes #1073

#### Relies on the following changes
In the moment there's a design "bug" which results in heavy UI lagging during fast double tapping in some fragments. The reason is that with each seeking the history and playlist fragment is reloaded. Since this feature increases the frequency such high reloading is noticeable.

It could be fixed if the lists fragments apply changes to single items instead of reloading the whole list. But that should be done in another PR, probably by changing to the Groupie library.

#### APK testing (2020-12-01)
When you try it, please make sure you haven't opened the history or playlist detail fragments opened (also in tabs).
[Releases page](https://github.com/vkay94/NewPipe/releases/tag/YT-Rewind-Forward)

Here are some visuals for the 2020-12-01-version with its states:

<p align="center">
<br>
<em>Normal video frame for reference</em>
</p>

<p align="center">
<br>
<em>Playing + dark arc shadow</em>
</p>

<p align="center">
<br>
<em>Paused + controls were visible (shadow + dark transparent arc)</em>
</p>

So let's continue discussing it here with the view of the code implementation ;)

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #3488, opened 2020-04-25
feat(add-rtl-layout-support): drawer, tabs, list items & about

Royosef:
#### What is it?
- [ ] Bug fix (user facing)
- [x] Feature (user facing)
- [ ] Code base improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

feat(add-rtl-layout-support): drawer, tabs, list items & about
* supportsRtl="true"
* replace Left/ Right properties with Start/ End
* tabs:
	* layoutDirection="ltr" for <TabsLayout>
	* reverse items for RTL in PagerAdapter
* properties for force right in RTL:
	* textAlignment="viewStart"
	* textDirection="anyRtl"

image

image

image

image

image

image

image

fixes https://github.com/TeamNewPipe/NewPipe/issues/3882

#### Testing apk

[NewPipe_AddRtlLayoutSupport-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/4533982/NewPipe_AddRtlLayoutSupport-debug.zip)

#### Agreement
- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #7652, opened 2022-01-13
Add RTL layout support

### LINKED ISSUE #3882: RTL UI for rtl languages such as ckb,ku,ar,urdu...etc
Hey
Can we have RTL UI when we will change the language to an RTL language? three dash menu goes to righ side and texts also.

Best Regards.

mhmdanas:
#### What is it?
- [x] Feature (user facing)

Add RTL layout support to NewPipe.

Based on https://github.com/TeamNewPipe/NewPipe/pull/3488 by @Royosef (thank you for your work!). Differences:
- Fixed merge conflicts
- Removed RTL handling from `AboutActivity`, since ViewPager2 supports RTL natively
- Other minor refactors

Issues to fix:

- [ ] Dialog action buttons should be reversed in order
- [ ] Broken with #4833
- [x] ~~fast-forward/reverse seek in player (double taps) are not reversed accordingly~~ seekbar isn't flipped anymore, so this isn't necessary (it might be necessary to investigate the player more closely in general)

| Before | After |
| ------ | ----- |
| Subscriptions before | Subscriptions after |
| Drawer before | Drawer after |
| Comments before | Comments after |
| Player before | Player after |
| About us before | About us after |
| Licenses before | Licenses after |

- Fixes #3882

#### APK testing 

#### Due diligence

**NOTE**: SQUASH MERGE


mhmdanas:
Regarding this suggestion from https://github.com/TeamNewPipe/NewPipe/pull/3488:

> I'd merge LocalizeLayoutUtils (used to know if the device should be RTL and one method around that) and AndroidTvUtils to DeviceUtils.

I don't think we should merged `LocalizeLayoutUtils` to `DeviceUtils`. A device can not be RTL; a locale _can_ be RTL. Also, if the user chooses a language in NewPipe different from their system, then it's not really related to their device anymore.

opusforlife2:
> It doesn't make sense to call a device RTL

Indeed.

mhmdanas:
The CI is failing because of this line:
https://github.com/TeamNewPipe/NewPipe/blob/dd9772cde20535fba1ff0372895795c91d1212f3/app/src/androidTest/java/org/schabi/newpipe/local/history/HistoryRecordManagerTest.kt#L71

On first look this doesn't seem to be related to the PR, but no such failure is present on `dev`, so either the test is flaky, or I've messed up something.

mhmdanas:
So the only issue I found so far with my PR is this:
App in Arabic but system in English

Both app and system in Arabic

In both screenshots, NewPipe's language is Arabic, but in the first one, the system language is English, while in the second it's Arabic. As you can see, the buttons are in different positions. I've checked with some other apps (e.g. Tasks) and they don't seem to exhibit this issue (i.e. they behave as the second screenshot when in RTL).

Do you think this is a blocker?

mhmdanas:
If I want to always use LTR in certain places, is it better to use `left` and `right`, or should I use `start` and `end` but with `android:layoutDirection="ltr"`?

Stypox:
@mhmdanas probably `left` and `right` so that only the things that you specify are affected, otherwise (I think) `android:layoutDirection="ltr"` would propagate to all attributes and children. Unless you just want the latter ;-)

mhmdanas:
The player is too complicated for me to touch in this PR, so I'll try to leave it as-is for now, and perhaps open an issue when this PR is merged.

mhmdanas:
I don't think I'll be picking this up, so I'll be closing it (though I'd be happy to see someone else finish up the work).

PR #7661, opened 2022-01-15
Increase playlist stuck target duration coefficient and catch BehindLiveWindowExceptions properly

TiA4f8R:
#### What is it?
- [x] Bugfix (user facing)

Set a bigger delay before throwing a [`PlaylistStuckException`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/source/hls/playlist/HlsPlaylistTracker.PlaylistStuckException.html) (`15 * TARGET_DURATION_OF_A_SEGMENT`) than ExoPlayer's delay ([`3.5 * TARGET_DURATION_OF_A_SEGMENT`](https://github.com/google/ExoPlayer/blob/r2.14.2/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/DefaultHlsPlaylistTracker.java#L721-L723)), to improve streaming experience on livestreams.

Indeed, I noticed that sometimes, YouTube HLS manifests of livestreams are being stuck sometimes randomly, for a few seconds, and sometimes very frequently (I don't know if that's a YouTube throttling or not), where livestreams on the website/biggest official clients are playing mostly fine. For low or ultralow latency livestreams, this stucking results right now in a exception which is thrown very quickly (sometimes after 1 or 2 second), which makes the livestream unplayable.

This PR also fixes a bug when ExoPlayer encounter a [`BehindLiveWindowException`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/source/BehindLiveWindowException.html): the play queue manager was reloaded to try to reload the stream and then an error was thrown. It makes no sense to do this, so this PR makes the player seek instead to the default position, preparing again ExoPlayer (like recommended by [ExoPlayer's website](https://exoplayer.dev/live-streaming.html#behindlivewindowexception-and-error_code_behind_live_window)) and don't throw an error.

No issue I found was opened for these bugs.

#### APK testing 

#### Due diligence


Redirion:
generally this looks good to me but we need to keep track of those "copy" classes somewhere. In the event that ExoPlayer gets updated we would also have to check if have to update / replace / remove these copy classes as well.

TiA4f8R:
Yes, unfortunately, that's the only disadvantage of this, but I don't think we have a choice if we want to improve ExoPlayer's default behavior.

litetex:
I also currently don't like the PR as it introduces a bunch of copy-paste code... jada jada see above comments

Wouldn't it be better if we implement the change into ``ExoPlayer`` directly / open a PR by them?
The PR could either 
* make all methods / fields protected so that they can be overridden (by us) or
* add an option for the MAXIMUM_PLAYLIST_STUCK_DURATION_MS or something similar

We could also fork Exoplayer and build a custom version with the fix. However that might be to complex (e.g. the building part).

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2516,50 +2516,75 @@ public void onPlayerError(@NonNull final ExoPlaybackException error) {
...
-            errorInfo = new ErrorInfo(error, UserAction.PLAY_STREAM,
-                    "Player error[type=" + error.type + "] occurred, currentMetadata is null");
-        } else {
-            errorInfo = new ErrorInfo(error, UserAction.PLAY_STREAM,
-                    "Player error[type=" + error.type + "] occurred while playing "
-                            + currentMetadata.getMetadata().getUrl(),
-                    currentMetadata.getMetadata());
-        }
-        ErrorUtil.createNotification(context, errorInfo);
+        boolean isBehindLiveWindowException = false;
```
litetex, line 2519: I think the following would be easier to understand:

```Java
        boolean couldErrorBeFixed = false;

        switch (error.type) {
            case ExoPlaybackException.TYPE_SOURCE:
                couldErrorBeFixed = processSourceError(error.getSourceException());
                break;
            case ExoPlaybackException.TYPE_UNEXPECTED:
                setRecovery();
                reloadPlayQueueManager();
                break;
            case ExoPlaybackException.TYPE_REMOTE:
            case ExoPlaybackException.TYPE_RENDERER:
            default:
                onPlaybackShutdown();
                break;
        }

        if (couldErrorBeFixed) {
            return;
        }

        createErrorNotification(error); // In theory we also don't need this method anymore - you might also remove it

        if (fragmentListener != null) {
            fragmentListener.onPlayerError(error);
        }
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2516,50 +2516,75 @@ public void onPlayerError(@NonNull final ExoPlaybackException error) {
...
         setRecovery();
 
         if (error instanceof BehindLiveWindowException) {
-            reloadPlayQueueManager();
+            simpleExoPlayer.seekToDefaultPosition();
+            simpleExoPlayer.prepare();
+            // Inform the user that we are reloading the stream by switching to the buffering state
+            onBuffering();
+            return true;
         } else {
```
litetex, line 2585: The else is no longer required (return)

opusforlife2:
> Wouldn't it be better if we implement the change into ExoPlayer directly / open a PR by them?

Let's please do this. It will be ironic and hilarious.

Redirion:
what I just noticed: it should already be customizable. See this commit:
https://github.com/google/ExoPlayer/commit/28434cff4c11b0e19a3145bfbaff6c9507a574b1

TiA4f8R:
Yes, but we need to pass an `HlsMediaSource.Factory`. The only problem is that the factory uses the constructor with the default playlist stuck target duration coefficient value. So what I should do?

app/src/main/java/org/schabi/newpipe/player/Player.java
TiA4f8R: Indeed, it will also reflect that we are not limited to one ExoPlayer's non-fatal exception that we can catch.

litetex:
> Yes, but we need to pass an `HlsMediaSource.Factory`. The only problem is that the factory uses the constructor with the default playlist stuck target duration coefficient value. So what I should do?

Uhm I don't see a problem here:
```diff
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
@@ -8,12 +8,17 @@ import com.google.android.exoplayer2.source.ProgressiveMediaSource;
 import com.google.android.exoplayer2.source.SingleSampleMediaSource;
 import com.google.android.exoplayer2.source.dash.DashMediaSource;
 import com.google.android.exoplayer2.source.dash.DefaultDashChunkSource;
+import com.google.android.exoplayer2.source.hls.HlsDataSourceFactory;
 import com.google.android.exoplayer2.source.hls.HlsMediaSource;
+import com.google.android.exoplayer2.source.hls.playlist.DefaultHlsPlaylistTracker;
+import com.google.android.exoplayer2.source.hls.playlist.HlsPlaylistParserFactory;
+import com.google.android.exoplayer2.source.hls.playlist.HlsPlaylistTracker;
 import com.google.android.exoplayer2.source.smoothstreaming.DefaultSsChunkSource;
 import com.google.android.exoplayer2.source.smoothstreaming.SsMediaSource;
 import com.google.android.exoplayer2.upstream.DataSource;
 import com.google.android.exoplayer2.upstream.DefaultDataSourceFactory;
 import com.google.android.exoplayer2.upstream.DefaultLoadErrorHandlingPolicy;
+import com.google.android.exoplayer2.upstream.LoadErrorHandlingPolicy;
 import com.google.android.exoplayer2.upstream.TransferListener;

 public class PlayerDataSource {
@@ -45,7 +50,16 @@ public class PlayerDataSource {
         return new HlsMediaSource.Factory(cachelessDataSourceFactory)
                 .setAllowChunklessPreparation(true)
                 .setLoadErrorHandlingPolicy(
-                        new DefaultLoadErrorHandlingPolicy(MANIFEST_MINIMUM_RETRY));
+                        new DefaultLoadErrorHandlingPolicy(MANIFEST_MINIMUM_RETRY))
+                .setPlaylistTrackerFactory(
+                        (dataSourceFactory, loadErrorHandlingPolicy, playlistParserFactory) ->
+                            new DefaultHlsPlaylistTracker(
+                                    dataSourceFactory,
+                                    loadErrorHandlingPolicy,
+                                    playlistParserFactory,
+                                    /* playlistStuckTargetDurationCoefficient */ 10.0
+                            )
+                        );
     }

     public DashMediaSource.Factory getLiveDashMediaSourceFactory() {
```

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
```diff
@@ -17,9 +18,12 @@
 import com.google.android.exoplayer2.upstream.TransferListener;
 
 public class PlayerDataSource {
+
+    public static final int LIVE_STREAM_EDGE_GAP_MILLIS = 10000;
+
+    private static final double PLAYLIST_STUCK_TARGET_DURATION_COEFFICIENT = 15;
```
litetex, line 24: Ah short documentation what this values does and why it was chosen would be great 😄

litetex (commented):
Almost perfect

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
TiA4f8R: > why it was chosen

It's more like by testing on (ultra)low latency livestreams but on livestreams with an higher segment target duration, the coefficient maybe too big.

What's your advice (and @Redirion's one)?

litetex: > What's your advice

Good question. I'm not really technically into this.

Note: The formula where this is used is:
```
if (currentTimeMs - lastSnapshotChangeMs > C.usToMs(playlistSnapshot.targetDurationUs) * playlistStuckTargetDurationCoefficient) {
   // do throw error
}
```
https://github.com/google/ExoPlayer/blob/029a2b27cbdc27cf9d51d4a73ebeb503968849f6/library/hls/src/main/java/com/google/android/exoplayer2/source/hls/playlist/DefaultHlsPlaylistTracker.java#L728-L733

Do you have any data regarding the values above that might helps choosing a value?

Anyway the current 15 would be ~4,3x greater then the default (3.5) so I think we could say/document "Choose a 4-times greater value than the default to ensure that livestreams which got stuck for a moment don't crash/fail"

opusforlife2:
Importing database into this APK is causing it to close at the splash screen without crashing. This isn't happening on other PR APKs.

TiA4f8R:
See #7674. I just need to rebase all my PRs.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
TiA4f8R: I will use the following JavaDoc:

```java
    /**
     * An approximately 4.3 times greater value than the
     * {@link DefaultHlsPlaylistTracker#DEFAULT_PLAYLIST_STUCK_TARGET_DURATION_COEFFICIENT default}
     * to ensure that (very) low latency livestreams which got stuck for a moment don't crash too
     * early.
     */
```

PR #7670, opened 2022-01-18
Ensure order of search history entries in tests

### LINKED ISSUE #7553: CI pipeline maybe flaky (HistoryRecordManagerTest#deleteSearchHistory)
### Checklist

The CI pipeline failed on the android tests. It fixed itself after rerunning it, which could indicate a flaky test.
This Issue should serve as place to collect the logs and discuss the problem if it occurs more often. If it does not occur again this can be closed.

### Logs
```
org.schabi.newpipe.local.history.HistoryRecordManagerTest > deleteSearchHistory[test(AVD) - 5.0.2] FAILED 
	java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)

> Task :app:connectedDebugAndroidTest FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:connectedDebugAndroidTest'.
```
Log of failed CI job: [logs_5027.zip](https://github.com/TeamNewPipe/NewPipe/files/7740706/logs_5027.zip)
https://github.com/TeamNewPipe/NewPipe/runs/4571396354?check_suite_focus=true

XiangRongLin:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Ensure order of search history entries in tests

- Fixes #7553

#### Due diligence

❗  squash merge

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
   test-android:
     # macos has hardware acceleration. See android-emulator-runner action
     runs-on: macos-latest
+    timeout-minutes: 20
     strategy:
       matrix:
         # api-level 19 is min sdk, but throws errors related to desugaring
--- a/app/src/androidTest/java/org/schabi/newpipe/local/history/HistoryRecordManagerTest.kt
+++ b/app/src/androidTest/java/org/schabi/newpipe/local/history/HistoryRecordManagerTest.kt
@@ -25,9 +25,6 @@ class HistoryRecordManagerTest {
     @get:Rule
     val trampolineScheduler = TrampolineSchedulerRule()
 
-    @get:Rule
-    val timeout = Timeout(1, TimeUnit.SECONDS)
-
     @Before
     fun setup() {
         database = TestDatabase.createReplacingNewPipeDatabase()
--- a/app/src/androidTest/java/org/schabi/newpipe/local/history/HistoryRecordManagerTest.kt
+++ b/app/src/androidTest/java/org/schabi/newpipe/local/history/HistoryRecordManagerTest.kt
@@ -12,7 +12,9 @@ import org.schabi.newpipe.database.AppDatabase
 import org.schabi.newpipe.database.history.model.SearchHistoryEntry
 import org.schabi.newpipe.testUtil.TestDatabase
 import org.schabi.newpipe.testUtil.TrampolineSchedulerRule
+import java.time.LocalDateTime
 import java.time.OffsetDateTime
+import java.time.ZoneOffset
 import java.util.concurrent.TimeUnit
 
 class HistoryRecordManagerTest {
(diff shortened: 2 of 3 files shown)


litetex (commented):
LGTM, however the same problem might also happen here:
https://github.com/TeamNewPipe/NewPipe/blob/4a7cfd1a6cc63a95140c03a937c9f607303e1a57/app/src/androidTest/java/org/schabi/newpipe/local/history/HistoryRecordManagerTest.kt#L147

Could you fix it also there? 
A constant would be great 😄

XiangRongLin:
> however the same problem might also happen here:

How would it also happen there? every line has the `.minusSeconds(x)`

litetex:
@XiangRongLin 

The problem here is that the value can't be controlled by the test.
``OffsetDateTime.now()`` is not a constant, if there is e.g. a leap second or system clock update between the two lines in https://github.com/TeamNewPipe/NewPipe/blob/4a7cfd1a6cc63a95140c03a937c9f607303e1a57/app/src/androidTest/java/org/schabi/newpipe/local/history/HistoryRecordManagerTest.kt#L147-L148 the test might fail.

I'm strongly in favor of implementing a hard-coded/current time independent timestamp like
```java
OffsetDateTime.of(LocalDateTime.of(2000, 1, 1, 1, 1), ZoneOffset.UTC)
```
and then use that one for the tests.

litetex:
There seems to be something wrong with the test on Emulator API 29:
```
org.schabi.newpipe.local.history.HistoryRecordManagerTest > onSearched[test(AVD) - 10] FAILED 
	org.junit.runners.model.TestTimedOutException: test timed out after 1 seconds
	at org.assertj.core.internal.Strings.instance(Strings.java:118)
```

not sure what's causing that but according to https://stackoverflow.com/questions/17016011/junit-test-times-out-despite-executing-quickly and 
grafik
 it has has be an ``InterruptedException`` inside
https://github.com/TeamNewPipe/NewPipe/blob/3c21be8fa5d66383d2fb1b2d4be28d2e409b2929/app/src/androidTest/java/org/schabi/newpipe/local/history/HistoryRecordManagerTest.kt#L44-L53

litetex:
Above error should be fixed with https://github.com/TeamNewPipe/NewPipe/pull/7670/commits/a6515d5450d4eac4b85b9644ea3dfa9d7140a49c.
There seems to be a performance problem with GH actions which causes the Android API 29 test do not run fast enough (needs more than 1s) on the VM which causes the above problems.

The internal was - for a unknown reason - changed from 10s to 1s inside https://github.com/TeamNewPipe/NewPipe/commit/7d6688f4976d1c998938843f5c29d5db8c0ea7fc#diff-52b3c5a209220512cb671827511b090b467afb544abcd94da776e7273072ffc0R24

litetex (approved):
LGTM

PR #7678, opened 2022-01-21
Fix some gradle deprecations

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Update gradle wrapper 7.3.1 -> 7.3.3
- Update Checkstyle 9.2 -> 9.2.1 ([changelog](https://checkstyle.sourceforge.io/releasenotes.html))
- Fixed some well-hidden deprecations (found [here](https://developer.android.com/reference/tools/gradle-api/7.0/classes))

#### APK testing 
Shouldn't be necessary.

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -8,14 +8,14 @@ plugins {
 }
 
 android {
-    compileSdkVersion 30
+    compileSdk 30
     buildToolsVersion '30.0.3'
 
     defaultConfig {
         applicationId "org.schabi.newpipe"
         resValue "string", "app_name", "NewPipe"
-        minSdkVersion 19
-        targetSdkVersion 29
+        minSdk 19
+        targetSdk 29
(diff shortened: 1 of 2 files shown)


mhmdanas:
> Update gradle wrapper 7.3.1 -> 7.3.3

Just to clarify, you're upgrading Gradle, not the Gradle wrapper.

Stypox (approved):
Gradle changes look good according to the documentation, though it's strange that they deprecate things without giving an explanation and an alternative method themselves. The checkstyle update also looks good, as there have only been internal bugfixes in `9.2.1`: https://checkstyle.sourceforge.io/releasenotes.html. The app builds just fine, both here and on my pc.

TacoTheDank:
Thanks :)

PR #7679, opened 2022-01-22
Update ACRA library

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Update ACRA 5.7.0 -> 5.8.4 ([changelog](https://github.com/ACRA/acra/releases))

#### APK testing 

You need to deliberately break the app to test this, ironically.

#### Due diligence


app/src/main/java/org/schabi/newpipe/App.java
```diff
@@ -212,7 +212,7 @@ protected void initACRA() {
 
         try {
             final CoreConfiguration acraConfig = new CoreConfigurationBuilder(this)
-                    .setBuildConfigClass(BuildConfig.class)
+                    .withBuildConfigClass(BuildConfig.class)
                     .build();
```
Redirion, line 216: from the docs:
http://www.acra.ch/docs/Setup

instead of calling build() you can also directly provide the CoreConfigurationBuilder to ACRA.init

I would say we should take this opportunity to init it compliant to the docs

Redirion:
```
            final CoreConfigurationBuilder builder = new CoreConfigurationBuilder(this)
                    .withBuildConfigClass(BuildConfig.class);
            ACRA.init(this, builder);
```

TacoTheDank: Actually we can't do that because then `ACRAConfigurationException` isn't thrown. Should we get rid of the try catch block?

mhmdanas:
> You need to deliberately break the app to test this, ironically.

Off-topic, but it would be neat if we added an option to trigger an error in a debug APK.

Stypox:
> Off-topic, but it would be neat if we added an option to trigger an error in a debug APK.

It already exists, Settings -> Debug -> Crash the app

mhmdanas:
@Stypox ooh, nice, I didn't know that.

app/src/main/java/org/schabi/newpipe/App.java
Redirion: well sounds good to me if we don't need the try catch anymore

TacoTheDank: Okay.

TacoTheDank:
I just realized I opened the PR with the wrong branch name 🤦 doesn't really matter I guess.

Redirion (approved):
lgtm. I guess you did a quick check whether the app still works?

Issue #7427, opened 2021-11-20
Video progress resets when switching to popup/background and viceversa


Dangelman:
# Workaround: https://github.com/TeamNewPipe/NewPipe/issues/7427#issuecomment-1014762780

### Checklist

- [x] I am using the latest version - 0.21.13 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug
1. Play a video
2. Watch if for some minutes in portrait mode
3. Switch to popup or background

### Actual behavior

The video progress will reset to either the start or the video or the last timestamp you paused it. The opposite is also true (especially switching from background to main player). I have auto-rotate disabled, if it can help.

### Expected behavior

The video should progress from the same point (or 1-2 seconds before).

### Screenshots/Screen recordings

https://github.com/TeamNewPipe/NewPipe/issues/7427#issuecomment-97482348
(change resolution if video freezes)

### Device info

 - Android version/Custom ROM version: 11
 - Device model: LM-V500EM (LG V50 ThinQ 5G)

mhmdanas:
Sounds like a duplicate of #7303 to me, but I'm not exactly sure what the reproduction steps in your issue are since you didn't fill out that section, so please do that. :)

Dangelman:
> Sounds like a duplicate of #7303 to me, but I'm not exactly sure what the reproduction steps in your issue are since you didn't fill out that section, so please do that. :)
mb, I forgot to delete the comment brackets.

It should be the same as #7303, tho the author did not mention the problem happening to popup or the possibility for the video to resuming from the last pause timestamp instead of restarting.

opusforlife2:
@Dangelman Could you provide the exact steps? As I've written in the issue, it takes 2 switches for the bug to happen for me. Your steps seem to contain only one switch. Something is missing here.

Dangelman:
@opusforlife2 It's as I described.
1. Open NewPipe
2. Select a new video
3. Play the video in portrait mode for 30s
4. Switch to either popup/background

The video restarts instead of continuing, or jump back to the last timestamp you paused previously.

Dangelman:
https://streamable.com/dsae9q
[...]

PR #7693, opened 2022-01-24
Fix first item always played in the play queue when reloading play queue manager

### LINKED ISSUE #7531: Resuming app from background plays wrong video from queue
### Checklist

- [x] I am using the latest version - 0.21.14 
- [X] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [X] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [X] This issue contains only one bug. I will open one issue for every bug report I want to file.

1. Create a queue of videos
2. Watch a few videos from the queue
3. Wait til a video starts buffering 
4. Minimize app and let background playback start 
5. App started playing the previous video from the queue 
6. Reopen app 
7. The video before that current starts playing 

### Actual behavior
Newpipe plays videos from that queue that have already been watched 

### Expected behavior
Newpipe should continue the current video instead of jumping to a previous queued one 

The video shows the currently running video. After minimizing the app another video starts ("what's up guys" in the background) and after opening the app again it immediately jumps to another video from my queue. Which you can see in the video again 
https://youtu.be/Tf4jVVlY7l8

### Device info

 - Android version/Custom ROM version: Android 11, Oxygen OS
 - Device model: Oneplus9 Pro

TiA4f8R:
#### What is it?
- [x] Bugfix (user facing)

This PR fixes the playback of the first item in the play queue when a resolution (when switching the quality played in the quality selector) or a player type change occurs (for e.g. switching from background to main player), when playing another item than the first one of the play queue.

The issue was caused by an ExoPlayer change, which when setting a new media source, resets now the current playback position and the current window index to the first item and its default position by default (the position was then set with the first item's recovery position).

It sets now to ExoPlayer to not reset the position when setting a new media source.

- Fixes #7531

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -2048,7 +2048,7 @@ public final class Player implements
         if (currentState == STATE_BLOCKED) {
             changeState(STATE_BUFFERING);
         }
-        simpleExoPlayer.setMediaSource(mediaSource);
+        simpleExoPlayer.setMediaSource(mediaSource, false);
         simpleExoPlayer.prepare();
     }
 


litetex (changes requested):
#7427 was already fixed.
What are the ``player.setRecovery`` blocks doing inside this PR?

TiA4f8R:
See https://github.com/TeamNewPipe/NewPipe/issues/7427#issuecomment-985465982: the player recovery position needs to be set when clicking on the content thumbnail (so playback position is not lost when you used background or popup player (by pressing the background button) and you are now using main player).

Other additions are for cases like this one: you put the app with a content playing in the main player on the background, so background player is used while the player is not visible and you came back to the player with Android's media notification. Playback position was not saved and the content starts to play again from the previous playback position saved (I was able to reproduce this once, now it seems I can't).

litetex:
@TiA4f8R 
And how is that related to #7531?

Please one bugfix at the time if they aren't directly linked together.
We could have fixed #7531 by now if these were 2 PRs.

It would be best if you limit this PR to the fixes of #7531 and create a new PR for the additional fixes of #7427.

Btw: I think the workaround for #7427 is incomplete because in grafik the workaround should also be applied to ``overlay_metadata_layout`` and ``overlay_buttons_layout``

TiA4f8R:
> the workaround should also be applied to `overlay_metadata_layout` and `overlay_buttons_layout`

In fact, no, because there is no need to set player recovery when pressing an element inside the metadata part of the description or when pressing other buttons than popup and background buttons.

> It would be best if you limit this PR to the fixes of #7531 and create a new PR for the additional fixes of #7427.

Ok, if you want.

Stypox (approved):
According to the [documentation](https://github.com/google/ExoPlayer/blob/029a2b27cbdc27cf9d51d4a73ebeb503968849f6/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayer.java#L1164-L1172) what you did is correct, and that's the only place where `setMediaSource` is used.
I tested on API 19 emulated, API 31 emulated and API 29 device and the fix worked (i.e. I could reproduce the bug before, but with the same steps now I can't).

litetex (approved):
LGTM

Nice and simple 👍

Issue #5558, opened 2021-02-11
Remove Duplicate Entries From Queue 


Zakkumaru:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want
A button or a passive feature in the options to have all duplicate entries in the playlist removed

#### Is your feature request related to a problem? Please describe it
My problem is related to bug https://github.com/TeamNewPipe/NewPipe/issues/5557 causing several duplicate entries in Playing Queue

#### How will you/everyone benefit from this feature?
1. Trying to create master playlists with hundreds of songs to shuffle, it becomes increasingly more difficult to tell which ones have been added to the list
2. Would be useful for creating a playlist out of multiple playlists without having duplicate entries.  ie: "Nonvocal Liquid Bass" and "Beethoven Remixes" could be combined for whatever study music you're suddenly in the mood for, but if they had overlapping interests (and thus, some of the same songs), you would have to manually remove those entries in order to not have them in the list multiple times.  This could be especially bothersome, again, if you had a lit of songs in each list.

(To be clear, I'm referring to songs of the same ID -- removing entries with similar titles would be a different issue)

AlexN1997:
Have you looked at #5334 ?

Zakkumaru:
> Have you looked at #5334 ?

I didn't notice it, although I had tried searching for something similar.  I would consider this a separate feature since the features apply to preexisting entries, as opposed to checking one at a time at the time of adding them to the list.  Plus, #5334 wouldn't apply retroactively.

onelharrison:
I can try implementing this functionality.

ShareASmile:
Fixed in [v0.25.1 ](https://github.com/TeamNewPipe/NewPipe/issues/9870) by #9707

Issue #5557, opened 2021-02-11
Cleared Playlist Doesn't Stay Cleared


Zakkumaru:
### Checklist
- [x] I am using the latest version - 0.20.8 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug
1. Queue a song
2. Go to the notification dropdown and click in a blank space of the song currently playing
3. This brings up the "Now Playing..." list
4. Stop the song
5. Erase the notification
6. Swipe away all of the songs currently in the playlist
7. Add the songs you want
8. Close the app
9. Reopen the app
10. Queue a song

### Actual behaviour
 - All of the erases songs will be back, sometimes even duplicated multiple times
 - Same behavior when simply erasing all songs instead of queuing more before closing the app
 - Same behavior when leaving the app open for a while before closing
 - Same behavior if the notification isn't closed

### Expected behavior
All erased songs should be gone, leaving only the ones previously queued before the last session was closed

Stypox:
This was fixed in 0.24.1 (https://github.com/TeamNewPipe/NewPipe/pull/8706). Please confirm whether that's true, and if it's not feel free to reopen the issue ;-)

opusforlife2:
@Stypox They can't reopen it if you close the issue. They can only comment.

tovam:
@Stypox I have 0.24.1 and sadly the issue is still not fixed here

Stypox:
Ok, then I don't understand what the issue is. Could you provide a video of the problem? I don't understand what "Swipe away all of the songs currently in the playlist" means, since you never mentioned that a playlist needs to be opened.

Issue #4186, opened 2020-08-25
Show a warning before adding a duplicate video to a playlist


AnonymousSeal:
#### Describe the feature you want

It would be nice to be alerted when trying to add a video to a playlist that the playlist already contains.

#### How will you/everyone benefit from this feature?

It prevents you from accidentally adding a video to a playlist twice. Especially helpful for long playlists.

Edit: adding keywords for searching: duplicate/entry/entries/playlist/playlists/videos

opusforlife2:
Closing #1473 in favour of this. Alerting the user seems more useful than removing after the fact.

ahmad-abdelbaset:
Hi, I would like to work on this issue as a part of my project to Master Degree course (Software Construction), I discussed it with my instructor @atamrawi

Stypox:
@ahmad-abdelbaset great! Go for it, nobody else has taken it up so far. Thank you :-)

ShareASmile:
@Stypox  I think there should { or would be great } to have an option to Both
1. {disallow adding twice the same video to a list} 
2. and {to batch remove all previous duplicates} from video playlists.

As Batch remove duplicates from playlist would greatly help in eliminating duplicates from prevously made playlists by users (especially longer ones) from which manual removal is most difficult.
.. discussion from #1473

Stypox:
@ShareASmile @ahmad-abdelbaset yeah, I agree, there should be:
- a new "Remove duplicates" button in the local playlist page's three-dot menu (i.e. in the same place as "Remove watched")
- a small indicator next to each local playlist in the "Add to playlist" dialog meaning that the stream being added is already in there
- a confim dialog that opens when a user tries to add a stream to a local playlist that already contains it

ahmad-abdelbaset:
Thank you @Stypox and @ShareASmile, it is really good idea, I will own this one.
@atamrawi

ahmad-abdelbaset:
Hey, 
I made a pull request (#5334 ) which is showing a message tells the user if it was any duplicated video/s before and asking him if he like to add it once again. If the user accept, the video will be added. if he reject the video will not be added. @atamrawi

I didn't have time to add the feature of "Remove Duplicated", sorry.

infinitewaveparticle:
Any updates on implementing this?

SameenAhnaf:
@infinitewaveparticle PR #5334 is in progress.

infinitewaveparticle:
[...]

PR #7703, opened 2022-01-25
Implement removal of duplicates from local playlists

onelharrison:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- #5558

Before

Before

After

After1

On click "Remove duplicates" menu item, the appropriate popup menu is displayed
After2

On click "Yes", duplicates in playlist are removed
After3

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
@@ -55,8 +55,10 @@ import org.schabi.newpipe.util.StreamDialogEntry;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -99,6 +101,8 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
     private AtomicBoolean isModified;
     /* Is the playlist currently being processed to remove watched videos */
     private boolean isRemovingWatched = false;
+    /* Is the playlist currently being processed to remove duplicate streams */
+    private boolean isRemovingDuplicateStreams = false;
 
     public static LocalPlaylistFragment getInstance(final long playlistId, final String name) {
         final LocalPlaylistFragment instance = new LocalPlaylistFragment();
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -367,12 +371,83 @@ public boolean onOptionsItemSelected(final MenuItem item) {
...
+
+        disposables.add(playlistManager.getPlaylistStreams(playlistId)
+                .subscribeOn(Schedulers.io())
+                .map((List<PlaylistStreamEntry> playlist) -> {
+                    // Playlist data
+                    final Iterator<PlaylistStreamEntry> playlistIter = playlist.iterator();
+
+                    // Remove duplicates, Functionality data
+                    final List<PlaylistStreamEntry> uniquePlaylistItems = new ArrayList<>();
+                    final Set<Long> uniquePlaylistItemStreamIds = new HashSet<>();
```
mhmdanas, line 407: Hash sets are unordered, right? Wouldn't that be an issue here, or am I mistaken?

onelharrison: Hash sets are unordered, but it doesn't present a problem here because it's only being used to check if a stream has already been seen. The List preserves the order of the steams in the playlist.

litetex:
Problems I saw while quickly taking a look at this PR:
1. I don't think the function will be used very often. The PR implements a workaround for the bugs #5557 and #4186 - I would like to see fixes for these bugs instead because they are the primary cause for duplicates in a playlist.
2. The code should be refactored out of the UI if possible. Also using the Java streaming API would be great.
3. I don't think it's a good idea to compare thumbnails to determine if a video is a duplicate
4. I think running the de-duplication on database level and then refreshing/reloading the UI would be the most performant solution here

Stypox:
About 4.: I agree, and a query like this should work to get the items. I took part of [this query](https://github.com/TeamNewPipe/NewPipe/blob/0113ad5e1490c483622dc7f4870ec2e18f9b8077/app/src/main/java/org/schabi/newpipe/database/playlist/dao/PlaylistStreamDAO.java#L56, added a `GROUP BY` to group items with the same stream id together (which also solves 3.) and then only keeping the minimum join index among each group (which means the first instance of the video in the playlist is kept).
```sql
// get ids of streams of the given playlist
"SELECT " + JOIN_STREAM_ID + ", MIN(" + JOIN_INDEX + ")"
+ " FROM " + PLAYLIST_STREAM_JOIN_TABLE
+ " WHERE " + JOIN_PLAYLIST_ID + " = :playlistId"
+ " GROUP BY " + JOIN_STREAM_ID
```

litetex:
@Stypox 

I don't think you need a query.
Something like
```SQL
DELETE 
FROM playlist_stream_join
WHERE playlist_id = <playlistId>
AND stream_id IN (
    SELECT stream_id
    FROM playlist_stream_join
    WHERE playlist_id = <playlistId>
    GROUP BY stream_id
    HAVING COUNT(*) > 1
)
AND join_index NOT IN (
    SELECT join_index
    FROM playlist_stream_join
    WHERE playlist_id = <playlistId>
    GROUP BY stream_id
    HAVING COUNT(*) > 1
)
```
should also be able to de-duplicate the data.
However the join_index should probably be rebuilt after that.

Stypox:
#5557 has been fixed separately in #8706, so this PR shouldn't need to fix it. Anyway, I think it's better to implement #4186 first, which I guess is simpler and more user-friendly (less cluttering options under hidden menus). Then if we see there is still need for removing duplicates from playlists, we can always implement that, too. Hence I am closing this PR, @onelharrison if you wish to work on #4186 feel free to do so :-)

Iron-E:
> I think it's better to implement https://github.com/TeamNewPipe/NewPipe/issues/4186 first, which I guess is simpler and more user-friendly (less cluttering options under hidden menus).

Unless I'm misunderstanding: as I said in https://github.com/TeamNewPipe/NewPipe/pull/7703#issuecomment-1281530474, even if #4186 is implemented, playlists with pre-existing duplicates won't be affected. It would help _future_ playlists, but nothing that already has duplicates, which is still a separate issue.

Stypox:
Well, in the long term playlists with already some duplicates would not be a problem anymore

Issue #7097, opened 2021-09-12
Improve description for videos


SameenAhnaf:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

Links in description covers too much space on Newpipe app sometimes. Links should be shortened just like YouTube app.

<details><summary> Newpipe app layout </summary>

![IMG_20210912_133624](https://user-images.githubusercontent.com/79245641/132978160-7d493a1e-ac92-442f-afb8-d693d8dae423.jpg)

</details>

<details><summary> YouTube app layout </summary>

![IMG_20210912_133754](https://user-images.githubusercontent.com/79245641/132978176-8b4cbb98-be79-4e95-bd75-d5d18fca75d6.jpg)

</details>

In my opinion, long press on any text should copy the whole description just like how we see in comments. Long press on any link should copy only a single link. (I wouldn't have mentioned about copying description in a separate issue. But copying shortened links might be buggy. So, I am willing to mention it here.)

mhmdanas:
Can you please post the link of the video?

SameenAhnaf:
@mhmdanas Here you go. https://youtu.be/mEw05znACp0

OmlineEditor:
I have a similar problem. I have Russian in the settings, but English text in the description of the videos, as well as English text in the title of the video. if you watch it in a real browser, then the description of the video is in Russian, not in English.

an example where there is a problem: https://www.youtube.com/watch?v=02Q-mWjQemI

PR #7725, opened 2022-01-30
Add long press action on hashtags and web links in descriptions

### LINKED ISSUE #3253: Links in description can't be copied
### Version

- 0.18.7, latest from F-Droid

1. Go to any video which has long links in the description.
2. Links can be clicked on and will open correctly on the browser.
3. Try to copy the link(s) and paste somewhere else.

### Expected behavior
The full link will be copied.

### Actual behaviour
A clipped version of the link is copied.

### Screenshots/Screen records
Can't provide screenshots right now.

### Logs
Not a crash, just a bug, so no logs can be provided.

AudricV:
#### What is it?
- [x] Feature (user facing)

This PR adds a long press action on web links and hashtags which copy them to clipboard.

To do so:

- Some changes in our `TextView` class have been required and metadata items are now using a `NewPipeTextView` instead of a standard `TextView`.

- Three new classes have been added: a custom `LinkMovementMethod` class (adapted from [a StackOverflow answer](https://stackoverflow.com/a/31786969)), a custom `ClickableSpan` class have been added in order to set a long press event and a class to avoid code duplication in `CommentTextOnTouchListener`, `TouchUtils`.

- Before:

https://user-images.githubusercontent.com/74829229/151714340-aa46635c-998a-48ee-9b87-13ba5e83a771.mp4

- After:

https://user-images.githubusercontent.com/74829229/151714470-88379f1f-9fb0-4da3-95f4-950847867725.mp4

- Fixes #3253

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/CommentTextOnTouchListener.java
+++ b/app/src/main/java/org/schabi/newpipe/util/CommentTextOnTouchListener.java
@@ -1,6 +1,7 @@
 package org.schabi.newpipe.util;
 
-import android.text.Layout;
+import static org.schabi.newpipe.util.TouchUtils.getOffsetForHorizontalLine;
+
 import android.text.Selection;
 import android.text.Spannable;
 import android.text.Spanned;
@@ -30,23 +31,9 @@ public class CommentTextOnTouchListener implements View.OnTouchListener {
 
             final int action = event.getAction();
 
-            if (action == MotionEvent.ACTION_UP
-                    || action == MotionEvent.ACTION_DOWN) {
-                int x = (int) event.getX();
-                int y = (int) event.getY();
-
-                x -= widget.getTotalPaddingLeft();
-                y -= widget.getTotalPaddingTop();
-
-                x += widget.getScrollX();
-                y += widget.getScrollY();
-
-                final Layout layout = widget.getLayout();
-                final int line = layout.getLineForVertical(y);
-                final int off = layout.getOffsetForHorizontal(line, x);
-
-                final ClickableSpan[] link = buffer.getSpans(off, off,
-                        ClickableSpan.class);
+            if (action == MotionEvent.ACTION_UP || action == MotionEvent.ACTION_DOWN) {
+                final int offset = getOffsetForHorizontalLine(widget, event);
+                final ClickableSpan[] link = buffer.getSpans(offset, offset, ClickableSpan.class);
 
                 if (link.length != 0) {
                     if (action == MotionEvent.ACTION_UP) {
@@ -58,8 +45,7 @@ public class CommentTextOnTouchListener implements View.OnTouchListener {
                             }
                         }
                     } else if (action == MotionEvent.ACTION_DOWN) {
-                        Selection.setSelection(buffer,
-                                buffer.getSpanStart(link[0]),
+                        Selection.setSelection(buffer, buffer.getSpanStart(link[0]),
                                 buffer.getSpanEnd(link[0]));
                     }
                     return true;
--- a/app/src/main/java/org/schabi/newpipe/util/TouchUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/TouchUtils.java
@@ -1,0 +1,38 @@
+package org.schabi.newpipe.util;
+
+import android.text.Layout;
+import android.view.MotionEvent;
+import android.widget.TextView;
+
+import androidx.annotation.NonNull;
+
+public final class TouchUtils {
+
+    private TouchUtils() {
+    }
+
+    /**
+     * Get the character offset on the closest line to the position pressed by the user of a
+     * {@link TextView} from a {@link MotionEvent} which was fired on this {@link TextView}.
+     *
+     * @param textView the {@link TextView} on which the {@link MotionEvent} was fired
+     * @param event    the {@link MotionEvent} which was fired
+     * @return the character offset on the closest line to the position pressed by the user
+     */
+    public static int getOffsetForHorizontalLine(@NonNull final TextView textView,
+                                                 @NonNull final MotionEvent event) {
+
+        int x = (int) event.getX();
+        int y = (int) event.getY();
+
+        x -= textView.getTotalPaddingLeft();
+        y -= textView.getTotalPaddingTop();
+
+        x += textView.getScrollX();
+        y += textView.getScrollY();
+
+        final Layout layout = textView.getLayout();
+        final int line = layout.getLineForVertical(y);
+        return layout.getOffsetForHorizontal(line, x);
+    }
+}
--- a/app/src/main/java/org/schabi/newpipe/util/external_communication/TextLinkifier.java
+++ b/app/src/main/java/org/schabi/newpipe/util/external_communication/TextLinkifier.java
@@ -2,8 +2,6 @@ package org.schabi.newpipe.util.external_communication;
 
 import android.content.Context;
 import android.text.SpannableStringBuilder;
-import android.text.method.LinkMovementMethod;
-import android.text.style.ClickableSpan;
 import android.text.style.URLSpan;
 import android.text.util.Linkify;
 import android.util.Log;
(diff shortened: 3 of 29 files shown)


litetex (commented):
Note:
* The videos are unplayable (can't view them in Firefox or Chroium-based browsers)
* There are some Sonar warnings

TiA4f8R:
> * The videos are unplayable (can't view them in Firefox or Chromium-based browsers)

I had to compress them, unfortunately, but it seems Video Transcoder converted them in an unplayable format for browsers (you can play them in VLC I think)

SameenAhnaf:
@litetex You may download the video with [Video Download Helper](https://www.google.com/url?sa=t&source=web&rct=j&url=https://addons.mozilla.org/en-US/firefox/addon/video-downloadhelper/&ved=2ahUKEwjc9Padg9r1AhXdILcAHWKKB9gQFnoECAgQAQ&usg=AOvVaw33rEvRz1dzj2d86z1mWTW1). I eventually had to download, the video is still loading on my browser.

@TiA4f8R I included some suggestions in #7097. I think, it'll be more suitable for copying links.

litetex:
Converted the videos from ``mpeg4 (mp4v)`` to ``h264 (avc1)`` (according to ffmpeg). Now they are also 8x smaller. However the resolution with 221x480 is not very good, but at least it's viewable now 😄

TiA4f8R:
So what's the best way to proceed when long pressing links and hashtags?

- copy them directly to clipboard
- show an alert dialog to allow opening/searching them and copying them to clipboard

I am not all a UX designer so feedback is appreciated.

opusforlife2:
@TiA4f8R The usual behaviour for Android apps is to take the specific action when directly tapping, and to copy to clipboard (or something else) when long-pressing.

TiA4f8R:
Ok, I will do this.

ktprograms:
@TiA4f8R Can you please add this functionality to timestamps in the description? (Example video with timestamps: https://www.youtube.com/watch?v=mOSirVeP5lo)

Thanks!

app/src/main/java/org/schabi/newpipe/util/external_communication/TextLinkifier.java
```diff
@@ -244,14 +284,33 @@ private static void changeIntentsOfDescriptionLinks(final TextView textView,
...
-                        if (!InternalUrlsHandler.handleUrlDescriptionTimestamp(
-                                new CompositeDisposable(), context, url)) {
-                            ShareUtils.openUrlInBrowser(context, url, false);
+                final ClickableSpan clickableSpan;
+                if (addLongPressListenerToLinks) {
+                    clickableSpan = new LongPressClickableSpan() {
+                        @Override
+                        public void onClick(@NonNull final View view) {
+                            if (!InternalUrlsHandler.handleUrlDescriptionTimestamp(
+                                    new CompositeDisposable(), context, url)) {
```
Stypox, line 293: Also in other places: creating a `new CompositeDisposable()` defeats the purpose of adding things to a disposable, since disposables should be cleared somewhere when the app/activity/fragment/dialog closes
```suggestion
                                    disposables, context, url)) {
```

Stypox (changes requested):
Looks almost good to me ;-)

AudricV:
> @TiA4f8R Can you please add this functionality to timestamps in the description? (Example video with timestamps: https://www.youtube.com/watch?v=mOSirVeP5lo)
> 
> Thanks!

@ktprograms Done (in the app side) on services on which timestamps supported (otherwise, a fallback to copy timestamp raw text is made).

AudricV:
Note that the Sonar bug is a false positive, as nullity is checked by `Utils.isBlank` method of the extractor as the first instruction of the method and that the TODO that I added, is of course, not faisable in this PR.

app/src/main/java/org/schabi/newpipe/util/CommentTextOnTouchListener.java
```diff
@@ -1,6 +1,7 @@
 package org.schabi.newpipe.util;
```
Stypox, line 1: Can't this be moved to `external_communication`?

app/src/main/java/org/schabi/newpipe/util/TouchUtils.java
```diff
@@ -0,0 +1,38 @@
+package org.schabi.newpipe.util;
```
Stypox, line 1: This, too, should be moved there

app/src/main/java/org/schabi/newpipe/views/LongPressLinkMovementMethod.java
```diff
@@ -0,0 +1,77 @@
+package org.schabi.newpipe.views;
```
Stypox, line 1: And also this I would say, so that we have everything related to sharing, spans, copying to clipboard, ... in a single folder

Stypox (approved):
At this point you should probably create a library for clipboard & sharing handling :joy: 

I tested on API 33 emulated, API 22 emulated and my API 29 phone and everything works well. Comments with links now behave correctly :-D

The few comments I left are really small things: once you solve them, you can merge this PR yourself. Code looks good.

app/src/main/java/org/schabi/newpipe/util/TouchUtils.java
AudricV: I am not agree with your comment: `TouchUtils` may be used to provide in the future more methods which are not used by classes interacting with other apps.

Also, the current content of this class has nothing to do with external communication.

app/src/main/java/org/schabi/newpipe/views/LongPressLinkMovementMethod.java
Stypox: Ok, makes sense

app/src/main/java/org/schabi/newpipe/util/TouchUtils.java
Stypox: Yeah ok, then `TouchUtils` should be moved to the spans package you mentioned below

Stypox (approved):
LGTM, thanks! The bug reported by Sonar is wrong: it says `content` might be null but actually we check for `isBlank`ness above.

yashpalgoyal1304:
> * The videos are unplayable (can't view them in Firefox or Chromium-based browsers)

kinda strange that yeah, it doesn't play on desktop browser ( Vivaldi | 5.5.2805.48 (Stable channel) (64-bit) ; Chrome/106.0.0.0), but is working fine on android browser (Vivaldi 5.6.2868.29)

yashpalgoyal1304:
but i still don't understand the premise of this PR.

like, selecting tags and links was available before this pr as well:
- "Enable selecting text in description"
- press and hold on links to select and copy them
- the only problem was that clicking them caused them to open

This pr shows that press and holding them automatically copies them - one tap saved.

But what happens in this pr when u single tap on links/tags? Did this problem was an effort to solve this single tap problem?

Following is a screenrecord from Newpipe v0.23.3 i.e. back from 2022.08'

https://user-images.githubusercontent.com/19423063/213942070-13371bb7-5819-40d5-a9d6-8518613c1218.mp4

Stypox:
> The videos are unplayable

I don't understand what you are referring to.

> Did this problem was an effort to solve this single tap problem?

What do you mean by "single tap problem"? For me single taps work normally. Also, this PR was also needed because soon we will have comments with copiable links and hashtags.

yashpalgoyal1304:
> I don't understand what you are referring to.

sorry, have updated with link. : https://github.com/TeamNewPipe/NewPipe/pull/7725#pullrequestreview-867148494



> single tap problem"? For me single taps work normall

this is shown in the screen recording - i meant that on single tapping, the links were opening. i was asking that was that considered the problem?



> Also, this PR was also needed because soon we will have comments with copiable links and hashtags.

yeah, that is understandable. the thing that confused me was the following quoted part: _"reopening as we are not able to copy full links"_ which we definitely were.

> Reopening as we are not able to copy full links on YouTube. I will open a PR which will add this ability soon.
\- @ AudricV at https://github.com/TeamNewPipe/NewPipe/issues/3253#issuecomment-1024970127

Stypox:
> which we definitely were

Oh ok, I see what you meant. Yeah, in the description you were already able to copy YouTube links, since those are always shown in full. I am not sure about Peertube though, where Markdown is employed, and `[link](https://example.org)` gets transformed into the literal "[link](https://example.org)". Unless Android is able to copy rich text, I doubt you would be able to copy links from there. So the YouTube description was just a special case among other services' descriptions and also comments.

yashpalgoyal1304:
> about Peertube though, where Markdown is employed,
> Unless Android is able to copy rich text,

ohw ohkayh, yeah, makes sense then :+1:

PR #7728, opened 2022-01-31
Remember adjustment step size for playback controls (speed and pitch)

### LINKED ISSUE #7031: Remember step for playback speed controls
Make playback the step width in percent a sticky option, which saves its last state. Usually it's always reset to 25% if you re-open the option dialog on the video screen.

I usually work with 1,15x average speed for speech channels and 5% step width, with the need to adjust according to content and situation, but I always need to adjust the width first, which is annoying.

ktprograms:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Save `adjustmentStepSize` to shared preferences, pass it back and forth
between PlaybackParameterDialog and Player/PlayQueueActivity.
- Made `DEFAULT_STEP` in PlaybackParameterDialog public so it can be used as
the default `adjustmentStepSize` value in the Player.

| Before | After |
|--|--|
| <video src="https://user-images.githubusercontent.com/71804605/151766501-807ca760-ad0c-494a-a1f3-9c1de85a2b05.mp4"> | <video src="https://user-images.githubusercontent.com/71804605/151766566-86420278-19d6-4bec-96c0-c5433dcca642.mp4"> |

- Fixes #7031

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java
@@ -71,7 +71,6 @@ public class PlaybackParameterDialog extends DialogFragment {
     private double tempo = DEFAULT_TEMPO;
     private double pitch = DEFAULT_PITCH;
     private int semitones = DEFAULT_SEMITONES;
-    private double stepSize = DEFAULT_STEP;
 
     @Nullable
     private SeekBar tempoSlider;
@@ -147,7 +146,6 @@ public class PlaybackParameterDialog extends DialogFragment {
             tempo = savedInstanceState.getDouble(TEMPO_KEY, DEFAULT_TEMPO);
             pitch = savedInstanceState.getDouble(PITCH_KEY, DEFAULT_PITCH);
             semitones = percentToSemitones(pitch);
-            stepSize = savedInstanceState.getDouble(STEP_SIZE_KEY, DEFAULT_STEP);
         }
     }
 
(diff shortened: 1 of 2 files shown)


Stypox (changes requested):
I don't think you should pass around the `adjustmentStepSize` as a playback parameter. It's not something influencing the playback, it's just a value to save to / read from preferences. You should just read it when the `PlaybackParameterDialog` is setup, and save it when the user changes it inside the same dialog.

ktprograms:
@Stypox Sorry, I was busy so I didn't have time to look at the code.

I'm guessing this should be done similarly to how `setupHookingControl` handles accessing the preferences?

https://github.com/TeamNewPipe/NewPipe/blob/af80d96b9ed701ec44c36d1ea8e37c11dbae56f5/app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java#L237-L260

Stypox:
@ktprograms yes, I think that's the way to go

ktprograms:
@Stypox Since the step size will now be saved to shared preferences, should I remove the code that saves it to/reads it from outState?

Stypox:
@ktprograms yes, there is no need to save/restore it there

ktprograms:
@Stypox Done, basically rewrote the entire commit, so here's the overview of everything changed:
- Add `adjustment_step_key` to `settings_keys.xml` to be used when
saving/loading the step size.
- Remove the global `stepSize` variable and the code that saves it to
`outState`/loads it from `savedInstanceState` because it's now saved to
Shared Preferences.
- Move initially setting step size to `setupStepSizeSelector` to be
consistent with the other view setup methods, using the value loaded
from Shared Preferences.
- Save the step size to Shared Preferences inside `setStepSize`.

Stypox (commented):
Thank you! Code looks good to me, I'm going to test it later :-)

Stypox (approved):
I tested and everything seems to work well :-)

ktprograms:
Thanks!

PR #7668, opened 2022-01-17
Workaround: Set recovery before switching player

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
Quick and dirty fix for #7427:
Calls ``setRecovery`` before switching the player.

See also https://github.com/TeamNewPipe/NewPipe/issues/7427#issuecomment-1014852363

#### Fixes the following issue(s)
- Fixes #7427 
⚠️ Please reopen the issue when merging as this is only a workaround or notify @litetex that he creates a followup issue

#### APK testing 
The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #7852, opened 2022-02-10
Set workaround for playback position reset when switching to main player with content thumbnail

TiA4f8R:
#### What is it?
- [x] Bugfix (user facing)

The workaround set with #7668 was not applied when switching to main player with content thumbnail from popup or background player. This PR fixes this by applying the workaround when switching to main player with content thumbnail from popup or background player.

- Fixes https://github.com/TeamNewPipe/NewPipe/issues/7427#issuecomment-985465982

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -500,6 +500,10 @@ public final class VideoDetailFragment
                 break;
             case R.id.detail_thumbnail_root_layout:
                 autoPlayEnabled = true; // forcefully start playing
+                // FIXME Workaround #7427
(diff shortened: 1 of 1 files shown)


tsiflimagas:
I tried it, but it doesn't work. I don't remember what was the case when I made that comment, but I found out that the right `case` is `case R.id.detail_thumbnail_root_layout:`. Sorry for the confusion, I'm not sure why I wrote that back then.

TiA4f8R:
Indeed, the `overlay_thumbnail` id refers to thumbnail of the content opened and/or played in mini player, as you can see in the following screenshot:

overlay_thumbnail_design_android_studio

Thank you for looking to this!

Are my last changes working (I used what you said) or not?

tsiflimagas:
Yes, it works as expected.

TobiGr (approved):

Issue #7731, opened 2022-01-31
Play/Pause buttons are not centered in fullscreen mode / Arc size of double tap gestures is not the same


SameenAhnaf:
### Checklist
- [x] I am using the latest version - dev branch after #4833 merged 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

View a video (with a cosshair) in fullscreen mode.
Or use the double tap gesture.

### Actual behavior

Note that the play/pause button are not centered.
There's a striking difference in arc size. Double tap to rewind tends to smudge with `Pause` icon.

### Expected behavior

The play/pause button are not centered.
Size of the arcs should be the same. Don't smudge the arc with `Pause` icon to keep the UI a little more clean.

### Screenshots/Screen recordings

| Rewind Gesture | Fast Forward Gesture |
| --- | --- |
| <img src="https://user-images.githubusercontent.com/79245641/151863574-e3519933-c293-4c03-9e77-085d1f1fe3f6.jpg" /> | <img src="https://user-images.githubusercontent.com/79245641/151863775-f1c561e5-6c2f-43f6-a1eb-518e16367749.jpg" /> |

Further examples with a crosshair from a minecraft video:
| Demo1 | Demo2 |
| --- | --- |
| ![grafik](https://user-images.githubusercontent.com/40789489/153716352-5fa8aadf-97b8-40e2-a4c6-60cc6e00d6d6.png) | ![grafik](https://user-images.githubusercontent.com/40789489/153716380-9517bf29-b5cb-420c-bc0b-50d4299b74e0.png) |

Portrait mode works as expected:
![grafik](https://user-images.githubusercontent.com/40789489/153716335-6e52cac6-4f9e-4b36-9d63-470d3e7ae04b.png)

litetex:
The arc size is perfect.
The buttons are off-centered (because of the system menu/buttons) inside the full screen mode and that is the problem.

<details><summary>I noticed that while I watched video which have a crosshair, e.g. </summary>

![grafik](https://user-images.githubusercontent.com/40789489/152872194-5d1bd870-19e1-4a19-a8b7-ca051d108d90.png)
[...]

PR #7894, opened 2022-02-18
Remove large-land player layout: not actually used

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

While debugging #7731 I found out that any change I made to the large-land layout of the player, didn't have any effect, which really confused me, since in the fullscreen/landscape player the large-land layout should have been used. But it seems like the normal layout is always used instead, so this PR is an experiment that removes the large-land layout completely. Removing it solves the problem of having 800 lines of duplicated layout code.

The diff between the normal layout and the large-land layout before removing it is below here, if it might come in handy for reviewing. As you can see, there are just some formatting differencies, and maybe a couple `dp` distances are not the same, but nothing big changes. So even if the large-land layout was being used at particular (*unknown*) times, removing it (which causes the normal layout to always be used) shouldn't make much of a difference.
<details><summary><code>diff app/src/main/res/layout/player.xml app/src/main/res/layout-large-land/player.xml</code></summary>

```diff
204c204,205
<                         tools:ignore="ContentDescription,RtlHardcoded" />
---
>                         tools:ignore="ContentDescription,RtlHardcoded"
>                         tools:visibility="visible" />
214c215
<                         android:paddingStart="6dp"
---
>                         android:paddingStart="3dp"
216c217
<                         android:paddingEnd="6dp"
---
>                         android:paddingEnd="3dp"
222c223,224
<                         tools:ignore="ContentDescription,RtlHardcoded" />
---
>                         tools:ignore="ContentDescription,RtlHardcoded"
>                         tools:visibility="visible" />
256a259
>                         android:textAllCaps="false"
496c499
<                 android:layout_height="40dp"
---
>                 android:layout_height="50dp"
511c514
<                 android:layout_height="60dp"
---
>                 android:layout_height="70dp"
522c525
<                 android:layout_height="40dp"
---
>                 android:layout_height="50dp"
539c542
<         android:layout_width="match_parent"
---
>         android:layout_width="380dp"
540a544
>         android:layout_alignParentEnd="true"
597a602
>                 android:src="@drawable/exo_controls_repeat_off"
599d603
<                 app:srcCompat="@drawable/exo_controls_repeat_off"
733d736
<         style="@style/Widget.AppCompat.Button.Borderless"
```

</details>

#### APK testing 

#### Due diligence


Atemu (commented):
This does not affect the tablet mode player in case anyone's wondering.

Stypox:
@B0pol @zubayy @epicincognito since you have a real Android TV and you are *used to using it*, could you test if this PR's apk causes any problem in the player (both fullscreen and not)?

peat80:
NewPipe delete-large-land-player
NewPipe nightly

Seems like the large-land-layout icludes some focus related stuff for android tv. 🤔

peat80:
NewPipe delete-large-land-player_20220220_133941

Same behaviour within fullscreen player. Buttons are not focusable anymore. Other than that the layout looks the same as in nightlies / releases.

Stypox:
Thank you for the feedback, I didn't notice there was some code to handle focus in the large-land layout but not in the normal one. I added such code to the normal one now.

Also, the title and channel views (the ones visible while in fullscreen) were clickable & focusable in the normal layout, but not in the large-land one; I made them not-clickable and not-focusable for everyone, since clicking on them actually did/does nothing. I tested that this does not prevent the text views from automatically scrolling when the title is too long, so the behavior is exactly the same as before.

I updated the diff in the PR description based on the above changes

Other actual things that change, by comparing the two files:
- some buttons are a little bit bigger, some a little bit smaller, but nothing that really has a motivation, so I'd leave it like this
- the play queue in large-land previously used only a part of the screen, now it uses the whole screen; can you tell me if this is bad? I personally prefer it covering the whole screen, so that you can read the whole stream titles. See the two screenshots below for a comparison, I tested on emulated API 30 Android TV.
<table>
  <tr>
    <th></th>
    <th>Before</th>
    <th>After removing large-land</th>
  </tr>
  <tr>
    <td>Non-fullscreen</td>
    <td>image</td>
    <td>image</td>
  </tr>
  <tr>
    <td>Fullscreen</td>
    <td>image</td>
    <td>image</td>
  </tr>
</table>

Atemu:
The spacing on non-fullscreen video and recommends is better on the left while, on the right, the playlist selector is better IMO.

How do the chapters look like? Those should probably be like playlists are on the left.

peat80:
The focusing now works exactly as before. 👌
I also like the play queue using full width more than the actual layout. 😎

While your at it could you also fix some focus / highlighting issues on ATV?

NewPipe delete-large-land-player_20220224_000313
The top row on the play queue could only be highlighted as a whole. So no buttons are acessible with remote.

NewPipe delete-large-land-player_20220224_001010
If you want to change the resolution on the fly the choosen resolution is not highlighted so one has to guess where you are (or count the button presses) to change to the desired resolution. But it works. 

Also on the speedmenu the 2 checkboxes are focusable but they are not highlighted. So another guess (or count). But they become highlighted when you activate / deactivate them. 🤔

These issues are of course not introduced with this pr. They exist in normal builds too. Just thought to mention them here. 
Would be ok if they will not be fixed in this pr but it could maybe fit in here.

litetex:
Thank you all for the feedback.

> These issues are of course not introduced with this pr. They exist in normal builds too. Just thought to mention them here.
Would be ok if they will not be fixed in this pr but it could maybe fit in here.

This PR only removes the large-hand layout and nothing else.
You might open an issue or more for these problems so we can track them (if there aren't already some open).

litetex (approved):
LGTM

I like it when code gets deleted 👍 

Did a quick test on the emulator and couldn't find any problems.

Stypox:
@peat80 thank you for the feedback! I just opened a follow-up PR: #7963

> Did a quick test on the emulator and couldn't find any problems.

I also tested extensively (while creating #7963), so I can confirm everything works fine

PR #7893, opened 2022-02-18
Fix fast seek overlay arc not correctly centered

Stypox:
#### What is it?
- [x] **Regression** bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
This PR makes sure that the fast seek overlay arcs share the same paddings as other controls (such as play-pause, brightness, volume, ...). The paddings are being set by the code and are "caused" by the system navigation bar. Play-pause, brightness, volume, ... buttons were correctly centered in the **app** UI (i.e. all of the screen excluding the navigation bar), while the fast seek overlay was centered in the **system** UI (i.e. the whole screen), causing a mismatch between the two that looked ugly, as seen in #7731. My way of fixing it might not be the best, as the arc is now not drawn under the system navigation bar anymore (see *After* screenshot below), but I just wanted to create a PR, I will try to improve it properly later today or tomorrow (though I think the current state is already acceptable).

#### Before/After Screenshots/Screen Record

| Before | After |
| --- | --- |
|  |  |

#### Fixes the following issue(s)

- Fixes #7731

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #7898, opened 2022-02-19
Have fast seek overlay arc go under system ui

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

In #7893 I added padding (equal to the window insets) to the fast seek overlay just as it was done with the player overlays (play-pause, brightness, ...). But as you can see in the "Before" screenshot below, that caused the fast-seek arc to not go under the system ui. This PR instead applies to the fast seek overlay negative margins, for each side respectively equal to minus the window inset of the opposite side. This way the view covers all of the player (overflowing on some sides, but that's not important) and its center still coincides with the center of other controls (which is a requirement, since we don't want #7731).

| Before | After |
| --- | --- |
|   |   |

#### APK testing 

I tested on API 19 and API 30 and it worked fine.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -580,11 +580,17 @@ public final class Player implements
                             v.getPaddingTop(),
                             v.getPaddingRight(),
                             v.getPaddingBottom());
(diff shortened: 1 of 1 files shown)


litetex (approved):
LGTM

PR #7904, opened 2022-02-19
Solve Java warning "Raw use of parameterized class"

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

I made sure there was no **raw** use of parameterized classes. Raw means without <> when the class requires it.
- Use `<?>` for `CacheData` since the cache map `LRU_CACHE` can contain items cache data of different types
- Replace raw use of `JsonSink` with the actual classes being passed to functions. This may decrease the generality of the functions having such items as parameters, since they would work for any `JsonSynk` but are actually only called with one type of `JsonSink` each time.  Maybe just using `JsonSink<?>` is better? (btw, the definition of `JsonSink` is recursive for some reason...: `public interface JsonSink<SELF extends JsonSink<SELF>>`)
- `BaseListInfoFragment` now has two parameters: the item type to be shown in the list (`I extends InfoItem`), and the list info type  (previously `I extends InfoItemsPage` without `<>`, now renamed to `L extends InfoItemsPage<I>`). This propagated to all subclasses and allowed solving problems in all of them.
- In `AbstractInfoPlayQueue` I turned the raw `ListInfo` in `ListInfo<StreamInfoItem>` (if you look at the class, it assumes the list info provides stream info items, and strange things were already being done to cast). I also removed an unused <> parameter `U` (no idea why that existed, its usage only in an unused constructor made no sense), refactored the constructors to deduplicate code for subclasses and refactored a method to use Java streams.
- In other places, I added the `<Type>` where it was missing
- I refactored some lines in `BaseListFragment` and `StreamItemAdapter`

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
@@ -272,7 +272,7 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I>
             }
         });
 
-        infoListAdapter.setOnChannelSelectedListener(new OnClickGesture<ChannelInfoItem>() {
+        infoListAdapter.setOnChannelSelectedListener(new OnClickGesture<>() {
             @Override
             public void selected(final ChannelInfoItem selectedItem) {
                 try {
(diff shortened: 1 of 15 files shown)


litetex (approved):
LGTM

PR #7910, opened 2022-02-19
Better equals check

avently:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

It ensures that queues are not the same. Without this check when you have multiple videos in the backstack and navigating back via Back button you'll get duplicated videos

Don't know. To hard to search for "duplicate" keyword:)

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
@@ -532,8 +532,11 @@ public abstract class PlayQueue implements Serializable {
             return false;
         }
         for (int i = 0; i < size(); i++) {
-            // Check is based on URL
-            if (!streams.get(i).getUrl().equals(other.streams.get(i).getUrl())) {
+            final PlayQueueItem stream = streams.get(i);
+            final PlayQueueItem otherStream = other.streams.get(i);
+            // Check is based on serviceId and URL
+            if (stream.getServiceId() != otherStream.getServiceId()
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
```diff
@@ -528,7 +528,16 @@ public boolean equals(@Nullable final Object obj) {
             return false;
         }
         final PlayQueue other = (PlayQueue) obj;
-        return streams.equals(other.streams);
+        if (size() != other.size()) {
+            return false;
+        }
+        for (int i = 0; i < size(); i++) {
+            // Check is based on URL
+            if (!streams.get(i).getUrl().equals(other.streams.get(i).getUrl())) {
```
Stypox, line 536: I know that the url should technically be enough to uniquely identify a stream, but in other places when comparing streams the service id is also used, so I'd add it.

Stypox (changes requested):
Other than the small comment I made, this looks good to me

app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
avently: Added this check

avently:
You can merge this PR if it's ok, then I'll merge dev into #6895 and it will make testing of #6895 much easier

Stypox (approved):
Thank you! I'm going to merge this in the afternoon, after having created a separate branch on dev for an almost-hotfix release 0.22.1

Issue #6927, opened 2021-08-15
Video player slow loading on first tap


swarup13:
### Checklist

- [x] I am using the latest version - 0.21.8  
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo, It is a completely different issue, please read before marking as closed
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

1. Go to ' What's New/Trending page'
2. Press on 'Any video'

### Actual behavior
It is taking too much time to load a video,  with the contents like description etc. Sometimes I have to close the window or app to play the video. Sometimes it doesn't even load the player.
See it's not related to buffering during already playing video, or not even related to Internet. 

### Expected behavior
The expected behaviour is fast video player loading

### Screenshots/Screen recordings

![Screenshot_20210815-154029864](https://user-images.githubusercontent.com/55395995/129474974-78045ad4-223e-4383-a2f1-05105bd17108.jpg)

### Device info

 - Android version/Custom ROM version: Android 11
 - Device model: Redmi note 10 (Sunny/Mojito)

Gottarocket:
When clicking on YT links it used to open newpipe fine now it says loading content getting info on notification and then opens app and above bug takes few more seconds again to load video

swarupbhanjachowdhury:
Yes correct @Gottarocket

akiyev:
I have this problem too.

akiyev:
strange, puretube opens video faster than newpipe

tsiflimagas:
Is this still a problem on 0.23.1?

opusforlife2:
Closing. Please open a new issue if needed.

Issue #3192, opened 2020-03-06
[Proposal/Feedback] Reorganisation/recategorisation of Settings


opusforlife2:
This issue is to collect ideas from the community on how to better organise various settings in the app. For example, #3184 does this for two such settings.

Stypox:
I am in favour of adding a "fine-tuning" settings section, inside which we could add all of the settings people ask for, and they wouldn't be cluttering the ui and reducing accessibility.

opusforlife2:
I think it is also worth separating and demarcating service-specific settings, because an increasing number of services would result in a jumble of settings. For example, download resolutions are different between Youtube, MediaCCC and Peertube, both for video and audio, yet the existing setting caters only to Youtube.

Stypox:
The existing setting shuld do the same for every service :thinking:

opusforlife2:
MediaCCC, for example, has a weird resolution of 576p, and Peertube has 300p. Soundcloud currently has only one bit rate for MP3's, but when it gets Opus support, then we'll need a bit rate selector for that. Plus I think we should have a default bit rate selector for Youtube audio playback as well.

opusforlife2:
Feed-related settings could be grouped into their own menu, when things like auto-refresh and pull to refresh are added.

SameenAhnaf:
>I am in favour of adding a "fine-tuning" settings section, inside which we could add all of the settings people ask for, and they wouldn't be cluttering the ui and reducing accessibility.

Are you talking about this type of sub-section to show related settings on each section of Settings? If yes, we should also consider 'Search' on the top right of settings separately. This search should not interfere with browsing history though.

<details><summary>Screenshot</summary>

![IMG_20210709_160005](https://user-images.githubusercontent.com/79245641/125061247-f26c4a00-e0ce-11eb-8b58-ebded4c5d68d.jpg)

</details>

>MediaCCC, for example, has a weird resolution of 576p, and Peertube has 300p. Soundcloud currently has only one bit rate for MP3's, but when it gets Opus support, then we'll need a bit rate selector for that. Plus I think we should have a default bit rate selector for Youtube audio playback as well.

Needless to say, smashing all settings, subscriptions and playlists into the same UI makes no sense. Peertube, YouTube and Bandcamp all platforms are different.

[...]

PR #7919, opened 2022-02-21
Mitigating long buffering on initial video playback

### LINKED ISSUE #7630: Slow initial video buffer
### Checklist

- [x] I am using the latest version - 0.21.15.981 

Start playing random unopened videos

### Actual behavior
Initial loading time of an unopened video is too high.

### Conditions while reproducing the bug
1. Default video format = webm/240p
2. Internet speed is 100Mbps
3. No issues in the device & it's OS.

### Expected behavior
Videos should start playing instantly.

[Here's the screen recorded proof.](https://youtu.be/9Ld63Br4aR0) 

### Logs

Not generated. I could verbose log, if been asked.

### Device info

 - Android version/Custom ROM version: 11/crDroid 7.13/4.14.190 perf
 - Device model: Redmi Note 10/sunny/mojito

karyogamy:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Hi team, long time no see. 
I've long since had issue with the buffering time when starting out a new video in the built-in player. Sometimes this gets so bad, there's a 10 seconds wait before any picture would even show up. Of course, your experience might differ depending on the phones, network conditions, or luck etc. 

After digging a bit, I found a `continueLoadingCheckIntervalBytes` parameter on ExoPlayer progressive media source that controls the size of data loaded per call from the playback loader. By default, ExoPlayer will load 1MB per call. Reducing this number seems to have solved the initial buffering woes on my side, though I have no idea how it affects other users. So if you have a similar issue with buffering under fast network speed, please give this a try and test extensively to see if it works and if lowering the values has any side effects.

- Fixes #7630, #6927

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
@@ -31,11 +31,13 @@ public class PlayerDataSource {
     private static final int MANIFEST_MINIMUM_RETRY = 5;
     private static final int EXTRACTOR_MINIMUM_RETRY = Integer.MAX_VALUE;
 
+    private final int continueLoadingCheckIntervalBytes;
     private final DataSource.Factory cacheDataSourceFactory;
     private final DataSource.Factory cachelessDataSourceFactory;
 
     public PlayerDataSource(@NonNull final Context context, @NonNull final String userAgent,
                             @NonNull final TransferListener transferListener) {
+        continueLoadingCheckIntervalBytes = PlayerHelper.getProgressiveLoadIntervalBytes(context);
         cacheDataSourceFactory = new CacheFactory(context, userAgent, transferListener);
         cachelessDataSourceFactory
                 = new DefaultDataSourceFactory(context, userAgent, transferListener);
@@ -91,6 +93,7 @@ public class PlayerDataSource {
 
     public ProgressiveMediaSource.Factory getExtractorMediaSourceFactory() {
         return new ProgressiveMediaSource.Factory(cacheDataSourceFactory)
+                .setContinueLoadingCheckIntervalBytes(continueLoadingCheckIntervalBytes)
                 .setLoadErrorHandlingPolicy(
                         new DefaultLoadErrorHandlingPolicy(EXTRACTOR_MINIMUM_RETRY));
     }
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
@@ -34,6 +34,7 @@ import androidx.preference.PreferenceManager;
 import com.google.android.exoplayer2.PlaybackParameters;
 import com.google.android.exoplayer2.Player.RepeatMode;
 import com.google.android.exoplayer2.SeekParameters;
+import com.google.android.exoplayer2.source.ProgressiveMediaSource;
 import com.google.android.exoplayer2.trackselection.AdaptiveTrackSelection;
 import com.google.android.exoplayer2.trackselection.ExoTrackSelection;
 import com.google.android.exoplayer2.ui.AspectRatioFrameLayout;
@@ -389,6 +390,19 @@ public final class PlayerHelper {
         // 0: Screen orientation is locked
         return android.provider.Settings.System.getInt(
                 context.getContentResolver(), Settings.System.ACCELEROMETER_ROTATION, 0) == 0;
+    }
+
+    public static int getProgressiveLoadIntervalBytes(@NonNull final Context context) {
+        final String preferredIntervalBytes = getPreferences(context).getString(
+                context.getString(R.string.progressive_load_interval_key),
+                context.getString(R.string.progressive_load_interval_default_value));
+
+        if (context.getString(R.string.progressive_load_interval_default_value)
+                .equals(preferredIntervalBytes)) {
+            return ProgressiveMediaSource.DEFAULT_LOADING_CHECK_INTERVAL_BYTES;
+        }
+        // Keeping the same KiB unit used by ProgressiveMediaSource
+        return Integer.parseInt(preferredIntervalBytes) * 1024;
     }
 
     ////////////////////////////////////////////////////////////////////////////
--- a/app/src/main/res/values/settings_keys.xml
+++ b/app/src/main/res/values/settings_keys.xml
@@ -55,6 +55,23 @@
         <item>30000</item>
     </string-array>
 
+    <string name="progressive_load_interval_key">progressive_load_interval</string>
+    <string name="progressive_load_interval_default_value">default</string>
+    <string-array name="progressive_load_interval_descriptions">
+        <item>1 KiB</item>
+        <item>16 KiB</item>
+        <item>64 KiB</item>
+        <item>256 KiB</item>
+        <item>@string/progressive_load_interval_default</item>
+    </string-array>
+    <string-array name="progressive_load_interval_values">
+        <item>1</item>
+        <item>16</item>
+        <item>64</item>
+        <item>256</item>
+        <item>default</item>
+    </string-array>
+
     <string name="minimize_on_exit_key">minimize_on_exit_key</string>
     <string name="minimize_on_exit_value">@string/minimize_on_exit_background_key</string>
     <string name="minimize_on_exit_none_key">minimize_on_exit_none_key</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -78,6 +78,8 @@
     <string name="use_inexact_seek_title">Use fast inexact seek</string>
     <string name="use_inexact_seek_summary">Inexact seek allows the player to seek to positions faster with reduced precision. Seeking for 5, 15 or 25 seconds doesn\'t work with this</string>
     <string name="seek_duration_title">Fast-forward/-rewind seek duration</string>
+    <string name="progressive_load_interval_title">Playback load interval size</string>
+    <string name="progressive_load_interval_summary">Change the load interval size (currently at %s). A lower value may speed up initial video loading. Changes require a player restart.</string>
     <string name="clear_queue_confirmation_title">Ask for confirmation before clearing a queue</string>
     <string name="clear_queue_confirmation_summary">Switching from one player to another may replace your queue</string>
     <string name="clear_queue_confirmation_description">The active player queue will be replaced</string>
@@ -717,4 +719,6 @@
     <!-- Show Channel Details -->
     <string name="error_show_channel_details">Error at Show Channel Details</string>
     <string name="loading_channel_details">Loading Channel Details…</string>
+    <!-- Progressive Load Interval -->
+    <string name="progressive_load_interval_default">ExoPlayer default</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -79,7 +79,7 @@
     <string name="use_inexact_seek_summary">Inexact seek allows the player to seek to positions faster with reduced precision. Seeking for 5, 15 or 25 seconds doesn\'t work with this</string>
     <string name="seek_duration_title">Fast-forward/-rewind seek duration</string>
     <string name="progressive_load_interval_title">Playback load interval size</string>
-    <string name="progressive_load_interval_summary">Change the load interval size (currently at %s). A lower value may speed up initial video loading. Changes require a player restart.</string>
+    <string name="progressive_load_interval_summary">Change the load interval size (currently %s). A lower value may speed up initial video loading. Changes require a player restart.</string>
     <string name="clear_queue_confirmation_title">Ask for confirmation before clearing a queue</string>
     <string name="clear_queue_confirmation_summary">Switching from one player to another may replace your queue</string>
     <string name="clear_queue_confirmation_description">The active player queue will be replaced</string>
--- a/app/src/main/res/xml/video_audio_settings.xml
+++ b/app/src/main/res/xml/video_audio_settings.xml
@@ -61,6 +61,16 @@
         app:singleLineTitle="false"
         app:iconSpaceReserved="false" />
 
+    <ListPreference
+        android:defaultValue="@string/progressive_load_interval_default_value"
+        android:entries="@array/progressive_load_interval_descriptions"
+        android:entryValues="@array/progressive_load_interval_values"
+        android:key="@string/progressive_load_interval_key"
+        android:summary="@string/progressive_load_interval_summary"
+        android:title="@string/progressive_load_interval_title"
+        app:singleLineTitle="false"
+        app:iconSpaceReserved="false" />
+
     <PreferenceCategory
         android:layout="@layout/settings_category_header_layout"
         android:title="@string/settings_category_player_title"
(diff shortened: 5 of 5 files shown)


app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -391,6 +391,13 @@ public static boolean globalScreenOrientationLocked(final Context context) {
                 context.getContentResolver(), Settings.System.ACCELEROMETER_ROTATION, 0) == 0;
     }
 
+    @NonNull
+    public static Integer getProgressiveLoadIntervalBytes(@NonNull final Context context) {
+        return 1024 * Integer.parseInt(Objects.requireNonNull(getPreferences(context).getString(
```
litetex, line 396: Shouldn't that be ``1000``?
https://en.wikipedia.org/wiki/Kilobyte

karyogamy: Oh man, my life has been a lie =D
image

That aside, kB is now KiB. Hope it won't be confusing to users.

app/src/main/res/values/settings_keys.xml
```diff
@@ -55,6 +55,25 @@
         <item>30000</item>
     </string-array>
 
+    <string name="progressive_load_interval_key">progressive_load_interval</string>
+    <string name="progressive_load_interval_bytes_default_value">16384</string>
```
litetex, line 59: That is not the default value according to https://github.com/google/ExoPlayer/blob/029a2b27cbdc27cf9d51d4a73ebeb503968849f6/library/core/src/main/java/com/google/android/exoplayer2/source/ProgressiveMediaSource.java#L256

I think adding an default entry that say's "use ExoPlayer's default value" (or something similar) would be great.
If this value (e.g. -1) is encountered the continueLoadingCheckIntervalBytes will be untoughed in the code.

app/src/main/res/values/settings_keys.xml
```diff
@@ -55,6 +55,25 @@
...
+    <string-array name="progressive_load_interval_descriptions">
+        <item>1 KiB</item>
+        <item>4 KiB</item>
+        <item>16 KiB</item>
+        <item>64 KiB</item>
+        <item>256 KiB</item>
```
litetex, line 65: I'm not sure if all those entries are useful.
Is there so much difference between 1 and 4 kB?
Usually network connections nowadays have at least 1MB/s and an under limit of ~8-64kB/s.

I think a list containing
* use default
* 1kB
* 16kB
* 256kB

would be enough.

app/src/main/res/values/strings.xml
```diff
@@ -78,6 +78,8 @@
     <string name="use_inexact_seek_title">Use fast inexact seek</string>
     <string name="use_inexact_seek_summary">Inexact seek allows the player to seek to positions faster with reduced precision. Seeking for 5, 15 or 25 seconds doesn\'t work with this</string>
     <string name="seek_duration_title">Fast-forward/-rewind seek duration</string>
+    <string name="progressive_load_interval_title">Playback Load Interval Size</string>
```
litetex, line 81: Not sure about the writing but I think it should look like this:
```suggestion
    <string name="progressive_load_interval_title">Playback load interval size</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -78,6 +78,8 @@
...
+    <string name="progressive_load_interval_summary">Change the load interval size (currently at %s) on progressive/normal streams. A lower value may speed up initial video loading. Restart player to take effect.</string>
```
litetex, line 82: I would short this a bit:
```suggestion
    <string name="progressive_load_interval_summary">Change the load interval size (currently at %s). A lower value may speed up initial video loading. Changes require a player restart.</string>
```

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
```diff
@@ -31,11 +31,13 @@
     private static final int MANIFEST_MINIMUM_RETRY = 5;
     private static final int EXTRACTOR_MINIMUM_RETRY = Integer.MAX_VALUE;
 
+    private final Integer continueLoadingCheckIntervalBytes;
```
litetex, line 34: Why is this a non primitive datatype?
Why not use ``int`` instead?

litetex (commented):
Preamble: I'm not sure if this fixes the above issues in the first place. Usually people have at least >1MB/s downlink - so transferring the required data should take 1-2s at worst.

Anyway a few more notes:

TobiGr:
Welcome back!

allgasno:
Hey @karyogamy whats your internet speed? I don't think theres much difference between 1mb and 256kb, but I'm willing to test could you please direct me to the test apk for myself and others, thank you. Have an internet speed around 250Mbps.

app/src/main/res/values/settings_keys.xml
karyogamy: Setting the default to 16KiB was to help testers to see the effect right away before muddling with the config.
I've updated that to reflect the ExoPlayer default now, and yes, using the default value from lib is better than having it hardcoded on our side. Fixed.

app/src/main/res/values/settings_keys.xml
```diff
@@ -55,6 +55,25 @@
         <item>30000</item>
     </string-array>
 
+    <string name="progressive_load_interval_key">progressive_load_interval</string>
+    <string name="progressive_load_interval_bytes_default_value">16384</string>
+    <string-array name="progressive_load_interval_descriptions">
+        <item>1 KiB</item>
```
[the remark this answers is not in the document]
karyogamy: Keeping this at 1024/KiB for now since that is the unit used in ExoPlayer. Will test *1000 on my side and see if there is anything noticeable before making the change.

app/src/main/res/values/settings_keys.xml
karyogamy: >  Is there so much difference between 1 and 4 kB?
Nope. At least on my device, there is no difference between 1/4/16 kB, but I do notice a very small difference going from 16 to 64. Will remove the 1kB option for now.

app/src/main/res/values/strings.xml
karyogamy: Done.

app/src/main/res/values/strings.xml
karyogamy: Done.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
karyogamy: Good catch, fixed.

karyogamy:
> Preamble: I'm not sure if this fixes the above issues in the first place. Usually people have at least >1MB/s downlink - so transferring the required data should take 1-2s at worst.

Thanks for the review @litetex!
This fix is not really dealing with issues arising from network speed. On my side, many of the videos load and play quickly enough, but sometimes, a video would buffer for a long time (~10s) at the start. When this buffering happens, I'd fast forward once and the video would start playing right away instead of buffering for another few seconds. It also didn't make sense to me why this initial buffering is so pronounced here but not on other similar apps or platforms. Though like I mentioned, this might just be affecting a small number of users, but if possible, please try clicking on a variety of individual videos and see if this can be reprod on your devices too. 

> Hey @karyogamy whats your internet speed? I don't think theres much difference between 1mb and 256kb, but I'm willing to test could you please direct me to the test apk for myself and others, thank you. Have an internet speed around 250Mbps.

Thanks @allgasno, my internet is at 500Mbps so this is not really caused by speed, but perhaps something internal to the player like mentioned. 
Please do test it, I will ask for help on the other issue threads as well. You can find the zipped app under Artifacts in https://github.com/TeamNewPipe/NewPipe/actions/runs/1878608107.

pc00per:
I've tested the debug app & it turned out okay when I switched the interval load time from ExoPlayer default to other value. Basically all 4 values (1k,16k, 64k, 256k) are same. That means the issue must be ExoPlayer being couldn't handle moov atom of the streams, probably the issue.

pc00per (approved):
Fixes buffer issues for me. I found all intervals have no difference in the latency.

opusforlife2:
(This is speculation; I haven't tested yet) If it turns out that the 1MB value is better for general cases (loading other than the start of a stream) is it possible to use the smaller value for just the first few seconds, and 1MB (or something higher?) for the rest of the stream?

karyogamy:
@opusforlife2 That's exactly the issue: This change adds a parameter that is passed directly from `ProgressiveMediaSource` factory to the media source and can't be changed on-the-fly unless we duplicate thousands of lines of code. 

Also, this issue doesn't only affect the first video played, but every video in a play queue that hasn't been pre-buffered for seamless playback (e.g. when you skip a few songs). 

Honestly, I'm not sure how many users are affected by the same buffering issue, though my guess is a good majority are. That is why I agree with @litetex to keep the exoplayer setting (1MB) default for now. So if buffering complaint keeps coming in and gets fixed with this settings change, we can then consider making a smaller value default instead.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -391,6 +392,20 @@ public static boolean globalScreenOrientationLocked(final Context context) {
...
 
+    public static int getProgressiveLoadIntervalBytes(@NonNull final Context context) {
+        final String preferredIntervalBytes = getPreferences(context).getString(
+                context.getString(R.string.progressive_load_interval_key),
+                context.getString(R.string.progressive_load_interval_default_value));
+
+        if (context.getString(R.string.progressive_load_interval_default_value)
+                .equals(preferredIntervalBytes)) {
+            return ProgressiveMediaSource.DEFAULT_LOADING_CHECK_INTERVAL_BYTES;
+        } else {
```
litetex, line 403: The else is not required here 😄

Otherwise the PR LGTM

app/src/main/res/values/settings_keys.xml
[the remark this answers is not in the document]
karyogamy: > Otherwise the PR LGTM

Still interested in this change? Base on my own tests, there seems to be little to no difference between 1000x vs 1024x, though I'd still go for what exo devs are using.

teaerror:
Does anyone know why 1MB doesn't work for NewPipe, when ExoPlayer uses it? 
Also, skipping 10 seconds resolves the issue but setting the playback load to 256 as default would help several users with buffering issues.

opusforlife2:
In my testing, none of the values give a consistently faster playback start. I've seen both slow and fast starts on every value. I'm assuming my internet latency isn't stable enough to test this. 🤷

karyogamy:
@teaerror Yup, that was the page that eventually led me to this solution. 
The workaround flags didn't seem to make a difference though. 
This comment might help with why `continueLoadingCheckIntervalBytes` worked: https://github.com/google/ExoPlayer/issues/4014#issuecomment-384673429

@opusforlife2 Interesting. Initial buffering should become more consistent with the change. 
Would you mind sharing your internet speed? 
Also, did you make sure the players were fully shutdown and restarted after changing the settings?

opusforlife2:
It varies between 50-100 Mbps, but as I said, latency is quite an issue.

I was using the main player only, and I closed it before every time I opened settings.

litetex (approved):
LGTM

Test worked fine. I think the initial loading is a bit better now 😄 

Applied minor changes:
* Updated the branch
* Moved the preference to a better matching section
* Shortened a string

(@karyogamy you might also cleanup your repo's branches)

Stypox (approved):
Code looks good to me, too, though it doesn't seem to make any difference for me, like opus

opusforlife2:
Oh. I thought this was meant to be a test PR. To find out what the best value is and make that the default. Are we going to advise users to adjust the value based on their own experiences?

Stypox:
Oh, I didn't get that ;-)
Should the PR be reverted and should we choose a specific value for everyone? This makes sense so that we don't introduce a new setting without a clear meaning except for us.

karyogamy:
@opusforlife2 @Stypox I think the plan is to have the setting available to all for the time being, since there isn't a clear answer as to which value is best for everyone, especially now we know changing the value doesn't seem to make a difference for some. The side effect of choosing a smaller value is still unclear too. My guess is there has to be some kind of a trade-off. 

We can either tell the users about this setting proactively, or when there's more buffering issues come in. A new default should be selected in a separate PR when the team is comfortable with it. 

On a side note, would it make sense to have a separate settings section for these kind of fine-tuning details? e.g. Player internal buffer/cache size, extractor cache size, etc. Maybe a public Debug section if settings might disappear later?

opusforlife2:
Sure. We can work with it in the blog. A temporary setting to poll our users. They could respond on our chosen platform(s), like here, Reddit, blog comments, etc.

> On a side note, would it make sense to have a separate settings section for these kind of fine-tuning details? e.g. Player inter

https://github.com/TeamNewPipe/NewPipe/issues/3192#issuecomment-599146600 ;)

pc00per:
The recent v0.22.2 still buffers like crazy, even though this fix claims to be merged. Something's strange.

TiA4f8R:
Nothing strange, only the extractor version and the required app changes have been made (that's why 0.22.2 is a hotfix version and not a normal version). It will be included in version 0.23.0.

Stypox:
This issue is present even on `dev` though. I added a TODO point to #8230. The cause might be completely different from the one this PR fixes, since switching resolution almost instantly loads the video, and this behaviour could be observed only after updating the NewPipeExtractor (since I could notice it in the player refactor PR before YouTube broke). #8224 is actually relevant (since I am experiencing this issue, too, on 0.23.0 builds).

karyogamy:
@Stypox Behaviour is still seems fine on my end, did you test by setting a lower interval size? 
Also, is the problem reproducible on every video? IIRC, there are some exceptions where the videos would still take a while to load even with the settings changed. In those cases, I added a hack to skip the first frames, but left it out of this PR since I'd much prefer a proper solution like #8153. 

But if you are interested, try adding these lines [here](https://github.com/TeamNewPipe/NewPipe/pull/8231/files#diff-68cd2f051cf461f1cee8f97ab4a793ab90276943eed26f8f1baf5105948b2b5eR2557) and see if it does anything:
```
        if (newPosition.contentPositionMs == 0
                && simpleExoPlayer.getTotalBufferedDuration() < 500L) {
            Log.d(TAG, "Playback - skipping to initial keyframe.");
            simpleExoPlayer.setSeekParameters(SeekParameters.CLOSEST_SYNC);
            simpleExoPlayer.seekTo(1L);
            simpleExoPlayer.setSeekParameters(PlayerHelper.getSeekParameters(context));
        }
```

litetex:
→ Might be related: https://github.com/TeamNewPipe/NewPipe/issues/8238

ColeWunderlich:
For what it's worth I am also having issues with very long buffer times on v0.22.0.  

I am able to consistently fix this by:
1. Attempting to play the video
2. Immediately hitting the "background" button (ie audio only) 
3. Immediately switching back to full video after the background playback has started. Then everything works as normal. 

This trick consistently works for me. Without the trick initial buffer times are usually over 30 sec for a video, but with this trick the video plays instantly without fail (background audio never takes more than a sec to kick in).

Perhaps this might provide some insight as to what's going on? I find it interesting that this "fix" doesn't involve changing any settings, which tells me the old defaults might be fine and something else might be going on. 

PS. I have gigabit internet, so there's no way internet speed is causing buffering issues

PR #7925, opened 2022-02-22
Replace CircleImageView with ShapeableImageView

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- The CircleImageView library repository hasn't been touched in over a year (though I guess that doesn't matter much). ImageView transformation exists now in Google Material anyway, so the library is no longer necessary.

- Not necessary. Although the margins and padding worked pretty differently between the two ImageView implementations (requiring quite a bit of adjustments), I matched the previous behavior as precisely as possible by using exact margins, adding some tiny padding where necessary, and other such things. I actually dare you to try and find a noticeable difference :) I DARE YOU-

#### APK testing 

#### Due diligence


app/build.gradle
```diff
@@ -246,8 +246,6 @@ dependencies {
     implementation "com.github.lisawray.groupie:groupie:${groupieVersion}"
     implementation "com.github.lisawray.groupie:groupie-viewbinding:${groupieVersion}"
 
-    // Circular ImageView
-    implementation "de.hdodenhof:circleimageview:3.1.0"
```
TiA4f8R, line 250: How much the APK size is decreased with the library removal?

TiA4f8R (commented):
Just a question and a suggestion. I didn't test your changes.

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
```diff
@@ -328,9 +332,13 @@ public void handleResult(@NonNull final PlaylistInfo result) {
...
             // this is an auto-generated playlist (e.g. Youtube mix), so a radio is shown
-            headerBinding.uploaderAvatarView.setDisableCircularTransformation(true);
-            headerBinding.uploaderAvatarView.setBorderColor(
-                    getResources().getColor(R.color.transparent_background_color));
+            final ShapeAppearanceModel model = ShapeAppearanceModel.builder()
+                    .setAllCorners(CornerFamily.ROUNDED, 0f)
+                    .build(); // this turns the image back into a square
+            headerBinding.uploaderAvatarView.setShapeAppearanceModel(model);
+            headerBinding.uploaderAvatarView.setStrokeColor(ColorStateList.valueOf(
+                    getResources().getColor(R.color.transparent_background_color))
```
litetex, line 340: Not sure why this is formatted like this.
I think something like this would be better:
```suggestion
            headerBinding.uploaderAvatarView.setStrokeColor(
                   ColorStateList.valueOf(
                          getResources().getColor(R.color.transparent_background_color))
```

litetex (commented):
LGTM

Code is ok and a short also revealed no problems.
Please change the code as @TiA4f8R noted above.

app/build.gradle
TacoTheDank: Debug builds are around 5KB smaller.

Release builds are around 1KB larger.

Nothing large 👍

app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
TacoTheDank: Sure 👍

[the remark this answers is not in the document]
TacoTheDank: Sure 👍

PR #2335, opened 2019-05-09
New streams notifications

nv95:
Checking and notifying about new streams from subscriptions
Closes #719 closes #1086

| Screenshot_2019-05-09-21-00-19 | Screenshot_2019-05-09-21-02-27 |
|---|---|

- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #7975, opened 2022-03-01
Migrate app update checker to AndroidX Work

### LINKED ISSUE #7397: CheckForNewAppVersion randomly fails (when app is in background)
### Checklist

No idea. NewPipe popped up from background with the error message.

### Actual behavior

NewPipe popped up from background with the error message.

### Expected behavior

Do not pop up with random

### Device info

 - Android version/Custom ROM version: lineage_enchilada-userdebug 11 RQ3A.211001.001 10036215
 - Device model: OnePlus 6

## Exception
* __User Action:__ ui error
* __Request:__ ACRA report
* __Content Country:__ RU
* __Content Language:__ ru-RU
* __App Language:__ ru_RU
* __Service:__ none
* __Version:__ 0.21.13
* __OS:__ Linux Android 11 - 30

```
java.lang.IllegalStateException: Not allowed to start service Intent { cmp=org.schabi.newpipe/.CheckForNewAppVersion }: app is in background uid UidRecord{27dcaf3 u0a248 CAC  bg:+1m47s807ms idle change:idle procs:1 seq(0,0,0)}
	at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1715)
	at android.app.ContextImpl.startService(ContextImpl.java:1670)
	at android.content.ContextWrapper.startService(ContextWrapper.java:720)
	at org.schabi.newpipe.CheckForNewAppVersion.startNewVersionCheckService(CheckForNewAppVersion.java:258)
	at org.schabi.newpipe.MainActivity.onPostCreate(MainActivity.java:175)
	at android.app.Instrumentation.callActivityOnPostCreate(Instrumentation.java:1383)
	at android.app.ActivityThread.handleStartActivity(ActivityThread.java:3484)
	at android.app.servertransaction.TransactionExecutor.performLifecycleSequence(TransactionExecutor.java:221)
	at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:201)
	at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:173)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
	at android.app.ClientTransactionHandler.executeTransaction(ClientTransactionHandler.java:58)
	at android.app.ActivityThread.handleRelaunchActivityLocally(ActivityThread.java:5414)
	at android.app.ActivityThread.access$3300(ActivityThread.java:237)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2076)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:223)
	at android.app.ActivityThread.main(ActivityThread.java:7664)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)

```
</details>
<hr>

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

IntentService is deprecated, so I migrated to what was supposed to be its replacement (JobIntentService, which is also now deprecated), and then I migrated to AndroidX Room (which is now the recommended replacement).

Unfortunately this means yet another library, but it is what it is.

1. I moved the utility methods out of the service class to make it smaller because that stuff didn't really belong in there. Also I cleaned up some parameters.
2. Convert the new utility class to Kotlin so I can merge it together with NewVersionManager.
3. Migrate the IntentService to what was supposed to be its replacement (JobIntentService). This itself would be enough to fix #7397. However, I wanted to go further. This is because JobIntentService is also terrible in general and has an infamous crashing bug on API 26. Also it's deprecated (lol). So...
4. Convert this Service class into a Worker. Workers are what we are now supposed to use. Also I renamed the class to make more sense.
5. Converted this new Worker class to Kotlin.

I tested these changes on API 29. I checked the manual update button in settings and it works.

Here's more info:
- https://developer.android.com/topic/libraries/architecture/workmanager
- https://developer.android.com/reference/androidx/work/WorkManager

Here's another PR in which I made this change: https://github.com/AntennaPod/AntennaPod/pull/5742

- Fixes #7397 (I think)

#### APK testing 
[app-release updateCheckWorker.zip](https://github.com/TeamNewPipe/NewPipe/files/8164958/app-release.updateCheckWorker.zip)

This is a specially generated APK where the `isReleaseApk()` method always returns true. This is only for testing purposes.

#### Due diligence


Stypox:
> and then I migrated to AndroidX Room

I assume this is a typo and you meant AndroidX Work

> Unfortunately this means yet another library, but it is what it is.

I think that's not a big issue, since AndroidX Work was going to be added anyway by #2335

TacoTheDank:
> I assume this is a typo and you meant AndroidX Work

Damn, and I checked it over like three times too LOL, fixed

> I think that's not a big issue, since AndroidX Work was going to be added anyway by #2335

👍

app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
```diff
@@ -0,0 +1,122 @@
...
+            return
+        }
+
+        val prefs = PreferenceManager.getDefaultSharedPreferences(applicationContext)
+        // Check if the last request has happened a certain time ago
+        // to reduce the number of API requests.
+        val expiry = prefs.getLong(applicationContext.getString(R.string.update_expiry_key), 0)
+        if (!isExpired(expiry)) {
+            return
+        }
```
Stypox, line 40: Can the expiration date be handled by some parameter in the worker?

app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
```diff
@@ -0,0 +1,122 @@
...
+
+        // Make a network request to get latest NewPipe data.
+        val response = DownloaderImpl.getInstance()[NEWPIPE_API_URL]
```
Stypox, line 43: [I understand](https://kotlinlang.org/docs/operator-overloading.html#indexed-access-operator) that Kotlin allows writing `[]` instead of `.get()`, but I'd use `.get()` here since it's not accessing a list ;-)
```suggestion
        val response = DownloaderImpl.getInstance().get(NEWPIPE_API_URL)
```

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
```diff
@@ -0,0 +1,160 @@
...
+import java.security.NoSuchAlgorithmException
+import java.security.cert.CertificateEncodingException
+import java.security.cert.CertificateException
+import java.security.cert.CertificateFactory
+import java.security.cert.X509Certificate
+import java.time.Instant
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
+
+object ReleaseUtil {
```
Stypox, line 29:
```suggestion
object ReleaseVersionUtil {
```

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
```diff
@@ -0,0 +1,160 @@
...
+        notificationManager.notify(2000, notificationBuilder.build())
+    }
+
+    private fun showRequestError(app: App, e: Exception, request: String) {
+        createNotification(
+            app, ErrorInfo(e, UserAction.CHECK_FOR_NEW_APP_VERSION, request)
+        )
+    }
+
+    fun isExpired(expiry: Long): Boolean {
```
Stypox, line 141:
```suggestion
    fun isLastUpdateCheckExpired(expiry: Long): Boolean {
```

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
```diff
@@ -0,0 +1,160 @@
...
+        return Instant.ofEpochSecond(expiry).isBefore(Instant.now())
+    }
+
+    /**
+     * Coerce expiry date time in between 6 hours and 72 hours from now
+     *
+     * @return Epoch second of expiry date time
+     */
+    fun coerceExpiry(expiryString: String?): Long {
```
Stypox, line 150:
```suggestion
    fun coerceUpdateCheckExpiry(expiryString: String?): Long {
```

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * Method to compare the current and latest available app version.
+     * If a newer version is available, we show the update notification.
+     *
+     * @param versionName    Name of new version
+     * @param apkLocationUrl Url with the new apk
+     * @param versionCode    Code of new version
+     */
+    fun compareAppVersionAndShowNotification(
```
Stypox, line 106: Why is this function here? I'd move it to `NewVersionWorker` as it is used only there and is strictly related to the worker

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
```diff
@@ -0,0 +1,160 @@
...
+            .setContentTitle(app.getString(R.string.app_update_notification_content_title))
+            .setContentText(
+                app.getString(R.string.app_update_notification_content_text)
+                        + " " + versionName
+            )
+        val notificationManager = NotificationManagerCompat.from(app)
+        notificationManager.notify(2000, notificationBuilder.build())
+    }
+
+    private fun showRequestError(app: App, e: Exception, request: String) {
```
Stypox, line 135: Same as my comment above

Stypox (changes requested):
Thank you! Note: this should be tested well, otherwise we might encounter the same problems that forced us to do 2 hotfix releases in a row a while ago ;-)

app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
TacoTheDank: Sure 👍

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
TacoTheDank: I felt that it was unnecessarily cluttering up the `NewVersionWorker`. I can move it back, I suppose.

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
TacoTheDank: This is only for the `certificateSHA1Fingerprint` getter, to shorten it and make it a bit more readable. It's supposed to be here :)

app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
TacoTheDank: I mean I guess, but I'm not sure why? Could you expand on this?

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
Stypox: In my opinion yes, it is not a "release utility" that could be useful throughout the app, it is a method that creates the notification as an output to the worker

app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
Stypox: Nevermind, there [doesn't seem](https://developer.android.com/reference/androidx/work/WorkRequest.Builder) [to be](https://developer.android.com/reference/androidx/work/Constraints) any method that allows scheduling a work only if it has not been already run in the previous X hours. Stick with this, sorry for the confusion.

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
Stypox: Right, keep it here ;-)

app/src/main/java/org/schabi/newpipe/util/ReleaseUtil.kt
TacoTheDank: Okies, done 👍

TacoTheDank:
Updated the attached testing APK.

TacoTheDank:
Any updates on this?

Stypox (approved):
I tested on API 19 and API 31 three different situations (last update check expired -> notification, last update check not expired -> nothing, manual update check -> notification) and I can confirm it works. Thank you and sorry for the delay :-)

PR #7024, opened 2021-08-30
Never use ``android.R.string``s; remove unused ``add`` string

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
- Rename string `finish`, whose content was "OK", to `ok`
- Replace `android.R.string.yes` and `android.R.string.ok` with `R.string.ok`
- Replace `android.R.string.no` and `android.R.string.cancel` with `R.string.cancel`
- Remove unused string `add` (legacy left behind from the downloader)
- Reformat some lines of code in `FeedGroupDialog`, solving Android Studio warnings

#### Fixes the following issue(s)
Fixes https://github.com/TeamNewPipe/NewPipe/pull/6848#issuecomment-907623835

#### APK testing 
The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.
@nadiration could you check if this works well?

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #5878, opened 2021-03-22
Add a confirmation button when deleting all files in downloader

SpinHit:
Adding an confirmation button while deleting all files.

#### Fixes the following issue(s)

- Fixes #5419

Issue #5454, opened 2021-01-18
Missing strings (Weblate)


nadiration:
### Checklist

- [x] I am using the latest version - x.xx.x 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

1. ~~Search any video and the published time/date (e.g 18 hours ago, 1day/month/year ago) of the content is stuck with English even though you switched the language (in this case Somali)~~ Edit: This happened to be social date formatter PrettyTime which NewPipe uses so I translated the language for the project. This is SOLVED.
2. "Next" on video page acts the same way.
3. "Cancel" key on the app language picker (edit: and some other places i just found out) is the same although cancel in other places of the app is translated fine. 
Edit: the cancel not being translated button isn't limited to language picker but also on other places like creating new chanel groups

See screenshots for context.

### Actual behaviour

"Next" on video page isn't following the app language, that's due to "Next" string missing from translation strings on Weblate.

The "Cancel" key on the app language picker and some other places in the app.

### Expected behavior

Complete strings on Weblate for the app to be consistent.

### Screenshots/Screen recordings

~~*on search (published time)~~ Edit: SOLVED.

*on video page ("Next")
![](https://user-images.githubusercontent.com/59365236/104948402-4cb50980-59ce-11eb-9622-bfa176837263.png)

*on creating new chanel groups and some other places ("Cancel") :

![Screenshot_20210119-143134](https://user-images.githubusercontent.com/59365236/105029423-66546080-5a63-11eb-929f-64f886bfddb0.png)

@TobiGr

TobiGr:
@stypox any idea, why `exo_controls_next_description` is used? I did not noticed, that it was changed.
I suggest to use something different, because "next track" is not always true (e.g. the queue can contain other streams which are up next). What about "related content" or "similar content" or "related streams"?

nadiration:
> @stypox any idea, why `exo_controls_next_description` is used? I did not noticed, that it was changed.
> I suggest to use something different, because "next track" is not always true (e.g. the queue can contain other streams which are up next). What about "related content" or "similar content" or "related streams"?
> 
> About the upload time: that seems to be a bug, because some lib should be responsible for that.

[...]

PR #7976, opened 2022-03-02
Replace `R.string.yes` with `R.string.ok`

Stypox:
Android doesn't use yes/no but ok/cancel usually, so this should be done here, too

#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

#### Related info
This can be considered a followup to #7024. I thought all `yes` were replaced by `ok` in that PR, but apparently it was not the case (as I could notice in #5878), hence this PR.

#### APK testing 

#### Due diligence


litetex (approved):
LGTM

Redirion:
any reason why we don't use android.R.string.yes or android.R.string.ok ?

TiA4f8R:
No, we can't for some languages: see the PR linked and #5454.

Redirion:
ok yeah but that was regarding the well known "cancel". But anyway :)

PR #7981, opened 2022-03-03
Deduplicate code for fetching stream info when sparse

### LINKED ISSUE #7941: Duplicated methods that fetch full info for sparse streams
### Checklist

### Affected version

After (but also before) #7570

### The problem

There are currently two methods which do almost the same thing: `StreamDialogEntry.fetchItemInfoIfSparse()` and `SaveUploaderUrlHelper.saveUploaderUrlIfNeeded()`. They both return some information from the info item if it is available right away (though using different checks, which is inconsistent), or fetch the full stream info if the info item was sparse. They should be deduplicated. See https://github.com/TeamNewPipe/NewPipe/pull/7570#discussion_r810488117

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

As explained in #7941 the method to fetch stream info item if sparse and to fetch uploader url if missing worked basically the same way, by fetching the `StreamInfo` if needed. This PR unifies those two methods in `SparseItemUtil`. They now show a toast with the correct information ("Fetching stream details" instead of "Fetching channel details"), and does so not only when loading uploader url but also when loading stream info before (en)queueing. I also fixed the way streams are saved to database: instead of only saving the channel url or saving the whole stream but setting its watch duration to 0 uselessly, now the stream info is `upsert`ed in the DB fully and without side effects. I also removed an unused related string.

- Fixes #7941

#### APK testing 

#### Due diligence


app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
```diff
@@ -0,0 +1,126 @@
...
+     * org.schabi.newpipe.extractor.feed.FeedExtractor}. FeedExtractors provide a fast and
+     * lightweight method to fetch info, but the info might be incomplete (see
+     * {@link org.schabi.newpipe.local.feed.service.FeedLoadService} for more details).
+     *
+     * @param context the Android context
+     * @param item the item which is checked and eventually loaded completely
+     * @param callback the callback to call with the single play queue built from the original item
+     *                 if all info was available, otherwise from the fetched {@link
+     *                 org.schabi.newpipe.extractor.stream.StreamInfo}
+     */
```
TiA4f8R, line 48: Indent the parameters description so they are all aligned on the same column (like that's the case in most of other places).

app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
```diff
@@ -0,0 +1,126 @@
...
+            // all info is already there, no need to fetch
+            callback.accept(new SinglePlayQueue(item));
+        }
+    }
+
+    /**
+     * Use this to certainly obtain an uploader url when the stream info item or play queue item you
+     * are handling might not have the uploader url (e.g. because it was fetched with {@link
+     * org.schabi.newpipe.extractor.feed.FeedExtractor}). A toast is shown if loading details is
+     * required.
```
TiA4f8R, line 66: You should probably move this sentence to a new paragraph on a new line, by using `<p>` and `</p>` (also add a "space line" first (line with only a `*`)).

app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @param context the Android context
+     * @param serviceId the serviceId of the item
+     * @param url the item url
+     * @param uploaderUrl the uploaderUrl of the item; if null or empty will be fetched
+     * @param callback the callback called with either the original uploaderUrl, if it was a valid
+     *                 url, otherwise with the uploader url obtained by fetching the {@link
+     *                 org.schabi.newpipe.extractor.stream.StreamInfo} corresponding to the item
+     */
```
TiA4f8R, line 75: Same as above for parameters' description.

app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
```diff
@@ -0,0 +1,126 @@
...
+                    streamInfo -> callback.accept(streamInfo.getUploaderUrl()));
+        } else {
+            callback.accept(uploaderUrl);
+        }
+    }
+
+    /**
+     * Loads the stream info corresponding to the given data on an I/O thread, stores the result in
+     * the database and calls the callback on the main thread with the result. A toast will be shown
+     * to the user about loading stream details, so this needs to be called on the main thread.
```
TiA4f8R, line 92: Same as my previous comment about new sentences on new lines.

app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @param context the Android context
+     * @param serviceId the service id of the stream to load
+     * @param url the url of the stream to load
+     * @param callback the callback to call with the result
```
TiA4f8R, line 97: Same as above (for parameters' description).

app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
```diff
@@ -0,0 +1,126 @@
...
+     */
+    private static void fetchStreamInfoAndSaveToDatabase(final Context context,
```
TiA4f8R, line 99: The context should be annotated with `@NonNull`, because it must not be null to show a toast (and maybe in more places).

TiA4f8R (commented):
Note that I didn't really take a look at the real code in my review, one about it is necessary.

Stypox:
@TiA4f8R I disagree with all of your review except for the `@NonNull` and possibly the alignment of the parameters ;-)

app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
```diff
@@ -0,0 +1,126 @@
...
+                                             @NonNull final StreamInfoItem item,
+                                             @NonNull final Consumer<SinglePlayQueue> callback) {
+        if ((!(item.getStreamType() == LIVE_STREAM || item.getStreamType() == AUDIO_LIVE_STREAM)
+                && item.getDuration() < 0) || isNullOrEmpty(item.getUploaderUrl())) {
+            fetchStreamInfoAndSaveToDatabase(context, item.getServiceId(), item.getUrl(),
+                    streamInfo -> callback.accept(new SinglePlayQueue(streamInfo)));
+        } else {
+            // all info is already there, no need to fetch
+            callback.accept(new SinglePlayQueue(item));
+        }
```
litetex, line 59: I think this would be better understandable:

```suggestion
        // Don't do anything when the stream is a livestream OR the duration is already fetched 
        // AND the uploader url is not empty
        if ((item.getStreamType() == LIVE_STREAM || item.getStreamType() == AUDIO_LIVE_STREAM
                || item.getDuration() >= 0) && !isNullOrEmpty(item.getUploaderUrl())) {
            // all info is already there, no need to fetch
            callback.accept(new SinglePlayQueue(item));
            return;
        }
        fetchStreamInfoAndSaveToDatabase(context, item.getServiceId(), item.getUrl(),
            streamInfo -> callback.accept(new SinglePlayQueue(streamInfo)));
```

litetex (commented):
LGTM, only one minor thing.

Btw how can I test the changes?

app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
XiangRongLin, line 66: Isn't this mixing UI stuff (Toast) and logic (ensure uploader url) thus making testing very hard?

I can understand that from the DRY standpoint, one wants to have both here, but from the SoC standpoint it seems wrong.

app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
Stypox: Done and did not find other places

app/src/main/java/org/schabi/newpipe/util/SparseItemUtil.java
Stypox: I will not do this

Stypox:
I solved or answered your comments, thank you for the reviews

litetex:
@Stypox 
I think you didn't answer my questions and suggestions...

Stypox:
Oh, I missed the "how can I test the changes".  Shall I create a database in which the feed has some items with missing data?

litetex:
> Oh, I missed the "how can I test the changes" (but I did implement the suggestion and forgot to push it, see latest commit). Shall I create a database in which the feed has some items with missing data?

You can if you want 😄

Stypox:
Here it is: just import, open the feed fragment and reload the feed and you will get a bunch of videos with missing data. If you then want to test videos with all data, just disable fast loading and reload. [NewPipeData-20220316_112803.zip](https://github.com/TeamNewPipe/NewPipe/files/8261282/NewPipeData-20220316_112803.zip)

PR #7565, opened 2021-12-21
Change pitch by semitones

haklc:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Added optional semitone slider to Playback Parameter Dialog
- Added toggle to enable/disable this slider
- Added code that enables use of semitones along percent for pitch adjustment

#### Before/After Screenshots/Screen Record

- Before:

- After:
 
 

#### Fixes the following issue(s)

- Fixes #7322

 

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #5927, opened 2021-03-27
Migrate to Android DayNight Theme, fix Light Theme, minor UI improvements

krlvm:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Fixed Dark Text & Dark Icons in Light Theme
- Fixed Light Navigation Bar Icons in Light Theme
- Navigation Bar in Splash Screen is red like it's background now
- All styles relies on Android's DayNight Theme now
- Light Theme is now looks properly, so I changed default theme from Dark to Automatic
- All `_white` and `_black` assets have been separated in `drawable` and `drawable-night`, `ThemeHelper.resolveResourceIdFromAttr` is no longer used - it should improve the overall performance, but I did not run benchmarks
- Styles, attributes XML code cleanup
- Slightly reduced installed app size (installed v0.21.0-release: 17MB -> 14,68MB debug version)

#### Fixes the following issue(s)

- Fixes #3042

#### APK testing 

On the website the APK can be found by going to the "Checks" tab below the title and then on "artifacts" on the right.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #7989, opened 2022-03-04
Rewrote ``PlaybackParameterDialog``

### LINKED ISSUE #7960: Overhaul ``PlaybackParameterDialog``
I think the [PlaybackParameterDialog](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java) needs an major overhaul/rework.
Many things are duplicated, could be done in a better way and the code is very confusing in some place. There is also a major lack of documentation/comments.

litetex:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

<details><summary>Backstory</summary>

* Noticed #7565 lately
* Took a look at changes inside there/``PlaybackParameterDialog`` <br/>
* This PR

</details>

* Refactored the code
  * Removed a ton of duplicated stuff
  * Use viewbinding
  * Use ``Icepick`` instead of saving everything manually
* Tuned UI
  * UI fit's now - without a scroller - onto a 5in smartphone (Pixel 3a) in fullscreen mode
  * Removed "Adjust pitch by musical semitones" checkbox and replaced it by a tab like component that can be expanded on demand
  * Semitones now work with "hook"-mode
  * Mark the currently selected steps/pitchmode inside the UI
* Other
  * Used more containers in the UI so that we don't have to do everything inside the code

- Before:
grafik

- After:
grafik

https://user-images.githubusercontent.com/40789489/156843380-566584ad-61d7-4b17-bf97-d43317059b5d.mp4

- Fixes #7960

#### APK testing 

#### Due diligence


app/src/main/java/org/schabi/newpipe/util/DrawableResolver.kt
```diff
@@ -0,0 +1,26 @@
+package org.schabi.newpipe.util
+
+import android.content.Context
+import android.graphics.drawable.Drawable
+import androidx.annotation.AttrRes
+
+/**
+ * Utility class for resolving [Drawables](Drawable)
+ */
+class DrawableResolver {
```
TacoTheDank, line 10: Regarding Kotlin, this should be an object class so that we don't need to use companion objects.

app/src/main/java/org/schabi/newpipe/util/DrawableResolver.kt
```diff
@@ -0,0 +1,26 @@
...
+    companion object {
+        @JvmStatic
+        fun resolveDrawable(context: Context, @AttrRes attrResId: Int): Drawable? {
+            return androidx.core.content.ContextCompat.getDrawable(
+                context,
+                android.util.TypedValue().apply {
```
TacoTheDank, line 16: I think it would be better if these were replaced with imports instead of manually having them be written out.

app/src/main/java/org/schabi/newpipe/util/DrawableResolver.kt
```diff
@@ -2,25 +2,24 @@ package org.schabi.newpipe.util
...
-                        this,
-                        true
-                    )
-                }.resourceId
-            )
-        }
+object DrawableResolver {
+    @JvmStatic
+    fun resolveDrawable(context: Context, @AttrRes attrResId: Int): Drawable? {
+        return androidx.core.content.ContextCompat.getDrawable(
```
TacoTheDank, line 14: Thanks for fixing! However, it seems you missed a spot :P

litetex: Thx for the hint.
I really shouldn't blindly copy past code from stackoverflow (or something like this) next time 😆

karyogamy:
Nice work! By the way, would it be possible to collapse the 2 checkboxes into one line similar to `Percent` and `Semitone`? 
If the text is too long to fit, maybe we can just call them `Unhook` and `Skip Silence`? Even with text and subtext it might save more space.

litetex:
@karyogamy 
I don't think that collapsing the checkboxes would be beneficial, because we don't need to save more space (already optimized for a standard 5" screen).
I might eventually do more UI optimizations in a future PR (e.g. representing the "Unhook" checkbox as a chain between speed and pitch) but the current one primarily aims at code improvements.

app/src/main/java/org/schabi/newpipe/util/DrawableResolver.kt
```diff
@@ -0,0 +1,26 @@
...
+import android.content.Context
+import android.graphics.drawable.Drawable
+import android.util.TypedValue
+import androidx.annotation.AttrRes
+import androidx.core.content.ContextCompat
+
+/**
+ * Utility class for resolving [Drawables](Drawable)
+ */
+object DrawableResolver {
```
Stypox, line 12: Why is this here and not in `ThemeHelper`? I'd prefer not to create yet another class in the `util` package when there is already one that contains similar functions

app/src/main/java/org/schabi/newpipe/player/helper/PlayerSemitoneHelper.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Converts between percent and 12-tone equal temperament semitones.
+ * <br/>
+ * @see
+ * <a href="https://en.wikipedia.org/wiki/Equal_temperament#Twelve-tone_equal_temperament">
+ *     Wikipedia: Equal temperament#Twelve-tone equal temperament
+ * </a>
+ */
+public final class PlayerSemitoneHelper {
+    public static final int TONES = 12;
```
Stypox, line 12: I'd rename this to `SEMITONE_COUNT` or something like that, since `TONES` is different from `SEMITONES` (a tone is twice a semitone)

app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java
```diff
@@ -1,122 +1,120 @@
...
+import icepick.State;
+
 public class PlaybackParameterDialog extends DialogFragment {
+    private static final String TAG = "PlaybackParameterDialog";
+
     // Minimum allowable range in ExoPlayer
-    private static final double MINIMUM_PLAYBACK_VALUE = 0.10f;
-    private static final double MAXIMUM_PLAYBACK_VALUE = 3.00f;
+    private static final double MIN_PLAYBACK_VALUE = 0.10f;
+    private static final double MAX_PLAYBACK_VALUE = 3.00f;
```
Stypox, line 48:
```suggestion
    private static final double MIN_PLAYBACK_SPEED = 0.10f;
    private static final double MAX_PLAYBACK_SPEED = 3.00f;
```

Stypox (changes requested):
> Used more containers in the UI so that we don't have to do everything inside the code

This is not good for performance afaik, but if we don't migrate to `ConstraintLayout` (which has view groups) there is no alternative

app/src/main/java/org/schabi/newpipe/util/DrawableResolver.kt
litetex: ``ThemeHelper`` is used for doing theme-related stuff (e.g. ``getThemeForService``). ``DrawableResolver`` is used for resolving Drawables.

Not sure how these classes should be related...

Stypox: No, `ThemeHelper` contains things that have to do with theme and Context resources. Those two things are tightly coupled since the resources are determined by the theme. Hence why, for example, `resolveColorFromAttr` is in that class, and `resolveDrawable` should go there too. Before #5927 `resolveResourceIdFromAttr` existed in that class, too.

litetex:
> > Used more containers in the UI so that we don't have to do everything inside the code
>
> This is not good for performance afaik, but if we don't migrate to `ConstraintLayout` (which has view groups) there is no alternative (so nothing to do)

1. It aren't so many new "containers" (1-3)
2. I think that `ConstraintLayout` is overrated in the first place - at least for this view:

So I quickly googled why `Constraintlayout` is better than `Relativelayout` and found these sources (from https://stackoverflow.com/questions/59628341/constraintlayout-vs-relativelayout-performance-2020):
* https://developer.android.com/topic/performance/rendering/optimizing-view-hierarchies.html#cheaper
  > In the N release of Android, the ConstraintLayout class provides similar functionality to RelativeLayout, but at a significantly lower cost.
  
  However it doesn't cite any sources why it performs better...
* https://android-developers.googleblog.com/2017/08/understanding-performance-benefits-of.html
  This article is in my opinion not representable:
  * It simply removes all relative layouts when using `ConstraintLayout`. They don't use any view groups.
  * According to their graph the performance in the "Measurement and Layout phase" increased from 0.65ms to 0.5ms (which is a performance increase in 25-30% not 40% as they stated)

TL;DR
I will not rewrite the complete layout now to ``ConstraintLayout`` because it's then 0.15ms faster. Especially regarding that the dialog/layout is rarely used in the first place.
We can create an issue that does this in the follow up, but for now I think that the current state should be enough.

Stypox:
> I will not rewrite the complete layout now to ConstraintLayout because it's then 0.15ms faster. Especially regarding that the dialog/layout is rarely used in the first place.
We can create an issue that does this in the follow up, but for now I think that the current state should be enough.

Yes, I do agree it shouldn't be done here, and I do agree there is no big need to spend time on it. It's just something worth noting in my opinion, since `ConstraintLayout`s are better not only in performance but also they are also simpler to code, cause less errors and are more powerful.

SameenAhnaf:
Do we really need to click twice just to save space? A checkbox could be placed here instead? 
IMG_20220403_173456

litetex:
@SameenAhnaf 
1. How often do you change between semitones and percent?Not that often or?
2. If the "pitch" and "Semtiones" are displayed on a small screen and have long localizations they won't fit

Redirion:
I resolved the conflict directly here on github and it resulted in that not really helpful commit message "Merge branch 'dev'". Can you clean that up please litetex? This should be ready to merge.

Stypox (approved):
I tested and it works well. Thank you!

yashpalgoyal1304:
> * Mark the currently selected steps/pitchmode inside the UI

hey! is this reverted in v0.24.1 ?? I was unable to see the rectangular dashed border around currently selected speed selector/speed adjustment step size buttons in that version.

----
<details>
<summary>Trail/please ignore</summary>

* blame on [dialog_playback_parameter.xml](https://github.com/TeamNewPipe/NewPipe/blame/dev/app/src/main/res/layout/dialog_playback_parameter.xml)
* <-- files in [!9145](https://github.com/TeamNewPipe/NewPipe/pull/9145/files#diff-9e225b66a55f2a93310079e15a6616443352f4008509b9db1e65b0bd821a0224)
* <-- linked in [#8962](https://github.com/TeamNewPipe/NewPipe/issues/8962#ref-pullrequest-1412514715)
* <-- gh issue 🔍 : [adjustment step size](https://github.com/TeamNewPipe/NewPipe/issues?q=adjustment+step+size)
* <-- search term found in [!7728](https://github.com/TeamNewPipe/NewPipe/pull/7728)
* <-- Release 0.23.0 [#8230](https://github.com/TeamNewPipe/NewPipe/issues/8230)
* <-- gh issue 🔍 : [speed selector](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+speed+selector+is%3Aclosed)

</details>

yashpalgoyal1304:
> I can see it just fine. Maybe your OS colours are different?

i just rechecked, and yeah, it was present there.... though it was very faint due to low resolution and small screen size of the device i saw that on.

Issue #7939, opened 2022-02-23
Rewrite the player (and all its components)


AudricV:
### Checklist

- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I'm aware that this is a request for NewPipe itself and that requests for adding a new service need to be made at [NewPipeExtractor](https://github.com/TeamNewPipe/NewPipeExtractor/issues).
- [X] I have taken the time to fill in all the required details. I understand that the feature request will be dismissed otherwise.
- [X] This issue contains only one feature request.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Feature description

The player codebase is pretty hard to understand, is very old, and uses several deprecated methods (which may have been already removed in the latest ExoPlayer versions)/old ways to do some things in ExoPlayer.

The current implementation also does not really allow the implementation of the Android Picture and Picture API (or it is really hard to do), required for Android 12+ devices to prevent untrusted touches outside the popup window, as an activity or a fragment is required to enter the PiP mode. The player seems to be added manually with a view within the `VideoDetailFragment` (that's probably not a good idea at all to enter this fragment into PiP mode).

We currently have some player bugs for which we don't really have a way to detect them and/or fix them, like the playback position reset when switching at least manually player type (that we fixed with a workaround).

We may split the player into two parts such as a part in a fragment and the ExoPlayer part in a separate class and file.

### Why do you want this feature?

I want this "feature" mainly to fix weird player bugs, update to new ExoPlayer APIs, fix popup mode on Android 12+ devices but also to improve code readability and maintenability and make comments about the player code such as `the player is a hot pile of 💩` or `the code is a spaghetti monster of doom` gone.

SC1040-TS2:
This is also necessary now more than ever due to the way both the player and NewPipe's Extractor are currently failing to reliably parse and play WebM videos. Particularly WebM VP9 videos are causing Extractor errors.

Stypox:
Something I was thinking about the other day is the following. We should differentiate between the player logic and the UI.
[...]

Issue #7916, opened 2022-02-20
Video sometimes restarts from beginning when backgrounding


bobdanek:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.22.0

### Steps to reproduce the bug

This is not the only way this is triggered, but it's how I've finally figured out how to reproduce the issue, though it does take a few attempts before it happens for me

1. Play YouTube video in the app, be somewhere far from the beginning of the video
2. Double-tap power button to bring up the camera (GrapheneOS' camera app)
3. Listen to the audio to confirm that it's started from the beginning - if it hasn't, goto step 1 and try again
4. Switch back to NewPipe and observe that your place in the video has been lost and that you're back at the beginning of the video

### Expected behavior

Background audio of a video I'm watching with NewPipe picks up where the video+audio left off when it goes into the background after I switch to another app, and when I return to NewPipe, the video+audio picks up where the background audio left off

### Actual behavior

Occasionally, when I switch to another app while watching a video in NewPipe, the audio restarts from the beginning of the video. And when I go back to the app, the video+audio picks up where that audio left off, at or close to the start of the video.

In other words, I lose my "place" in a video sometimes when switching apps.

### Screenshots/Screen recordings

n/a

### Logs

n/a

### Affected Android/Custom ROM version

GrapheneOS 2022021721

### Affected device model

Google Pixel 5

### Additional information

"Minimize on app switch" is set to "Minimize to background player"

bobdanek:
My apologies, this is not the latest version. I'll update and try to reproduce, and update version number in this report if it still occurs.

bobdanek:
issue also occurs in 0.22.0. I initially reported this for version 0.21.16, but edited the app version in the report so that it doesn't look like I lied when I checked the very first box about reproducing with the latest version. :P

[...]

Issue #6989, opened 2021-08-27
Livestream -> Background: unrecoverable player error


CBSkarmory:
### Checklist

- [x] I am using the latest version - 0.21.9 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug
1. Go to a Youtube livestream (unfortunately I can't provide any links since they are no longer live)
2. Press on `Background` button (icon of headphones)

### Actual behavior

pop-up: "Unrecoverable player error occurred"

However, playing as video and then pressing the android home button allows it to play in the background (though tapping the notification will bring back the video rather than the usual audio-only control interface)

### Expected behavior

It should play in the background (audio only)

### Logs

N/A, no crash/prompt to report bug

### Device info

 - Android version/Custom ROM version: Android 11
 - Device model: Redmi Note 9S

Denisdjt07:
Hi.
I have noticed exactly the same.
No crash log to report.
Device Model: LG Q6
Android 7

mhmdanas:
Are you facing the issue on a specific video? What service are you using (e.g. YouTube or SoundCloud)?

Denisdjt07:
> Are you facing the issue on a specific video? What services do you face the issue on (e.g. YouTube or SoundCloud)?

It's happened again today. This time I was trying to listen to this livestream as background (https://youtu.be/6vTb5Tmg14g).
I just can run as background when I am playing the video and then a press the android home button.
Every time I try to run using the "background option" I got the error but without trace log.

CBSkarmory:
Edited issue: the "steps to reproduce the bug" section was commented out

boringhexi:
Seems to have been introduced by dfa941a9e78a72a54801be9da6b8b79590121543, going by my testing. 

My steps for testing were: add Lofi Girl stream (currently [here](https://www.youtube.com/watch?v=5qap5aO4i9A)) to a new playlist, then Background play the playlist and see if I get an unrecoverable player error.

saurabhinabys:
> ### Checklist
> * [x]  I am using the latest version - 0.21.9
> * [x]  I checked, but didn't find any duplicates (open OR closed) of this issue in the repo.
[...]

PR #8020, opened 2022-03-12
ExoPlayer 2.17.1 update and MediaSource management rework

karyogamy:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Hi team,
I took a shot at upgrading ExoPlayer to 2.17.0 and realized that quite a few API has been made obsolete. For example, `getTag` has been dropped and will need even more hacks to get it to work. Since a rework is needed anyways, I took the liberty to toss out the old custom MediaSource management in favour of a more vanilla ExoPlayer model, following  [this comment](https://github.com/google/ExoPlayer/issues/7087#issuecomment-599581609). This is an attempt to help with refactoring the player so we can shift to expose ExoPlayer interface for player management instead of managing our own `PlayQueue` for playlist modifications.

This means ExoPlayer will now be responsible for managing the internal playlist, and `PlayQueue` will no longer play a central role to playback management. It will become a player proxy for user issued commands, and acts as a view for the player playlist, since ExoPlayer doesn't provide an API for the entire playlist. As an added bonus, shuffle now works properly. You may also notice buffer will start immediately when seeking to an unprepared stream, as opposed to the old model, where playback resumes until `StreamInfo` is fetched before buffering starts. This is because `StreamInfo` are now loaded on-demand inside the `MediaSource` in a custom `CompositeMediaSource` that dispatches a call to resolve/build the actual `MediaSource` using the `PlaybackResolvers`. 

Also, each `MediaSource` is now associated with a `MediaItem` that holds metadata. `Tag` objects are now contained inside this `MediaItem` instead. As a result, the old `MediaSourceTag` are now revamped into a generic interface that bridges between `PlayQueueItem` and `StreamInfo` since they both hold similar metadata. This can be expanded later so this interface object is exposed to external adapters and properly display if a stream is properly loaded or has errors (e.g. content error).

Going back to `MediaSource` management, a challenge arising from the new model is its lifecycle, where `MediaSource` in ExoPlayer aren't really allowed to self update. This means when a stream expires, we cannot update the `Playlist MediaSource` with a new one, nor should we since ExoPlayer should take care of it instead. This brings in `DataSpec` resolution, which is an interface ExoPlayer exposes that lets us examine an URI and replace it with another. So the solution here is to update our `PlaybackResolver` to encapsulate the Extractor Stream metadata (e.g. `AudioStream`, `VideoStream`, `SubtitleStream`, etc.) in an URI, and deliver that to the new `DataSpecResolver` to convert our custom URI to a real stream URI. This means we can rely on our own Extractor cache to handle expiration and resolve stream metadata into the real URL and ExoPlayer will handle the rest.

Concerns:
1. Because ExoPlayer now takes care of managing its own playlist, the `MediaSourceManager` (now `PlayQueueReactor`) no longer correct desynchronizations between the `PlayQueue` and ExoPlayer's own playlist. Since `PlayQueue` and `PlayQueueReactor` communicate with a message bus, non-atomic calls to `PlayQueue` can cause desyncs. For example, `Enqueue next` was broken because it works by appending to the queue and moving the last item to the current playing. When the append event message gets to the reactor, the item has already moved on the play queue. This is temporarily fixed by introducing a new event `Insert` to avoid mixing logic with auto-queuing on `Append` events. But a proper solution would be to refactor `PlayQueue` into an interface and create a new wrapper `ActivePlayQueue` on the base `PlayQueue` to handle all the player controlled play queue logic. The ultimate end-state for `PlayQueue` in this new model is to mainly act as a quick-access view for the adapters, and all actions to the queue are exposed via ExoPlayer interface instead.

2. Since ExoPlayer controls the shuffling now, we also lose some control (for the time being) in shuffled mode. For example, currently `PlayQueue` shuffle reorders the queue and is presented to the user (I actually like this better) instead of ExoPlayer's model of jumping to a random index. But this now also breaks the `Enqueue next` feature when used in shuffle mode. We can potentially create our own shuffle order as an individual component, though it means more code and kind of defeats the purpose of lifting the feature to ExoPlayer.

3. As mentioned before, `StreamInfo` preparation will now also be part of the buffering. This means, when selecting a stream that is not yet loadedm at all, buffering will feel a bit slower than before. Whereas before, the `StreamInfo` would load in the background while the stream continues to play before the buffering starts. I tried to mitigate this by pre-emptively loading the `StreamInfo` of the next `MediaItem`, but this issue will never fully go away since we can't load every item on the queue.

As you can see, there are quite a few more things that can be worked on to get this to a perfect state. But since the PR size is getting larger, I wanted to get this PR out now to get some opinion on if we want to make this shift in player modelling as part of the greater refactor work (or if this path isn't really worth proceeding). And if we want to make the shift, have the changes tested by more people and the code reviewed since it might break existing features.

#### APK testing 

- relate to https://github.com/TeamNewPipe/NewPipe/issues/7939

#### Due diligence


app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -270,25 +271,26 @@
...
     @Nullable private PlayQueueItem currentItem;
-    @Nullable private MediaSourceTag currentMetadata;
+    @Nullable private MediaItemTag currentMetadata;
     @Nullable private Bitmap currentThumbnail;
 
     /*//////////////////////////////////////////////////////////////////////////
     // Player
     //////////////////////////////////////////////////////////////////////////*/
 
-    private SimpleExoPlayer simpleExoPlayer;
```
karyogamy, line 283: `SimpleExoPlayer` is being deprecated.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -270,25 +271,26 @@
...
+    private ExoPlayer simpleExoPlayer;
     private AudioReactor audioReactor;
     private MediaSessionManager mediaSessionManager;
     @Nullable private SurfaceHolderCallback surfaceHolderCallback;
 
-    @NonNull private final CustomTrackSelector trackSelector;
```
karyogamy, line 288: `CustomTrackSelector` interface has conflicts now, `DefaultTrackSelector` already fixed the issue with language selection, so making the switch here.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -1980,17 +1983,39 @@ private void hideSystemUIIfNeeded() {
     //////////////////////////////////////////////////////////////////////////*/
     //region Playback states
 
-    @Override // exoplayer listener
-    public void onPlayerStateChanged(final boolean playWhenReady, final int playbackState) {
```
karyogamy, line 1984: `onPlayerStateChanged` interface is now broken into separate callbacks below.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4498,9 +4469,13 @@ private void setupVideoSurface() {
             surfaceHolderCallback = new SurfaceHolderCallback(context, simpleExoPlayer);
             binding.surfaceView.getHolder().addCallback(surfaceHolderCallback);
             final Surface surface = binding.surfaceView.getHolder().getSurface();
-            // initially set the surface manually otherwise
-            // onRenderedFirstFrame() will not be called
-            simpleExoPlayer.setVideoSurface(surface);
+            // ensure player is using an unreleased surface, which the surfaceView might not be
+            // when starting playback on background or during player switching
+            if (surface.isValid()) {
```
karyogamy, line 4474: @TiA4f8R This should fix the issue with starting livestream on background. Let me know if this works and if you want this in another PR instead.

app/src/main/java/org/schabi/newpipe/player/helper/CacheFactory.java
```diff
@@ -49,7 +49,9 @@ private CacheFactory(@NonNull final Context context,
                          final long maxFileSize) {
         this.maxFileSize = maxFileSize;
 
-        dataSourceFactory = new DefaultDataSourceFactory(context, userAgent, transferListener);
```
karyogamy, line 52: More ExoPlayer deprecation.

app/src/main/java/org/schabi/newpipe/player/helper/MediaSessionManager.java
```diff
@@ -19,7 +19,6 @@
 import org.schabi.newpipe.MainActivity;
 import org.schabi.newpipe.player.mediasession.MediaSessionCallback;
 import org.schabi.newpipe.player.mediasession.PlayQueueNavigator;
-import org.schabi.newpipe.player.mediasession.PlayQueuePlaybackController;
```
karyogamy, line 22: Deprecated and removed from Exo.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerDataSource.java
```diff
@@ -80,17 +83,6 @@ public PlayerDataSource(@NonNull final Context context, @NonNull final String us
         return new DefaultDashChunkSource.Factory(dataSourceFactory);
     }
 
-    public HlsMediaSource.Factory getHlsMediaSourceFactory() {
```
karyogamy, line 83: Removed because I don't think there is caching for livestreams. Let me know if otherwise to revert.

app/src/main/java/org/schabi/newpipe/player/mediasession/PlayQueueNavigator.java
```diff
@@ -44,17 +43,17 @@ public long getSupportedQueueNavigatorActions(@Nullable final Player player) {
     }
 
     @Override
-    public void onTimelineChanged(final Player player) {
+    public void onTimelineChanged(@NonNull final Player player) {
```
karyogamy, line 46: More deprecations here.

app/src/main/java/org/schabi/newpipe/player/resolver/ExtractorDataSourceResolver.java
```diff
@@ -0,0 +1,321 @@
...
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+public final class ExtractorDataSourceResolver implements ResolvingDataSource.Resolver {
```
litetex, line 32: I'm not sure if 3 interfaces and 4 classes inside that class are a good idea. Maybe create separate classes?

litetex, line 32: Why is this method final?

litetex (changes requested):
Thank you for the PR!

A ExoPlayer update was long overdue in my opinion, good that you picked it up.

Very good PR in my opinion, I like the coding style and the usage of new Java 8 features such as the Streams API 👍 

Didn't test it yet but I found some things in the review that should be processed:

Update: Also please have a look at the Sonarcloud warnings

app/src/main/java/org/schabi/newpipe/player/resolver/ExtractorDataSourceResolver.java
```diff
@@ -0,0 +1,321 @@
...
+        public Uri resourceUriFrom(@NonNull final StreamInfo info, @NonNull final String protocol) {
+            final Uri manifestUri;
+            if (DASH_PROTOCOL.equals(protocol) && info.getDashMpdUrl() != null) {
+                manifestUri = Uri.parse(info.getDashMpdUrl());
+            } else if (HLS_PROTOCOL.equals(protocol) && info.getHlsUrl() != null) {
+                manifestUri = Uri.parse(info.getHlsUrl());
+            } else {
+                throw new NoSuchElementException("Protocol [" + protocol + "] not found.");
+            }
+            return manifestUri;
```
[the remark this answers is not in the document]
Redirion: subjective. I actually prefer this coding style with preferably one return statement or an Exception (which is done here) or at least as few as possible return statements if the code would otherwise get ugly.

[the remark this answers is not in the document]
litetex: Sorry but I think that 
```java
            if (DASH_PROTOCOL.equals(protocol) && info.getDashMpdUrl() != null) {
                return Uri.parse(info.getDashMpdUrl());
            } else if (HLS_PROTOCOL.equals(protocol) && info.getHlsUrl() != null) {
                return Uri.parse(info.getHlsUrl());
            }
            throw new NoSuchElementException("Protocol [" + protocol + "] not found.");
```
is a lot better readable then:
```java
            final Uri manifestUri;
            if (DASH_PROTOCOL.equals(protocol) && info.getDashMpdUrl() != null) {
                manifestUri = Uri.parse(info.getDashMpdUrl());
            } else if (HLS_PROTOCOL.equals(protocol) && info.getHlsUrl() != null) {
                manifestUri = Uri.parse(info.getHlsUrl());
            } else {
                throw new NoSuchElementException("Protocol [" + protocol + "] not found.");
            }
            return manifestUri;
```

Shorter, less expressions, less ``{ }``, same functionality.

karyogamy:
Many thanks for the thorough review @litetex. 

Given the serialization usage coincides with the `DataSpec` issue that might conflict with @TiA4f8R's `delivery-methods-v2`, I'd recommend not going forward with this PR as-is, and instead continue with the existing model where we have control `MediaSource`. The player changes though will largely remain the same, so the review is still effective. I've acked most of the change requests, and will make the change and add docs when time permits. 

Eventually when we decide to move to the ExoPlayer model, we will have the same issue using `DataSpec`s and need to choose to go with serialization or cache.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -1950,10 +1962,12 @@ private void showOrHideButtons() {
         final boolean showNext = playQueue.getIndex() + 1 != playQueue.getStreams().size();
         final boolean showQueue = playQueue.getStreams().size() > 1 && !popupPlayerSelected();
         boolean showSegment = false;
-        if (currentMetadata != null) {
-            showSegment = !currentMetadata.getMetadata().getStreamSegments().isEmpty()
-                    && !popupPlayerSelected();
-        }
+        showSegment = /*only when stream has segment and playing in fullscreen player*/
```
litetex, line 1965: Maybe also do that in the line above?

litetex, line 1965:
```suggestion
        showSegment = /*only when stream has segments and is not playing in popup player*/
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -1993,9 +2007,30 @@ private void hideSystemUIIfNeeded() {
...
     //////////////////////////////////////////////////////////////////////////*/
     //region Playback states
+    @Override
+    public void onPlayWhenReadyChanged(final boolean playWhenReady, final int reason) {
+        if (DEBUG) {
+            Log.d(TAG, "ExoPlayer - onPlayWhenReadyChanged() called with: "
+                    + "playWhenReady = [" + playWhenReady + "], "
+                    + "reason = [" + reason + "]");
+        }
+        final int playbackState = simpleExoPlayer == null
```
litetex, line 2017: Use `exoPlayerIsNull()`

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -1993,9 +2007,30 @@ private void hideSystemUIIfNeeded() {
...
 
-    @Override // exoplayer listener
-    public void onPlayerStateChanged(final boolean playWhenReady, final int playbackState) {
+    @Override
+    public void onPlaybackStateChanged(final int playbackState) {
+        if (DEBUG) {
+            Log.d(TAG, "ExoPlayer - onPlaybackStateChanged() called with: "
+                    + "playbackState = [" + playbackState + "]");
+        }
+        final boolean playWhenReady = simpleExoPlayer != null && simpleExoPlayer.getPlayWhenReady();
```
litetex, line 2029: Use `getPlayWhenReady()` here (no variable required then)

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2461,26 +2487,37 @@ private void setMuteButton(@NonNull final ImageButton button, final boolean isMu
...
-        if (DEBUG) {
-            Log.d(TAG, "ExoPlayer - onTimelineChanged() called with "
-                    + "timeline size = [" + timeline.getWindowCount() + "], "
-                    + "reason = [" + reason + "]");
-        }
-
-        maybeUpdateCurrentMetadata();
-        // force recreate notification to ensure seek bar is shown when preparation finishes
-        NotificationUtil.getInstance().createNotificationIfNeededAndUpdate(this, true);
+    public void onEvents(@NonNull final com.google.android.exoplayer2.Player player,
```
litetex, line 2490: Still waiting for the documentation what is happening here...

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2559,94 +2599,106 @@ public void onCues(@NonNull final List<Cue> cues) {
...
-     */
+     * @see com.google.android.exoplayer2.Player.Listener#onPlayerError(PlaybackException)
+     * */
+    @SuppressLint("SwitchIntDef")
     @Override
-    public void onPlayerError(@NonNull final ExoPlaybackException error) {
+    public void onPlayerError(@NonNull final PlaybackException error) {
         Log.e(TAG, "ExoPlayer - onPlayerError() called with:", error);
 
+        setRecovery();
```
litetex, line 2630: Why is setRecovery called here (a comment in the code would be great)

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3410,12 +3412,15 @@ private void updateStreamRelatedViews() {
                 break;
 
             case VIDEO_STREAM:
-                if (info.getVideoStreams().size() + info.getVideoOnlyStreams().size() == 0) {
+                if (currentMetadata == null
+                        || !currentMetadata.getMaybeQuality().isPresent()
+                        || info.getVideoStreams().size() + info.getVideoOnlyStreams().size() == 0) {
```
litetex, line 3417:
```suggestion
                        || (info.getVideoStreams().isEmpty() && info.getVideoOnlyStreams().isEmpty())) {
```

I would write it like this - better understandable IMHO

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3679,7 +3684,10 @@ private void onTextTracksChanged() {
         }
 
         // Normalize mismatching language strings
-        final String preferredLanguage = trackSelector.getPreferredTextLanguage();
+        final List<String> preferredLanguages =
+                trackSelector.getParameters().preferredTextLanguages;
+        final String preferredLanguage =
+                preferredLanguages.isEmpty() ? null : preferredLanguages.get(0);
```
litetex, line 3690:
```suggestion
        final String preferredLanguage = trackSelector.getParameters().preferredTextLanguages
                .stream().findFirst().orElse(null);
```

Would write it like this (no additional variable required)

app/src/main/java/org/schabi/newpipe/player/mediaitem/MediaItemTag.java
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Metadata container and accessor used by player internals.
+ *
+ * This interface ensures consistency of fetching metadata on each stream,
+ * which is encapsulated in a {@link MediaItem} and delivered via ExoPlayer's
+ * {@link Player.Listener} on event triggers to the downstream users.
+ **/
+public interface MediaItemTag {
+
+    List<Throwable> getErrors();
```
litetex, line 29: I'm pretty sure that these should be Exceptions.
https://rules.sonarsource.com/java/RSPEC-1181

app/src/main/java/org/schabi/newpipe/player/mediasource/FailedMediaSource.java
```diff
@@ -2,88 +2,162 @@
...
+     * while still provide a smooth playback experience. A duration lower than 1 second is
+     * not recommended, it may cause ExoPlayer to buffer for a while.
+     * */
+    public static final long SILENCE_DURATION_US = TimeUnit.SECONDS.toMicros(2);
+    public static final MediaPeriod SILENT_MEDIA = makeSilentMediaPeriod(SILENCE_DURATION_US);
+
     private final String TAG = "FailedMediaSource@" + Integer.toHexString(hashCode());
     private final PlayQueueItem playQueueItem;
-    private final FailedMediaSourceException error;
+    private final Throwable error;
```
litetex, line 39: https://rules.sonarsource.com/java/RSPEC-1181

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2984,10 +3010,10 @@ public void saveStreamProgressState() {
     }
 
     public void saveStreamProgressStateCompleted() {
-        if (currentMetadata != null) {
+        getCurrentStreamInfo().ifPresent(info -> {
             // current stream has ended, so the progress is its duration (+1 to overcome rounding)
-            saveStreamProgressState((currentMetadata.getMetadata().getDuration() + 1) * 1000);
-        }
+            saveStreamProgressState((info.getDuration() + 1) * 1000);
+        });
```
litetex, line 3016: https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=8020&id=TeamNewPipe_NewPipe&open=AX9__pDcgqUh_HHl9bWr

https://rules.sonarsource.com/java/RSPEC-1602

litetex (changes requested):
Some minor things I noticed:

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2460,27 +2484,51 @@ private void setMuteButton(@NonNull final ImageButton button, final boolean isMu
     //////////////////////////////////////////////////////////////////////////*/
     //region ExoPlayer listeners (that didn't fit in other categories)
 
+    /**
+     * <p>Listens for event or state changes on ExoPlayer. When any event happens, we check for
```
litetex, line 2488: Shouldn't the <p> be also closed somewhere?

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2559,94 +2610,100 @@ public void onCues(@NonNull final List<Cue> cues) {
...
+     * For any error above that is <b>not</b> explicitly <b>catchable</b>, the player will
+     * create a notification so users are aware.
      * </ul>
-     *
-     * @see #processSourceError(IOException)
-     * @see com.google.android.exoplayer2.Player.Listener#onPlayerError(ExoPlaybackException)
-     */
+     * @see com.google.android.exoplayer2.Player.Listener#onPlayerError(PlaybackException)
+     * */
+    @SuppressLint("SwitchIntDef")
```
litetex, line 2637: A reason for the suppression would be good 😄

e.g.
```java
@SuppressLint("SwitchIntDef") // My cat doesn't like it,. No, seriously write a real reason here ;)
```

app/src/main/java/org/schabi/newpipe/player/mediasource/FailedMediaSource.java
```diff
@@ -2,88 +2,162 @@
...
+        return new FailedMediaSource(playQueueItem, error,
+                System.currentTimeMillis() + retryWaitMillis);
     }
 
     public PlayQueueItem getStream() {
         return playQueueItem;
     }
 
-    public FailedMediaSourceException getError() {
+    public Throwable getError() {
```
litetex, line 81: Shouldn't that be also an  ``Exception`` here?

app/src/main/java/org/schabi/newpipe/player/playback/MediaSourceManager.java
```diff
@@ -417,20 +423,27 @@ private void maybeLoadItem(@NonNull final PlayQueueItem item) {
...
-            return new LoadedMediaSource(source, stream, expiration);
-        }).onErrorReturn(throwable -> new FailedMediaSource(stream,
-                new StreamInfoLoadException(throwable)));
+            return new LoadedMediaSource(source, tag, stream, expiration);
+        }).onErrorReturn(throwable -> {
+            if (throwable instanceof ExtractionException) {
+                return FailedMediaSource.of(stream, new StreamInfoLoadException(throwable));
+            }
+            return FailedMediaSource
+                    .of(stream, new Exception(throwable), /*immediatelyRetryable=*/0L);
```
litetex, line 445: ``/*immediatelyRetryable=*/0L``
Couldn't that lead to a deadlock/loop?

litetex (changes requested):
Nearly perfect, only minor things:

litetex (approved):
LGTM

Found one problem while testing (however this is likely unrelated): When enqueuing a video it gets enqueued 2x.

Otherwise it looks like it' working fine 👍

karyogamy:
Other issues that should be fixed in tandem:
- Video progress resets during player switch: #7916
- Video Surface Released: #6989

TiA4f8R:
It seems I found a regression with this PR: use the Play all button on the [Creative Commons channel](https://www.youtube.com/c/CreativeCommonsVids/videos), open the video with Stacy Allison-Cassin, enable autogenerated subtitles, then switch to the next video. Wait for lyrics and you will notice that English subtitles are enabled, even if subtitles are marked as disabled by the player.

karyogamy:
Confirmed - can reproduce by queuing any 2 videos, first with a manual subtitle (e.g. english) and the second with only auto-generated subtitle of the previous language (e.g. english auto-generated) and the rest is the same as @TiA4f8R describes. 

I suspect this is because of how ExoPlayer compare language names, we can probably fix this by updating the selected ExoPlayer subtitle track according to the selected caption dropdown on metadata update. Will take a look when next available.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2460,27 +2484,51 @@ private void setMuteButton(@NonNull final ImageButton button, final boolean isMu
...
-    public void onTimelineChanged(@NonNull final Timeline timeline, final int reason) {
-        if (DEBUG) {
-            Log.d(TAG, "ExoPlayer - onTimelineChanged() called with "
-                    + "timeline size = [" + timeline.getWindowCount() + "], "
-                    + "reason = [" + reason + "]");
-        }
-
-        maybeUpdateCurrentMetadata();
-        // force recreate notification to ensure seek bar is shown when preparation finishes
-        NotificationUtil.getInstance().createNotificationIfNeededAndUpdate(this, true);
```
Stypox, line 2473: Why was this removed @karyogamy?

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -4519,9 +4538,13 @@ private void setupVideoSurface() {
...
             final Surface surface = binding.surfaceView.getHolder().getSurface();
-            // initially set the surface manually otherwise
-            // onRenderedFirstFrame() will not be called
-            simpleExoPlayer.setVideoSurface(surface);
+            // ensure player is using an unreleased surface, which the surfaceView might not be
+            // when starting playback on background or during player switching
+            if (surface.isValid()) {
+                // initially set the surface manually otherwise
+                // onRenderedFirstFrame() will not be called
+                simpleExoPlayer.setVideoSurface(surface);
```
Stypox, line 4546: Could this be replaced by `simpleExoPlayer.setVideoSurfaceHolder(binding.surfaceView.getHolder());`, or would there be issues? After some testing in #8170 I couldn't notice any difference.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3043,48 +3071,29 @@ private void onMetadataChanged(@NonNull final MediaSourceTag tag) {
...
-        maybeAutoQueueNextStream(metadata);
-
-        if (currentMetadata == metadata) {
-            return;
-        }
-        currentMetadata = metadata;
-        onMetadataChanged(metadata);
+        maybeAutoQueueNextStream(streamInfo);
+        onMetadataChanged(streamInfo);
+        NotificationUtil.getInstance().createNotificationIfNeededAndUpdate(this, true);
```
Stypox, line 3081: Also, why does this now have `forceRecreate = true`? Also, it seems like you are calling `createNotificationIfNeededAndUpdate` twice when `updateMetadataWith` is called, since there is a `createNotificationIfNeededAndUpdate(this, false);` also inside `onMetadataChanged`. And as you can see that one has `forceRecreate = false`.

app/src/main/java/org/schabi/newpipe/player/Player.java
karyogamy: Updates still happen on timeline changes. This method, and other calls to `maybeUpdateCurrentMetadata` are now conslidated into `onEvents` which is triggered on any ExoPlayer change instead of scattering them into different callbacks.

app/src/main/java/org/schabi/newpipe/player/Player.java
karyogamy: Oversight on my part. `createNotificationIfNeededAndUpdate ` in `onMetadataChanged` should be removed.

app/src/main/java/org/schabi/newpipe/player/Player.java
karyogamy: Absolutely, that is the better solution. `setVideoSurfaceHolder` is doing the same null checks here: https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImpl.java#L1262

karyogamy:
Sorry about that folks, got too used to Stash PR workflow and didn't realize on Github all my review replies stayed in `Pending` and never got out =\

PR #8064, opened 2022-03-18
Simplify some sharing utilities

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- These are handled internally by the methods I implemented in the PR.
  - At least I'm pretty sure.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
@@ -8,7 +8,6 @@ import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.content.pm.ResolveInfo;
 import android.net.Uri;
-import android.os.Build;
 import android.text.TextUtils;
 import android.widget.Toast;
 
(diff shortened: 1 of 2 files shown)


TiA4f8R:
Sorry, but I will probably close this PR.

Let me explain why: you are reverting my share sheet improvements on EMUI devices, which allow these devices to get the preview title (and soon the thumbnail of the content) of the content shared by using Android system chooser instead of EMUI chooser. See #5187 (the PR which introduced the changes) for the difference.

I know that's something we didn't have to do if Huawei didn't change this, but as you can see, the improvement by the use of the Android share sheet is really big in terms of information show to the user.

Hope you can understand.

opusforlife2:
@TiA4f8R Is the Huawei-specific code marked as such? Or do you have a comment somewhere which explains why that code shouldn't be touched?

TacoTheDank:
image

Curses, foiled again! Damn you Android fragmentation!

TacoTheDank:
But yeah I understand, I didn't even know that was intentional (I guess I should've looked at the git blame though).

Perhaps some code commenting is in order?

(Also @TiA4f8R did you install the debug build and see if this PR actually did screw up the share sheet (seeing as you have an EMUI device)? I'd check, just in case. If you already did, then alrighty. 👍)

TiA4f8R:
@TacoTheDank I did and the Android share sheet is still shown when sharing a content (for an unknown reason).

But the EMUI one is back when using the `Open in browser` command or when sharing a download (which was not the case, even before my PR (the Android one was shown, unless Huawei did more changes in the `Intent` class with a security patch) (it tries to generate an `hwCHOOSER` intent first and fallbacks to the Android one if the first one is null).

cketti:
What is special about EMUI? Does `Intent.createChooser()` create an intent that is different from a manually created `Intent.ACTION_CHOOSER` intent?

TiA4f8R:
> What is special about EMUI? Does `Intent.createChooser()` create an intent that is different from a manually created `Intent.ACTION_CHOOSER` intent?

Yes, see my comment above:

> it tries to generate an hwCHOOSER intent first and fallbacks to the Android one if the first one is null

(where the Android one is `Intent.ACTION_CHOOSER`)

litetex:
I think we can close this or?

TacoTheDank:
I'll close it.

PR #8134, opened 2022-04-02
Limit speed at live edge

### LINKED ISSUE #7667: Livestreams should not have playback parameters (speed adjustment or skip slience)
### Checklist
- [x] I am using the latest version - 0.21.15

1. Open a livestream

### Actual behavior
Notice that you can change the speed / skip silence / etc

### Expected behavior
None of the options should be available and a livestream, because it is LIVE.

grafik

KosOrKosm:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

- Upon reaching the live edge of a livestream, if playback speed is greater than 100%, set it to 100%.
- The maximum playback speed that can be set in the PlaybackParameterDialog is reduced to 100% when on the live edge.
- The PlaybackParameterDialog skip silence button is also hidden on the live edge.

Note it is still possible to use playbacks speeds <= 100% at all times on livestreams, and playback speeds > 100% when not on the live edge.

Changes tested on emulator (Pixel 3a w/ API 28).

#### Before/After Screenshots
| Before: | After: |
| PR2 - NonlivePlaybackParamDialog | PR2 - LivePlaybackParamDialog |

- Fixes #7667

#### APK testing 

#### Due diligence


TiA4f8R:
Thank you for the pull request!

I know I am bit against these changes, but what do you think of setting only the playback speed to 1x by default only on livestreams? This will allow users to watch streams from the start of the live window up to the maximum speed supported in the app (see what I wrote in https://github.com/TeamNewPipe/NewPipe/issues/7667#issuecomment-1082296021 for more details).

So if users increased playback speed and they have issues when using it at the live edge, then they can understand the issue is from them, not from us.

KosOrKosm:
@TiA4f8R Thank you for the feedback!

I was considering setting the default playback speed to 1x for livestreams, but I thought this solution might be less intrusive. This approach allows the user to speed up in the live window, just not in the live edge. In my testing, Youtube livestreams had an ~30 second live window, and only ~5 seconds of that window was the live edge. So only those last ~5 seconds cannot be sped up, and the rest of the window can still be freely controlled. I think this allows the user the most freedom without causing an unpleasant experience due to buffering at the live edge.

In terms of notifying the user, I wasn't sure if any notification was necessary. It could be surprising for the user that the speed resets at the live edge, but I also think it might be clear why it occurs. Perhaps a popup message could appear when the speed is reset to explain why it is happening?

TiA4f8R:
> In my testing, Youtube livestreams had an ~30 second live window, and only ~5 seconds of that window was the live edge.

This will be changed with my PR, depending of the streams, up to one hour (for livestreams with DVR of course).

> This approach allows the user to speed up in the live window, just not in the live edge

Why not using what YouTube does? When coming to the live edge or being behind the live window, YouTube player(s) reset(s) the playback speed to 1x and seek(s) to the default position. This would solve all the issues without preventing users to select the playback speed. In this case, of course, 1x should be not saved as the default playback speed for videos.

KosOrKosm:
> Why not using what YouTube does? 
I had also considered this method. I preferred locking to the default seek position (the current implementation), rather than allowing the user to briefly accelerate past it only to send them back moments later once buffering occurs.

I can use the same strategy as Youtube simply by moving the new logic from onUpdateProgress to onBuffering + adding a seek to default. Would this be the best / preferred approach?

KosOrKosm:
Hello @litetex, can you review my pull request?

litetex:
Blocked by #7989

KosOrKosm:
Ahh, I see. @litetex  Could I close this PR and open a new one without the speed dialogue changes? I can just use the same strategy as Youtube and reset playback speed when buffering at the live edge. That will solve the underlying issue without interfering with any incoming changes to the dialogue.

Sorry, I am relatively new to Github, so I'm not sure if this would be appropriate.

Stypox:
@KosOrKosm no, you PR is perfectly fine, don't worry. What litetex was saying is probably that he wants to merge #7989 (which I will do now) before this PR since they both edit the same file and that causes merge conflicts. By merging #7989 I will create some merge conflicts for you, though, let me know if you need help with solving them ;-)

Stypox:
@KosOrKosm you might have to rewrite from scratch your changes to `PlaybackParameterDialog` since #7989 basically rewrote it from scratch, too.

KosOrKosm:
I reimplemented my tempo limit logic for `PlaybackParameterDialog` on the #7989 changes. I also merged with the latest version of dev.

app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java
```diff
@@ -63,13 +64,20 @@ public class PlaybackParameterDialog extends DialogFragment {
...
     private static final boolean DEFAULT_SKIP_SILENCE = false;
+    private static final boolean DEFAULT_LIVE_EDGE = false;
 
     private static final SliderStrategy QUADRATIC_STRATEGY = new SliderStrategy.Quadratic(
             MIN_PITCH_OR_SPEED,
             MAX_PITCH_OR_SPEED,
             1.00f,
             10_000);
 
+    private static final SliderStrategy QUADRATIC_LIVE_STRATEGY = new SliderStrategy.Quadratic(
```
Stypox, line 75: I'd call it `LIVE_EDGE_QUADRATIC_STRATEGY`

app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java
```diff
@@ -49,6 +49,7 @@ public class PlaybackParameterDialog extends DialogFragment {
     // Minimum allowable range in ExoPlayer
     private static final double MIN_PITCH_OR_SPEED = 0.10f;
     private static final double MAX_PITCH_OR_SPEED = 3.00f;
+    private static final double MAX_LIVE_PITCH_OR_SPEED = 1.00f;
```
Stypox, line 52: This only applies to speed, doesn't it? I'd call it `MAX_LIVE_EDGE_SPEED` then

app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java
```diff
@@ -533,7 +558,8 @@ private void setSliders(final double newValue) {
     private void setAndUpdateTempo(final double newTempo) {
         this.tempo = calcValidTempo(newTempo);
 
-        binding.tempoSeekbar.setProgress(QUADRATIC_STRATEGY.progressOf(tempo));
+        final SliderStrategy tempoStrategy = getTempoStrategy();
+        binding.tempoSeekbar.setProgress(tempoStrategy.progressOf(tempo));
```
Stypox, line 562:
```suggestion
        binding.tempoSeekbar.setProgress(getTempoStrategy().progressOf(tempo));
```

Stypox (changes requested):
Code looks almost good. Thank you!

KosOrKosm:
I pushed the requested changes.

app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java
```diff
@@ -91,20 +99,25 @@ public double valueOf(final int progress) {
     double initialPitchPercent = DEFAULT_PITCH_PERCENT;
     @State
     boolean initialSkipSilence = DEFAULT_SKIP_SILENCE;
+    @State
+    boolean initialLiveEdge = DEFAULT_LIVE_EDGE;
```
Stypox, line 103: Oh, I just noticed, probably `initialiLiveEdge` is not needed. The other "initial" variables are for parameters that the dialog can change and that the dialog should be able to restore to their initial values when the user presses "Reset". But in this case "live edge" is not something the user can change, it is just a state of the player, so there is no need to store its initial value because it doesn't make sense to restore it, and also because the `liveEdge` variable will never change throughout the lifetime of the dialog.

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -1717,6 +1733,10 @@ private void onUpdateProgress(final int currentProgress,
         }
         binding.playbackLiveSync.setClickable(!isLiveEdge());
 
+        if (isLiveEdge() && getPlaybackSpeed() > 1.0f) {
```
litetex, line 1736: Did you notice that this is run constantly in a loop?

litetex (changes requested):
Just scrolled down a bit, noticed problem, block catastrophe.

I will review/test this in more detail once I have time and higher priority PRs are done.

app/src/main/java/org/schabi/newpipe/player/Player.java
KosOrKosm: I did realize this is a function that is executing very frequently, but I wanted the playback speed to reset the moment the live edge was reached and wasn't sure if there was a better place to put the logic. 

This logic could be moved to `onBuffering()`, which would match Youtube's livestreaming behavior and avoid calling the check in a loop. The downside is that the speed reset won't occur immediately after hitting the live edge.

app/src/main/java/org/schabi/newpipe/player/helper/PlaybackParameterDialog.java
KosOrKosm: Ah, I overlooked that. I am not actually using the initialLiveEdge value anywhere. I will remove it.

app/src/main/java/org/schabi/newpipe/player/Player.java
Stypox: The more resource-heavy part here is probably [calculating `isLiveEdge`](https://github.com/TeamNewPipe/NewPipe/blob/81dd93e8bfd5dea32b777af03a42577094f24c9c/app/src/main/java/org/schabi/newpipe/player/Player.java#L2768), seeing it is not a trivial function.

Stypox, line 1736: Then probably we can just do this:
```suggestion

        // if live edge of a livestream -> no need for sync button; playback speed must be <= 1.0
        final boolean liveEdge = isLiveEdge();
        binding.playbackLiveSync.setClickable(!liveEdge);
        if (liveEdge && getPlaybackSpeed() > 1.0f) {
```

Stypox (changes requested):

GameKyuubi:
As someone who listens to live streams frequently, I strongly disagree with this change. I often use these features to catch up to the live point without missing anything. What *should* happen is those features should automatically disable themselves (reset to 1x) once the live point in the stream is hit. This way, the problem is resolved without removing features.

Stypox:
I'm closing this due to inactivity and due to the fact that we're soon starting the rewrite and we don't want to touch the player and possibly introduce regressions. Thank you anyway for your effort!

Issue #7811, opened 2022-02-05
Issue with different subtitles of the same language


AquaVirus:
### Checklist

- [x] I am using the latest version - 0.21.16 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

1. Start watching https://www.youtube.com/watch?v=wsQiKKfKxug in the app.
2. There are two different sets of English subtitles, "English" and "English - French Speakers Only", as can be seen if you try to watch the video in a web browser. Try to select the "English" subtitles.

### Actual behavior

I can only select the "French Speakers Only" subtitles, which is duplicated in the subtitle selection menu.

### Expected behavior

I should be able to pick between the two different subtitles.

### Device info

 - Android version/Custom ROM version: 11
 - Device model: Moto G9 Power

AquaVirus:
The subtitle download function seems to be unaffected. I can download the files for both subtitles just fine.

TiA4f8R:
@AquaVirus Does #8150 fixes this issue?

PR #8150, opened 2022-04-05
Fix caption auto-selection not reflected in player GUI

karyogamy:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This PR fixes an issue introduced by #8020, where `CustomTrackSelector` was replaced by `DefaultTrackSelector`, delegating preferred caption language selection to ExoPlayer instead. When transitioning from a video with manual subtitles selected, e.g. `english`, to a video with only auto-generated subtitles, e.g. `english (auto-generated)`, ExoPlayer would apply the previously selected language as preferred language and use the auto-generated subtitles instead of disabling captions. This is not reflected in the GUI since it is still trying to find that previously selected language in the list of available languages for the second video, which isn't there, so the GUI shows `No Captions`. This is fixed by looking at the currently select tracks `onTracksInfoChanged`, and show the ExoPlayer's selected tracks instead.

This change also fixes an unreported caption bug when selecting language variants, where [a video has one caption language with multiple formats in parentheses](https://www.youtube.com/watch?v=lheapd7bgLA), e.g. `english (auto-generated)` and `english (united kingdom)` in this order. When any variant of a language is selected (e.g. `english (auto-generated)`), the selected caption would jump to the last variant of a language group instead, in this case, `english (united kingdom)`. This happens because of parentheses detection logic when setting `userPreferredLanguage`, and is also fixed by delegating preferred language selection to ExoPlayer.

- Fixes https://github.com/TeamNewPipe/NewPipe/pull/8020#issuecomment-1086863037

#### APK testing 

#### Due diligence


TiA4f8R:
> This change also fixes an unreported caption bug when selecting language variant

It was reported :) See #7811.

karyogamy:
I think #7811 is referring to a different issue that is extractor related. For https://www.youtube.com/watch?v=wsQiKKfKxug, the actual subtitles are 'English' and 'English - French Speakers Only', but on the player GUI, only 'english' and 'english' are shown. Preferred text language update won't be able to help with that.

TiA4f8R:
Oh, ok, thanks for letting me know.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3667,41 +3661,44 @@ private void setupSubtitleView() {
...
-        final int textRenderer = getCaptionRendererIndex();
-
+    private void onTextTracksChanged(@NonNull final TracksInfo currentTrackInfo) {
         if (binding == null) {
             return;
         }
-        if (trackSelector.getCurrentMappedTrackInfo() == null
-                || textRenderer == RENDERER_UNAVAILABLE) {
+
+        if (!currentTrackInfo.isTypeSupportedOrEmpty(C.TRACK_TYPE_TEXT)) {
```
Redirion, line 3669: when there is no subtitle track available, this method would return true. I think you need to restore the previous conditions

Redirion:
I have tested a bit, so far so good. But now having english (united kingdom) and choosing a video with english (automatic) or vise versa would instead unselect the captions.

I preferred the previous behaviour actually. If I understood correctly you wanted to makre that if both (united kingdom) and (automatic) are available it would keep the previous. But this now feels like a downgrade when not both are available at the same time.

Redirion:
I have tested a bit, so far so good. But now having english (united kingdom) and choosing a video with english (automatic) or vise versa would instead unselect the captions.

I preferred the previous behaviour actually. If I understood correctly you wanted to makre that if both (united kingdom) and (automatic) are available it would keep the previous. But this now feels like a downgrade when not both are available at the same time.

karyogamy:
> But now having english (united kingdom) and choosing a video with english (automatic) or vise versa unselects the captions.

I think this is the default behavior for Youtube browser player, probably why it is replicated on ExoPlayer as default behavior.
This was remedied by that auto-generated caption seeking logic that converts `english (united kingdom)` to `english (automatic)`, but as mentioned, it has a bug that breaks selection on multiple language variants, which is problematic for [videos like this](https://www.youtube.com/watch?v=8iuLXODzL04) that has 6 variants of Chinese captions. 

This is now fixed by using multiple preferred languages in ExoPlayer caption support, though it requires parsing for the "stem" of language descriptor: e.g. `english (united kingdom)` -> `english`. So when `english (united kingdom)` is selected, both `english (united kingdom)` and `english` will be applied, in that order, and ExoPlayer will try to find a caption in the same order when the next video starts playing. Proper support of this should probably be included in the extractor, though that is out of scope on my  side time-wise, at least in the near future.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3668,41 +3664,45 @@ private void setupSubtitleView() {
...
+                .stream()
+                .filter(trackGroupInfo -> C.TRACK_TYPE_TEXT == trackGroupInfo.getTrackType())
+                .collect(Collectors.toList());
+        final List<String> availableLanguages = new ArrayList<>(textTracks.size());
+        for (int i = 0; i < textTracks.size(); i++) {
+            final TrackGroup textTrack = textTracks.get(i).getTrackGroup();
             if (textTrack.length > 0) {
                 availableLanguages.add(textTrack.getFormat(0).language);
             }
         }
```
litetex, line 3690: I think this can be converted to java-streams api:

Example:
```java
...
final List<String> availableLanguages = textTracks 
  .filter(info -> info.getTrackGroup().length > 0) // Maybe use isEmpty if available
  .map(textTrack -> textTrack.getFormat(0))
  .collect(Collectors.toList())
```

Redirion:
Tested again.

I just tested a video with english (united kingdom) and then switched to another video which has both english (automatisch erzeugt) and english (united states).

It appears to be random / nondeterministic. Sometimes cc even get's competely unselected. I don't know what's happening there.

karyogamy:
> I just tested a video with english (united kingdom) and then switched to another video which has both english (automatisch erzeugt) and english (united states).

Thanks for testing, not sure how I missed this one.
It seems to be caused by auto-generated tracks sharing the same priority inside `DefaultTrackSelector` as manual tracks since both tracks have the same stems (`english`) with no track having an exactly matching language name as the preferred. Thankfully there is a way to distinguish auto-generated tracks in the resolver and tracks are now given different roles that ExoPlayer can use in track selection. The updated track selection order is left as a comment.

> Sometimes cc even get's competely unselected. 

I'm not able to reprod this one. Could you test the updated apk and provide some steps if this still happens?

app/src/main/java/org/schabi/newpipe/player/Player.java
karyogamy: Thanks, fixed.

Redirion (approved):
Tested again, working now as expected. Thank you!

PR #8170, opened 2022-04-09
Refactor player and extract UI components

Stypox:
This big PR extracts ~2500 lines of code from the `Player`, namely the code that handles the various UI states, into plug-and-play ui classes. The UI classes I created so far are:
- `PlayerUi`: this is the base class with all of the **base methods** with an empty implementation. It holds a *reference to the player*.
- `NotificationPlayerUi`: still needs to be finished, but will contain all of the **calls to the `NotificationUtil`** class. Maybe we could merge it with `NotificationUtil`, in order to solve the singleton issues? **Tell me what you think**
- `VideoPlayerUi`: contains the code that **main and popup player have in common**, for example to handle controls, button presses, video surface, ... It also contains the *reference to the `VideoPlayerBinding`*, that is, the instantiation of `player.xml` that currently is the same for both the popup and the main player (and probably it's ok to be that way for now). This class is where most of the `Player` code went, as it is ~1500 lines long.
- `MainPlayerUi`: contains the code specific to the **main player**: handling fullscreen, stream segments, chapters & play queue list, minimizing to popup/background, main player gesture listener ...
- `PopupPlayerUi`: contains the code specific to the **popup player**: window management, close overlay, popup player gesture listener, ...

The `Player` object now holds a `PlayerUiList UIs` member variable through which UIs can be **added, removed or queried**. I used stream API and `Optional` to simplify things: when the player wants to call some method on all of its UIs it just has to call e.g. `UIs.call(PlayerUi::onComplete);`. And when classes from outside want to make changes to a specific UI, they can do `player.UIs().get(MainPlayerUi.class)` and use the `Optional` API on the returned object.

`MainPlayer` was renamed to `PlayerService` as that's what it actually is.

#### Highlights

- Now the background player doesn't keep all of the player views loaded in the background, since the `Main/PopupPlayerUi` is not added when starting the player in background, and removed when switching to it!
- The `PlayerGestureListener` class made no sense before the refactoring: why make both a base abstract class and a subclass, if they need to keep interchanging information and there is no difference in the nature of the code in the two classes?
- The exo player's video surface can be setup only when the surface view root has a parent and is not just a view floating around. But other than that, the `setupVideoSurfaceIfNeeded` method can be called anytime without issues, since the player works fine even without surface, and will start rendering happily as soon as a valid surface is connected. Btw, for Android >= API23, I used `setVideoSurfaceHolder` instead of manually extracting the surface from the holder and then calling `setVideoSurface`, see `VideoPlayerUi` line ~1500 (@Redirion is this ok)?
- When switching from main to popup player and viceversa, the view binding is reused, like before.
- For some reason the queue activity was using the queue adapter of the player to show the queue items in the list... Now it uses its own adapter as expected.

#### TODO / known issues
- [ ] Fix all `TODO TODO` comments
- [ ] Investigate why the initial video stream loading is far slower than when switching resolution
- [ ] More notification calls to `NotificationPlayerUi`
- [ ] Understand what to do about `BackgroundPlayerUi`: it can probably be removed
- [ ] Maybe turn fragment/activity player listeners into `ListenerPlayerUi` or something like that
- [ ] Remove the playback speeds `PopupMenu` in the popup player and just use the usual `PlaybackParametersDialog`, to save a few lines of code, complexity and inconsistency.
- [ ] Reenable checkstyle (to be done at the end)
- [ ] Improve `//region` comments
- [ ] Test tést tEST Test tesT tEst tèSt

Fixes part of #7939

#### Please test!
Could you **test** and tell me all of the issues you can find? The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

<sub>I removed other parts of the template as there is already too much text here...</sub>


opusforlife2:
And so it begins.

> Could you test and tell me all of the issues you can find?

Do you want bug reports only related to this PR? Or bug reports for the player in general, like playback exceptions?

Stypox:
@opusforlife2 possibly only related to this PR, but if in doubt send anyway.

opusforlife2:
Do you intend to fix some bugs along the way or is this strictly a 1:1 refactor without any change in what the code actually does, bugs included?

Stypox:
The code had to change in many places, and so what I rewritten may work differently than before and solve some bugs. But the point of the PR is the refactor, not the bug fixing (which will hopefully become simpler after this PR is merged).

Stypox:
I rebased on top of dev (which contained #8020, and I have a few questions about that PR I posted there). 

I used the first diff below here when rebasing, so that I could make sure that I didn't miss any changes in the `Player` class. You can take a look at the second diff for a comparison.
1. The changes that happened from the `dev` commit this PR was based on before rebasing (ac00c8f6aebb3164fababe175d80eac15c2d6bf9) and the `dev` commit this PR is currently based on after rebasing (4040cb36bb8280fa6fc218e23414a8d22d2fb45b): https://github.com/TeamNewPipe/NewPipe/compare/ac00c8f6aebb3164fababe175d80eac15c2d6bf9...4040cb36bb8280fa6fc218e23414a8d22d2fb45b?expand=1
2. The diff between this PR before rebasing and this PR after rebasing: https://github.com/TeamNewPipe/NewPipe/compare/b9e6a6a411a2e35a91c2e2a0ee32a3b1e9f5b67f...1d3dd1545951347cffb9ed78d79374bc2073068e?expand=1

Stypox:
I finished the TODO list, now this PR is completely ready for review (to be done at the beginning of the next release, if possible).
I didn't solve all of the Sonar issues as most were only related to the code I moved, not the code I wrote anew. In particular the two security hotspots are just `sendBroadcast` being called in a way that allows all apps to see the broadcast, but that's not an issue that should be solved here.

Redirion:
which broadcasts do you mean? Just create the Intent not directly in sendBroadcast and call intent.setPackage("org.schabi.newpipe") -> "security" problem solved

app/src/main/java/org/schabi/newpipe/player/gesture/MainPlayerGestureListener.kt
```diff
@@ -0,0 +1,224 @@
...
+
+    override fun onScrollEnd(event: MotionEvent) {
+        super.onScrollEnd(event)
+        if (binding.volumeRelativeLayout.visibility == View.VISIBLE) {
+            binding.volumeRelativeLayout.animate(false, 200, AnimationType.SCALE_AND_ALPHA, 200)
+        }
+        if (binding.brightnessRelativeLayout.visibility == View.VISIBLE) {
+            binding.brightnessRelativeLayout.animate(false, 200, AnimationType.SCALE_AND_ALPHA, 200)
+        }
+    }
```
litetex, line 139: Could ``visibility == View.VISIBLE`` be replaced by ``isVisible``? (applies to whole class)

app/src/main/java/org/schabi/newpipe/player/gesture/PopupPlayerGestureListener.kt
```diff
@@ -0,0 +1,287 @@
...
+        if (playerUi.isInsideClosingRadius(event)) {
+            playerUi.closePopup()
+        } else if (!playerUi.isPopupClosing) {
+            playerUi.closeOverlayBinding.closeButton.animate(false, 200)
+            binding.closingOverlay.animate(false, 200)
+        }
+    }
+
+    private fun handleMultiDrag(event: MotionEvent): Boolean {
+        if (initPointerDistance != -1.0 && event.pointerCount == 2) {
```
litetex, line 109: I would revert that if and do a quick exit with ``return false`` here

app/src/main/java/org/schabi/newpipe/player/gesture/PopupPlayerGestureListener.kt
```diff
@@ -0,0 +1,287 @@
...
+        return when {
+            e.x < playerUi.popupLayoutParams.width / 2.0 -> DisplayPortion.LEFT_HALF
+            else -> DisplayPortion.RIGHT_HALF
+        }
+    }
+
+    companion object {
+        private val TAG = PopupPlayerGestureListener::class.java.simpleName
+        private val DEBUG = MainActivity.DEBUG
+        private const val TOSS_FLING_VELOCITY = 2500
```
litetex, line 285: What is this for? Maybe a short jdoc comment?

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -546,27 +544,26 @@ public static WindowManager.LayoutParams retrievePopupLayoutParamsFromPrefs(
...
-    public static void savePopupPositionAndSizeToPrefs(final Player player) {
-        if (player.getPopupLayoutParams() != null) {
-            player.getPrefs().edit()
-                    .putFloat(player.getContext().getString(R.string.popup_saved_width_key),
-                            player.getPopupLayoutParams().width)
-                    .putInt(player.getContext().getString(R.string.popup_saved_x_key),
-                            player.getPopupLayoutParams().x)
-                    .putInt(player.getContext().getString(R.string.popup_saved_y_key),
-                            player.getPopupLayoutParams().y)
+    public static void savePopupPositionAndSizeToPrefs(final PopupPlayerUi playerUi) {
```
litetex, line 557: Sounds like this and ``retrievePopupLayoutParamsFromPrefs`` should be inside the PopupPlayer or a PopupPlayerHelper.
I'm not sure if this is even accessed outside the Player, so I think it might be good if it's encapsulated/non-public.

app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java
```diff
@@ -0,0 +1,988 @@
...
+        binding.topControls.setClickable(true);
+        binding.topControls.setFocusable(true);
+
+        if (isFullscreen) {
+            binding.titleTextView.setVisibility(View.VISIBLE);
+            binding.channelTextView.setVisibility(View.VISIBLE);
+        } else {
+            binding.titleTextView.setVisibility(View.GONE);
+            binding.channelTextView.setVisibility(View.GONE);
+        }
```
litetex, line 271: I think this should be replaced by ``binding.titleTextView.setVisibility(isFullscreen ? View.VISIBLE : View.GONE);`` and so on:
It's shorter and easier to read in my opinion.

app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java
```diff
@@ -0,0 +1,988 @@
...
+    protected float calculateMaxEndScreenThumbnailHeight(@NonNull final Bitmap bitmap) {
+        final int screenHeight = context.getResources().getDisplayMetrics().heightPixels;
+
+        if (DeviceUtils.isTv(context) && !isFullscreen()) {
+            final int videoInfoHeight =
+                    DeviceUtils.dpToPx(85, context) + DeviceUtils.spToPx(16, context);
+            return Math.min(bitmap.getHeight(), screenHeight - videoInfoHeight);
+        } else if (DeviceUtils.isTablet(context) && isLandscape() && !isFullscreen()) {
+            final int videoInfoHeight =
+                    DeviceUtils.dpToPx(85, context) + DeviceUtils.spToPx(15, context);
```
litetex, line 482: Where are the "magical" constants 85, 16 and 15 coming from?
Also the tv and table procedures are nearly identical, maybe we should abstract them?

app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java
```diff
@@ -0,0 +1,988 @@
...
+
+        if (isFullscreen) {
+            binding.titleTextView.setVisibility(View.VISIBLE);
+            binding.channelTextView.setVisibility(View.VISIBLE);
+            binding.playerCloseButton.setVisibility(View.GONE);
+        } else {
+            binding.titleTextView.setVisibility(View.GONE);
+            binding.channelTextView.setVisibility(View.GONE);
+            binding.playerCloseButton.setVisibility(View.VISIBLE);
+        }
```
litetex, line 945: Same as above, I would also replace this by  ``binding.titleTextView.setVisibility(isFullscreen ? View.VISIBLE : View.GONE);``

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.java
```diff
@@ -0,0 +1,77 @@
...
+                .map(playerUiType::cast)
+                .findFirst();
+    }
+
+    /**
+     * Calls the provided consumer on all player UIs in the list.
+     * @param consumer the consumer to call with player UIs
+     */
+    public void call(final Consumer<PlayerUi> consumer) {
+        //noinspection SimplifyStreamApiCallChains
```
litetex, line 74: Why is this ignore? (Comment in code would be good)

app/src/main/java/org/schabi/newpipe/player/ui/PopupPlayerUi.java
```diff
@@ -0,0 +1,485 @@
...
+
+        animatePopupOverlayAndFinishService();
+    }
+
+    public boolean isPopupClosing() {
+        return isPopupClosing;
+    }
+
+    public void removePopupFromView() {
+        // wrap in try-catch since it could sometimes generate errors randomly
```
litetex, line 320: Are the random errors new or have they been there before?

If they were there before we should link the corresponding issue.
If they haven't we should maybe create a new issue.

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
```diff
@@ -0,0 +1,1578 @@
...
+        PopupMenu.OnMenuItemClickListener, PopupMenu.OnDismissListener {
+    private static final String TAG = VideoPlayerUi.class.getSimpleName();
+
+    // time constants
+    public static final long DEFAULT_CONTROLS_DURATION = 300; // 300 millis
+    public static final long DEFAULT_CONTROLS_HIDE_TIME = 2000;  // 2 Seconds
+    public static final long DPAD_CONTROLS_HIDE_TIME = 7000;  // 7 Seconds
+    public static final int SEEK_OVERLAY_DURATION = 450; // 450 millis
+
+    // other constants (TODO remove playback speeds and use normal menu for popup, too)
```
litetex, line 98: TODO detected

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
```diff
@@ -0,0 +1,1578 @@
...
+                break;
+            case REPEAT_MODE_ONE:
+                binding.repeatButton.setImageResource(R.drawable.exo_controls_repeat_one);
+                break;
+            case REPEAT_MODE_ALL:
+                binding.repeatButton.setImageResource(R.drawable.exo_controls_repeat_all);
+                break;
+            default:
+                break; // unreachable
+        }
```
litetex, line 927: Optional: I think an if else if else block would be shorter here.

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
```diff
@@ -0,0 +1,1578 @@
...
+    @Override
+    public void onPlaybackParametersChanged(@NonNull final PlaybackParameters playbackParameters) {
+        super.onPlaybackParametersChanged(playbackParameters);
+        binding.playbackSpeed.setText(formatSpeed(playbackParameters.speed));
+    }
+
+    @Override
+    public void onRenderedFirstFrame() {
+        super.onRenderedFirstFrame();
+        //TODO check if this causes black screen when switching to fullscreen
```
litetex, line 967: TODO detected

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
```diff
@@ -0,0 +1,1578 @@
...
+    //region Popup menus ("popup" means that they pop up, not that they belong to the popup player)
+
+    private void buildQualityMenu() {
+        if (qualityPopupMenu == null) {
+            return;
+        }
+        qualityPopupMenu.getMenu().removeGroup(POPUP_MENU_ID_QUALITY);
+
+        @Nullable final List<VideoStream> availableStreams
+                = Optional.ofNullable(player.getCurrentMetadata())
```
litetex, line 1060: I think the ``=`` should be one line higher

litetex (changes requested):
Thank you for the PR!
This cleanup was long overdue and it has max priority on my personal PR todo-list (so that we don't end up with a ton of merge-conflicts).

Disclaimer: Didn't test it so far.

Your coding style is really good, found only some minor things while reviewing (might be leftovers from the migration/previous code):

Also please rebase the PR (merge conflicts).

PS: You might also have a look at Sonarqube

Stypox:
@litetex I won't do much to this PR until 0.23.1 with delivery methods is released, as we decided a while ago. This PR will go into 0.24.0, which will also be the first version to drop support for KitKat.

app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistDialog.java
```diff
@@ -147,4 +152,30 @@ public static Disposable createCorrespondingDialog(
...
+     * @param fragmentManager the fragment manager to use to show the dialog
+     * @return the disposable that was created
+     */
+    public static Disposable showForPlayQueue(
+            final Player player,
+            @NonNull final FragmentManager fragmentManager) {
+
+        final List<StreamEntity> streamEntities = Stream.of(player.getPlayQueue())
+                .filter(Objects::nonNull)
+                .flatMap(playQueue -> Objects.requireNonNull(playQueue).getStreams().stream())
```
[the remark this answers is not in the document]
Stypox: Android studio problems... I agree with you it can't be null due to filtering. I removed it.

app/src/main/java/org/schabi/newpipe/player/PlayerService.java
```diff
@@ -0,0 +1,159 @@
...
+    @Override
+    public int onStartCommand(final Intent intent, final int flags, final int startId) {
+        if (DEBUG) {
+            Log.d(TAG, "onStartCommand() called with: intent = [" + intent
+                    + "], flags = [" + flags + "], startId = [" + startId + "]");
+        }
+
+        if (!Intent.ACTION_MEDIA_BUTTON.equals(intent.getAction())
+                || player.getPlayQueue() != null) {
+            // ^ no need to process media button's action if player is not working
```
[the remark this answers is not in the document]
Stypox: It was introduced here, I'm not gonna change it, but I'm gonna improve the comment: e41218c46b54e35c9d9aad9608891adb5fea12d2

app/src/main/java/org/schabi/newpipe/player/PlayerService.java
```diff
@@ -0,0 +1,159 @@
...
+
+            // Releases wifi & cpu, disables keepScreenOn, etc.
+            // We can't just pause the player here because it will make transition
+            // from one stream to a new stream not smooth
+            player.smoothStopForImmediateReusing();
+
+            // Notification shows information about old stream but if a user selects
+            // a stream from backStack it's not actual anymore
+            // So we should hide the notification at all.
+            // When autoplay enabled such notification flashing is annoying so skip this case
```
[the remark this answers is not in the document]
Stypox: It is my fault in a previous PR: cf9b482be25b412f183635cb0fc4bfc877c7d035 :grimacing:. It is like this on `dev`, too. When it was introduced it had a meaning: [`d1609cb`](https://github.com/TeamNewPipe/NewPipe/commit/d1609cba90006e18a0dde03f09d133aedf73e985#diff-37a86e64fd7d66a570bb1da94ce02cd463a0ae04596e1575800d7f3cba1a0e40R137). Now it does not have anymore as the function does a different thing and there is no `autoPlay` parameter anymore. I'm removing it.

app/src/main/java/org/schabi/newpipe/player/gesture/PopupPlayerGestureListener.kt
Stypox: This is the commit that introduced it: 38b2ffd4504cefaa79885d2f6b90a9c129645445. From the PR (#810) description:

> Added popup player tossing when flung with weaker strength (useful for getting the popup out of keyboard's way).

I personally don't understand [the code](https://github.com/TeamNewPipe/NewPipe/blob/26099bb47b821978a35ee961d7ffaf2c4a89a393/app/src/main/java/org/schabi/newpipe/player/gesture/PopupPlayerGestureListener.kt#L173-L190) (it sets a velocity to a position... it does not make sense from a metrics point of view).
Btw, "toss" and "fling" mean roughly the same thing, I don't know why there is such a redundancy.

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
Stypox: Moved to `PopupPlayerUi`

app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java
Stypox: The explanation is in the method javadoc. I created three constants, but I will not proceed further into abstracting (nothing good comes to mind) and/or improving the meaning of those values in this PR.

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.java
Stypox: There are just too many of these and it's just an Android Studio bug. In this case it says to use `playerUis.forEach(consumer);` instead of `playerUis.stream().forEach(consumer);`, but the former is not allowed on APIs < 24.

app/src/main/java/org/schabi/newpipe/player/ui/PopupPlayerUi.java
Stypox: I will not take care of this in this PR as it's just copied code: https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/Player.java#L1571

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
Stypox: Again, not in this PR

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
Stypox: Unrelated to this PR: https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/Player.java#L2584

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
Stypox: Then please open a checkstyle PR adding such a rule, this repo is full of things like this (many done by me). If there is nothing enforcing it, it is a waste of time to change it, just to forget next time ;-)

Stypox:
@litetex
- rebased on top of `dev` (it was not so difficult)
- fixed an issue with the volume slider (it was using the incorrect multiplier)
- applied your review comments (see my answers)

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
litetex: @Stypox 
Created #8601

litetex:
@Stypox Looks pretty good to me now.

I think a few things are still not resolved/commented (see my previous review above) but the rest looks good.
I will install the debug APK and test it for a few days on my smartphone to see if I find any additional problems.

opusforlife2:
Whoa, the player now reorients to portrait when it detects the video's aspect ratio with 'always fullscreen' on! Thanks, @Stypox!

It's much better than doing it manually, fumbling with the button, and having to rewind.

app/src/main/java/org/schabi/newpipe/player/PlayerService.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @Override
+    public void onTaskRemoved(final Intent rootIntent) {
+        super.onTaskRemoved(rootIntent);
+        if (!player.videoPlayerSelected()) {
+            return;
+        }
+        onDestroy();
+        // Unload from memory completely
+        Runtime.getRuntime().halt(0);
```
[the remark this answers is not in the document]
Stypox: It was apparently introduced by @avently in c45514b9896d83d858daefd5f6a58e2ae61c5bfc: https://github.com/TeamNewPipe/NewPipe/blame/dev/app/src/main/java/org/schabi/newpipe/player/MainPlayer.java#L170. According to its description, `onTaskRemoved` "is called if the service is currently running and the user has removed a task that comes from the service's application.". So destroying and completely killing the app here makes sense, in order to respond correctly to the user swiping the app from recents. Closing it not gracefully is apparently a good way to "unload from memory completely"?!?

app/src/main/java/org/schabi/newpipe/player/PlayerService.java
```diff
@@ -0,0 +1,159 @@
...
+ */
+public final class PlayerService extends Service {
+    private static final String TAG = PlayerService.class.getSimpleName();
+    private static final boolean DEBUG = Player.DEBUG;
+
+    private Player player;
+
+    private final IBinder mBinder = new PlayerService.LocalBinder();
+
+    public enum PlayerType {
```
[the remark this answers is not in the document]
Stypox: Done

Stypox:
> I think a few things are still not resolved/commented (see my previous review above) but the rest looks good.

Oh yes, sorry. I pushed a commit regarding the `PlayerType` and added a comment about `halt(0)`

> Whoa, the player now reorients to portrait when it detects the video's aspect ratio with 'always fullscreen' on!

~Oh yeah, that's definitely something I worked hard on, yes!~ Uh, great, it was fixed by accident then ;-)

litetex (approved):
Tested the APK a few days, everything is working fine :)

Code also LGTM

opusforlife2:
Got a crash. I can't tell you the context of the crash because... I fell asleep while using the app and woke up minutes later to Scoop with a crash report. 😅

```
FATAL EXCEPTION: main
Process: org.schabi.newpipe.debug.playerrefactor, PID: 31478
java.lang.NullPointerException: Attempt to read from field 'android.widget.FrameLayout org.schabi.newpipe.databinding.FragmentVideoDetailBinding.playerPlaceholder' on a null object reference
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.lambda$addVideoPlayerView$18$org-schabi-newpipe-fragments-detail-VideoDetailFragment(VideoDetailFragment.java:1316)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment$$ExternalSyntheticLambda18.accept(Unknown Source:4)
	at j$.util.Optional.ifPresent(Optional.java:159)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.lambda$addVideoPlayerView$19$org-schabi-newpipe-fragments-detail-VideoDetailFragment(VideoDetailFragment.java:1314)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment$$ExternalSyntheticLambda6.run(Unknown Source:2)
```

Stypox:
@opusforlife2 thank you for reporting; should be fixed in https://github.com/TeamNewPipe/NewPipe/pull/8170/commits/0e8cc72b134bf5755f59cbbfd6e01180509e7c57 :-)

litetex:
I also got 2 random crashes over the past week with the debug APK from 2022-07-10 however I was unable to see what happened because there was no ACRA report and I didn't have a PC with ADB near me. But I guess it was the same problem as opus pointed out above.

Merging for now to ensure no conflicts.

Subsequent problems should be addressed by new PRs.

PJansky:
I've found a minor problem with some videos with non standard aspect ratios like album covers. After opening the video in portrait mode, the video is stretched to use the full width of the screen. It's only reproducible with videos from  "Artist - Topic"  Youtube channels, videos from other channels with squarish aspect ratios are somehow not effected. Going into full screen playback and back to portrait resolves this issue.

The testing APK otherwise works flawlessly in my two days of use.

Screenshot of problem below:

Screenshot_20220720-170832_NewPipe player-refactor

opusforlife2:
Confirmed ^

Stypox:
@opusforlife2 @PJansky see #8731

opusforlife2:
From the OP:

> Maybe we could merge it with NotificationUtil? Tell me what you think

This might have been addressed in the review comments, but just in case it wasn't, I'm leaving this here for future reference.

Stypox:
No, it wasn't addressed ;-)

Issue #7870, opened 2022-02-13
Sorting playlists/subscriptions


IErnest134:
### Checklist

- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I'm aware that this is a request for NewPipe itself and that requests for adding a new service need to be made at [NewPipeExtractor](https://github.com/TeamNewPipe/NewPipeExtractor/issues).
- [X] I have taken the time to fill in all the required details. I understand that the feature request will be dismissed otherwise.
- [X] This issue contains only one feature request.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Feature description

Add two bars for playlists and subscriptions to move up/down

Example ▼
![example](https://user-images.githubusercontent.com/99507421/153772450-e6f4d547-d027-4c5c-b2c3-4c1541666ac2.jpg)

### Why do you want this feature?

I would like to sort playlists or subscriptions some down others up for convenience.
Thanks in advance.

GGAutomaton:
I would like to work on this issue.

GGAutomaton:
Current progress:
- Add a new column "display_index" in the database.
- Implement an algorithm to merge local playlists and remote playlists by index. If they have the same index, sort them by name.
- If the saved index doesn't match the actual display index for some reason (e.g., setting to all zero after database migration, or the index is not continuous after deletion), save the new index to the database.
- A newly added playlist is always on the top of the bookmark.

GGAutomaton:
Almost done.

https://user-images.githubusercontent.com/32899400/163572697-42c74fe6-dacf-4750-968e-4ce4f494db9c.mp4

nerdguy1338:
> Current progress:
> 
> * Add a new column `display_index` in the database.
> * Implement an algorithm to merge local playlists and remote playlists by index. If they have the same index, sort them by name.
> * If the saved index doesn't match the actual display index for some reason (e.g., setting to all zero after database migration, or the index is not continuous after deletion), save the new index to the database.
> * A newly added playlist is always on the top of the bookmark.
> * UI design. Add item handle.
> * Debounced saver.

[...]

PR #8221, opened 2022-04-15
Sort bookmarked playlists

### LINKED ISSUE #3316: Sort bookmarked playlists in main page
NOT related to #3263, because it's about inner playlist sorting. I'm talking about bookmarked playlists you can add or create in main page. I'll explain below.

#### Describe the solution you'd like
I've several playlists created or added in main page, and they are currently sorted by name (automatically). I'd like to sort them as I want, to move some playlists up, other down.
Open a playlist **that you have created**, you'll see at the right two —, and when you click on it, you can move up or down a video. That's exactly what I want (not mandatory this design), but in main page.

Convince us! How does it change your NewPipe experience and/or your life?
The better this paragraph is, the more likely a developer will think about working on it.
It is really cool when you have a lot of playlists, you want to sort them. For example, I'd like to put all the music playlists at the top, and the coding playlists at the bottom.

GGAutomaton:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Database migration
Add a new column `display_index (INTEGER NOT NULL DEFAULT 0)` in `playlists` and `remote_playlists`.
- A playlists merge algorithm
Implement an algorithm to merge local playlists and remote playlists by index. If they have the same index, sort them by name.
- Changing index
If the index saved in the database doesn't match the actual display index for some reason (e.g., setting to all zero after database migration, or a new playlist is created with index -1 (thus being on the top), or the index is not continuous after deletion, or the user changes the index), save the new index to the database.
- Debounced saver
    Save the list 10 seconds after the last change occurred. The logic is almost the same as the one in `LocalPlaylistFragment`.
    **Update:** now the code is reused.
- UI and Item holder
  - Create `list_playlist_bookmark_item.xml`. It's just `list_playlist_mini_item.xml` adding an handle on the right.
  - Create `LocalBookmarkPlaylistItemHolder` and `RemoteBookmarkPlaylistItemHolder`.
  - `LocalItemListAdapter` will choose the item holder with/without the handle.

Before:

Before

After:

https://user-images.githubusercontent.com/32899400/163595997-0f3c0214-f75b-47c8-865f-7d5fdab76790.mp4

- Fixes #3316
- Half of #7870

#### APK testing 

#### Due diligence


GGAutomaton:
https://user-images.githubusercontent.com/32899400/163661533-00f6446b-346d-4cce-abe1-0777357bf6b3.mp4


The data storage logic has been modified, so more detailed testing is required to avoid data loss.
I'm not familiar with rxjava, therefore some code needs extra checking.

GGAutomaton:
Two of the three code smells are about high complexity methods. But their logic is continuous, I don't think refactoring them is a necessity.
Another warning is about the debounced saver. It's similar to the one in `LocalPlaylistFragment`. But `BookmarkFragment` and `LocalPlaylistFragment` have each extended another abstract class, reusing the debounced saver could be difficult.

Please keep me informed if there's anything else to do.

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
```diff
@@ -247,12 +314,242 @@ protected void resetFragment() {
...
+    @Override
+    public void saveImmediate() {
+        if (itemListAdapter == null) {
+            return;
+        }
+
+        // List must be loaded and modified in order to save
+        if (isLoadingComplete == null || debounceSaver == null
+                || !isLoadingComplete.get() || !debounceSaver.getIsModified()) {
+            Log.w(TAG, "Attempting to save playlists in bookmark when bookmark "
```
Stypox, line 402: Since this is always called in `onPause`, I would not print a warning here

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
```diff
@@ -247,12 +314,242 @@ protected void resetFragment() {
...
+
+            // Updating display index in the item does not affect the value inserts into
+            // database, which will be recalculated during the database update. Updating
+            // display index in the item here is to determine whether it is recently modified.
+            // Save the index read from the database.
+            if (item instanceof PlaylistMetadataEntry) {
+
+                displayIndexInDatabase.put(new Pair<>(((PlaylistMetadataEntry) item).uid,
+                        LocalItem.LocalItemType.PLAYLIST_LOCAL_ITEM), item.getDisplayIndex());
+                ((PlaylistMetadataEntry) item).displayIndex = i;
```
Stypox, line 376: I would make `setDisplayIndex` and `getUid` abstract methods of `PlaylistLocalItem`

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
```diff
@@ -247,12 +314,242 @@ protected void resetFragment() {
...
+
+                final Long uid = ((PlaylistMetadataEntry) item).uid;
+                final Pair<Long, LocalItem.LocalItemType> key = new Pair<>(uid,
+                        LocalItem.LocalItemType.PLAYLIST_LOCAL_ITEM);
+                final Long databaseIndex = displayIndexInDatabase.remove(key);
+
+                // The database index should not be null because inserting new item into database
+                // is not handled here. NullPointerException has occurred once, but I can't
+                // reproduce it. Enhance robustness here.
+                if (databaseIndex != null && databaseIndex != i) {
```
Stypox, line 428: I don't understand why you need to store the original database index and all of the other data. I would do it like this:
- when initially receiving the items, just set them in the list
- keep a `List<Pair<Long, LocalItem.LocalItemType>> deletedItems` and when the user deletes an item, add its id+type here
- when doing any change (except deletion), do not update any variable except `debounceSaver.isModified`
- let the debounce saver or `onPause` call `saveImmediate`
- in `saveImmediate`:
  - delete all items found in `deletedItems` and clear `deletedItems`
  - cycle through `items` and if there is an item with a `displayIndex` different than its position in the list, add it to the items to update
  - update the `displayIndex` of items in the list

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
```diff
@@ -247,12 +314,242 @@ protected void resetFragment() {
...
+
+        displayIndexInDatabase.clear();
+
+        // 1. Update local playlists
+        // 2. Update remote playlists
+        // 3. Set isModified false
+        disposables.add(localPlaylistManager.updatePlaylists(localItemsUpdate, localItemsDeleteUid)
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(() -> disposables.add(remotePlaylistManager.updatePlaylists(
+                        remoteItemsUpdate, remoteItemsDeleteUid)
```
Stypox, line 462: Instead of nesting things, you can just use some RxJava functions to do two actions in parallel and then calling `debounceSaver.setIsModified(false);` when both have finished.

app/src/main/java/org/schabi/newpipe/local/holder/LocalBookmarkPlaylistItemHolder.java
```diff
@@ -0,0 +1,54 @@
...
+        final PlaylistMetadataEntry item = (PlaylistMetadataEntry) localItem;
+
+        itemHandleView.setOnTouchListener(getOnTouchListener(item));
+
+        super.updateFromItem(localItem, historyRecordManager, dateTimeFormatter);
+    }
+
+    private View.OnTouchListener getOnTouchListener(final PlaylistMetadataEntry item) {
+        return (view, motionEvent) -> {
+            view.performClick();
```
Stypox, line 45: Why perform click? In other places, too

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -109,12 +108,11 @@ public static LocalPlaylistFragment getInstance(final long playlistId, final Str
     public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         playlistManager = new LocalPlaylistManager(NewPipeDatabase.getInstance(requireContext()));
-        debouncedSaveSignal = PublishSubject.create();
 
         disposables = new CompositeDisposable();
 
         isLoadingComplete = new AtomicBoolean();
-        isModified = new AtomicBoolean();
+        debounceSaver = new DebounceSaver(10000, this);
```
Stypox, line 115: Maybe use the same constant from the other fragment

app/src/main/java/org/schabi/newpipe/util/DebounceSaver.java
```diff
@@ -0,0 +1,68 @@
+package org.schabi.newpipe.util;
```
Stypox, line 1: I would create a sub-package `util.debounce` for this and `DebounceSavable`

app/src/main/java/org/schabi/newpipe/util/DebounceSaver.java
```diff
@@ -0,0 +1,68 @@
...
+public class DebounceSaver {
+
+    private final long saveDebounceMillis;
+
+    private final PublishSubject<Long> debouncedSaveSignal;
+
+    private final DebounceSavable debounceSavable;
+
+    // Has the object been modified
+    private final AtomicBoolean isModified;
```
Stypox, line 22: I don't get why `isModified` is stored here, as it actually has nothing to do with the debounce saver. Either you move more of the code having to do with it here, or you move it out of this class.

Stypox (changes requested):
Thank you! I have a few comments

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
GGAutomaton: `item` is an object implementing `LocalItem`, not `PlaylistLocalItem`. Still need to write `((PlaylistMetadataEntry) item)` if making `getUid` abstract methods of `PlaylistLocalItem`.

app/src/main/java/org/schabi/newpipe/util/DebounceSaver.java
GGAutomaton: This is because `saveChanges` is frequently called by `BookmarkFragment` and `LocalPlaylistFragment`. If `isModified` is not saved in the debounce saver, I need to write 
```java
if (isModified != null) isModified.set(true);
debounceSaver.saveChanges();
```
everytime I want to save changes.

The original method is here. The debounce saver is integrated in `LocalPlaylistFragment` before this pr.
https://github.com/TeamNewPipe/NewPipe/blob/a4037a826822a7cbf56619dd9a782e3583d17287/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java#L617-L624

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
GGAutomaton: Nope. It will also be called 10 seconds after the last change, not only in `onPause`.

https://github.com/GGAutomaton/NewPipe/blob/6526ff1612ad430325a5bf8af7b425055306a0e6/app/src/main/java/org/schabi/newpipe/util/DebounceSaver.java#L51-L58

app/src/main/java/org/schabi/newpipe/local/holder/LocalBookmarkPlaylistItemHolder.java
GGAutomaton: Copied from
https://github.com/TeamNewPipe/NewPipe/blob/a4037a826822a7cbf56619dd9a782e3583d17287/app/src/main/java/org/schabi/newpipe/local/holder/LocalPlaylistStreamItemHolder.java#L127-L137

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
```diff
@@ -26,28 +31,47 @@
...
     private Subscription databaseSubscription;
     private CompositeDisposable disposables = new CompositeDisposable();
     private LocalPlaylistManager localPlaylistManager;
     private RemotePlaylistManager remotePlaylistManager;
+    private ItemTouchHelper itemTouchHelper;
+
+    /* Have the bookmarked playlists been fully loaded from db */
+    private AtomicBoolean isLoadingComplete;
+
+    private DebounceSaver debounceSaver;
```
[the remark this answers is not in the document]
GGAutomaton: The debounce saver is first introduced in `LocalPlaylistFragment`. Everytime the change occurred, all the videos in the playlist will be deleted and insert again. (This implementation is a bit of silly, I think.) Perhaps deleting 100 records and adding 100 records when the item moves a single row. This will lead to heavy db load.

As for `BookmarkFragment`, the db load is not as heavy as that in `LocalPlaylistFragment`, since it only modifies records it needs to modify. I use the debounce saver in `BookmarkFragment` for consistency. But not using it might be fine.

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
Stypox: Yes, I know. But I mean, this is, among other places, always called in `onPause`. So the warning will be generated each time this method is called from `onPause`, which is not good...

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
Stypox: No, `item` here is implementing `PlaylistLocalItem`: https://github.com/TeamNewPipe/NewPipe/blob/d32490a4be288348a8377c84dbf8c82253240bf6/app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java#L363

app/src/main/java/org/schabi/newpipe/local/holder/LocalBookmarkPlaylistItemHolder.java
Stypox: Mmmh, ok, I guess AS gives the `ClickableViewAccessibility` warning otherwise ([never understood what those warnings mean](https://stackoverflow.com/questions/47107105/android-button-has-setontouchlistener-called-on-it-but-does-not-override-perform)).

app/src/main/java/org/schabi/newpipe/util/DebounceSaver.java
Stypox: Taking a second look, I think you should change all calls to `setIsModified(false)` to `setNoChangesToSave()` or something like that, to make it clearer that calling such method indicates that we either finished saving, or we just loaded data so nothing needs saving. Feel free to suggest a better name ;-)

app/src/main/java/org/schabi/newpipe/database/playlist/PlaylistLocalItem.java
```diff
@@ -11,17 +11,68 @@
...
                 localPlaylists.size() + remotePlaylists.size());
-        items.addAll(localPlaylists);
-        items.addAll(remotePlaylists);
+        final List<PlaylistLocalItem> itemsWithSameIndex = new ArrayList<>();
+
+        // The data from database may not be in the display index order
+        Collections.sort(localPlaylists,
+                Comparator.comparingLong(PlaylistMetadataEntry::getDisplayIndex));
+        Collections.sort(remotePlaylists,
+                Comparator.comparingLong(PlaylistRemoteEntity::getDisplayIndex));
```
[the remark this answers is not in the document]
GGAutomaton: Unfortunately, I found that `merge` may be called in other situations which I cannot trace the call stack. In these cases, the playlists are out of order, so sorting them may still be a necessity.

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
GGAutomaton: Deleted.

app/src/test/java/org/schabi/newpipe/database/playlist/PlaylistLocalItemTest.java
```diff
@@ -0,0 +1,105 @@
...
+        assertEquals(0, mergedPlaylists.size());
+    }
+
+    @Test
+    public void onlyLocalPlaylists() {
+        final List<PlaylistMetadataEntry> localPlaylists = new ArrayList<>();
+        final List<PlaylistRemoteEntity> remotePlaylists = new ArrayList<>();
+        localPlaylists.add(new PlaylistMetadataEntry(1, "name1", "", 0, 1));
+        localPlaylists.add(new PlaylistMetadataEntry(2, "name2", "", 1, 1));
+        localPlaylists.add(new PlaylistMetadataEntry(3, "name3", "", 3, 1));
```
Stypox, line 29: Shouldn't the indexes be in a random order, so that the test can check whether `merge` sorts correctly? Also in other functions here

app/src/main/java/org/schabi/newpipe/database/playlist/PlaylistLocalItem.java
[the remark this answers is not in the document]
Stypox: In the end I removed those checks and made sure `merge` is called in only one place

app/src/test/java/org/schabi/newpipe/database/playlist/PlaylistLocalItemTest.java
Stypox: See https://github.com/TeamNewPipe/NewPipe/pull/8221#discussion_r1545370806

Stypox (approved):

Stypox:
The 5 sonartype issues are due to:
- some code duplication due to both BookmarkFragment and LocalPlaylistFragment allowing to move items around (but it would make things just more complicated to try to deduplicate this, it's not so much after all)
- an empty onComplete() method (there's no action to do, and there isn't a specific reason for this, so no need to add a comment to the method)
- some "commented out code" which is actually a legitimate comment explaining what methods should be used for future reference
- constructor with 8 parameters required by the Room database

PR #8222, opened 2022-04-15
Update NewPipeExtractor again to fix throttling on age restricted videos

Stypox:
The release apk contains this fix, but I accidentally forgot to push the local change before merging the release PR :sweat_smile:


opusforlife2:
Does this mean... for a short while there... the GPL was being violated? 👀

PR #8227, opened 2022-04-16
Fix keyboard showing after the search box acquiring focus

### LINKED ISSUE #7647: Keyboard doesn't open - Disabled animations
### Checklist

- [x] I am using the latest version - 0.21.15 
- [x] I checked, found duplicates (open OR closed) of this issue in the repo.  #7607 

0. ``Dev note: Animations have to be disabled - otherwise the bug is not reproducible``
1. Pull down a video to the bottom of the player as seen below.
2. Search bar won't call they keyboard even after multiple clicks.
3. Pull up the video to the normal view and pull it down again.
4. Search bar now will call the keyboard as intended.

### Actual behavior

Search fails to call the keyboard.

### Expected behavior

Search should call the keyboard.

IMG_000023

### Logs

NA

### Device info

 - Android version/Custom ROM version: 11
 - Device model: OnePlus

### Note
Tested https://github.com/TeamNewPipe/NewPipe/suites/4810100849/artifacts/135282702 found at #7607 but didn't work.

dtcxzyw:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- The code block on onResume() for showing/hiding the soft keyboard failed in a specific scenario. It is because the current view of InputMethodManager is not searchEditText. This problem is handled just when searchEditText gets the focus.
- (**Update**) Fix the underlying problem as described in #7647. It just tries again to request focus and show the soft keyboard when it fails.

- Before:

https://user-images.githubusercontent.com/15650457/163761718-ba6fc388-73be-4425-9af8-2fbc86bb07bd.mp4

- After:

https://user-images.githubusercontent.com/15650457/163761735-2a86f962-7e62-4182-843b-d6ded03292cc.mp4

- Fixes #7647

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
@@ -141,8 +141,7 @@ public class SearchFragment extends BaseListFragment<SearchInfo, ListExtractor.I
     @State
     boolean wasSearchFocused = false;
 
-    @Nullable
-    private Map<Integer, String> menuItemToFilterName = null;
+    @Nullable private Map<Integer, String> menuItemToFilterName = null;
     private StreamingService service;
     private Page nextPage;
     private boolean showLocalSuggestions = true;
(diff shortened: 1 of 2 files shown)


litetex (changes requested):
Thank you for the PR.

However:
* This PR does not fix the underlying problem https://github.com/TeamNewPipe/NewPipe/issues/7647#issuecomment-1100943166 and only applies a questionable workaround.
  * As I can't verify that this workaround works correctly and this is only fixes the SearchFragment and not other places where the keyboard is shown, I would suggest you implement the fix from the above comment instead.
* If possible don't upload demo videos with background music

dtcxzyw:
Thank @litetex for the review! I tried a similar fix on KeyboardUtil before this PR and it didn't work. But after I called ```cleanFocus()``` before re-requesting focus as you suggested, it worked fine. It does solve the problem without confusion compared to previous solutions. Please let me know if it's ready to be merged into the dev branch.

litetex (approved):
LGTM :)

litetex:
To the merger: This should be squash merged.

PR #8242, opened 2022-04-17
Trim search string and remove duplicate records from the database

### LINKED ISSUE #8012: Trim trailing/preceding whitespaces from search inputs
### Checklist

### Feature description

I think it would be better if trailing and/or preceding whitespaces are trimmed from search inputs. 

By not having those whitespaces trimmed, it's easy to get duplicates in the search history. YouTube app does it this way too. 

### Additional information

Not sure if this could be considered a bug in some way, so pardon me for labeling it as a feature request,  if that's the case.

dtcxzyw:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Strip leading and trailing white spaces of search string before passing it into ```suggestionPublisher``` ~~after typing or pasting text~~
- Trim the search string of the search bar when the user presses search
- **Add new database migration to remove duplicate records in search history**
- Add a test for the database migration

- Before:

https://user-images.githubusercontent.com/15650457/163706467-497441a9-5213-456a-8255-64575c51acb8.mp4

- After:

https://user-images.githubusercontent.com/15650457/163706473-f08b8102-ea62-4b96-ba08-82e47a1bf645.mp4

- Fixes #8012

#### APK testing 

[Debug APK](https://github.com/TeamNewPipe/NewPipe/suites/6149649612/artifacts/214883611)

#### Due diligence

### DIFF
--- a/app/src/androidTest/java/org/schabi/newpipe/database/DatabaseMigrationTest.kt
+++ b/app/src/androidTest/java/org/schabi/newpipe/database/DatabaseMigrationTest.kt
@@ -111,8 +111,10 @@ class DatabaseMigrationTest {
         )
 
         testHelper.runMigrationsAndValidate(
-            AppDatabase.DATABASE_NAME, Migrations.DB_VER_6,
-            true, Migrations.MIGRATION_5_6
+            AppDatabase.DATABASE_NAME,
+            Migrations.DB_VER_8,
+            true,
+            Migrations.MIGRATION_7_8
         )
 
         val migratedDatabaseV3 = getMigratedDatabase()
@@ -150,10 +152,13 @@ class DatabaseMigrationTest {
     }
 
     @Test
-    fun migrateDatabaseFrom5to6() {
-        val databaseInV5 = testHelper.createDatabase(AppDatabase.DATABASE_NAME, Migrations.DB_VER_5)
+    fun migrateDatabaseFrom7to8() {
+        val databaseInV7 = testHelper.createDatabase(AppDatabase.DATABASE_NAME, Migrations.DB_VER_7)
 
-        databaseInV5.run {
+        val defaultSearch1 = " abc "
+        val defaultSearch2 = " abc"
+
+        databaseInV7.run {
             insert(
                 "search_history", SQLiteDatabase.CONFLICT_FAIL,
                 ContentValues().apply {
@@ -186,12 +191,12 @@ class DatabaseMigrationTest {
         }
 
         testHelper.runMigrationsAndValidate(
-            AppDatabase.DATABASE_NAME, Migrations.DB_VER_6,
-            true, Migrations.MIGRATION_5_6
+            AppDatabase.DATABASE_NAME, Migrations.DB_VER_8,
+            true, Migrations.MIGRATION_7_8
         )
 
-        val migratedDatabaseV6 = getMigratedDatabase()
-        val listFromDB = migratedDatabaseV6.searchHistoryDAO().all.blockingFirst()
+        val migratedDatabaseV8 = getMigratedDatabase()
+        val listFromDB = migratedDatabaseV8.searchHistoryDAO().all.blockingFirst()
 
         assertEquals(2, listFromDB.size)
         assertEquals("abc", listFromDB[0].search)
--- a/app/src/androidTest/java/org/schabi/newpipe/database/DatabaseMigrationTest.kt
+++ b/app/src/androidTest/java/org/schabi/newpipe/database/DatabaseMigrationTest.kt
@@ -28,9 +28,6 @@ class DatabaseMigrationTest {
 
         private const val DEFAULT_SECOND_SERVICE_ID = 1
         private const val DEFAULT_SECOND_URL = "https://www.youtube.com/watch?v=ncQU6iBn5Fc"
-
-        private const val DEFAULT_SEARCH1 = " abc "
-        private const val DEFAULT_SEARCH2 = " abc"
     }
 
     @get:Rule
@@ -163,28 +160,28 @@ class DatabaseMigrationTest {
                 "search_history", SQLiteDatabase.CONFLICT_FAIL,
                 ContentValues().apply {
                     put("service_id", DEFAULT_SERVICE_ID)
-                    put("search", DEFAULT_SEARCH1)
+                    put("search", defaultSearch1)
                 }
             )
             insert(
                 "search_history", SQLiteDatabase.CONFLICT_FAIL,
                 ContentValues().apply {
                     put("service_id", DEFAULT_SERVICE_ID)
-                    put("search", DEFAULT_SEARCH2)
+                    put("search", defaultSearch2)
                 }
             )
             insert(
                 "search_history", SQLiteDatabase.CONFLICT_FAIL,
                 ContentValues().apply {
                     put("service_id", DEFAULT_SECOND_SERVICE_ID)
-                    put("search", DEFAULT_SEARCH1)
+                    put("search", defaultSearch1)
                 }
             )
             insert(
                 "search_history", SQLiteDatabase.CONFLICT_FAIL,
                 ContentValues().apply {
                     put("service_id", DEFAULT_SECOND_SERVICE_ID)
-                    put("search", DEFAULT_SEARCH2)
+                    put("search", defaultSearch2)
                 }
             )
             close()
--- a/app/src/androidTest/java/org/schabi/newpipe/database/DatabaseMigrationTest.kt
+++ b/app/src/androidTest/java/org/schabi/newpipe/database/DatabaseMigrationTest.kt
@@ -8,6 +8,7 @@ import androidx.test.core.app.ApplicationProvider
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
 import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
 import org.junit.Assert.assertNull
 import org.junit.Rule
 import org.junit.Test
@@ -25,8 +26,11 @@ class DatabaseMigrationTest {
         private const val DEFAULT_UPLOADER_NAME = "Uploader Test"
         private const val DEFAULT_THUMBNAIL = "https://example.com/example.jpg"
 
-        private const val DEFAULT_SECOND_SERVICE_ID = 0
+        private const val DEFAULT_SECOND_SERVICE_ID = 1
         private const val DEFAULT_SECOND_URL = "https://www.youtube.com/watch?v=ncQU6iBn5Fc"
+
+        private const val DEFAULT_SEARCH1 = " abc "
+        private const val DEFAULT_SEARCH2 = " abc"
     }
 
     @get:Rule
@@ -106,6 +110,11 @@ class DatabaseMigrationTest {
             Migrations.MIGRATION_6_7
         )
 
+        testHelper.runMigrationsAndValidate(
+            AppDatabase.DATABASE_NAME, Migrations.DB_VER_6,
+            true, Migrations.MIGRATION_5_6
+        )
+
         val migratedDatabaseV3 = getMigratedDatabase()
         val listFromDB = migratedDatabaseV3.streamDAO().all.blockingFirst()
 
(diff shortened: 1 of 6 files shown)


tsiflimagas:
Migration worked great, and trimming also works as expected. Thank you!

litetex:
@TiA4f8R 
Do we really need a database migration for this?
The users can simply delete their search history and the problems should be gone.
Looks a little bit overkill to me, also regarding the fact that it doesn't affect that many users.

Without the database migration, only a single class is affected, now we have 6...

TiA4f8R:
> @TiA4f8R 
> Do we really need a database migration for this?

Disclaimer: the idea is not from me but from @TobiGr:

> the dev might want to add a migration finding duplicate entries in the search history and delete them, leaving just the trimmed one

TobiGr:
@litetex I personally do not like to leave it up to the user to clean up the database to contain expected values.
It's better to know exactly how the data is structured. It's just the search history, but we should keep the db clean. I've learned that the hard way... (not in this project though)

dtcxzyw:
I don't think it's a good idea to have the user manually clear the search history for this feature. Actually, NewPipe works fine with or without database migrations because the PR does not add a unique constraint to the ```search``` fields.

app/src/main/java/org/schabi/newpipe/database/Migrations.java
```diff
@@ -188,6 +189,15 @@ public void migrate(@NonNull final SupportSQLiteDatabase database) {
         }
     };
 
+    public static final Migration MIGRATION_5_6 = new Migration(DB_VER_5, DB_VER_6) {
+        @Override
+        public void migrate(@NonNull final SupportSQLiteDatabase database) {
+            database.execSQL("DELETE FROM search_history WHERE id NOT IN (SELECT id FROM "
+                    + "(SELECT id FROM search_history GROUP BY trim(search), service_id) tmp)");
```
Stypox, line 196: I don't know why SQL lets you do this, but you can't select one `id` from the `GROUP BY`, since there could be multiple `id`s in the same group, as you are not grouping by id. I would suggest to do `SELECT MIN(id)` for a more well-defined behavior, even though there would be no actual difference.

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -395,8 +397,8 @@ public void onSaveInstanceState(@NonNull final Bundle bundle) {
 
     @Override
     public void reloadContent() {
-        if (!TextUtils.isEmpty(searchString)
-                || (searchEditText != null && !TextUtils.isEmpty(searchEditText.getText()))) {
+        if (!TextUtils.isEmpty(searchString) || (searchEditText != null
+                && TextUtils.getTrimmedLength(searchEditText.getText()) > 0)) {
```
Stypox, line 401: No need for a new function, use `!isBlank(searchEditText.getText())`, from Android's libraries (also in other places)

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -604,8 +610,13 @@ public void afterTextChanged(final Editable s) {
                     s.removeSpan(span);
                 }
 
-                final String newText = searchEditText.getText().toString();
+                final String newText = searchEditText.getText().toString().trim();
                 suggestionPublisher.onNext(newText);
+
+                if (isPastedText) {
+                    // trim pasted text
+                    searchEditText.setText(newText);
```
Stypox, line 618: Why trim it in the search field, too? Let the user type or paste what he wants and:
- trim the string before passing it to `suggestionPublisher`
- when the user presses search, set the trimmed string in the search bar, search for the trimmed string and save the trimmed string in the db

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -811,6 +823,11 @@ private void search(final String theSearchString,
                         final String theSortFilter) {
         if (DEBUG) {
             Log.d(TAG, "search() called with: query = [" + theSearchString + "]");
+            final String trimmedSearchString = theSearchString.trim();
+            if (!trimmedSearchString.equals(theSearchString)) {
+                Log.d(TAG, "The precondition is not satisfied. "
```
Stypox, line 828: I'd do `Log.w`

Stypox (changes requested):
I agree the database migration should be done, to keep data consistent

app/src/main/java/org/schabi/newpipe/database/Migrations.java
dtcxzyw: Oops! It is my mistake due to oversight.

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
dtcxzyw: The above two behaviors have been implemented.

As trimming when pasting, it is inconsistent with the behavior of other search bars. I thought it's often unintentional for users to paste text with leading and trailing spaces. So I made this additional change.

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
dtcxzyw: Should we handle other non-ASCII spaces? Like ```'\u3000'```. 
```String.trim()``` and ```org.schabi.newpipe.extractor.utils.Utils.isBlank()``` do not handle these code points. 
We should use ```String.strip()``` and ```String.isBlank()```, but they do not available in the current version of the Android SDK.

File: app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java [hunk 5]
```diff
@@ -395,8 +398,8 @@ public void onSaveInstanceState(@NonNull final Bundle bundle) {
 
     @Override
     public void reloadContent() {
-        if (!TextUtils.isEmpty(searchString)
-                || (searchEditText != null && !TextUtils.isEmpty(searchEditText.getText()))) {
+        if (!TextUtils.isEmpty(searchString) || (searchEditText != null
+                && !isBlank(searchEditText.getText().toString()))) {
```
litetex, line 402: I think a helper method for ``isBlank(searchEditText.getText().toString())`` would be good because it's used 3 times here.

Maybe something like
```java
public static boolean isTextViewBlank(final TextView textView) {
  return Utils.isBlank(textView.getText().toString());
}
```

litetex (dismissed):

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
Stypox: There is no need to, in my opinion

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
Stypox: Or even non static and without the parameter

app/src/androidTest/java/org/schabi/newpipe/database/DatabaseMigrationTest.kt
```diff
@@ -26,8 +27,11 @@ class DatabaseMigrationTest {
         private const val DEFAULT_UPLOADER_NAME = "Uploader Test"
         private const val DEFAULT_THUMBNAIL = "https://example.com/example.jpg"
 
-        private const val DEFAULT_SECOND_SERVICE_ID = 0
+        private const val DEFAULT_SECOND_SERVICE_ID = 1
```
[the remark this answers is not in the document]
dtcxzyw: I think ```service_id``` is used to distinguish different services. Search records with the same search string but different ```service_id``` should not be considered equal. However, ```service_id``` is always 0 currently.

app/src/androidTest/java/org/schabi/newpipe/database/DatabaseMigrationTest.kt
```diff
@@ -84,6 +88,11 @@ class DatabaseMigrationTest {
             true, Migrations.MIGRATION_4_5
         )
 
+        testHelper.runMigrationsAndValidate(
+            AppDatabase.DATABASE_NAME, Migrations.DB_VER_6,
+            true, Migrations.MIGRATION_5_6
+        )
+
```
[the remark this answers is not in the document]
dtcxzyw: This method also confuses me. Should I split ```migrateDatabaseFrom2to3``` into multiple functions?

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -811,6 +816,11 @@ private void search(final String theSearchString,
                         final String theSortFilter) {
         if (DEBUG) {
             Log.d(TAG, "search() called with: query = [" + theSearchString + "]");
+            final String trimmedSearchString = theSearchString.trim();
+            if (!trimmedSearchString.equals(theSearchString)) {
+                Log.w(TAG, "The precondition is not satisfied. "
+                        + "\"theSearchString\" is not allowed to have leading or trailing spaces");
+            }
```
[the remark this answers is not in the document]
dtcxzyw: The current implementation ensures that all strings passed into ```search``` satisfy the precondition. This check might help solve the problem if someone else calls ```search``` with a non-trimmed search string in the future.

app/src/main/java/org/schabi/newpipe/database/Migrations.java
```diff
@@ -188,6 +189,15 @@ public void migrate(@NonNull final SupportSQLiteDatabase database) {
         }
     };
 
+    public static final Migration MIGRATION_5_6 = new Migration(DB_VER_5, DB_VER_6) {
+        @Override
+        public void migrate(@NonNull final SupportSQLiteDatabase database) {
+            database.execSQL("DELETE FROM search_history WHERE id NOT IN (SELECT id FROM (SELECT "
+                    + "MIN(id) as id FROM search_history GROUP BY trim(search), service_id ) tmp)");
```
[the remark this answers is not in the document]
dtcxzyw: It removes duplicates before updating the string. The order doesn't matter.

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
```diff
@@ -383,6 +385,7 @@ public void readFrom(@NonNull final Queue<Object> savedObjects) throws Exception
 
     @Override
     public void onSaveInstanceState(@NonNull final Bundle bundle) {
+        searchEditText.setText(searchEditText.getText().toString().trim());
```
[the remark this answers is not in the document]
dtcxzyw: This is a small mistake.

app/src/androidTest/java/org/schabi/newpipe/database/DatabaseMigrationTest.kt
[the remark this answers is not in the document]
Stypox: Mmmh, this looks strange to me. In the UI all searches are shown, independently of the service you are currently on. So there should be no information about the service saved in search history, and there also shouldn't be duplicate searches with different service.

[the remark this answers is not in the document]
TobiGr: Fixed in 15fd47c7f24c79b0be2e0e98f216c3140e88c8fb

app/src/main/java/org/schabi/newpipe/fragments/list/search/SearchFragment.java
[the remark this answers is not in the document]
TobiGr: I forgot to remove this, will do later.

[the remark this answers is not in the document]
TobiGr: Removed the check and added a JDoc instead.

TobiGr (approved):
Thank you for the initial work!

Stypox (approved):

PR #8248, opened 2022-04-17
Fix inconsistency between user interaction and database commit order when re-adding videos to the playlist

### LINKED ISSUE #7402: Cannot readd video to playlist
### Checklist

1. Open a Playlist
2. Reproduce a video inside it in foreground
3. While still reproducing, remove it from the playlist, even without waiting
4. Without exiting the playlist, try readding it to the same playlist.
5. It should not work, until you close and reopen the playlist and readd it again.

### Actual behavior

The video cannot be readded to the same playlist it was just removed from without closing and reopening the playlist first.

### Expected behavior

I'd expect to readd it without having to close and reopen the playlist.

### Device info

 - Android version/Custom ROM version: Lineage OS 18.1 for microG
 - Device model: Oneplus 5

dtcxzyw:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

The reason for this issue is an inconsistency between user interaction and database commit order.

###### Expected behavior
delete - commit (delete) - add - commit (add)
###### Actual behavior
delete - add - commit (add, commit immediately) - commit (delete, commit after ~10s due to the debouncedSaveSignal)
###### Best Solution
Removes all debounced commit strategies. Sequential consistency is guaranteed by database transactions.

But I don't know why the delete operation uses the deferred commit strategy. BTW, I realize PR #8221 could have many conflicts with this solution. So I took a conservative approach to fix it.
###### Current Solution
Concurrent operations on the playlist database occur if and only if the user opens both LocalPlaylistFragment and VideoDetailFragment as described in the related issue. So I force all delete operations to be committed to the database before adding.

- Before:

https://user-images.githubusercontent.com/15650457/163764963-2ef2b4ee-7e0f-426d-9967-7e1d1381e943.mp4

- After:

https://user-images.githubusercontent.com/15650457/163764983-3438a231-a65e-4b3e-965a-8aedad61fafe.mp4

From main page tab:

https://user-images.githubusercontent.com/15650457/166681029-af08dfa5-6c94-4afa-b72b-0679cbef122e.mp4

- Fixes #7402

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -85,6 +85,7 @@ import org.schabi.newpipe.extractor.stream.VideoStream;
 import org.schabi.newpipe.fragments.BackPressable;
 import org.schabi.newpipe.fragments.BaseStateFragment;
 import org.schabi.newpipe.fragments.EmptyFragment;
+import org.schabi.newpipe.fragments.MainFragment;
 import org.schabi.newpipe.fragments.list.comments.CommentsFragment;
 import org.schabi.newpipe.fragments.list.videos.RelatedItemsFragment;
 import org.schabi.newpipe.ktx.AnimationType;
@@ -482,6 +483,8 @@ public final class VideoDetailFragment
                 // commit previous pending changes to database
                 if (fragment instanceof LocalPlaylistFragment) {
                     ((LocalPlaylistFragment) fragment).commitChanges();
+                } else if (fragment instanceof MainFragment) {
+                    ((MainFragment) fragment).commitPlaylistTabs();
                 }
 
                 disposables.add(PlaylistDialog.createCorrespondingDialog(requireContext(),
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -54,6 +54,7 @@ import androidx.appcompat.content.res.AppCompatResources;
 import androidx.appcompat.widget.Toolbar;
 import androidx.coordinatorlayout.widget.CoordinatorLayout;
 import androidx.core.content.ContextCompat;
+import androidx.fragment.app.Fragment;
 import androidx.preference.PreferenceManager;
 
 import com.google.android.exoplayer2.PlaybackException;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -445,6 +447,14 @@ public void onClick(final View v) {
                 break;
             case R.id.detail_controls_playlist_append:
                 if (getFM() != null && currentInfo != null) {
+                    final Fragment fragment = getParentFragmentManager().
+                            findFragmentById(R.id.fragment_holder);
+
+                    // commit previous pending changes to database
+                    if (fragment instanceof LocalPlaylistFragment) {
+                        ((LocalPlaylistFragment) fragment).commitChanges();
+                    }
```
Stypox, line 456: What about the cases when the local playlist fragment is a main page tab?

Stypox (changes requested):
Thank you!

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
dtcxzyw: This case can also be handled by these lines.
I tested this case on my device. It works fine.

Stypox: Mmmh, I don't understand why those lines would handle it. If the fragment is a main page tab, then it is not the parent of the VideoDetailFragment anymore

dtcxzyw: I have added a supplementary video for this case. Did I understand you correctly?

TobiGr: Fixed in 2de828289f91ef28ad0960fc784f102fc52bde5b

TobiGr:
- Rebased
- Fixed the behaviour when LocalPlaylistFragment is used as frontpage tab

| Before | After |
|---|---|
| [MainTab_LocalPlaylist_Before.webm](https://github.com/TeamNewPipe/NewPipe/assets/17365767/20c7b98f-47b4-4bcb-a173-fafb48d1b920) | [MainTab_LocalPlaylist_After.webm](https://github.com/TeamNewPipe/NewPipe/assets/17365767/6983d147-b886-4aac-844e-31e062768183) |

TobiGr (approved):
Sorry for the delay and thank you for submitting the fix!

Stypox (approved):

Issue #8230, opened 2022-04-16
Release 0.23.0 (please TEST!)


Stypox:
## Testing for regressions

Debug APK (built by our CI in #8231): [NewPipe_0.23.0_RC2.zip](https://github.com/TeamNewPipe/NewPipe/files/8511654/NewPipe_0.23.0_RC2.zip)
Release APK (signed with Stypox's keys): [NewPipe_0.23.0_RC2.zip](https://github.com/TeamNewPipe/NewPipe/files/8513027/NewPipe_0.23.0_RC2.zip)
Please report **only regressions** (i.e. new issues) here, not issues that were already present in the previous release!

## TODO

- [x] Merge Weblate
- [ ] Unlock Weblate when 0.23.0 is released
- [x] Possibly including @TiA4f8R's extractor improvement to solve potential future throttling issues
  - [x] In this case a new NewPipeExtractor version needs to be released,
  - [x] and the extractor commit should be updated in the gradle dependencies.
- [x] Possibly see if #7630 can be solved in this release
  - [x] #8302
- [x] #8233
- [x] #8236
- [x] #8244
- [x] #8249
- [x] DEFINITELY #8279

## NewPipeExtractor version

This version of NewPipe will ship with [NewPipeExtractor version 0.22.1](https://github.com/TeamNewPipe/NewPipeExtractor/releases/tag/v0.22.1)

### New
- Add seamless transition between background and video players when putting the app in background (for video-only streams and audio-only streams only) #7349 #8139
- Change pitch by semitones #7565
- Implement appending queue to playlist in main player #8008
- New streams notifications #2335
- Load YouTube mixes in related items TeamNewPipe/NewPipeExtractor#788 -- note that this has already been broken by YouTube

### Improved
- Remember adjustment step size for playback controls (speed and pitch) #7728
- Mitigating long buffering on initial video playback #7919
- Replace `R.string.yes` with `R.string.ok` #7976
- Improve player UI and navigability for Android TV #7963
- Add a confirmation button when deleting all files in downloader #5878
- ExoPlayer 2.17.1 update and MediaSource management rework #8020 #8150

### Fixed
- Fix player controls not hiding if resumed from media button #7341
- Migrate app update checker to AndroidX Work #7975
- Fix error notification on KitKat #7977
- Fix playback being reset to beginning when player type was changed #8020
- Fix download dialog selector layout #7516
- Fix crash when rotating device with playlist dialog open #8146
[...]

PR #8249, opened 2022-04-17
Fix disabled caption to no longer automatically re-enable on new player instance

karyogamy:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This fixes another bug introduced by #8150 where `DefaultTrackSelector` will have text tracks automatically selected at start, where captions will be enabled for a playback session (i.e. new player instance) even if `No Caption` was selected in previous session. 
This change will disable text tracks when building the caption menu when `userPreferredLanguage` does not exist.

- https://github.com/TeamNewPipe/NewPipe/issues/8230#issuecomment-1100921774

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -3557,14 +3557,19 @@ public final class Player implements
         }
 
         // apply caption language from previous user preference
-        final List<String> selectedPreferredLanguages =
-                trackSelector.getParameters().preferredTextLanguages;
         final String userPreferredLanguage =
                 prefs.getString(context.getString(R.string.caption_user_set_key), null);
         final int textRendererIndex = getCaptionRendererIndex();
 
-        if (userPreferredLanguage != null
-                && !selectedPreferredLanguages.contains(userPreferredLanguage)
+        if (userPreferredLanguage == null) {
+            trackSelector.setParameters(trackSelector.buildUponParameters()
+                    .setRendererDisabled(textRendererIndex, true));
+            return;
+        }
+
+        final List<String> selectedPreferredLanguages =
+                trackSelector.getParameters().preferredTextLanguages;
+        if (!selectedPreferredLanguages.contains(userPreferredLanguage)
                 && textRendererIndex != RENDERER_UNAVAILABLE) {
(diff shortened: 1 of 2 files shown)


litetex:
Note: Changed to targeted branch to the current release-branch.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -3557,14 +3557,19 @@ private void buildCaptionMenu(@NonNull final List<String> availableLanguages) {
...
         // apply caption language from previous user preference
-        final List<String> selectedPreferredLanguages =
-                trackSelector.getParameters().preferredTextLanguages;
         final String userPreferredLanguage =
                 prefs.getString(context.getString(R.string.caption_user_set_key), null);
         final int textRendererIndex = getCaptionRendererIndex();
 
-        if (userPreferredLanguage != null
-                && !selectedPreferredLanguages.contains(userPreferredLanguage)
+        if (userPreferredLanguage == null) {
```
litetex, line 3564: 1. Please add some comments why this stuff is happening
2. What is happening if ``textRendererIndex == RENDERER_UNAVAILABLE``?

karyogamy: 1. Fixed.
2. Fixed. Though nothing would happen in practice since here `RENDERER_UNAVAILABLE` is being disabled. But it is bad form and inconsistent with the other checks.

litetex:
I adjusted the CI so that it also runs when the PR target is a release branch.

@karyogamy 
You may cleanup your repo, there are a lot of dead branches ;)

litetex (approved):
LGTM

Tested it by switching between a few videos:
* "No captions" stays "No captions"
* "english" stays the same
* "english (auto-generated)" stays the same
* If nothing matching is available it falls back to "No captions"
* Closing and reopening the player also works

Stypox:
> I adjusted the CI so that it also runs when the PR target is a release branch.

Then we have to remember to always use `release/0.*` and not any other form like `release_0.*` like it has sometimes been done in the past ;-)

TacoTheDank:
Did you forget to add this PR to the draft release description?

PR #8262, opened 2022-04-20
Fix enqueuing behavior

### LINKED ISSUE #2611: Queueing logic when queue is empty (bug?)

Long pressing on background/popup when the queue is empty adds the audio/video to the queue, but it also starts playing them, since they are the first entry. This is the same behavior as would arise when tapping on background/popup.

I propose that long presses should add to queue but keep it paused until the user plays it. Maybe it should start buffering meanwhile?

### LINKED ISSUE #8151: Show Enqueue menu item even when no current video playing
### Checklist

### Feature description

Always show the "Enqueue" menu item even when there is no current video playing, like how YouTube allows you to hover a thumbnail and "add to queue" at any time.

I like to queue up a bunch of videos to watch all in a row.
But currently I have to start playing the first video and minimise it, and then I can enqueue other videos.

### Additional information

This is just for convenience to save a few seconds.

dtcxzyw:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Show ```Enqueue``` menu item even when the play queue is empty.
- First enqueued video/audio will remain paused by setting ```playWhenReady``` to false.
- Buffering the paused video/audio (Tested on my device, but I can't tell from the visual feedback that it's working).

- Before:

https://user-images.githubusercontent.com/15650457/164134776-c806b891-b5ac-4070-b18a-29f9a8e688fc.mp4

- After:

https://user-images.githubusercontent.com/15650457/164134787-ebc99119-f109-492d-9f10-f217f0685b2d.mp4

- Fixes #2611
- Fixes #8151

#### APK testing 

[Debug APK](https://github.com/TeamNewPipe/NewPipe/suites/6183214268/artifacts/217353359)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/dialog/InfoItemDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/dialog/InfoItemDialog.java
@@ -245,18 +245,16 @@ public final class InfoItemDialog {
         }
 
         /**
-         * Adds {@link StreamDialogDefaultEntry#ENQUEUE} if the player is open and
+         * Adds {@link StreamDialogDefaultEntry#ENQUEUE} unconditionally and
          * {@link StreamDialogDefaultEntry#ENQUEUE_NEXT} if there are multiple streams
          * in the play queue.
          * @return the current {@link Builder} instance
          */
         public Builder addEnqueueEntriesIfNeeded() {
-            if (PlayerHolder.getInstance().isPlayQueueReady()) {
-                addEntry(StreamDialogDefaultEntry.ENQUEUE);
+            addEntry(StreamDialogDefaultEntry.ENQUEUE);
 
-                if (PlayerHolder.getInstance().getQueueSize() > 1) {
-                    addEntry(StreamDialogDefaultEntry.ENQUEUE_NEXT);
-                }
+            if (PlayerHolder.getInstance().getQueueSize() > 1) {
+                addEntry(StreamDialogDefaultEntry.ENQUEUE_NEXT);
             }
             return this;
         }
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -898,6 +898,11 @@ public final class Player implements
 
         simpleExoPlayer.setVolume(isMuted ? 0 : 1);
         notifyQueueUpdateToListeners();
+
+        // Buffer paused streams when users add streams to the play queue.
+        if (!playOnReady && !playQueue.isEmpty()) {
+            simpleExoPlayer.prepare();
+        }
     }
     //endregion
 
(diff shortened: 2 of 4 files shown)


opusforlife2:
> Enqueue means "Add to the queue", so it doesn't make sense as an action if there is no queue.

@Stypox This is how music apps behave. Try Vanilla Music, for example.

> Why start paused? This would look like a random action happening for the user who isn't aware, and anyway we don't want new random behaviors without a natural meaning.

See the linked #2611. I would argue _this_ is the natural meaning.

> Also, there could be issues when the player is still starting and receives another enqueue intent: it might not be happy and react with a crash.

Can that not be guarded against programmatically? And is it not already possible to cause this in the current release? Enqueuing something while the player is still starting?

> Which player should be opened? Background, popup or main? There is no single answer.

I think the enqueuing should be done in the background player. If you're not actively opening the popup or main player using the relevant buttons, and are only building a queue from scratch, you likely don't want to switch to video watching just yet.

Stypox:
@opusforlife2 then instead of overloading the Enqueue button, I would create a new button named "Start paused in the background" that is only shown when there is no player active/starting.

opusforlife2:
@Stypox It will have the same purpose, but sure. Will it then be a swap between Enqueue and Start paused based on whether a player instance is already open?

Stypox:
> Will it then be a swap between Enqueue and Start paused based on whether a player instance is already open?

Yes, always either one or the other (or none, while the player is starting), in the same place

dtcxzyw:
Creating a new button might be more user-friendly.
I try to handle the case that the player is still starting when enqueuing.
Is there anything else that needs to be modified? @Stypox

Stypox:
No, I think adding a new button "Start paused in the background" to the long-press menu would be enough. Make sure to only show that button if **the player is not open** (if the player is open but the **play queue is null**, *nothing* should be shown; if the player is open and the play queue is not null, then the "Enqueue" should be shown)

opusforlife2:
> I try to handle the case that the player is still starting when enqueuing.

Are you saying this _cannot_ be done, @Stypox? Or that it should not be done in _this_ PR?

Stypox:
Both ;-)

I mean, everything is doable, but enqueueing things while the player is starting sounds complicated, thus I'd suggest to do it after the player and the ways it can be started hage been refactored

Stypox:
Both ;-)

I mean, everything is doable, but enqueueing things while the player is starting sounds complicated, thus I'd suggest to do it after the player and the ways it can be started have been refactored

opusforlife2:
@dtcxzyw Then if you're interested in handling the player-still-starting case, keep an eye on #8170 and similar future PRs. Is that okay, @Stypox?

opusforlife2:
This PR wasn't being blocked by #8170 except for a particular case. Are you interested in going further with this, @dtcxzyw?

Stypox:
Closing since there has not been an answer for a while, feel free to reopen ;-)

PR #8263, opened 2022-04-20
Added higher resolutions to the default res settings when they are enabled

### LINKED ISSUE #2941: Ability to select default 2K/4K resolution

Coder246:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Added higher resolutions the the default resolution settings.
They are only visible, when show higher resolutions are turned on.

When the user has selected a higher resolution as default and he turns "show higher resolutions" off, the highest now available resolution will be used as a fallback.

- Before:
Screenshot_20220420-092024_NewPipe Debug
- After:
Screenshot_20220420-092033_NewPipe Debug

- Fixes   #2941 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
@@ -15,17 +15,30 @@ import com.google.android.material.snackbar.Snackbar;
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.util.PermissionHelper;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 
 public class VideoAudioSettingsFragment extends BasePreferenceFragment {
     private SharedPreferences.OnSharedPreferenceChangeListener listener;
+    private ListPreference defaultRes;
+    private ListPreference defaultPopupRes;
+    private ListPreference limitMobDataUsage;
+
 
     @Override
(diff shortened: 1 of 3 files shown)


litetex (changes requested):
Didn't review it in detail but had a quick look:
There are serious formatting and code style problems

Example:
```java
    private void updateResolutions() {


        if (getPreferenceManager().getSharedPreferences()
                .getBoolean(getString(R.string.show_higher_resolutions_key), false)) {

            //if setting was turned on, enable additional resolutions
            showHigherResolutions(true);

        } else {

            showHigherResolutions(false);


        }
    }
```

I'm also pretty sure that not all fields are required

Coder246:
You don't allow lines longer than 100 characters.
What is the preferred way to deal with them?

litetex:
> You don't allow lines longer than 100 characters.
> What is the preferred way to deal with them?

Uhm using a new line?

Coder246:
Then everything should be fine.
I simply used a new line when i run out of characters.

litetex:
@Coder246 
Not sure what you are talking about but the above method alone has at least the following problems:
* way to many empty new lines
* The comment is formatted incorrectly
* The if block is useless, this can be done inline

opusforlife2:
Isn't code formatting enforced by a linter? And what about Checkstyle?

PR #8279, opened 2022-04-22
Adapt opacity of popup close button to allow touches in other apps on Android >=12

### LINKED ISSUE #6770: Can't interact with app/screen below popup player on Android 12

TiA4f8R:
#### What is it?
- [x] Bugfix (user facing)

This PR sets the opacity, on Android 12 and higher, of the popup close button to [the maximum one which allows touches on other apps for these Android versions](https://developer.android.com/reference/android/view/WindowManager.LayoutParams#maximum-obscuring-opacity).

See the documentation of the [`FLAG_NOT_TOUCHABLE`](https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_NOT_TOUCHABLE) flag, used in the popup close button, for more details.

These changes have been tested on an Android 12 emulator and by several users (from [the corresponding issue](https://github.com/TeamNewPipe/NewPipe/issues/6770) and from [Reddit](https://www.reddit.com/r/NewPipe/comments/u69r9c/about_non_interactivity_on_other_apps_than/)) and the changes restore the ability to interact with other apps on Android 12 (and probably higher, I didn't test the changes on Android 13 Developer Previews at all).

Remember that **these changes only apply on Android 12 and higher**.

- Before:

  before_opacity_change

- After:

  after_opacity_change

- Fixes #6770

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
@@ -78,6 +78,20 @@ public final class PlayerHelper {
     private static final NumberFormat SPEED_FORMATTER = new DecimalFormat("0.##x");
     private static final NumberFormat PITCH_FORMATTER = new DecimalFormat("##%");
 
+    /**
+     * Maximum opacity allowed for Android 12 and higher to allow touches on other apps when using
+     * NewPipe's popup player.
+     *
+     * <p>
+     * This value is hardcoded instead of being get dynamically with the method linked of the
+     * constant documentation below, because it is not static and popup player layout parameters
+     * are generated with static methods.
+     * </p>
+     *
+     * @see WindowManager.LayoutParams#FLAG_NOT_TOUCHABLE
+     */
+    private static final float MAXIMUM_OPACITY_ALLOWED_FOR_S_AND_HIGHER = 0.8f;
+
     @Retention(SOURCE)
     @IntDef({AUTOPLAY_TYPE_ALWAYS, AUTOPLAY_TYPE_WIFI,
             AUTOPLAY_TYPE_NEVER})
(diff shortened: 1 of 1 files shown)


litetex (approved):
A pretty small fix for a pretty big problem.
This is too good to be true, but it looks like it is.

LGTM

Good work!

litetex:
@TiA4f8R 
I have one question though: Why is the full opacity player not a problem?

TiA4f8R:
@litetex

I don't really know what to say:

- this should not really work according to the rule `Multiple SAW windows with enough transparency` described in the [`FLAG_NOT_TOUCHABLE`](https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_NOT_TOUCHABLE) flag, unless this rule only applies for windows with this flag, which seems to be the case (in this case, the documentation is not clear at all...);
- if we are in the case above, the first part of the rule `One SAW window with enough transparency` is now respected and the second part seems to be respected since the beginning (maybe not when the popup player itself and the close button are overlapping): the rule `One SAW window with enough transparency` is now fully respected and the touch events are delivered.

rawlife56:
@TiA4f8R The touch inputs are not still working in some protected screens like Google Play Store, app updates, Google app security settings screen. Is this expected behavior?

Bruceforce:
@rawlife56 this was/is also the case for Android 11 (at least for me) so this would be a separate issue in my opinion.

rawlife56:
> @rawlife56 this was/is also the case for Android 11 (at least for me) so this would be a separate issue in my opinion.

Good to know. Couldn't test as there wasn't any A11 device around. Hopefully work on native PIP mode support wouldn't be less prioritized.

PR #8287, opened 2022-04-23
replace possible dirty URL with clean URL when sharing

### LINKED ISSUE #3247: "Clean" URL when sharing (remove all parameters that are not video ID)

SydneyDrone:
#### What is it?
- [x] Bugfix (user facing)

- The share and open in browser options use a clean URL instead of possibly "dirty" origin URL
- Finished undone work in #3262
- every call of `ShareUtils.shareText()` and `ShareUtils.openUrlInBrowser()` is checked

- Fixes #3247 

#### Test Building
The basic sharing function works well, but in order to build a dirty URL for further testing, we need to add some parameters. I've tried every parameter listed in the [guidance of youtube player parameters](https://developers.google.com/youtube/player_parameters), and the only parameter that works in the web player is `start`, but it neither work at NewPipe(`start=30` should work the same as `t=30`) nor shown in the original URL shared in the prefix version.
So I didn't write any test yet, and helps are needed here.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
@@ -207,12 +207,12 @@ public class ChannelFragment extends BaseListInfoFragment<StreamInfoItem, Channe
                 break;
             case R.id.menu_item_openInBrowser:
                 if (currentInfo != null) {
-                    ShareUtils.openUrlInBrowser(requireContext(), currentInfo.getOriginalUrl());
+                    ShareUtils.openUrlInBrowser(requireContext(), currentInfo.getUrl());
                 }
                 break;
             case R.id.menu_item_share:
                 if (currentInfo != null) {
-                    ShareUtils.shareText(requireContext(), name, currentInfo.getOriginalUrl(),
+                    ShareUtils.shareText(requireContext(), name, currentInfo.getUrl(),
                             currentInfo.getAvatarUrl());
                 }
                 break;
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/playlist/PlaylistFragment.java
@@ -226,11 +226,11 @@ public class PlaylistFragment extends BaseListInfoFragment<StreamInfoItem, Playl
                 NavigationHelper.openSettings(requireContext());
                 break;
             case R.id.menu_item_openInBrowser:
-                ShareUtils.openUrlInBrowser(requireContext(), url);
+                ShareUtils.openUrlInBrowser(requireContext(), currentInfo.getUrl());
                 break;
             case R.id.menu_item_share:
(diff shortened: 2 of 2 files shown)


opusforlife2:
> So I didn't write any test yet, and helps are needed here.

So let's make this a draft.

SydneyDrone:
> So let's make this a draft.
@opusforlife2 Make a PR draft seems to prevent it from being reviewed. But right now I really need some suggestions. Should I open it again?

mhmdanas:
@SydneyDrone draft PRs can still be reviewed.

opusforlife2:
They just cannot be merged. Which is what we want here.

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
```diff
@@ -188,6 +188,9 @@ public void onCreateOptionsMenu(@NonNull final Menu menu,
         }
     }
 
+    /**
+     * fix github issue #3262: replace possible dirty URL with clean URL when sharing
+     */
```
Stypox, line 193: I don't think it is a good idea to add these javadocs here, I would just add a
```
// share the clean url instead of the dirty one
```
before every changed line

Stypox (changes requested):
> I've tried every parameter listed in the [guidance of youtube player parameters](https://developers.google.com/youtube/player_parameters), and the only parameter that works in the web player is start, but it neither work at NewPipe(start=30 should work the same as t=30) nor shown in the original URL shared in the prefix version.

Oh, so we just need to find some urls with many parameters? Maybe there are other services which use parameters?

By the way, you don't seem to have changed any line related to the video detail fragment, so the share button there will behave the same. Please check that.

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelFragment.java
SydneyDrone: I've reset the head and deleted the annotation.

SydneyDrone:
> Oh, so we just need to find some urls with many parameters? Maybe there are other services which use parameters?

Since we only clean those parameters in the share step (which is the last step), I suppose this action will not affect other services.

> By the way, you don't seem to have changed any line related to the video detail fragment, so the share button there will behave the same. Please check that.

I've checked every other use of methods (including those in the `VideoDetailFragment.Java`) such as `ShareUtils.openUrlInBrowser()` or `ShareUtils.shareUrl()`, and they all use `currentInfo.getUrl()` as parameter instead of `currentInfo.getOriginalUrl()`, so video detail fragment doesn't need any change.

Stypox:
Closing as there has not been any recent activity, and the correct way to solve this issue would be to refactor link handling in the extractor and then allow to user to choose how to share the URL among various options (e.g. shortened, alternative frontend, include time, ...). Thank you anyway to everyone involved!

Issue #8238, opened 2022-04-16
[YT] Throttling of (60fps) streams


litetex:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.22.2 / dev

### Steps to reproduce the bug

This is sometimes quite hard to reproduce reliable but I can reproduce it doing the following:

0. Clearing all caches and stop the app
1. Open a YT video with 60fps and quality >720p - it might work normally but let it run for at least 1min.
2. Open another YT video with 60fps
3. Notice that it (initially) loads forever and the network speed is at ~33kB/s.

I can reproduce the issue on multiple phones/emulators with the current released version (for 60fps video streams) and the current dev branch (happens on ALL video streams).

### Expected behavior

No throttling occurs

### Actual behavior

YT (video-only) streams are (initially) throttled to ~33kB/s = ~256kBit/s (suspicious number since this is exactly or very close to 2^8)

### Screenshots/Screen recordings

Example video:
https://www.youtube.com/watch?v=hWihEDctPKA

**Version 0.22.2**

Huawei P20 lite:
<img src="https://user-images.githubusercontent.com/40789489/163690306-6d28bc62-af8e-41f2-b947-07fd3a7e6541.png" height=400 />

Android studio network inspector with emulator:
![grafik](https://user-images.githubusercontent.com/40789489/163692358-bf9ccdfe-9d5f-415e-a13e-a9991f98ff68.png)

**Dev-Branch**
Android emulator:
![grafik](https://user-images.githubusercontent.com/40789489/163691040-94906679-04ee-415b-9397-29ba971219e2.png)

### Affected Android/Custom ROM version

doesn't matter

### Affected device model

doesn't matter

### Additional information

After some digging down we believe that the problem is likely caused by a ExoPlayer bug or bad combination with YT:
[...]

PR #8288, opened 2022-04-24
Set the progressive_load_interval_default_value to 64 KiB

litetex:
#### What is it?
- [x] Workaround (user facing)

Set the progressive_load_interval_default_value to 64 KiB

This should workaround the problems from #8238.
https://github.com/TeamNewPipe/NewPipe/issues/8238#issuecomment-1106096644

As a theoretical drawback stuttering may occur on very low bandwith connections, but I think this is negligible (also users can change it in the settings).

- Works around #8238

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/res/values/settings_keys.xml
+++ b/app/src/main/res/values/settings_keys.xml
@@ -56,7 +56,7 @@
     </string-array>
 
     <string name="progressive_load_interval_key">progressive_load_interval</string>
-    <string name="progressive_load_interval_default_value">default</string>
+    <string name="progressive_load_interval_default_value">64</string>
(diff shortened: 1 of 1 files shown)


Stypox:
I don't think this PR solves anything, see how `progressive_load_interval_default_value` is used here. You are basically using a different *name* for the default value, without changing its effect, that is, setting the default loading check interval to `1MiB`. https://github.com/TeamNewPipe/NewPipe/blob/9b3e43ffc172023ce26e8f368f2cd761369a2db7/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java#L424-L435

PR #8302, opened 2022-04-28
Use 64 KiB as the default progressive load interval

Stypox:
#### What is it?
- [x] Workaround (user facing)

#### Description of the changes in your PR (partially copied from #8288)
Use 64 KiB as the default progressive load interval. This ensures a small value is used by default, solving buffering issues at the beginning of videos.

This should workaround the problems from #8238.
https://github.com/TeamNewPipe/NewPipe/issues/8238#issuecomment-1106096644

As a theoretical drawback stuttering may occur on very low bandwith connections, but I think this is negligible (also users can change it in the settings).

- Works around #8238
- Supersedes #8288

#### APK testing

#### Due diligence


app/src/main/res/values/strings.xml
```diff
@@ -733,7 +733,7 @@
     <string name="on">On</string>
     <string name="off">Off</string>
     <!-- Progressive Load Interval -->
-    <string name="progressive_load_interval_default">ExoPlayer default</string>
+    <string name="progressive_load_interval_exoplayer_default">ExoPlayer default</string>
```
mhmdanas, line 736: Shouldn't the corresponding string in other languages be renamed too?

Stypox: Yes ;-)

Issue #4422, opened 2020-10-06
Hover events trigger a white overlay on Samsungs and other devices


hipunk:
### Checklist

- [X] I am using the latest version. 
- [ ] I checked, but didn't find any duplicates of this issue in the repo. This is a duplicate of: #4357 but it doesn't only involve a mouse but also loads of Samsung phones. o_O  
- [X] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md
- [X] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug and the actual behaviour:

1. Open a video in fullscreen
2. Hover your finger over the screen within a 3cm distance on a Samsung Galaxy S4 or newer. Or use a S-Pen with a Samsung Galaxy Note then hover it over the screen or move Androids mouse cursor using a bluetooth or a USB mouse 
3. A milky white transparent overlay appears over the whole screen, some hover event is apparently being triggered and makes an overlay appear! :)

### Expected behavior

No white overlay.
Suggestion: Show controls when you detect a hover event! It's fine like it was before 0.20 as well.

### Device info

 - Device models affected: GT-I9500, GT-I9506, GT-I9515, GT-I9505, GT-N7000, GT-N7100, SM-G930, SM-G935, probably many more + all devices with bluetooth and USB mice

opusforlife2:
What?

PR #8316, opened 2022-04-30
Fix persistent hover overlay when in desktop/DeX mode or using a mouse/non-touch input

### LINKED ISSUE #4357: Physical Mouse: remove the white overlay on mousover from the videoplayer window
#### Version

- 0.20.0 RC

It would be nice if you disable the selection feedback for the video window. At least in full screen.

When using a mouse with android there is this problem that the mouse keeps selecting the video window on mouse-over. This causes a white haze in front of the video, messing up the colours. _The big problem is: this also happens in fullscreen._

After 15 whole seconds the mouse fades, but the issue persists because the mouse seems to be there still.
#### Additional context

It looks like looking at the video through a tinted glass pane.

Clicking the video and holding seems to alleviate the problem. Except, then the controls pop up again.

This only happens in the RC build #4333. (edited: I think I saw it happen in 19.8, but no matter what I try cannot reproduce it in that version.)

People using a mouse can view the video in full glory, not through some selection feedback filter.

han-sz:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

A non-touch input device that supports hover events results in a persistent transparent overlay on top of the video player. When using an Android device in desktop mode or with a mouse, the content is obscured with said overlay. 

The bug was first reported here: https://github.com/TeamNewPipe/NewPipe/issues/4357

This PR adds an `isDesktopMode` check and removes the hover overlay for the `detailThumbnailRootLayout` only when in desktop mode (DeX, Android desktop, etc.)

- Before:
Screenshot_20220501-022300_NewPipe

- After:
Screenshot_20220501-022400_NewPipe fix_video_mouse_hover_overlay

- Fixes #4357 

#### APK testing 

APK for testing: [app-debug.apk.zip](https://github.com/TeamNewPipe/NewPipe/files/8597415/app-debug.apk.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
@@ -100,20 +100,18 @@ public final class DeviceUtils {
      */
     public static boolean isDesktopMode(final Context context) {
         // Adapted from https://stackoverflow.com/a/64615568
-        // to check for all devices that have an active cursor
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-            final InputManager im = (InputManager) context.getSystemService(INPUT_SERVICE);
-            for (final int id : im.getInputDeviceIds()) {
-                final InputDevice inputDevice = im.getInputDevice(id);
-                if (
-                        inputDevice.supportsSource(InputDevice.SOURCE_BLUETOOTH_STYLUS)
-                                || inputDevice.supportsSource(InputDevice.SOURCE_MOUSE)
-                                || inputDevice.supportsSource(InputDevice.SOURCE_STYLUS)
-                                || inputDevice.supportsSource(InputDevice.SOURCE_TOUCHPAD)
-                                || inputDevice.supportsSource(InputDevice.SOURCE_TRACKBALL)
-                ) {
-                    return true;
-                }
+        // to check for all input devices that have an active cursor
+        final InputManager im = (InputManager) context.getSystemService(INPUT_SERVICE);
+        for (final int id : im.getInputDeviceIds()) {
+            final InputDevice inputDevice = im.getInputDevice(id);
+            if (
+                    inputDevice.supportsSource(InputDevice.SOURCE_BLUETOOTH_STYLUS)
+                            || inputDevice.supportsSource(InputDevice.SOURCE_MOUSE)
+                            || inputDevice.supportsSource(InputDevice.SOURCE_STYLUS)
+                            || inputDevice.supportsSource(InputDevice.SOURCE_TOUCHPAD)
+                            || inputDevice.supportsSource(InputDevice.SOURCE_TRACKBALL)
+            ) {
+                return true;
             }
         }
 
--- a/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
@@ -89,11 +89,11 @@ public final class DeviceUtils {
     }
 
     public static boolean isDesktopMode(final Context context) {
-        final boolean isDesktopMode = ContextCompat.getSystemService(context, UiModeManager.class)
-                .getCurrentModeType() == Configuration.UI_MODE_TYPE_DESK;
-
+        if (ContextCompat.getSystemService(context, UiModeManager.class)
+                .getCurrentModeType() == Configuration.UI_MODE_TYPE_DESK) {
+            return true;
+        }
         // DeX check for standalone and multi-window mode
-        boolean isDeXMode = false;
         try {
             final Configuration config = context.getResources().getConfiguration();
             final Class<?> configClass = config.getClass();
@@ -102,7 +102,7 @@ public final class DeviceUtils {
             final int currentMode =
                     configClass.getField("semDesktopModeEnabled").getInt(config);
             if (semDesktopModeEnabledConst == currentMode) {
-                isDeXMode = true;
+                return true;
             }
         } catch (final NoSuchFieldException | IllegalAccessException e) {
             // empty
@@ -122,14 +122,14 @@ public final class DeviceUtils {
                 final int enabled = (int) getEnabledMethod.invoke(desktopModeState);
                 if (enabled == desktopModeStateClass
                         .getDeclaredField("ENABLED").getInt(desktopModeStateClass)) {
-                    isDeXMode = true;
+                    return true;
                 }
             } catch (NoSuchFieldException | NoSuchMethodException
                     | IllegalAccessException | InvocationTargetException e) {
                 // Device does not support DeX 3.0
             }
         }
-        return isDesktopMode || isDeXMode;
+        return false;
     }
 
     public static boolean isTablet(@NonNull final Context context) {
--- a/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
@@ -1,5 +1,6 @@
 package org.schabi.newpipe.util;
 
+import android.annotation.SuppressLint;
 import android.app.UiModeManager;
 import android.content.Context;
 import android.content.pm.PackageManager;
@@ -21,6 +22,9 @@ import androidx.preference.PreferenceManager;
 
 import org.schabi.newpipe.App;
 import org.schabi.newpipe.R;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 
 public final class DeviceUtils {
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -85,6 +89,50 @@ public static boolean isTv(final Context context) {
         return DeviceUtils.isTV;
     }
 
+    public static boolean isDesktopMode(final Context context) {
+        final boolean isDesktopMode = ContextCompat.getSystemService(context, UiModeManager.class)
+                .getCurrentModeType() == Configuration.UI_MODE_TYPE_DESK;
```
Stypox, line 94:
```suggestion
        if (ContextCompat.getSystemService(context, UiModeManager.class)
                .getCurrentModeType() == Configuration.UI_MODE_TYPE_DESK) {
            return true;
        }
```

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -85,6 +89,50 @@ public static boolean isTv(final Context context) {
...
+        boolean isDeXMode = false;
+        try {
+            final Configuration config = context.getResources().getConfiguration();
+            final Class<?> configClass = config.getClass();
+            final int semDesktopModeEnabledConst =
+                    configClass.getField("SEM_DESKTOP_MODE_ENABLED").getInt(configClass);
+            final int currentMode =
+                    configClass.getField("semDesktopModeEnabled").getInt(config);
+            if (semDesktopModeEnabledConst == currentMode) {
+                isDeXMode = true;
```
Stypox, line 106:
```suggestion
                return true;
```

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -85,6 +89,50 @@ public static boolean isTv(final Context context) {
...
+                        .getDeclaredMethod("getDesktopModeState");
+                final Object desktopModeState = getDesktopModeStateMethod
+                        .invoke(desktopModeManager);
+                final Class<?> desktopModeStateClass = desktopModeState.getClass();
+                final Method getEnabledMethod = desktopModeStateClass
+                        .getDeclaredMethod("getEnabled");
+                final int enabled = (int) getEnabledMethod.invoke(desktopModeState);
+                if (enabled == desktopModeStateClass
+                        .getDeclaredField("ENABLED").getInt(desktopModeStateClass)) {
+                    isDeXMode = true;
```
Stypox, line 126:
```suggestion
                    return true;
```

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -85,6 +89,50 @@ public static boolean isTv(final Context context) {
...
+                }
+            } catch (NoSuchFieldException | NoSuchMethodException
+                    | IllegalAccessException | InvocationTargetException e) {
+                // Device does not support DeX 3.0
+            }
+        }
+        return isDesktopMode || isDeXMode;
```
Stypox, line 133:
```suggestion
        return false;
```

Stypox (changes requested):
Thank you, code looks almost good. But return directly ;-)

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
han-sz: Thanks, resolved all suggestions

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -85,6 +89,50 @@ public static boolean isTv(final Context context) {
...
+                if (enabled == desktopModeStateClass
+                        .getDeclaredField("ENABLED").getInt(desktopModeStateClass)) {
+                    return true;
+                }
+            } catch (NoSuchFieldException | NoSuchMethodException
+                    | IllegalAccessException | InvocationTargetException e) {
+                // Device does not support DeX 3.0
+            }
+        }
+        return false;
```
litetex, line 133: Where did you get these values from?

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -85,6 +89,50 @@ public static boolean isTv(final Context context) {
...
+    }
+
```
litetex, line 135: Note: 
This always executed on ALL devices when the VideoDetailFragment is initialized. 
Combined with the use of Reflection - which is also not that performant (https://stackoverflow.com/a/435568) - this might be an bigger issue.

litetex (changes requested):
Considering NewPipe is designed for Android which is designed for touch input devices I'm not sure if we should implement this in the first place.

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
han-sz: Samsung DeX check code can be found from Samsung's developer guide:
https://developer.samsung.com/samsung-dex/modify-optimizing.html

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
han-sz: How long do you think  a one-time check will take? On an old device this code takes <0.001ms and is run only once in view init. Line 114 guards against any reflection calls and on line 101 there is one reflection call which will throw and fail fast on non-desktop devices.

Stypox: @litetex performance is negligible here. This is a one-time check, and even though reflection is ten times slower than normal class access, according to the SO you linked it still takes <10^-6s, which is not a problem. Maybe let's just add a note in the javadoc so that this method is not misused in the future.

han-sz: Yup, I have added a jsdoc comment with a note that this should ideally only be invoked once on view init

Stypox (approved):
This looks good to me, thanks :-)
Let's wait for what @litetex has to say

cybersphinx:
I've added a check for devices with cursors: https://github.com/cybersphinx/NewPipe/commit/e86a34b1c4eedead1bc791b7869c03f833f5265a ([full branch](https://github.com/TeamNewPipe/NewPipe/compare/dev...cybersphinx:NewPipe:fix_video_mouse_hover_overlay))

With this it works on my tv box (and phone with attached mouse), I don't have any of the fancy devices mentioned in #4422 with stylus or finger hover detection.

There's an SDK version check around it that can/should be removed if minSdk is raised to 21.

opusforlife2:
> There's an SDK version check around it that can/should be removed if minSdk is raised to 21.

The barrier has been lifted.

cybersphinx:
Updated [my branch](https://github.com/TeamNewPipe/NewPipe/compare/dev...cybersphinx:NewPipe:fix_video_mouse_hover_overlay) to remove the check, and fix a silly oversight.

Stypox:
@cybersphinx do you want me to add your commits here? Or do you want to open a new PR?

cybersphinx:
If you can add them here I guess that's the best place.

Stypox:
Done; @litetex could you review again?

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -84,6 +92,73 @@ public static boolean isTv(final Context context) {
         return DeviceUtils.isTV;
     }
 
+    /**
+     * Checks if the device is in desktop or DeX mode. This function should only
+     * be invoked once on view load as it is using reflection for the DeX checks.
+     * @param context the context to use for services and config.
+     * @return true if the Android device is in desktop mode or using DeX.
+     */
+    public static boolean isDesktopMode(final Context context) {
```
TobiGr, line 101: this method is complex and should only be evaluated once. Is there a reason not to cache the result in a class var ? e.g. 
``` Java
private static Boolean isDesktopMode
```

han-sz: The check relies on both input devices connected and the state of desktop mode when the function is called.

Example scenario where cacheing once will result in wrong output: Changing to Desktop mode/docking a device can happen after the variable is computed and cached. When changing to desktop mode the view lifecycle hooks rerun to update the layout and is when this is computed

TobiGr: oh ok. I did not think of that scenario :+1:

TobiGr (approved):

AudricV:
I rebased the PR and pushed a commit to the branch which improves the code added, let me know if it is good for you :)

han-sz:
Thanks @AudricV, looks good to me!

Andronx:
Issue solved in v0.25.0.

PR #8332, opened 2022-05-01
Fix Routeractivity theming

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
When the app is not running in the foreground and started via ``RouterActivity`` (e.g. clicking/sharing a link to NewPipe) incorrect things start to happens:
* The UI has incorrect colors (seen in #8179)
  * The day/night mode of the app is ignored
* The language is the device language

Why does this happen?
* The RouterActivityThemes are overriding the ``color*``-attributes (``colorPrimary``, ``colorPrimaryDark``, ...). I'm not sure why this was done in the first place, however it causes the theming issues on the UI.
* RouterActivity misses some code from MainActivity, e.g. ``assureCorrectAppLanguage`` and ``ThemeHelper.setDayNightMode``. This causes further problems like the incorrect language or wrong colored theme elements.

#### Before/After Screenshots/Screen Record
Done via the test manual below:

- Before: 
grafik grafik grafik grafik

- After: 
grafik grafik grafik grafik

#### Fixes the following issue(s)

- Fixes #8179; maybe more issues, hard to check when there are >1k open ones

#### APK testing 
The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

##### Test manual
* Change theme to dark
* Change the app-language to something other then the phone-languguage (e.g. German)
* Restart the app (so that the language is applied)
* Close the app completely
* Share a stream link to NewPipe (that you never watched before)
* the ask what to do dialog should come up
* click add to playlist

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #3282, opened 2020-03-27
Check code style with Checkstyle

wb9688:
#### What is it?

- [ ] Bug fix
- [ ] Feature
- [x] Meta

#### Long description of the changes in your PR

This PR will let Travis check the code style of all PRs. Do you think this should be run before every build as well?

I'm also going to fix all the current code style issues, except in GigaCrap, so I excluded that already in `build.gradle`.

The Checkstyle configuration is based on [this](https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/sun_checks.xml), but I disabled some checks.

I'm going to implement this for NewPipeExtractor as well.

#### Agreement

- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

PR #8340, opened 2022-05-02
Fix "Add to playlist" not working and cleanup "RouterActivity" choice handling

litetex:
#### What is it?
- [x] Bugfix (user facing)

Problem: Setting _Preferred 'open' action_ to "Add to playlist" does nothing when opening a video
Discovered while working on #8332 

This PR also cleans up how the choices are handled inside ``RouterActivity``. It was until now possible to e.g. download a channel (which resulted in an error/unsupported url page) due to incorrect choice-availability-checks.
Now everything is centralized in open place and should be easier to maintain.

- Before: _Nothing happens when clicking a video; Behaves like "Show info"_
- After:

https://user-images.githubusercontent.com/40789489/166318116-2016161a-d8be-4ba4-922f-0bab05bb821c.mp4

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -264,15 +264,10 @@ public class RouterActivity extends AppCompatActivity {
                         getString(R.string.preferred_open_action_default));
 
         final String showInfoKey = getString(R.string.show_info_key);
-        final String videoPlayerKey = getString(R.string.video_player_key);
-        final String backgroundPlayerKey = getString(R.string.background_player_key);
-        final String popupPlayerKey = getString(R.string.popup_player_key);
-        final String downloadKey = getString(R.string.download_key);
-        final String alwaysAskKey = getString(R.string.always_ask_open_action_key);
 
-        if (selectedChoiceKey.equals(alwaysAskKey)) {
-            final List<AdapterChoiceItem> choices
-                    = getChoicesForService(currentService, currentLinkType);
+        if (selectedChoiceKey.equals(getString(R.string.always_ask_open_action_key))) {
+            final List<AdapterChoiceItem> choices =
+                    getChoicesForService(currentService, currentLinkType);
 
             switch (choices.size()) {
                 case 1:
@@ -285,36 +280,39 @@ public class RouterActivity extends AppCompatActivity {
                     showDialog(choices);
                     break;
             }
-        } else if (selectedChoiceKey.equals(showInfoKey)) {
-            handleChoice(showInfoKey);
-        } else if (selectedChoiceKey.equals(downloadKey)) {
-            handleChoice(downloadKey);
+        } else if (selectedChoiceKey.equals(showInfoKey)
+                || selectedChoiceKey.equals(getString(R.string.download_key))
+                || selectedChoiceKey.equals(getString(R.string.add_to_playlist_key))
+        ) {
+            handleChoice(selectedChoiceKey);
         } else {
             final boolean isExtVideoEnabled = preferences.getBoolean(
                     getString(R.string.use_external_video_player_key), false);
             final boolean isExtAudioEnabled = preferences.getBoolean(
                     getString(R.string.use_external_audio_player_key), false);
-            final boolean isVideoPlayerSelected = selectedChoiceKey.equals(videoPlayerKey)
-                    || selectedChoiceKey.equals(popupPlayerKey);
-            final boolean isAudioPlayerSelected = selectedChoiceKey.equals(backgroundPlayerKey);
+            final boolean isVideoPlayerSelected =
+                    selectedChoiceKey.equals(getString(R.string.video_player_key))
+                            || selectedChoiceKey.equals(getString(R.string.popup_player_key));
+            final boolean isAudioPlayerSelected =
+                    selectedChoiceKey.equals(getString(R.string.background_player_key));
 
-            if (currentLinkType != LinkType.STREAM) {
-                if (isExtAudioEnabled && isAudioPlayerSelected
-                        || isExtVideoEnabled && isVideoPlayerSelected) {
-                    Toast.makeText(this, R.string.external_player_unsupported_link_type,
-                            Toast.LENGTH_LONG).show();
-                    handleChoice(showInfoKey);
-                    return;
-                }
+            if (currentLinkType != LinkType.STREAM
+                    && (isExtAudioEnabled && isAudioPlayerSelected
+                    || isExtVideoEnabled && isVideoPlayerSelected)
+            ) {
+                Toast.makeText(this, R.string.external_player_unsupported_link_type,
+                        Toast.LENGTH_LONG).show();
+                handleChoice(showInfoKey);
+                return;
             }
 
-            final List<StreamingService.ServiceInfo.MediaCapability> capabilities
-                    = currentService.getServiceInfo().getMediaCapabilities();
+            final List<StreamingService.ServiceInfo.MediaCapability> capabilities =
+                    currentService.getServiceInfo().getMediaCapabilities();
 
             boolean serviceSupportsChoice = false;
             if (isVideoPlayerSelected) {
                 serviceSupportsChoice = capabilities.contains(VIDEO);
-            } else if (selectedChoiceKey.equals(backgroundPlayerKey)) {
+            } else if (isAudioPlayerSelected) {
                 serviceSupportsChoice = capabilities.contains(AUDIO);
             }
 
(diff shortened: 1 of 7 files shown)


opusforlife2:
Code blocks in the title break npbot's PR link fetching mechanism in the IRC channel.

Stypox:
> Code blocks in the title break npbot's PR link fetching mechanism in the IRC channel.

Time to fix npbot then, not to change the PR title ;-)

opusforlife2:
Just something to keep in mind when creating PR titles, until the bug _is_ fixed.

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -425,87 +466,61 @@ private void showDialog(final List<AdapterChoiceItem> choices) {
...
                 R.drawable.ic_headset);
-        final AdapterChoiceItem addToPlaylist = new AdapterChoiceItem(
-                getString(R.string.add_to_playlist_key), getString(R.string.add_to_playlist),
-                R.drawable.ic_add);
+        final AdapterChoiceItem popupPlayer = new AdapterChoiceItem(
+                getString(R.string.popup_player_key), getString(R.string.popup_player),
+                R.drawable.ic_picture_in_picture);
 
+        final List<AdapterChoiceItem> returnedItems = new ArrayList<>();
+        returnedItems.add(showInfo); // Always present
```
Stypox, line 483: Uhm, why is it always present? Shouldn't it be hidden in cases where it would do the exact same thing as `videoPlayer`, to save some space?

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -257,70 +258,110 @@ private void showUnsupportedUrlDialog(final String url) {
...
-                }
+            final boolean isVideoPlayerSelected =
+                    selectedChoice.equals(getString(R.string.video_player_key))
+                            || selectedChoice.equals(getString(R.string.popup_player_key));
+            final boolean isAudioPlayerSelected =
+                    selectedChoice.equals(getString(R.string.background_player_key));
+
+            if (currentLinkType != LinkType.STREAM
+                    && (isExtAudioEnabled && isAudioPlayerSelected
+                    || isExtVideoEnabled && isVideoPlayerSelected)
```
Stypox, line 295: Parenthesis

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -257,70 +258,110 @@ private void showUnsupportedUrlDialog(final String url) {
...
+                Toast.makeText(this, R.string.external_player_unsupported_link_type,
+                        Toast.LENGTH_LONG).show();
+                handleChoice(getString(R.string.show_info_key));
+                return;
             }
 
-            final List<StreamingService.ServiceInfo.MediaCapability> capabilities
-                    = currentService.getServiceInfo().getMediaCapabilities();
+            final List<StreamingService.ServiceInfo.MediaCapability> capabilities =
+                    currentService.getServiceInfo().getMediaCapabilities();
```
Stypox, line 304: Uhm, why isn't checkstyle complaining? The `=` should be on the newline

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -257,70 +258,110 @@ private void showUnsupportedUrlDialog(final String url) {
...
 
-            boolean serviceSupportsChoice = false;
-            if (isVideoPlayerSelected) {
-                serviceSupportsChoice = capabilities.contains(VIDEO);
-            } else if (selectedChoiceKey.equals(backgroundPlayerKey)) {
-                serviceSupportsChoice = capabilities.contains(AUDIO);
+            // Check if the service supports the choice
+            if (isVideoPlayerSelected && capabilities.contains(VIDEO)
+                    || isAudioPlayerSelected && capabilities.contains(AUDIO)) {
```
Stypox, line 308: Parenthesis

Stypox (changes requested):
Code mostly looks good

app/src/main/java/org/schabi/newpipe/RouterActivity.java
litetex: 1. You would end up with an action "Video player" that behaves like "Show Info". If thought logically it should be exactly opposite: We shouldn't show "Video player" when it can be done with "Show info".
2. This would require a ton of if logic and another logical layer on top of the choices. Which is in my opinion overkill.
3. It may be unexpected for the user if no "Show info" or "Video player" is present.
4. Saving space is not relevant because there is still a huge portion of users that have all entries anyway visible (e.g. if the have auto-play enabled - which is default).

app/src/main/java/org/schabi/newpipe/RouterActivity.java
Stypox: Actually, from the checkstyle link you provided, the default value is `nl`, i.e. putting the operator on new line. It always has been this way in NewPipe, so I think you should stick to that (as you did e.g. for `&&` and `||` in `if`s), though this is really minor.

app/src/main/java/org/schabi/newpipe/RouterActivity.java
Stypox: Mmmh ok, then in a separate PR the router activity should be made scrollable. When I have the phone in landscape I can't see all options.

litetex: Oh, I didn't notice that.
Sound's like a new issue 😉

app/src/main/java/org/schabi/newpipe/RouterActivity.java
litetex: > Actually, from the checkstyle link you provided, the default value is nl, i.e. putting the operator on new line.

Uhm no:
grafik

It only checks the tokens in the right (not left) column.


> It always has been this way in NewPipe...

Also no. Example: https://github.com/TeamNewPipe/NewPipe/blob/1daece3bee8a2955423750706199819338724f8f/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L138-L147 or https://github.com/TeamNewPipe/NewPipe/blob/1daece3bee8a2955423750706199819338724f8f/app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java#L131-L139

Update: I went a bit through <sub>git</sub> history and found these commits:
* https://github.com/TeamNewPipe/NewPipe/pull/3282 - https://github.com/TeamNewPipe/NewPipe/pull/3282/commits/fda5405e48f7cb1f4f4b9df89052d55c90e8b167 (looks like the origin of the ``=`` on new line)
* https://github.com/TeamNewPipe/NewPipe/pull/2907 - https://github.com/TeamNewPipe/NewPipe/commit/d8f7db4715cea3c405e11ba28cfa809d9c6447a4 (overwrites/reverts some of these changes)

So apparently someone (*cough* [wb9688](https://github.com/wb9688)) introduced his personal code "style" but didn't persist the changes inside checkstyle and also no one noticed it inside the PR...

Anyway I - and I also  think most other developers (as seen above) - favor the ``=`` after assignment style. So I would stick with that ;)

app/src/main/java/org/schabi/newpipe/RouterActivity.java
litetex: ``&&`` binds stronger than ``||`` and I know that (also most Java devs should probably also know that...).

However - as I have to admit - this is not very clearly readable. 

Fixed.

app/src/main/java/org/schabi/newpipe/RouterActivity.java
Stypox: Mmmh ok. I still don't see why all operators should be on newlines (like you did for `||` below here) but `=` should not, but anyway... If this is the standard across the industry, if you have a little spare time please open a PR to enforce such style in checkstyle

Stypox (approved):
Thank you! Code looks good to me. I also tested and everything works well.

A note for a separate PR: the *"processing, will take a moment"* toast should be shown on any action that waits for video info to be obtained. Currently it is only shown for "Add to playlist". I think the same string should also be used for actions in the long-press menu.

Stypox (changes requested):
Wait, when the router activity closes (because the dialog is dismissed), a strange animation is shown with a red line swiping through the screen. I thought this happened on dev, too, but it doesn't. I think it is some activity theming issue.

litetex:
> Wait, when the router activity closes (because the dialog is dismissed), a strange animation is shown with a red line swiping through the screen. I thought this happened on dev, too, but it doesn't. I think it is some activity theming issue.

<details><summary>I think you mean this or?</summary>

grafik NPRouterActivityP1

</details>

I will have a look into it...

litetex:
Look a easy fix:
https://stackoverflow.com/a/26927698

Oh we are still supporting Android 4.4.... So not an easy fix... yay
grafik

Fixed it only for Android 5+ up because Android 4.4 will soon be no longer supported.

And btw, this issue arose in https://github.com/TeamNewPipe/NewPipe/pull/8332. So a new PR would have been better :)

Stypox (approved):
Great, thank you :-)

PR #8380, opened 2022-05-10
Use ArraySet.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use `ArraySet` in more places, as it is a more memory-efficient implementation of `Set`.

- Fixes #

#### APK testing 

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -202,6 +202,7 @@ dependencies {
 /** AndroidX **/
     implementation 'androidx.appcompat:appcompat:1.3.1'
     implementation 'androidx.cardview:cardview:1.0.0'
+    implementation 'androidx.collection:collection-ktx:1.2.0'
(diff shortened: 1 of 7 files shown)


litetex (changes requested):
Thank you for the PR.

However vetoed from my side.
1. Please show me how this effects the performance on a up-to-date smartphone. I'm pretty sure that we are inside the ``ns`` (time) and ``kB``(ram) area which is insufficient considering the following trade-offs.
2. ArraySet is exclusively made for Android and not available inside Java. One possible effect: A lot of developers won't use it/may mishandle it because they don't know it and HashSet is normally sufficient.
3. It adds an additional library which has to be maintained, etc.
4. An arrayset doesn't work like an hashset, it may break code (https://stackoverflow.com/a/11934110)

PR #8381, opened 2022-05-10
Use image resource methods.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use the Drawable resource methods, as the compat view implementations inflate vector drawables on older API levels.

- Fixes #

#### APK testing 

#### Due diligence


litetex:
I'm not sure what the current PR tries to improve or why the current (before-PR) solution is bad.

However considering this notice:
https://developer.android.com/reference/android/widget/ImageView#setImageResource(int)
grafik

I don't have the feeling that our code becomes better with this PR...

Isira-Seneviratne:
> I'm not sure what the current PR tries to improve or why the current (before-PR) solution is bad.
> 
> However considering this notice: https://developer.android.com/reference/android/widget/ImageView#setImageResource(int) grafik
> 
> I don't have the feeling that our code becomes better with this PR...

The compatibility implementations of those methods already call `AppCompatResources.getDrawable()`.

TacoTheDank:
Mm, I still fail to see how this would be better, considering the warning above. I did look through the relevant compat code for more context.

Also, after checking the calls, I'm finding that some of them are using the platform `setImageResource`, not the compat version.

And there's some unrelated changes in the PR too.

This doesn't seem like a worthwhile change IMO.

Appreciate the effort, though 👍

PR #8382, opened 2022-05-10
Remove unnecessary compat method calls.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Remove compatibility method calls that aren't needed due to the `minSdk` being 19 (KitKat).

- Fixes #

#### APK testing 

#### Due diligence


litetex (approved):
Thank you for the PR.

LGTM however someone with more Android experience might have a look at this before merging :)

Worked well while doing a quick test.

Redirion:
even though these methods might not be of use currently, I see no advantage in removing them. androidx is updateable, the core Android OS on most devices is not due to lacking manufacturer support. So sticking to an updatable API seems advantageous to me.

Stypox:
I agree with @Redirion, I see no advantage

Isira-Seneviratne:
> even though these methods might not be of use currently, I see no advantage in removing them. androidx is updateable, the core Android OS on most devices is not due to lacking manufacturer support. So sticking to an updatable API seems advantageous to me.

I was under the impression that NewPipe Legacy is no longer being maintained, so I removed those method calls.

* `ViewCompat.setBackground()` isn't needed on API levels >= 16.
* `GestureDetectorCompat` isn't needed on API levels >= 17.
* `TextViewCompat.setCompoundDrawablesRelativeWithIntrinsicBounds()` isn't needed on API levels >= 18.

Stypox:
Mmmh ok. This is a micro-optimization, since using wrapped functions is slower and uses less apk size than using non-wrapped ones, but it is really a minuscole difference. But it actually makes sense to use non-compat methods to reduce code complexity.

> even though these methods might not be of use currently, I see no advantage in removing them. androidx is updateable, the core Android OS on most devices is not due to lacking manufacturer support. So sticking to an updatable API seems advantageous to me.

Actually, that does not hold well, since otherwise we should use androidx methods for every place possible, but that would be a nightmare. Let's remove them if possible.

PR #8405, opened 2022-05-16
Use tint methods.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use tint methods to set drawable colors.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
+++ b/app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
@@ -17,6 +17,7 @@ import static org.schabi.newpipe.player.helper.PlayerHelper.nextResizeModeAndSav
 import static org.schabi.newpipe.player.helper.PlayerHelper.retrieveSeekDurationFromPreferences;
 
 import android.content.Intent;
+import android.content.res.ColorStateList;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.Color;
@@ -162,8 +163,7 @@ public abstract class VideoPlayerUi extends PlayerUi
         binding.resizeTextView
                 .setText(PlayerHelper.resizeTypeOf(context, binding.surfaceView.getResizeMode()));
 
-        binding.playbackSeekBar.getThumb()
-                .setColorFilter(new PorterDuffColorFilter(Color.RED, PorterDuff.Mode.SRC_IN));
+        binding.playbackSeekBar.setThumbTintList(ColorStateList.valueOf(Color.RED));
         binding.playbackSeekBar.getProgressDrawable()
                 .setColorFilter(new PorterDuffColorFilter(Color.RED, PorterDuff.Mode.MULTIPLY));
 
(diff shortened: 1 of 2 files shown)


Stypox:
What is the advantage of using these methods instead of `PorterDuffColorFilter`?

TacoTheDank:
Hmm, this doesn't seem necessary in light of the coming support removal for APIs < 21. Opinions?

Isira-Seneviratne:
> > Opinions?
> 
> This depends on whether we drop support for API 19 in 0.24.0 (spoiler: YES!) or not. 
> 
> > What is the advantage of using these methods instead of PorterDuffColorFilter?
> 
> Answer is awaiting ;-)

The PorterDuff mode does not need to be specified when setting the tint (the default is SRC_IN).

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -484,10 +485,19 @@ private void initViews(@NonNull final PlayerBinding playerBinding) {
         binding.resizeTextView
                 .setText(PlayerHelper.resizeTypeOf(context, binding.surfaceView.getResizeMode()));
 
-        binding.playbackSeekBar.getThumb()
-                .setColorFilter(new PorterDuffColorFilter(Color.RED, PorterDuff.Mode.SRC_IN));
-        binding.playbackSeekBar.getProgressDrawable()
-                .setColorFilter(new PorterDuffColorFilter(Color.RED, PorterDuff.Mode.MULTIPLY));
+        tintThumbRed();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
Stypox, line 489: Please remove these checks, as now the minsdk is lollipop (21), then this PR is ready

Stypox (changes requested):

TacoTheDank:
Could you rebase the PR (on dev) and force-push to only have one commit? That will make it much easier to read :)

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
```diff
@@ -173,8 +172,9 @@ private void initViews() {
         playbackSpeedPopupMenu = new PopupMenu(context, binding.playbackSpeed);
         captionPopupMenu = new PopupMenu(themeWrapper, binding.captionTextView);
 
-        binding.progressBarLoadingPanel.getIndeterminateDrawable()
-                .setColorFilter(new PorterDuffColorFilter(Color.WHITE, PorterDuff.Mode.MULTIPLY));
+        binding.progressBarLoadingPanel.setIndeterminateTintList(ColorStateList.valueOf(Color
+                .WHITE));
```
Stypox, line 176:
```suggestion
        binding.progressBarLoadingPanel.setIndeterminateTintList(
                ColorStateList.valueOf(Color.WHITE));
```

Stypox (changes requested):
It doesn't seem to work correctly: the top image is the color of the seekbar before this PR, the bottom one is from this PR.
image
image

TacoTheDank:
Sorry for being annoying lol, but I'd like to request another rebase for readability :P

Stypox:
Now it seems to work as it should but... why does commit 763f0bf7ab31f3e7badb621c71cd4c8e83ffc086 solve the problem? I mean, why does `setThumbTintList` do the correct thing while `setProgressTintList` doesn't, requiring you to use `getProgressDrawable().setTint`? I still have some doubts about this PR, since it doesn't seem to provide better code than before (I mean, what was wrong with specifying `PorterDuff` mode?) but seems to introduce possibility for bugs.

Isira-Seneviratne:
I undid a couple of the changes.

Stypox:
I'm sorry, I have to reject this pull request. The original code calls `setColorFilter`, while the proposed code calls (I delved into library code) `setTintList` on the same `Drawable`. The two functions behave similarly, according to the [documentation](https://developer.android.com/reference/android/graphics/drawable/Drawable#setTintList(android.content.res.ColorStateList)), but there appear to be some differences, and I don't want to risk having non-working code. Also, [this Stack Overflow](https://stackoverflow.com/questions/56716093/setcolorfilter-is-deprecated-on-api29#56717316) thread suggests the differences actually are there.

Issue #2477, opened 2019-07-26
Feature request :  Add a tab in Main Page for the Background Player


gituby:
Hi NewPipe Team,

I wish to make a suggestion for a feature in the main page content. Could it be possible to add in the main content page, a tab for the background player to be able to access it easily (picture 1).

For this, in the parameters menu "Content of Main Page", when clicking on the "+" there could be a new option to add a tab to access directly the background player (picture 2). This will let the user decide to show this tab or not as he wish.

Thanks for reading

![Main Page - Background Player](https://user-images.githubusercontent.com/53324825/61981397-e0846000-aff9-11e9-8cbd-a133d2eae53c.jpg)

![Parameters - Content of Main¨Page](https://user-images.githubusercontent.com/53324825/61981415-eb3ef500-aff9-11e9-9d65-12d1738372cf.jpg)

Stypox:
This is a great idea!

Wilker-a:
i think that an always available button would be easier to navigate, as well as the plus button available in the same way in the playlist screen would be easier to navigate with.

![a mock-up of the button](https://user-images.githubusercontent.com/10854855/62001450-519e4300-b0e0-11e9-80b8-27579d36ff1e.jpg)

Stypox:
@Wilker-a 
What if someone never uses the background player? We could add an option to enable/disable that floating-action-button (FAB), but adding too many options is never good.
Also, a custom background tab is less obstructive, since you can switch back and forth any tab seamlessly.
And adding both the button and the background tab is not great either, since there should only be one way to do something.

Material Design Guidelines suggests to use FABs only when the action they link to is very obvious and frequently used, and I don't think the button you suggest to add meets either of these requirements.

Wilker-a:
i think the solution then should be to make it so the button could be shown only whenever there is a queue or when the media is playing at the background, as these seems to be the only use cases that would most likely happen for the use of the playlist button, and in these cases, it would be frequently used, because the notifications or going away from a video page is not always a reliable way to control media.

the tabs requires the user to go back once and maybe twice if they're browsing a video or searching for one, so i think that makes it a bit more difficult to access without obstructing the other tasks, and i think it would be useful only when the user starts up the app to get a playlist to play with a link for example.

one alternative would be to have the menu panel available at all times with a swipe and put the playlist button there, or put it in the 3 dots menu (together with the downloads and settings buttons and stuff).

[...]

PR #8419, opened 2022-05-19
Add play queue to menu

HybridAU:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
There is a feature (bug?) in Android 11 and up where media controls can be completely dismissed. The problem is the only way to get the the play queue is clicking on the media controls notification.
- Adds "Play Queue" button to the main menu
- Play Queue activity does not finish when the queue is empty. (So if you click on the Play Queue menu item while it's empty it doesn't flash open for a second and then immediately kick you back out)

#### Before/After Screenshots/Screen Record

- Before:
1_before

- After:
2_after

 - Empty play queue screen:
3_empty_queue

#### Fixes the following issue(s)

**EDITED**: See comments below.

This PR might help address #2477,  #5235 and #7208 although not exactly what they are looking for.

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [X] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

Issue #7208, opened 2021-10-07
Return to Play Queue screen from video details


TippyLion28:
### Checklist

- [x] I am using the latest version - x.xx.x 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

Fire up the background audio player and open the Play Queue screen from the notification. Long-press a video and tap "details". Now swipe down the video screen as to minimise the video details. 

### Actual behavior

App returns to the main NewPipe window. 

### Expected behavior

App should return back to the Play Queue. 

### Screenshots/Screen recordings

Can provide a screen recording if necessary. 

### Device info

 - Android version/Custom ROM version: 11, LineageOS 18.1
 - Device model: Poco F1

yashpalgoyal1304:
hi, one issue related to playing queue that i am facing is that the playing queue is not opening when taping the notification. and i dont know any other way to reach that playing queue page. I am facing this on newpipe v0.21.16

is it known? or should i file it? as always, i tried searching for duplicates, but i can never find the related issues in this repo for whatever reasons.

opusforlife2:
@yashpalgoyal1304 It works fine for me. Please check again.

@TippyLion28 It could lead to complications, as currently the main player shows the play queue already. So you would have access to the same play queue using 2 different actions.

TippyLion28:
The video info screen does not show the play queue when the background player is being used.

If I do Background Queue > Long press a video and select "Details" > then press the hardware back button on my device nothing happens. If I press it again it just minimises the video info screen.

What I am proposing is that pressing the hardware back button should send you back to the background play queue

opusforlife2:
Ah. That makes sense. Please update your description to say that.

Issue #5235, opened 2020-12-20
Swipe-up queue


ggdorman:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want
Hi. I would like to see a swipe-up queue implemented into the NewPipe player. I created a few mockups to give a rough idea of what it should look like. It should be accessible with a simple, you guessed it, swipe up gesture. I find it to be really useful—better than accessing it from the notification.

An alternative option would be to use the space below the video player to always show the queue (and option I think would be useful to implement either way). This would behave just like the "comments" and "next and similar videos" fragments.

#### Is your feature request related to a problem? Please describe it

Accessing the playing queue from the notification is a bit cumbersome.

#### Additional context

Mockups:
P.S. Sorry if they are big, they seem to show up full-size in the preview :/
![20201220_174614](https://user-images.githubusercontent.com/76399981/102719213-b74a2b00-42e4-11eb-800f-c216d9a4dc69.jpg)
![20201220_174707](https://user-images.githubusercontent.com/76399981/102719214-b913ee80-42e4-11eb-8422-61b309058f08.jpg)

#### How will you/everyone benefit from this feature?

It will improve the UX of the app, making it easier for people to access the playing queue, especially if they don't know it can only be accessed from the notification.

mhmdanas:
I totally agree that accessing the play queue is more cumbersome than necessary, and I like your idea, but I have one concern. When the video is minimized, will both the minimized player (the one you see at the bottom when you minimize a video while in the app) and the "minimized queue" (I think that's a good name for the feature) be shown? That would be a bit crowded.

vkay94:
>An alternative option would be to use the space below the video player to always show the queue (and option I think would be useful to implement either way). This would behave just like the "comments" and "next and similar videos" fragments.

@ggdorman What do you think about (https://github.com/TeamNewPipe/NewPipe/pull/4534#issuecomment-742448622)?

i5513:
I think queue should be accesible from principal menu, at left, Trends, suscriptions, news, playlists, queue, Download and history. I had to search at github to know that I could access to queue from notifications

Thank you very much for NewPipe !!


T

[...]

PR #8424, opened 2022-05-20
Add play queue button to video details fragment

HybridAU:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

There is a feature (bug?) in Android 11 and up where media controls can be completely dismissed. The problem is the only way to get the the play queue is clicking on the media controls notification.

I originally opened #8419 to add the button to the main menu, but after feedback I've moved it to the video fragment.

- Adds "Play Queue" button to the VideoDetailFragment (small player bar at the bottom of the screen)

- Before:
before

- After:
after

This PR might help address #2477,  #5235 and #7208 although not exactly what they are looking for.

#### APK testing 

#### Due diligence


SameenAhnaf:
It'll surely solve https://github.com/TeamNewPipe/NewPipe/issues/3515 as way to background player controls is not specified. It just asks for any simple method.

TacoTheDank:
From a cursory glance, I still don't think that's the greatest location, but perhaps that's just me.

Anyway, I need to ask: where is it that you're getting the vector drawables from? It doesn't mesh well with the other icons; I don't think it's Google-designed. I'd recommend using Google-designed vector drawables where possible.

HybridAU:
Hey @TacoTheDank thanks for taking a look at this PR :slightly_smiling_face:

I've switched to `ic_list.xml` that looks nicer, I drew the vector files in Inkscape and imported them in Android Studio. I'm curious when you say "doesn't mesh well" do you just aesthetically they are a little off (I won't be offended, I've never considered myself a graphic designer) or was there something technical like they don't render properly because of something I did or didn't do when I was importing them?

I'm not super happy with the icon location, but I'm not sure of a better place. I liked it in the main menu (my original PR) because I mainly use NewPipe as a music player. But I know that's not everyone's use case and space on the main menu is valuable.

Also I've found I introduced a bug, if I start a video playing, go to the play queue and remove the video from the play queue, I can go back to the main screen and the video details fragment is still there. If I click on the play queue icon now it crashes the app. I think I need to hide the video details fragment when the last item is removed from the queue. I'll work out how to do that and then I'll update this PR.

litetex:
So before this is added we should clarify if we even want this.

As far as I can see none of the following issues is somehow fixed:
#2477 - Asks for the playqueue activtiy on the main screen (inside a tab)
#5235 - Asks for a swipe up queue. Similar to YT's swipe up comments.
#7208 - Asks to open the playqueue instead when opening the videodetails?

I think this would only kind of address #3515.

----

Currently I don't think the button inside the mini player is the best solution.
IMHO we would need something like this:
grafik
inside the VideoDetails and not in the mini player.
Some ideas from my side:
* Maybe an additional button here inside the actions?
grafik
* Or maybe adding that a long press of the playqueue button leads to the PlayQueue activtity?
* Or maybe somewhere inside the PlayQueue section of the player?
grafik


Anyway as #3515 has currently 5 upvotes I would postpone this because there is currently more important stuff to be done.

Stypox:
> I've switched to ic_list.xml that looks nicer, I drew the vector files in Inkscape and imported them in Android Studio. I'm curious when you say "doesn't mesh well" do you just mean aesthetically they are a little off (I won't be offended, I've never considered myself a graphic designer) or was there something technical like they don't render properly because of something I did or didn't do when I was importing them?

I don't think he meant anything about the icon looking "bad": it just looks a little bit off with respect to other icons because it is not material-design styled like [all other google icons](https://fonts.google.com/icons?selected=Material+Icons) which follow strict standards.

> I would postpone this because there is currently more important stuff to be done.

I agree: I also don't think this is really a solution and we first need to decide together how to solve the problem at a deeper level.

I am closing this for now, but thank you @HybridAU for your work anyway :-). If you still want to play around with making the queue more accessible, you could try to implement #2477, which personally I think is the more elegant and fitting solution for everyone. But be advised that this is only my opinion and if you work on that your work might be rejected because others disagree. So maybe you could try to summarize the problems people have in the various issues and make a comment explaining the possible ways (with pros and cons) a solution could be achieved. See this comment, for example: https://github.com/TeamNewPipe/NewPipe/pull/4534#issuecomment-741700417

MD77MD:
this is unfortunate...I think #8424  the best solution that solves everyone problem with no compromises...

@HybridAU is it possible to implement your solution to polymorphicshade fork until the developers decide... link below 

https://github.com/polymorphicshade/NewPipe

M1CK431:
I would like to thanks you @HybridAU for your work. You exactly did what I need and wait for more than 2 years now: adding a simple button solving an issue which haven't any workaround at this time. I follow your instruction to get the `apk` but didn't find it. Is there another way to get it? That button is really game changing for me since I don't have any access to notifications...

@litetex @Stypox Ok, you have your opinion about priority and what should be the finale solution, I respect.
However, what it the drawback to merge this excessively simple PR and revert it once the "perfect" solution will be implemented/available? Especially since fixing that long standing issue isn't even planned?

Roy-Orbison:
@HybridAU Since you clearly have the ability to work with the code, would you please consider [the suggestion I made](https://github.com/TeamNewPipe/NewPipe/issues/3515#issuecomment-1063639608)? It seems about the same complexity as your PR, should work on non-touch devices, and would be a decent solution until #2477 or #5235 are implemented.

Stypox:
> However, what it the drawback to merge this excessively simple PR and revert it once the "perfect" solution will be implemented/available? 

The drawback is that that button is not in the correct place and clutters the UI, for example making the title of the video less visible and confusing people who don't use a queue. The button in the drawer was a better solution in my opinion, though still not the best.

M1CK431:
@Stypox Ok there is (a few) less space for video title (but probably still enough to identify the video I guess, since on the screenshot the difference is five letters so not a big deal). Not sure to well understand "confusing people who don't use a queue". Isn't that part of the UI only visible when the user add at least on item in the queue? Also, what is "the drawer"?

MD77MD:
the drawer is the main menue ☰ ... however #8424 is much much better... i hope from stypox to  reconsider

M1CK431:
For other people which, like me, really need an in app access to the background player, here is the debug apk build of this PR: [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/8985558/app-debug.zip)

Hope it helps. I will try to rebase and provide an updated apk after futur releases.

@Stypox I really can't understand your point of view, honestly.

Your first arg is about "clutters the UI" but it was not a problem for existing "play/pause" and "clear" buttons... and now it's a big deal for that new "playlist" button? :roll_eyes: If "clutters the UI" is really the issue, so remove "play/pause" and "clear" buttons and keep only that new "playlist" button, it will maximize the space for the video title while keeping an access to the background player controls (including "play/pause" and "clear"... and many more)!

Your second arg about "confusing people who don't use a queue", sorry it's not possible at all since that part of the UI is only displayed when the user add something in the queue (and because of that, we can be sure it's the right place for that new button, at least far better compare to the drawer).

You stand yourself that the "perfect" solution (aka background player in a tab) isn't even planned due to priority, but please take also into account that some user are already waiting for more than 2 years for an access to the background player from within the app (for various reasons). If not merging that PR, how long will it take until this user need is addressed, even in a basic manner?

So please please please, reconsider this PR as an acceptable compromise, waiting for a better solution :pray: :candle:

Stypox:
@M1CK431 I get it that many people need this, and thank you for providing some thoughtful criticism.

I noticed that this PR always shows the queue button independently of whether the player is open or not, which is not the correct behavior. The queue should be openable only if there actually is one ;-)

After thinking a little bit more about it:
- button in the drawer:
  - pro: fits in a nice place, since the drawer is made exactly for actions that bring the user to other places in the app
  - con: more difficult to implement properly, since the `MainFragment` does not have knowledge of whether the player is open or not, so that would need to be added
- button in the bottom sheet:
  - pro: simpler to access?
  - pro: simpler to implement, since `VideoDetailFragment` already knows whether the player is open or not
  - con: it clutters the UI (I think the play/pause button should stay there, since it is a basic action, while opening the queue from there just doesn't feel right and would be unclear for most users, but this is just my opinion)
- queue tab in the main page (#2477), the final solution:
  - pro: inserts nicely in the current app structure (though we should decide what to do if the player is closed... just show a "There is no player open" message and don't let the user do anything?)
  - pro: allows users to hide it if they don't want it
  - con: takes some time to implement, but most of the code should already be there in `PlayQueueActivity`, which just needs to be turned into a fragment. After #8170 will become a little bit simpler to take care of.
  
@M1CK431 do you have the knowledge and time to try to implement #2477? Otherwise yeah, maybe you are right that in this case it better to just merge a temporary solution. I don't know if "button in the drawer" is implementable properly with the current player structure, but maybe after #8170 will become simpler. So yes, you can go with "button in the bottom sheet" and I will accept the PR, even though I don't like the position of the button too much.

MD77MD:
thank you stypox

M1CK431:
Hello @Stypox, sorry for late reply.

First of all, thanks A LOT for your message :hugs: 

The last time I wrote something in java was a long time ago and it was desktop oriented (a deadly simple exercice for a friend studies). I have absolutely no experience in mobile app development.
Also, despite I'm a curious person who love to learn new things, I will not have enough time to work on this, at least in the next months (not before winter).

Sorry for that, I well know it's not ideal :disappointed: 

I just build the latest master with that PR and here is the APK (inside a zip, for github): [app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/9130368/app-debug.zip)
Perhaps you could give it a try before thinking about this PR merge? I'm a personally using it since weeks now and it does the job very well :star_struck:

Stypox:
Just tried it, it crashes when I follow these steps:
- tap on a video and play it in the main player
- tap on the X in the notification (this will close the player but leave the video detail fragment open)
- minimize the video detail fragment
- press on the new queue button
- NewPipe crashes because there is no player open for which to show a queue

M1CK431:
You are right, I just reproduce it even quicker:
- tap on a video
- scroll down from the video preview
- press the new queue button
- NewPipe crashes

@HybridAU could you please add a check to hide the new queue button when the queue is empty?

HybridAU:
No worries, I'll take a look when I get a chance and update this PR.

Gabr-F:
> I noticed that this PR always shows the queue button independently of whether the player is open or not, which is not the correct behavior. The queue should be openable only if there actually is one ;-)

@Stypox, a queue can be empty (and still exist), every single media player I remember to have ever used had a permanent queue window that was simply empty when nothing was queued!

What is **extremely** weird is the current state of having to jump through hoops, and finding "secret" commands, to open such a basic functionality as a media player's queue!

So while I strongly agree that by far the best, most intuitive solution is the #2477 tab bar, if that requires more work a temporary, less perfect solution is much better than nothing, especially for new NewPipe users (which only learn about the "tapping the notification secret" by asking someone else)

HybridAU:
I like the idea of showing an empty queue more than the idea of hiding the button when the queue is empty. 

I've updated [my branch](https://github.com/HybridAU/NewPipe/tree/add_play_queue_button_to_video_details_fragment) so that if you click on the play queue button when the queue is empty it won't crash and will instead just show the empty play queue page.

I'm not sure if I can re-open this PR or should I just create a new one?

M1CK431:
@HybridAU thanks a lot! I will give it a try asap!
@Stypox I guess you will reopen this PR because we have all the history here?

Stypox:
Yes, you can reopen this PR

MD77MD:
> I like the idea of showing an empty queue more than the idea of hiding the button when the queue is empty.
> 

I agree  with you... this is also  great for consistency ...  but i guess its would be better to try it for week or so then give final verdict

opusforlife2:
@Stypox You closed this PR, so it can't be opened by the author.

But I think a new PR should be opened instead, linking back to this PR.

Stypox:
Oh, I didn't know that. But yeah ok, go with a new PR then.

Issue #8436, opened 2022-05-24
Add loop feature


FrankFacundo:
### Checklist

- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I'm aware that this is a request for NewPipe itself and that requests for adding a new service need to be made at [NewPipeExtractor](https://github.com/TeamNewPipe/NewPipeExtractor/issues).
- [X] I have taken the time to fill in all the required details. I understand that the feature request will be dismissed otherwise.
- [X] This issue contains only one feature request.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Feature description

In player interface, there will be available a button that allow to loop the current video.

### Why do you want this feature?

- People who want to repeat a song.
- People who want to repeat a video playlist.

SameenAhnaf:
Duplicate of https://github.com/TeamNewPipe/NewPipe/issues/6914

mhmdanas:
Note that you can loop by using the loop button on the media notification.

PR #8438, opened 2022-05-26
add loop player feature

FrankFacundo:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Allow loop video 
- https://github.com/TeamNewPipe/NewPipe/issues/8436

- Before:
- After:

- Fixes #

- 

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/res/drawable/ic_replay_off.xml
+++ b/app/src/main/res/drawable/ic_replay_off.xml
@@ -1,0 +1,19 @@
+<vector
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <group android:name="group">
+        <path
+            android:name="path"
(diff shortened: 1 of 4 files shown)


Stypox (changes requested):
As you have been told in the issue you opened:
- you should't have referenced your issue in the "Fixes" section, but the one it is duplicate to
- a way to turn on looping is to click on the loop button in the player notification. So you should have seen there is no need to reinvent the wheel and add a "private boolean loop" field. Just look at how it is being done in `NotificationUtil` and reuse the same method (spoiler: enabling/disabling looping is a feature built-in of `simpleExoPlayer`).

Also, you did not fill the PR template properly.

app/src/main/res/layout/player.xml
```diff
@@ -328,6 +328,21 @@
                         app:tint="@color/white"
                         tools:ignore="RtlHardcoded" />
 
+                    <androidx.appcompat.widget.AppCompatImageButton
```
Stypox, line 331: It should not be put here, but rather in the queue menu

app/src/main/res/values/strings.xml
```diff
@@ -432,6 +432,7 @@
     <string name="name">Name</string>
     <string name="add_to_playlist">Add to playlist</string>
     <string name="processing_may_take_a_moment">Processing… May take a moment</string>
+    <string name="loop">Loop</string>
```
Stypox, line 435: This shouldn't be needed, as there is already a "Repeat" string somewhere

app/src/main/res/drawable/ic_replay_off.xml
```diff
@@ -0,0 +1,19 @@
+<vector
```
Stypox, line 1: This also isn't needed, use exoplayer's icons

Stypox (changes requested):

Stypox:
Wait, no, I'm closing this PR. A button to enable looping already exists in the queue menu. It is not accessible when there is just one song, but this should not be solved with a duplicate button.

PR #8456, opened 2022-05-29
Add database test for SubscriptionManager

### LINKED ISSUE #2758: Only one emply playlist is shown
Steps to reproduce:

1. Select item, `Add To Playlist` -> `New Playlist` -> `test1`
2. Go to `test1` playlist, `Delete` item. (`test1` playlist will be shown with 0 tracks)
3. Select item, `Add To Playlist` -> `New Playlist` -> `test2`
4. Go to `test2` playlist, `Delete` item. -> `test1` playlist become hidden!
5. Delete `test2` playlist -> `test1` playlist become visible again!

### LINKED ISSUE #8080: Add more tests for methods involving the database
### Checklist

### Affected version

Every version, but in particular this issue was opened after #2335

### The problem

We should add more tests for methods that query/use the database, after mocking the NewPipeExtractor. These kind of tests would not be too difficult to do (android tests would, instead, I think), but would anyway allow us to ensure things work correctly faster. For example we could start by adding a test for the [`SubscriptionManager.rememberAllStreams()`](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionManager.kt#L126) method introduced in #2335, see https://github.com/TeamNewPipe/NewPipe/pull/2335#discussion_r764894903.

SydneyDrone:
#### What is it?

- A JUnit test file is added to test SubscriptionManager.kt (see #2335). However, the test result shows that either the original methods or my understanding of those methods has some problem, so I made this a draft.
- The first problem is that the `ChannelInfo` I insert doesn't equal to the one I searched. The test result shows that it's because of their difference in the uid. IDK if it is designed to operate like that. If it is, I will delete the corresponding test cases in later commit.
- The second problem is that the `updateNotificationMode()` method doesn't seems to function correctly. The `testUpdateNotificationMode()` method 2 and 4 seems to be problematic and redundant, but I will temporarily keep them until this problem is solved or clairfied.
- The third problem is that the `rememberAllStreams()` method doesn't seems to fetch all videos as it should be (in my test case, it just fetched two instead of the total number of 100+)

- The total test result:
image

- [x] Fixes #8080
- [ ] Related to #2758

#### Due diligence

### DIFF
--- a/app/src/androidTest/java/org/schabi/newpipe/local/subscription/SubscriptionManagerTest.java
+++ b/app/src/androidTest/java/org/schabi/newpipe/local/subscription/SubscriptionManagerTest.java
@@ -1,0 +1,144 @@
+package org.schabi.newpipe.local.subscription;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+
+import androidx.test.core.app.ApplicationProvider;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.schabi.newpipe.database.AppDatabase;
+import org.schabi.newpipe.database.subscription.SubscriptionEntity;
+import org.schabi.newpipe.extractor.channel.ChannelInfo;
+import org.schabi.newpipe.extractor.exceptions.ExtractionException;
+import org.schabi.newpipe.extractor.stream.StreamInfo;
+import org.schabi.newpipe.testUtil.TestDatabase;
+import org.schabi.newpipe.testUtil.TrampolineSchedulerRule;
+
+import java.io.IOException;
+import java.lang.reflect.Method;
+
+public class SubscriptionManagerTest {
+    private AppDatabase database;
+    private SubscriptionManager manager;
+    private SubscriptionEntity subscription;
+    private SubscriptionEntity anotherSubscription;
+
+    @Rule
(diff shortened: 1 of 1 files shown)


opusforlife2:
Are you adding all possible tests that involve the database? If not, then I don't think you should link this PR to that issue. This would then be one of several PRs that work towards completion of the objective stated there.

opusforlife2:
@SydneyDrone Please try to provide a reason when you close a PR. It helps when we're looking at it in the future.

Stypox:
@SydneyDrone thank you for writing these tests, I think they are valid. Could you reopen?

SydneyDrone:
> @SydneyDrone thank you for writing these tests, I think they are valid. Could you reopen?

It's reopened now. Sorry for the inconvenience.

Stypox:
I made some changes:
- fix checkstyle errors
- tests do not run in order, so each one has to do its own assertions separately from what others did
- the uid of an entity in the database needn't be the same of the one created in-memory, since the uid gets assigned upon inserting in the database
- some database functions return a `Completable` that doesn't do anything until it is subscribed to or awaited, so I added `.awaitBlocking()` where needed
- the data of an entity in-memory does not get updated automatically when the corresponding entity in the database is changed, so some tests have been removed
- `manager.insertSubscription` only inserts recent streams, so they need to have a date set on them (I also made related items hardcoded and not dependent on what the channel is currently doing)

Now the tests pass and I think they are reasonable. Tell me what you think ;-)

TobiGr (approved):

PR #8468, opened 2022-06-01
Clean up proguard file

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

I went down the git blame hole and found that PrettyTime was first added using 4.0.1. Meanwhile, consumer proguard rules were [added to PrettyTime in 4.0.3](https://github.com/ocpsoft/prettytime/commit/602b3a9bb6efbc9f7032fb7307f4866227aa4135). Therefore we no longer need this rule (and also ours is now [outdated](https://github.com/ocpsoft/prettytime/blob/master/core/src/main/resources/META-INF/proguard/prettytime.pro) anyway).

```-keep class org.ocpsoft.prettytime.i18n.** { *; }```

These are from before AndroidX. I don't think I need to explain these :)

```
-dontwarn android.arch.util.paging.CountedDataSource
-dontwarn android.arch.persistence.room.paging.LimitOffsetDataSource
```

The OkHttp rules are embedded into the library as of [3.11.0](https://square.github.io/okhttp//changelogs/changelog_3x/#version-3110:~:text=New%3A%20Embed%20R8/ProGuard%20rules%20in%20the%20jar). OkHttp was [first added using 3.9.1](https://github.com/TeamNewPipe/NewPipe/commit/76e082159dbbf5b7bc14fc26ad6030bffed7484b). Therefore, these specific ones are now obsolete and safe to remove.

```
-dontwarn javax.annotation.**
-keepnames class okhttp3.internal.publicsuffix.PublicSuffixDatabase
```

#### APK testing 

#### Due diligence

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -18,7 +18,6 @@
 
 -dontobfuscate
 -keep class org.schabi.newpipe.extractor.timeago.patterns.** { *; }
--keep class org.ocpsoft.prettytime.i18n.** { *; }
 
 -keep class org.mozilla.javascript.** { *; }
 
@@ -26,9 +25,6 @@
 -keep class com.google.android.exoplayer2.** { *; }
 
 -dontwarn org.mozilla.javascript.tools.**
--dontwarn android.arch.util.paging.CountedDataSource
--dontwarn android.arch.persistence.room.paging.LimitOffsetDataSource
-
 
 # Rules for icepick. Copy paste from https://github.com/frankiesardo/icepick
 -dontwarn icepick.**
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/proguard-rules.pro [hunk 1]
```diff
@@ -39,12 +35,11 @@
 }
 -keepnames class * { @icepick.State *;}
 
-# Rules for OkHttp. Copy paste from https://github.com/square/okhttp
+## Rules for OkHttp. Copy paste from https://github.com/square/okhttp
```
mhmdanas, line 38: What's the idea behind the double hash?

Stypox:
@TacoTheDank in order to test these changes, the only thing I have to do is checking that the app builds fine? Or should I also check things at runtime?

app/proguard-rules.pro
litetex: I think it would be better if you just leave the ``#`` and a a line above 
```
-keepclassmembers class * implements java.io.Serializable
```

that says something like ``# keep required serialized stuff``

TacoTheDank: I wasn't intending to add any note to that, I just wanted to section off the OKHttp stuff from the rest of the file. I guess I could add a note for that though, if you want.

Stypox (approved):
Actually, there does seem to be a difference, 57576 vs 57561 methods

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

But as it can be seen in the images, it's just a difference in "prettytime", and as you said our rule was outdated anyway. This saves ~1KB :-D

Anyway, then this looks good, since everything builds normally and the app runs normally (prettytime works).

PR #8153, opened 2022-04-05
Support delivery methods other than progressive HTTP

AudricV:
<sub>Previous too-long title: Add support of other delivery methods than progressive HTTP (in the player only), use DASH to fetch YouTube progressive contents, fix extraction of PeerTube streams with separate audio stream and more</sub>

#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)

#### Description of the changes in your PR
This PR superseeds #6537 and adds so support of other delivery methods of contents than progressive HTTP, in the player only. It also:

  - fixes seeking of PeerTube HLS streams, by using streams' HLS manifests to play HLS streams;
  - fixes the lack of some streams on some YouTube videos (OTF streams);
  - improves a lot loading times of YouTube streams, after a quality change or a playback start (progressive streams are now using DASH to fetch contents, and the current way is used as a fallback if something goes wrong when generating the DASH manifest associated to this stream);
  - fixes the extraction of YouTube ended livestreams as livestreams (this type of streams is not interpreted anymore as them but as a separate type instead, look at the extractor PR for more details), which also:
    - fixes watch count interpreted as watching count;
    - fixes the watchable time of these contents (full content instead of the last 2 or 4 hours, depending of the length returned in the HLS manifest).
  - fixes the crash when trying to play PeerTube videos with a separate audio stream;
  - improves the quality selection for external players dialog, by adding a title to it and a message when no streams are available for external players (see `Before/After Screenshots/Screen Record` section);
  - fixes playback of SoundCloud HLS-only tracks (which cannot be downloaded anymore, as the workaround I introduced is being removed by SoundCloud and has been already removed in the extractor with TeamNewPipe/NewPipeExtractor#859).

<details>
<summary>Detailed changes</summary>

## External players:

  - Add a message instruction about stream selection;
  - Add a message when there is no stream available for external players;
  - Return now HLS, DASH and SmoothStreaming URL contents, in addition to progressive HTTP ones.

## Player:

  - Support DASH, HLS and SmoothStreaming streams for videos, whether they are content URLs or the manifests themselves, in addition to progressive HTTP ones;
  - Use a custom `HttpDataSource` to play YouTube contents, named `YoutubeHttpDataSource`, based of ExoPlayer's default one, which allows better spoofing of official clients (custom user-agent and headers (depending of the client used), use of `range` and `rn` (set dynamically by the `DataSource`) parameters);
[...]

PR #8469, opened 2022-06-01
Specify used ExoPlayer libraries

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

I don't know how this hadn't caught my eye before, considering I've made this exact change in other apps. Anyway, I digress.

Before, when just using `exoplayer`, *every* ExoPlayer library would be added (regardless of whether they were used). This would cause some unnecessary bloat. Now after being specific, the unused libraries (like rtsp) are not being added.

This shaves off around 103KB for debug builds and 49KB for release builds.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -238,7 +238,13 @@ dependencies {
     implementation "com.squareup.okhttp3:okhttp:3.12.13"
 
     // Media player
-    implementation "com.google.android.exoplayer:exoplayer:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-core:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-dash:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-database:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-datasource:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-hls:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-smoothstreaming:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-ui:${exoPlayerVersion}"
     implementation "com.google.android.exoplayer:extension-mediasession:${exoPlayerVersion}"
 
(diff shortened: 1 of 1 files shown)


TacoTheDank:
> or maybe directly switch to [androix media3](https://developer.android.com/jetpack/androidx/releases/media3)? :)

I'd probably wait until it's stable :)

app/build.gradle
```diff
@@ -238,7 +238,13 @@ dependencies {
     implementation "com.squareup.okhttp3:okhttp:3.12.13"
 
     // Media player
-    implementation "com.google.android.exoplayer:exoplayer:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-core:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-dash:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-database:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-datasource:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-hls:${exoPlayerVersion}"
+    implementation "com.google.android.exoplayer:exoplayer-smoothstreaming:${exoPlayerVersion}"
```
litetex, line 246: Do we really use smoothstreaming inside NewPipe? 
AFAIK we only use DASH, HLS and progressive HTTP but I could be wrong...

TiA4f8R: SmoothStreaming code is currently used in `PlaybackResolver` and `PlayerDataSource`, but after #8153, more code related to SmoothStreaming (support of URI and non-URI contents of this technology) will be added (even if no content with this adaptive streaming technology is currently extracted by the extractor).

litetex: @TiA4f8R
Do we have anything inside the extractor that uses smooth streaming in the first place?

TiA4f8R: @litetex No.

Stypox: Then I would remove support for SS in NewPipe until it actually becomes needed. It makes no sense to keep untested and unused code into the repo, which also increases app size by a little.

TacoTheDank: That can be done in another PR if so.

PR #8477, opened 2022-06-03
make video title & controls more compact

jaromeyer:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

- Increase default video title to 2 lines: I have found that almost every video title fits in two lines, but not in one (which is the current default). Also everywhere else in the app (and in the official YT app iirc) video titles currently can take up two lines, so this will unify the UI throughout the app.
- merge primary and secondary controls: This decouples showing additional options from expanding the title and thus makes the UI more intuitive in my opinion. I have tested it on various screen sizes and even on very small screens it works fine, so I currently don't see any downsides of always showing all controls.
- only show the expansion arrow when the title is ellipsized (longer than 2 lines)

<table>
  <tr>
    <th>Before (default)</th>
    <th>Before (expanded)</th>
    <th>After</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
    <td></td>
  </tr>
</table>

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -579,19 +579,21 @@ public final class VideoDetailFragment
     }
 
     private void toggleTitleAndSecondaryControls() {
-        if (binding.detailSecondaryControlPanel.getVisibility() == View.GONE) {
-            binding.detailVideoTitleView.setMaxLines(10);
-            animateRotation(binding.detailToggleSecondaryControlsView,
-                    Player.DEFAULT_CONTROLS_DURATION, 180);
-            binding.detailSecondaryControlPanel.setVisibility(View.VISIBLE);
-        } else {
-            binding.detailVideoTitleView.setMaxLines(1);
-            animateRotation(binding.detailToggleSecondaryControlsView,
-                    Player.DEFAULT_CONTROLS_DURATION, 0);
-            binding.detailSecondaryControlPanel.setVisibility(View.GONE);
+        // only toggle if title is ellipsized
+        final int lines = binding.detailVideoTitleView.getLineCount();
+        if (binding.detailVideoTitleView.getLayout().getEllipsisCount(lines - 1) > 0) {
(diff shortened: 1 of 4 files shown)


SameenAhnaf:
Keeping the `Share` and `Open in` buttons in the middle instead of the end would make the experience a little more symmetrical.

peat80:
Since the control icons are more or less self explaining, why not remove the labels and just show the icons? Would look even better i think. 🤔 😀

Stypox:
I agree the buttons should be more compact, since on most devices there is a lot of wasted space in between buttons. But on some devices 4 buttons barely fit, adding 3 more just makes it worse there (remember, there is also Kodi, which is hidden unless enabled)... And the buttons are also more difficult to tap.

I tried a while ago to make that row of buttons horizontally scrollable, so that if they were too large, they would become scrollable, but was unable to do it. Maybe you could try?

Btw, take a look at how I edited your PR description. Now images are of the correct size ;-)

> Since the control icons are more or less self explaining, why not remove the labels and just show the icons? Would look even better i think.

@peat80 well, I also agree it would look even better. But I don't know about discoverability... Once we have an "introductory course" the first time a user opens NewPipe, then yeah, we can remove labels completely, but otherwise... I don't know. Many people to which I share NewPipe do not realize what the background button does even if currently there is a label. I think it would only be worse if we removed the label. I would put this to votes

opusforlife2:
Please try to do only one thing per PR, not just for easier reviewing, but also so that discussions on separate things happen separately.

- Labels should not be removed at all. These icons aren't universal. As Stypox says, without the labels, our somewhat cryptic UI would become even less usable by 'the normies'. That's practically everyone I've shared this app with.
- The 7 buttons in a row would be horrifying to use on a 4.7 inch screen. Hard no.
- I fully agree with the title changes, both the 2 line thing and the ellipsis for longer titles.

This PR should only focus on the video title changes, I think.

Stypox:
Closing as there is still no consensus, because of the problems already highlighted, and because of inactivity. If you still wish to work on this feel free to try to reach consensus and then reopen @jaromeyer :-)

PR #8484, opened 2022-06-05
Use DisplayCompat.getMode().

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use [`DisplayCompat.getMode()`](https://developer.android.com/reference/androidx/core/view/DisplayCompat#getMode(android.content.Context,android.view.Display)) to get the current display mode. This method provides workarounds for retrieving the display size for Android TV devices.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -10,7 +10,6 @@ import android.content.SharedPreferences;
 import android.content.pm.ActivityInfo;
 import android.database.ContentObserver;
 import android.graphics.Color;
-import android.graphics.Point;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
@@ -22,6 +21,7 @@ import android.provider.Settings;
 import android.util.DisplayMetrics;
 import android.util.Log;
 import android.util.TypedValue;
+import android.view.Display;
 import android.view.LayoutInflater;
 import android.view.MotionEvent;
 import android.view.View;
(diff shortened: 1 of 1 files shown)


Isira-Seneviratne:
> Please, provide a "why" to your changes when opening PRs. I have no idea what `DisplayCompat` does and why this change would be an improvement, I only know it increases code size.

Sorry about that, I amended the change description.

Stypox:
Thank you

> This method provides workarounds for retrieving the display size for Android TV devices.

Didn't it already work before, or is there an issue this PR fixes? What is the difference in behaviour?

Isira-Seneviratne:
> Thank you
> 
> > This method provides workarounds for retrieving the display size for Android TV devices.
> 
> Didn't it already work before, or is there an issue this PR fixes? What is the difference in behaviour?

The documentation mentions that some 4K TVs might have trouble running the framework code, so a workaround is provided in the method.

Stypox:
According to the Android docs, [Display.getSize()](https://developer.android.com/reference/android/view/Display#getSize(android.graphics.Point)) does something different than [DisplayCompat.getMode().getPhysicalHeight()](https://developer.android.com/reference/androidx/core/view/DisplayCompat.ModeCompat#getPhysicalHeight())/[display.getMode().getPhysicalHeight()](https://developer.android.com/reference/android/view/Display.Mode#getPhysicalHeight()). The first one gets the size of the screen used by the app (so removing system parts), while `getPhysicalHeight()` just returns the full screen size.

Isira-Seneviratne:
> According to the Android docs, [Display.getSize()](https://developer.android.com/reference/android/view/Display#getSize(android.graphics.Point)) does something different than [DisplayCompat.getMode().getPhysicalHeight()](https://developer.android.com/reference/androidx/core/view/DisplayCompat.ModeCompat#getPhysicalHeight())/[display.getMode().getPhysicalHeight()](https://developer.android.com/reference/android/view/Display.Mode#getPhysicalHeight()). The first one gets the size of the screen used by the app (so removing system parts), while `getPhysicalHeight()` just returns the full screen size.

You're right, my bad. The backport implementation calls `Display#getRealSize()` on API levels 17-22 and `Display#getSize()` below API 17.

The documentation recommends using `WindowManager#getCurrentWìndowMetrics()` instead of `Display#getSize()`. I'll update this PR.

Isira-Seneviratne:
I created a separate PR for the change I mentioned here: https://github.com/TeamNewPipe/NewPipe/pull/8656

PR #7518, opened 2021-12-08
Remove icon duplicates and fix some theming issues

mauriciocolli:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

Some icons are not consistently themed (it affects the other pull request #7516):

<table><tr>
  <td>Drawing</td>
  <td>Drawing</td>
</tr></table>

---

- Remove unnecessary duplicates of some icons, as it is a bad practice that leads to mistakes. As vectors are used, color resources are available.
  - [The documentation recommends the `?colorControlNormal` attribute.](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#themes-styles) It's the default value when creating a new vector drawable.
  I chose the night-qualified resources route to stay consistent with current behavior. Do share your thoughts on this.
  - Take a look at the "live tv" icon, [it was left at the default value (`?colorControlNormal`)](https://github.com/TeamNewPipe/NewPipe/blob/ad405d9e0b63ac96949953f9fca0c125ee35f3c4/app/src/main/res/drawable/ic_live_tv.xml#L2), so it's already using the attribute and its color is different than the rest.
- Use menu resources instead of manually creating it, let the system do its job and get the correct resources.
  - Some menu items were not themed properly because the toolbar itself had a custom theme overriding some values.

---

Now only some drawables are different in behavior than the rest of the `ic_` prefixed icons:
- The heart icon.
- Headphone with shadow.
- Play with shadow.

Do you think they should be renamed to something else to make it consistent?

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #5928, opened 2021-03-27
Replace UniversalImageLoader with Picasso

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
This PR replaces the outdated [Android-Universal-Image-Loader](https://github.com/nostra13/Android-Universal-Image-Loader) (which has not had any new release in the last 6 years) with the more stable and updated [Picasso](https://github.com/square/picasso). This hopefully fixes a [nasty unfixed issue](https://github.com/nostra13/Android-Universal-Image-Loader/issues/908) in Android-Universal-Image-Loader that was causing #4638. Also, the loading seems smoother to me, and retrieving images from memory cache is now instant, instead of taking a couple of milliseconds (try scrolling fast up and down after all thumbnails have been loaded).
I also added a switch in the *debug* settings to enable [indicators](https://square.github.io/picasso/static/debug.png).

#### Fixes the following issue(s)
Fixes #4638

#### APK testing 
@gulachev @k3a @Sekitan @UserX404 @nonremittal @arielshulman @dboydor @pauseee @astatine (taken from #4638)
Could you test [this APK from the CI](https://github.com/TeamNewPipe/NewPipe/suites/2358978908/artifacts/49936658) and see if it fixes the issue? Thank you ;-)

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #8530, opened 2022-06-21
Improve image placeholders

krlvm:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)

Currently, placeholders are shown instead of avatars or video previews only in case of an error during loading, including a timeout. This is not a good solution for a slow or unstable internet connection, or if the government blocks the domain with user avatars (`yt3.ggpht.com`) - the application waits for a very long time to load and in the end it will display the placeholder, but all this time on this place is empty, which makes it difficult to read comments and use the application as a whole.

Also, the placeholders didn't display very nicely on high resolution screens, so I replaced them with vector ones based on the SVG originals from the assets directory. I adapted them to the light theme, as they don't look very consistent and don't fit very well.

- Show placeholders until the image is loaded because timeout can be very long and missing profile pictures and video thumbnails make app inconvenient to use
- Adapt profile picture and video thumbnail placeholders to light theme
- Replace profile picture and video thumbnail placeholders with vector graphics

| Before  | After |
| before  | after |
| before2 |  after2 |

#### APK testing 

#### Due diligence


Stypox:
What about using [this material icon](https://fonts.google.com/icons?selected=Material+Icons&icon.query=person) instead of creating a new SVG icon? This way we surely don't have styling problems, we don't have to maintain one more icon and we have a more standard look.

krlvm:
I thought using of custom icon was intended, this look better indeed

Stypox (changes requested):
- Is the file `dummy_thumbnail_dark.png` still used? Remove it otherwise
- I would rename `buddy.xml` to `dummy_person.xml`
- I think you have colors swapped for the video thumbnail: image
- Even though you created different colors in `values-night` and then you used them in all placeholders, it doesn't seem to have worked, as the same image is shown regardless of the theme (light, dark or black)

krlvm:
Can't reproduce the last one, it works like #7518 does

`dummy_thumbnail_dark.png` is still used [here](https://github.com/TeamNewPipe/NewPipe/pull/8530/files#diff-b2e369e39e24aea82f68bcb17acd1135173d776a871986ac8ef85ad73e767333R2351), but I've resized to 1x1px because it is just filled with black color.

Stypox:
> Can't reproduce the last one, it works like https://github.com/TeamNewPipe/NewPipe/pull/7518 does

These are the same colors, I tested with a color picker, but the theme is "black" in the first image and "light" in the second.
image
image
I tried to tackle this, but unfortunately it seems to be a problem in Picasso, see square/picasso#1275. I therefore removed the differences between themes for now, leaving the colors as they have always been in NewPipe.

-----

Also, the play arrow's left line does not seem to be completely straight:
image
I pushed a commit improving the svg paths (also for the playlist), tell me if they look good to you ;-)

-----

> dummy_thumbnail_dark.png is still used [here](https://github.com/TeamNewPipe/NewPipe/pull/8530/files#diff-b2e369e39e24aea82f68bcb17acd1135173d776a871986ac8ef85ad73e767333R2351), but I've resized to 1x1px because it is just filled with black color.

I would remove `dummy_thumbnail_dark.png` completely and set the overlay thumbnail to a `null` image when nothing is to be shown. I pushed a commit for this, too, tell me if it looks good to you.

-----

I also renamed all of the placeholder images (also e.g. the channel banner) to `placeholder_*` in another commit.

krlvm:
LGTM, night mode colors worked for me though, at least on Android 10+.
I think there should be a light version of `placeholder_person` - the dark version is too contrasting with the background.

Stypox:
> LGTM, night mode colors worked for me though, at least on Android 10+.

Yes, they do follow the device's night/day theme. But if you set the theme manually in NewPipe's settings, then it doesn't work anymore.

Stypox:
Mmmh, I think the background and foreground colors should be the same across all placeholders. For example, I think the first image below better fits the app style when thumbnails are disabled.
image
image

Maybe we should just increase the darkness of the foreground color for each placeholder; would something like this do, in your opinion?
image

Btw, I rebased and resized SVGs to 16:9 so that with the new `scaleType:fitCenter` they appear at the correct size in thumbnails.

Stypox (approved):
Ok, then I am good with these changes. Thank you for the PR and the feedback :-D

> Unfortunately, proper drawable resource can't be selected too due to Picasso bug, so I think we will have to use the same placeholders for both themes.

Maybe by looking more into it it is possible to find a solution, but I don't think it is so important as of now. Though, if you want to give it a try in a new PR, it would be accepted.

Stypox:
> Currently, placeholders are shown instead of avatars or video previews only in case of an error during loading, including a timeout. This is not a good solution ...

For reference, this behavior was initially requested in #5928

opusforlife2:
You mean this https://github.com/TeamNewPipe/NewPipe/pull/5928#issuecomment-895666105?

Stypox:
Yes

PR #8531, opened 2022-06-21
Updated extractor to latest revision and fix compile errors

litetex:
#### What is it?

* Updates the extractor to the latest revision
* Fixes overlooked compile errors from https://github.com/TeamNewPipe/NewPipeExtractor/pull/846

#### APK testing 

- https://github.com/TeamNewPipe/NewPipe/pull/8153

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ServiceHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ServiceHelper.java
@@ -1,9 +1,12 @@
 package org.schabi.newpipe.util;
 
+import static org.schabi.newpipe.extractor.ServiceList.SoundCloud;
+
 import android.content.Context;
 import android.content.SharedPreferences;
 
 import androidx.annotation.DrawableRes;
+import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
 import androidx.preference.PreferenceManager;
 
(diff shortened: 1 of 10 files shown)


litetex:
Note: Because the changes for #8153 got already merged into the extractor but not into NewPipe itself there are currently build errors.
Once #8153 is merged this PR can be rebased and merged.

app/src/main/java/org/schabi/newpipe/util/ServiceHelper.java
```diff
@@ -113,18 +117,32 @@ public static int getImportInstructionsHint(final int serviceId) {
...
 
-        return serviceId;
+    @Nonnull
+    public static String getNameOfServiceById(final int serviceId) {
+        return ServiceList.all().stream()
+                .filter(s -> s.getServiceId() == serviceId)
+                .findFirst()
+                .map(StreamingService::getServiceInfo)
+                .map(StreamingService.ServiceInfo::getName)
+                .orElse("<unknown>");
```
Stypox, line 145: String resource?

litetex: Wouldn't recommend:
1. This should never happen in the first place (however it's possible because we use ``int``s instead of valid objects)
2. We are inside a static method we would need some kind of context here (and in all subsequent calls)
3. I don't think we have to translate this
4. It was a hardcoded string before

Stypox (approved):
LGTM

PR #8542, opened 2022-06-24
Add 'Share playlist' option to Playlist fragment

### LINKED ISSUE #1788: Share playlist or export it to text or clipboard
Please add ability to Export playlists in readable text just in case Newpipe becomes blocked by YouTube and the internal playlists are inaccessible.

carmebar:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Added ability to share a playlist using a list of video URLs

- Before:
image
- After:
image

- Fixes #1788

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
@@ -1,5 +1,6 @@
 package org.schabi.newpipe.local.playlist;
 
+import static org.schabi.newpipe.error.ErrorUtil.showUiErrorSnackbar;
 import static org.schabi.newpipe.ktx.ViewUtils.animate;
 import static org.schabi.newpipe.util.ThemeHelper.shouldUseGridLayout;
 
@@ -50,6 +51,7 @@ import org.schabi.newpipe.util.Localization;
 import org.schabi.newpipe.util.NavigationHelper;
 import org.schabi.newpipe.util.OnClickGesture;
 import org.schabi.newpipe.info_list.dialog.StreamDialogDefaultEntry;
+import org.schabi.newpipe.util.external_communication.ShareUtils;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


SameenAhnaf:
https://github.com/TeamNewPipe/NewPipe/issues/1788#issuecomment-484678883 could have been used instead. YouTube allows anonymous playlists for upto 50 videos.

Example playlist URL: http://www.youtube.com/watch_videos?video_ids=5aTwJEW_Se4,7zplexrqshk,i9ZzrGYEhhw

carmebar:
@SameenAhnaf I choose a link per video to allow more flexible usage and easier implementation. If there's some interest I might add an option in the settings panel to choose what format to share the playlist. If there are more than 50 videos in the playlist I could generate multiple links.

SameenAhnaf:
Ok, thanks. I actually need to turn on this toggle in content settings as there's a limit for characters in text messages.

Toggle: `Allow Anonymous YouTube Playlist Sharing` 
Sub-note: `This option will only work only if 50 or less YouTube videos are about to be shared`

Stypox (approved):
I rebased and pushed a commit myself, since there were various problems:
- There was no need to use a `isSharingPlaylist` boolean to prevent the user from tapping the button multiple times in a row. Such a check would be needed if the button modified the playlist, in which case pressing multiple times would lead to possible problems, but in this case if the user wants to share twice (in a row, before waiting for the previous share to be done), let them do whatever they want.
- The list elements (loading, error panel, ...) were used to show loading/errors for the sharing process. This is not a good idea, since those elements have to do with the list itself, not the actions you perform on it, otherwise strange situations would occur. Also, there is no need to show a loading indicator for the share button, it probably only takes a few microseconds to complete.
- A new "Share playlist" string was added, but I think the already-translated "Share" works better and is more standard.
- The share button was being hidden in the three-dot menu, but it is better if it is shown, when there is room.

Thank you anyway @carmebar, and please tell me if these changes look good to you ;-)

carmebar:
Looks great. Thanks for taking the time to write such informative feedback. Much appreciated.

nlof:
This seems to generate a list of youtube links. I didn't find a way to add them all at once to a youtube playlist. Am I missing something?

carmebar:
@nlof this feature is intended to share all the videos from a playlist outside NewPipe in a simple list. To import videos into a Youtube playlist you can follow these steps: https://webapps.stackexchange.com/a/106340

nlof:
@camebar Thanks for that link. It's pretty cumbersome though, I don't like getting that techy, it's too time consuming and I don't know linux. There was however a google spreadsheet  which might work... hopefully :)

PR #8336, opened 2022-05-02
Fixed viewed counting

Mamadou78130:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

When watching a video that you have not seen before and going to the history.
Note that the history entry shows 2 views instead of 1 view for said video.

I then modified a function to solve the problem.
So when we go to the video history we see that he has 1 marked view.

#### Before/After Screenshots/Screen Record

- Before:
Capture d’écran du 2022-05-02 15-25-46

- After:
Capture d’écran du 2022-05-02 15-24-11

#### Fixes the following issue(s)

- Fixes https://github.com/TeamNewPipe/NewPipe/issues/8330

##### Test Manuel
- Watch a video you've never seen
- Go down in history
- And you see the video shows 1 view

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #8564, opened 2022-07-01
Actually fix history view count

### LINKED ISSUE #8330: NewPipe client counts 1 view as 2 views
### Checklist

### Affected version

0.23.0

1. Go view a video you haven't seen before
2. Go to history
3. Notice the history entry shows 2 views instead of 1 view for said video

### Expected behavior

To count 1 view every time I watch something

### Actual behavior

Counts 2 views every time I watch something

_No response_

### Logs

_No response_

Android 12

### Affected device model

Samsung Galaxy Tab S6 Lite

### Additional information

_No response_

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

#8336 did not really fix the problem, it just made so that each item in history has 1 less view. This meant that, even though the bug appeared to be fixed the first time a video is opened (since that video's views are correctly set to 1), the second time the same video is opened the views are incorrectly set to 3. And #8336 actually introduced a regression, since e.g. enqueued videos were registered as "No views" as reported in https://github.com/TeamNewPipe/NewPipe/pull/8336#issuecomment-1149998538.

The underlying problem was an incorrect usage of Java's `==` (*same*) operator: it was used in the player's `onEvents` function to check if the media item tag in the **ExoPlayer** was equal to the one currently set in our `Player`. The problem with this is, when the player is started from scratch, the media item tag is first updated with a tag with no `LoadedMediaSource` attached (since the player is still loading the media source), and then updated again when the media source was loaded. The second update created a new instance of the media item tag (see `StreamInfoTag.withExtras()` and its usage in `LoadedMediaSource` and `FailedMediaSource`), so the `==` operator is not suitable for comparing them. When the player is not started from scratch, but rather just switches to the next item in the queue, only 1 view was (correctly) registered, since the media source is usually already loaded because of the mechanism of pre-fetching the next item in the queue. 

This PR solves the issue by comparing the stream info url associated with the media item tag before calling `updateMetadataWith()` with the new stream info.

This PR also makes it so that marking a video as watched does not add one view to it, and also does not update its last access time. Now a new history entry **with 0 views** is added, and this happens only if one did not exist before. Tell me if I should revert this.

- Replaces #8336
- Really fixes #8330

#### APK testing 
[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/9029513/app-debug.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -2485,22 +2485,31 @@ public final class Player implements
         Listener.super.onEvents(player, events);
         MediaItemTag.from(player.getCurrentMediaItem()).ifPresent(tag -> {
             if (tag == currentMetadata) {
-                return;
+                return; // we still have the same metadata, no need to do anything
             }
+            final StreamInfo previousInfo = Optional.ofNullable(currentMetadata)
+                    .flatMap(MediaItemTag::getMaybeStreamInfo).orElse(null);
             currentMetadata = tag;
-            if (!tag.getErrors().isEmpty()) {
+
+            if (!currentMetadata.getErrors().isEmpty()) {
+                // new errors might have been added even if previousInfo == tag.getMaybeStreamInfo()
                 final ErrorInfo errorInfo = new ErrorInfo(
-                        tag.getErrors().get(0),
+                        currentMetadata.getErrors(),
                         UserAction.PLAY_STREAM,
-                        "Loading failed for [" + tag.getTitle() + "]: " + tag.getStreamUrl(),
-                        tag.getServiceId());
+                        "Loading failed for [" + currentMetadata.getTitle()
+                                + "]: " + currentMetadata.getStreamUrl(),
+                        currentMetadata.getServiceId());
                 ErrorUtil.createNotification(context, errorInfo);
             }
-            tag.getMaybeStreamInfo().ifPresent(info -> {
+
+            currentMetadata.getMaybeStreamInfo().ifPresent(info -> {
                 if (DEBUG) {
                     Log.d(TAG, "ExoPlayer - onEvents() update stream info: " + info.getName());
                 }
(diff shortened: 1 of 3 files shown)


litetex (approved):
LGTM

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2485,22 +2485,31 @@ public void onEvents(@NonNull final com.google.android.exoplayer2.Player player,
         Listener.super.onEvents(player, events);
         MediaItemTag.from(player.getCurrentMediaItem()).ifPresent(tag -> {
             if (tag == currentMetadata) {
-                return;
+                return; // we still have the same metadata, no need to do anything
             }
+            final StreamInfo previousInfo = Optional.ofNullable(currentMetadata)
+                    .flatMap(MediaItemTag::getMaybeStreamInfo).orElse(null);
```
AudricV, line 2491: You can use `getCurrentStreamInfo()` method to do so, unless you want to avoid confusion between the previous info and the current one :)

```suggestion
            final StreamInfo previousInfo = getCurrentStreamInfo().orElse(null);
```

<br>

<details>

<summary>Reminder of getCurrentStreamInfo() method</summary>

```java
    private Optional<StreamInfo> getCurrentStreamInfo() {
        return Optional.ofNullable(currentMetadata).flatMap(MediaItemTag::getMaybeStreamInfo);
    }
```

</details>

AudricV (commented):
I didn't really reviewed the whole code but found an optimization in it that you may use.

I trust @litetex for the rest of the code :)

app/src/main/java/org/schabi/newpipe/player/Player.java
Stypox: Even tough to my eye your suggestion looks better, I'd prefer to keep the current code as it shows where the stream info comes from

PR #8601, opened 2022-07-10
Checkstyle assign on same line

litetex:
#### What is it?

See https://github.com/TeamNewPipe/NewPipe/pull/8170#discussion_r917278123.

After some discussion with @Stypox  a while ago (can't find the issue/PR anymore :/ ) where to put the ``=`` on a new line or not we agreed to put it not on the next line.

This PR was create to enforce this behavior.

This PR also removes some unused code that was found on the fly.

After the PR is merged the changes should also be applied to https://github.com/TeamNewPipe/NewPipeExtractor

#### Due diligence


app/src/main/java/org/schabi/newpipe/util/urlfinder/PatternsCompat.java
```diff
@@ -18,159 +18,13 @@
 
 package org.schabi.newpipe.util.urlfinder;
```
Stypox, line 19: I don't think you can just remove things from this file at random. This file was taken from AOSP and modified, as the licence explains (though a comment would be better), so any change should be documented I guess.

Stypox (changes requested):
LGTM, except for the unused code removal. Thanks ;-)

app/src/main/java/org/schabi/newpipe/player/MainPlayer.java
```diff
@@ -65,24 +65,25 @@ public enum PlayerType {
...
+    static final String ACTION_PLAY_PREVIOUS =
+            BASE_ACTION + "ACTION_PLAY_PREVIOUS";
+    static final String ACTION_FAST_REWIND =
+            BASE_ACTION + "ACTION_FAST_REWIND";
+    static final String ACTION_FAST_FORWARD =
+            BASE_ACTION + "ACTION_FAST_FORWARD";
+    static final String ACTION_SHUFFLE =
+            BASE_ACTION + "ACTION_SHUFFLE";
+    public static final String ACTION_RECREATE_NOTIFICATION =
+            BASE_ACTION + "ACTION_RECREATE_NOTIFICATION";
```
TacoTheDank, line 86: You could probably just remove the newlines for most of these 🤔

Stypox: I think they are this way for alignment purposes... I'd leave them this way

TacoTheDank: ¯\_(ツ)_/¯ alrighty

Stypox (approved):
Thanks :-)

PR #8603, opened 2022-07-10
Add quality selector for livestreams, fix some seamless transition issues and more

### LINKED ISSUE #1239: Quality selector in lives

AudricV:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)

Replacement of #7260, rebased with some changes in the feature.

This pull request adds the quality selector for livestreams. This quality selector, due to limitations, will only include qualities' resolution (or its bitrate if its height is unknown and as a last resort, the `Quality unknown` string).

Contrary to the previous attempt, there is no `Auto` quality, in order to prevent confusion of this availability on videos.

The default resolution is set according to the user choices in Video and audio settings.

However, resolutions are now set using minimum and maximum video size constraints instead of selection overrides, in order to simply the code. Available resolutions are got from the first video `TrackGroupInfo` found and rebuilt each time the `onTracksChanged(TracksInfo)` event of ExoPlayer is triggered.

This PR also refactors how `SourceType`s are used by the player. Instead of being returned only by `VideoPlaybackResolver`, it is now included in `MediaItemTag`s (and the corresponding enum has been extracted from `VideoPlaybackResolver`). This allow detection of sources which require building a quality selector using the first video `TrackGroupInfo` found (like I said above), setting an `AnalyticsListener` to update the quality `TextView` depending of the quality played, and setting the default resolution using `TrackSelectionParameters`.

The `SourceType` enum has been changed to contain more values, including a one used for livestreams: `MANIFEST`. As the name suggests, it represents `MediaSource`s generated from a "master" manifest URL. This will allow us in the future to use, if needed, a manifest to play a non-live content with the ability to choose the video quality played (even if a few changes in player and resolver codes will be needed).

This enum is now also used to do the seamless transition, instead of the last resolved source, and disabling video and text tracks used for the transition is now always done (so even when the play queue manager is reloaded), in order to fix some blank video issues.

Finally, this PR also fixes playability of contents which doesn't have an audio stream and only video streams when trying to play them background (by using `Start playing in the background` or `Enqueue` commands). The default video stream will be played, with potentially text tracks if they are enabled. This is the case on several PeerTube videos.

For more detailed changes, please take a look at the code changes.

- Before:

Quality selection livestreams before

- After:

Quality selection livestreams after

- Fixes #1239

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -2289,11 +2289,11 @@ public final class Player implements PlaybackListener, Listener, AnalyticsListen
 
     private boolean isLive() {
         try {
-            return !exoPlayerIsNull() && simpleExoPlayer.isCurrentMediaItemDynamic();
+            return !exoPlayerIsNull() && simpleExoPlayer.isCurrentMediaItemLive();
         } catch (final IndexOutOfBoundsException e) {
             // Why would this even happen =(... but lets log it anyway, better safe than sorry
             if (DEBUG) {
-                Log.d(TAG, "player.isCurrentWindowDynamic() failed: ", e);
+                Log.d(TAG, "player.isCurrentMediaItemLive() failed: ", e);
             }
             return false;
         }
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -43,6 +43,7 @@ import static org.schabi.newpipe.player.notification.NotificationConstants.ACTIO
 import static org.schabi.newpipe.player.notification.NotificationConstants.ACTION_RECREATE_NOTIFICATION;
 import static org.schabi.newpipe.player.notification.NotificationConstants.ACTION_REPEAT;
 import static org.schabi.newpipe.player.notification.NotificationConstants.ACTION_SHUFFLE;
+import static org.schabi.newpipe.util.ListHelper.computeDefaultResolution;
 import static org.schabi.newpipe.util.ListHelper.getPopupResolutionIndex;
 import static org.schabi.newpipe.util.ListHelper.getResolutionIndex;
 import static org.schabi.newpipe.util.Localization.assureCorrectAppLanguage;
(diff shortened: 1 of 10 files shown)


AudricV:
Note: as you can see in the last commit, I have an issue when a track switch occurs: the `onTracksInfoChanged` event is fired before the track switch, as the quality selection is changed or the available tracks change, so the reset is cancelled and all the process made when a livestream is initialized is done again.

I don't know how to fix this issue, so I gladly ask for help on it.

opusforlife2:
My dude, I'm no dev, but this looks like 2 (maybe even 3) PRs rolled into one. Are your changes so tightly integrated that they cannot be split into individual PRs?

You could mark one PR as draft if it is dependent on another being merged.

Stypox:
I support opusforlife2 on this: seeing the title already made me question why you made one big PR instead of two or more

AudricV:
> Are your changes so tightly integrated that they cannot be split into individual PRs?

> why you made one big PR instead of two or more

Yes, most of changes are linked between them and are required, except the background playback fix for contents without an audio stream.

opusforlife2:
> except the background playback fix for contents without an audio stream.

Then separate this out into another PR?

opusforlife2:
All the more reason to separate out the background fix so that it isn't held back by this PR! 💯

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -218,6 +223,8 @@ public final class Player implements PlaybackListener, Listener {
     private boolean isPrepared = false;
     private boolean wasPlaying = false;
 
+    private boolean defaultResolutionSetForManifestsSources = false;
```
Stypox, line 226: What is this variable? Why are you setting it to `false` only when Next and Previous are called? Since `setDefaultResolutionForManifestSourcesIfNeeded` does not depend on anything, shouldn't it be called only when `trackSelector` is initialized, i.e. in the `Player` constructor?

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -1281,7 +1289,8 @@ public void onTracksChanged(@NonNull final Tracks tracks) {
             Log.d(TAG, "ExoPlayer - onTracksChanged(), "
                     + "track group size = " + tracks.getGroups().size());
         }
-        UIs.call(playerUi -> playerUi.onTextTracksChanged(tracks));
+        setDefaultResolutionForManifestSourcesIfNeeded();
```
Stypox, line 1292: Why do you need to call it again here? It shouldn't make any changes.

Stypox (changes requested):
Looks mostly good, thank you! I tested a little and everything seems to work fine.

AudricV:
I am closing this PR, as the current player structure doesn't allow us to implement properly this feature.

It we rely directly on ExoPlayer track selection APIs to build quality menus when rewriting the player, we will be able to implement this in a proper way.

For instance, LibreTube uses these APIs for quality selection and has support of this feature.

For other fixes and improvements in this PR, a part of them (or all of them?) have been already done in other PRs and the remaining parts, if any, could be done if needed in the future.

PR #8621, opened 2022-07-13
Deduplicate SQL queries to get feed streams

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

While merging #8545 I noticed there was much duplicated code in the feed loading. There were 4 different SQL queries based on the wanted filtering (all the combination of "filter or not by group" and "include or not played items"). Furthermore #8545 included another condition, but it wasn't implemented in SQL (as it would have meant adding 4 more queries) but rather in code. This PR removes duplicate queries by making the now-only-one query take as input the various filtering conditions. I tested and noticed no changes in the UI behavior.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/database/feed/dao/FeedDAO.kt
+++ b/app/src/main/java/org/schabi/newpipe/database/feed/dao/FeedDAO.kt
@@ -9,6 +9,7 @@ import androidx.room.Update
 import io.reactivex.rxjava3.core.Flowable
 import io.reactivex.rxjava3.core.Maybe
 import org.schabi.newpipe.database.feed.model.FeedEntity
+import org.schabi.newpipe.database.feed.model.FeedGroupEntity
 import org.schabi.newpipe.database.feed.model.FeedLastUpdatedEntity
 import org.schabi.newpipe.database.stream.StreamWithState
 import org.schabi.newpipe.database.stream.model.StreamStateEntity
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/database/feed/dao/FeedDAO.kt
```diff
@@ -122,24 +47,37 @@ abstract class FeedDAO {
         INNER JOIN feed f
         ON s.uid = f.stream_id
 
-        INNER JOIN feed_group_subscription_join fgs
+        LEFT JOIN feed_group_subscription_join fgs
         ON fgs.subscription_id = f.subscription_id
 
-        WHERE fgs.group_id = :groupId
+        WHERE (
+            :groupId == ${FeedGroupEntity.GROUP_ALL_ID}
```
litetex, line 54: Is there a reason why there are two ``=``?
Is this a typo?

litetex (commented):
Looks good to me. Did a quick test, found no bigger problems.

Only one minor thing:

app/src/main/java/org/schabi/newpipe/database/feed/dao/FeedDAO.kt
Stypox: Uuups, I mixed java with sql (but in theory it should work)

Stypox:
@litetex done ;-)

litetex (approved):

PR #8656, opened 2022-07-19
Use WindowMetrics API.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use Android R's `WindowMetrics` API in `VideoDetailFragment` and `PopupPlayerUi`, as `Display#getSize()` and `Display#getMetrics()` have been deprecated in favor of it.

- Fixes #

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -21,7 +21,6 @@ import android.content.SharedPreferences;
 import android.content.pm.ActivityInfo;
 import android.database.ContentObserver;
 import android.graphics.Color;
-import android.graphics.Point;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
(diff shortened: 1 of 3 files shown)


Stypox:
> I tried using the Android API, but faced some SonarQube issues

Oh, which was the issue?

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1078,6 +1077,20 @@ public void updateTabLayoutVisibility() {
         }
     }
 
+    private int getWindowHeight(@NonNull final WindowManager windowManager) {
```
Stypox, line 1080: Put this method in `util/DeviceUtils`

Stypox (changes requested):
I tested and it works fine. Thank you for implementing this without libraries. The Sonar warnings are normal, because the usage of deprecated methods was not removed, but that's normal in this case, as the deprecated methods are only used on older devices.

Stypox (approved):
Works, thanks

Issue #8646, opened 2022-07-17
Tablet layout encountered after full screen video ends, despite Tablet mode being off


opusforlife2:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.23.1

### Steps to reproduce the bug

1. Lock system rotation in portrait mode.
2. Turn off Tablet mode from Newpipe settings.
3. Play a video in full screen and let it finish.

### Expected behavior

The app should fully respect the Tablet mode setting, and go into the phone version of landscape mode after exiting full screen. (Not optimal, but at least it should match the currently expected behaviour.)

### Actual behavior

After the video ends, the app exits full screen and goes into the tablet version of the landscape layout, instead of the phone version. Thankfully, the "Tapping the Back button" workaround to go back to portrait mode still works on this tablet version, like it does on the phone version.

I've encountered this behaviour on a phone with Smallest width at 514 dp. This bug does not occur on a different phone which is at 449 dp. Both phones have Display size set to "Smaller" from Display settings. The Smallest width developer setting wasn't directly touched. Changing a user-facing option shouldn't cause this sort of bug.

### Additional information

On the ~500+ dp phone, turn on system auto-rotate. Open a video details fragment. Rapidly tilt the phone so you keep quickly shifting between landscape and portrait modes. You will have a ~50-50 chance of seeing either the phone version or the tablet version of the landscape layout each time the app rotates.

Ensure Autoplay is off. Turn on the "Start player in full screen" setting. Tap on a video thumbnail to start the video. The app will just rotate to the tablet version of landscape. The video doesn't play until you tap the thumbnail. The 449 dp phone correctly goes into full screen landscape and the video starts automatically.

---

[...]

PR #8661, opened 2022-07-19
Fix random NullPointerException when adding video player view

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Followup to #8170, bug reported by @opusforlife2. I don't know why the `binding` could be `null` when `getView()` returns a non-null value, unless the view is destroyed exactly in between `new Handler().post()` being called and the passed function being actually called, but anyway... 

#### Fixes the following *comments*
- https://github.com/TeamNewPipe/NewPipe/pull/8170#issuecomment-1184735382 https://github.com/TeamNewPipe/NewPipe/pull/8170#issuecomment-1185762001
- https://github.com/TeamNewPipe/NewPipe/issues/8646#issuecomment-1186609499

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -1313,9 +1313,11 @@ public final class VideoDetailFragment
         // Prevent from re-adding a view multiple times
         new Handler(Looper.getMainLooper()).post(() ->
                 player.UIs().get(MainPlayerUi.class).ifPresent(playerUi -> {
-                    playerUi.removeViewFromParent();
-                    binding.playerPlaceholder.addView(playerUi.getBinding().getRoot());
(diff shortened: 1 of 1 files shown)


opusforlife2:
I discovered something stupid :D

I tried the player-refactor APK again just to compare it with this PR. The bug didn't occur!

And then some testing led me to realise that the crash occurs _only when_ the full screen button is tapped while the device is moving! So if the device stays still on a surface or you keep your hand still while holding it, the full screen button works properly. But if the accelerometer registers movement, there is a crash.

This bug is so strange that I think it must be CalyxOS-specific.

Anyway, this PR fixes this weird bug.

Stypox:
Merging as it is such a simple change and opus said it is now fixed for him.

> only when the full screen button is tapped while the device is moving

It kinda makes sense, since the fixed code has to do with adding the video player view (also) after layout changes, but strange nonetheless ;-)

PR #8668, opened 2022-07-21
Show toast when no updates are available.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Display a toast if no new versions of NewPipe are available when checking for updates.

- Before:
- After:

- Fixes #

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
+++ b/app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
@@ -3,16 +3,18 @@ package org.schabi.newpipe
 import android.content.Context
 import android.content.Intent
 import android.util.Log
+import android.widget.Toast
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
+import androidx.core.content.ContextCompat
 import androidx.core.content.edit
 import androidx.core.net.toUri
 import androidx.preference.PreferenceManager
-import androidx.work.OneTimeWorkRequest
+import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.WorkManager
-import androidx.work.WorkRequest
 import androidx.work.Worker
 import androidx.work.WorkerParameters
+import androidx.work.workDataOf
 import com.grack.nanojson.JsonParser
 import com.grack.nanojson.JsonParserException
 import org.schabi.newpipe.extractor.downloader.Response
@@ -42,26 +44,34 @@ class NewVersionWorker(
         versionCode: Int
     ) {
         if (BuildConfig.VERSION_CODE >= versionCode) {
+            if (inputData.getBoolean(IS_MANUAL, false)) {
+                // Show toast stating that the app is up-to-date if the update check was manual.
+                ContextCompat.getMainExecutor(applicationContext).execute {
+                    Toast.makeText(applicationContext, R.string.app_update_unavailable_toast,
+                        Toast.LENGTH_SHORT).show()
+                }
+            }
             return
         }
-        val app = App.getApp()
 
         // A pending intent to open the apk location url in the browser.
         val intent = Intent(Intent.ACTION_VIEW, apkLocationUrl?.toUri())
         intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
-        val pendingIntent = PendingIntentCompat.getActivity(app, 0, intent, 0)
-        val channelId = app.getString(R.string.app_update_notification_channel_id)
-        val notificationBuilder = NotificationCompat.Builder(app, channelId)
+        val pendingIntent = PendingIntentCompat.getActivity(
+            applicationContext, 0, intent, 0
+        )
+        val channelId = applicationContext.getString(R.string.app_update_notification_channel_id)
+        val notificationBuilder = NotificationCompat.Builder(applicationContext, channelId)
             .setSmallIcon(R.drawable.ic_newpipe_update)
             .setVisibility(NotificationCompat.VISIBILITY_PUBLIC)
-            .setContentIntent(pendingIntent)
             .setAutoCancel(true)
-            .setContentTitle(app.getString(R.string.app_update_notification_content_title))
-            .setContentText(
-                app.getString(R.string.app_update_notification_content_text) +
-                    " " + versionName
-            )
-        val notificationManager = NotificationManagerCompat.from(app)
+            .setContentIntent(pendingIntent)
+            .setContentTitle(applicationContext.getString(
+                    R.string.app_update_available_notification_title))
+            .setContentText(applicationContext.getString(
+                    R.string.app_update_available_notification_text, versionName))
+
+        val notificationManager = NotificationManagerCompat.from(applicationContext)
         notificationManager.notify(2000, notificationBuilder.build())
     }
 
@@ -72,12 +82,14 @@ class NewVersionWorker(
             return
         }
 
-        val prefs = PreferenceManager.getDefaultSharedPreferences(applicationContext)
-        // Check if the last request has happened a certain time ago
-        // to reduce the number of API requests.
-        val expiry = prefs.getLong(applicationContext.getString(R.string.update_expiry_key), 0)
-        if (!isLastUpdateCheckExpired(expiry)) {
-            return
+        if (!inputData.getBoolean(IS_MANUAL, false)) {
+            val prefs = PreferenceManager.getDefaultSharedPreferences(applicationContext)
+            // Check if the last request has happened a certain time ago
+            // to reduce the number of API requests.
+            val expiry = prefs.getLong(applicationContext.getString(R.string.update_expiry_key), 0)
+            if (!isLastUpdateCheckExpired(expiry)) {
+                return
+            }
         }
 
         // Make a network request to get latest NewPipe data.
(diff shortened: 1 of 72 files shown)


TacoTheDank:
Wouldn't this show every time the app is launched? That would be pretty annoying, I think.

mhmdanas:
I agree with @TacoTheDank, it makes little sense to me to show a notification if there is _no_ update available, plus it would be quite irritating. What problem are you trying to solve by this PR @Isira-Seneviratne?

Isira-Seneviratne:
> I agree with @TacoTheDank, it makes little sense to me to show a notification if there is _no_ update available, plus it would be quite irritating. What problem are you trying to solve by this PR @Isira-Seneviratne?

The idea was to show some feedback if no update was available, since at the moment nothing is shown if the app is up-to-date.

mhmdanas:
I think that would be more suited for a manual update check perhaps.

TobiGr:
I would not show a notification, but rather a toast. A toast saying "Checking for updates..." is shown when the "check for updates" button is pressed (settings > updates).
I agree with the others that this info should not be shown when the check was performed automatically.

Isira-Seneviratne:
@TobiGr Yeah, I'll make that change.

app/src/main/java/org/schabi/newpipe/settings/UpdateSettingsFragment.java
```diff
@@ -32,7 +32,7 @@ private void checkNewVersionNow() {
         // Reset the expire time. This is necessary to check for an update immediately.
         defaultPreferences.edit()
                 .putLong(getString(R.string.update_expiry_key), 0).apply();
-        NewVersionWorker.enqueueNewVersionCheckingWork(getContext());
+        NewVersionWorker.enqueueNewVersionCheckingWork(requireContext(), true);
```
Stypox, line 35: I think you should use the MANUAL flag inside `NewVersionWorker.checkNewVersion` to determine whether to ignore the expiry time, instead of resetting the expiry time here. Also, since only one statement would be left here then, I would remove `checkNewVersionNow()` and inline the work enqueueing.

app/src/main/res/values/strings.xml
```diff
@@ -551,8 +551,9 @@
     <string name="low_quality_smaller">Low quality (smaller)</string>
     <string name="dont_show">Don\'t show</string>
     <!-- App update notification -->
-    <string name="app_update_notification_content_title">NewPipe update is available!</string>
-    <string name="app_update_notification_content_text">Tap to download</string>
+    <string name="app_update_unavailable_toast">You are running the latest version of NewPipe.</string>
+    <string name="app_update_available_notification_title">NewPipe update is available!</string>
+    <string name="app_update_available_notification_text">Tap to download %s</string>
```
Stypox, line 556: Since you added `%s` to this string, I think you should remove all of the copies in other languages so that they are retranslated with the `%s` in the correct position.

Stypox (changes requested):
Looks almost good to me

app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
```diff
@@ -120,43 +126,49 @@ class NewVersionWorker(
...
-            val workRequest: WorkRequest =
-                OneTimeWorkRequest.Builder(NewVersionWorker::class.java).build()
+        fun enqueueNewVersionCheckingWork(context: Context, isManual: Boolean) {
+            // Search for updates immediately when update checks are enabled. Reset the expire time
+            // if the check was manual. This is necessary to check for an update immediately.
+            if (isManual) {
+                PreferenceManager.getDefaultSharedPreferences(context).edit {
+                    putLong(context.getString(R.string.update_expiry_key), 0)
+                }
+            }
```
Stypox, line 168: This is not what I meant: remove this and just add an `if (!manual)` around [these lines](https://github.com/TeamNewPipe/NewPipe/blob/6ca5b4108ba98274ab7602bbbb1ebee03439147a/app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt#L81-L87)

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
Isira-Seneviratne: Is the change okay?

app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
```diff
@@ -42,26 +44,30 @@ class NewVersionWorker(
         versionCode: Int
     ) {
         if (BuildConfig.VERSION_CODE >= versionCode) {
+            if (inputData.getBoolean(IS_MANUAL, false)) {
+                // Show toast stating that the app is up-to-date if the update check was manual.
+                Toast.makeText(applicationContext, R.string.app_update_unavailable_toast, Toast.LENGTH_SHORT).show()
```
Stypox, line 49: No lines longer than 100 characters... ktlint seems to be misconfigured, it should detect these problems
```suggestion
                Toast.makeText(applicationContext, R.string.app_update_unavailable_toast,
                        Toast.LENGTH_SHORT).show()
```

app/src/main/java/org/schabi/newpipe/NewVersionWorker.kt
```diff
@@ -42,26 +44,30 @@ class NewVersionWorker(
...
             .setAutoCancel(true)
-            .setContentTitle(app.getString(R.string.app_update_notification_content_title))
-            .setContentText(
-                app.getString(R.string.app_update_notification_content_text) +
-                    " " + versionName
-            )
-        val notificationManager = NotificationManagerCompat.from(app)
+            .setContentIntent(pendingIntent)
+            .setContentTitle(applicationContext.getString(R.string.app_update_available_notification_title))
+            .setContentText(applicationContext.getString(R.string.app_update_available_notification_text, versionName))
```
Stypox, line 68:
```suggestion
            .setContentTitle(applicationContext.getString(
                    R.string.app_update_available_notification_title))
            .setContentText(applicationContext.getString(
                    R.string.app_update_available_notification_text, versionName))
```

app/src/main/java/org/schabi/newpipe/settings/UpdateSettingsFragment.java
```diff
@@ -16,25 +16,17 @@ public class UpdateSettingsFragment extends BasePreferenceFragment {
                 .apply();
 
         if (checkForUpdates) {
-            checkNewVersionNow();
+            NewVersionWorker.enqueueNewVersionCheckingWork(requireContext(), true);
```
Stypox, line 19: I would make this `false` intead of `true`, since the user did not directly ask for an update check, he just enabled them. So if the last check was done a while ago (or never), then it's ok to check again, but if it has been done minutes earlier since the user just disabled and enabled twice in a row, it should not be done. Also, it's better to consider it automatic since we don't want a toast to be shown to the user in this case, as he does not expect feedback I think.

app/src/main/res/values/strings.xml
```diff
@@ -551,8 +551,9 @@
     <string name="low_quality_smaller">Low quality (smaller)</string>
     <string name="dont_show">Don\'t show</string>
     <!-- App update notification -->
-    <string name="app_update_notification_content_title">NewPipe update is available!</string>
-    <string name="app_update_notification_content_text">Tap to download</string>
+    <string name="app_update_unavailable_toast">You are running the latest version of NewPipe.</string>
```
Stypox, line 554:
```suggestion
    <string name="app_update_unavailable_toast">You are running the latest version of NewPipe</string>
```

Stypox (changes requested):
I get the following error

```
java.lang.NullPointerException: Can't toast on a thread that has not called Looper.prepare()
        at ...
        at org.schabi.newpipe.NewVersionWorker.compareAppVersionAndShowNotification(NewVersionWorker.kt:49)

Isira-Seneviratne:
> I get the following error, probably the wrong thread or context are used?
> 
> ```
> java.lang.NullPointerException: Can't toast on a thread that has not called Looper.prepare()
>         at ...
>         at org.schabi.newpipe.NewVersionWorker.compareAppVersionAndShowNotification(NewVersionWorker.kt:49)

Yeah, it seems like the fix is to use a `Handler` or `Executor` to show the toast.

By the way, how can I check for updates in the debug build?

TacoTheDank:
> By the way, how can I check for updates in the debug build?

I would assume that debug builds are for debugging, and that update checking isn't needed for them.

Isira-Seneviratne:
> > By the way, how can I check for updates in the debug build?
> 
> I would assume that debug builds are for debugging, and that update checking isn't needed for them.
> 

Right, my bad. I'll check with a release build.

app/src/main/java/org/schabi/newpipe/settings/UpdateSettingsFragment.java
TobiGr: I'd keep this `true` because the user enabled the check manually, in other words wants NewPipe to check for new versions. Reporting back, that we did not find any new version *this time* is ok. It shows that the app does something as requested.

TobiGr (approved):
LGTM. Thank you

TobiGr:
btw. a rebase is required. I just pulled the latest translations from weblate. 

To the maintainer merging this PR: please make sure to sync weblate once this PR is merged

Stypox (approved):
Thanks!

PR #8669, opened 2022-07-21
Remove uses of setBottomSheetCallback().

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Remove calls to the deprecated `setBottomSheetCallback()` method.

- Fixes #

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -192,6 +192,7 @@ public final class VideoDetailFragment
     private Disposable positionSubscriber = null;
 
     private BottomSheetBehavior<FrameLayout> bottomSheetBehavior;
+    private BottomSheetBehavior.BottomSheetCallback bottomSheetCallback;
     private BroadcastReceiver broadcastReceiver;
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -207,8 +208,6 @@ public final class VideoDetailFragment
     private PlayerService playerService;
     private Player player;
     private final PlayerHolder playerHolder = PlayerHolder.getInstance();
-
-    private BottomSheetBehavior.BottomSheetCallback bottomSheetCallback;
 
     /*//////////////////////////////////////////////////////////////////////////
     // Service management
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -208,6 +208,8 @@
     private Player player;
     private final PlayerHolder playerHolder = PlayerHolder.getInstance();
 
+    private BottomSheetBehavior.BottomSheetCallback bottomSheetCallback;
```
Stypox, line 211: Mmmh, I would put this field near `setBottomSheetCallback`, I think it makes more sense

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -386,7 +388,7 @@ public void onDestroy() {
         disposables.clear();
         positionSubscriber = null;
         currentWorker = null;
-        bottomSheetBehavior.setBottomSheetCallback(null);
+        bottomSheetBehavior.removeBottomSheetCallback(bottomSheetCallback);
```
Stypox, line 391: Are you sure nothing bad happens if `bottomSheetCallback` is `null` or it has already been removed?

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
Isira-Seneviratne: Sure, I'll make that change.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
Isira-Seneviratne: The callbacks are stored in an `ArrayList`, and the method calls the list's remove method.

Just in case, I'll check.

Stypox (approved):

PR #8677, opened 2022-07-22
Keep strong references to Picasso notification icon loading targets

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Before the Target would sometimes be garbage collected before being called with the loaded channel icon, since Picasso holds weak references to targets. This meant that sometimes a new streams notification would not be shown, because the lambda that should have shown it had already been garbage collected.

I don't know if there was a way to properly reproduce something wrong with the previous behavior, but Picasso clearly states in `RequestCreator#into(Target)`:

> This method keeps a weak reference to the `Target` instance and will be garbage collected if you do not keep a strong reference to it.

Anyway, I tested the new code by subscribing to the usual Roel Van De Paar and it works.

Note that this is also one of the causes of a missing thumbnail in the notification, that will be fixed with #8678.

None that I could find

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/feed/notifications/NotificationHelper.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/feed/notifications/NotificationHelper.kt
@@ -4,6 +4,8 @@ import android.app.NotificationManager
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
+import android.graphics.Bitmap
+import android.graphics.drawable.Drawable
 import android.net.Uri
 import android.os.Build
 import android.provider.Settings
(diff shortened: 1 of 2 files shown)


litetex (approved):
Code LGTM

Is it somehow possible to test this?

Stypox:
Yes, just subscribe to Roel Van De Paar and enable its notifications, wait for a few minutes (until he uploads a new video, or find a channel which uploads more frequently), then manually trigger the streams notification update from debug settings.

litetex (approved):
LGTM, quick test also successful

Issue #8616, opened 2022-07-12
Player-related codethat should be improved


litetex:
I and Stypox noticed some strange things while reviewing/refactoring the player https://github.com/TeamNewPipe/NewPipe/pull/8170 (which were also there before the refactoring):

1. When the playerservice is removed the complete app is killed  (not gracefully terminated) by calling ``Runtime.getRuntime().halt(0)``: https://github.com/TeamNewPipe/NewPipe/pull/8170#discussion_r865168043
2. https://github.com/TeamNewPipe/NewPipe/pull/8170#discussion_r915709223 - here somehow a velocity is set to a position.
3. https://github.com/TeamNewPipe/NewPipe/blob/d9af7885140560a6373c1a8863aa381cdd8ed170/app/src/main/java/org/schabi/newpipe/player/Player.java#L2584 should be addressed (or removed) https://github.com/TeamNewPipe/NewPipe/pull/8170#discussion_r917277853

J-Stutzmann:
I think I have a decent explanation for what item 2 is/does:

- For a given `velocityX` and `velocityY` of a popup:
- if the **Magnitude** of either is greater than `TOSS_FLING_VELOCITY`:
- for each that is greater, set the axial position of the popup (`playerUi.popupLayoutParams.*`) to the rounded **Value** of its co-axial velocity (`velocity*`)
- clamp popup position into `popupLayoutParams` width/height

What this does, for a well chosen `TOSS_FLING_VELOCITY` that:
- As a velocity is high, but attainable
- As a position lies outside the `popupLayoutParams` width/height when either negative or positive

is to snap a flung popup to the edge of what `popupLayoutParams` says is the available space in the (approximate) direction it was sent.

PR #8678, opened 2022-07-22
Create media session UI and fix player notification

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This PR solves one of the points in #8616: it introduces `MediaSessionPlayerUi`, moving more code out of the `Player`, and fixes various serious issues related to the player notification and the media session. I tried to put as much info as possible in commit messages and comments, so that it is simpler to debug and git blame.
- Some useless classes were removed
  - `MediaSessionCallback` and `PlayerMediaSession` was just a wrapper for some player functions (half of which used in one place, half in another), but it didn't make any sense not to use the player instance directly bca9543a54b04035a33227d65d6bf6be8dcc1672
  - All of `MediaSessionManager` was moved in `MediaSessionPlayerUi` 49be75e6a00240480c1b40c596007c35dd472f5b
- Make sure to trigger a metadata update after the thumbnail finishes loading, otherwise it ~~could happen~~ almost always happened that the new thumbnail would never be shown. 80df363d11a612826dd4e3e9f71fa00ee62d6548
- Once a new video info starts loading in the player, the thumbnail of the previous video info is unset, so that any metadata update would just get (correctly) a `null` thumbnail, until the new thumbnail loading finishes. The placeholder thumbnail is not used anymore, as it makes more sense to just not set a thumbnail. 25f40a63c92c086ba3de2fe7596f8d86a3a7a65f
- Since the builder used to create player notifications is reused, now the thumbnail is set to `null` when the user disables all thumbnails from settings while the notification is open (this requires a notification update to kick in), otherwise the lastly set thumbnail would be incorrectly shown. c9c55b10803cc7662ae2619e6b226b35f4adaaf7
- Let exoplayer decide when to update metadata, instead of deciding ourselves. This seems to fix the wrong seekbar properties (duration and current position) being shown in the notification sometimes. `MediaSessionConnector.setMetadataDeduplicationEnabled(true)` was used to keep the previous behavior of not updating metadata if nothing changed from the already-set metadata. Now the number of updates is a little bit higher (tested by setting "noisy non-suspending breakpoints"), but there is no random update while playing or with other actions that do not change the current stream info, which would drain battery. E.g. 7 metadata updates instead of 2 when the player is started, and 3 instead of 2 when a new stream info plays. (740ab8af0cd2a8a57f800923d37ebf39f09ef8f8) and most importantly d0677bbb8c8b1ed3ab057817a56dbec47fb63489
- Keep strong reference to Picasso thumbnail loading target. Before the Target would sometimes be garbage collected before being called with the loaded thumbnail, since Picasso holds weak references to targets. Also see #8677 for a similar change. e9e9e8096ca6838a259250ed0f83d3c986dc64b2

You can test changes to the media session by installing [this neat media session testing app](https://github.com/googlesamples/android-media-controller/). Some of the `Test`s provided there are unsuccessful, but I don't know why since the actions they test are actually performed correctly (I included this sentence just so you know, as fixing the tests, which are related to player actions, is out of the scope of this PR).

This is related to #7166, so I'd like @litetex to review.

I am not providing screenshots of all of the various issues there were before, since they take a while to reproduce and if you have used NewPipe for a while you know about them by now.

Strangely enough I could not find any... Maybe somebody of you has a clue?

#### APK testing 

#### Due diligence


app/src/main/java/org/schabi/newpipe/player/mediasession/MediaSessionPlayerUi.java
```diff
@@ -0,0 +1,140 @@
...
+            mediaSession.setActive(false);
+            mediaSession.release();
+            mediaSession = null;
+        }
+    }
+
+    @Override
+    public void onBroadcastReceived(final Intent intent) {
+        super.onBroadcastReceived(intent);
+        // TODO decide whether to handle ACTION_HEADSET_PLUG or not
```
litetex, line 70: TODO detected

app/src/main/java/org/schabi/newpipe/player/mediasession/MediaSessionPlayerUi.java
```diff
@@ -0,0 +1,140 @@
...
+
+        // set album art, unless the user asked not to, or there is no thumbnail available
+        final boolean showThumbnail = player.getPrefs().getBoolean(
+                context.getString(R.string.show_thumbnail_key), true);
+        Optional.ofNullable(player.getThumbnail())
+                .filter(bitmap -> showThumbnail)
+                .ifPresent(bitmap -> {
+                    builder.putBitmap(MediaMetadataCompat.METADATA_KEY_ALBUM_ART, bitmap);
+                    builder.putBitmap(MediaMetadataCompat.METADATA_KEY_DISPLAY_ICON, bitmap);
+                });
```
litetex, line 136: Wouldn't something like
```java
        if (player.getPrefs().getBoolean(context.getString(R.string.show_thumbnail_key), true)) {
            final BitMap bitmap = player.getThumbnail(); // might be done inline
            builder.putBitmap(MediaMetadataCompat.METADATA_KEY_ALBUM_ART, bitmap);
            builder.putBitmap(MediaMetadataCompat.METADATA_KEY_DISPLAY_ICON, bitmap);
        }
```
be better here?

app/src/main/java/org/schabi/newpipe/player/notification/NotificationUtil.java
```diff
@@ -104,9 +107,11 @@ private synchronized NotificationCompat.Builder createNotification() {
             compactSlots[i] = compactSlotList.get(i);
         }
 
-        builder.setStyle(new androidx.media.app.NotificationCompat.MediaStyle()
-                    .setMediaSession(player.getMediaSessionManager().getSessionToken())
-                    .setShowActionsInCompactView(compactSlots))
+        final MediaStyle mediaStyle = new MediaStyle().setShowActionsInCompactView(compactSlots);
+        player.UIs().get(MediaSessionPlayerUi.class).flatMap(MediaSessionPlayerUi::getSessionToken)
+                        .ifPresent(mediaStyle::setMediaSession);
```
litetex, line 112: I would write it in this style. Easier readable in my opinion since the relevant calls are structure under each other.
```suggestion
        player.UIs()
            .get(MediaSessionPlayerUi.class)
            .flatMap(MediaSessionPlayerUi::getSessionToken)
            .ifPresent(mediaStyle::setMediaSession);
```

app/src/main/java/org/schabi/newpipe/player/mediasession/PlayQueueNavigator.java
```diff
@@ -1,106 +1,152 @@
...
+        if (item == null) {
+            return null;
+        }
+
+        final MediaDescriptionCompat.Builder descBuilder = new MediaDescriptionCompat.Builder()
+                .setMediaId(String.valueOf(index))
+                .setTitle(item.getTitle())
+                .setSubtitle(item.getUploader());
+
+        // set additional metadata for A2DP/AVRCP
```
litetex, line 129: What's ``A2DP/AVRCP``?

litetex (changes requested):
Code LGTM mostly.

but there are build failures and some minor problems:

PS: I will test this when the build works again

app/src/main/java/org/schabi/newpipe/player/mediasession/PlayQueueNavigator.java
opusforlife2: I thought you would have Wiki'd it immediately. https://en.wikipedia.org/wiki/List_of_Bluetooth_profiles

app/src/main/java/org/schabi/newpipe/player/mediasession/MediaSessionPlayerUi.java
Stypox: [Previously](https://github.com/TeamNewPipe/NewPipe/blob/cacce6d2d0946ecf2b3cd79e20129031c2c0ca45/app/src/main/java/org/schabi/newpipe/player/helper/MediaSessionManager.java#L137) the bitmap was set only if it was non null, so I made sure the same behavior is kept, avoiding setting `null` thumbnails, even though it should not really be a problem.

app/src/main/java/org/schabi/newpipe/player/mediasession/PlayQueueNavigator.java
Stypox: I have no idea, I didn't write this code, I just copied it from [here](https://github.com/TeamNewPipe/NewPipe/blob/cacce6d2d0946ecf2b3cd79e20129031c2c0ca45/app/src/main/java/org/schabi/newpipe/player/playback/PlayerMediaSession.java#L69) ;-)
I will improve the comment anyway.

Stypox:
I applied two review suggestions and rebased.

> there are build failures

Not a problem in this PR, see #8679

app/src/main/java/org/schabi/newpipe/player/mediasession/MediaSessionPlayerUi.java
Stypox: Actually, I don't know. Maybe we should remove this TODO and call it a day, since it has been there since forever and NewPipe's behavior when headsets are plugged or unplugged is currently correct. What do you think?

opusforlife2:
> new thumbnail would never be shown

> wrong seekbar properties (duration and current position) being shown in the notification sometimes

Anything else apart from these 2 issues?

> since they take a while to reproduce and if you have used NewPipe for a while you know about them by now.

Well, yes, but we wouldn't know which ones to look for between the current release and this PR.

Gotta test whatever we can, I guess.

opusforlife2:
In the middle of a video, I suddenly got a buffering indicator for ~half a second, and then the app completely crashed, with no error caught by Newpipe. Scoop's crash report:

```
FATAL EXCEPTION: main
Process: org.schabi.newpipe.debug.mediasessionui, PID: 3279
java.lang.NullPointerException: Attempt to invoke virtual method 'org.schabi.newpipe.player.ui.PlayerUiList org.schabi.newpipe.player.Player.UIs()' on a null object reference
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.lambda$addVideoPlayerView$19$org-schabi-newpipe-fragments-detail-VideoDetailFragment(VideoDetailFragment.java:1314)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment$$ExternalSyntheticLambda9.run(Unknown Source:2)
```

app/src/main/java/org/schabi/newpipe/player/mediasession/PlayQueueNavigator.java
litetex: Hmm okay, maybe we should fine tune the comment in the future to something like "set additional metadata for Bluetooth" - save a Wikipedia page call :D

app/src/main/java/org/schabi/newpipe/player/mediasession/MediaSessionPlayerUi.java
litetex: Then I think we should remove it.

If someone has a problem, we can create an issue and handle it there.

litetex (approved):
Code LGTM

Did a quick test and everything seems to work fine.
Profiler also said that there are no bigger performance problems.

However you may address opus comment above before merging.

And btw you may also cleanup your fork-repo, there are a ton of unused branches there which are fetched by Git when I check it out :/

Stypox:
@opusforlife2's problem is unrelated to this PR. It is instead another instance of #8661, I'm gonna open yet another PR...
<sub>Which problems does your phone have with that little piece of code? :joy:</sub>

opusforlife2:
> Which problems does your phone have with that little piece of code? 😂

Here's possibly another. <3

Rough STR (might not work): Opened a video, it didn't load. Realised internet was off. Turned it on. Tried to close and reopen the video details screen. <- crash occurred somewhere here.

## Exception
* __User Action:__ ui error
* __Request:__ ACRA report
* __Content Country:__ CH
* __Content Language:__ en-IN
* __App Language:__ en_IN
* __Service:__ none
* __Version:__ 0.23.1
* __OS:__ Linux Android 12 - 32

```
android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{2fc9070 u0 org.schabi.newpipe.debug.mediasessionui/org.schabi.newpipe.player.PlayerService}
	at android.app.ActivityThread.generateForegroundServiceDidNotStartInTimeException(ActivityThread.java:1965)
	at android.app.ActivityThread.throwRemoteServiceException(ActivityThread.java:1936)
	at android.app.ActivityThread.access$2700(ActivityThread.java:256)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2190)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7870)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
Caused by: android.app.StackTrace: Last startServiceCommon() call for this service was made here
	at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1877)
	at android.app.ContextImpl.startForegroundService(ContextImpl.java:1832)
	at android.content.ContextWrapper.startForegroundService(ContextWrapper.java:781)
	at androidx.core.content.ContextCompat$Api26Impl.startForegroundService(ContextCompat.java:931)
	at androidx.core.content.ContextCompat.startForegroundService(ContextCompat.java:703)
	at org.schabi.newpipe.player.helper.PlayerHolder.startService(PlayerHolder.java:126)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.openMainPlayer(VideoDetailFragment.java:1217)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.$r8$lambda$7cA0nfOW4EXxLpA03fs92UQohwM(Unknown Source:0)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment$$ExternalSyntheticLambda13.run(Unknown Source:2)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.replaceQueueIfUserConfirms(VideoDetailFragment.java:2130)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.openVideoPlayer(VideoDetailFragment.java:1184)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.openVideoPlayerAutoFullscreen(VideoDetailFragment.java:1197)
	at org.schabi.newpipe.fragments.detail.VideoDetailFragment.onClick(VideoDetailFragment.java:511)
	at android.view.View.performClick(View.java:7455)
	at android.view.View.performClickInternal(View.java:7432)
	at android.view.View.access$3700(View.java:835)
	at android.view.View$PerformClick.run(View.java:28810)
	at android.os.Handler.handleCallback(Handler.java:938)
	... 6 more

```
</details>
<hr>

litetex (approved):
Rebased and retestet, LGTM

Also thank you for the tip Stypox but I will stay at least for now with my
``git remote add Stypox https://github.com/Stypox/NewPipe.git`` and do the rest via GH desktop :D

PR #8682, opened 2022-07-23
Refactor LicenseFragmentHelper.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Make some improvements to `LicenseFragmentHelper`.

- Fixes #

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.kt
+++ b/app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.kt
@@ -64,29 +64,25 @@ private fun getHexRGBColor(context: Context, color: Int): String {
     return context.getString(color).substring(3)
 }
 
-fun showLicense(context: Context?, license: License): Disposable {
-    return showLicense(context, license) { alertDialog ->
-        alertDialog.setPositiveButton(R.string.ok) { dialog, _ ->
-            dialog.dismiss()
-        }
-    }
-}
-
 fun showLicense(context: Context?, component: SoftwareComponent): Disposable {
-    return showLicense(context, component.license) { alertDialog ->
-        alertDialog.setPositiveButton(R.string.dismiss) { dialog, _ ->
+    return showLicense(context, component.license) {
+        setPositiveButton(R.string.dismiss) { dialog, _ ->
             dialog.dismiss()
         }
-        alertDialog.setNeutralButton(R.string.open_website_license) { _, _ ->
+        setNeutralButton(R.string.open_website_license) { _, _ ->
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.kt
```diff
@@ -67,11 +67,10 @@ private fun getHexRGBColor(context: Context, color: Int): String {
...
 
 fun showLicense(context: Context?, component: SoftwareComponent): Disposable {
-    return showLicense(context, component.license) { alertDialog ->
-        alertDialog.setPositiveButton(R.string.dismiss) { dialog, _ ->
+    return showLicense(context, component.license) {
+        setPositiveButton(R.string.dismiss) { dialog, _ ->
             dialog.dismiss()
-        }
-        alertDialog.setNeutralButton(R.string.open_website_license) { _, _ ->
+        }.setNeutralButton(R.string.open_website_license) { _, _ ->
```
TacoTheDank, line 66: I would keep this on a new line, IMO.

app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.kt
```diff
@@ -1,3 +1,5 @@
+@file:JvmName("LicenseFragmentHelper")
```
TacoTheDank, line 1: None of these are used in Java, so I don't believe we need this.

TacoTheDank:
Also, I don't believe the default parameter commit is particularly useful.

app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.kt
Isira-Seneviratne: > Also, I don't believe the default parameter commit is particularly useful.

It allows the method overload to be removed.

TacoTheDank: I'm aware, it's just that IMO it didn't really do anything useful here, considering that both `showLicense` have different block arguments for the "backend" `showLicense`'s AlertDialog. I think the previous one is simpler to read, even if it makes the class have a few more lines. If it's merged as is, then the block's default parameter (with its arguments) is just overridden by the block arguments set in `showLicense(Context, SoftwareComponent)`. I do understand the reasoning behind it, I just think that it makes it overall harder to interpret.

I can't explain very well right now, my brain is kinda fried.

Stypox (approved):
Tested, works, thanks. The build was failing because of ktlint, I fixed that.

Issue #4638, opened 2020-10-25
System UI keeps stopping [Image Loader issue]


gulachev:
### Checklist

- [x] I am using the latest version - 0.20.1 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

1. Search the concerned video like 'Hello World'.
2. Enqueue in the background
3. Open the video. 
(Both the cases 2. and 3. produce the similar behaviour)

### Actual behaviour
The video or audio(when playing in background) starts playing, the screen starts flickering , in the sense that display gets on and off , flickers for some time displaying the messege -- 'System UI keeps stopping' (giving two options -- app info or close app). The app gets closed for the latter option, but if pressed anywhere else on the screen the video plays continuosly with audio, screen flickering for some time and then phone gets restarted.

### Expected behavior

The video to be played till the end.

### Screenshots/Screen recordings

![Screenshot_20201025-053028](https://user-images.githubusercontent.com/73409058/97095970-5dffab80-1655-11eb-8c3f-a6899094d436.png)

### Logs
 None

### Device info

 - Android version/Custom ROM version: Android 10
 - Device model: Motorola One Power (P30 Note)

k3a:
Related logcat logs https://pastebin.com/raw/c65CcAyN

Sekitan:
Same here, same issue with my Sony Xperia XZ2 Compact...

UserX404:
@k3a 
- Do you encouter the same behavior as described here?
https://github.com/TeamNewPipe/NewPipe/issues/4644#issuecomment-716726661

k3a:
I think it may be the same problem. But in my case, I was using normal player (no background one). The app opened fine, allowed me to search a query. Clicking on a search result started playing the video but soon after System UI would crash, forcing the lock screen to appear. As I have fingerprint unlocking I was able to unlock, then after a second or two, System UI would crash again and so on until I force-closed NewPipe from the multitask switcher quickly. After hour or so I tried to open NewPipe again and the same thing would happen, suggesting it is a bug. Then I found this github issue.

[...]

PR #8686, opened 2022-07-24
Use smoother bitmap downscaling for thumbnails.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Use [`BitmapCompat.createScaledBitmap()`](https://developer.android.com/reference/androidx/core/graphics/BitmapCompat#createScaledBitmap(android.graphics.Bitmap,int,int,android.graphics.Rect,boolean)) for smoother thumbnail downscaling.

- Fixes #

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.kt
+++ b/app/src/main/java/org/schabi/newpipe/about/LicenseFragmentHelper.kt
@@ -36,21 +36,17 @@ private fun getFormattedLicense(context: Context, license: License): String {
  */
 private fun getLicenseStylesheet(context: Context): String {
     val isLightTheme = ThemeHelper.isLightThemeSelected(context)
-    return (
-            "body{padding:12px 15px;margin:0;" + "background:#" + getHexRGBColor(
-                context,
-                if (isLightTheme) R.color.light_license_background_color
-                else R.color.dark_license_background_color
-            ) + ";" + "color:#" + getHexRGBColor(
-                context,
-                if (isLightTheme) R.color.light_license_text_color
-                else R.color.dark_license_text_color
-            ) + "}" + "a[href]{color:#" + getHexRGBColor(
-                context,
-                if (isLightTheme) R.color.light_youtube_primary_color
-                else R.color.dark_youtube_primary_color
-            ) + "}" + "pre{white-space:pre-wrap}"
-            )
+    val licenseBackgroundColor = getHexRGBColor(
+        context, if (isLightTheme) R.color.light_license_background_color else R.color.dark_license_background_color
+    )
+    val licenseTextColor = getHexRGBColor(
+        context, if (isLightTheme) R.color.light_license_text_color else R.color.dark_license_text_color
+    )
+    val youtubePrimaryColor = getHexRGBColor(
+        context, if (isLightTheme) R.color.light_youtube_primary_color else R.color.dark_youtube_primary_color
+    )
+    return "body{padding:12px 15px;margin:0;background:#$licenseBackgroundColor;color:#$licenseTextColor}" +
+        "a[href]{color:#$youtubePrimaryColor}pre{white-space:pre-wrap}"
 }
 
 /**


Isira-Seneviratne:
When checking the change in `PicassoHelper` on my Galaxy Tab A6 (Android 8.1), I found that the result bitmap was mutable while the source was immutable, so I removed the bug fix for the issue mentioned. I'll restore it if needed.

app/src/main/java/org/schabi/newpipe/util/PicassoHelper.java
```diff
@@ -133,28 +134,18 @@ public Bitmap transform(final Bitmap source) {
...
                                 source,
                                 (int) notificationThumbnailWidth,
                                 (int) (source.getHeight()
                                         / (source.getWidth() / notificationThumbnailWidth)),
+                                null,
                                 true);
 
-                        if (result == source) {
-                            // create a new mutable bitmap to prevent strange crashes on some
-                            // devices (see #4638)
```
Stypox, line 145: You should not ignore this comment. See #4638

Stypox (changes requested):
Could you provide screenshots for comparison? Does the benefits exceed the cost of possibly having bugs or crashes?

Isira-Seneviratne:
> Could you provide screenshots for comparison? Does the benefits exceed the cost of possibly having bugs or crashes?



The documentation mentions that a mutable bitmap is returned if the original was immutable.

app/src/main/java/org/schabi/newpipe/util/PicassoHelper.java
Stypox: Oh, sorry, I didn't understand. I would rather not remove the workaround, since it caused many severe problems for users. But let's first ask the ones having problems to test themselves.

Stypox:
@gulachev @arielshulman could you test if this PR still works fine, or if it reintroduces #4638? Assuming you still have the same phone.

Isira-Seneviratne:
@Stypox I added a copy operation if the resulting bitmap is immutable (this only happens if the source bitmap was returned and it was immutable).

app/src/main/java/org/schabi/newpipe/util/PicassoHelper.java
```diff
@@ -133,28 +134,21 @@ public Bitmap transform(final Bitmap source) {
...
-                                    true);
-                            source.recycle();
-                            return copied;
-                        } else {
-                            source.recycle();
-                            return result;
+                        // create a new mutable bitmap to prevent strange crashes on some devices
+                        // (see #4638)
+                        if (!result.isMutable()) {
+                            result = result.copy(result.getConfig(), true);
```
Stypox, line 148: Nope, you can't just copy. We found out this would lead to crashes anyway (yeah, it doesn't make sense). Instead a scaled bitmap needs to be created. But anyway, let's wait for what users have to say.

Isira-Seneviratne: Possibly because the bitmap was a hardware bitmap, and those are always immutable ([the copy method](https://developer.android.com/reference/android/graphics/Bitmap#copy(android.graphics.Bitmap.Config,%20boolean)) throws an exception if a mutable copy is requested for such a bitmap with the hardware config). `BitmapCompat.createScaledBitmap()` creates a copy of the bitmap with a different config if it's a hardware bitmap.

Stypox: No, the copy worked correctly as far as I remember. But for some users the system crashed when that copied bitmap was displayed as the lock screen thumbnail.

Isira-Seneviratne: > No, the copy worked correctly as far as I remember. But for some users the system crashed when that copied bitmap was displayed as the lock screen thumbnail.

Interesting.

Maybe this line is the cause of the issue? https://github.com/androidx/androidx/blob/a66defb2607fc20c7d42ea455a0883d80033529c/palette/palette/src/main/java/androidx/palette/graphics/Palette.java#L610

arielshulman:
> @gulachev @arielshulman could you test if this PR still works fine, or if it reintroduces #4638? Assuming you still have the same phone.

Hey, brought that phone to my boy... 
I've tested the apk from CI and it worked great with those specific bitmaps which crashed the phone when it displayed on notification bar.

app/src/main/java/org/schabi/newpipe/util/PicassoHelper.java
Stypox: Mmmh, I think the problem was deeper than that. I think it was a crash in some C code, and since it happened only rarely on some phones, I think it was a bug in their operating systems themselves.

arielshuman said his phone has no problem, but actually, I think it's better to keep the old code and just upgrade from `Bitmap` to `BitmapCompat.createScaledBitmap` in both places where it is used. The only difference I would make is to create a new thumbnail both if `result == source` and if `!result.isMutable()` (remember to cleanup the intermediate `result` also if it is different from `source`, if the `if` is entered). Sorry for not providing a definite answer earlier, but I think it's better to be on the safe side, considering this was a bug that costed us many hours of debugging a while back.

TobiGr (approved):
lgtm

app/src/main/java/org/schabi/newpipe/util/PicassoHelper.java
```diff
@@ -139,21 +140,23 @@ public Bitmap transform(final Bitmap source) {
...
+                        final Bitmap result = BitmapCompat.createScaledBitmap(
                                 source,
                                 (int) notificationThumbnailWidth,
                                 (int) (source.getHeight()
                                         / (source.getWidth() / notificationThumbnailWidth)),
+                                null,
                                 true);
 
-                        if (result == source) {
+                        if (result == source && !result.isMutable()) {
```
Stypox, line 151: What I meant with

> create a new thumbnail both if result == source and if !result.isMutable()

was `||`, not `&&`, sorry for the confusion.
```suggestion
                        if (result == source || !result.isMutable()) {
```

Stypox (changes requested):

Stypox:
Something strange happened. Anyway, this is approved and merged in #9479, thanks @Isira-Seneviratne :-)

PR #8708, opened 2022-07-30
Reduce View.kt size.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use nested functions in `View.kt`.
- Create reusable class which overrides `AnimatorListenerAdapter`.

- Fixes #

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/ktx/View.kt
+++ b/app/src/main/java/org/schabi/newpipe/ktx/View.kt
@@ -90,7 +90,9 @@ fun View.animate(
  */
 fun View.animateBackgroundColor(duration: Long, @ColorInt colorStart: Int, @ColorInt colorEnd: Int) {
     if (MainActivity.DEBUG) {
-        Log.d(TAG, "animateBackgroundColor() called with: view = [$this], duration = [$duration], " +
+        Log.d(
+            TAG,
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/ktx/View.kt
```diff
@@ -332,6 +274,19 @@ fun View.animateHideRecyclerViewAllowingScrolling() {
...
+    private val execOnEnd: Runnable?,
+    private val shouldHideView: Boolean
+) : AnimatorListenerAdapter() {
+    override fun onAnimationEnd(animation: Animator) {
+        if (shouldHideView) {
+            view.isGone = true
+        }
+        execOnEnd?.run()
+    }
+}
```
Stypox, line 288: I'd create two classes, one for when `shouldHideView` is `true`, one for when it's `false`, in order to avoid that runtime check (maybe the second can extend the first, but idk). The names for the two classes should be imo: `HideAndExecOnEndListener` and `ExecOnEndListener`. `RunnableExecutorAdapter` is not so meaningful.

Stypox (changes requested):
Thanks, looks almost good

Stypox (approved):

PR #8715, opened 2022-07-31
Update dimens.xml to better display in landscape mode.

TheDawnOvO:
Hi, I slightly reduced the font size of the upload date to avoid that it is not displayed fully in landscape mode.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

### DIFF
--- a/app/src/main/res/values-land/dimens.xml
+++ b/app/src/main/res/values-land/dimens.xml
@@ -7,7 +7,7 @@
     <dimen name="video_item_search_title_text_size">16sp</dimen>
     <dimen name="video_item_search_duration_text_size">12sp</dimen>
     <dimen name="video_item_search_uploader_text_size">14sp</dimen>
-    <dimen name="video_item_search_upload_date_text_size">14sp</dimen>
+    <dimen name="video_item_search_upload_date_text_size">12sp</dimen>
     <!-- Elements Size -->
     <!-- 16 / 9 ratio-->
     <dimen name="video_item_search_thumbnail_image_width">142dp</dimen>


app/src/main/res/values-land/dimens.xml
```diff
@@ -7,7 +7,7 @@
     <dimen name="video_item_search_title_text_size">16sp</dimen>
     <dimen name="video_item_search_duration_text_size">12sp</dimen>
     <dimen name="video_item_search_uploader_text_size">14sp</dimen>
-    <dimen name="video_item_search_upload_date_text_size">14sp</dimen>
```
Stypox, line 10: All of these values above are related to each other. If you change one, you should change others too, otherwise the text sizes will look off (as you can see in your screenshots, the uploader is bigger than the upload date).

How does it look like if you just remove all of these, so that [the corresponding smaller values in `values/dimens.xml`](https://github.com/TeamNewPipe/NewPipe/blob/bbf134bc6b19d0c1836029bce1181325dd54284b/app/src/main/res/values/dimens.xml#L17-L21) are used?

Please try out some sets of values (including the suggestion above) and provide some screenshots, so that we can choose the best.

Also, are you sure somebody using a TV or a tablet is able to read text so small?

Stypox (changes requested):

Stypox:
You seem to have ignored or removed the Pull Request template; please don't do that. Also, make sure your images are not gigantic (see how I edited the PR description).

Stypox:
Closing as there has not been any recent activity, and we're going to rewrite layouts in Compose soon anyway

PR #8683, opened 2022-07-23
Update Lifecycle to 2.5.1.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [x] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Update AndroidX Lifecycle from 2.3.1 to 2.5.0. [Changelog](https://developer.android.com/jetpack/androidx/releases/lifecycle)

#### Fixes the following issue(s)

- Fixes #

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #8718, opened 2022-07-31
Use Paging.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use the [AndroidX Paging](https://developer.android.com/jetpack/androidx/releases/paging) library.

- Fixes #

- https://github.com/TeamNewPipe/NewPipe/pull/8683

#### APK testing 

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -98,6 +98,7 @@ ext {
     checkstyleVersion = '10.3.1'
 
     androidxLifecycleVersion = '2.5.1'
+    androidxPagingVersion = '3.1.1'
     androidxRoomVersion = '2.4.3'
     androidxWorkVersion = '2.7.1'
 
@@ -208,6 +209,8 @@ dependencies {
     implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:${androidxLifecycleVersion}"
     implementation 'androidx.localbroadcastmanager:localbroadcastmanager:1.1.0'
     implementation 'androidx.media:media:1.6.0'
+    implementation "androidx.paging:paging-runtime:${androidxPagingVersion}"
+    implementation "androidx.paging:paging-rxjava3:${androidxPagingVersion}"
     implementation 'androidx.preference:preference:1.2.0'
     implementation 'androidx.recyclerview:recyclerview:1.2.1'
     implementation "androidx.room:room-runtime:${androidxRoomVersion}"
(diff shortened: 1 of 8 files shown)


TacoTheDank:
This PR contains the commits from the Lifecycle PR. Could you fix this please?

app/build.gradle
```diff
@@ -218,6 +219,8 @@ dependencies {
     implementation 'androidx.viewpager2:viewpager2:1.1.0-beta01'
     implementation "androidx.work:work-runtime-ktx:${androidxWorkVersion}"
     implementation "androidx.work:work-rxjava3:${androidxWorkVersion}"
+    implementation "androidx.paging:paging-runtime:${androidxPagingVersion}"
```
TacoTheDank, line 222: Please keep the AndroidX libraries in alphabetical order (applies to the ext block too).

app/src/main/java/org/schabi/newpipe/fragments/list/search/SuggestionListPagingAdapter.kt
```diff
@@ -0,0 +1,37 @@
...
+        fun onSuggestionItemLongClick(item: SuggestionItem)
+    }
+
+    override fun onBindViewHolder(holder: SuggestionItemHolder, position: Int) {
+        TODO("Not yet implemented")
+    }
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): SuggestionItemHolder {
+        TODO("Not yet implemented")
+    }
```
TacoTheDank, line 24: TODO detected, is there something to be done here?

TacoTheDank (commented):
I like the idea of this a lot.

Some comments:

Stypox:
Interesting! Since I don't know much about good practices for Android design, if you could provide some insight into how you did things I would be really glad :-)

Isira-Seneviratne:
It seems like making use of the pagination library will require changes to the extractor as well, since it assumes that the additional page items will be of the same type as the initial ones.

Stypox:
I think each and every list present in NewPipe should be the same, and should hence be able to display all kinds of info items. E.g. info items that come to my mind are: video, channel, playlist, local playlist, video from history, subscription, ... So `Paging` should be made so that it can handle all item types we use at the same time. Does this help you?

Isira-Seneviratne:
> I think each and every list present in NewPipe should be the same, and should hence be able to display all kinds of info items. E.g. info items that come to my mind are: video, channel, playlist, local playlist, video from history, subscription, ... So `Paging` should be made so that it can handle all item types we use at the same time. Does this help you? 

Yeah, definitely. I think it would be best to start with using a view model for the search fragment (I made a separate PR for that) before switching to Paging.

PR #8728, opened 2022-08-03
Use Comparator factory methods.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Simplify audio and video stream comparisons using the `Comparator` factory methods.
- Use `Comparator.comparingDouble()` in the preference search to avoid boxing.

- Fixes #

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
@@ -343,7 +343,10 @@ public final class ListHelper {
      */
     private static List<VideoStream> sortStreamList(final List<VideoStream> videoStreams,
                                                     final boolean ascendingOrder) {
-        final Comparator<VideoStream> comparator = ListHelper::compareVideoStreamResolution;
+        // Compares the quality of two video streams.
+        final Comparator<VideoStream> comparator = Comparator.nullsLast(Comparator
+                .comparing(VideoStream::getResolution, ListHelper::compareVideoStreamResolution)
+                .thenComparingInt(s -> VIDEO_FORMAT_QUALITY_RANKING.indexOf(s.getFormat())));
         Collections.sort(videoStreams, ascendingOrder ? comparator : comparator.reversed());
         return videoStreams;
     }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/apache/commons/text/similarity/FuzzyScore.java
```diff
@@ -84,7 +84,7 @@ public FuzzyScore(final Locale locale) {
      * @return result score
      * @throws IllegalArgumentException if the term or query is {@code null}
      */
-    public Integer fuzzyScore(final CharSequence term, final CharSequence query) {
+    public int fuzzyScore(final CharSequence term, final CharSequence query) {
```
Stypox, line 87: I think this was taken from [here](https://commons.apache.org/sandbox/commons-text/jacoco/org.apache.commons.text.similarity/FuzzyScore.java.html). Does [the apache license](http://www.apache.org/licenses/LICENSE-2.0) allow modifying code? If so, does it require changes to be documented?

Stypox (changes requested):
Code looks good to me

app/src/main/java/org/apache/commons/text/similarity/FuzzyScore.java
Isira-Seneviratne: I undid the change.

app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -374,11 +376,15 @@ static int getHighestQualityAudioIndex(@Nullable final MediaFormat format,
      */
     static int getMostCompactAudioIndex(@Nullable final MediaFormat format,
                                         @Nullable final List<AudioStream> audioStreams) {
-
         return getAudioIndexByHighestRank(format, audioStreams,
                 // The "-" is important -> Compares ascending (first = highest rank)
```
Stypox, line 380:
```suggestion
                // The "reversed()" is important -> Compares ascending (first = highest rank)
```

Stypox (changes requested):

Stypox (approved):
I tested the download dialog, also `ListHelper` tests pass, so this is fine

PR #8746, opened 2022-08-08
Move search suggestion logic into a view model.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Created a view model for the search fragment, to separate the search logic from the UI, and moved the search suggestion loading logic there. I did not encounter any issues while making this change, and it should work as it did prior to the change.
- Renamed the extractor helper's `suggestionsFor` method to `getSuggestionsFor`.

#### Fixes the following issue(s)

- Fixes #

#### Relies on the following changes

- https://github.com/TeamNewPipe/NewPipe/pull/8719

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

Issue #5662, opened 2021-02-21
Add more search filters for videos


SameenAhnaf:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

Although, Newpipe has most basic filters, advanced video search is a little mess. Please add all types of filters soon.

#### Additional context

![IMG_20210219_212821](https://user-images.githubusercontent.com/79245641/108641489-b6429f00-74c9-11eb-8008-f086ab9a1574.jpg)

#### How will you/everyone benefit from this feature?

It will help users search for the most appropriate video in the most accurate way.

HTSamurai:
Oh yes!

SameenAhnaf:
Duplicate of #110

Issue #3824, opened 2020-06-28
A global language filter


MD77MD:
#### Describe the feature you want

It would be nice if a global language filter can be implemented so that newPipe would only show videos from selected languages... the idea was inspired from Thorium project, they have an option in settings where you can select multiple languages then the app would only show videos from those selected languages globally, this includes trending videos, searches,...etc.
#### Is your feature request related to a problem? Please describe it

This would solve the problem when I search for videos on a certain subjects... many videos would show in the search list with languages that I cannot understand, to the point where sometimes they would overtake the whole search list... for example: if I search for "redmi note 9", most of the videos that come up are in Hindi. but there is no indication to whether it's in English or Hindi (because they use English titles). this forces me to open them one by one. 
#### How will you/everyone benefit from this feature?

it would make our lives a lot easier and stop wasting our precious time and data for opening videos we do not want. 
also this would stop the false increase in views number which I assume is a strategy used by some channels.

Stypox:
Unfortunately there is no way to determine the language of a youtube video before fully fetching its info. Also, content country does not work and that's a known issue (only english is supported).

B0pol:
> Unfortunately there is no way to determine the language of a youtube video before fully fetching its info. Also, content country does not work and that's a known issue (only english is supported).

Even when you fetch it. Unless I don't know something, but auto-generated captions are not reliable

MD77MD:
although skyTube uses newPipe extractor library, it has this feature implemented. the filtering in note perfect by any means, but still looking into how it's implemented could be useful.
https://skytube-app.com

leo-wiz:
Ukraine here. Our country was being Russified for centuries. Russia spent lots of money for this.
We really need language restrictions to protect our kids from russification here in Ukraine.
It would be great to protect the content-language option by password so a kid can't change the option.
Thanks.

PR #8837, opened 2022-08-19
Add support for advanced content and sort filters in searches

### LINKED ISSUE #2251: Search fitler UI
This is a discussion on search and filter UI based on the work and discussion from https://github.com/TeamNewPipe/NewPipeExtractor/pull/124

@Bluesir9 @theScrabi Here are my ideas for the search and filter UI. I think it is better to put this into the app repo, because UI discussions do not belong into the extractor repo. I tried to explain briefly, what my intentions and ideas were to make the benefits and disadvantages a bit clearer.

-----

TLDR; you can find the drafts at the end. It might  be good to take a look at them before reading the text so the text is better understandable.

-----

First of all, there were some things to consider before taking a look at the actual design.

NewPipe downloads a complete website for each search. That's why, I decided to make the filters directly visible, so people can see and change them __before performing the search__. This should be quite time and data saving.
Nevertheless, I am not that happy with the current position of the filter bar. While typing the search suggestions change and should help the user to suggest the search query he is looking for. Therefore, the filter bar seems to be "in the way" - just like a visual barrier. I considered to place the bar at the bottom of the screen. But this was also quite annoying, because the keyboard needs much space and the filter bar seems to be lost down there. Do you have any suggestions where to put it instead?

Because I decided to show the filters, I needed to group them, otherwise the bar contains too much info at once and is messy. Fortunately, the filters are already grouped by YouTube. Displaying text is a bad thing as it a) needs to be translated and b) people need way more time to read a text compared to see and understand an icon. Icons are intuitive and can be recognized later on. The difficult part here was to find icons which match the meaning of a group. I could not decided which icon to use for the `duration` category. I'd say, I prefer the clock icon above the timer because it matches with the two filters best. In other words, the disabled state looks quite similar to the to other states (`short` and `long`).

Apropos __state__. I didn't mention it before, so I do now. In the filter bar, there are five icons, each for one category. Most categories allow only one of their filters to be applied at a time. For example, the type category has six states: `video`, `channel`, `playlist`, `movie`, `show` and `disabled / inactive` (currently represented in NewPipe by `all`). Filtering by `all` and `video` is not possible. The second idea behind showing the filters in the bar represented by icons is to display their current state. This means, we replace the category icon when a filter is selected.. This needs  a bunch of icons which represent the active filter. For most filters, good icons already exist (e.g. `video`, `channel`, `playlist`, `thumb up` for `best rated` etc.) and are easy to find. 
But finding good icons for the `upload date` and `features` groups drove me crazy.
I couldn't find a way to abstract their filters without using numbers and letters. My decision is to represent the times by using the calendar and putting the time into it (see below). 
There are too many `features` and in difference to all the other categories YouTube allows to apply multiple feature filters. Therefore we need to handle them different to the others.

For the "normal" categories I suggest to show a normal dialog containing the icon and filter name whereas the feature filter needs a bit more work. There are eleven feature filters provided by YouTube. I am not sure how many we want to show (We could either show all and ignore that our players do not support all features yet and provide the download of the videos so the user can an an other player, or just reduce the list to those filter which are supported by exoplayer).

Nevertheless, we need a different UI, because the single click dialogs are quite nonfunctional for this purpose. All together I needed to decided between two concepts: display the features as "badge" / "chip" or use the settings style with a toggle.
One one hand, we can put the chips after selecting them right into the filter bar. This allows the user to still see all applied filters. As said above, I'd like to achieve this behaviour. As a bonus, users are able to remove the filter chip with ony one click/touch. The big questions is how to select the chips. The [material guidelines for chips](https://material.io/design/components/chips.html#filter-chips) suggest to use horizontal scrolling whenever the chips need more than two rows, because otherwise the screen looks quite messy. However, I think we can ignore this suggestion when we use a blank fragment to select the chips from and group then in a smart way.
On the other hand, we do not need to show the currently applied feature filters in the filter bar. This allows us to not use chips to select the filters. I think it is not intuitive to use setting-like toggles to enable something which than appears as a chip somewhere else.
I'd like to go with the chips, but am open to your feedback, which is btw. appreciated to all things here, but in particular to this topic.

Just my two cents regarding the __type filter__: Apart from the video, channel and playlist filter, we can now support the movie and show filter. We might consider to disable the movie filter in the app, because the extractor currently does only render the `video-render` objects, but not `movie-render` objects. Nevertheless, movies need do be purchased, a login is therefore required. So the movie filter might not bring the expected results because only free "movies", which are uploaded as video, are shown in the results.

### Filter Icons
Filter_11

### General filter UI and behaviour
We can also put the feature filters into a new row instead of appending them.
Filter_21

evermind-zz:
#### What is it?
- [x] Bugfix (user facing)
- [x] Feature (user facing)

- provide sort filters (and content filters) that are available by the supported platforms itself

There are 4 different UI to test
- (1) dialog like (default)
- (2) action menu look alike dialog (my favourite)
- (3) action menu (legacy)
- (4) chip dialog

To choose go to settings -> appearance -> 'Select Search Filter UI'

Notes:
- UI (3) is slow but working (slow maybe because of misusing the action menu
##### Feedback needed
- which UIs do you want (to merge)? I would like to keep  1 and 2 and 4. 3 seems to be also liked. So keep them all!. ~~I will then rework the code as there is some duplication right now that should be eliminated first~~ Any duplication has been removed. 
##### Todo
- ~~fix sonar findings~~
- ~~fix your findings~~
- ~~string L10n~~

<table>
  <tr>
    <th>Select the UI</th>
    <th>(Ui 1) YouTube</th>
    <th>(UI 2) YouTube</th>
    <th>(UI 2) Peertube</th>
    <th>(UI 3) Live in action</th>
    <th>(UI 4) YouTube</th>
  </tr>
  <tr>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
     <td></td>
  </tr>
</table>

- Fixes #2251

- https://github.com/TeamNewPipe/NewPipeExtractor/pull/904

#### APK testing 

#### Due diligence


opusforlife2:
Maybe the UI could be changed from this giant scroll. Ideas?

mhmdanas:
@opusforlife2 something like https://github.com/TeamNewPipe/NewPipe/issues/2251#issuecomment-477288259 looks great to me.

opusforlife2:
@evermind-zz Is that UI something you would be interested in implementing? If not, we could always go with the current one as a minimum viable product and improve it later.

evermind-zz:
> @evermind-zz Is that UI something you would be interested in implementing? If not, we could always go with the current one as a minimum viable product and improve it later.

For me the interaction with the current UI is quick and I would like to keep it. But I am willing to give implementing the other UI a shot. Maybe I will like the new UI too or I could make it configurable in settings if someone like me prefer the 'old way'

MD77MD:
first thank you for this... is it possible to have all the options to show up in one box without scrolling... same like the playback speed mene... with proper layout for both portrait and landscape mode

also i think it is important to have [ok] [reset] buttons... so we can choose the filter /sort options in one go... not in multiple times and reloading results in each time

TacoTheDank:
> For me the interaction with the current UI is quick and I would like to keep it. But I am willing to give implementing the other UI a shot. Maybe I will like the new UI too or I could make it configurable in settings if someone like me prefer the 'old way'

If we go with the current design for now, I would at least suggest using a dedicated filter icon (via Google's [material icons](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:filter_list:FILL@0;wght@400;GRAD@0;opsz@48&icon.platform=android)). Having it be stuffed into the hamburger menu isn't very intuitive, IMO.

evermind-zz:
@MD77MD I do not really understand what you mean by you have to reload the gui. This version modifies the current menu filter  behaviour. Here you first select everything in the menu and than you have to scroll to the top and push search inside the menu. Than the menu will disappear. Have you tested this PR here?

But nevertheless I will come up with a dialog style UI

evermind-zz:
> > For me the interaction with the current UI is quick and I would like to keep it. But I am willing to give implementing the other UI a shot. Maybe I will like the new UI too or I could make it configurable in settings if someone like me prefer the 'old way'
> 
> If we go with the current design for now, I would at least suggest using a dedicated filter icon (via Google's [material icons](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:filter_list:FILL@0;wght@400;GRAD@0;opsz@48&icon.platform=android)). Having it be stuffed into the hamburger menu isn't very intuitive, IMO.

no matter which UI but a filter icon would be nice

Angelk90:
@evermind-zz : Are there any news?

evermind-zz:
> @evermind-zz : Are there any news?

still working on it! I've rewrote the logic part and have a new UI! But my time is very limited so it will still take some time

Angelk90:
@evermind-zz : Will it be a dialog box like youtube does?

MD77MD:
> first thank you for this... is it possible to have all the options to show up in one box without scrolling... same like the playback speed mene... with a proper layout for both portrait and landscape mode
> 

This is the closest thing to what i was trying to convey 

**Portrait mode**:
108641489-b6429f00-74c9-11eb-8008-f086ab9a1574
from #5662 by @SameenAhnaf 

**landscape mode**:
 the same as the above idea but stacked like pic below
94719607-ae843180-0353-11eb-8646-a957e1d6f41e

MD77MD:
With Ok, Reset buttons 
108641489-b6429f00-74c9-11eb-8008-f086ab9a1574

MD77MD:
Also do you think adding #3824 is possible? ... even later on.

Angelk90:
@evermind-zz, @MD77MD  : Are there any news?

opusforlife2:
@Angelk90 13 days since the last author reply is not nearly enough time to be asking for updates. This is closer to spam territory.

evermind-zz:
I'm still working on it. I'll post an update ASAP

nicoursi:
I would love to test it but the apk does not install on Android 11. Can you provide a suitable apk?  Normally I can install the artifacts from the workflows,  so there must be something wrong with this specific one.

evermind-zz:
> I would love to test it but the apk does not install on Android 11. Can you provide a suitable apk? Normally I can install the artifacts from the workflows, so there must be something wrong with this specific one.


the problem is that this one has no signature: but with
`adb install -t app-debug.apk`
it should be installable. If not possible I will create a new one with a signature

opusforlife2:
It became obvious in just a few tries that the dialog menu is _far_ better to use compared to the action menu. Not only is it comfortably info dense without the need for lots of scrolling, it lets you change multiple settings before applying the filter/sort.

evermind-zz:
> @evermind-zz : It does not allow to be installed.

The new 'artifact' one should work. See first post where to find it.

evermind-zz:
@Angelk90 thx! Are screens in portrait mode?

evermind-zz:
> @evermind-zz : Yes, in portrait mode.

Could you @Angelk90 do one screenshot for landscape too? That would be great.

evermind-zz:
> Action Menu (legacy), weird bug first, once I click I get first image then second image.:
> 
> 
> 
> 

This seems due to the fact that the keyboard is removed programmatically. But as I think this UI will not make it into NewPipe, we can ignore it as I probably remove it if no one wants to have it:)

app/src/main/java/org/schabi/newpipe/fragments/list/search/filter/SearchFilterDialogFragment.java
```diff
@@ -0,0 +1,55 @@
...
+ * A search filter dialog that also looks like a dialog aka. 'dialog style'.
+ */
+public class SearchFilterDialogFragment extends BaseSearchFilterDialogFragment {
+
+    private SearchFilterDialogFragmentBinding binding;
+
+    public static DialogFragment newInstance(
+            final int serviceId,
+            final ArrayList<Integer> userSelectedContentFilter,
+            final ArrayList<Integer> userSelectedSortFilter) {
```
Isira-Seneviratne, line 27: Likewise.

```suggestion
    public static DialogFragment newInstance(
            final int serviceId,
            final List<Integer> userSelectedContentFilter,
            final List<Integer> userSelectedSortFilter) {
```

app/src/main/java/org/schabi/newpipe/fragments/list/search/filter/SearchFilterDialogSpinnerAdapter.java
```diff
@@ -0,0 +1,220 @@
...
+
+public class SearchFilterDialogSpinnerAdapter extends BaseAdapter {
+
+    private final Context context;
+    private final FilterGroup group;
+    private final BaseSearchFilterUiGenerator.UiWrapperMapDelegate wrapperDelegate;
+    private final Spinner spinner;
+    private final Map<Integer, Integer> id2PosMap = new HashMap<>();
+    private final Map<Integer, UiItemWrapperSpinner>
+            viewWrapperMap = new HashMap<>();
```
Isira-Seneviratne, line 33: You might want to use [`SparseIntArray`](https://developer.android.com/reference/android/util/SparseIntArray) and [`SparseArray`](https://developer.android.com/reference/android/util/SparseArray) / [`SparseArrayCompat`](https://developer.android.com/reference/androidx/collection/SparseArrayCompat) instead to reduce memory usage.

app/src/main/java/org/schabi/newpipe/fragments/list/search/filter/SearchFilterDialogSpinnerAdapter.java
```diff
@@ -0,0 +1,220 @@
...
+        return view;
+    }
+
+    @SuppressLint("WrongConstant")
+    private void initViewWithData(final int position,
+                                  final FilterItem item,
+                                  final TextView view) {
+        final UiItemWrapperSpinner wrappedView =
+                viewWrapperMap.get(position);
+        Objects.nonNull(wrappedView);
```
Isira-Seneviratne, line 84: I'm assuming this is a non-null enforcement check.

```suggestion
        Objects.requireNonNull(wrappedView);
```

app/src/main/java/org/schabi/newpipe/fragments/list/search/filter/SearchFilterOptionMenuAlikeDialogFragment.java
```diff
@@ -0,0 +1,88 @@
...
+import androidx.fragment.app.DialogFragment;
+
+public class SearchFilterOptionMenuAlikeDialogFragment extends BaseSearchFilterDialogFragment {
+
+    private SearchFilterOptionMenuAlikeDialogFragmentBinding binding;
+
+    public static DialogFragment newInstance(
+            final int serviceId,
+            final ArrayList<Integer> userSelectedContentFilter,
+            final ArrayList<Integer> userSelectedSortFilter) {
```
Isira-Seneviratne, line 30: See the earlier suggestion.

```suggestion
    public static DialogFragment newInstance(
            final int serviceId,
            final List<Integer> userSelectedContentFilter,
            final List<Integer> userSelectedSortFilter) {
```

app/src/main/java/org/schabi/newpipe/fragments/list/search/filter/BaseSearchFilterUiDialogGenerator.java
```diff
@@ -0,0 +1,78 @@
...
+            view.getKey().setOnClickListener(view.getValue());
+        }
+    }
+
+    @Override
+    public void onPause() {
+        for (final Map.Entry<View, View.OnClickListener> view : viewListeners.entrySet()) {
+            view.getKey().setOnClickListener(null);
+        }
+    }
```
Isira-Seneviratne, line 59: Why is setting/unsetting the view listeners needed here?

app/src/main/java/org/schabi/newpipe/fragments/list/search/filter/BaseSearchFilterDialogFragment.java
```diff
@@ -0,0 +1,188 @@
...
+        dismiss();
+    }
+
+    /**
+     * Listener to be implemented by the parent Fragment so it can receive data.
+     */
+    public interface Listener {
+
+        void onFinishSearchFilterDialog(ArrayList<Integer> userSelectedContentFilterList,
+                                        ArrayList<Integer> userSelectedSortFilterList,
```
Isira-Seneviratne, line 184: See the earlier suggestion.

```suggestion
        void onFinishSearchFilterDialog(List<Integer> userSelectedContentFilterList,
                                        List<Integer> userSelectedSortFilterList,
```

Isira-Seneviratne:
In addition to the above changes, you'll want to consider making use of a [view model](https://developer.android.com/topic/libraries/architecture/viewmodel), to separate the search logic from the UI. For reference, you can check this PR: #8746

app/src/main/java/org/schabi/newpipe/fragments/list/search/filter/BaseSearchFilterUiDialogGenerator.java
evermind-zz: > Why is setting/unsetting the view listeners needed here?

I think this was a left over. Once I created the TextView's for the Spinners I kept those TextView's for reuse. There I got a null pointer exception. Removing/ and readding solved the problem. But later I figured out I missued the Spinners TextView's  instead of using the convertView parameter of the Adapters getView() method.

So this is a left over. I will remove the listener stuff there. Thanks!

app/src/main/java/org/schabi/newpipe/fragments/list/search/filter/SearchFilterDialogSpinnerAdapter.java
evermind-zz: I've converted all of them to SparseArrayCompat even though I could use SparseIntArray to some extend.
The problem is that my unit test otherwise has to be converted to androidTest and that would slow testing down so I wanted to avoid it. Another options would be to use roboelectric. If adding roboelectric to the test dependencies would be an option let me know.

Skillanswer:
@evermind-zz 

 I don't know if this is  the right place or right way to ask but...

Can you please also add an "Custom upload Date Range" option  under "sort by / date" section? 

It would have a Start Date and an End Date.So, only the results uploaded between those dates will appear.

It would be like Duckduckgo custom date range filter.🙏🏼

Screenshot_20221115-120840-188

Screenshot_20221115-120840-188

evermind-zz:
@Skillanswer It is not easily possible as here we are not filtering data/videos information we have already receive, instead we just request what is possible for each service and then we get the results. I think for Peertube this kind of scenario is possible as their API might support it. Duckduckgo has their own database they can query.

evermind-zz:
> In addition to the above changes, you'll want to consider making use of a [view model](https://developer.android.com/topic/libraries/architecture/viewmodel), to separate the search logic from the UI. For reference, you can check this PR: #8746

I'll have a look

evermind-zz:
> > In addition to the above changes, you'll want to consider making use of a [view model](https://developer.android.com/topic/libraries/architecture/viewmodel), to separate the search logic from the UI. For reference, you can check this PR: #8746
> 
> I'll have a look

Done

RomainBertrand:
> @Isira-Seneviratne :
> 
> It seems that some tests have failed.
> 
> 

Specifically, only the building off the app fails due to a StackOverflowError : 
image

AudricV:
No, there is a compile error that needs to be fixed by @evermind-zz: see https://github.com/TeamNewPipe/NewPipe/actions/runs/3512433178/jobs/5884127881#step:6:75

evermind-zz:
> No, there is a compile error that needs to be fixed by @evermind-zz: see https://github.com/TeamNewPipe/NewPipe/actions/runs/3512433178/jobs/5884127881#step:6:75

yeah it fails because it depends on changes in TeamNewPipe's NewPipeExtractor. I'll add a commit to point to my NewPipeExtractor PR branch so it will build

evermind-zz:
@Angelk90 you refer to the default UI?

evermind-zz:
@Angelk90 you refer to the default UI?

evermind-zz:
Everything should be fine now
- DividerItem stuff moved to NewPipe
- default search filter dialog will no longer stretch vertically

tsiflimagas:
You can just open GitHub in a browser and download it from there.

evermind-zz:
@Angelk90 

> @evermind-zz : Much better, could you make it like ui like the youtube one?

No I will not imitate the youtube UI completely for me it is enough.
I tweaked the appearance of the Chip's a bit
> java.lang.ClassCastException: androidx.appcompat.view.menu.MenuItemImpl cannot be cast to org.schabi.newpipe.fragments.list.search.filter.InjectFilterItem$DividerItem

Thx and fixed!

MD77MD:
This UI is much simpler, beautiful and functional ... no need to hop from one menu to another

190019317-4e61378c-ebb3-4e30-8509-1c42daffc22a

as a bounce,  it would work with TVs much better ... more suitable for remote controls

evermind-zz:
> This UI is much simpler, beautiful and functional ... no need to hop from one menu to another
> 
> 190019317-4e61378c-ebb3-4e30-8509-1c42daffc22a
> 
> as a bounce, it would work with TVs much better ... more suitable for remote controls

@MD77MD  Here we go
 205103650-656bac31-3585-4960-b556-2265b4e7735

It is selectable in the options as 'Chip dialog'

app/src/main/java/org/schabi/newpipe/fragments/list/search/filter/BaseSearchFilterUiGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+        this.logic = logic;
+        this.contentFilterWorker = createContentFilterWorker();
+        this.sortFilterWorker = createSortFilterWorker();
+    }
+
+    // common helpers
+    public static float dipToPixels(@NonNull final Context ctx, final float dipValue) {
+        final DisplayMetrics metrics = ctx.getResources().getDisplayMetrics();
+        return TypedValue.applyDimension(COMPLEX_UNIT_DIP, dipValue, metrics);
+    }
```
TobiGr, line 42: You can use `DeviceUtils.dpToPx(int, Context)`

evermind-zz: done

evermind-zz:
> @evermind-zz : Seems to work fine, when will the merge be done?

I guess then when the core developers have the time for more reviews and accept the changes.

Skillanswer:
I am using this for months now. it is a really useful feature. It is working without any issues or hiccups. So when would it get  merged?
I am commenting this because of the last update. I was hoping that  this PR would be a part of that.

avatar1024:
The feature works really well, thank you!! FYI importing database from latest version (v0.25), either with or without settings, makes the current test apk (based on 0.24.1) crash on launch.

javiimo:
If the artifact has expired, where could I download the APK?

evermind-zz:
> If the artifact has expired, where could I download the APK?

You could try this feature here:
https://github.com/bravenewpipe/NewPipe/releases

jcperil:
> https://github.com/bravenewpipe/NewPipe/releases

Hi, I installed this to check the filter feature. As others have mentioned, yes, it is working. Although, I am facing an issue in terms of the UI. It is tightly spaced for someone with large fingers.



I have a few ideas on how to improve it. Please let me know what you think.

1. The highlight words "Content filters" and "Sort filters" are taking up most of the space. Perhaps, you can write them in capital/block letters, change the font and reduce font size. This will free up some space and yet remain a highlight.

2. Sufficient spacing between options as well as in drop-down submenus.

#### Illustration:

AudricV:
@evermind-zz

After updating your extractor changes, could you please rebase this PR on the `dev` branch (or merge it into yours if that's too difficult to rebase for you) and include your updated extractor in it? Thank you in advance.

> which UIs do you want (to merge)? I would like to keep 1 and 2 and 4. 3 seems to be also liked. So keep them all!

I don't think this is a good idea to maintain 4 UIs of the search filters, especially with the current app codebase.

I personally prefer UIs 1 and 4, with minor changes (like converting them to an alert dialog instead of a dialog similar to the download one if possible and not done in a hacky way).

Stypox:
I already started the extractor rebase (and almost finished, I just need to integrate 2 things), sorry for not writing that anywhere

evermind-zz:
> @evermind-zz
> 
> After updating your extractor changes, could you please rebase this PR on the `dev` branch (or merge it into yours if that's too difficult to rebase for you) and include your updated extractor in it? Thank you in advance.
> 
> > which UIs do you want (to merge)? I would like to keep 1 and 2 and 4. 3 seems to be also liked. So keep them all!
> 
> I don't think this is a good idea to maintain 4 UIs of the search filters, especially with the current app codebase.
> 
> I personally prefer UIs 1 and 4, with minor changes (like converting them to an alert dialog instead of a dialog similar to the download one if possible and not done in a hacky way).

we can go with UI 1 and 4, but maybe we should decide after rebasing so everyone can have a hands on test.
 I've started rebasing, I am nearly finished only have to test and will push tomorrow - also some fixes for the extractor will be pushed.

Happy New Year!!

evermind-zz:
rebased finished and all seems to work properly.

MD77MD:
I vote for UI 4... but i wish selected option could be more highlighted... something like this but in red instead of gold color 

190013629-40df50a5-3f42-4fe1-908d-3845ae177da7

throwaway242685:
I also vote for UI 4.

it's the most intuitive one.

Stypox:
> Would it be okay to convert this PR into an issue for the time being, and tackle it again after the rewrite?

Yes, unfortunately we (as in, the TeamNewPipe members) were unable to work on this properly as it requires a lot of structural changes (especially in the extractor). After the extractor side will be ready, we can implement this again in the rewritten NewPipe, and we will probably be able to do so with much less code on the NewPipe side.

I want to apologize to everyone involved for how this PR was handled. And thank you evermind-zz a lot for the effort you put into this PR, and for developing various possibilities for the UI. The feedback the users gave in this thread will be really valuable once we start working on this again.

Issue #8016, opened 2022-03-12
ExoPlaybackException - MediaCodecVideoDecoderException


Gitogitos:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.22.1

### Steps to reproduce the bug

When I play YouTube, with video rotation sometimes without video rotation 

### Expected behavior

The video should play

### Actual behavior

It's stop with error message 

## Exception
* __User Action:__ play stream
* __Request:__ Player error[type=1] occurred while playing https://www.youtube.com/watch?v=qrIDMfPRF6I
* __Content Country:__ PL
* __Content Language:__ pl-PL
* __App Language:__ pl_PL
* __Service:__ YouTube
* __Version:__ 0.22.0 and 0.22.1
* __OS:__ Linux motorola/athene/athene:7.0/NPJ25.93-14/16:user/release-keys 7.0 - 24
<details><summary><b>Crash log </b></summary><p>

```
com.google.android.exoplayer2.ExoPlaybackException: MediaCodecVideoRenderer error, index=0, format=Format(1, null, null, video/avc, avc1.64001F, -1, null, [1280, 720, 30.0], [-1, -1]), format_supported=YES
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:555)
	at android.os.Handler.dispatchMessage(Handler.java:98)
	at android.os.Looper.loop(Looper.java:154)
	at android.os.HandlerThread.run(HandlerThread.java:61)
Caused by: com.google.android.exoplayer2.video.MediaCodecVideoDecoderException: Decoder failed: OMX.qcom.video.decoder.avc
	at com.google.android.exoplayer2.video.MediaCodecVideoRenderer.createDecoderException(MediaCodecVideoRenderer.java:1410)
[...]

Issue #7599, opened 2021-12-30
v0.21.15 Old regression returned: slow playback on LineageOS


FrostKiwi:
As previously with v0.21.2, updating to v0.21.15 regressed the playback performance to a 10fps slideshow on the OnePlus One with 1080p60 footage.
Previously reported and discussed indepth here with screen recordings and so on: https://github.com/TeamNewPipe/NewPipe/issues/6207

Thus it is most likely in connection to ExoPlayer not properly detecting the hardware decoding features of the OnePlus One and thus falling back to a software decoded slideshow on 1080p60 footage.

I downgraded to v0.21.14 and the issue is gone.

### Checklist

- [x] I am using the latest version - 0.21.15
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Device info

 - Android version/Custom ROM version: Android 11, LineageOS 18.1
 - Device model: OnePlus One (A0001)

Redirion:
connect to your OnePlus One with adb.
"adb shell"
"getprop | grep build"
and paste the output here.
We should be able to reanable the media parser feature particularly for LinageOS as it seems to be working there

FrostKiwi:
Here it is:
<details><summary>Output of greprop | grep build</summary>

```
[build.version.extensions.r]: [0]
[ro.bootimage.build.date]: [Tue Aug 10 00:29:06 UTC 2021]
[ro.bootimage.build.date.utc]: [1628555346]
[ro.bootimage.build.fingerprint]: [oneplus/bacon/A0001:5.0.2/LRX22G/YNG1TAS0YL:user/release-keys]
[ro.build.characteristics]: [default]
[ro.build.date]: [Tue Aug 10 00:29:06 UTC 2021]
[ro.build.date.utc]: [1628555346]
[ro.build.description]: [bacon-user 5.0.2 LRX22G YNG1TAS0YL release-keys]
[ro.build.display.id]: [lineage_bacon-userdebug 11 RQ3A.210805.001.A1 eng.root.20210810.003107 dev-keys]
[ro.build.expect.trustzone]: [TZ.BF.2.0-2.0.0109|TZ.BF.2.0-2.0.0123|TZ.BF.2.0-2.0.0134]
[...]

PR #8875, opened 2022-08-24
Add an ExoPlayer settings page

AudricV:
#### What is it?
- [x] Feature (user facing)

This PR adds an ExoPlayer settings page to NewPipe's settings, at the current place of the playback load interval size setting (`Video and audio` settings page).

This page contains three settings:

- the playback load interval size one, for which its description has been updated to reflect its effect on progressive sources only (its translations have been removed);
- a new setting to enable [ExoPlayer's decoder fallback feature](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/DefaultRenderersFactory.html#setEnableDecoderFallback(boolean)), which could help for issues such as #8016;
- the media tunneling setting, which is now controllable by the user and not only for debug builds. Media tunneling may be not supported by more devices than the ones on which we disabled this feature before, so I think it is a good idea to allow users to not wait for us to disable this setting on their device on a future release, and also for future code complexity purposes. **Users of devices on which media tunneling has been disabled will have to enable manually this setting again on their devices with this change.**

More settings could be added in the future, such as ones to change buffer settings.

**These settings require a player restart to take effect** (this information has been written on the link of this page, in `Video and audio` settings).

- Before:
Video and audio settings before

- After:

| Video and audio settings after | ExoPlayer settings |
|:-:|:-:|

- Fixes no issue, but it could help for #8016 and #9023.

#### APK testing 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -263,7 +263,9 @@ public final class Player implements PlaybackListener, Listener {
         final PlayerDataSource dataSource = new PlayerDataSource(context,
                 new DefaultBandwidthMeter.Builder(context).build());
         loadController = new LoadController();
-        renderFactory = new DefaultRenderersFactory(context);
+        renderFactory = new DefaultRenderersFactory(context)
+                .setEnableDecoderFallback(prefs.getBoolean(
+                        context.getString(R.string.use_exoplayer_decoder_fallback_key), false));
 
         videoResolver = new VideoPlaybackResolver(context, dataSource, getQualityResolver());
         audioResolver = new AudioPlaybackResolver(context, dataSource);
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -69,7 +69,6 @@ import com.google.android.exoplayer2.ExoPlayer;
 import com.google.android.exoplayer2.PlaybackException;
 import com.google.android.exoplayer2.PlaybackParameters;
 import com.google.android.exoplayer2.Player.PositionInfo;
-import com.google.android.exoplayer2.RenderersFactory;
 import com.google.android.exoplayer2.Timeline;
 import com.google.android.exoplayer2.Tracks;
 import com.google.android.exoplayer2.source.MediaSource;
@@ -95,6 +94,7 @@ import org.schabi.newpipe.local.history.HistoryRecordManager;
 import org.schabi.newpipe.player.event.PlayerEventListener;
 import org.schabi.newpipe.player.event.PlayerServiceEventListener;
 import org.schabi.newpipe.player.helper.AudioReactor;
+import org.schabi.newpipe.player.helper.CustomRenderersFactory;
 import org.schabi.newpipe.player.helper.LoadController;
 import org.schabi.newpipe.player.helper.PlayerDataSource;
 import org.schabi.newpipe.player.helper.PlayerHelper;
@@ -196,7 +196,7 @@ public final class Player implements PlaybackListener, Listener {
 
     @NonNull private final DefaultTrackSelector trackSelector;
     @NonNull private final LoadController loadController;
-    @NonNull private final RenderersFactory renderFactory;
+    @NonNull private final DefaultRenderersFactory renderFactory;
 
     @NonNull private final VideoPlaybackResolver videoResolver;
     @NonNull private final AudioPlaybackResolver audioResolver;
@@ -261,9 +261,16 @@ public final class Player implements PlaybackListener, Listener {
         final PlayerDataSource dataSource = new PlayerDataSource(context,
                 new DefaultBandwidthMeter.Builder(context).build());
         loadController = new LoadController();
-        renderFactory = new DefaultRenderersFactory(context)
-                .setEnableDecoderFallback(prefs.getBoolean(
-                        context.getString(R.string.use_exoplayer_decoder_fallback_key), false));
+
+        renderFactory = prefs.getBoolean(
+                context.getString(
+                        R.string.always_use_exoplayer_set_output_surface_workaround_key), false)
+                ? new CustomRenderersFactory(context) : new DefaultRenderersFactory(context);
+
+        renderFactory.setEnableDecoderFallback(
+                prefs.getBoolean(
+                        context.getString(
+                                R.string.use_exoplayer_decoder_fallback_key), false));
 
         videoResolver = new VideoPlaybackResolver(context, dataSource, getQualityResolver());
         audioResolver = new AudioPlaybackResolver(context, dataSource);
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -77,7 +77,6 @@ import com.google.android.exoplayer2.text.CueGroup;
 import com.google.android.exoplayer2.trackselection.DefaultTrackSelector;
 import com.google.android.exoplayer2.trackselection.MappingTrackSelector;
 import com.google.android.exoplayer2.upstream.DefaultBandwidthMeter;
-import com.google.android.exoplayer2.util.Util;
 import com.google.android.exoplayer2.video.VideoSize;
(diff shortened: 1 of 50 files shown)


TacoTheDank:
I'd recommend that, when using multiple sentences for a preference summary, that each sentence end with a period.

E.g. "These changes require a player restart to take effect." Same with the ExoPlayer settings.

AudricV:
> I'd recommend that, when using multiple sentences for a preference summary, that each sentence end with a period.

This behavior is used in the majority of settings description with multiple settings, so I try to kept the same style.

TacoTheDank:
> This behavior is used in the majority of settings description with multiple settings, so I try to kept the same style.

I just checked all the settings, and that does seem to be the case. That's kinda weird IMO, but alright. 👍

glbyers:
I agree that having the ability to disable tunneling from settings in the release build is a good thing, but I also believe a tunneling blacklist should be maintained for those devices that don't support it properly. Otherwise you're just going to end up with frustrated users searching forums and whatnot trying to find out why the app is unusable or constantly crashing for them. Even though that switch was available in the debug build, it wasn't once mentioned on the bug report I was following, and not once did I consider it a potential fix. I had to actually go through the whole debugging process to find it. Maybe a knowledgeable dev would have found that much sooner, but I'm just a lowly sys admin; it took some time. 

Anyway, give users of the application a nice experience from the get go, without the need for having to hunt down why it won't work for them in the first place.

For those with issues that do manage to find the switch, perhaps also adding the internal device name & API version to the about page would help aid in maintaining the blacklist in future? Rather than someone reporting that "disabling tunneling works for me on a Zephir TS43UHD-2", they could instead report that it works for them on a cvt_mt5886_eu_1g with API version 24 for example.

Anyway, just my 2c.

opusforlife2:
> but I also believe a tunneling blacklist should be maintained for those devices that don't support it properly

@glbyers The thing is, media tunneling was created for Android TVs in the first place. So bypassing it on TVs makes no sense at all. o_O

> Anyway, give users of the application a nice experience from the get go, without the need for having to hunt down why it won't work for them in the first place.

That's the idea. Since all these TVs were working fine before 0.24.0, this is a regression, and a fix for this should fix full screen on all TVs, not specific models. You finding out that media tunneling causes this has been a huge help, so thank you for debugging. ^_^

glbyers:
> @glbyers The thing is, media tunneling was created for Android TVs in the first place. So bypassing it on TVs makes no sense at all. o_O

There are a significant number of issues both open and closed in the exoplayer github related to media tunneling. I've seen some of the devs suggest that it's a risky thing to enable by default as it has a high number of device specific problems. I get how valuable it is, and sweet if it works, especially at 4k resolutions. But it's kinda reliant on every device manufacturer that claims to support it to implement it properly. Even with my TV, it's not that the codecs don't support media tunneling, it's more that the audio codec doesn't support changing of the surface after it has been started with tunneling enabled.

I'm not at all arguing against a more general fix. I'm not sure an obscure option is necessarily a general fix though, especially if it defaults to on. It should still default to off for devices it's known not to work with in my mind. Perhaps it needs some clarity so it's obvious what it does, not what it's called. "Hardware decoding" is instantly more recognisable than '"media tunneling" IMO.

FWIW, as I said in the issue. It wasn't working properly prior to 0.24 either. At least for my TV. It's just that having to relaunch it between every video was still a usable solution.

Anyway, you know what they say. Opinions are like something or other. Everyone has one.

opusforlife2:
> I'm not sure an obscure option is necessarily a general fix though

@glbyers Oh, this PR is not the intended fix. It just happens to contain the workaround you found already. We've asked users in that thread to test it so that we can be sure that everyone's problem was the same and is indeed solved this way.

Don't worry. We'll try to look for a proper fix that doesn't need users to mess around with decoding settings.

hassaanarif:
I am facing the same issue as mentioned above in the comments. 1080p 60fps videos are extremely stuttery though 7200 60fps works fine.

I was just googling the issue and found this page. Seems like adding this new settings page solves the issue but I don't this exo player settings page added in 0.24.1. Any idea how can I fix this stuttery video issue and how can I get this exo player setting page ? 

Device: Galaxy Note 3
Custom Rom : Resurrection Remix (Android 10)

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -35,22 +35,6 @@ public final class DeviceUtils {
...
-            && Build.DEVICE.equals("Hi3798MV200");
-    // Zephir TS43UHD-2
-    private static final boolean CVT_MT5886_EU_1G = Build.VERSION.SDK_INT == 24
-            && Build.DEVICE.equals("cvt_mt5886_eu_1g");
-    // Hilife TV
-    private static final boolean REALTEKATV = Build.VERSION.SDK_INT == 25
-            && Build.DEVICE.equals("RealtekATV");
-    // Philips QM16XE
-    private static final boolean QM16XE_U = Build.VERSION.SDK_INT == 23
-            && Build.DEVICE.equals("QM16XE_U");
```
Stypox, line 52: Mmmh, why are you removing this list completely? We know all these TVs don't work, so by default on such devices tunneling should be disabled. Then the user can go into settings and enable it, but at least it should be off by default.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -521,16 +526,11 @@ private void initPlayer(final boolean playOnReady) {
         // Setup UIs
         UIs.call(PlayerUi::initPlayer);
 
-        // enable media tunneling
-        if (DEBUG && PreferenceManager.getDefaultSharedPreferences(context)
+        // Disable media tunneling if requested by the user from ExoPlayer settings
+        if (!PreferenceManager.getDefaultSharedPreferences(context)
                 .getBoolean(context.getString(R.string.disable_media_tunneling_key), false)) {
```
Stypox, line 531: So I would make sure the preference does not get auto-created, and then use `!DeviceUtils.shouldSupportMediaTunneling()` instead of `false` as the value to use when the preference is not set.

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/player/helper/OldMediaCodecSelector.java
```diff
@@ -0,0 +1,33 @@
...
+import androidx.annotation.NonNull;
+
+import com.google.android.exoplayer2.mediacodec.MediaCodecInfo;
+import com.google.android.exoplayer2.mediacodec.MediaCodecSelector;
+import com.google.android.exoplayer2.mediacodec.MediaCodecUtil;
+
+import java.util.Collections;
+import java.util.List;
+
+public final class OldMediaCodecSelector implements MediaCodecSelector {
```
Stypox, line 12: Maybe add a javadoc for this class, possibly with references to an exoplayer issue

Stypox (changes requested):
Maybe `use_inexact_seek_key` should also go into the exoplayer settings?

Do we also want to add an option to enable the `setSurface` workaround manually?

Stypox:
Well, it's a setting that changes how exoplayer seeks, so it's more a tecnichal setting. But yeah, it's arguable whether that setting should be under the exoplayer menu or not.

app/src/main/java/org/schabi/newpipe/player/helper/OldMediaCodecSelector.java
AudricV: Not applicable anymore, due to our decision to remove this setting entirely (see https://github.com/TeamNewPipe/NewPipe/issues/7599#issuecomment-1453663276).

AudricV:
> Do we also want to add an option to enable the setSurface workaround manually?

Done.

twasjustadream:
Hey could you please rebase as disabling media tunneling makes tvs work in fullscreen, thanks for the commit.

Stypox (approved):

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
TobiGr: Someone needs to write a settings migration for that. Otherwise, NewPipe is going to stop working for some users and they certainly not search for a solution in the settings.

PR #8881, opened 2022-08-25
Updated AGP and other dependepcies.

xareal:
Updated mockitoVersion, desugar_jdk, appcompat, fragment_ktx, preference, kotlin and AGP

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -109,7 +109,7 @@ ext {
 
     leakCanaryVersion = '2.5'
     stethoVersion = '1.6.0'
-    mockitoVersion = '4.0.0'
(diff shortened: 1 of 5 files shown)


TacoTheDank:
This tries to update too many big things at once. Fragment needs to have deprecations fixed, we don't need preference-ktx, AppCompat already has a PR made for it, we cannot update desugar until AGP 7.3.x, Kotlin may need to be tested separately, etc.

AudricV:
Closing due to the reasons stated above and due to inactivity.

AudricV:
Closing due to the reasons stated above and due to inactivity.

PR #8883, opened 2022-08-25
Only show "Enqueue next" when in the middle of the queue

Douile:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Add a check that the queue position is not the last in the queue before
showing "Enqueue next".

Previously the "Enqueue next" action would always be shown if the queue
length was greater than one, this meant even if you were at the end of
the queue (when "Enqueue" would have the same effect as "Enqueue next")
the action would still be shown.

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/dialog/InfoItemDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/dialog/InfoItemDialog.java
@@ -252,10 +252,11 @@ public final class InfoItemDialog {
          * @return the current {@link Builder} instance
          */
         public Builder addEnqueueEntriesIfNeeded() {
-            if (PlayerHolder.getInstance().isPlayQueueReady()) {
+            final PlayerHolder holder = PlayerHolder.getInstance();
+            if (holder.isPlayQueueReady()) {
                 addEntry(StreamDialogDefaultEntry.ENQUEUE);
 
-                if (PlayerHolder.getInstance().getQueueSize() > 1) {
+                if (holder.getQueuePosition() < holder.getQueueSize() - 1) {
                     addEntry(StreamDialogDefaultEntry.ENQUEUE_NEXT);
                 }
             }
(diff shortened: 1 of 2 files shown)


Master2050:
Well it will degrade newpipe experience for some users as when you have a long queue and you want to play something after current video is played ensue next is used else for making a long queue you should use 'add to queue' button

Douile:
I'm confused how this will degrade experience, it will only hide the enqueue next button when it has the same effect as the enqueue button: when you are in watching the last video in the queue

Master2050:
enqueue and enqueue next doesn't have the same effect. Let's take an example - consider a queue of 5 items namely A, B, C, D, E now clicking on enqueue button will add a sixth element F at the sixth position in queue whereas using enqueue next button it will add a new element in between A and B the final list will be A, F, B, C, D, E

Douile:
This patch only hides the enqueue next button when you are currently playing video E, therefore whichever button you press there is only 1 place for F to go: the sixth position.

app/src/main/java/org/schabi/newpipe/info_list/dialog/InfoItemDialog.java
```diff
@@ -252,10 +252,12 @@ public Builder setAction(@NonNull final StreamDialogDefaultEntry entry,
...
          */
         public Builder addEnqueueEntriesIfNeeded() {
-            if (PlayerHolder.getInstance().isPlayQueueReady()) {
+            final PlayerHolder holder = PlayerHolder.getInstance();
+            if (holder.isPlayQueueReady()) {
                 addEntry(StreamDialogDefaultEntry.ENQUEUE);
 
-                if (PlayerHolder.getInstance().getQueueSize() > 1) {
+                final int size = holder.getQueueSize();
+                if (size > 1 && holder.getQueuePosition() < size - 1) {
```
Stypox, line 260: I think you don't need the `size > 1` condition. If `size <= 1`, then `size - 1 <= 0`, and since `queuePosition` is always `>= 0`, it is never `< size - 1 <= 0`.
```suggestion
                if (holder.getQueuePosition() < holder.getQueueSize() - 1) {
```

Stypox (changes requested):

Stypox (approved):
Thank you! I tested and it works well

Issue #8890, opened 2022-08-27
Release v0.24.0 (please TEST!)


Stypox:
## Testing for regressions
Debug APK (built by our CI in #8889): [NewPipe_v0.24.0_RC3_debug.zip](https://github.com/TeamNewPipe/NewPipe/files/9597719/NewPipe_v0.24.0_RC3_debug.zip)
Release APK (built and signed by @Stypox): [NewPipe_v0.24.0_RC3_release.zip](https://github.com/TeamNewPipe/NewPipe/files/9597686/NewPipe_v0.24.0_RC3_release.zip)
Please report **only regressions** (i.e. new issues) here, not issues that were already present in previous releases!

## TODO

- [x] Merge weblate
- [x] #8899
- [x] TeamNewPipe/NewPipeExtractor#920
- [x] TeamNewPipe/NewPipeExtractor#925
- [x] TeamNewPipe/NewPipeExtractor#926
- [x] Thumbnails in the feed tab are incorrectly resized https://github.com/TeamNewPipe/NewPipe/issues/8890#issuecomment-1236708859 https://github.com/TeamNewPipe/NewPipeExtractor/pull/927
- [x] Strange `Show future videos` icon behavior https://github.com/TeamNewPipe/NewPipe/issues/8890#issuecomment-1230466338 #8966
- [ ] Crash when closing mini player before it loads https://github.com/TeamNewPipe/NewPipe/issues/8890#issuecomment-1242885157 https://github.com/TeamNewPipe/NewPipe/issues/8646#issuecomment-1236162209
- [x] Search suggestions ListAdapter issues #8972

## NewPipeExtractor version

This version of NewPipe will ship with NewPipe Extractor commit https://github.com/TeamNewPipe/NewPipeExtractor/commit/5c710da160f488bb40ab2cf4469bec9bd4cefd38.
-> we are going to change the exposed extractor API again soon, so no official release will be done until then ;-)

<details><summary><h2>App changelog </h2></summary><p>

### New
- :warning: Bump `minSdk` to 21 (Android 5 Lollipop) and **drop support for Android 4.4 KitKat** #7613
- Add Download entry to long-press menu #8397 
- Add option to hide future videos in feed #8545
- Add Share menu item to local playlists #8542

### Improved
- Crop the notification thumbnail in 1:1 mode instead of stretching it #8533
- Make thumbnails' `scaleType` `fitCenter` #8573 
- Improve image placeholders #8530
[...]

PR #8899, opened 2022-08-28
Fix wrong thumbnail in notification on Android 13

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This PR fixes yet another issue about the notification thumbnail not changing when going to the previous or next item in the player queue (see https://github.com/TeamNewPipe/NewPipe/issues/8890#issuecomment-1229464123).

What I think the problem was (though this might need more investigation) is this:
1. when tapping on "next", NewPipe's queue implementation would go to the next item
2. then the ExoPlayer's internal media item queue gets synchronized, by telling the player to go to the next item
3. the player internally switches to the next item, and it starts loading its stream details to be able to play it, also updating the item index in the media session (a component that the system can read to get info)
4. the player switches state to `onBuffering`, which triggers a notification update for NewPipe
5. notification updates are implemented not based on the selected queue item (`currentItem`) but rather on the StreamInfo (`currentMetadata`) and the `currentThumbnail` stored in the player. That StreamInfo, though, still belongs to the previous item in the queue, since stream details have not finished loading yet. This causes the wrong title, uploader and thumbnail to be temporarily set in the notification.
6. the stream details finally load, they are set in the player and the notification is updated again, this time with the correct information
7. title and uploader are correctly updated in the notification, but the thumbnail is not. This happens probably because the system associated the thumbnail of the previous item to the index of the next, and then tries to avoid reloading the thumbnail again to save cpu, assuming the first set thumbnail was the correct one. Triggering notification updates in other ways (e.g. play-pause) does not make a difference.

So in the PR that will merge `NotificationUtil` and `NotificationPlayerUi`, the notification, the player and the media session should be switched to using data not about the player StreamInfo, but rather the actual current play queue item (except for the thumbnail, for which we need the high-res one from the fully-loaded stream details). Though it might not be this simple.

This PR fixes the problem by detecting when the queue item changes and then setting `currentThumbnail` to `null`. This happens in-between points 1 and 2 above, i.e. before the exoPlayer internal queue is synced and before the notification update is triggered.

The commits strictly necessary to solve the issue are the first two, though the other two are also important in my opinion.

- https://github.com/TeamNewPipe/NewPipe/issues/8890#issuecomment-1229464123

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -1509,48 +1509,50 @@ public final class Player implements PlaybackListener, Listener {
             Log.d(TAG, "Playback - onPlaybackSynchronize(was blocked: " + wasBlocked
                     + ") called with item=[" + item.getTitle() + "], url=[" + item.getUrl() + "]");
         }
-        if (exoPlayerIsNull() || playQueue == null) {
-            return;
+        if (exoPlayerIsNull() || playQueue == null || currentItem == item) {
+            return; // nothing to synchronize
         }
 
-        final boolean hasPlayQueueItemChanged = currentItem != item;
+        final int playQueueIndex = playQueue.indexOf(item);
+        final int playlistIndex = simpleExoPlayer.getCurrentMediaItemIndex();
+        final int playlistSize = simpleExoPlayer.getCurrentTimeline().getWindowCount();
+        final boolean removeThumbnailBeforeSync = currentItem == null
+                || currentItem.getServiceId() != item.getServiceId()
+                || !currentItem.getUrl().equals(item.getUrl());
 
-        final int currentPlayQueueIndex = playQueue.indexOf(item);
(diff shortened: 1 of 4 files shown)


Stypox:
@karyogamy could you explain why the `!isPlaying()` is needed on [this line](https://github.com/TeamNewPipe/NewPipe/blob/2572a8c1a977dec7e5f172f1009ea86921dacdd9/app/src/main/java/org/schabi/newpipe/player/Player.java#L1537)? You initially introduced it in 1d136c6c353c5ff7206812acbe377f347f48df46 (#810), then accidentally removed it in 4e459b33839e428c26c5730d629ba07356ca88e3 and readded it in b81eb35f3d045e9bbaa77e8fcfc13ab1fb43cd09 (#8020).

opusforlife2:
Has the reddit reporter been asked to test this?

AudricV:
Yes, and it was claimed to be fixed by the user but their comments have been now deleted.

PR #8986, opened 2022-09-17
Update Android Gradle Plugin to 7.3.0.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Update the Android Gradle Plugin to 7.3.0.

- Fixes #

#### APK testing

#### Due diligence


AudricV (approved):
Thank you, this will be merged after 0.24.0 release, to ensure that most of developers updated Android Studio to the latest version :)

app/build.gradle
```diff
@@ -10,6 +10,7 @@ plugins {
 android {
     compileSdk 31
     buildToolsVersion '31.0.0'
```
Redirion, line 12: I think this line can be removed. If we just keep the Gradle Plugin updated.

https://developer.android.com/studio/releases/build-tools

"If you're using [Android plugin for Gradle 3.0.0](https://developer.android.com/studio/releases/gradle-plugin#3-0-0) or higher, your project automatically uses a default version of the build tools that the plugin specifies."

This way we make sure that the Plugin uses the recommended/tested/default version.

TobiGr: We should nevertheless specify it to ensure the usage of the same version when building NewPipe by hand. Otherwise, F-Droid might make problems resulting in non-reproducible builds.

TobiGr (approved):

Issue #7644, opened 2022-01-11
Option to Turn off or hide YouTube shorts


sahlers83:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### What feature do you want?

Add a filter in "Settings -> Content " with title "Include YouTube Shorts" with description "Turn off to hide YouTube shorts".  (Lets say one can place this option after Load Thumbnails.

#### Why do you want this feature?

1. Subscribe to a channel
2. Go to Hamburger menu -> Subscriptions -> Open the channel

Some channels are just having too many "shorts", and the user can avoid it.

Note that this feature is NOT connected to "Whats New" option. In the list of Whats New, YouTube shorts vides DO NOT appear

SameenAhnaf:
I highly dislike Shorts and find disturbing when mixed with usual content. YouTube app shows Shorts both in row and other videos. However, Newpipe should exclusively show them in a row which could be optionally hidden just like clicking on arrow of YouTube video description.

eladkarako:
`/shorts/` links will as of start of April 2022 get redirects (`HTTP 303 See Other`)  to back to `/watch?v=` normal video. you can still handle "short" videos (under 2 minutes).  

I'm applying <a href="https://gist.github.com/eladkarako/2549ac9cdfd7eae43a957404ec5e339b">a uBlock Origin rule</a> with `:upward` for DOM-traversing (avoiding APIs), 
but you shouldn't bother with that. You pretty much have all the video details already so you can introduce hiding to "short" video based on length (in seconds).

Magneticdud:
A filter to hide shorts is becoming an essential feature 

More and more channels are having too many shorts, very irritating

atul1503:
How to see shorts? Is there any exclusive button for shorts in NP?

atul1503:
Can channel groups be configured to only show shorts from the channels which are part of it?

Magneticdud:
At the moment it's not possible. The purpose of this proposal is to **hide** shorts from the channels, though, not to make them prominent

terminal-root:
I don't know much about the specs behind shorts, but I imagine one way to do it would be to filter by aspect ratio.

[...]

PR #9049, opened 2022-10-01
Add option to hide YouTube Shorts

chowder:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Adds an options to hide YouTube Shorts from the feed. 

- Before:

- After:

- https://github.com/TeamNewPipe/NewPipe/issues/7644

#### APK testing

#### Due diligence


app/src/main/java/org/schabi/newpipe/local/feed/FeedViewModel.kt
```diff
@@ -54,22 +60,24 @@ class FeedViewModel(
...
         .throttleLatest(DEFAULT_THROTTLE_TIMEOUT, TimeUnit.MILLISECONDS)
         .subscribeOn(Schedulers.io())
         .observeOn(Schedulers.io())
-        .map { (event, showPlayedItems, showFutureItems, notLoadedCount, oldestUpdate) ->
+        .map { (event, showPlayedItems, showFutureItems, showShorts, notLoadedCount, oldestUpdate) ->
             val streamItems = if (event is SuccessResultEvent || event is IdleEvent)
                 feedDatabaseManager
                     .getStreams(groupId, showPlayedItems, showFutureItems)
                     .blockingGet(arrayListOf())
+                    .filter { s -> showShorts || s.stream.duration > 0 }
```
TobiGr, line 80: Please explain this filter. Do shorts have an unknown duration (i.e. -1))? 

What about other services? When there is an unknown duration for e.g. a media.ccc.de or PeerTube stream, they are removed by your filter, too. I suggest to only remove streams with the YouTube service id.

AudricV:
This is a non-viable solution, as it is planned in the future to extract shorts duration from channels using the accessibility data.

Note: YouTube will roll out in the future dedicated shorts and live tabs, as you can see in [this video](https://www.youtube.com/watch?v=czIyqEC4V-s&t=12s).

chowder:
Thanks for the comments, both.

I spent some more time on this, and struggled to find a way to elegantly implement this, so I raised [this PR](https://github.com/TeamNewPipe/NewPipeExtractor/pull/946) in `NewPipeExtractor` to be able to identify streams that are short-form content (i.e. content in the style of TikTok, YouTube Shorts, or Instagram Moment videos), that this PR is contingent on. 

I'd like to keep this PR open in the meantime to provide some context for the `NewPipeExtractor` PR, and for any tangential discussions, e.g. 

- whether such setting should be applied globally (channel feeds, trending feeds), or only to subscription feeds
- whether such toggle would be better placed in the settings view
- a more appropriate icon to use for the toggle

Douile:
It would be nice if this was a unified option to filter types of content: a filter content menu where you can tick on/off
- regular video
- shorts
- live streams
- possibly future videos (premiers), although this would duplicate the show future content button in feed
- other types of content from other services

opusforlife2:
> although this would duplicate the show future content button in feed

@Douile This is the exact button that would need to be changed into a content filter menu. We shouldn't have the same functionality in different buttons/menus.

> other types of content from other services

This raises an important point: the filter should be service-specific. So when you change the service from the hamburger menu, the filter menu should change accordingly. Also, it should remember the last set of selections for that service.

> whether such setting should be applied globally 

@chowder How about only feeds for now?

> whether such toggle would be better placed in the settings view

As mentioned above, the "hide premieres" button could be changed into a menu with check boxes. Probably, the last ticked item in the check box shouldn't be untick-able.

> a more appropriate icon to use for the toggle

A filter icon? Is that available?

Honestly, I would prefer the eye icon, which is unfortunately taken up by the "hide watched videos" button. But I don't know if we could find another icon for that button, either. Something like a play icon with a tick at the bottom.

Or maybe the "watched videos" button could become just one more check box in the filter menu itself, though visually separated from the content-type check boxes. After all, the functionality is the same - hiding certain videos based on a condition. In that case, the eye icon could be used for this single unified drop-down.

opusforlife2:
Question: should the filter be feed-specific as well? e.g. you might only want regular videos from a news channel, but livestreams from a gaming channel. And this preference is unlikely to change much. So someone who frequently switches between feeds would have to keep adjusting the filters, which seems inconvenient.

Stypox:
Thank you!
I think there are some database-related issues, since the 5.json schema shouldn't change. Could you revert changes in that file?
Also, at the moment there does not seem to be any code that actually sets the "is short content" value to stream entities, and so that field is always set to false by default. You will want to use https://github.com/TeamNewPipe/NewPipeExtractor/pull/946 for this :-)

chowder:
Thanks for reviewing, @Stypox! 

Indeed NewPipeExtractor doesn't support the new channel tabs layout at the moment; as such, shorts are currently not being extracted, making this PR somewhat pointless (for now). 

I'll mark this as a draft until https://github.com/TeamNewPipe/NewPipeExtractor/pull/951 makes further progress.

magicweirdoo:
I don't know the code of the NewPipe Extractor but there is already a shorts filter in youtube. If you click on a channels videos page only long form content will be displayed, no shorts.

WhatsNew:


Channel View:


So if there is a video shorter than 1min and not in the channel video view it won't be displayed. This would be the filter.

As I said I didn't look into the code but it's just a thought that came to me and I had to share it.

PR #9098, opened 2022-10-10
Fix Open action menu not working on cold start

### LINKED ISSUE #3129: Share to newpipe menu does not open at first click on link
- [no] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [no ] I checked if the issue/feature exists in the latest version.
- [ dont know what is] I did use the [incredible 

Im not a pro user who knows much about bug report and etc.

A problem.

As I close every app as "kill foreground app", whenever I try to open a utube video with newpipe again, the video doesnt come the first time. the forst attempt, the app shows up on the default page, what ever it is, then u have to click on the link again to see the video in newpipe.
I checked this with official app and its not like this.
Can u fix this that even if the app is closed from the base, when it opens with a link, to go directly to the link and not to the main page?

I play videos, By searching a video in google and clicking on a vid coming up in search results. Then it asks me to open woth chrome or newpipe, I choose newpipe.

This way, first the app starts without showing the content I want and I have to click on the link again and it shows the vid as the app has started now.
This bothers a little.

Another thing is that the videos dont load sometimes and the rest of the time, load slower than other apps.

But still, its better than vanced or snaptube or original app. thanks.

devlearner:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

#### Description

So I think I hit a bug (similar to #3129).  Sorry if it seems like a rolled up bug report + pull request, but I'll just set out the steps to reproduce below, for the sake of clarity/testing:
##### Preparation
- Have the 0.24.0 release apk installed
- Ensure that `Settings` -> `Video & audio` -> `Preferred 'open' action` is set to `Always ask` (the default setting)
- Change the theme and/or app language to something _different from_ the system default
(say, my phone is on light, I choose the dark theme)
##### Steps to reproduce
- Force stop the app (in system `Settings` -> `Apps`)
- Share a YouTube video link to NewPipe (eg from a browser)
##### Observation
- (if you've set the dark app theme like me, you might see the background dims for a brief moment)
- Nothing happens (the Open action selection dialog does not pop up as expected)
- If you try again (by sharing/clicking on the link the second time), this time `RouterActivity` dialog appears as expected

[Repeat the above with the debug apk to see if this fixes the issue ie the dialog should appear as expected the first time (on cold start)]

#### Hypothesis and PR changes
- It's a bit tricky to identify since there's no error message and the selection dialog just seems to be silently dropped (but only) the first time (on cold start).
- A hypothesis (my suspicion) is that the theme/locale setup in `onCreate()` caused the `RouterActivity` to recreate.  Thus I move up those lines to before `super.onCreate()` (for this I'm indebted to the generous answers on StackOverflow) and voila, it seems issue solved.
- The second commit addresses the balance of the issue by avoiding (accidentally) calling `finish()` in `onDismissListener()` (which will be triggered in `onStop()` when cleaning up the dialog) to allow the activity to be recreated on screen rotation.

- (Potentially) fixes #3129
(sorry for pinging if it doesn't solve the exact same issue; but I tried)

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -114,6 +114,11 @@ public class RouterActivity extends AppCompatActivity {
 
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
+        ThemeHelper.setDayNightMode(this);
+        setTheme(ThemeHelper.isLightThemeSelected(this)
+                ? R.style.RouterActivityThemeLight : R.style.RouterActivityThemeDark);
+        Localization.assureCorrectAppLanguage(this);
+
         super.onCreate(savedInstanceState);
         Icepick.restoreInstanceState(this, savedInstanceState);
 
@@ -125,11 +130,6 @@ public class RouterActivity extends AppCompatActivity {
                 finish();
             }
         }
-
-        ThemeHelper.setDayNightMode(this);
-        setTheme(ThemeHelper.isLightThemeSelected(this)
-                ? R.style.RouterActivityThemeLight : R.style.RouterActivityThemeDark);
-        Localization.assureCorrectAppLanguage(this);
     }
 
     @Override
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -397,7 +397,8 @@ public class RouterActivity extends AppCompatActivity {
                 .setNegativeButton(R.string.just_once, dialogButtonsClickListener)
                 .setPositiveButton(R.string.always, dialogButtonsClickListener)
                 .setOnDismissListener(dialog -> {
-                    if (!selectionIsDownload && !selectionIsAddToPlaylist) {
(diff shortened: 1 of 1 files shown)


opusforlife2:
What happened?

devlearner:
Hey, thanks for your message.
I reproduce the bug on my Android 8, but somehow (after I made the PR) couldn't reproduce this on an Android 11 (both Samsung) phone, so I realized most of us have probably 'moved on', both in terms of a more modern Android, and perhaps also for the OP two years on.
The 'fix' is actually a simple one, just moving up three lines.  I still think it's valid (and good practice) but the second half (on `finish()`) probably needs some more love.
So closing this in favour of a Take 2 (which I hope) to address the screen rotation thing in general (across Android versions old and new) in `RouterActivity` (and I'll slip in the 'fix' for my own benefit 😅).
Saying has it 'orientation change is a beast'; I think I've got a local copy working, but I'll just spend a couple more days to tame it before Take 2 is out.
Sorry for the noise.  You're welcome to test my Take 2 which hopefully will be up for review this weekend 🙂

PR #9135, opened 2022-10-15
Improve screen rotation handling in Open action menu

### LINKED ISSUE #3129: Share to newpipe menu does not open at first click on link

devlearner:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Currently when sharing to NewPipe (when set to `Always ask`), the Open action menu dialog would rotate away on orientation change (since `finish()` would be called by onDismissListener during dialog cleanup thus preventing `RouterActivity` to be recreated when screen rotates).  This PR tries to improve that by making the dialogs stay across screen rotations.

~~_Known issue:_ Ongoing network requests, however, won't survive config change (which would somehow require keeping track of the `Observable`s across recreation of activity).  If user selects Add to Playlist or Download and rotates away while stream info is being fetched, the network request underway would be dropped and `RouterActivity` would start over again when recreated on orientation change.  That would (unfortunately) be another PR for another day.~~ Addressed this using a retained fragment the other day

##### Misc changes
- Fixes a bug on older Android where the Open action dialog won't appear the first time on cold start
- Shows on Download action as well the toast that asks the user to wait
- When fetching stream info (for Add to Playlist and Download), the transparent activity no longer blocks touch events and inadvertently locks UI thus giving the impression that NewPipe freezes

##### To test:
- Make sure `Settings` -> `Video & audio` -> `Preferred 'open' action` is set to `Always ask` (the default setting)
- Share a (YouTube) URL to NewPipe (Debug apk) (say, from a browser)
- Happy screen rotating :)

- Fixes #3129

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/RouterActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/RouterActivity.java
@@ -32,6 +32,7 @@ import androidx.appcompat.content.res.AppCompatResources;
 import androidx.core.app.NotificationCompat;
 import androidx.core.app.ServiceCompat;
 import androidx.core.math.MathUtils;
+import androidx.fragment.app.DialogFragment;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
 import androidx.preference.PreferenceManager;
@@ -113,6 +114,7 @@ public class RouterActivity extends AppCompatActivity {
     private boolean selectionIsDownload = false;
     private boolean selectionIsAddToPlaylist = false;
     private AlertDialog alertDialogChoice = null;
+    private FragmentManager.FragmentLifecycleCallbacks dismissListener = null;
 
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
@@ -141,6 +143,27 @@ public class RouterActivity extends AppCompatActivity {
 
         super.onCreate(savedInstanceState);
         Icepick.restoreInstanceState(this, savedInstanceState);
+
+        // FragmentManager will take care to recreate (Playlist|Download)Dialog when screen rotates
+        // We used to .setOnDismissListener(dialog -> finish()); when creating these DialogFragments
+        // but those callbacks won't survive a config change
+        // Try an alternate approach to hook into FragmentManager instead, to that effect
+        // (ref: https://stackoverflow.com/a/44028453)
+        final FragmentManager fm = getSupportFragmentManager();
+        if (dismissListener == null) {
+            dismissListener = new FragmentManager.FragmentLifecycleCallbacks() {
+                @Override
+                public void onFragmentDestroyed(@NonNull final FragmentManager fm,
+                                                @NonNull final Fragment f) {
+                    super.onFragmentDestroyed(fm, f);
+                    if (f instanceof DialogFragment && fm.getFragments().isEmpty()) {
+                        // No more DialogFragments, we're done
+                        finish();
+                    }
+                }
+            };
+        }
+        fm.registerFragmentLifecycleCallbacks(dismissListener, false);
 
         if (TextUtils.isEmpty(currentUrl)) {
             currentUrl = getUrl(getIntent());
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -114,6 +116,19 @@ public class RouterActivity extends AppCompatActivity {
 
     @Override
     protected void onCreate(final Bundle savedInstanceState) {
+        ThemeHelper.setDayNightMode(this);
+        setTheme(ThemeHelper.isLightThemeSelected(this)
+                ? R.style.RouterActivityThemeLight : R.style.RouterActivityThemeDark);
+        Localization.assureCorrectAppLanguage(this);
+
+        // Pass-through touch events to background activities
+        // so that our transparent window won't lock UI in the mean time
```
Stypox, line 125: Note that this does not work on Android 13 / API 33, but it does work e.g. on API 22, since recent Android versions made some changes regarding transparency. Maybe add a comment.

Stypox (approved):
Code looks good to me. I tested and it seems to work well. Thanks!

app/src/main/java/org/schabi/newpipe/RouterActivity.java
```diff
@@ -111,12 +117,57 @@ public class RouterActivity extends AppCompatActivity {
...
+
+        // Android never fails to impress us with a list of new restrictions per API.
+        // Starting with S (Android 12) one of the prerequisite conditions has to be met
+        // before the FLAG_NOT_TOUCHABLE flag is allowed to kick in:
+        // @see WindowManager.LayoutParams#FLAG_NOT_TOUCHABLE
+        // For our present purpose it seems we can just set LayoutParams.alpha to 0
+        // on the strength of "4. Fully transparent windows" without affecting the scrim of dialogs
+        final WindowManager.LayoutParams params = getWindow().getAttributes();
+        params.alpha = 0f;
+        getWindow().setAttributes(params);
```
devlearner, line 145: Thanks both for the heads up and reference!  I've addressed it here, by simply setting LayoutParams.alpha to 0 to check off point 4 in their list.  Do you think I'm getting it right? 🤔
From my testing (at least on my device) this won't have a side-effect on the scrim (which, in Material speak, refers to the dimmed background) behind AlertDialog/DialogFragments: the background will still be dimmed instead of fully transparent even with this put in, which is good for our purpose

devlearner:
(Oops, I messed up a bit while fast-forwarding the branch; the first five commits are the old commits from last time, sorry for the force-push noise.)

Thanks both for the review and reference!  I've updated it above 👍 

So while we're at it, I've also taken the opportunity to revisit the last part of the issue, and tried to address it utilizing a retained fragment.  It should now get everything covered from start to end for the duration and purpose of `RouterActivity` any potential orientation changes without losing network requests in flight in the interim.  (The only case uncovered for the time being is if the user switches to another app during the waiting window, but I guess it'd be as good as it can get for now :) )

Stypox (approved):
Thank you! I tested and it seems to work well. The only inconsistency I noticed is that the initial selection menu stays shown even when the download dialog is showing above it, except after rotating, but this is not really a problem. I pushed a small commit that uses `Optional` for the activity context and simplifies a couple of other things. Thank you!

Issue #9038, opened 2022-09-27
[YouTube] 403 HTTP error when trying to play some videos


coding-dynasty:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.24.0

### Steps to reproduce the bug

I clicked on a random video

## Exception
* __User Action:__ play stream
* __Request:__ Player error[type=ERROR_CODE_IO_BAD_HTTP_STATUS] occurred while playing https://www.youtube.com/watch?v=6NqK9uS3wxA
* __Content Country:__ GB
* __Content Language:__ en-GB
* __App Language:__ en_GB
* __Service:__ YouTube
* __Version:__ 0.24.0
* __OS:__ Linux Android 10 - 29
<details><summary><b>Crash log </b></summary><p>

```
com.google.android.exoplayer2.ExoPlaybackException: Source error
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleIoException(ExoPlayerImplInternal.java:632)
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:604)
	at android.os.Handler.dispatchMessage(Handler.java:103)
	at android.os.Looper.loop(Looper.java:237)
	at android.os.HandlerThread.run(HandlerThread.java:67)
Caused by: com.google.android.exoplayer2.upstream.HttpDataSource$InvalidResponseCodeException: Response code: 403
	at org.schabi.newpipe.player.datasource.YoutubeHttpDataSource.open(YoutubeHttpDataSource.java:422)
	at com.google.android.exoplayer2.upstream.DefaultDataSource.open(DefaultDataSource.java:258)
[...]

Issue #8933, opened 2022-09-05
Loop video in background crashes


ninja-comprised-given:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.23.3

### Steps to reproduce the bug

1. Start a lomg video (~1.5 hour)
2. Switch to background play
3. From phones notification drop down set it to loop indefinitely.
4. After some time, the play stops and gives below error.

### Expected behavior

The looping background video should continue to loop indefinitely.

### Actual behavior

The video stops and newpipe gives error.

## Exception
* __User Action:__ play stream
* __Request:__ Player error[type=ERROR_CODE_IO_BAD_HTTP_STATUS] occurred while playing https://www.youtube.com/watch?v=Mk1NaOjVSBs
* __Content Country:__ US
* __Content Language:__ en-US
* __App Language:__ en_US
* __Service:__ YouTube
* __Version:__ 0.23.3
* __OS:__ Linux Android 11 - 30
<details><summary><b>Crash log </b></summary><p>

```
com.google.android.exoplayer2.ExoPlaybackException: Source error
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleIoException(ExoPlayerImplInternal.java:641)
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:613)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:223)
	at android.os.HandlerThread.run(HandlerThread.java:67)
Caused by: com.google.android.exoplayer2.upstream.HttpDataSource$InvalidResponseCodeException: Response code: 403
[...]

Issue #8627, opened 2022-07-14
Can't resume YouTube videos after not playing for a long time


timsu92:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.23.1

### Steps to reproduce the bug

1. Goto a channel in YouTube
2. Press `background` to play in the background
3. Shuffle the playlist using the notification
4. ...Listen to music...
5. Pause and wait for a long time
6. Play it

### Expected behavior

Continue fluently or try to load the video/audio again just like a click on the video, but start from the current time

### Actual behavior

HTTP 403 and play the next video.
This behavior is new in this new version.
Before, it stucks somewhat like #5730, except I didn't face what that user said:
> 8. Try to go back to previously paused video via History, Playlists tab, etc (problem is here again)

Details below.

### Logs

HTTP error 403 happened after a long time not listening to YouTube (eg. a good night's sleep😴; a nap isn't suffice)
Before updating to 0.23.1, it used to behave somewhat like #5730. 
## Exception
* __User Action:__ play stream
* __Request:__ Player error[type=ERROR_CODE_IO_BAD_HTTP_STATUS] occurred while playing https://www.youtube.com/watch?v=mnvsE7UtFVA
* __Content Country:__ TW
* __Content Language:__ zh-TW
* __App Language:__ zh_TW_#Hant
* __Service:__ YouTube
* __Version:__ 0.23.1
* __OS:__ Linux Android 12 - 31
<details><summary><b>Crash log </b></summary><p>

```
com.google.android.exoplayer2.ExoPlaybackException: Source error
	at com.google.android.exoplayer2.ExoPlayerImplInternal.handleIoException(ExoPlayerImplInternal.java:641)
[...]

Issue #5730, opened 2021-03-02
App can't continue YouTube video that was paused some hours ago


JCU0:
### Checklist

- [x] I am using the latest version - 0.20.10 
- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one bug. I will open one issue for every bug report I want to file.

### Steps to reproduce the bug

1. Start a video to play
2. Pause a video.
3. Lock your device.
4. Sleep and relax for some hours
5. Unlock your device
6. Try to resume a video (problem is here)
7. Try to play another video (no problem)
8. Try to go back to previously paused video via History, Playlists tab, etc (problem is here again)

### Actual behaviour
A video is starting to play for up to a minute (pre-loaded part), then it is freezing in an infinite load loop. When you try to switch to another video and return to that one, it's still not playable. You need to force close and reload the app to be able to continue with that video.

### Expected behavior
A video should continue to play smoothly until end.

jonnyb49:
This behavior happens even if paused in background mode. However this is not a new problem and has been observed in several last releases. Sometimes if there is enough of data queued up it could play for couple of seconds even a minute after this long pause , but as soon as it should fetch new data the play stops and stays like that forever.

It appears as if there was some kind of timer that overflows in this long pause and then it cannot start where it left off. Or the timestamp of the last queued data does not match current time thus unable to fetch new data from the source.
 
One workaround found is to stop the app, clear the  storage cache in app settings, restart the app and via history restart the playback. However there is one big drawback to this action. It seems that the comments disappear completely after clearing the cache for all videos played afterwards. They sometimes resurface again after couple of day, but have not figured out what triggers this condition. But this is probably another bug issue that needs to be reported separately.

XiangRongLin:
You can also just change the video quality to fix there current instance of the problem

davidasunmo:
> You can also just change the video quality to fix there current instance of the problem

And how to fix in background mode without having to force close the app and lose your current queue?

FlareHeart:
I am still having this issue in 21.1. Now it won't even play buffered video and it seems newpipe thinks the video has ended since the icon changes to the "replay" arrow rather than the play triangle. Resuming is not possible and requires restarting the video.

Lomanic:
[...]

PR #9167, opened 2022-10-21
Added response to error code for bad http status

JasonT962:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

A lot of people (including me) have been getting a network error 403 response code randomly when starting a new video. 403 response code is when the client is forbidden from accessing a URL. A common fix for this was to wipe your cached metadata in settings and then reloading the video.

This PR works by wiping your metadata cache and then reloading the video playback if the player runs into any bad http status error.

- Fixes https://github.com/TeamNewPipe/NewPipe/issues/9038
- also fixes https://github.com/TeamNewPipe/NewPipe/issues/8933
- also fixes https://github.com/TeamNewPipe/NewPipe/issues/8627
- should fix https://github.com/TeamNewPipe/NewPipe/issues/5730 as seems like same issue directly above from comments and marked duplicates. However, no logs given for this issue.

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -246,6 +246,11 @@ public final class Player implements PlaybackListener, Listener {
     @NonNull private final SharedPreferences prefs;
     @NonNull private final HistoryRecordManager recordManager;
 
+    /*//////////////////////////////////////////////////////////////////////////
+    // Error states
+    //////////////////////////////////////////////////////////////////////////*/
+
+    private int badHttpStatusRetry = 0;
 
     /*//////////////////////////////////////////////////////////////////////////
     // Constructor
@@ -1407,11 +1412,16 @@ public final class Player implements PlaybackListener, Listener {
                 break;
             case ERROR_CODE_IO_INVALID_HTTP_CONTENT_TYPE:
             case ERROR_CODE_IO_BAD_HTTP_STATUS:
-                isCatchableException = true;
-                // Clears metadata cache and then reloads playback
-                InfoCache.getInstance().clearCache();
-                setRecovery();
-                reloadPlayQueueManager();
+                if (badHttpStatusRetry < 3) {
+                    badHttpStatusRetry += 1;
+                    isCatchableException = true;
+                    // Clears metadata cache and then reloads playback
+                    InfoCache.getInstance().clearCache();
+                    setRecovery();
+                    reloadPlayQueueManager();
+                } else {
+                    isCatchableException = false;
+                }
                 break;
             case ERROR_CODE_IO_FILE_NOT_FOUND:
             case ERROR_CODE_IO_NO_PERMISSION:
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -1407,6 +1407,12 @@ public final class Player implements PlaybackListener, Listener {
                 break;
             case ERROR_CODE_IO_INVALID_HTTP_CONTENT_TYPE:
             case ERROR_CODE_IO_BAD_HTTP_STATUS:
+                isCatchableException = true;
+                // Clears metadata cache and then reloads playback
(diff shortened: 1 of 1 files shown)


AudricV (changes requested):
Hmm, that's not a good behavior in my opinion to clear the cache and retry each time this issue occurs.

For YouTube, this is an A/B test, they could roll out this change globally in UK or in the entire world, making your workaround useless.

This error affects other apps, even reVanced if I am not wrong, and must be fixed by extractor changes (and that's pretty easy in fact, by changing the primary InnerTube client used for playback (`ANDROID`) or removing it, unless a real fix with it is found).

However, I think we could retry to get stream info only one time when this type of error occurs, without clearing cache. The expiration time is managed by NewPipe's cache system for each service. This would solve 403 HTTP errors when stream URLs have expired (for instance on YouTube and SoundCloud).

JasonT962:
> Hmm, that's not a good behavior in my opinion to clear the cache and retry each time this issue occurs.
> 
> For YouTube, this is an A/B test, they could roll out this change globally in UK or in the entire world, making your workaround useless.
> 
> This error affects other apps, even reVanced if I am not wrong, and must be fixed by extractor changes (and that's pretty easy in fact, by changing the primary InnerTube client used for playback (`ANDROID`) or removing it, unless a real fix with it is found).
> 
> However, I think we could retry to get stream info only one time when this type of error occurs, without clearing cache. The expiration time is managed by NewPipe's cache system for each service. This would solve 403 HTTP errors when stream URLs have expired (for instance on YouTube and SoundCloud).

I don't mind changing it so it only retries to get stream info 1 time without clearing cache but I don't see how I would be able to do that without introducing a new private variable for Player.java to keep track of if it encounters the error twice in a row. So, am I allowed to do that or no?

JasonT962:
I can also keep the original clear cache but it retries it up to x times before giving an error message. So maybe like 3 times?

JasonT962:
So I limited how many times the playback could be reloaded before it gave an error message. I've currently set the limit to 3 as with 1 attempt, the problem is a lot less common but still happens occasionally, with 2 attempts the problem was basically gone so I just went with 3 attempts to be safe.

If you still want to go with your original suggestion where it doesn't wipe cache and only reloads once, then all you need to do is set the limit to 1 and remove the InfoCache.getInstance().clearCache(); line.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -1407,6 +1412,17 @@ public void onPlayerError(@NonNull final PlaybackException error) {
                 break;
             case ERROR_CODE_IO_INVALID_HTTP_CONTENT_TYPE:
             case ERROR_CODE_IO_BAD_HTTP_STATUS:
+                if (badHttpStatusRetry < 3) {
+                    badHttpStatusRetry += 1;
+                    isCatchableException = true;
+                    // Clears metadata cache and then reloads playback
```
Stypox, line 1418: Add more comments here, with what you have in the PR description.

Stypox (changes requested):
Would it be possible to clear the cache only of the currently playing video? That way we prevent the whole cache from being deleted (cache is there for a reason, to make things faster, so deleting it completely does not seem like a good idea).

Also, I think the `badHttpStatusRetry` variable should be reset to 0 whenever the currently playing stream changes. Otherwise, with the current implementation, only the first 3 videos with failures would benefit from retrying.

opusforlife2:
@JasonT962 ?

JasonT962:
Sorry, I didn't think my workaround was necessary anymore, I also don't have the time to work on this right now.

app/src/main/java/org/schabi/newpipe/player/Player.java
JasonT962: I'm not sure if you can only clear the cache of the currently playing video.

However, I have tested the badHttpStatusRetry variable thing and it does reset back to 0 automatically whenever the player changes. For example, if you watch a different video. I think it's probably because the Player.java class gets created every time you watch a new video.

Issue #8800, opened 2022-08-15
Add a 'Open in Browser' button to the broken video page


astatine:
### Checklist

- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I'm aware that this is a request for NewPipe itself and that requests for adding a new service need to be made at [NewPipeExtractor](https://github.com/TeamNewPipe/NewPipeExtractor/issues).
- [X] I have taken the time to fill in all the required details. I understand that the feature request will be dismissed otherwise.
- [X] This issue contains only one feature request.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Feature description

Currently, NewPipe shows the 'Sorry, something went wrong' page with the big frowny face when the video hasn't premiered yet, and when NewPipe itself is broken. That page has `Report` and `Retry` buttons with expected functionality.

For both these cases, it would be great if a `Open in Browser` button is added to the page.

<img alt="Sorry, something went wrong." src="https://user-images.githubusercontent.com/1077402/184637375-a6e2670b-2b01-4779-a39c-9680ae2f3bd4.png" height=600 />

### Why do you want this feature?

This would enable the user to know when the video will premiere, and in the second case would enable the user to watch the video.

SameenAhnaf:
@mhmdanas Duplicate of https://github.com/TeamNewPipe/NewPipe/issues/6624

mhmdanas:
@SameenAhnaf no, this isn't a duplicate of #6624. This issue requests an "Open in browser" button when errors occur, while that issue asks for showing more information in case of errors.

opusforlife2:
Would really prefer it if the browser is sent a Piped/Invidious link instead of a YT one.

mhmdanas:
I suppose that could be added as a setting.

opusforlife2:
#2351 would be a prerequisite then.

YonghaoDeng:
Hey, could I work on this issue? :))

opusforlife2:
@YonghaoDeng Please give some idea of how you will implement this? A button under Report and Retry saying "Open in Browser"? Or something else?

YonghaoDeng:
> s

[...]

PR #9180, opened 2022-10-23
Add an Open in browser button on error panel

YonghaoDeng:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

With the addition of the open_in_browser button design, the The user can watch the video in case the new pipe itself is faulty.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

- Fixes  https://github.com/TeamNewPipe/NewPipe/issues/8800

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
+++ b/app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
@@ -82,8 +82,10 @@ class ErrorInfo(
 
         fun throwableListToStringList(throwableList: List<Throwable>) =
             throwableList.map { it.stackTraceToString() }.toTypedArray()
+
         private fun getInfoServiceName(info: Info?) =
             if (info == null) SERVICE_NONE else ServiceHelper.getNameOfServiceById(info.serviceId)
+
         @StringRes
         private fun getMessageStringId(
             throwable: Throwable?,
--- a/app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
+++ b/app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
@@ -82,10 +82,8 @@ class ErrorInfo(
 
         fun throwableListToStringList(throwableList: List<Throwable>) =
             throwableList.map { it.stackTraceToString() }.toTypedArray()
-
         private fun getInfoServiceName(info: Info?) =
             if (info == null) SERVICE_NONE else ServiceHelper.getNameOfServiceById(info.serviceId)
-
         @StringRes
         private fun getMessageStringId(
             throwable: Throwable?,
--- a/app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
+++ b/app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
@@ -2,12 +2,14 @@ package org.schabi.newpipe.error
 
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
 import android.util.Log
 import android.view.View
 import android.widget.Button
 import android.widget.TextView
 import androidx.annotation.Nullable
 import androidx.annotation.StringRes
+import androidx.core.content.ContextCompat.startActivity
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import com.jakewharton.rxbinding4.view.clicks
@@ -52,6 +54,8 @@ class ErrorPanelHelper(
         errorPanelRoot.findViewById(R.id.error_action_button)
     private val errorRetryButton: Button =
         errorPanelRoot.findViewById(R.id.error_retry_button)
+    private val errorOpenbrowserButton: Button =
+        errorPanelRoot.findViewById(R.id.error_openinbrowser)
 
     private var errorDisposable: Disposable? = null
 
@@ -69,18 +73,16 @@ class ErrorPanelHelper(
         errorServiceExplanationTextView.isVisible = false
         errorActionButton.isVisible = false
         errorRetryButton.isVisible = false
+        errorOpenbrowserButton.isVisible = false
     }
 
     fun showError(errorInfo: ErrorInfo) {
-
         if (errorInfo.throwable != null && errorInfo.throwable!!.isInterruptedCaused) {
             if (DEBUG) {
                 Log.w(TAG, "onError() isInterruptedCaused! = [$errorInfo.throwable]")
             }
             return
         }
-
-        ensureDefaultVisibility()
 
         if (errorInfo.throwable is ReCaptchaException) {
             errorTextView.setText(R.string.recaptcha_request_toast)
@@ -99,6 +101,7 @@ class ErrorPanelHelper(
             }
 
             errorRetryButton.isVisible = true
+            errorOpenbrowserButton.isVisible = true
         } else if (errorInfo.throwable is AccountTerminatedException) {
             errorTextView.setText(R.string.account_terminated)
 
@@ -127,6 +130,12 @@ class ErrorPanelHelper(
             ) {
                 // show retry button only for content which is not unavailable or unsupported
                 errorRetryButton.isVisible = true
+                errorOpenbrowserButton.isVisible = true
+                errorOpenbrowserButton.setOnClickListener {
+                    val uriUrl = Uri.parse(errorInfo.request)
+                    val intent = Intent(Intent.ACTION_VIEW, uriUrl)
+                    context.startActivity(intent)
+                }
             }
         }
 
--- a/app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
+++ b/app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
@@ -2,14 +2,12 @@ package org.schabi.newpipe.error
 
 import android.content.Context
 import android.content.Intent
-import android.net.Uri
 import android.util.Log
 import android.view.View
 import android.widget.Button
 import android.widget.TextView
 import androidx.annotation.Nullable
 import androidx.annotation.StringRes
-import androidx.core.content.ContextCompat.startActivity
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import com.jakewharton.rxbinding4.view.clicks
@@ -32,6 +30,7 @@ import org.schabi.newpipe.ktx.animate
 import org.schabi.newpipe.ktx.isInterruptedCaused
 import org.schabi.newpipe.ktx.isNetworkRelated
 import org.schabi.newpipe.util.ServiceHelper
+import org.schabi.newpipe.util.external_communication.ShareUtils
 import java.util.concurrent.TimeUnit
 
 class ErrorPanelHelper(
@@ -54,8 +53,8 @@ class ErrorPanelHelper(
         errorPanelRoot.findViewById(R.id.error_action_button)
     private val errorRetryButton: Button =
         errorPanelRoot.findViewById(R.id.error_retry_button)
-    private val errorOpenbrowserButton: Button =
-        errorPanelRoot.findViewById(R.id.error_openinbrowser)
+    private val errorOpenInBrowserButton: Button =
+        errorPanelRoot.findViewById(R.id.error_open_in_browser)
 
     private var errorDisposable: Disposable? = null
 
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
```diff
@@ -82,10 +82,8 @@ class ErrorInfo(
 
         fun throwableListToStringList(throwableList: List<Throwable>) =
             throwableList.map { it.stackTraceToString() }.toTypedArray()
-
         private fun getInfoServiceName(info: Info?) =
             if (info == null) SERVICE_NONE else ServiceHelper.getNameOfServiceById(info.serviceId)
-
```
AudricV, line 88: Remove these unneeded changes.

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
```diff
@@ -69,19 +73,17 @@ class ErrorPanelHelper(
         errorServiceExplanationTextView.isVisible = false
         errorActionButton.isVisible = false
         errorRetryButton.isVisible = false
+        errorOpenbrowserButton.isVisible = false
     }
 
     fun showError(errorInfo: ErrorInfo) {
-
```
AudricV, line 75: Same as above.

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
```diff
@@ -69,19 +73,17 @@ class ErrorPanelHelper(
...
         if (errorInfo.throwable != null && errorInfo.throwable!!.isInterruptedCaused) {
             if (DEBUG) {
                 Log.w(TAG, "onError() isInterruptedCaused! = [$errorInfo.throwable]")
             }
             return
         }
 
-        ensureDefaultVisibility()
-
```
AudricV, line 84: Could you explain why you removed this?

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
```diff
@@ -127,6 +130,12 @@ class ErrorPanelHelper(
             ) {
                 // show retry button only for content which is not unavailable or unsupported
                 errorRetryButton.isVisible = true
+                errorOpenbrowserButton.isVisible = true
+                errorOpenbrowserButton.setOnClickListener {
+                    val uriUrl = Uri.parse(errorInfo.request)
+                    val intent = Intent(Intent.ACTION_VIEW, uriUrl)
+                    context.startActivity(intent)
```
AudricV, line 137: Use [`ShareUtils.openUrlInBrowser`](https://github.com/TeamNewPipe/NewPipe/blob/2a98cca80100d87562f0c9ff1fadeac3838f6ff6/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java#L125) here instead.

app/src/main/res/layout/error_panel.xml
```diff
@@ -72,4 +72,19 @@
         android:visibility="gone"
         tools:visibility="visible" />
 
+    <Button
+        android:id="@+id/error_openinbrowser"
```
AudricV, line 76:
```suggestion
        android:id="@+id/error_open_in_browser"
```

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
```diff
@@ -52,6 +54,8 @@ class ErrorPanelHelper(
         errorPanelRoot.findViewById(R.id.error_action_button)
     private val errorRetryButton: Button =
         errorPanelRoot.findViewById(R.id.error_retry_button)
+    private val errorOpenbrowserButton: Button =
+        errorPanelRoot.findViewById(R.id.error_openinbrowser)
```
AudricV, line 58:
```suggestion
    private val errorOpenInBrowserButton: Button =
        errorPanelRoot.findViewById(R.id.error_open_in_browser)
```

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
```diff
@@ -69,19 +73,17 @@ class ErrorPanelHelper(
         errorServiceExplanationTextView.isVisible = false
         errorActionButton.isVisible = false
         errorRetryButton.isVisible = false
+        errorOpenbrowserButton.isVisible = false
```
AudricV, line 76:
```suggestion
        errorOpenInBrowserButton.isVisible = false
```

app/src/main/res/layout/error_panel.xml
```diff
@@ -72,4 +72,19 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="4dp"
+        android:layout_marginBottom="8dp"
```
AudricV, line 80: As you can notice on the screenshot you provided, the margin between the `Open in browser` and `Retry` buttons is not the same as the one between the `Report` and `Retry` buttons. Could you change this?

app/src/main/res/layout/error_panel.xml
```diff
@@ -72,4 +72,19 @@
...
+        android:text="@string/error_openinbrowser"
```
AudricV, line 81: There is already a string for `Open in browser`: https://github.com/TeamNewPipe/NewPipe/blob/2a98cca80100d87562f0c9ff1fadeac3838f6ff6/app/src/main/res/values/strings.xml#L10

```suggestion
        android:text="@string/open_in_browser"
```

app/src/main/res/values/strings.xml
```diff
@@ -251,6 +251,7 @@
     <string name="error_report_open_github_notice">Please check whether an issue discussing your crash already exists. When creating duplicate tickets, you take time from us which we could spend with fixing the actual bug.</string>
     <string name="error_snackbar_message">Sorry, something went wrong.</string>
     <string name="error_snackbar_action">Report</string>
+    <string name="error_openinbrowser">Open in Browser</string>
```
AudricV, line 254: See above.

AudricV (changes requested):
Thank you for this improvement! Here are changes that you need to apply in order to have this pull request merged.

app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
YonghaoDeng: I didn't change this line of code, it was already there, are you sure I want to remove it?

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
YonghaoDeng: I didn't change this line of code either, it was already there, are you sure I want to remove it? Also if I delete it, the file will error out

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
YonghaoDeng: It's my fault, I shouldn't have deleted it

app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
AudricV: In the case you didn't understand, I meant don't remove these lines with your changes in this PR.

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
AudricV: See my last comment above.

app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
YonghaoDeng: Oh! make sense!

YonghaoDeng: I have modified!

opusforlife2:
What will the button do, BTW? Will it directly open the default browser selected in the OS? Or show a list of apps that can handle the link type?

YonghaoDeng:
> it will directly open the default browser selected in the OS

opusforlife2:
I think it would be better to show the list of apps. The user may want to open it in an app other than the default browser, like a media player.

YonghaoDeng:
> But I think the video needs to be viewed online at the YouTube website and the requirement is to create an open in browser button, so I think it should be redirected to the browser, don't you think so?

opusforlife2:
It's perfectly fine to open it in the default browser. Nothing wrong with that behaviour.

I'm just providing a suggestion for extra functionality. If a user wants to open it in a different browser, or in a media app, they should be able to. They can always select "Always open" so that their selected choice becomes the default way to open.

YonghaoDeng:
> show the list of apps

Hmmmmm I see what you mean but I'm afraid I won't be able to do well with this extra feature. As I am an Anu student, this is an example of PR practice in one of my classes and the deadline for this assignment is this weekend

opusforlife2:
No problems at all. 👍

settings.gradle
```diff
@@ -4,8 +4,8 @@ include ':app'
 // We assume, that NewPipe and NewPipe Extractor have the same parent directory.
 // If this is not the case, please change the path in includeBuild().
 
-//includeBuild('../NewPipeExtractor') {
```
Stypox, line 7: Do not push these changes, they should be local-only

app/src/main/res/layout/error_panel.xml
```diff
@@ -72,4 +72,19 @@
         android:visibility="gone"
         tools:visibility="visible" />
 
+    <Button
+        android:id="@+id/error_open_in_browser"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="-4dp"
```
Stypox, line 79: Remove `layout_marginBottom` from `error_retry_button` and remove this line: negative margins are not a good idea, and also at the moment the buttons are not equally spaced out.

app/src/main/res/layout/error_panel.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    android:layout_width="wrap_content"
+    android:layout_width="match_parent"
```
Stypox, line 4: Why this change?

Stypox (changes requested):
Thank you! Looks almost good

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
```diff
@@ -99,6 +103,7 @@ class ErrorPanelHelper(
             }
 
             errorRetryButton.isVisible = true
+            errorOpenInBrowserButton.isVisible = true
```
Stypox, line 106: Shouldn't the on click listener be set here, too? I suggest that you create a `showAndSetOpenInBrowserButtonAction` function which sets the button to visible and sets the on click listener. Then you can call that function both here and 30 lines below here ;-)

YonghaoDeng:
I added the function as you said @Stypox

app/src/main/res/layout/error_panel.xml
```diff
@@ -62,7 +62,7 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_marginTop="4dp"
-        android:layout_marginBottom="8dp"
+
```
Stypox, line 65: Remove this blank line
```suggestion
```

app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
```diff
@@ -127,6 +132,7 @@ class ErrorPanelHelper(
             ) {
                 // show retry button only for content which is not unavailable or unsupported
                 errorRetryButton.isVisible = true
+                showAndSetOpenInBrowserButtonAction(errorInfo)
```
Stypox, line 135: I think it's better if you put this line outside the `if (errorInfo.throwable !is ContentNotAvailableException &&
                errorInfo.throwable !is ContentNotSupportedException
            )`, since it's ok to allow the user to go to the browser even when the content is unavailable or unsupported

Stypox (changes requested):
Just a few more little changes

YonghaoDeng:
OK，I modified the code as you said @Stypox

Stypox (approved):
Looks good to me. Thank you!

PR #9203, opened 2022-10-26
Added timestamped link sharing from the start of a chapter with a long hold press

### LINKED ISSUE #8343: chapter timestamp share
### Checklist

### Feature description

Can be implemented either in two ways:
1. Put share button beside each chapter
2. Or Hold the chapter to share the link with the chapter timestamp included

To quickly share the link with timestamp included, so that the stream would play from that timestamp instead of beginning.

### Additional information

As an example, the timestamp option append looks like `&t=1m30s` at the end of the link

Callisto404:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Added the ability to share timestamped links starting from the beginning of a particular chapter when the user long clicks on them

- Fixes #8343 

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java
+++ b/app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java
@@ -3,6 +3,7 @@ package org.schabi.newpipe.player.ui;
 import static android.view.ViewGroup.LayoutParams.MATCH_PARENT;
 import static org.schabi.newpipe.MainActivity.DEBUG;
 import static org.schabi.newpipe.QueueItemMenuUtil.openPopupMenu;
+import static org.schabi.newpipe.extractor.ServiceList.YouTube;
 import static org.schabi.newpipe.ktx.ViewUtils.animate;
 import static org.schabi.newpipe.player.Player.STATE_COMPLETED;
 import static org.schabi.newpipe.player.Player.STATE_PAUSED;
@@ -61,6 +62,7 @@ import org.schabi.newpipe.player.gesture.BasePlayerGestureListener;
 import org.schabi.newpipe.player.gesture.MainPlayerGestureListener;
 import org.schabi.newpipe.player.helper.PlaybackParameterDialog;
 import org.schabi.newpipe.player.helper.PlayerHelper;
+import org.schabi.newpipe.player.mediaitem.MediaItemTag;
 import org.schabi.newpipe.player.playqueue.PlayQueue;
 import org.schabi.newpipe.player.playqueue.PlayQueueAdapter;
 import org.schabi.newpipe.player.playqueue.PlayQueueItem;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java
```diff
@@ -696,10 +698,24 @@ public void onScrolledDown(final RecyclerView recyclerView) {
...
+            }
+
+            @Override
+            public void onItemLongClick(@NonNull final StreamSegmentItem item, final int seconds) {
+                final PlayQueueItem currentItem = player.getCurrentItem();
+                if (currentItem != null) {
+                    String videoUrl = player.getVideoUrl();
+                    videoUrl += ("&t=" + seconds);
+                    ShareUtils.shareText(context, currentItem.getTitle(),
+                            videoUrl, currentItem.getThumbnailUrl());
```
Stypox, line 716: This won't work for services other than YouTube. You should use the same logic as in the player: https://github.com/TeamNewPipe/NewPipe/blob/4a7af6f9ac50c0a5a3ed3b306d08d4b8aa622aea/app/src/main/java/org/schabi/newpipe/player/Player.java#L1782

Stypox (changes requested):
Thanks for the PR, looks almost good

<sup>Are you an ANU student with the strict deadline on the 31st of October?</sup>

app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java
Callisto404: Hey sorry for clarification, do you mean that I should be surrounding line 714 with an if statement checking that the service ID is indeed YouTube? I did look at this part in the player but thought it might not be necessary since I thought that these click handlers on the segments would only ever be called from a YouTube service. Sorry if I am misunderstanding and thanks for the quick replies!
<sub>And yes I am an ANU student :)</sub>

Stypox: > I thought that these click handlers on the segments would only ever be called from a YouTube service

That's probably true, but who knows if a future service will support segments ;-)

Yes, sorround it with a "is youtube" check. If it's not YouTube, don't allow long-pressing the item at all.

Callisto404: Hey, I have been attempting to add this check to the `buildSegements` function with `setLongClickable(false)`, however even with `setOnLongClickListener(null)`, the long click event still occurs. Similarly, the existing `onClick` method used for seeking to different chapters does not get disabled by setting clickable to false. I have also tried adding the check to the beginning of the long click call, however since the call consumes the click event, this is not the desired result.

> don't allow long-pressing the item at all.  

Is there some better way of doing this that I am unaware of? Thanks

Stypox: Ok, then it's fine if you just don't do anything when a long-press is detected but the url is non-youtube

Stypox (approved):
I tested and it works well. Thank you and good luck with the exams! ;-)

Issue #8605, opened 2022-07-11
Add option to display channel groups as list instead of grid


pc00per:
### Checklist

- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I'm aware that this is a request for NewPipe itself and that requests for adding a new service need to be made at [NewPipeExtractor](https://github.com/TeamNewPipe/NewPipeExtractor/issues).
- [X] I have taken the time to fill in all the required details. I understand that the feature request will be dismissed otherwise.
- [X] This issue contains only one feature request.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Feature description

A list form of channel group instead of Grid form

### Why do you want this feature?

Some users have long lengthy list of channel groups, which also make them hard to sort & also scrolling them from left to right, seems unintuitive.

### Additional information

And I personally don't like manual channel group sorting & Instead prefer Alphanumeric order (might need to open another issue for this)

And also, an another problem of Grid Layout is that it masks bigger Group titles & show only first 8 of their characters on the small dpi displays.

plasticanu:
Hi, we are an assignment group (6 people) studying computer science at the Australian National University. We would like to work on this issue, and we have 19 days left to write code and open a pull request. Can you assign this issue to us?

mhmdanas:
@plasticanu somebody will assign you to the issue eventually, but you don't have to wait for that to happen before you actually start working on it.

opusforlife2:
@plasticanu You've offered to work on a bunch of issues. Do you plan to work on any one, a few, or all of them? Please clarify so that I can start assigning.

plasticanu:
> 

Hi, sorry for the late reply. I am going to work with 2 people from my team. So we'll be a 3 people team.

cernunnos1710:
Hello, I would like to also worked on this issue with @plasticanu.

cernunnos1710:
Hi! For the list layout, Is it alright if I add a toggle to change it into a vertical version of the card layout instead of completely overhauling into a list view?

pc00per:
[...]

PR #9204, opened 2022-10-26
Add option to display channel groups as list instead of grid

cernunnos1710:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

- Created a list view for channel groups in the subscriptions page
- Added a toggle button to toggle between the original card view to the newly added one
- Unlike sort, toggle button will exist at all times
- The newly formatted buttons conforms to the same logic as the ones previously. All existing buttons work as intended as they should and will automatically update the ones on the other display format when toggled

Screenshot_20221026-205343
Screenshot_20221026-205345

#### Fixes the Following issue: 
- Fixes: [https://github.com/TeamNewPipe/NewPipe/issues/8605](https://github.com/TeamNewPipe/NewPipe/issues/8605)
#### APK testing

#### Due diligence


cernunnos1710:
The issue right now is it's not possible to mess with the toggle function without rewriting a big chunk of the code. These are two different entities and it's not possible to do @SetOnTouchListener and @SetOnClickListener at the same time, so you end up with a list that's merely draggable and nothing more.

cernunnos1710:
There is a commit where I retain the old list view with a new card view, which I think is a nice implementation of the code.

cernunnos1710:
image

This would be the final implementation of the list view if we do not need a toggle next to each list element

PR #9230, opened 2022-10-27
Fix duplicate videos in feed group "All"

### LINKED ISSUE #9024: v 0.24.0 gives me double videos in feed list
### Checklist

### Affected version

0.24.0

1. Updated to v0.24.0
2. Refreshed feed list of subscribed channels

### Expected behavior

List of all new videosz one entry per video.

### Actual behavior

Some videos appear twice in the list.

Screenshot here: https://www.reddit.com/r/NewPipe/comments/xnq9l0/why_this_after_updating_to_v24/

### Logs

_No response_

CalyxOS 3.9.0

### Affected device model

Google Pixel 4a

### Additional information

_No response_

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

In #8621 I deduplicated the queries to get feed info, but I made a mistake regarding the "All" group. The LEFT JOIN of the subscription table with the feed groups table rightfully generates multiple results for each subscription (since it is a join), but, for user-created groups, all results except for at most one are, in the end, filtered out by the first condition in the WHERE, because a subscription can appear in a group at most once. When the query was used for the "All" group, though, the first condition in the WHERE would not do anything (the reasoning behind this was to allow subscriptions independently of groups), causing duplicate subscription results to be generated (i.e. one per group the subscription belonged to). This PR fixes the problem by just making the condition on the LEFT JOIN stricter: since when we are selecting all subscriptions we don't need data about groups, the LEFT JOIN now always fails since the condition in the ON would evaluate to false when groupId=all. Therefore, since we are using a left join, it is ok for the groups table columns to not exist, and hence exactly one result gets generated.

- Fixes #9024

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/database/feed/dao/FeedDAO.kt
+++ b/app/src/main/java/org/schabi/newpipe/database/feed/dao/FeedDAO.kt
@@ -48,7 +48,10 @@ abstract class FeedDAO {
         ON s.uid = f.stream_id
 
         LEFT JOIN feed_group_subscription_join fgs
-        ON fgs.subscription_id = f.subscription_id
+        ON (
+            :groupId <> ${FeedGroupEntity.GROUP_ALL_ID}
+            AND fgs.subscription_id = f.subscription_id
+        )
 
         WHERE (
             :groupId = ${FeedGroupEntity.GROUP_ALL_ID}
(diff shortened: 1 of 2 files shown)


TobiGr (commented):
Code changes lgtm

PR #9285, opened 2022-11-03
Clean up Optional-related code.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Remove `SimplifyOptionalCallChains`, as the `Optional#isEmpty()` method is now available in the Android SDK starting with version 33 and is desugared.
- Make use of `Optional` method chaining.

- Fixes #

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -255,11 +255,8 @@ public final class VideoDetailFragment
             playerUi.ifPresent(MainPlayerUi::toggleFullscreen);
         }
 
-        //noinspection SimplifyOptionalCallChains
-        if (playAfterConnect
-                || (currentInfo != null
-                && isAutoplayEnabled()
-                && !playerUi.isPresent())) {
+        if (playAfterConnect || (currentInfo != null && isAutoplayEnabled()
+                && playerUi.isEmpty())) {
             autoPlayEnabled = true; // forcefully start playing
             openVideoPlayerAutoFullscreen();
         }
@@ -1174,16 +1171,15 @@ public final class VideoDetailFragment
      * be reused in a few milliseconds and the flickering would be annoying.
      */
     private void hideMainPlayerOnLoadingNewStream() {
-        //noinspection SimplifyOptionalCallChains
-        if (!isPlayerServiceAvailable() || !getRoot().isPresent()
-                || !player.videoPlayerSelected()) {
+        final var root = getRoot();
+        if (!isPlayerServiceAvailable() || root.isEmpty() || !player.videoPlayerSelected()) {
             return;
         }
 
         removeVideoPlayerView();
         if (isAutoplayEnabled()) {
             playerService.stopForImmediateReusing();
-            getRoot().ifPresent(view -> view.setVisibility(View.GONE));
+            root.ifPresent(view -> view.setVisibility(View.GONE));
         } else {
             playerHolder.stopService();
         }
@@ -1887,10 +1883,8 @@ public final class VideoDetailFragment
     @Override
     public void onFullscreenStateChanged(final boolean fullscreen) {
         setupBrightness();
-        //noinspection SimplifyOptionalCallChains
-        if (!isPlayerAndPlayerServiceAvailable()
-                || !player.UIs().get(MainPlayerUi.class).isPresent()
-                || getRoot().map(View::getParent).orElse(null) == null) {
+        if (!isPlayerAndPlayerServiceAvailable() || player.UIs().get(MainPlayerUi.class).isEmpty()
+                || getRoot().map(View::getParent).isEmpty()) {
             return;
         }
 
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -255,7 +255,9 @@ public final class VideoDetailFragment
             playerUi.ifPresent(MainPlayerUi::toggleFullscreen);
         }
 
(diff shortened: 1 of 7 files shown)


Stypox:
Why do you think `SimplifyOptionalCallChains` is not needed anymore? It was not added by accident: it was added because the IDE was complaining. So why does Android Studio say it's not a valid method, but at the same time suggests it? Is it desugared properly?

image
image

app/src/main/java/org/schabi/newpipe/player/playback/MediaSourceManager.java
```diff
@@ -422,21 +421,24 @@ private void maybeLoadItem(@NonNull final PlayQueueItem item) {
...
-                        FailedMediaSource.of(stream, new MediaSourceResolutionException(message));
-            }
-
-            final MediaItemTag tag = MediaItemTag.from(source.getMediaItem()).get();
-            final long expiration = System.currentTimeMillis()
-                    + ServiceHelper.getCacheExpirationMillis(streamInfo.getServiceId());
-            return new LoadedMediaSource(source, tag, stream, expiration);
+            final var source = playbackListener.sourceOf(stream, streamInfo);
+
+            return Optional.ofNullable(source)
```
Stypox, line 426: And also turn `source1` into `source` below

```suggestion
            return Optional.ofNullable(playbackListener.sourceOf(stream, streamInfo))
```

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2036,40 +2030,35 @@ public void useVideoSource(final boolean videoEnabled) {
...
-        }
+                trackSelector.setParameters(parametersBuilder);
+            }
 
-        setRecovery();
+            setRecovery();
+        }, () -> {
+            reloadPlayQueueManager();
+            setRecovery();
+        });
```
Stypox, line 2061: Maybe add a comment in this lambda to make it clear that it is called when we have no current stream info

Isira-Seneviratne:
> Why do you think `SimplifyOptionalCallChains` is not needed anymore? It was not added by accident: it was added because the IDE was complaining. So why does Android Studio say it's not a valid method, but at the same time suggests it? Is it desugared properly?
> 
> image image

Yeah, the `isEmpty()` method was added to the Android SDK starting with version 33, and it gets desugared.

Stypox (changes requested):
Other simplifications look good to me

Stypox:
> Yeah, the isEmpty() method was added to the Android SDK starting with version 33, and it gets desugared.

Oh great! I had gradle out of sync with the ide

app/src/main/java/org/schabi/newpipe/player/mediaitem/MediaItemTag.java
```diff
@@ -61,12 +61,11 @@ default Optional<Quality> getMaybeQuality() {
...
     static Optional<MediaItemTag> from(@Nullable final MediaItem mediaItem) {
-        if (mediaItem == null || mediaItem.localConfiguration == null
-                || !(mediaItem.localConfiguration.tag instanceof MediaItemTag)) {
-            return Optional.empty();
-        }
-
-        return Optional.of((MediaItemTag) mediaItem.localConfiguration.tag);
+        return Optional.ofNullable(mediaItem)
+                .map(item -> item.localConfiguration)
+                .map(localConfiguration -> localConfiguration.tag)
```
Stypox: `map` will crash if the return value of the lambda is `null`. I looked it up but there does not seem to be a `mapNullable` (just like `ofNullable` exists). So I used `flatMap(... -> Optional.ofNullable(...))`

```java
jshell> Optional.of(new Object()).map(null)
|  Exception java.lang.NullPointerException
|        at Objects.requireNonNull (Objects.java:233)
|        at Optional.map (Optional.java:256)
|        at (#2:1)

jshell> Optional.of(new Object()).flatMap(o -> Optional.ofNullable(null))
$4 ==> Optional.empty
```

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -1887,10 +1883,8 @@ public void onServiceStopped() {
     @Override
     public void onFullscreenStateChanged(final boolean fullscreen) {
         setupBrightness();
-        //noinspection SimplifyOptionalCallChains
-        if (!isPlayerAndPlayerServiceAvailable()
-                || !player.UIs().get(MainPlayerUi.class).isPresent()
-                || getRoot().map(View::getParent).orElse(null) == null) {
+        if (!isPlayerAndPlayerServiceAvailable() || player.UIs().get(MainPlayerUi.class).isEmpty()
+                || getRoot().map(View::getParent).isEmpty()) {
```
Stypox: Your code is perfectly equivalent, but I think the `.map(View::getParent)` was wrongly being used to allow `getParent` to return `null`.

Stypox (approved):
I made some small changes, in part described below, in the last commit, to fix some things and make further improvements. Thank you!

app/src/main/java/org/schabi/newpipe/player/mediaitem/MediaItemTag.java
Isira-Seneviratne: That's weird, `null` values should be converted to empty `Optional`s: https://developer.android.com/reference/java/util/Optional#map(java.util.function.Function%3C?%20super%20T,%20?%20extends%20U%3E)

Stypox: I will revert that part of my commit

Stypox: I pushed ad605e2c5a38e57af1b0ec7d4c991a9a9240dfff

PR #9463, opened 2022-11-25
Fixed bug that lead to faulty playlist thumbnail updates

### LINKED ISSUE #9458: Playlist thumbnail doesn't get updated after the "Remove Watched" dialog is used
### Checklist

### Affected version

0.24.1

1. Create a new playlist with two videos
2. Go to playlist overview
3. Go to the newly created playlist
4. Click "Background"
5. Listen for a couple of seconds
6. Click "Removed Watched"
7. Click "Yes, and partially watched videos"
8. Go to Playlist Overview

### Expected behavior

The thumbnail of the playlist should have changed to the only item in the playlist. 

### Actual behavior

The thumbnail doesn't get updated.

https://user-images.githubusercontent.com/26669009/203877279-809cfc90-4c17-4635-b995-1dde58416bce.mp4

### Logs

_No response_

Android 12

### Affected device model

Pixel 5

### Additional information

I am a developer and plan to work on this problem myself.

Jared234:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)

This is an attempt to fix the bug mentioned in #9458.

I found out that this bug not only happens if you play videos in the background but also if you watch them “normally”.
The problem is that there are two types of thumbnails. There is a low- and a high-resolution thumbnail for every video. So, after you start watching a video in a particular playlist, the thumbnail of the watched item changes from low to high resolution. However, this change is not detected in the playlist. Therefore, all of the following equality checks no longer work because you are comparing the low-resolution and high-resolution thumbanails, which have completely different URLs. 

I thought about adding a field that stores both versions. However, this implementation felt like a temporary hack and would most likely cause unnecessary database bloat. 
Therefore, I decided to check the "Thumbnail" video for a change of the thumbnail every time the items in the playlist are changed or replaced. This solution has another advantage as well. When the creator of a YouTube video changes the thumbnail, it should be automatically updated in NewPipe.

If there is anything I did wrong, please let me know and I will fix it. 

- Fixes #9458 

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
@@ -73,6 +73,8 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
     // Save the list 10 seconds after the last change occurred
     private static final long SAVE_DEBOUNCE_MILLIS = 10000;
     private static final int MINIMUM_INITIAL_DRAG_VELOCITY = 12;
+    private static final int VIDEO_ID_NO_LONGER_IN_PLAYLIST = -1;
+
     @State
     protected Long playlistId;
     @State
@@ -472,6 +474,49 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
                         "Removing watched videos, partially watched=" + removePartiallyWatched))));
     }
 
+    /**
+     * If the current thumbnail has changed but the video is still the same, this method will update
+     * the playlist thumbnail accordingly.
+     *
+     * @param currentItems List of items currently in the playlist
+     * @param updatedItems List of items that will replace the current items in the playlist
+     */
+    private void updateCurrentThumbnail(@NonNull final List<LocalItem> currentItems,
+                                        @NonNull final List<PlaylistStreamEntry> updatedItems) {
+
+        final String playlistThumbnail = playlistManager.getPlaylistThumbnail(playlistId);
+        long streamId = VIDEO_ID_NO_LONGER_IN_PLAYLIST;
+
+        // Find the streamId of the current thumbnail video
+        for (final LocalItem item : currentItems) {
+            if (playlistThumbnail.equals(((PlaylistStreamEntry) item).getStreamEntity()
+                    .getThumbnailUrl())) {
+                streamId = ((PlaylistStreamEntry) item).getStreamId();
+            }
+        }
+
+        if (streamId == VIDEO_ID_NO_LONGER_IN_PLAYLIST) {
+            return;
+        }
+
+        // If the video has a different thumbnail now, update the playlist thumbnail accordingly
+        for (final PlaylistStreamEntry item : updatedItems) {
+
+            if (streamId == item.getStreamId()
+                    && !playlistThumbnail.equals(item.getStreamEntity().getThumbnailUrl())) {
+                final String newThumbnailUrl = item.getStreamEntity().getThumbnailUrl();
+
+                final Disposable disposable = playlistManager
+                        .changePlaylistThumbnail(playlistId, newThumbnailUrl)
+                        .observeOn(AndroidSchedulers.mainThread())
+                        .subscribe(ignored -> Log.d(TAG, "Updating current thumbnail with new"
+                                + "url=[" + newThumbnailUrl + "]"));
+                disposables.add(disposable);
+            }
+
+        }
+    }
+
     @Override
     public void handleResult(@NonNull final List<PlaylistStreamEntry> result) {
         super.handleResult(result);
@@ -479,6 +524,7 @@ public class LocalPlaylistFragment extends BaseLocalListFragment<List<PlaylistSt
             return;
         }
 
+        updateCurrentThumbnail(itemListAdapter.getItemsList(), result);
         itemListAdapter.clearStreamItemList();
 
         if (result.isEmpty()) {


app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -472,13 +472,57 @@ public void removeWatchedStreams(final boolean removePartiallyWatched) {
...
     }
 
+    /**
+     * If the current thumbnail has changed but the video is still the same, this method will update
+     * the playlist thumbnail accordingly.
+     *
+     * @param currentItems List of items currently in the playlist
+     * @param updatedItems List of items that will replace the current items in the playlist
+     */
+    private void updateCurrentThumbnail(final List<LocalItem> currentItems,
```
evermind-zz, line 482: Could you please annotate  your parameters  if they are @Nullable or @NotNull

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -472,13 +472,57 @@ public void removeWatchedStreams(final boolean removePartiallyWatched) {
...
+                                        final List<PlaylistStreamEntry> updatedItems) {
+
+        final String playlistThumbnail = playlistManager.getPlaylistThumbnail(playlistId);
+        long streamId = -1;
```
evermind-zz, line 486: Could you please avoid magic numbers?

Jared234:
Thanks for the review. I hope it's fine now.

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
Jared234: I added a comment explaining this variable/number. I hope this is understandable now.

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -472,13 +472,59 @@ public void removeWatchedStreams(final boolean removePartiallyWatched) {
...
+     * @param currentItems List of items currently in the playlist
+     * @param updatedItems List of items that will replace the current items in the playlist
+     */
+    private void updateCurrentThumbnail(@NonNull final List<LocalItem> currentItems,
+                                        @NonNull final List<PlaylistStreamEntry> updatedItems) {
+
+        final String playlistThumbnail = playlistManager.getPlaylistThumbnail(playlistId);
+        // StreamId of the video that provides the thumbnail for the playlist (stays at index -1 if
+        // the video is no longer in the playlist).
+        long streamId = -1;
```
evermind-zz, line 488: Thx for your comment. Actually I was thinking about something like this:

```suggestion
        // StreamId of the video that provides the thumbnail for the playlist (stays at index -1 if
        // the video is no longer in the playlist).
        private static final int VIDEO_ID_NO_LONGER_IN_PLAYLIST = -1;
        
        long streamId = VIDEO_ID_NO_LONGER_IN_PLAYLIST;
```

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
```diff
@@ -472,13 +472,59 @@ public void removeWatchedStreams(final boolean removePartiallyWatched) {
...
+
+        // Find the streamId of the current thumbnail video
+        for (final LocalItem item : currentItems) {
+            if (playlistThumbnail.equals(((PlaylistStreamEntry) item).getStreamEntity()
+                    .getThumbnailUrl())) {
+                streamId = ((PlaylistStreamEntry) item).getStreamId();
+            }
+        }
+
+        if (streamId == -1) {
```
evermind-zz, line 498:
```suggestion
        if (streamId == VIDEO_ID_NO_LONGER_IN_PLAYLIST ) {
```
likewise

app/src/main/java/org/schabi/newpipe/local/playlist/LocalPlaylistFragment.java
Jared234: I was actually thinking of adding another constant, but since the "-1" would just be a helper index and could be any random negative number, I thought a small comment would suffice.
But I understand your perspective as well and want to align with NewPipes style, so I'll implement it now. 😄

evermind-zz (commented):
Sorry for the delay! Code looks good! But as I've no glue in this part of NewPipe code hopefully someone else will review your fix soon.

One last thing. Can you stash your 3 commits into one?

Stypox:
This does not really fix the issue when the playlist fragment is not open. Steps to reproduce:
- go to trending
- long-press a video
- add it to a new playlist (do not open the playlist!)
- play that video for some seconds
- open the playlist
- long press the only video
- delete it

=> the playlist thumbnail is not reset to the default one

I don't think the approach taken in this PR is the correct one, as it will only work in the rare case in which the playlist fragment was open in the background listening for changes when a specific video is played and its thumbnail gets updated. A better approach would probably to be to store in the database not the thumbnail, but rather the primary key of the video (belonging to the playlist) to look up to find the thumbnail. And then each time you get the list of playlists you also join with streams to get the connected thumbnails.

Jared234:
Thanks for the review. While working on your suggestions, I have stumbled upon an issue that I am not sure how to fix.

The problem is that the database migration will most likely not work entirely. My plan was to add a new column `thumbnail_stream_id` and insert the correct streamIds by joining the playlist table with the stream table on the thumbnail_url.
However, this won't work on playlists that already exist. AFAIK there are cases in which you won't be able to link the thumbnailURLs of these playlists to the uid of the streams. Thus, the best we could do in these cases is setting the first video as the video_thumbnail or -1 if there are no videos in the playlist.

Do you have a better solution for this problem?

Stypox:
> However, this won't work on playlists that already exist. AFAIK there are cases in which you won't be able to link the thumbnailURLs of these playlists to the uid of the streams. Thus, the best we could do in these cases is setting the first video as the video_thumbnail or -1 if there are no videos in the playlist.

I think first video thumbnail is fine, and in case there is no video the `drawable://...` one. I thought about other options but this seems like the best one to me.

PR #9523, opened 2022-12-09
Allow the user to permanently set a thumbnail

### LINKED ISSUE #9468: Enable users to permanently lock the thumbnail of a playlist
### Checklist

### Feature description

Right now, there is no feature that allows you to permanently set the thumbnail of a playlist.
Such a feature could be introduced by pressing on the three dots in the playlist. There you could have an option to permanently lock the thumbnail. If it is pressed, the thumbnail of the playlist won't change. If you then press "unlock thumbnail" (only visible if thumbnail is currently locked) it can be changed again.
If the thumbnail video is not in the playlist anymore the first video gets selected as the new thumbnail video.

I want my thumbnail to stay the same for a particular playlist. The current implementation would lead to the playlist thumbnail constantly changing and thus not always being representative of my chosen playlist topic.

### Additional information

I am a developer and would work on this issue myself.

Jared234:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

- added an an option to permanently set a thumbnail with the already existing "Set Thumbnail" feature in the long press menu
- changed the long press menu of the local playlists
    - Before: Rename/Delete dialog opens
    - After: AlertDialog opens with 3 options: `Rename`, `Delete`, `Unset Thumbnail`. `Rename` and `Delete` open the respective dialog. `Unset Thumbnail` is only enabled if the thumbnail has been permanently set. If you press this option, the permanent playlist thumbnail is unset and a new one gets chosen automatically (first video in the playlist or a default thumbnail if there is no video in the playlist). 

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

- Fixes #9468 

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
@@ -259,11 +259,42 @@ public final class BookmarkFragment extends BaseLocalListFragment<List<PlaylistL
     }
 
     private void showLocalDialog(final PlaylistMetadataEntry selectedItem) {
+        final String rename = getString(R.string.rename);
+        final String delete = getString(R.string.delete);
+        final String unsetThumbnail = getString(R.string.unset_playlist_thumbnail);
         final boolean isPlaylistThumbnailSet = localPlaylistManager
                 .getIsPlaylistThumbnailSet(selectedItem.uid);
+
         final AlertDialog.Builder builder = new AlertDialog.Builder(activity);
 
-        final ArrayAdapter<String> arrayAdapter = new ArrayAdapter<>(getContext(),
+        final ArrayAdapter<String> arrayAdapter = getLocalDialogArrayAdapter(isPlaylistThumbnailSet,
+                unsetThumbnail);
+        arrayAdapter.addAll(rename, delete, unsetThumbnail);
+
+        final DialogInterface.OnClickListener action = (dialog, index) -> {
+            if (index == arrayAdapter.getPosition(rename)) {
+                showRenameDialog(selectedItem);
+            } else if (index == arrayAdapter.getPosition(delete)) {
+                showDeleteDialog(selectedItem.name, localPlaylistManager
+                        .deletePlaylist(selectedItem.uid));
+                dialog.dismiss();
+            } else if (isPlaylistThumbnailSet) {
+                final String thumbnail_url = localPlaylistManager
+                        .getAutomaticPlaylistThumbnail(selectedItem.uid);
+                localPlaylistManager.changePlaylistThumbnail(selectedItem.uid, thumbnail_url, false)
+                        .observeOn(AndroidSchedulers.mainThread())
+                        .subscribe();
+            }
+        };
+
+        builder.setAdapter(arrayAdapter, action)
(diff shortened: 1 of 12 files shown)


Jared234:
Sorry for another failed test. All of them worked locally, but i forgot to add the unversioned file "6.json". I will fix this asap.

opusforlife2:
Should I run it now?

Jared234:
Yes, that would be great. I have really pushed everything now and tested it again locally. Everything should hopefully work now.

app/src/main/res/values/strings.xml
```diff
@@ -438,6 +438,7 @@
     <string name="mute">Mute</string>
     <string name="unmute">Unmute</string>
     <string name="set_as_playlist_thumbnail">Set as playlist thumbnail</string>
+    <string name="unset_playlist_thumbnail">Unset thumbnail</string>
```
Stypox, line 441: "Unset playlist thumbnail" would probably be less confusing

Stypox (changes requested):
Looks almost good to me, just two small issues

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
```diff
@@ -256,6 +259,60 @@ private void showRemoteDeleteDialog(final PlaylistRemoteEntity item) {
...
+        final String delete = getString(R.string.delete);
+        final String unsetThumbnail = getString(R.string.unset_playlist_thumbnail);
+        final boolean isThumbnailPermanent = localPlaylistManager
+                .getIsPlaylistThumbnailPermanent(selectedItem.uid);
+
+        final AlertDialog.Builder builder = new AlertDialog.Builder(activity);
+
+        final ArrayAdapter<String> arrayAdapter = getLocalDialogArrayAdapter(isThumbnailPermanent,
+                unsetThumbnail);
+        arrayAdapter.addAll(rename, delete, unsetThumbnail);
```
Stypox, line 272: Mmmh, after trying out the app I think it's better if you just don't add the "unsetThumbnail" entry instead of disabling it. At the moment the disabled item is still clickable and once clicked closes the dialog, which feels strange. Also, it's expected across the app for things to appear only when they do make sense.

app/src/main/res/values/strings.xml
Stypox: Soooorry, I mean "Unset permanent thumbnail"... I don't know why I wrote what I wrote :sweat_smile: :sweat_smile:

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
Stypox: I'd prefer to go with simplicity of implementation here, so that there is less code that may fail. As I said there is no big difference anyway from the user perspective (grayed out things are rarely useful to find out features, as they don't point to anything, imho). So go with the simple approach of just not adding the `unsetThumbnail` when it does not apply.

Stypox (approved):
Code looks good and seems to work fine. Thank you!

PR #9526, opened 2022-12-09
Add check for data saver

john-peterson:
This setting is called data saver for the user. (To match the equivalent setting battery saver.) I suppose anyone that use it would expect the lowest bitrate in the background. Unless i made a mistake with the logic this always use low bitrate when data saver is enabled regardless of resolutioin limit

I had no idea the resolution limit also set the background bitrate until i checked to add a setting for it. 160 or 50 kbit makes a big difference when listenting for along time.

A few comments on my use. I poersonally never use video. I will enable resolution limit now to get cheap audio. When i checked i had not enabled data saver but i enabled it now since i cant think of any reason i wouild not want it. maybe there was a reason that i forgot, like drive upl;loads not working or something. battery saver ius horrendoues but i think i like data saver fine it doesnt have any strange side effects. I would like to see more aggressive use of it like firefox not showing any ppictures for example would be perfect

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
@@ -3,6 +3,7 @@ package org.schabi.newpipe.util;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.net.ConnectivityManager;
+import android.os.Build;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


TobiGr:
Please do not ignore the template. We have put it there for a reason.

> Unless i made a mistake with the logic this always use low bitrate when data saver is enabled regardless of resolutioin limit

Did you test your changes?

From a brief look at the code, I'd assume that the check needs to be added on line 581. `getResolutionLimit()` is used in another method to retrieve the max resolution which should be reduced when the data saver is enabled.  If my previous assumption is correct, I'd suggest you to move the data saver check into a separate method for better readability.

https://github.com/TeamNewPipe/NewPipe/blob/5942add141ec34f0378806336c96e0ea81cfea75/app/src/main/java/org/schabi/newpipe/util/ListHelper.java#L573-L590

Isira-Seneviratne:
The check could be simplified using [`ConnectivityManagerCompat.getRestrictBackgroundStatus()`](https://developer.android.com/reference/androidx/core/net/ConnectivityManagerCompat#getRestrictBackgroundStatus(android.net.ConnectivityManager)).

AudricV:
Closing due to lack of author's response.

Issue #5352, opened 2021-01-05
Support for QUIC through chronet


FireMasterK:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want
https://developer.android.com/guide/topics/connectivity/cronet

I want to support quic though chronet (chromium's network stack), so we can replicate a chrome browser and bypass recaptchas.

#### Is your feature request related to a problem? Please describe it
Without quic, you may be limited to captchas as explained in https://github.com/iv-org/invidious/issues/957#issuecomment-576424042.

#### How will you/everyone benefit from this feature?
Lesser captchas for everyone who uses QUIC! (this does not include tor users as tor is tcp only)

FireMasterK:
There is a GMS version of cronet, and a non GMS version: 
```
"In cases where Cronet cannot be loaded from Google Play services, there is a less performant implementation of Cronet's API that can be used. To use this fall-back implementation, depend on org.chromium.net:cronet-fallback and call new JavaCronetProvider(context).createBuilder()."
```

We want to use the non GMS version ideally.

FireMasterK:
There also appears to be a cronet extension for exoplayer: https://github.com/google/ExoPlayer/tree/release-v2/extensions/cronet

eladkarako:
<a href="https://developer.android.com/guide/topics/connectivity/cronet">cronet</a> is awesome, it supports QUIC and <a href="https://tools.ietf.org/html/rfc7932">Brotli Compressing</a> seamlessly (JSON/manifest/xmls are compressed to almost nothing!!!), which is why I'm using it exclusively for almost a year now (instead of few Apache libs), connections are a LOT stable (especially for Google servers), and keepalive is FAR more efficiently implemented (various other libs kept giving me timeouts for long/slow downloads but this one didn't).

I personally don't mind the Google Play Services dependency, if you do (what's the official policy regarding it anyway?) the chromium-lib variation is a decent replacement (it <a href="https://github.com/GoogleChromeLabs/cronet-sample/commit/6f21eefe7a2787cf59b62bebecaa8a407cd4ec68">bloats the APK a bit</a> but nothing too noticeable).

Stypox:
[...]

PR #9530, opened 2022-12-11
Use Cronet for playback, simplify YouTube DataSource, and increase connection timeouts

AudricV:
#### What is it?

- [x] Feature (user facing)

This PR adds support of QUIC and Brotli content-encoding for playback, by using the [Cronet network stack](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/cronet/README.md).

It replaces usage of Android's network stack for playback (with `HttpUrlConnection`), which is OS version and device specific, and so can be easily outdated and can contain some issues that are not present on Cronet or OkHttp.

While OkHttp could be used as a network stack of ExoPlayer, I noticed when testing the usage of the OkHttp extension for playback that throttling issues with playback of YouTube contents may happen. That's the reason why Cronet is used instead, on which the issue does not happen, according to my tests since a few months.

The addition of this network library, embedded into the application, increases unfortunately the application size by around 10MB, because the Cronet library contains native code, shipped for all CPU architectures supported by Android's NDK (`arm64-v8a`, `armeabi-v7a`, `x86` and `x86_64`).

This **doubles the release application size**, due to the fact that the application is universal and not architecture dependant. I can enable split APKs support to allow creation of single architecture APKs (while also keeping an universal APK) in a separate pull request.

The QUIC protocol has a User-Agent ID property. On Cronet, it is composed of the name of the application using Cronet and ` Cronet/` + Cronet's version. This ID cannot be edited by third-party clients, unless they replace the code which sets this User-Agent ID, by using a custom `cronet-common` module, module on which Cronet's embedded variant depends.

In order to avoid services to fingerprint NewPipe with QUIC's User-Agent ID while being able to use QUIC protocol, a Gradle "hack" of the `cronet-common` module, which replaces the `UserAgent` class by our custom implementation, has been added, thanks to Stypox for that.

As `YoutubeHttpDataSource` was a copy of ExoPlayer's DefaultHttpDataSource, based on Android's network stack, which is now unused, it has been removed and the modifications made for YouTube playback on streaming URLs are now made using a [`ResolvingDataSource` `Resolver`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/upstream/ResolvingDataSource.Resolver.html), `YoutubeDataSourceResolver`.
    
A workaround is used in it to prevent ExoPlayer sending the Range HTTP header. If usage of the range query parameter/path parameter is enabled and that it doesn't already exist as a URL parameter, this header is converted into a range query parameter or "path parameter".

This may break playback if YouTube is serving streaming data with compression, which is unlikely to happen as:
- it is inefficient;
- Cronet doesn't allow editing the `Accept-Encoding` HTTP header (see https://groups.google.com/a/chromium.org/g/net-dev/c/VbWPhyGrxD4/m/VC71PP3DAQAJ for details about how Cronet handles content encoding)
- the YouTube official Android app uses it, according to reverse engineering that I made on a recent build of the app (the build was a 18 major one).

This resolver adds support of YouTube `videoplayback` URLs using "path parameters" instead of query parameters. For example, `https://YOUTUBE_VIDEO_HOST.com/videoplayback/PARAM_1/VALUE_1/PARAM_2/VALUE_2` is using "path parameters", when `https://YOUTUBE_VIDEO_HOST.com/videoplayback?PARAM_1=VALUE_1&PARAM_2=VALUE_2` query parameters. "Path parameters" are used for streaming URLs of DASH and HLS manifests.

This `Resolver` is compatible with multiple network stacks and is easier to maintain than the previous approach.

Timeouts of network sockets connect and write operations have been increased, in the app except images, from the default values (10 seconds for each operation) to 30 seconds. It is now also the case for playback, while this value where has been set where it was possible (connect and read operations, their default value is 8 seconds in ExoPlayer).

#### Known issues
If you switch from a network allowing QUIC connections to a one blocking the protocol, you may encounter network errors. There is nothing I can do to prevent this, except adding switches to disable QUIC on the app side.

- #5352 for playback

#### APK testing

#### Due diligence


opusforlife2:
> enable split APKs support 

Is this something we can do on our own or is coordination with F-Droid devs needed?

opusforlife2:
From reading the PR description, I don't really know what benefit the user is getting out of this, except for longer timeouts.

opusforlife2:
- Videos start nearly instantly, definitely much faster than 0.24.1 even in the worst cases.
- On average, it also gets stuck less often when jumping to random timestamps in a 1080p60 video.

mhmdanas:
> Is this something we can do on our own or is coordination with F-Droid devs needed?

I believe what this would require, at least on the F-Droid side, is merely some changes to the F-Droid build manifest. You can look at some MRs to split APKs based on architecture for guidance ([here's one](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/12037/diffs)). Asking the F-Droid devs is still probably a good idea, just to get some help.

app/src/main/java/org/schabi/newpipe/player/datasource/YoutubeDataSourceResolver.java
```diff
@@ -0,0 +1,234 @@
...
+            final String rangeParameter = buildRangeParameter(
+                    originalDataSpec.position, originalDataSpec.length);
+            if (rangeParameter != null) {
+                // Workaround: prevent ExoPlayer HttpDataSources sending a Range HTTP header by
+                // setting the DataSpec position to 0 and its length to unset
+                // See HttpUtil#buildRangeRequestHeader(long, long)
+                // If the content is served with compression, this workaround may make ExoPlayer
+                // not work properly
+                dataSpecBuilder.setPosition(0);
+                dataSpecBuilder.setLength(C.LENGTH_UNSET);
```
Stypox, line 141: What happens if you were to remove these two lines? You would start sending the range header, but why is that unwanted? According to what you wrote above: "the server responds correctly when requesting a range of a resource with the header".

app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
```diff
@@ -179,4 +184,22 @@ public Response execute(@NonNull final Request request)
...
+                    // by default
+                    .enableQuic(true)
+                    .enableHttp2(true)
+                    .build();
+        }
+    }
+
+    public CronetEngine getCronetEngine() {
+        return cronetEngine;
+    }
```
Stypox, line 204:
```suggestion
    public void getCronetEngine() {
        if (cronetEngine == null) {
            cronetEngine = new NativeCronetProvider(App.getApp().getApplicationContext())
                    .createBuilder()
                    .setUserAgent(USER_AGENT)
                    .enableBrotli(true)
                    // Be sure that QUIC and HTTP/2 are enabled, even it should be already the case
                    // by default
                    .enableQuic(true)
                    .enableHttp2(true)
                    .build();
        }
        return cronetEngine;
    }
```

Stypox (changes requested):
As a first option it's fine, but in the end I think we need to find a better way to build the cronet-common module without having a complex updating mechanism and without including so much code here. Here are a few options I could think of:
- We can set up a GitHub action that automatically does the updating when we tell it to do so. In the end we want to copy over everything from the `cronet-common` to our own `cronet-common` except from the `UserAgent` class
- The best option would be to find some Gradle way of excluding things. Maybe some of the following links might be useful. I looked for such a method the last time we talked about this on IRC but couldn't find anything. Now I have searched for "gradle exclude specific class from library" on DuckDuckGo and found some promising results.
  - https://stackoverflow.com/questions/38383440/gradle-exclude-java-class-from-lib-replaced-by-own-class-to-avoid-duplicate
  - https://stackoverflow.com/questions/36473673/how-to-exclude-particular-class-from-a-jar-in-gradle
  - https://stackoverflow.com/questions/33400774/gradle-how-to-exclude-a-class-package-from-dependency-jar-file

Overall I really like the simplicity and the structure of this PR. Thank you!

app/src/main/java/org/schabi/newpipe/player/datasource/YoutubeDataSourceResolver.java
AudricV: We would send the HTTP header and the query parameter in this case, which is something uncommon and could be used by YouTube to detect us.

YouTube clients send either the header (rarely used) or the query parameter (mostly used).

Stypox:
After almost 4 hour of going crazy, I was able to do it in Gradle. Will push code soon.

Stypox:
I pushed two commits: the first one removes the files and will need to be squashed once we decide what to do; the second one actually implements the gradle tasks.

app/build.gradle
```diff
@@ -177,6 +182,38 @@ sonar {
...
+}
+// these three tasks are actually just a chain of tasks that are executed one after the other
+task cronetCommonNoUserAgentLib(type: Copy) {
+    from configurations.cronetCommon
+    into (cronetCommonPath)
+    from zipTree("$cronetCommonPath/cronet-common-${rootProject.ext.cronetVersion}.aar")
+    into (cronetCommonPath)
+    from zipTree("$cronetCommonPath/classes.jar")
+    into (cronetCommonPath)
+    // we want every class except the UserAgent class, since we want to define that ourself
```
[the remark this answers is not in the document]
Stypox: Yeah it was kind of a "myself" but with "us" 😂

Stypox (commented):
I pushed a commit that prevents even more leakage of user information. Now even the default user agent is `DownloaderImpl.USER_AGENT`. We would overwrite that field manually anyway by calling `setUserAgent()` while building the engine, but let's be completely sure ;-)

Other code looks good to me and I think this PR is ready, but we need to test it thoroughly. If you haven't already done so, maybe ask people on reddit to test it, or also on the #newpipe channel.

TobiGr:
@opusforlife2 The crash is unrelated. See #9692

AudricV:
As we plan to rewrite the app and change the network stack used in it for all network requests (extraction, image loading, downloader, playback, ...), shouldn't I close this PR and wait for the relevant network stack, downloader and player changes in the rewrite process?

AudricV:
Closing this PR. We are not 100% sure if we want to use this network stack, and YouTube uses a custom format called UMP, which we should use in the future. It would require a custom ExoPlayer `DataSource` which makes the related simplifications in this PR useless.

Anyway, this PR gave the procedure we need to use to prevent Cronet to send the QUIC user agent ID, which is not a trivial task. This can be reused in a future PR about this network stack.

Connection timeouts can be increased in a separate PR, if we want to do so.

PR #9538, opened 2022-12-12
Handle duplicate streams in the "Add to playlist" dialog

### LINKED ISSUE #4186: Show a warning before adding a duplicate video to a playlist

Jared234:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

- Indicator added to the "Add to playlist" dialog to alert users if they are attempting to add a duplicate stream
- AlertDialog opens if a user tries to add the same stream to the same playlist, allowing them to confirm or cancel their action

This helps prevent users from accidentally adding the same stream multiple times to their playlist, but stills allows them do it if they intend to do so. 

- Before: The "Add to playlist" dialog did not handle duplicate streams properly. When a user attempted to add a duplicate stream, they were not warned or given any indication that the stream was a duplicate. This allowed users to unknowingly add the same stream multiple times to their playlist.
- After:
If a user attempts to add a stream to a playlist twice, the playlist will be grayed out to indicate that it is a duplicate: 

Screenshot_20221212-225124

If a user attempts to add a stream to their playlist that is already in the playlist anyway, a dialog will appear to alert them that the stream is a duplicate and ask them if they want to proceed with adding it to the playlist:

Screenshot_20221212-225111

- Fixes #4186 

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/LocalItemListAdapter.java
+++ b/app/src/main/java/org/schabi/newpipe/local/LocalItemListAdapter.java
@@ -11,6 +11,7 @@ import androidx.recyclerview.widget.GridLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 
 import org.schabi.newpipe.database.LocalItem;
+import org.schabi.newpipe.database.playlist.PlaylistMetadataEntry;
 import org.schabi.newpipe.database.stream.model.StreamStateEntity;
 import org.schabi.newpipe.local.history.HistoryRecordManager;
 import org.schabi.newpipe.local.holder.LocalItemHolder;
@@ -342,6 +343,16 @@ public class LocalItemListAdapter extends RecyclerView.Adapter<RecyclerView.View
         } else {
             onBindViewHolder(holder, position);
         }
+    }
+
+    @Override
+    public long getItemId(final int position) {
+        final LocalItem item = localItems.get(0);
+        if (item != null && item.getLocalItemType() == LocalItem.LocalItemType.
+                PLAYLIST_LOCAL_ITEM) {
+            return ((PlaylistMetadataEntry) item).uid;
+        }
+        return super.getItemId(position);
     }
 
     public GridLayoutManager.SpanSizeLookup getSpanSizeLookup(final int spanCount) {
--- a/app/src/main/java/org/schabi/newpipe/local/LocalItemListAdapter.java
+++ b/app/src/main/java/org/schabi/newpipe/local/LocalItemListAdapter.java
@@ -347,7 +347,7 @@ public class LocalItemListAdapter extends RecyclerView.Adapter<RecyclerView.View
 
     @Override
     public long getItemId(final int position) {
-        final LocalItem item = localItems.get(0);
+        final LocalItem item = localItems.get(position);
         if (item != null && item.getLocalItemType() == LocalItem.LocalItemType.
                 PLAYLIST_LOCAL_ITEM) {
             return ((PlaylistMetadataEntry) item).uid;
--- a/app/src/main/java/org/schabi/newpipe/local/LocalItemListAdapter.java
+++ b/app/src/main/java/org/schabi/newpipe/local/LocalItemListAdapter.java
@@ -11,7 +11,6 @@ import androidx.recyclerview.widget.GridLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 
 import org.schabi.newpipe.database.LocalItem;
-import org.schabi.newpipe.database.playlist.PlaylistMetadataEntry;
 import org.schabi.newpipe.database.stream.model.StreamStateEntity;
 import org.schabi.newpipe.local.history.HistoryRecordManager;
 import org.schabi.newpipe.local.holder.LocalItemHolder;
@@ -343,16 +342,6 @@ public class LocalItemListAdapter extends RecyclerView.Adapter<RecyclerView.View
         } else {
             onBindViewHolder(holder, position);
         }
-    }
-
-    @Override
-    public long getItemId(final int position) {
-        final LocalItem item = localItems.get(position);
-        if (item != null && item.getLocalItemType() == LocalItem.LocalItemType.
-                PLAYLIST_LOCAL_ITEM) {
-            return ((PlaylistMetadataEntry) item).uid;
-        }
-        return super.getItemId(position);
     }
 
     public GridLayoutManager.SpanSizeLookup getSpanSizeLookup(final int spanCount) {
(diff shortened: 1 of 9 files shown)


Jared234:
I was wondering if it would be possible to rerun the pipeline. All of the tests work locally, and in this pipeline the operation was cancelled without any error messages. I don't think there is a problem with the code, so could you please rerun it @AudricV?

Thank you for your help.

opusforlife2:
I don't think the AlertDialog is needed? Since the user has been warned that the item will be duplicated, and there is a visual indication as well, the dialog seems overkill.

Also, for users who intentionally want to add duplicates, this would be an annoying extra step each and every time they go through with it.

Instead of an AlertDialog, I would rather go with a toast saying "Duplicate added", which also informs the user that they did indeed go ahead and add a duplicate, but doesn't need a tap for confirmation. For users who ignore/don't read the warning about the greyed out playlists, the toast would act as a second warning.

Jared234:
I see your point. The only downside to removing this dialog would be that users would not know how many duplicate videos are already in the playlist. At first glance, though, I don't know why anyone would need that information.... So I will change it.

Jared234:
I removed the duplicate dialog and if a video was already at least once in the playlist I replaced the toast message "Playlisted" with "Playlisted duplicate".

opusforlife2:
> how many duplicate videos are already in the playlist

"Duplicate added `x` time(s)"

There you go, if needed.

app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
```diff
@@ -26,8 +26,12 @@
 public final class PlaylistAppendDialog extends PlaylistDialog {
     private static final String TAG = PlaylistAppendDialog.class.getCanonicalName();
 
+    private static final float DEFAULT_ALPHA = 1f;
+    private static final float GRAYED_OUT_ALPHA = 0.3f;
```
Stypox, line 30: 0.6 would also be fine I think. Greying out too much could make text not readable anymore.

app/src/main/res/values/strings.xml
```diff
@@ -432,6 +432,7 @@
     <string name="preferred_player_fetcher_notification_message">"Loading requested content"</string>
     <!-- Local Playlist -->
     <string name="create_playlist">New Playlist</string>
+    <string name="duplicate_in_playlist">The playlists that are grayed out already contain this item.</string>
```
Stypox, line 435:
```suggestion
    <string name="duplicate_in_playlist">The playlists that are grayed out already contain this item</string>
```

app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
```diff
@@ -26,8 +26,12 @@
...
+
     private RecyclerView playlistRecyclerView;
     private LocalItemListAdapter playlistAdapter;
+    private List<Long> duplicateIds;
```
Stypox, line 34: Mmmh, I don't like this approach. It is slow, since at the moment queries run in `O(n)`, while with this code showing the list would require `O(n²)` time (where `n` is the number of playlists), since for each playlist being shown you need to check if its id is among those duplicate. Can't you just return that information in the `getPlaylists` database query, after changing it to take the video id as parameter (or -1 in order not to show duplicate information in other places where playlists are shown)? You could add a `boolean wouldVideoBeDuplicate` field to `PlaylistMetadataEntry` items, that is not stored in the database but is instead returned only in queries. Then you can use that field in local item views to change the alpha of the whole item view.

Stypox (changes requested):
Thank you! I agree with @opusforlife2 on the fact that a dialog is not needed and a toast is better. Let me know if the changes I requested for the list view are comprehensible.

app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
Jared234: Thanks for your input. However, after having another look, I have another suggestion.
When you refer to the `O(n^2)` time required, you probably talk about the the checks in the `showDuplicateIndicators` method. If I adjust this method a bit, the loop within this method is only repeated a maximum of 10 times (only checking the currently visible playlists). In this loop, their Id gets only compared to the duplicates, which is usually a really low number (`duplicates << playlists`), since 99% of people won't add one video to all of their playlists.

Furthermore, the current implementation would be faster if the are no duplicates at all (which is the large majority of all requests), right? Since you can simply check if the list is empty before you go through any of the playlist items.
IMO, with a simple change this would be performant and keep us from changing the `getPlaylist` query and adding another field to the `PlaylistMetadataEntry`.

What's your opinion? Is there anything I missed?

Stypox: I pushed a commit containing what I mean. It would have been harder to explain than to do ;-)

I agree that your approach wouldn't really be slow, but it also has to access the list from various unusual places. The solution I pushed I think is more neat.

app/src/main/java/org/schabi/newpipe/local/LocalItemListAdapter.java
```diff
@@ -344,6 +345,16 @@ public void onBindViewHolder(@NonNull final RecyclerView.ViewHolder holder, fina
         }
     }
 
+    @Override
+    public long getItemId(final int position) {
```
Stypox, line 349: (probably can be removed)

app/src/main/java/org/schabi/newpipe/local/dialog/PlaylistAppendDialog.java
Jared234: I just looked at you changes and agree that your solution is preferable. However, at the moment there is no text that explains what a grayed out playlist actually means. I initially created an invisible explainer text (`NewPipeTextView`) that you can set to visible if there are duplicates. Did you think this was unnecessary or did you just forget to include it in your current solution approach?

Stypox: I forgot to include it, sorry

Jared234: Ok, I can add this and your other suggestions. Just one more question: Do you think using a for loop until you find the first one with a duplicate is performant enough? Since `PlaylistDuplicatesEntry` has the field `timesStreamIsContained`, it should be in O(N) (worst case). And since this field is not stored in the database, a query wouldn`t be faster as well.

Stypox: Yeah, that's fine, I doubt there is any clever trick (and also O(n) is perfectly ok and does not increase complexity if the data has already been collected in O(n) anyway)

Stypox (approved):
I pushed two more small commits and tested thoroughly. It seems to work well. Thank you!

PR #9608, opened 2022-12-30
Place downloaded files in default folder without file permission

john-peterson:
Thats how it should wok in my experience. Solid explorer has full access and the others do zero. the master does everything. they just work

i updated newpipe a year ago and downloads were crashing incomprehensibly. something was more complicated than it had to be. the old settings had not been tested and a change in the published release crashed for everyone with previous settings and certain os versions.

this beautiful method was introduced in api 19. works like magic. no clutter. no bugs. it just works. i still have apps that cling on the the legacy override and clutter my homefgolder with useless empty folders. apps lovew to make empty folders for no reason. and everyone thinks their app is the only thing the os will ever servce. most people dont car e much about your app and just want it to do no harm

the gui settings still have to be removed. i am only one man only haver so much ork. 

if the storage backup can be removed it will completely remove all traces of file pickers and permissions since api 19. and homespace horrow https://blog.codinghorror.com/dont-pollute-user-space/

i certainly store files. notes. but no browser data. i dont know why someone would collect playlists et c and guard them like a treasure. i would remove it to purge all home space pollution or traces of it

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -6,7 +6,6 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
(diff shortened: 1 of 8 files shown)


Stypox:
I am sorry but I strongly disagree with many of your points and most users will disagree even more than me. Before making (drastic) changes you should open an issue and discuss it with users and maintainers, as explained in the contribution guidelines.

PR #9631, opened 2023-01-05
Update NewPipeExtractor and properly linkify comments

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Update the extractor to the latest commit, with breaking changes from https://github.com/TeamNewPipe/NewPipeExtractor/pull/987: now comments have a `Description` and not a `String` anymore
- Changes in `TextLinkifier`:
  - Move the method that determines how to linkify a description based on its type from `DescriptionFragment` to `TextLinkifier`
  - Rename methods in `TextLinkifier` to not contain the redundant "link" word already contained in "Linkifier"
  - Allow passing a callback to `TextLinkifier` methods to be notified when the text has been set to the `TextView`, needed to set ellipsis and movement method in comments
  - Since `TextLinkifier` only needs stream url and service id to generate popup-timestamps and hashtags, and since comments only have that information and not a full `Info`, now the `TextLinkifier` methods take those two parameters instead of an `Info.
- Rework the comments touch handling because it wouldn't work anymore since the links are now linkified with a different approach than before (I don't know why it stopped working)
  - Now calling `getText()` does not return a `Spannable`, but rather a `Spanned` (even though in `TextLinkifier` a `SpannableBuilder` is used). This means that the text selection (i.e. the highlight that appears when tapping on links) can't be manually set by us as done before in `CommentTextOnTouchListener`, unfortunately.
  - Note that long clicks still don't work, I'll leave it to a future PR. The same logic as `LongPressLinkMovementMethod` needs to be implemented in `CommentTextOnTouchListener`.
- Rework comment ellipsis handling, since it added ... by cutting away too many characters in case of links, and also because it made the text view size smaller on some cases for some reason (?)
  - Now links are not clickable when the comment is ellipsized (this should have been the case before, too, but it used to be buggy)
  - Only the last 2 characters are chopped off, and are only chopped off if there would not already be space for the ellipsis

It definitely fixes some problems with comments, but I couldn't find any related issue.

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
+++ b/app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
@@ -1,9 +1,10 @@
 package org.schabi.newpipe.info_list.holder;
 
+import android.graphics.Paint;
+import android.text.Layout;
 import android.text.TextUtils;
 import android.text.method.LinkMovementMethod;
 import android.text.style.URLSpan;
-import android.text.util.Linkify;
 import android.util.Log;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/schabi/newpipe/info_list/holder/CommentsMiniInfoItemHolder.java
```diff
@@ -91,18 +110,19 @@ public void updateFromItem(final InfoItem infoItem,
 
         itemThumbnailView.setOnClickListener(view -> openCommentAuthor(item));
 
+        try {
+            streamService = NewPipe.getService(item.getServiceId());
+        } catch (final ExtractionException e) {
+            // should never happen
+            Log.w(TAG, "Cannot obtain service from comment service id, defaulting to YouTube", e);
```
TobiGr, line 117: Shouldn't we display an error snackbar? Otherwise, we won't notice the bug itself, but might get complains about wrong behavior.

Stypox: Right, done

TobiGr (dismissed):
Code looks good. I tested briefly and did not find any bugs related to the changed linking.

There are a few bugs related to other extractor PRs which are introduced to NP in this PR, but are fixed separately. The extractor version can be updated independently.

Stypox:
I'm merging this now so that we can have it in the RC

TobiGr:
we should also update the extractor commit. otherwise, there will be reports about broken comments

Issue #4478, opened 2020-10-10
Always show the full screen button


nbmrjuhneibkr:
### Checklist

- [x] I checked, but didn't find any duplicates of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want
In v0.20, full screen button is hidden when automatic screen rotation is enabled in the OS.  
The button should always be accessible, regardless of the system settings.

#### Is your feature request related to a problem? Please describe it
When the device lies on any flat surface and auto-rotate is enabled in the OS, there is no way to play videos in full screen mode without picking the device up to trigger the sensor, or disabling auto-rotate.

#### Additional context
A user may want to watch videos without picking the device up. Disabling auto-rotation globally is not an acceptable solution.

#### How will you/everyone benefit from this feature?
When combined with the manual video rotation control (which should also be restored), this will restore all manual controls that were available in NewPipe before the 0.20 update.

avently:
I disagree with the issue. Your example with flat surface is like "I need to say something related to the issue". 
I can agree with some of your arguments related to fullscreen button in case of locked orientation but not unlocked.
Whenever I see disappeared fullscreen button it is a sign of enabled autorotation. And I'm 100% sure that I can rotate the device and get whatever orientation I want. With always shown button I need to use it or always look at status bar's panel (which is not an option on android 4.4 because the autorotation button isn't present there). This is something I know usefulness of.

nbmrjuhneibkr:
I disagree with your disagreement. Your responses often look like "I need to find a reason to dismiss this issue".

Sometimes I keep my phone on a table and play YouTube videos on it while doing work. I didn't have to pick up the phone to change the orientation before v0.20.
Please stop forcing your personal usage scenarios on everyone else.

>With always shown button I need to use it or always look at status bar's panel (which is not an option on android 4.4 because the autorotation button isn't present there).

Then why not have two slightly different icons? This is an easy fix.

avently:
@nbmrjuhneibkr 
>Please stop forcing your personal usage scenarios on everyone else.

Please stop saying what I should do and I will not say where you need to go.

nbmrjuhneibkr:
I'm only mentioning things that have direct impact on the app and the way people use it.

[...]

PR #9693, opened 2023-01-16
Orientation is locked if there is no sensor for it

Redirion:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)

This https://github.com/TeamNewPipe/NewPipe/issues/4478#issuecomment-1382828147 in the ever lasting full screen button discussion is actually valid. If a device does not feature an accelerometer (STB, Terminal, TV), the screen cannot rotate. It is more some sort of "sanity" check, as "def_accelerometer_rotation" in the SettingsProvider in AOSP is "false", which means, that after a factory reset, auto rotation of the screen should be disabled anyway. So a device manufacturer must have deliberately broken it (actually a device bug, not a bug in NewPipe) or a user must have selected screen auto rotation, even though it doesn't work. So this probably affects only a handful of users of NewPipe.

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
@@ -14,6 +14,7 @@ import static java.lang.annotation.RetentionPolicy.SOURCE;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
 import android.provider.Settings;
 import android.view.accessibility.CaptioningManager;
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
```diff
@@ -382,8 +383,11 @@ public static void setScreenBrightness(@NonNull final Context context,
     public static boolean globalScreenOrientationLocked(final Context context) {
         // 1: Screen orientation changes using accelerometer
         // 0: Screen orientation is locked
+        // if the accelerometer sensor is missing completely, assume locked orientation
         return android.provider.Settings.System.getInt(
-                context.getContentResolver(), Settings.System.ACCELEROMETER_ROTATION, 0) == 0;
+                context.getContentResolver(), Settings.System.ACCELEROMETER_ROTATION, 0) == 0
```
Stypox, line 388: I wonder what the `getInt` returns when there is no accelerometer for rotation... Shouldn't it already return 0, as it's the default value passed for when the stored value is not valid?

Stypox (approved):
Lgtm. Let's see if it works for users :-)

app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
Redirion: the setting exists in aosp. The manufacturer would have to deliberately remove the setting. So it most likely is there. And if not: all the guys with their realy odd hardware are happy, that the button is there for them

PR #9706, opened 2023-01-20
Fixed a bug that caused the background player to stop working

### LINKED ISSUE #9131: Start playing in background after clicking background in playlists bug
### Checklist

### Affected version

0.24.0

1. Open any playlist 
2. Click on background header 
3. Long press on any video except first one(which is already playing) select 'start playing in background'

### Expected behavior

The selected video should be played in background.

### Actual behavior

The first element(video/shorts) continue to play in background and no change occurs. The player does not register the aforesaid user command.

_No response_

### Logs

_No response_

Color OS v6.0

### Affected device model

Oppo Realme 1 CPH1861

### Additional information

_No response_

Jared234:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)

The bug was caused by an incorrect comparison between the existing `PlaylistQueue` and the new `PlaylistQueue`. In the `handleIntent` method in the `Player` class, both are compared, and if they are considered equal, the existing `PlaylistQueue` remains unchanged (because why reinitialize the same queue?). However, the overridden equality check was incorrect, as it only checks if the streams are the same, without considering the `queueIndex`. By inserting this check, the bug is fixed and only if the queues are actually identical, the existing queue is not reinitialized anymore.
I also had to change a test that tests if the equality check disregards the index. However, IMO if you do that the "equal" check doesn't make sense, since these queues cannot be considered equal.

- Fixes #9131 

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
@@ -527,6 +527,9 @@ public abstract class PlayQueue implements Serializable {
         if (size() != other.size()) {
             return false;
         }
+        if (other.getIndex() != getIndex()) {
+            return false;
+        }
         for (int i = 0; i < size(); i++) {
             final PlayQueueItem stream = streams.get(i);
             final PlayQueueItem otherStream = other.streams.get(i);
--- a/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
+++ b/app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
@@ -518,16 +518,11 @@ public abstract class PlayQueue implements Serializable {
      * This method also gives a chance to track history of items in a queue in
      * VideoDetailFragment without duplicating items from two identical queues
      */
-    @Override
-    public boolean equals(@Nullable final Object obj) {
-        if (!(obj instanceof PlayQueue)) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueue.java
```diff
@@ -527,6 +527,9 @@ public boolean equals(@Nullable final Object obj) {
         if (size() != other.size()) {
             return false;
         }
+        if (other.getIndex() != getIndex()) {
```
Stypox, line 530: This method is also used in other places where checking for the index is not wanted, e.g. in `VideoDetailFragment`. I would completely remove the `equals()` and `hashCode()` methods and create two new methods: `equalStreams(PlayQueue)` and `equalStreamsAndIndex(PlayQueue)`. Then the two methods can be called when needed, providing an easier to read code, and also preventing casting from `Object`.

Stypox (changes requested):

Jared234:
The "SonarCloud Quality Gate" failed. However, the "bug" it found cannot occur. If you take a look at the `equalStreams` method, you can see that it checks if `other` is null. Therefore, the `getIndex()` method can only be called on `other` in the `equalStreamsAndIndex` method if it is not null.  
Is there any way to ignore these kinds of issues? Or should I implement an additional check?

Stypox (dismissed):
LGTM, thank you!

Stypox (approved):

PR #9708, opened 2023-01-20
Created a setting to switch the sides of volume and brightness

### LINKED ISSUE #1857: Allow configuring volume and brightness sides
Some other apps use the exact opposite sides for brightness and volume control as this app so perhaps the two sides should be configurable.
(Same issue as https://github.com/ram-on/SkyTube/issues/387)

Marius1501:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Added a setting in Video and Audio settings under behavior
- When you switch it on the sides of volume and brightness gesture are switched
- The toggle is disabled if one of the gesture control toggles above is switched off

<table>
  <tr>
    <th>Normal Video and Audio settings</th>
    <th>When one toggle is switched off</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

- Fixes #1857
#### Due diligence


app/src/main/java/org/schabi/newpipe/player/gesture/MainPlayerGestureListener.kt
```diff
@@ -193,10 +194,16 @@ class MainPlayerGestureListener(
         isMoving = true
 
         // -- Brightness and Volume control --
-        val isBrightnessGestureEnabled = PlayerHelper.isBrightnessGestureEnabled(player.context)
-        val isVolumeGestureEnabled = PlayerHelper.isVolumeGestureEnabled(player.context)
+        var isBrightnessGestureEnabled = PlayerHelper.isBrightnessGestureEnabled(player.context)
+        var isVolumeGestureEnabled = PlayerHelper.isVolumeGestureEnabled(player.context)
```
Stypox, line 198: Why `var`?

app/src/main/java/org/schabi/newpipe/player/gesture/MainPlayerGestureListener.kt
```diff
@@ -193,10 +194,16 @@ class MainPlayerGestureListener(
...
+        var displaySide = DisplayPortion.LEFT_HALF
+        val sidesSwitched = PreferenceManager.getDefaultSharedPreferences(player.context)
+            .getBoolean(R.string.switch_gesture_sides_key.toString(), false)
+        if (sidesSwitched) {
+            displaySide = DisplayPortion.RIGHT_HALF
+        }
```
Stypox, line 204: I'd do:
```suggestion
        val brightnessSide = if (PreferenceManager.getDefaultSharedPreferences(player.context)
            .getBoolean(R.string.switch_gesture_sides_key.toString(), false)) DisplayPortion.RIGHT_HALF
            else DisplayPortion.LEFT_HALF
```

app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
```diff
@@ -21,6 +24,17 @@
...
 
+    @Override
+    public void onViewCreated(@NonNull final View rootView,
+                              @Nullable final Bundle savedInstanceState) {
+        super.onViewCreated(rootView, savedInstanceState);
+        findPreference(getString(R.string.switch_gesture_sides_key))
+            .setEnabled(getPreferenceManager().getSharedPreferences()
+                    .getBoolean(getString(R.string.volume_gesture_control_key), true)
+                    && getPreferenceManager().getSharedPreferences()
+                    .getBoolean(getString(R.string.brightness_gesture_control_key), true));
```
Stypox, line 35: I think you can replace this custom code with two preference dependencies in the XML file using `android:dependency`.

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/settings/VideoAudioSettingsFragment.java
Marius1501: I don't think that this works because you can only set one `android:dependency` in the XML file.

app/src/main/java/org/schabi/newpipe/player/gesture/MainPlayerGestureListener.kt
Marius1501: I made this change but now the function doesn't work anymore. The sides ar not switched when the toggle is switched on.

Marius1501:
I fixed a small bug. Now everything should be fine.

Stypox:
Mmmh, after thinking about it, I think it's better to have two list options ("Gesture for left side", "Gesture for right side"), a where each option allows choosing amongst "None", "Audio control", "Brightness control". This also means that the user will be able to use one half of the screen for volume, and leave the other half without any gesture, which may help in some situations. This design is also future-proof if we decide to add more possible gesture actions.

Marius1501:
grafik

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -2038,7 +2038,10 @@ private void setupBrightness() {
             restoreDefaultBrightness();
         } else {
             // Do not restore if user has disabled brightness gesture
-            if (!PlayerHelper.isBrightnessGestureEnabled(activity)) {
+            if (!(PlayerHelper.getRightSideGesture(activity)
+                    .equals(getString(R.string.right_brightness_control_key))
+                    && PlayerHelper.getLeftSideGesture(activity)
```
Stypox, line 2043: Shouldn't this be an `||`?

app/src/main/res/values/settings_keys.xml
```diff
@@ -191,6 +189,38 @@
...
+        <item>@string/left_brightness_control_key</item>
+        <item>@string/left_volume_control_key</item>
+        <item>@string/left_none_control_key</item>
+    </string-array>
+
+    <string name="right_gesture_control_key">right_gesture_control</string>
+    <string name="default_right_gesture_control_value">@string/right_volume_control_key</string>
+    <string name="right_brightness_control_key">right_brightness_control</string>
+    <string name="right_volume_control_key">right_volume_control</string>
+    <string name="right_none_control_key">right_none_control</string>
```
Stypox, line 212: Why did you duplicate these? I'd just call them "brightness_control_key", "volume_control_key" and "none_control_key" and use them both for right and left.

Stypox (changes requested):
I feel like "left/right gesture control" is not easy to understand. What about "Action for gesture on left/right side"?

app/src/main/java/org/schabi/newpipe/player/gesture/MainPlayerGestureListener.kt
```diff
@@ -193,18 +193,23 @@ class MainPlayerGestureListener(
...
         // -- Brightness and Volume control --
-        val isBrightnessGestureEnabled = PlayerHelper.isBrightnessGestureEnabled(player.context)
-        val isVolumeGestureEnabled = PlayerHelper.isVolumeGestureEnabled(player.context)
-        if (isBrightnessGestureEnabled && isVolumeGestureEnabled) {
-            if (getDisplayHalfPortion(initialEvent) === DisplayPortion.LEFT_HALF) {
-                onScrollBrightness(distanceY)
-            } else /* DisplayPortion.RIGHT_HALF */ {
-                onScrollVolume(distanceY)
+        val rightSide = PlayerHelper.getRightSideGesture(player.context)
+        val leftSide = PlayerHelper.getLeftSideGesture(player.context)
```
Stypox, line 197: Move these two directly inside the `when`s below

Stypox (changes requested):

Stypox:
Did you think of a better name&description for the setting?

app/src/main/res/values/strings.xml
```diff
@@ -99,10 +99,13 @@
...
     <string name="auto_queue_summary">Continue ending (non-repeating) playback queue by appending a related stream</string>
     <string name="auto_queue_toggle">Auto-enqueuing</string>
-    <string name="volume_gesture_control_title">Volume gesture control</string>
-    <string name="volume_gesture_control_summary">Use gestures to control player volume</string>
-    <string name="brightness_gesture_control_title">Brightness gesture control</string>
-    <string name="brightness_gesture_control_summary">Use gestures to control player brightness</string>
+    <string name="left_gesture_control_summary">Change the left gesture control</string>
+    <string name="left_gesture_control_title">Left gesture control</string>
+    <string name="right_gesture_control_summary">Change the right gesture control</string>
+    <string name="right_gesture_control_title">Right gesture control</string>
```
Stypox, line 105: I mainly meant from the user perspective:

> I feel like "left/right gesture control" is not easy to understand. What about "Action for gesture on left/right side"?

```suggestion
    <string name="left_gesture_control_summary">Choose the action for gestures on the left side of the player</string>
    <string name="left_gesture_control_title">Left gesture action</string>
    <string name="right_gesture_control_summary">Choose the action for gestures on the right side of the player</string>
    <string name="right_gesture_control_title">Right gesture action</string>
```

@opusforlife2 do you have any suggestion?

Stypox (changes requested):

app/src/main/res/values/strings.xml
Marius1501: @opusforlife2 I used your suggestion for the summarys. @Stypox I used your suggestion for the titles.

app/src/main/res/values/settings_keys.xml
```diff
@@ -16,8 +16,6 @@
     <string name="use_external_video_player_key">use_external_video_player</string>
     <string name="use_external_audio_player_key">use_external_audio_player</string>
 
-    <string name="volume_gesture_control_key">volume_gesture_control</string>
-    <string name="brightness_gesture_control_key">brightness_gesture_control</string>
```
Stypox, line 20: Since you are replacing these two settings with two other settings, I think you should create a settings migration to make sure the previous preferences are still respected after an app update.

app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
```diff
@@ -2038,7 +2038,10 @@ private void setupBrightness() {
             restoreDefaultBrightness();
         } else {
             // Do not restore if user has disabled brightness gesture
-            if (!PlayerHelper.isBrightnessGestureEnabled(activity)) {
+            if (PlayerHelper.getActionForRightGestureSide(activity)
+                    .equals(getString(R.string.brightness_control_key))
+                    || PlayerHelper.getActionForLeftGestureSide(activity)
+                    .equals(getString(R.string.brightness_control_key))) {
```
Stypox, line 2044: I think you have inverted the conditions here. The brightness should be **not** restored only if **both** gestures are **not** brightness.
```suggestion
            if (!PlayerHelper.getActionForRightGestureSide(activity)
                    .equals(getString(R.string.brightness_control_key))
                    && !PlayerHelper.getActionForLeftGestureSide(activity)
                    .equals(getString(R.string.brightness_control_key))) {
```

Stypox (changes requested):

app/src/main/res/values/settings_keys.xml
Marius1501: Alright. I am sorry, I have never done something like that before. Could you give me a hint how to implement the settings migration? Thank you in advance!

Stypox: The instructions are in the [SettingsMigrations](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java) file

Marius1501: Thank you!

app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
```diff
@@ -108,6 +108,39 @@ protected void migrate(final Context context) {
         }
     };
 
+    public static final Migration MIGRATION_4_5 = new Migration(4, 5) {
+        @Override
+        protected void migrate(final Context context) {
+            boolean brightnessGestureSwitch = sp.getBoolean(
+                    context.getString(R.string.left_gesture_control_key), false);
+            boolean volumeGestureSwitch = sp.getBoolean(
+                    context.getString(R.string.right_gesture_control_key), false);
```
Stypox, line 117: Isn't the default `true`?

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/settings/SettingMigrations.java
Marius1501: Yes, you are right. 👍🏻

PR #9719, opened 2023-01-25
Added bottom main-tabs feature

### LINKED ISSUE #7658: Add option to show tabs in home page on bottom of screen
### Checklist

#### What feature do you want?

Make the Menu Panel on bottom.
Search bar easy acces on bottom corner.

#### Why do you want this feature?
 want a better ui

Marius1501:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Added the feature that the main-tabs can be set to the bottom
- Added a toggle in the apperance-settings for this feature

| Main-Page  | Appearance-Settings |
|NewPipe2 – MainFragment java  NewPipe2 app main  25 01 2023 19_29_17 |NewPipe2 – MainFragment java  NewPipe2 app main  25 01 2023 19_29_33|

- Fixes #7658 

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
@@ -1,7 +1,15 @@
 package org.schabi.newpipe.fragments;
 
+import static android.widget.RelativeLayout.ABOVE;
+import static android.widget.RelativeLayout.ALIGN_PARENT_BOTTOM;
+import static android.widget.RelativeLayout.ALIGN_PARENT_TOP;
+import static android.widget.RelativeLayout.BELOW;
+import static com.google.android.material.tabs.TabLayout.INDICATOR_GRAVITY_BOTTOM;
+import static com.google.android.material.tabs.TabLayout.INDICATOR_GRAVITY_TOP;
+
 import android.content.Context;
 import android.content.res.ColorStateList;
+import android.graphics.Color;
 import android.os.Bundle;
 import android.util.Log;
 import android.view.LayoutInflater;
@@ -12,10 +20,10 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.RelativeLayout;
 
+import androidx.annotation.ColorInt;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.ActionBar;
-import androidx.core.content.ContextCompat;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
 import androidx.fragment.app.FragmentStatePagerAdapterMenuWorkaround;
@@ -93,8 +101,6 @@ public class MainFragment extends BaseFragment implements TabLayout.OnTabSelecte
 
         binding.mainTabLayout.setupWithViewPager(binding.pager);
         binding.mainTabLayout.addOnTabSelectedListener(this);
-        binding.mainTabLayout.setTabRippleColor(binding.mainTabLayout.getTabRippleColor()
-                .withAlpha(32));
 
         setupTabs();
     }
@@ -112,7 +118,8 @@ public class MainFragment extends BaseFragment implements TabLayout.OnTabSelecte
         } else if (hasTabsChanged) {
             setupTabs();
         }
-        updateTabsPosition();
+
+        updateTabLayoutPosition();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -213,6 +213,8 @@
     <string name="delete_search_history_alert">Delete entire search history?</string>
     <string name="search_history_deleted">Search history deleted</string>
     <string name="fast_mode">Fast mode</string>
+    <string name="main_tabs_position_summary">Change the position of the main tabs</string>
+    <string name="main_tabs_position_title">Main Tabs Position</string>
```
Stypox, line 217:
```suggestion
    <string name="main_tabs_position_summary">Move main tab selector to the bottom</string>
    <string name="main_tabs_position_title">Main tabs position</string>
```

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -189,6 +194,38 @@ private void updateTabsIconAndDescription() {
...
+                .getBoolean(getString(R.string.main_tabs_position_key), false)) {
+            tabParams.removeRule(RelativeLayout.ALIGN_PARENT_TOP);
+            tabParams.addRule(RelativeLayout.ALIGN_PARENT_BOTTOM);
+            pagerParams.removeRule(RelativeLayout.BELOW);
+            pagerParams.addRule(RelativeLayout.ABOVE, R.id.main_tab_layout);
+            tabLayout.setSelectedTabIndicatorGravity(TabLayout.INDICATOR_GRAVITY_TOP);
+            final TypedValue typedValue = new TypedValue();
+            getContext().getTheme().resolveAttribute(R.attr.colorPrimary, typedValue,
+                    true);
+            getActivity().getWindow().setNavigationBarColor(typedValue.data);
```
Stypox, line 214: The navigation bar seems to remain red even after going to e.g. the search fragment (which doesn't have tabs). Anyway, this comment would become outdated if you just made the tab selector the same color of the app's background color.

Stypox (changes requested):
I don't think it looks so good when the mini player is open. Maybe it would be better if you made the tab layout colored like the app background, when it is at the bottom? When at the top, it should still be service-colored.


Other than that, looks simple enough and mostly good to me. Thank you!

tsiflimagas:
I think what makes it look odd isn't the colour (could be too), but the placement when the player is minimized. Bottom bar should always stay at the bottom, mini player shouldn't move it from there, but go on top of it. Then it would also look uniform with the system navigation bar.

Stypox:
Yeah, that's true. But moving the bottom player to a different position in the UI would be pretty difficult I think.

Stypox:
> Making bottom tabs a permanent change could be a solution.

No, you would still have two nested tab layouts which is not a good idea. Also, this is unrelated to this PR.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -189,6 +196,46 @@ private void updateTabsIconAndDescription() {
...
+            pagerParams.addRule(RelativeLayout.ABOVE, R.id.main_tab_layout);
+            tabLayout.setSelectedTabIndicatorGravity(TabLayout.INDICATOR_GRAVITY_TOP);
+            final TypedValue typedValueBackground = new TypedValue();
+            getContext().getTheme().resolveAttribute(R.attr.colorSecondary, typedValueBackground,
+                    true);
+            tabLayout.setBackgroundColor(typedValueBackground.data);
+            tabLayout.setTabRippleColor(ColorStateList.valueOf(
+                    getResources().getColor(R.color.gray)));
+            tabLayout.setTabIconTint(ColorStateList.valueOf(getResources().getColor(R.color.gray)));
+            tabLayout.setSelectedTabIndicatorColor(getResources().getColor(R.color.gray));
```
Stypox, line 220: I don't think it should be grey. `colorAccent` would fit better.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -189,6 +196,46 @@ private void updateTabsIconAndDescription() {
...
+            pagerParams.addRule(RelativeLayout.BELOW, R.id.main_tab_layout);
+            tabLayout.setSelectedTabIndicatorGravity(TabLayout.INDICATOR_GRAVITY_BOTTOM);
+            final TypedValue typedValue = new TypedValue();
+            getContext().getTheme().resolveAttribute(R.attr.colorPrimary, typedValue,
+                    true);
+            tabLayout.setBackgroundColor(typedValue.data);
+            tabLayout.setTabRippleColor(binding.mainTabLayout.getTabRippleColor().withAlpha(32));
+            tabLayout.setTabIconTint(binding.mainTabLayout.getTabIconTint());
+            tabLayout.setSelectedTabIndicatorColor(ContextCompat
+                    .getColor(requireContext(), R.color.white));
```
Stypox, line 234: Please note that `binding.mainTabLayout` is the exact same thing as `tabLayout`. Therefore the first two lines either do not do anything, or creates the following situation after switching the setting two times.
image

Here it would be ok to use white everywhere as that seems like what's being done in `fragment_main.xml`.

app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
```diff
@@ -189,6 +196,46 @@ private void updateTabsIconAndDescription() {
...
+        if (PreferenceManager.getDefaultSharedPreferences(requireContext())
+                .getBoolean(getString(R.string.main_tabs_position_key), false)) {
+            tabParams.removeRule(RelativeLayout.ALIGN_PARENT_TOP);
+            tabParams.addRule(RelativeLayout.ALIGN_PARENT_BOTTOM);
+            pagerParams.removeRule(RelativeLayout.BELOW);
+            pagerParams.addRule(RelativeLayout.ABOVE, R.id.main_tab_layout);
+            tabLayout.setSelectedTabIndicatorGravity(TabLayout.INDICATOR_GRAVITY_TOP);
+            final TypedValue typedValueBackground = new TypedValue();
+            getContext().getTheme().resolveAttribute(R.attr.colorSecondary, typedValueBackground,
+                    true);
```
Stypox, line 215: Use ThemeHelper.resolveColorFromAttr instead (also below).

Stypox (changes requested):
Could you briefly try to see if there is a way to put the tab layout below the mini player? If that's not possible don't worry.

Stypox:
Thanks! There are still some comments you didn't solve (also from the previous round of reviewing). Also, please mark comments you have solved as "Resolved", in order for you to keep track of them in an easier way. Also, did you read the review top content?

> Could you briefly try to see if there is a way to put the tab layout below the mini player? If that's not possible don't worry.

Marius1501:
@Stypox I tried to put the tab layout below the mini player but it hasn`t worked yet. Maybe you have a hint how to solve the problem?

MD77MD:
i would also wish if you can make main tab always visible when navigating newpipe except when watching a video... same like YouTube app... there will make navigating the app much easier

Stypox (approved):
I pushed two commits fixing up the comment that were not solved yet (i.e. the one about only updating when needed). I also improved the code in `updateTabLayoutPosition` to make it simpler to grasp. I tested on my device and two emulators (API 22 and API 33) and everything worked as expected. Don't worry about the player position, it's not the best but it would be too difficult to change. Do you think my changes are ok? Thanks again!

TheHCJ:
When will this be merged?

TobiGr (approved):

PR #9746, opened 2023-02-04
Add an option to Ignore hardware media button events

### LINKED ISSUE #9745: Add an option to disable prev/next track buttons on headsets
### Checklist

### Feature description

My plan is to add a new option under Video and audio/Behavior, which would look like this:
```
Ignore headset media button presses  
Notice: In the case of a Bluetooth headset, all media button events will be ignored, which breaks the Media Control UI in the notification area
```

When the option is enabled, it should do precisely what its description states.
Because it is impossible to distinguish between events that come from a Bluetooth headset and the control UI, it will ignore events from both sides. However, this should only occur when a Bluetooth headset is detected.
See also: https://code.videolan.org/videolan/vlc-android/-/merge_requests/1633

I sleep with my headset on. However, I often press the prev/next track buttons on the headset by accident, which causes the video to rewind or skip. Needless to say, this is quite inconvenient.  
Unfortunately, on Android, there is no way to disable media buttons globally without rooting the phone or physically blocking the pin (which also disables the mic). This means I need the option to make the player ignore those presses.
I'm planning to implement it myself today. But I would also like to know if someone else is interested in such a feature.

### Additional information

_No response_

NyanCatTW1:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing)

- Adds an option called "Ignore hardware media button events", which disables any pause/play and the prev/next track buttons, including ones on a headset. It helps prevent accidental touches and makes the app more usable on a headset with broken buttons.

The option is inside the Video and audio category.
- Before:
image

- After:
image

- Fixes #9745

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/mediasession/MediaSessionPlayerUi.java
+++ b/app/src/main/java/org/schabi/newpipe/player/mediasession/MediaSessionPlayerUi.java
@@ -44,6 +44,15 @@ public class MediaSessionPlayerUi extends PlayerUi {
         sessionConnector = new MediaSessionConnector(mediaSession);
         sessionConnector.setQueueNavigator(new PlayQueueNavigator(mediaSession, player));
         sessionConnector.setPlayer(getForwardingPlayer());
+        sessionConnector.setMediaButtonEventHandler((p, i) -> {
+            // It seems like events from the Media Control UI
+            // in the notification area don't go through this function,
+            // so it's safe to just ignore all events in case we want to
+            // ignore the hardware media buttons.
+            // Returning true stops all further event processing of the system
+            return player.getPrefs().getBoolean(
+                    context.getString(R.string.ignore_hardware_media_buttons_key), false);
+        });
 
         sessionConnector.setMetadataDeduplicationEnabled(true);
         sessionConnector.setMediaMetadataProvider(exoPlayer -> buildMediaMetadata());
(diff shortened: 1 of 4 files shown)


Stypox (changes requested):
Thanks! I don't think it's a good idea to add the `BLUETOOTH_CONNECT` permission. Unfortunately unless there is a different way to tell media button events apart, I don't think this will be merged.

NyanCatTW1:
Actually, it seems like only hardware media events go through the custom handler. Therefore, it should be safe to just not distinguish them.
I will make the changes in two or three days

Stypox:
Ok, thanks. So the setting description should be updated to explain that all hardware actions will be blocked (including bluetooth ones). Then it will be up to users to choose if they want to enable this setting or not.

NyanCatTW1:
I made a mess with the git history on the fork, but it should be ready for preview now  
It now ignores all events that go through the MediaButtonEventHandler, which should hopefully be exactly what's needed to achieve the desired behavior

Stypox:
I pushed some improvements to make code shorter and a bit more performant. I tested with a bluetooth speaker and the hardware buttons get correctly blocked when the setting is on. Are the changes I pushed ok, in your opinion? Thanks!

NyanCatTW1:
Looks nice! So I suppose it's ready for merge now?

Stypox (approved):
Yes, thanks for confirming :-)

PR #9748, opened 2023-02-04
Add whitelist to only retrieve supported YouTube itags/streams

TobiGr:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

This PR makes NewPipe compatible with the changes from https://github.com/TeamNewPipe/NewPipeExtractor/pull/488 and https://github.com/TeamNewPipe/NewPipeExtractor/pull/706

https://github.com/TeamNewPipe/NewPipeExtractor/pull/488 and https://github.com/TeamNewPipe/NewPipeExtractor/pull/706 can, but do not have to be merged

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/resolver/AudioPlaybackResolver.java
+++ b/app/src/main/java/org/schabi/newpipe/player/resolver/AudioPlaybackResolver.java
@@ -1,6 +1,6 @@
 package org.schabi.newpipe.player.resolver;
 
-import static org.schabi.newpipe.util.ListHelper.getNonTorrentStreams;
+import static org.schabi.newpipe.util.ListHelper.getPlayableStreams;
 
 import android.content.Context;
 import android.util.Log;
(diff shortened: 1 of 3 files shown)


Stypox:
Great!

This is service-dependent, isn't it? Though I doubt there is any way around this.

TobiGr:
oh yes, I could also add a condition to check the current service to make sure that there are no side effects. I did not test this PR with other services.

TobiGr:
Tested with the other services as well. I did not encounter any problems

Stypox:
Shouldn't the filtering happen on a client level? What if NewPipe isn't able to play a specific video, while Piped is, or viceversa?

TobiGr:
> Shouldn't the filtering happen on a client level? What if NewPipe isn't able to play a specific video, while Piped is, or viceversa?

This PR is opened in the client / app repo and not the extractor repo :)

Stypox (approved):
Sorry, I don't know why I thought we were on the extractor... Anyway, looks good to me. Note that we have `getUrlAndNonTorrentStreams` too, which is used for external players, and we might want to prevent/allow other kinds of streams from being opened in external streams, too.

PR #9726, opened 2023-01-29
Fix opening URLs in browser on API 30+

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

See https://medium.com/androiddevelopers/package-visibility-in-android-11-cc857f221cd9 and https://github.com/TeamNewPipe/NewPipe/issues/9615. I used what @plasticanu suggested, except that I also added "http" and "market" to the used schemas.

I tried to install the Kore app and everything works well even without adding `<package android:name="org.xbmc.kore" />` to the possible queries, so I didn't add it in the end. I tested on API 33.

#### Fixes the following issue(s)

- Fixes #9615

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

Issue #9712, opened 2023-01-22
Release v0.25.0 (please TEST!)


Stypox:
## Testing for regressions
Debug APK (built by @Stypox): [NewPipe_v0.25.0_RC2_debug.zip](https://github.com/TeamNewPipe/NewPipe/files/10529530/NewPipe_v0.25.0_RC2_debug.zip)
Release APK (built and signed by @Stypox): [NewPipe_v0.25.0_RC2_release.zip](https://github.com/TeamNewPipe/NewPipe/files/10529519/NewPipe_v0.25.0_RC2_release.zip)
Please report **only regressions** (i.e. new issues) here, not issues that were already present in previous releases!

## TODO

- [x] https://github.com/orgs/TeamNewPipe/projects/12
- [ ] Merge weblate
- [x] https://github.com/TeamNewPipe/NewPipe/pull/9553#issuecomment-1398217296, see #9742

## NewPipeExtractor version

This version of NewPipe will ship with [NewPipeExtractor version v0.22.5](https://github.com/TeamNewPipe/NewPipeExtractor/releases/tag/v0.22.5) (still WIP)

<details><summary><h2>App changelog </h2></summary><p>

### New
- Show toast when no updates are available #8668
- Show subscriber count on video details page #9173
- Download items from the queue #9502
- Allow permanently setting a playlist thumbnail #9523
- Add long press action on hashtags and web links in descriptions #7725
- Full width thumbnails aka card view mode #9310

### Improved
- Remove the redundant/overlapping toast "Copied to clipboard" for Android 13+ #8961
- Make mini-player close button area larger #8743
- Use smoother bitmap downscaling for thumbnails #8686 #9479
- Rename "Help" to "Fast mode" #9500
- Add hint to improve discoverability of subscription import #9501
- Disable night theme selection if auto theme is not used #9488
- Set `compileSdk` and `targetSdk` to 33 (Android 13) #9306 #9671 #9696
- Seeking no longer pauses the played stream #9496
- Added icon for "Play All" #9568
- Add language suffix to downloaded subtitle filenames #9499
- Check availability of Samsung DeX only on Samsung devices #9619
- Add "What's New" to the default main page tabs #9664
- Change "What's New" icon #9678
- Made channel images in grid lists bigger #9555
- Change "Chapters" icon #9691

### Fixed
- Fix persistent hover overlay when in desktop/DeX mode or using a mouse/non-touch input #8316
[...]

PR #9758, opened 2023-02-07
Fix opening links on Android 12+

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

#9726 wasn't enough. Android 12+ seems to return null to calls to `resolveActivity`, so the previous code would have said "No app can open this". Now instead, we try to open the activity in any case, even when we don't have a resolved package to set. I will take care of removing the now unused string if this PR is considered correct.
@AudricV do you think this is ok?

- See https://github.com/TeamNewPipe/NewPipe/issues/9712#issuecomment-1400843696

#### APK testing

#### Due diligence


TobiGr:
Is ``null` returned on any Android version or just A12+? If the latter is the case, I'd suggest to implement the old behaviour for Android <= 11

Stypox:
I thought about that, too. But what's the advantage of showing an error to the user instead of trying again in a different way?

Issue #9773, opened 2023-02-09
Interaction needs one tap before beeing able to interact


IronGibbet:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.25

### Steps to reproduce the bug

1. tap and hold
2. swipe right side up (expected behavior not happening) 
3. only nav bar and pulldown bar appear
4. tap and hold again you can right/left side swipe up/down

### Expected behavior

1. Tap and hold
2. swipe up/down to adjust brightness/volume
or
1.double tap (or more) to skip/rewind 

### Actual behavior

1. tap and hold screen
2. nav bar and pulldown bar pop up
3. no interaction possible
4. tap and hold to adjust brightness or volume
5. 50:50 chance nav and pulldown bar vanish after time

### Affected Android/Custom ROM version

Android 11

### Affected device model

Xiaomi veux

### Additional information

Will rollback to 0.24.1

Stypox:
Does #9812 solve the issue? You can find an APK in the Checks tab as soon as the build finishes

Stypox:
I created a small testing application that simulates NewPipe's fullscreen behavior. When tapping on the FAB at bottom right, does the system UI appear on the first tap or only on the second? If it does so only on the second tap, then your OSs probably wrongly implement `BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE` and treat touches as swipes. We should be able to fix it anyway by showing NewPipe's controls whenever system bars are swiped into view by the user. 

[app-debug.zip](https://github.com/TeamNewPipe/NewPipe/files/10730773/app-debug.zip)

IronGibbet:
[...]

PR #9785, opened 2023-02-11
Avoid jumping of the player on devices with a cutout.

Isira-Seneviratne:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Restore the code to prevent the player jumping on devices with a cutout.

- Fixes #

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -27,6 +27,7 @@ import android.graphics.Color;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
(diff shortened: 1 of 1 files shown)


Stypox:
@SameenAhnaf could you check if this also fixes your "requires two taps" problem?

Stypox:
@Isira-Seneviratne then I think it's best to just revert #8894. It required to fix many bugs before merging, and even after merging new bugs keep arising. I don't think it's worth it to use `WindowCompat` if it just causes bugs.

Stypox:
Actually, I found out all flags were deprecated, so let's try a bit more before reverting. I pushed a commit that properly sets window flags only when needed, and also does not use `BEHAVIOR_SHOW_BARS_BY_TOUCH` which is deprecated on Android 12+.

Isira-Seneviratne:
@Stypox I simplified one of the methods you added.

Stypox:
Since this still does not solve #9773, I will merge #9812 instead, sorry. Let's not forget about the code in this PR as it might come in handy

PR #9801, opened 2023-02-12
Optimize svg assets

RiedleroD:
I wanted to finish optimizing all of them before opening the PR but I think that's not happening anymore. As far as I can see, none of those SVGs are actually directly used inside the app. Otherwise, this would probably be a significant performance improvement. So this PR really only saves about 60KiB when downloading the repo and not much else. Maybe next time someone wants to change some of those icons it'll be easier because they're less messy. idk.

I'll probably optimize some of the worse xml files that actually get used some other time and open another PR.

#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- optimized svg files in various ways

#### APK testing

#### Due diligence


Profpatsch:
The github preview shows that all optimized svgs now have a size of 0x0px, e.g.:
 
image

I feel like this might lead to problems if some code implicitly depends on these default height/widths?

RiedleroD:
huh, will look into it in a month or so. Currently absolutely swamped with work from my thesis, diploma project & finals  😓

Profpatsch:
@RiedleroD I’d close this PR for now, given that we only save about 60k and haven’t really checked whether everything still works as expected … I’d rather focus on the UI rewrite for now.

But feel free to open a separate one (and link this discussion) if you ever get around to it again!

RiedleroD:
yeah fair point. I don't think I'll be revisiting this anytime soon though, there's just too much to do 😓

Profpatsch:
No worries! Thanks for giving it a shot!

Issue #9403, opened 2022-11-15
Open in browser can't use any app other than the current default browser


SebiderSushi:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.24.1

### Steps to reproduce the bug

1. Install NewPipe
2. Install another app that can open youtube links
   For example the official YouTube app or [ViMusic](https://f-droid.org/en/packages/it.vfsfitvnm.vimusic)
3. Navigate to an arbitrary youtube video in NewPipe and tap "Open in browser"
4. Go to System Settings > Apps and notifications > Default apps > Opening links
5. Set all apps to "Don't open supported links"
6. Set only the official YouTube app OR ViMusic to "Open youtu.be and other URLs" or "Ask every time"
7. Navigate to an arbitrary youtube video in NewPipe and tap "Open in browser"
8. Disable the preinstalled "Browser" app
9. Navigate to an arbitrary youtube video in NewPipe and tap "Open in browser"

### Expected behavior

After clicking "Open in browser" the default android activity picker shows up (see screenshots) with all apps set up to handle the URL or android launches the default activity if previously selected.

If only one app is set to "Open youtu.be and other URLs" or "Ask every time" it should open the URL.

### Actual behavior

NewPipe [currently enforces the default package](https://github.com/TeamNewPipe/NewPipe/blob/b7a44560f5c28b8e252cbf49185da40f13c61e72/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java#L63) or refuses to operate if there is none set.

If only one app is set to "Open youtu.be and other URLs" or "Ask every time" the URL is still opened in the default browser.

[...]

PR #9833, opened 2023-02-18
Fix opening downloads and streams on external players on Android 11+ and refactor external intents

### LINKED ISSUE #9771: [Android 11+] Cannot open downloads from the app ("No app installed to play this file" toast)
### Checklist

### Affected version

v0.25.0

1. Download any audio from a youtube video in either .m4a or opus
2. Within the Downloads page tap on the downloaded audio file
3. A notification appears saying 'No app installed to play this file'

### Expected behavior

File should play in assigned player or a notification should appear for a suitable audio player to be chosen

### Actual behavior

A notification appears saying 'No app installed to play this file'

Screenshot_2023-02-09-17-24-01-97_ee8807d2de22f1831269d80920e33a3b

### Logs

_No response_

Android 12

### Affected device model

Oneplus 10 Plus 5G

### Additional information

In v0.24.1 everything works as expected, it was only when I upgraded to v0.25 that the problem manifests.

I tried installing and uninstalled v0.24.1 and v0.25 a numvber of time to test the issue, and I was able to reproduce it every time.

AudricV:
#### What is it?
- [x] Bugfix (user facing)

This PR fixes opening streams on external players and downloads on Android 11 and higher and refactors how external intents are handled.

It does more precisely the following changes:

- the check of the availability of a default package has been removed and replaced by `try` `catch` blocks. As a result, the HTTP default browser test for web links has been removed;
- an app chooser on Android 10+ is always shown for chooser intents, even if there is only one action to handle the original intent, so the user can use actions in the chooser, such as copying a text to clipboard in the share sheet for `ACTION_SEND` intents;
- the method to open an app chooser in `ShareUtils`, `openAppChooser`, has been made `public` and is now used when sharing a download in `MissionAdapter`, in order to avoid code duplication;
- some other changes about the parameters of the methods of `ShareUtils` have been also made: see the commits messages and the code changes for more details;
- as we only use now standard intents, we do not need to query packages anymore: see https://web.archive.org/web/20200817164113/https://developer.android.com/preview/privacy/package-visibility#use-cases-not-affected. That's the reason why the `queries` element in the Android manifest has been removed.
- the `toast_no_player` string resource has been removed, as it was only used in `MissionAdapter` when trying to a play a download, and has been replaced by the generic one `No app on your device can open this` with the usage of the method `openIntentInApp` in  `ShareUtils` instead of starting manually the intent.
- opening links on the about and licenses pages, RSS feed in the account/channel page where applicable, the issues GitHub page and the privacy policy links in the error activity and web links in comments, descriptions, metadata and meta-info no longer always use an app chooser, this task is now handled by the Android system. Let me know if I should revert completely or partially these changes.

- Fixes #9771.

#### APK testing

Please test that the changes work fine with this build, by using the following locations where external chooser and view intents are used:

- stream dialog entry, channel/account, playlist, content details, in-app player: open in browser and share;
- comment: open links;
- content details: open description and metadata;
- channel/account: open RSS feed, where applicable;
- external player dialog: open a stream URL in an external app and open a content in browser;
- play queue: share a stream;
- license dialog: open license's website;
- about page: open links;
- download context menu: open and share files;
- app install dialog: install VLC and Kodi;
- error report actions: create an email to send an error report, open report on GitHub and privacy policy links and share error report;
- editable or selectable texts: share selected text.

I personally tested the changes on an Android 14 Developer Preview 1 emulator and my Honor 9X device, running EMUI 10.0 with Android 10, and everything about external intents I have been able to test was working as excepted.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
@@ -7,8 +7,6 @@ import android.content.ClipData;
 import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.Intent;
-import android.content.pm.PackageManager;
-import android.content.pm.ResolveInfo;
 import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.Build;
@@ -35,24 +33,28 @@ public final class ShareUtils {
     }
 
     /**
-     * Open an Intent to install an app.
+     * Open an {@link Intent} to install an app.
+     *
      * <p>
      * This method tries to open the default app market with the package id passed as the
      * second param (a system chooser will be opened if there are multiple markets and no default)
      * and falls back to Google Play Store web URL if no app to handle the market scheme was found.
+     * </p>
+     *
      * <p>
-     * It uses {@link #openIntentInApp(Context, Intent, boolean)} to open market scheme
-     * and {@link #openUrlInBrowser(Context, String, boolean)} to open Google Play Store
-     * web URL with false for the boolean param.
+     * It uses {@link #openIntentInApp(Context, Intent, boolean)} to open the market scheme
+     * and {@link #openUrlInBrowser(Context, String, boolean)} to open Google Play Store's
+     * web URL with false for the boolean parameter.
+     * </p>
      *
      * @param context   the context to use
      * @param packageId the package id of the app to be installed
      */
     public static void installApp(@NonNull final Context context, final String packageId) {
         // Try market scheme
-        final boolean marketSchemeResult = openIntentInApp(context, new Intent(Intent.ACTION_VIEW,
-                Uri.parse("market://details?id=" + packageId))
-                .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK), false);
+        final boolean marketSchemeResult = openIntentInApp(context,
+                new Intent(Intent.ACTION_VIEW, Uri.parse("market://details?id=" + packageId))
+                        .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK), false);
         if (!marketSchemeResult) {
             // Fall back to Google Play Store Web URL (F-Droid can handle it)
             openUrlInBrowser(context,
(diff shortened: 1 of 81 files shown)


app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
```diff
@@ -35,24 +33,28 @@ private ShareUtils() {
...
      * and falls back to Google Play Store web URL if no app to handle the market scheme was found.
+     * </p>
+     *
      * <p>
-     * It uses {@link #openIntentInApp(Context, Intent, boolean)} to open market scheme
-     * and {@link #openUrlInBrowser(Context, String, boolean)} to open Google Play Store
-     * web URL with false for the boolean param.
+     * It uses {@link #openIntentInApp(Context, Intent, boolean)} to open the market scheme
+     * and {@link #openUrlInBrowser(Context, String, boolean)} to open Google Play Store's
+     * web URL with false for the boolean parameter.
```
Stypox, line 47:
```suggestion
     * web URL without an app chooser.
```

app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
```diff
@@ -203,36 +127,82 @@ private static void openAppChooser(@NonNull final Context context,
...
-        final ResolveInfo resolveInfo = context.getPackageManager().resolveActivity(intent,
-                PackageManager.MATCH_DEFAULT_ONLY);
-
-        if (resolveInfo == null) {
-            return "";
-        } else {
-            return resolveInfo.activityInfo.packageName;
+    public static boolean openIntentInApp(@NonNull final Context context,
+                                          @NonNull final Intent intent,
+                                          final boolean showNoAppToOpenToast) {
```
Stypox, line 189: What about creating two methods and removing this `boolean` parameter? I would call the two methods:
- `tryOpenIntentInApp` which just does `try{...}catch{return false}; return true`
- `openIntentInApp` which does `if (!tryOpenIntentInApp()) showToast` and has return type `void`

app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
```diff
@@ -203,36 +127,82 @@ private static void openAppChooser(@NonNull final Context context,
...
+     * @param url        the URL to browse
+     * @param useChooser whether to use a system chooser instead of opening the link in the default
+     *                   app or browser (the default app associated with this link or the default
+     *                   browser if there is no default app will be only shown in the system
+     *                   chooser on Android 12 and higher)
+     * @return whether the URL can be opened
+     */
+    public static boolean openUrlInBrowser(@NonNull final Context context,
+                                           final String url,
+                                           final boolean useChooser) {
```
Stypox, line 156: Also here. I think it would be better to create two different methods, so that it is clear what's happening just by reading the code (no need to look up parameter documentation).
- `openUrlInDefaultBrowser`
- `openUrlInBrowserWithChooser`

Stypox (changes requested):
Code looks good to me, I just have some suggestions on how to improve the `ShareUtils` methods structure. Thanks!

The behavior requested in https://github.com/TeamNewPipe/NewPipe/issues/9403 seems now to be NewPipe's behavior. So when opening in browser, the official youtube app would show up instead. Only if I disable the official app's "Open by default" system setting, then the browser shows up. And if I then set NewPipe to open supported links, opening in browser results in the NewPipe share dialog coming up, which is surely not wanted. Note that this happens both if I set `useChooser` to `false` and `true. Is this the way Android is supposed to work?

AudricV:
> And if I then set NewPipe to open supported links, opening in browser results in the NewPipe share dialog coming up, which is surely not wanted. Note that this happens both if I set useChooser to false and true, with the only difference that in the first case no chooser shows up. Is this the way Android is supposed to work?

Yes, it is, starting with Android 12.

Another solution would be to reintroduce query packages permission for `http` and `https` link, which are removed with this PR, query all packages without Android filtering for web links (excluding disabled apps) (like done [here](https://github.com/kuesji/link_eye/blob/8bc3f84ff8247be8b5a391779b2baf5dab4bd352/app/src/main/java/kuesji/link_eye/LinkHandler.java#L117) but without the [`MATCH_DISABLED_COMPONENTS` flag](https://developer.android.com/reference/android/content/pm/PackageManager#MATCH_DISABLED_COMPONENTS) instead of the [`MATCH_DISABLED_UNTIL_USED_COMPONENTS`](https://developer.android.com/reference/android/content/pm/PackageManager#MATCH_DISABLED_UNTIL_USED_COMPONENTS)), and then show a dialog with all apps which can open this link and open the corresponding app with the intent when it's tapped.

app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
```diff
@@ -350,32 +355,20 @@ private void viewWithFileProvider(Mission mission) {
...
         }
-        intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+
+        final Intent intent = new Intent(Intent.ACTION_SEND);
+        intent.setType(resolveMimeType(mission));
+        intent.putExtra(Intent.EXTRA_STREAM, resolveShareableUri(mission));
         intent.addFlags(FLAG_GRANT_READ_URI_PERMISSION);
 
-        mContext.startActivity(intent);
+        openAppChooser(mContext, intent, mContext.getString(R.string.share_dialog_title));
```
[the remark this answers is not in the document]
AudricV: No, this is the title that will be shown when something is shared, so using "Open with" for share intent is not a standard behavior, but "Share with" is (and that's what Android 10+ versions use).

app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
AudricV: Not a bad idea, but I would use `openUrlInDefaultApp` and `openUrlWithAppChooser`, as we do not look at a browser anymore.

app/src/main/java/us/shandian/giga/ui/adapter/MissionAdapter.java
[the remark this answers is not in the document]
Stypox: Oh ok, I see, thanks for clarifying

app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
Stypox: Right

Stypox:
> Yes, it is, starting with Android 12.

Mmmh, then I guess this is the right thing to do, let's hope it does not create issues for users. I don't think we should implement non-standard things: users have the ability to copy the url e.g. from the share sheet (including older android versions if there is a "Copy to clipboard" app installed).

Then I think the "Open in browser" button should become "Open in..." with [this icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aopen_in_new%3A) (or similar)

AudricV:
Closing in favor of #9850, which aims to keep the current behavior as much as possible and embed the requested changes.

PR #9878, opened 2023-03-03
Use custom relative time formatting logic instead of PrettyTime.

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Remove the PrettyTime dependency.
- Rely on Android's built-in functionality ([`DateUtils.getRelativeTimeSpanString`](https://developer.android.com/reference/android/text/format/DateUtils.html#getRelativeTimeSpanString(long)), [`RelativeDateTimeFormatter`](https://developer.android.com/reference/android/icu/text/RelativeDateTimeFormatter)) to format relative times.
- Add relative time localization strings for years, months and weeks, for use on APIs < 24.

- Before:
Screenshot_20230303-120519_NewPipe

- After:
Screenshot_20230303-120528_NewPipe Relative_time_format

#### APK testing

#### Due diligence


app/src/main/res/values/strings.xml
TobiGr, line 1: I think the translations were one of the main reasons for using PrettyTime, weren't they? I do not see the benefit of removing PrettyTime until the minSdk is at 24. The current minSdk is at 21, which equals Android 5. Having a minSdk of 24 is equal to Android 7. The minSdk won't be increased by two Android versions in the near future. 
Relying on translations made (or especially not made) by users will result in a worse UX.
What is the benefit of removing PrettyTime? The APK size might be decreased a bit, however the translations are missing atm.

Isira-Seneviratne: > What is the benefit of removing PrettyTime? The APK size might be decreased a bit, however the translations are missing atm.

True, but there are only three sets of quantity strings that need to be translated, as [`DateUtils.getRelativeTimeSpanString`](https://developer.android.com/reference/android/text/format/DateUtils#getRelativeTimeSpanString(long)) handles the formatting if the difference is less than 1 week. Also, using quantity strings allows for proper quantity localization in some locales: https://googlesamples.github.io/android-custom-lint-rules/checks/ImpliedQuantity.md.html

Stypox:
What are the advantages and disadvantages of this PR? I am talking about supported languages, available format methods, APK size, ...

Isira-Seneviratne:
> What are the advantages and disadvantages of this PR? I am talking about supported languages, available format methods, APK size, ...

Android includes some of its own relative time formatting functionality in methods like `[DateUtils.getRelativeTimeSpanString](https://developer.android.com/reference/android/text/format/DateUtils.html#getRelativeTimeSpanString(long))` (which allows for 1 day's difference to be formatted as "Yesterday" instead of "1 day ago", for instance.) Due to that, the PrettyTime dependency can be removed, but it does require localization strings to be defined for API levels < 24 (`RelativeDateTimeFormatter` is being used for API 24+).

I compared the APK sizes, and the APK with this changeset is 37 KB smaller.

Isira-Seneviratne:
It's not an issue for me.

PR #9887, opened 2023-03-05
Use JSON serialization

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Replaces Java serialization with JSON serialization

This was changed inside
* Importing/Exporting of settings
* Unfinished Downloads (Downloadmission)
* [StateSaver](https://github.com/litetex/NewPipe/blob/json-serialization/app/src/main/java/org/schabi/newpipe/util/StateSaver.java)

Additionally a bug was fixed where the export of the database was not possible because it was not initialized.

⚠️ Note that after merging this PR importing of existing settings won't work, due to the new format!
It's highly recommended to do a new backup afterwards (when this is released in a new version).
If a old backup - that was created before these changes took effect - should be imported, it has to be done in an old version of the app and then the app needs to be updated.

#### APK testing

#### Due diligence


lm41:
If I may give some thoughts: It would be much more user friendly if there is still (maybe in a transition period) the possibility to import the old backups. If you want to load an old newpipe backup in a newer version and have to uninstall the app first and install an old version, this is in my opinion not the best solution, if not no solution at all.

app/src/main/java/org/schabi/newpipe/util/JSONSerializer.java
```diff
@@ -0,0 +1,37 @@
...
+ * Centralized class for JSON serialization.
+ */
+public final class JSONSerializer {
+    private JSONSerializer() {
+
+    }
+
+    private static ObjectMapper newMapper() {
+        return new ObjectMapper();
+    }
```
XiangRongLin, line 20: According to Phind an `ObjectMapper`should be reused, because it's creation is relativly expensive.

SO source: https://stackoverflow.com/a/57671444

It kind of makes sense to me because in the spring boot world you have one instance of it and use it for all http requests unless specific config is required

app/src/main/java/org/schabi/newpipe/util/JSONSerializer.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public static void toJson(final Object objToConvert, final OutputStream target) {
+        try {
+            newMapper().writeValue(target, objToConvert);
+        } catch (final IOException ioe) {
+            throw new UncheckedIOException(ioe);
+        }
```
XiangRongLin, line 27: Any reason to wrap it in `UncheckedIOException`?

It seems that exception is used in the stream API and ideally one should also throw it there.
https://stackoverflow.com/a/23402640

Profpatsch:
> ⚠️ Note that after merging this PR importing of existing settings won't work, due to the new format!
> It's highly recommended to do a new backup afterwards (when this is released in a new version).

As a user, I’d expect settings import to continue working even for old formats. So even if the format is changed, I do not believe the deserializer of the old format should go away. Only the serializer.

Stypox:
Will be taken care of separately

Issue #9870, opened 2023-03-01
Release v0.25.1 (please TEST!)


Stypox:
## Testing for regressions

Debug APK (built by our CI in #9869): ...
Release APK (built and signed by @Stypox): [NewPipe_v0.25.1_RC1_release.zip](https://github.com/TeamNewPipe/NewPipe/files/10859622/NewPipe_v0.25.1_RC1_release.zip)
Please report **only regressions** (i.e. new issues) here, not issues that were already present in previous releases!

## TODO

- [x] https://github.com/orgs/TeamNewPipe/projects/13
- [x] Update weblate
- [x] Relase NewPipeExtractor and update it in `app/build.gradle`

## NewPipe Extractor version

This version of NewPipe will ship with [NewPipe Extractor version v0.22.6](https://github.com/TeamNewPipe/NewPipeExtractor/releases/tag/v0.22.6)

<details><summary><h2>App changelog </h2></summary><p>

### New
- Handle duplicate streams in the "Add to playlist" dialog #9538 
- Larger channel cards in search results #9728 
- Add an option to ignore hardware media button events #9746 
- Add option to remove partially watched videos from the 'Whats new' feed #9747 
- Add button to remove duplicates from playlist #9707 

### Improved
- Reduce the size of thumbnails on big screens to fit more grid columns #9858 
- Make UI behavior for playback information display more consistent #8180 

### Fixed
- Orientation is locked if there is no sensor for it #9693 
- Fix crash when clicking on an empty comment #9777 
- [Android 11+] Correctly open URLs in browser and fix opening downloads and external players #9850
- Fix playlist streams progress bar size in card mode #9769
- Fix erroneous updates of playlist thumbnails #9755 

### Development
- Update to ExoPlayer 2.18.5 #9847 #9890
</details>

Yowlen:
Went to test and the first thing I noticed was when importing a database so that I could have my settings from the main app, it refused to load thumbnails and other static images until I restarted the app manually. Video lists for channels and video comments loaded fine. Only thumbnails and other images appear affected. Testing in 0.24.1 (I'm skipping 0.25.0 due to the external player bug) confirmed it's not present there and everything loaded after import just fine.

Stypox:
Can you reproduce that consistently? I cannot reproduce on my phone @Yowlen

Yowlen:
[...]

PR #9968, opened 2023-03-25
Use a system chooser when opening links in browser in the case there is no browser available

AudricV:
#### What is it?
- [x] Bugfix (user facing)

This PR makes the app using the behavior when there is no default browser on Android 11 and lower, by opening a system chooser when there is no browser available (on all Android versions).

Also catch any exception when the system chooser cannot be opened and show the "No app on your device can open this" toast in this case, as an `ActivityNotFoundException` would be thrown if no app is available to open a web link.

These changes have been tested on an Android 14 developer preview 2 emulator and a system chooser is now displayed where there is no browser available, no default app set to open the requested link and an available app.
They also work fine on my physical device running Android 10 when there is no app available to open a web link (the "No app on your device can open this" toast is still shown with this change).

- Fixes regression mentioned in https://github.com/TeamNewPipe/NewPipe/issues/9870#issuecomment-1460462646

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
@@ -86,15 +86,19 @@ public final class ShareUtils {
                     PackageManager.MATCH_DEFAULT_ONLY);
         }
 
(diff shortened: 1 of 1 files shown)


TobiGr (approved):
code lgtm

Stypox (approved):

PR #10015, opened 2023-04-10
Remove invidious.namazso.eu

namazso:


Stypox:
Why? You ignored the PR template and now I have no idea why this change is needed

namazso:
Because it's no longer functional and will not be in the future either. I assumed it's pretty obvious.

namazso:
Oh, my bad then. I just went through the the top starred repos that came up with a code search.

My redirect are still getting a lot of hits, and just trying to eliminate usages where people might not be informed of the changes:

image

PR #10035, opened 2023-04-20
Upgrade Gradle and AGP from 7.x to 8.x and use Java 17

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

Android Studio Flamingo just came out and it asked me to upgrade to Gradle 8.0, so I just ran its upgrade assistant. The version of Gradle it brought me to was 8.0, so I manually set it to 8.1 and it works fine. I also had to manually set the correct `distributionSha256Sum` taken from [here](https://gradle.org/release-checksums/).

The release notes for Gradle 8.0 are [here](https://docs.gradle.org/8.0/release-notes.html), for 8.1 [here](https://docs.gradle.org/8.1/release-notes.html), and release notes for all versions can be found [here](https://gradle.org/releases/). I didn't find anything relevant for us in the release notes.

As per Android Gradle Plugin, [here](https://developer.android.com/build/releases/gradle-plugin#8-0-0) are the release notes. The "namespace" breaking change does not affect us as we are already setting `namespace "org.schabi.newpipe"` under `android {`. Other breaking changes have to do with default values, and an interesting one might be the `R` class now having non-final resource IDs by default. Since the app builds fine (I tried to build after a `gradle clean`), I guess all of these changes have not broken our build.

I pushed a third commit adding a JVM argument to `gradle.properties`, otherwise the build will just crash with this strange error message. Previously I avoided this issue by using Gradle with Java 11, but Java 11 is not supported by Gradle 8.1 anymore, and higher Java versions give this error (I tried 17 and 19). I have no idea where this came from and by searching on the internet I only found bug reports about tools we don't use, although the common temporary solution for these kind of messages just seems to be adding `--add-opens jdk.compiler/PACKAGE_NAME=ALL-UNNAMED` (where `PACKAGE_NAME` in this case is `com.sun.tools.javac.model`).
```
class clojure.lang.Reflector cannot access class com.sun.tools.javac.model.JavacElements (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.model to unnamed module
```

Now `compileOptions` and `kotlinOptions` use JDK 17 (the version shipped with Android Studio by default) instead of JDK 11 to adapt to JDK 11 not being valid anymore. The CI failed for the same reason, so I also updated it to use JDK 17. While I was at it I also upgraded the API 29 Android emulator CI tests to API 33 (Android 13, the latest version), which required changing the sdkmanager target from `default` to `google_apis` since there is no API 33 emulator without Google APIs.

#### APK testing

@Isira-Seneviratne can you confirm this is ok?

#### Due diligence


Isira-Seneviratne (approved):
I did a debug build, and I didn't encounter any issues with the build. This looks good to me.

PR #10165, opened 2023-06-15
Fix downloads of streams with missing MediaFormat

### LINKED ISSUE #10153: SoundCloud Download "Unknown Quality" audio cannot be played
### Checklist

### Affected version

0.25.1

1. Go to NewPipe
2. Go to the SoundCloud option
3. Find a song that allows you to download in bitrate higher than 125kbps.  Normally it says "Unknown Quality" with a filesize significantly bigger than the default mp3 128kbps.
4. Download it
5. Try to play it with AIMP
6. Go to terminal of your phone and do `file $DOWNLOADED_FILE` it says file type is just "data".

### Expected behavior

I should be able to play this file.

### Actual behavior

Music players cannot play the downloaded file.
Linux does not detect the extension.

screenshot

### Logs

No crash logs.

Android 13, OxygenOS

### Affected device model

OnePlus 8 Pro

### Additional information

_No response_

TobiGr:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Rertrieve `MediaFormat`s for streams in case the extractor was not able to get them. Sometimes the MediaFormat can be retrieved from the HTTP response header that is fetched when getting the stream size. 

- The dialog now also displays "Unknown Quality" if the quality is unknown. Previously, the media format name was displayed as title and as quality info.

- Some code improvements simplifying castings by directly casting in if-conditions using `instanceof`

<table>
    <thead>
        <tr>
            <th>Before</th>
            <th>After</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td> 

 </td>
            <td>  </td>
        </tr>
    </tbody>
</table>

- Fixes #10153 

TeamNewPipe/NewPipeExtractor#1074

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/StreamItemAdapter.java
+++ b/app/src/main/java/org/schabi/newpipe/util/StreamItemAdapter.java
@@ -41,7 +41,7 @@ import us.shandian.giga.util.Utility;
  */
 public class StreamItemAdapter<T extends Stream, U extends Stream> extends BaseAdapter {
     @NonNull
-    private final StreamSizeWrapper<T> streamsWrapper;
+    private final StreamInfoWrapper<T> streamsWrapper;
     @NonNull
     private final SparseArrayCompat<SecondaryStreamHelper<U>> secondaryStreams;
 
@@ -53,7 +53,7 @@ public class StreamItemAdapter<T extends Stream, U extends Stream> extends BaseA
     private final boolean hasAnyVideoOnlyStreamWithNoSecondaryStream;
 
     public StreamItemAdapter(
-            @NonNull final StreamSizeWrapper<T> streamsWrapper,
+            @NonNull final StreamInfoWrapper<T> streamsWrapper,
             @NonNull final SparseArrayCompat<SecondaryStreamHelper<U>> secondaryStreams
     ) {
         this.streamsWrapper = streamsWrapper;
@@ -63,7 +63,7 @@ public class StreamItemAdapter<T extends Stream, U extends Stream> extends BaseA
                 checkHasAnyVideoOnlyStreamWithNoSecondaryStream();
     }
 
-    public StreamItemAdapter(final StreamSizeWrapper<T> streamsWrapper) {
+    public StreamItemAdapter(final StreamInfoWrapper<T> streamsWrapper) {
         this(streamsWrapper, new SparseArrayCompat<>(0));
     }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/schabi/newpipe/util/StreamItemAdapter.java
```diff
@@ -271,13 +289,147 @@ public static <X extends Stream> Single<Boolean> fetchSizeForWrapper(
...
+                            final Charset charset = Charset.forName(encoding);
+                            // decode info with the given encoding
+                            if (charset == null) {
+                                continue;
+                            }
+                            if (charset.equals(Charset.defaultCharset())) {
+                                fileName = part;
+                            } else {
+                                fileName = new String(part.getBytes(charset),
+                                        StandardCharsets.UTF_8);
```
Stypox, line 374: I am not sure whether this code does the right thing. If the data is already in a String then it has already been decoded in the default format, so the only way to re-decode it in a different charset would be to reverse the initial decoding (with `encode(default charset)`) and then decode it with the different charset.

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/util/StreamItemAdapter.java
TobiGr: Hm, you are right. In that case, we cannot do much about it. And the encoding should not matter because we only want to get the file extension which should always be in from the Latin alphabet. Shall I remove the encoding part?

Stypox: Yes in my opinion

TobiGr: Done

TobiGr:
- Rebased
- Removed encoding stuff as discussed in the review above
- Added a few more comments & docs

app/src/main/java/org/schabi/newpipe/util/StreamItemAdapter.java
```diff
@@ -271,13 +288,145 @@ public static <X extends Stream> Single<Boolean> fetchSizeForWrapper(
...
+
+        @VisibleForTesting
+        public static <X extends Stream> boolean retrieveMediaFormatFromContentTypeHeader(
+                @NonNull final X stream,
+                @NonNull final StreamInfoWrapper<X> streamsWrapper,
+                @NonNull final Response response) {
+            // try to get the format by content type
+            // some mime types are not unique for every format, those are omitted
+            final List<MediaFormat> formats = MediaFormat.getAllFromMimeType(
+                    response.getHeader("Content-Type"));
```
Stypox, line 404: `response.getHeader("Content-Type")` will return null if there are no such headers, but fortunately that's not a problem since `getAllFromMimeType` works normally with null strings.

PR #10175, opened 2023-06-18
refactor isTv

thonsi:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [*] Codebase improvement (dev facing)

- Prepare way to detect the TV 

#### APK testing

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
@@ -1,6 +1,7 @@
 package org.schabi.newpipe.util;
 
 import static android.content.Context.INPUT_SERVICE;
+import static android.content.Context.UI_MODE_SERVICE;
 
 import android.annotation.SuppressLint;
 import android.app.UiModeManager;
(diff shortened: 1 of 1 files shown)


Redirion:
so the functional difference is a null check. Is there any device that happens to not have UiModeManager? Seems to be a core service always launched by SystemServer. I don't see any advantages of this refactoring. Sorry

app/src/main/java/org/schabi/newpipe/util/DeviceUtils.java
```diff
@@ -57,8 +58,11 @@ public static boolean isTv(final Context context) {
         final PackageManager pm = App.getApp().getPackageManager();
 
         // from doc: https://developer.android.com/training/tv/start/hardware.html#runtime-check
-        boolean isTv = ContextCompat.getSystemService(context, UiModeManager.class)
-                .getCurrentModeType() == Configuration.UI_MODE_TYPE_TELEVISION
+        @Nullable
+        UiModeManager uiModeManager =
+                (UiModeManager) context.getApplicationContext().getSystemService(UI_MODE_SERVICE);
```
snaik20, line 63: Any reason to prefer `UI_MODE_SERVICE`. Please provide a reason or a doc reference to cross verify this. Also, provide a test case that returns `null` for existing function as pointed in another comment.

PR #10213, opened 2023-07-06
Update screenshots

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

I used #9182 to take screenshots, so that they don't become outdated when we release 0.26.0. The screenshots resemble the previous screenshots, except that they are all the same size (16:9), they try to show off as much public domain material as possible, and there are no screenshots for Downloads.

All images have been stripped of metadata and optimized with both `mogrify -strip -taint -compress Lossless` and `optipng`.

You can view a preview [here](https://github.com/Stypox/NewPipe/tree/update-screenshots).

#### APK testing

#### Due diligence


TobiGr (commented):
Looks good. Can you add one screenshot using the light theme on a phone, too? Maybe take one screenshot twice so the difference is clearly visible. You could also make sure to display the two tablet screenshots in the same line by adding an extra lilnebreak after the last phone screenshot.

What about the translated READMEs? Do we need to modify their image URLs, too?

Stypox:
@TobiGr I have changed the "phone search" screenshot to be white-themed, and the "tablet search" one to be dark-themed. I added two `<br/><br/>` between the tablet and phone images. I also updated all translated READMEs.

TobiGr (approved):

PR #8737, opened 2022-08-05
Replace coerceIn() with MathUtils.clamp().

Isira-Seneviratne:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Use [`MathUtils.clamp()`](https://developer.android.com/reference/androidx/core/math/MathUtils#clamp(float,float,float)) instead of Kotlin's `coerceIn()` in `PopupPlayerGestureListener`. This avoids an issue found by @Stypox  in #8651, as `coerceIn()` throws an exception if the maximum is less than the minimum while `MathUtils.clamp()` doesn't.

#### Fixes the following issue(s)

- Fixes #

#### APK testing 

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #10224, opened 2023-07-12
Replace MathUtils.clamp with Kotlin coerceIn

TacoTheDank:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Replace 'MathUtils.clamp' with Kotlin 'coerceIn' where possible
- Also utilize the isGone extension function

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/gesture/PopupPlayerGestureListener.kt
+++ b/app/src/main/java/org/schabi/newpipe/player/gesture/PopupPlayerGestureListener.kt
@@ -4,7 +4,7 @@ import android.util.Log
 import android.view.MotionEvent
 import android.view.View
 import android.view.ViewConfiguration
-import androidx.core.math.MathUtils
+import androidx.core.view.isVisible
 import org.schabi.newpipe.MainActivity
 import org.schabi.newpipe.ktx.AnimationType
 import org.schabi.newpipe.ktx.animate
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/player/gesture/PopupPlayerGestureListener.kt
```diff
@@ -251,8 +249,7 @@ class PopupPlayerGestureListener(
         // -- Determine if the ClosingOverlayView (red X) has to be shown or hidden --
         val showClosingOverlayView: Boolean = playerUi.isInsideClosingRadius(movingEvent)
         // Check if an view is in expected state and if not animate it into the correct state
-        val expectedVisibility = if (showClosingOverlayView) View.VISIBLE else View.GONE
-        if (binding.closingOverlay.visibility != expectedVisibility) {
+        if (binding.closingOverlay.isGone) {
```
Stypox, line 252: I think this is what you meant
```suggestion
        if (binding.closingOverlay.isVisible != showClosingOverlayView) {
```

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/player/gesture/PopupPlayerGestureListener.kt
TacoTheDank: oh my god

7sj2qs

AudricV (dismissed):
Are you sure about your changes? A part of them has been intentionally reverted with #8737.

TacoTheDank:
@AudricV That's interesting, I wasn't aware of that.

If the problem is that `(playerUi.screenWidth - playerUi.popupLayoutParams.width).toFloat()` or `(playerUi.screenHeight - playerUi.popupLayoutParams.height).toFloat()` can somehow result in a negative number and therefore crash because it is less than the minimum (0f), then I think we can just use Kotlin's `coerceAtLeast` to make sure numbers can't go below '0f'. `coerceAtLeast` looks like this:

```
public fun Float.coerceAtLeast(minimumValue: Float): Float {
    return if (this < minimumValue) minimumValue else this
}
```

This would ensure we always have a maximum of at least '0f' and avoid a crash.

Stypox (approved):

Stypox:
ktlint fails

TobiGr (approved):
LGTM

PR #10436, opened 2023-09-20
Fix restoring software license dialog

TobiGr:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

The last viewed license was re-opened after an orientation change in the About and License Fragments even if the license view/dialog has been closed by the user earlier. I needed to move the method creating the AlertDialog into the Fragment class to access `activeLicense`. An alternative implementation would have been to pass the listener as variable. But I think that the AlertDialog creation is part of the Fragment and should therefore be in the LicenseFragment and not a helper class.

The dialog buttons to open the software's website is also restored correctly after a rotation change now.

| Before | After |
|--------|--------|
| https://github.com/TeamNewPipe/NewPipe/assets/17365767/6efedaaf-8680-4481-be06-0ef86357ce21 | https://github.com/TeamNewPipe/NewPipe/assets/17365767/82f54007-f2c3-4139-8f87-a2452daa5eba | 

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/about/LicenseFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/about/LicenseFragment.kt
@@ -14,6 +14,7 @@ import io.reactivex.rxjava3.core.Observable
 import io.reactivex.rxjava3.disposables.CompositeDisposable
 import io.reactivex.rxjava3.disposables.Disposable
 import io.reactivex.rxjava3.schedulers.Schedulers
+import org.schabi.newpipe.BuildConfig
 import org.schabi.newpipe.R
 import org.schabi.newpipe.databinding.FragmentLicensesBinding
 import org.schabi.newpipe.databinding.ItemSoftwareComponentBinding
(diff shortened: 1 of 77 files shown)


Stypox (changes requested):
The neutral button "open website license" disappears on rotation, right? Since you don't distinguish between `SoftwareComponent` and `License`. Also, setting the `activeLicense` should happen from inside `showLicense`, not right before calling it.

app/src/main/java/org/schabi/newpipe/about/LicenseFragment.kt
```diff
@@ -57,26 +66,73 @@ class LicenseFragment : Fragment() {
...
+                    val webView = WebView(context)
+                    webView.loadData(webViewData, "text/html; charset=UTF-8", "base64")
+
+                    Localization.assureCorrectAppLanguage(context)
+                    val builder = AlertDialog.Builder(requireContext())
+                        .setTitle(softwareComponent.name)
+                        .setView(webView)
+                        .setOnCancelListener { activeSoftwareComponent = null }
+                        .setOnDismissListener { activeSoftwareComponent = null }
+                    if (softwareComponent == NEWPIPE_SOFTWARE_COMPONENT) {
```
Stypox, line 109: I think it's perfectly fine if you show "dismiss" and "open website" even for NewPipe, to further simplify the code and make things consistent. After solving (or dismissing) this comment feel free to merge directly, thanks! The code looks much better

Stypox (approved):

app/src/main/java/org/schabi/newpipe/about/LicenseFragment.kt
TobiGr: I think it is ok to handle NewPipe differently and not show the website button. We already do that in the AboutFragment.
However, we can use the same text on the positive button to close the dialog. I'd not use "ok", but rather "dismiss" or even better just "close".
 @opusforlife2 What button text would you use to close the license dialog? _ok_, _dismiss_ or _close_

Stypox: "done" was not an option 😂

TobiGr: Done is also ok

Issue #10509, opened 2023-10-24
Magnify button to change layout from straight to landscape


joe61fa6f1awe:
### Checklist

- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [X] I'm aware that this is a request for NewPipe itself and that requests for adding a new service need to be made at [NewPipeExtractor](https://github.com/TeamNewPipe/NewPipeExtractor/issues).
- [X] I have taken the time to fill in all the required details. I understand that the feature request will be dismissed otherwise.
- [X] This issue contains only one feature request.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Feature description

I noticed that the video-playing activity can add one more feature which is a magnify button to change the layout from straight to landscape, which means change the video to full screen.

### Why do you want this feature?

Not too many problems or limitations. 
I'm a student at the Australian National University. I have one course that asks us to write a feature, Junit test, or debug for Open Repo on GitHub. I read the contribution. md and I know you don't want a new feature in the current development phase. Therefore, I don't expect the PR will be accepted. However, we have to write a repo according to this development (No matter whether the PR is accepted or not, we have to at least publish a PR and receive the developer's reply.)  Therefore, I still wish you can let me open this issue to finish my assignment, which is really close to due time.

### Additional information

No

joe61fa6f1awe:
I'm due at this weekend 11:59 p.m.
I'm glad you replied to me and I will also move to the issue you provide.

PR #10521, opened 2023-10-29
Magnify button

joe61fa6f1awe:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

-  I added the full-screen button and exit-full screen button.
-  I try to implement the rotation method. However, it seems the APP's layout is locked and the developers don't want the APP to rotate. Therefore, even though I implemented those rotation methods on the screen, the screen doesn't change the layout.

- Before
There is no magnify button showing up on the screen.
You can see the picture in the [4478](https://github.com/TeamNewPipe/NewPipe/issues/4478)

- After

Normal layout

image

image

The are two full-screen buttons on the video. One is the default, which is settled by developers on the previous version. The second one on the list is settled by me. However, no matter which one both of them can't do the rotation

- Fixes 
- #  [10509](https://github.com/TeamNewPipe/NewPipe/issues/10509)
- #  [4478](https://github.com/TeamNewPipe/NewPipe/issues/4478) but no completed fix it because I'm not sure if the developer meant to let the APP screen can't rotate or not

-  I changed the setting of the string.xml 
-  I added the full-screen and exit full-screen button on fragment_layout_detail.xml
-  I also added new methods and buttons on the PlayQueueActivity class to show the buttons and implement the rotation method
-  I modify some designs on the globalScreenOrientationLocked method and the MainPlayerUI class because they seems make the button disappear and cause the screen to be locked.

#### APK testing

#### Due diligence


AudricV:
Thank you for your pull request.

It seems you didn't look at the right place to edit the current fullscreen button, as you would have found how to make the app rotate.

Here is what you should have noticed and done in my opinion (this list is not exhaustive and should not be used as an "ultimate reference"):

- The screen orientation with the fullscreen button is currently set in [`VideoDetailFragment.onScreenRotationButtonClicked`](https://github.com/TeamNewPipe/NewPipe/blob/6d694518fe1a6747bdb9599394766f996c0ac324/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java#L1876-L1893).
- In this [`VideoDetailFragment`](https://github.com/TeamNewPipe/NewPipe/blob/6d694518fe1a6747bdb9599394766f996c0ac324/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java) class, a [`ContentObserver`](https://developer.android.com/reference/android/database/ContentObserver?hl=en) attribute is used to remove the forced orientation when the device's automatic screen rotation is disabled.
- As the button is currently used when the device's automatic screen rotation is disabled, the usage of the [`SCREEN_ORIENTATION_PORTRAIT` flag of the `ActivityInfo` class](https://developer.android.com/reference/android/content/pm/ActivityInfo?hl=en#SCREEN_ORIENTATION_PORTRAIT) is made and would need to be changed to the [`ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED flag`](https://developer.android.com/reference/android/content/pm/ActivityInfo?hl=en#SCREEN_ORIENTATION_UNSPECIFIED).
- In the [`MainPlayerUI`](https://github.com/TeamNewPipe/NewPipe/blob/6d694518fe1a6747bdb9599394766f996c0ac324/app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java) class, the [`setupScreenRotationButton` method](https://github.com/TeamNewPipe/NewPipe/blob/6d694518fe1a6747bdb9599394766f996c0ac324/app/src/main/java/org/schabi/newpipe/player/ui/MainPlayerUi.java#L891-L898) is, like its name implies, responsible of hiding or showing the fullscreen button. This operation would need to be removed and the method could be renamed to `setupScreenRotationButtonIcon`.
- In this `MainPlayerUI` class, an other `ContentObserver` attribute is used to hide or show the fullscreen button using the `setupScreenRotationButton` method.
- The fullscreen button is also hidden by default in the XML layout in which it is declared, so you would need to remove the `android:visibility` property set to the `gone` value to the button.

Note that this approach has the disadvantage that if the device is physically in landscape (and so that the corresponding sensor reports this), the user enabled automatic rotation, and the app is in landscape, when the user will press the fullscreen button, nothing will happen.

Please don't do unrelated changes to the topic of your changes, so removing blank lines in unrelated files or changing a dependency should not be committed.

If you didn't do so, you should have asked yourself if putting a fullscreen button in a play queue activity makes sense or not (the answer is no).

I noticed this comment in your changes of `PlayQueueActivity`: `// If you're using an ActionBar, hide it`. I don't want to judge you, but in general, I think you should adapt answers from forums, blogs or AIs to the project on which you work on. This comment feels that you copy-pasted an answer.

Pull requests you made should be generally in the default branch we set, so except in very rare cases or bug regression fixes during a release candidate, you should use the `dev` branch and not another one like you did (this could be changed without closing a PR).

Finally, please try to explain your changes in a more direct way and follow the template properly in the future.

Due to the reasons above, we can't merge your pull request unfortunately.

You don't need to work on this feature anymore, I already started a while ago to work on this without submitting a pull request.

I am closing your pull request, please do not take this result as a bad experience but as a learning one :)

P.S.: That's a fullscreen button, not a magnify one.

PR #10618, opened 2023-12-02
Improve app signature check

Isira-Seneviratne:
#### What is it?

- Simplify the app signature check using [`PackageInfoCompat.hasSignatures`](https://developer.android.com/reference/androidx/core/content/pm/PackageInfoCompat#hasSignatures(android.content.pm.PackageManager,java.lang.String,java.util.Map%3Cbyte[],java.lang.Integer%3E,boolean)).

#### APK testing

#### Due diligence


app/src/main/java/org/schabi/newpipe/util/ReleaseVersionUtil.kt
```diff
@@ -1,97 +1,39 @@
...
-        }
-        return str.toString()
-    }
-
-    private fun showRequestError(app: App, e: Exception, request: String) {
-        createNotification(
-            app, ErrorInfo(e, UserAction.CHECK_FOR_NEW_APP_VERSION, request)
+        @Suppress("NewApi")
+        val certificates = mapOf(
+            RELEASE_CERT_PUBLIC_KEY_SHA256.toByteArray() to PackageManager.CERT_INPUT_SHA256
```
AudricV, line 22: Why are you using a constant available only on Android API 28 and higher and suppressing its availability warning? Is it a thing which would work on all devices thanks to compilation steps and `compileSdk` set currently to Android API 34?

Also, maybe you should move the certificates map as a constant in order to not run the map creation each time the method is called, as it can be called multiple times (see the usages of the `isReleaseApk` method). Let me know what you think about this.

AudricV (changes requested):

app/src/main/java/org/schabi/newpipe/util/ReleaseVersionUtil.kt
Isira-Seneviratne: > Why are you using a constant available only on Android API 28 and higher and suppressing its availability warning? Is it a thing which would work on all devices thanks to compilation steps and `compileSdk` set currently to Android API 34?

It gets inlined during compilation, as it's an integer value, so it won't cause issues on older devices.

> 
> Also, maybe you should move the certificates map as a constant in order to not run the map creation each time the method is called, as it can be called multiple times (see the usages of the `isReleaseApk` method). Let me know what you think about this.

Yeah, that sounds good.

Isira-Seneviratne:
> Did you test your changes by using your own singing key? I tried to do so and it did not work for me. What I did:
> 
> * Replace `RELEASE_CERT_PUBLIC_KEY_SHA256` with the fingerprint of my key
> * build release APK and sign it
> * install signed APK
> * no "Updates" option in the settings, thus `ReleaseVersionUtil.INSTANCE.isReleaseApk()` yields `false` in SettingsActivity.

My bad, I added the certificate SHA-256 digest earlier. I've updated the PR with the public key hash, could you recheck?

PR #10671, opened 2023-12-20
Fix application lagging with many main page tabs

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

A lot of time ago [a 1700-changes commit](https://github.com/TeamNewPipe/NewPipe/blame/6675d3e2cd4263b7c09f50322b27885fb1b7b5bd/app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java#L110) decided that all main page tabs would be fetched at once and stay in memory. Before customizable tabs were a feature, this made sense, but now users can add as many tabs as they want. Therefore if a user had a lot of tabs, they would all be in memory at once, making the application lag, as observed in https://github.com/TeamNewPipe/NewPipe/issues/10471#issuecomment-1749842668. This PR fixes the problem by just removing the line that tells the view pager to keep all tabs loaded at the same time, and instead rely on the default behaviour of only keeping in memory the current tab and the adjacent ones.

- Fixes https://github.com/TeamNewPipe/NewPipe/issues/10471#issuecomment-1749842668

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/MainFragment.java
@@ -194,7 +194,6 @@ public class MainFragment extends BaseFragment implements TabLayout.OnTabSelecte
         }
 
         binding.pager.setAdapter(null);
-        binding.pager.setOffscreenPageLimit(tabsList.size());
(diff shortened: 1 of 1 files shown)


TobiGr (approved):
LGTM

PR #10673, opened 2023-12-20
Fix transaction too large in channel tab fragments

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This PR contains a really hacky workaround, that avoids storing useless data in the `ChannelTabFragment` state. `ChannelTabFragment` uses the corresponding tab link handler to be able to fetch data, since unlike other things in the extractor, channel tabs are not supposed to be obtained from a url (e.g. provided by the user), but rather always from a channel page internally in the app.

The problem is, `ReadyChannelTabListLinkHandler` might contain raw JSON data that uses a lot of memory (e.g. ~800KB for YouTube). While 800KB doesn't seem much, if you combine just a couple of channel tab fragments you easily go over the 1MB save&restore transaction limit, and get `TransactionTooLargeException`s.

What this PR does is to discard the `ReadyChannelTabListLinkHandler` after the first time the raw data has been parsed (i.e. the fragment's `handleResult` has been called). A new link handler with the same properties is then created. The good thing is, even if now the raw data is not in the link handler anymore, if `getChannelTab` is called on the newly created link handler, the app cache will kick in to still prevent network requests. Note that I didn't test this link-handler-recreation with all services, I will let users do this (or @AudricV could provide some feedback).

A proper solution would require rethinking about `ReadyChannelTabListLinkHandler`s, but that's out of scope for 0.26.0. We might need to rethink how the cache works, and instead of returning link handlers with raw data (which turned out to be a really bad idea) , we should put the raw data directly in an extractor cache. 

You can test the changes by putting one channel from all NewPipe services (Youtube, Soundcloud, ...) as a main page tab. Then open settings and the app should not crash. All video/audio tabs from all channels should still allow e.g. playing in the background without crashing.

- Fixes https://github.com/TeamNewPipe/NewPipe/issues/10471#issuecomment-1826689278

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelTabFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelTabFragment.java
@@ -1,6 +1,7 @@
 package org.schabi.newpipe.fragments.list.channel;
 
 import android.os.Bundle;
+import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
@@ -14,7 +15,9 @@ import org.schabi.newpipe.error.UserAction;
 import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.ListExtractor;
 import org.schabi.newpipe.extractor.channel.tabs.ChannelTabInfo;
+import org.schabi.newpipe.extractor.exceptions.ParsingException;
 import org.schabi.newpipe.extractor.linkhandler.ListLinkHandler;
+import org.schabi.newpipe.extractor.linkhandler.ReadyChannelTabListLinkHandler;
 import org.schabi.newpipe.extractor.stream.StreamInfoItem;
 import org.schabi.newpipe.fragments.list.BaseListInfoFragment;
 import org.schabi.newpipe.fragments.list.playlist.PlaylistControlViewHolder;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelTabFragment.java
```diff
@@ -114,6 +117,27 @@ public void setTitle(final String title) {
...
+        //  solution would require rethinking about `ReadyChannelTabListLinkHandler`s.
+        if (tabHandler instanceof ReadyChannelTabListLinkHandler) {
+            try {
+                // once `handleResult` is called, the parsed data was already saved to cache, so
+                // we can discard any raw data in ReadyChannelTabListLinkHandler and create a
+                // link handler with identical properties, but without any raw data
+                tabHandler = result.getService()
+                        .getChannelTabLHFactory()
+                        .fromQuery(tabHandler.getId(), tabHandler.getContentFilters(),
+                                tabHandler.getSortFilter());
```
TobiGr, line 134: This causes a crash. Use media.ccc.de service, open a channel / conference. `getChannelTabLHFactory()` is `null` for that service:

https://github.com/TeamNewPipe/NewPipeExtractor/blob/f276caf54a85a271904943e4aa6c57efd4249fec/extractor/src/main/java/org/schabi/newpipe/extractor/services/media_ccc/MediaCCCService.java#L53-L56


## Exception
* __User Action:__ ui error
* __Request:__ ACRA report
* __Content Country:__ DE
* __Content Language:__ de-
* __App Language:__ en
* __Service:__ none
* __Version:__ 0.26.0
* __OS:__ Linux Android 12 - 32

```
java.lang.NullPointerException: Attempt to invoke virtual method 'org.schabi.newpipe.extractor.linkhandler.ListLinkHandler org.schabi.newpipe.extractor.linkhandler.ListLinkHandlerFactory.fromQuery(java.lang.String, java.util.List, java.lang.String)' on a null object reference
	at org.schabi.newpipe.fragments.list.channel.ChannelTabFragment.handleResult(ChannelTabFragment.java:133)
	at org.schabi.newpipe.fragments.list.channel.ChannelTabFragment.handleResult(ChannelTabFragment.java:37)
	at org.schabi.newpipe.fragments.list.BaseListInfoFragment.lambda$startLoading$0(BaseListInfoFragment.java:150)
	at org.schabi.newpipe.fragments.list.BaseListInfoFragment.$r8$lambda$yX6S2nKGSbR70YowNdSpGemC7q4(BaseListInfoFragment.java:0)
	at org.schabi.newpipe.fragments.list.BaseListInfoFragment$$ExternalSyntheticLambda4.accept(R8$$SyntheticClass:0)
	at io.reactivex.rxjava3.internal.observers.ConsumerSingleObserver.onSuccess(ConsumerSingleObserver.java:62)
	at io.reactivex.rxjava3.internal.operators.single.SingleObserveOn$ObserveOnSingleObserver.run(SingleObserveOn.java:81)
	at android.os.Handler.handleCallback(Handler.java:938)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7870)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)

```
</details>
<hr>

Issue #9764, opened 2023-02-09
[Android 13+] Missing notification player actions


benazarenko:
### Checklist

- [X] I am able to reproduce the bug with the [latest version](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.25.0

### Steps to reproduce the bug

Play any song,
Swipe down to see the notification player
Swipe the notification player, or just tap on it

### Expected behavior

Next/previous track buttons, pause/play, repeat, close buttons are visible.

### Actual behavior

Only next/previous track and play/pause buttons are visible. Changing it in preferences does nothing. Not mentioning that I have enough space to have all five of them.

### Screenshots/Screen recordings

<img alt="photo_2023-02-09_15-26-51" src="https://user-images.githubusercontent.com/7333912/217812763-deab4776-c695-4b90-9f66-abf87e144939.jpg" width=270 />

### Affected Android/Custom ROM version

Android 13

### Affected device model

Galaxy M12

mendhak:
Seeing the same on Android 13, Pixel 6. I've got items checked for the player notification under settings, but they don't appear in the notification.

Additionally, maybe it's related, also noticed that the background image of the notification disappears if you switch from normal playing mode to background playing mode. 

![Screenshot_20230212-231721~2](https://user-images.githubusercontent.com/746276/218343292-1a324084-b83e-49b1-878f-3444e4d94b2f.png)

![Screenshot_20230212-231729~2](https://user-images.githubusercontent.com/746276/218343312-b4f82c95-da7d-4ef5-8140-9fd2c7a6d32d.png)

gqqjhffr:
I have this same issue on GrapheneOS on Pixel 4a

Graves98:
Same problem here, on Galaxy S20, Android 13

[...]

PR #10580, opened 2023-11-16
Android 13+ custom notification actions

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
- Supersedes https://github.com/TeamNewPipe/NewPipe/pull/10567 because the code there was too hacky.
- Create class `NotificationActionData` and move there (from `NotificationUtil`) the function that turns actions from the enum format (i.e. the number stored into preferences) into the triple (action string, name, icon). This allows deduplication of such function, instead of passing arrays around like in https://github.com/TeamNewPipe/NewPipe/pull/10567.
- Set custom action providers in the media session to customize actions on Android 13+.
- Use the workaround initially suggested in https://github.com/TeamNewPipe/NewPipe/pull/10567 to basically tell the system that we're not able to handle "prev" and "next", in order to have 4 customizable action slots instead of just 2.
- On Android 13+ normal notification actions are useless, so they are not set into the notification builder anymore, to save battery
- The opposite happens on Android 12-, where media session actions are not set because they don't seem to do anything
- Exclude the play-pause and play-pause-buffering actions from ever being set to media session actions, as the play-pause-buffering action is always shown by the system. The only difference between the system-provided play-pause-buffering and ours is that when the a video completes, the replay icon is not shown, but just a normal "play".
- Currently the UI in the settings has not been changed. This can be done in another PR, which should just remove one of the 5 customizable actions and the play-pause* actions in the settings (just for Android 13+), along with a settings migration that should kick in on Android system updates.

#### Before/After Screenshots/Screen Record

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

#### Fixes the following issue(s)

- Fixes #9764

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

Issue #7880, opened 2022-02-15
Reassign hardware Next/Prev Buttons


Mrnofish:
### Checklist

- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I'm aware that this is a request for NewPipe itself and that requests for adding a new service need to be made at [NewPipeExtractor](https://github.com/TeamNewPipe/NewPipeExtractor/issues).
- [X] I have taken the time to fill in all the required details. I understand that the feature request will be dismissed otherwise.
- [X] This issue contains only one feature request.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Feature description

AntennaPod has an incredibly neat feature that allows reassigning the NEXT and PREV hardware buttons to different functions. 

Specifically there are two separate settings, one for NEXT and one for PREV, each button can be assigned to any of four functions:
- PREV (Restart Episode / Video)
- NEXT (Skip Episode / Next Video)
- FFWD (Fast Forward)
- REW (Rewind)

This allows for:
1. Replacing the default NEXT / PREV function set with FFWD / REW.
2. Swapping the controls, so PREV can become NEXT and viceversa.
3. Replacing AND swapping, so NEXT => REW and PREV => FFWD

The rationale for case 3 is that, depending on the specifics of the control scheme, one gesture or multi-tap might be easier / faster, and the user will want to assign their most oft-used function to the more accessible control.

For instance, on device might use double tap on right for NEXT, and triple tap on right for PREV. But you use REW more than FFWD, so you will want to use the double tap for REW.

### Why do you want this feature?

Almost all True Wireless headsets only have gestures or multi-taps for Next and Prev, not REW / FFWD. This seems to be very common in general with Bluetooth headphones and headsets that have hardware buttons.

On YouTube there is a lot of podcast-like content that isn't available in true podcast format elsewhere.

[...]

PR #10712, opened 2023-12-29
[Android 13+] Restore support of custom notification actions

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Supersedes https://github.com/TeamNewPipe/NewPipe/pull/10580 and https://github.com/TeamNewPipe/NewPipe/pull/10567 because the **hacky workaround** to allow customizing 4 actions [turned out to cause problems](https://github.com/TeamNewPipe/NewPipe/pull/10580#discussion_r1435593734)
- Set notification actions only via `MediaSessionConnector` on Android 13+, and only via the notification builder on Android 12-, to save battery
- Notification actions in `MediaSessionPlayerUi` are only updated when something really changes, to avoid `setCustomActionProviders()` calling `invalidateMediaSessionPlaybackState()` more times as needed, which I figure might be costly
- Adapts the player notification settings page to allow customizing only two actions on Android 13+. The preference keys for the first three actions are actually still set, to avoid complicating the logic too much, and to allow seamless transitioning between Android 12- and 13+ back and forth (e.g. when a user exports and imports a database).
- Allows each notification slot to contain any possible action (which required changing how compact notification slot indices are computed, since the assumption that "Nothing" slots are always last does not hold anymore)
- Allows play/pausing from the notification when the player is buffering, which is in line with what happens in the main player ui
- Cleans up the code by extracting `NotificationSlot` from `NotificationActionsPreference`, and `NotificationActionData` from `NotificationUtil`

I tested a bit and caught some small inconsistencies in the code that now are fixed, and everything seems to work as expected.

<table>
  <tr>
    <th>Before notification</th>
    <th>After notification</th>
    <th>Settings in 12-</th>
    <th>Settings in 13+</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
    <td>After</td>
    <td>After</td>
  </tr>
</table>

- Fixes https://github.com/TeamNewPipe/NewPipe/issues/9764

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/custom/NotificationSlot.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/custom/NotificationSlot.java
@@ -89,7 +89,7 @@ class NotificationSlot {
     void setupCheckbox(final View view, final boolean isCompactSlotChecked) {
         final CheckBox compactSlotCheckBox = view.findViewById(R.id.notificationActionCheckBox);
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
-            // there are no compact slots to customize on Android 33+
+            // there are no compact slots to customize on Android 13+
             compactSlotCheckBox.setVisibility(View.GONE);
             view.findViewById(R.id.notificationActionCheckBoxClickableArea)
                     .setVisibility(View.GONE);
--- a/app/src/main/java/org/schabi/newpipe/settings/custom/NotificationSlot.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/custom/NotificationSlot.java
@@ -1,0 +1,172 @@
+package org.schabi.newpipe.settings.custom;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.content.res.ColorStateList;
+import android.os.Build;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.CheckBox;
+import android.widget.ImageView;
+import android.widget.RadioButton;
+import android.widget.RadioGroup;
+import android.widget.TextView;
+
+import androidx.appcompat.app.AlertDialog;
+import androidx.appcompat.content.res.AppCompatResources;
+import androidx.core.widget.TextViewCompat;
+
+import org.schabi.newpipe.R;
+import org.schabi.newpipe.databinding.ListRadioIconItemBinding;
+import org.schabi.newpipe.databinding.SingleChoiceDialogViewBinding;
+import org.schabi.newpipe.player.notification.NotificationConstants;
+import org.schabi.newpipe.util.DeviceUtils;
+import org.schabi.newpipe.util.ThemeHelper;
+import org.schabi.newpipe.views.FocusOverlayView;
+
+import java.util.Objects;
+import java.util.function.BiConsumer;
+
+class NotificationSlot {
+
+    private static final int[] SLOT_ITEMS = {
+            R.id.notificationAction0,
+            R.id.notificationAction1,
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/schabi/newpipe/settings/custom/NotificationSlot.java
```diff
@@ -0,0 +1,172 @@
...
+        ((TextView) view.findViewById(R.id.notificationActionTitle))
+                .setText(SLOT_TITLES[i]);
+        view.findViewById(R.id.notificationActionClickableArea).setOnClickListener(
+                v -> openActionChooserDialog());
+    }
+
+    void setupCheckbox(final View view, final boolean isCompactSlotChecked) {
+        final CheckBox compactSlotCheckBox = view.findViewById(R.id.notificationActionCheckBox);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            // there are no compact slots to customize on Android 33+
```
TobiGr, line 92: Either Android 13+ or API 33+
```suggestion
            // there are no compact slots to customize on Android 13+
```

Stypox:
Is https://github.com/TeamNewPipe/NewPipe/pull/10580#issuecomment-1868289126 a bug of this PR? I think it's more of a consequence of AudricV's PR that fixed ANR popups by showing a notification asap. I know it's not perfect but it's better to show the notification asap to avoid getting errors for taking too much time to show the foreground service notification.

About cutting down actions: you're right but maybe some users would find some use for some of those options (even if they (partially) overlap with the already shown ones), so I wouldn't remove them as of now.

TobiGr (approved):
Code looks good to me. I did some testing and was not able to find any issues on API 33, 32, 29, 26

dvalter:
For me it brings a huge regression compared to #10580 and pre-33 notifications.
On my devices (A13, LineageOS) it removes rewind/fast-forward from the default notification actions (and an option to configure force rewind/fast-forward for playlists).

It would be better to keep prev/next buttons semi-configurable for 33+ in settings rather than crippling their functionality.

Stypox:
> For me it brings a huge regression compared to https://github.com/TeamNewPipe/NewPipe/pull/10580 and pre-33 notifications.

Yeah, I know... But that's how Google intended it, and I agree they made a bad choice there...

Unfortunately we can't add a configurable switch to enable the hacky workaround for #10580 for various reasons
- maintenance is difficult when there are too many settings, due to more unexpected bugs popping up
- most users wouldn't stumble upon the setting anyway, even if they need it, considering how niche this change is
- users that enable the setting and then forget about it will start arguing that their bluetooth controls don't work

Stypox:
I guess that would be simple to do, yeah.

dvalter:
> Yeah, I know... But that's how Google intended it, and I agree they made a bad choice there...

They rarely make something new without screwing up something good and old.

> users that enable the setting and then forget about it will start arguing that their bluetooth controls don't work

That made me think, I have an idea, how to improve the UX in the most broken case. I'll open a new issue then.

dvalter:
@Stypox looks like #7880 is what I was thinking about.

Reassigning next/prev actions for the whole player should also do the trick with A13 notifications. And that may even be used in a "smart next/forward" mode to make notification less useless for a single video playback (which is 99% of my usage of NewPipe).

Stypox:
Thanks for looking it up! Yeah #7880 makes more sense than an option just for the notification

Stypox:
Ok

einzel:
Where should I post if I am still encountering a problem to this? I am on Android 14 and my notification icon shows no control buttons and only forwards to the app or closes by swiping.

PR #10717, opened 2023-12-30
Calculate cache key based on info type instead of item type

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

It didn't really make sense to consider two cache keys as equal based on the type of items contained within that list. The proper way is to distinguish the actual kind of `Info` being saved/retrieved, so I created a new `enum` independent from the extractor's one

This PR includes the extractor from https://github.com/TeamNewPipe/NewPipeExtractor/pull/1148 to showcase that the problem is fixed, since there MediaCCC channels and channel tabs have the same url.

#### APK testing

#### Due diligence


acrodemocide (approved):

Stypox:
>  InfoCache does not need to be reset after applying these changes because it has a lifetime of one session. Is this correct?

Yes, and also if it wasn't correct, all cache keys would have changed so there won't be conflicts

PR #10730, opened 2024-01-05
Simplify selection of video stream

Profpatsch:
I was trying to understand the logic here, and noticed the indirection via a QualityResolver interfaces is pretty unnecessary. Just branching directly makes the logic a lot easier to follow.

The `-999` sentinel value is a bit dumb, but java does not recognize that videoIndex is always initialized.

Nice side-effect, the `Resolver` interface was completely unused and can be dropped.

#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- record videos
- create clones
- take over the world

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/util/ListHelper.java
@@ -371,9 +371,8 @@ public final class ListHelper {
                 PreferenceManager.getDefaultSharedPreferences(context);
 
         // Load the preferred resolution otherwise the best available
-        String resolution = preferences != null
-                ? preferences.getString(context.getString(key), context.getString(value))
-                : context.getString(R.string.best_resolution_key);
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/schabi/newpipe/util/ListHelper.java
```diff
@@ -680,6 +674,14 @@ private static MediaFormat getMediaFormatFromKey(@NonNull final Context context,
         return format;
     }
 
+    /** #Comparator for two two resolution strings.
```
pihug12, line 677:
```suggestion
    /** #Comparator for two resolution strings.
```

Profpatsch:
Should I change the `switch` to an `if`? In my mind the switch tells the reader that this is intended to be exhaustive, but I don’t care either way

PR #10732, opened 2024-01-06
Don't write defaultFormat setting, use default value

Profpatsch:
Nowhere else does this (write a setting if it’s not set).

It took me a while to see that this code does not do what it intends, because `defaultFormat` is already the default value in the first `context.getString`, so calling `getMediaFormatFromKey` again is the exact same call (“do you know the definition of insanity…”) and will return `null` again …

So let’s drop the setting write and just rely on the default values.

#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

#### Due diligence


Profpatsch:
It’s a 2 line straight-forward function, what good would a test do?

PR #10735, opened 2024-01-06
DownloadDialog: remove thread selector

Profpatsch:
Exposing the number of download threads to users does not provide any value, because how should a user know how to set this value or what it means.

Instead, we just set the default to 4 threads for now (the previous default), with the possibility of improving the default later (e.g. depending on other running downloads or device speed).

#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

image

^ removes the threads slider on the bottom of the download popup

#### Due diligence


AudricV:
Thanks for your time, but I don't think it is a good idea to remove a feature like this one without any reason, especially as we will have a rewrite on which we will probably rework the download UI.

Your PR is also a "feature" PR, which we don't currently accept.

Due to the reasons stated above, I am closing your pull request.

PR #10781, opened 2024-01-23
BaseDescriptionFragment: Assert member is initialized

Profpatsch:
`streamInfo` and `channelInfo` have to be initialized, since the only way to construct the class it to pass them. So we can remove the null check boilerplate and make some of the accessors `NonNull`.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelAboutFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelAboutFragment.java
@@ -2,12 +2,12 @@ package org.schabi.newpipe.fragments.list.channel;
 
 import static org.schabi.newpipe.extractor.stream.StreamExtractor.UNKNOWN_SUBSCRIBER_COUNT;
 
-import android.content.Context;
 import android.os.Bundle;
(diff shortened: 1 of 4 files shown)


Stypox (approved):
Thanks!

app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelAboutFragment.java
```diff
@@ -26,15 +26,10 @@ public class ChannelAboutFragment extends BaseDescriptionFragment {
...
-        final ChannelAboutFragment fragment = new ChannelAboutFragment();
-        fragment.channelInfo = channelInfo;
-        return fragment;
+    ChannelAboutFragment(@NonNull final ChannelInfo channelInfo) {
+        this.channelInfo = channelInfo;
     }
 
-    public ChannelAboutFragment() {
-        super();
-    }
```
Stypox, line 37: Actually, the empty constructors are needed when resuming a fragment from memory:

`Caused by: java.lang.NoSuchMethodException: org.schabi.newpipe.fragments.list.channel.ChannelAboutFragment.<init> []`

PR #10782, opened 2024-01-23
SubscriptionsImportFragment: Inline setup stuff

Profpatsch:
Very small simplification

#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionsImportFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionsImportFragment.java
@@ -50,11 +50,11 @@ public class SubscriptionsImportFragment extends BaseFragment {
     @State
     int currentServiceId = Constants.NO_SERVICE_ID;
 
-    private List<SubscriptionExtractor.ContentSource> supportedSources;
-    private String relatedUrl;
+    private List<SubscriptionExtractor.ContentSource> supportedSources = Collections.emptyList();
+    private String relatedUrl = null;
 
     @StringRes
-    private int instructionsString;
+    private int instructionsString = 0;
 
     /*//////////////////////////////////////////////////////////////////////////
     // Views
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionsImportFragment.java
```diff
@@ -50,11 +50,11 @@ public class SubscriptionsImportFragment extends BaseFragment {
     @State
     int currentServiceId = Constants.NO_SERVICE_ID;
 
-    private List<SubscriptionExtractor.ContentSource> supportedSources;
-    private String relatedUrl;
+    private List<SubscriptionExtractor.ContentSource> supportedSources = Collections.emptyList();
+    private String relatedUrl = null;
```
acrodemocide, line 54: I know this was originally set based on the function that was removed, but, if I recall correctly, a declared String in Java is already null, so this assignment is not needed.

Profpatsch: Probably, but this just inlines the existing code, which I don’t want to change here.

Stypox (approved):

app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionsImportFragment.java
Stypox: Yeah there's no problem in making it explicit that it's initialized to null

Stypox:
Sorry, I just realized that according to Android fragment lifecycle, the `onCreate` function might be called multiple times, so this PR actually does change semantics. I would not risk making this change right before the rewrite, as we want to ship as stable of a release as possible. However, it's true that the `setupServiceVariables` method could be inlined, but it's not too important, and will probably be rewritten sooner or later anyway.

PR #10790, opened 2024-01-26
Ask for consent before checking for updates

TobiGr:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Ask for consent before enabling auto update check.

NewPipe is contacting its servers without asking for the users' consent. This is categorized as "tracking" by F-Droid (see https://github.com/TeamNewPipe/NewPipe/discussions/10785).

This PR disables checking for updates by default and adds a dialog asking for the user's consent to automatically check for updates if the app version is eligible for them. After upgrading to a version containing this commit the user is asked directly on the first app start. On fresh installs however, showing it on the first app start contributes to a bad onboarding an welcoming experience. Therefore, the dialog is shown at the second app start.

#### To Do

- [x] Implement settings migration triggering the consent check on update
- [x] Refactor the `App.isFirstRun` to be set inside App and read only (~~**should the variable be static?~~**)
- [x] Provide testing APK which does not rely on release signature check to ask for consent
- [x] Do testing

grafik

- Closes https://github.com/TeamNewPipe/NewPipe/discussions/10785

#### APK testing
Install this release APK for testing, singed by my dev key:
[NewPipe-check-consent.zip](https://github.com/TeamNewPipe/NewPipe/files/14692892/NewPipe-check-consent.zip)

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -79,6 +79,7 @@ import org.schabi.newpipe.player.Player;
 import org.schabi.newpipe.player.event.OnKeyDownListener;
 import org.schabi.newpipe.player.helper.PlayerHolder;
 import org.schabi.newpipe.player.playqueue.PlayQueue;
+import org.schabi.newpipe.settings.UpdateSettingsFragment;
 import org.schabi.newpipe.util.Constants;
 import org.schabi.newpipe.util.DeviceUtils;
 import org.schabi.newpipe.util.KioskTranslator;
@@ -86,6 +87,7 @@ import org.schabi.newpipe.util.Localization;
 import org.schabi.newpipe.util.NavigationHelper;
 import org.schabi.newpipe.util.PeertubeHelper;
 import org.schabi.newpipe.util.PermissionHelper;
+import org.schabi.newpipe.util.ReleaseVersionUtil;
 import org.schabi.newpipe.util.SerializedCache;
 import org.schabi.newpipe.util.ServiceHelper;
 import org.schabi.newpipe.util.StateSaver;
@@ -166,6 +168,11 @@ public class MainActivity extends AppCompatActivity {
             // Schedule worker for checking for new streams and creating corresponding notifications
             // if this is enabled by the user.
             NotificationWorker.initialize(this);
+        }
+        if (!UpdateSettingsFragment.wasUserAskedForConsent(this)
+                && ReleaseVersionUtil.INSTANCE.isReleaseApk()
+                && !App.getApp().isFirstRun()) {
+            UpdateSettingsFragment.askForConsentToUpdateChecks(this);
         }
     }
 
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -183,7 +183,7 @@ public class MainActivity extends AppCompatActivity {
         final App app = App.getApp();
         final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(app);
 
-        if (prefs.getBoolean(app.getString(R.string.update_app_key), true)) {
+        if (prefs.getBoolean(app.getString(R.string.update_app_key), false)) {
             // Start the worker which is checking all conditions
             // and eventually searching for a new version.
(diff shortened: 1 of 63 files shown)


AudricV (commented):
If the user closes the app when it gets the dialog and reopens it, does the dialog appear again? From what I understood of the code changes in this PR, it doesn't seem to be the case.

Code looks good to me, even if the placement of the alert dialog code in the update settings fragment seems a bit weird to me.

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -167,6 +169,10 @@ protected void onCreate(final Bundle savedInstanceState) {
             // if this is enabled by the user.
             NotificationWorker.initialize(this);
         }
+        if (ReleaseVersionUtil.INSTANCE.isReleaseApk()
+                && !UpdateSettingsFragment.wasUserAskedForConsent(this)) {
```
AudricV, line 173: Isn't running the checks in the contrary order better? I think for release builds, in the case when consent has been already asked, checking for consent first has a lower cost than searching for signatures.

AudricV (commented):
> I put it there because the dialog is working with the setting to keep that stuff in one place. For that reason, I did not put it into the MainFragment. If you want me to move it there, I'll do it.

That's fine, anyway, when this will be "compostified" in the future, it will be probably placed in the right place :)

Do you think that the first run code migration should be reverted? I don't think that's necessary, but do what you think is the best :)

The rest of the changes looks good to me except this comment (a small optimization). Thank you for your efforts!

AudricV:
> The migration is not necessary anymore because `wasUserAskedForConsen()` is checked on app start

I meant the refactoring of the first run code, sorry for the confusion.

Stypox:
Mmmh, I'm not sure about this change. It will be shown to all users when the app is started the first time, right? Wouldn't it be a not-so-welcoming first screen?

Maybe it would be possible to show it the second time the app is started, by checking if the database file has already been created to determine whether it is the first time the app is opened or not (or maybe we already have some mechanism in place, I don't remember). What do you think?

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -167,6 +169,10 @@ protected void onCreate(final Bundle savedInstanceState) {
             // if this is enabled by the user.
             NotificationWorker.initialize(this);
         }
+        if (!UpdateSettingsFragment.wasUserAskedForConsent(this)
+                && ReleaseVersionUtil.INSTANCE.isReleaseApk()) {
+            UpdateSettingsFragment.askForConsentToUpdateChecks(this);
+        }
```
Stypox, line 175: Mmmh, I'm not sure about this change. It will be shown to all users when the app is started the first time, right? Wouldn't it be a not-so-welcoming first screen?

I would prefer if it were shown the _second_ time the app is started, like this:
```suggestion
        if (!UpdateSettingsFragment.wasUserAskedForConsent(this)
                && ReleaseVersionUtil.INSTANCE.isReleaseApk()
                && !App.getApp().isFirstRun()) {
            UpdateSettingsFragment.askForConsentToUpdateChecks(this);
        }
```

Stypox (changes requested):
Code looks good :-)

Stypox:
Here an APK for you to test: [NewPipe updatechecker.zip](https://github.com/TeamNewPipe/NewPipe/files/14802832/NewPipe.updatechecker.zip)
This is the expected behavior:
- The **first** time the app is opened, no update check is run, and no consent is asked (this is to avoid a bad onboarding and welcoming experience)
- The **second** time, no update check is run, but the user is asked for consent
- If the user accepted to check for updates, from the **third** time onwards, the app will check for updates, otherwise it will never do so`

<sub>The APK signed with @Stypox keys; the fingerprint for which to check for updates has been set to `f3c6b6ddf4f4c9e2fa5c5e65799dc388d17eae626acfa6e29f571b3520883b75`, i.e. the fingerprint of my key; the version code of the APK is that of 0.26.0 (995) instead of 0.26.1 (996) so that version checks show a notification</sub>

Stypox (approved):
Thank you! I pushed two more commits to improve the checks a bit more (previously the default value when geting update_app_key was `true`, which would have resulted in update checks running without consent)

PR #10077, opened 2023-05-07
Migrate to non-transitive R classes

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Switch to [non-transitive R classes](https://blog.blundellapps.co.uk/speed-up-your-build-non-transitive-r-files/), which became the default starting in Android Gradle plugin 8.0. This fixes build issues with the CI pipeline and helps speed up the build process.

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #10918, opened 2024-03-28
Migrate to non-transitive R classes

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Replaces #10077.
- Switch to [non-transitive R classes](https://blog.blundellapps.co.uk/speed-up-your-build-non-transitive-r-files/), which became the default starting in Android Gradle plugin 8.0. This fixes build issues with the CI pipeline and helps speed up the build process.
- These were the cases where `R.id.*` did not exist anymore and had to be replaced by the fully-qualified path.
  - `R.id.content` -> `android.R.id.content` to find the root view of an activity. I tested this manually and it works.
  - `R.attr.colorSecondary` -> `android.R.attr.windowBackground` as the bottom toolbar background color. I don't know why it was `colorSecondary` before, but `windowBackground` yields the same color as before, I tested on API 33 and API 22 on all themes. Moreover `windowBackground` is defined directly [in styles.xml](https://github.com/TeamNewPipe/NewPipe/blob/9648525ac19207643c18d85c5b1afef8dfd21b08/app/src/main/res/values/styles.xml#L30), so it should be correct.
  - `R.attr.colorAccent` -> `android.R.attr.colorAccent` as the bottom toolbar foreground color. Again, I tested on API 33 and API 22 on all themes and noticed no difference.
  - `R.attr.selectableItemBackground` -> `android.R.attr.selectableItemBackground`
  - all exoplayer string/drawable resources (which btw are all private and are not meant to be used ;-) )

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/error/ErrorUtil.kt
+++ b/app/src/main/java/org/schabi/newpipe/error/ErrorUtil.kt
@@ -54,7 +54,7 @@ class ErrorUtil {
          */
         @JvmStatic
         fun showSnackbar(context: Context, errorInfo: ErrorInfo) {
-            val rootView = if (context is Activity) context.findViewById<View>(R.id.content) else null
+            val rootView = (context as? Activity)?.findViewById<View>(android.R.id.content)
             showSnackbar(context, rootView, errorInfo)
         }
 
@@ -71,7 +71,7 @@ class ErrorUtil {
         fun showSnackbar(fragment: Fragment, errorInfo: ErrorInfo) {
             var rootView = fragment.view
             if (rootView == null && fragment.activity != null) {
-                rootView = fragment.requireActivity().findViewById(R.id.content)
+                rootView = fragment.requireActivity().findViewById(android.R.id.content)
             }
             showSnackbar(fragment.requireContext(), rootView, errorInfo)
         }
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/schabi/newpipe/error/ErrorUtil.kt
```diff
@@ -54,7 +54,10 @@ class ErrorUtil {
          */
         @JvmStatic
         fun showSnackbar(context: Context, errorInfo: ErrorInfo) {
-            val rootView = if (context is Activity) context.findViewById<View>(R.id.content) else null
+            val rootView = if (context is Activity)
+                context.findViewById<View>(android.R.id.content)
+            else
+                null
```
Isira-Seneviratne, line 60:
```suggestion
            val rootView = (context as? Activity)?.findViewById<View>(android.R.id.content)
```

Isira-Seneviratne (commented):
Looks good to me overall, there's one small change I'd recommend making.

app/src/main/java/org/schabi/newpipe/error/ErrorUtil.kt
Stypox: Thanks, solved

Stypox:
Merging, so that nightly testers can report bugs in case there is anything we missed. As explained in the PR description, I tested all changes and found no issue.

PR #10952, opened 2024-04-06
Add an intuitive prefix for the duration of lists in the UI

bg172:
…and avoid the new prefix for single videos

#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- improvement for https://github.com/TeamNewPipe/NewPipe/pull/6045 - now a prefix is shown infront of the duration so that the meaning is more intuitive for the user

<table>
  <tr>
    <th>Before: </th>
    <th> After: </th>
  </tr>
  <tr>
    <th></th>
    <th>  </th>
  </tr>
</table>

- Fixes # https://github.com/TeamNewPipe/NewPipe/pull/6045#issuecomment-2035434620

- https://github.com/TeamNewPipe/NewPipe/pull/6045

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/Localization.java
+++ b/app/src/main/java/org/schabi/newpipe/util/Localization.java
@@ -245,7 +245,7 @@ public final class Localization {
      * @return a formatted duration String or {@code 0:00} if the duration is zero.
      */
     public static String getDurationString(final long duration) {
-        return getDurationString(duration, true);
+        return getDurationString(duration, true, false);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -274,8 +276,9 @@ public static String getDurationString(final long duration, final boolean isDura
         } else {
             output = String.format(Locale.US, "%d:%02d", minutes, seconds);
         }
+        final String durationPrefix = showDurationPrefix ? "⏱ " : "";
         final String durationPostfix = isDurationComplete ? "" : "+";
-        return output + durationPostfix;
+        return durationPrefix + output + durationPostfix;
```
AudricV, line 281: Shouldn't we make this a string resource, especially for RTL languages?

Stypox: I think it's better to keep it this way, since:
- "⏱" is an icon, not really a string, and most of NewPipe's UI sadly does not move icons around based on RTL/LTR
- The duration is formatted with `Locale.US` anyway
- We would probably need 4 strings then

Stypox (approved):

Issue #10930, opened 2024-04-01
Release v0.27.0 (please TEST!)


Stypox:
## Testing for regressions

Debug APK (built by our CI in https://github.com/TeamNewPipe/NewPipe/pull/10929): [NewPipe_v0.27.0_RC4_debug.zip](https://github.com/TeamNewPipe/NewPipe/files/15047944/NewPipe_v0.27.0_RC4_debug.zip)
Release APK (built and signed by @Stypox): [NewPipe_v0.27.0_RC4_release.zip](https://github.com/TeamNewPipe/NewPipe/files/15047940/NewPipe_v0.27.0_RC4_release.zip)

Please report **only regressions** (i.e. new issues) here, not issues that were already present in previous releases

## TODO

- [ ] Merge weblate
- [ ] Merge preferences fix
- [x] https://github.com/TeamNewPipe/NewPipeExtractor/pull/1163
- [x] #10957
- [x] #10959
- [x] #10952
- [x] https://github.com/TeamNewPipe/NewPipeExtractor/pull/1168
- [ ] https://github.com/TeamNewPipe/NewPipeExtractor/pull/1170

## NewPipeExtractor version

This version of NewPipe will ship with NewPipeExtractor version 0.23.2 (yet to be released)

<details><summary><h2>App changelog </h2></summary><p>

### New
- Add support for comment replies #10018 
- Add playlist description to playlist fragment #10091 
- Add option to reset settings #9236 
- Show overall duration of items in a playlist #6045 
- Allow ordering bookmarked playlists #8221 

### Improved
- [Android 13+] Restore support of custom notification actions #10712
  :warning: The first three notification actions (play/pause, previous and next) cannot be customized anymore on these Android versions, as they are set by the system
- Allow playing/pausing from the notification when the player is buffering #10712
- Added "free memory" check before downloading [Android N / API 24+] #10505 #10896
- Change the landscape layout of list card item #9975 
- Ask for consent before checking for updates #10790 
- Moved player notification setting to notification section #9580 

### Fixed
- Don't write defaultFormat setting, use default value #10732 
- Calculate cache key based on info type instead of item type #10717 
- Fix not considering nullability when comparing getAudioTrackType #10909 
- Trim search string and remove duplicate records from the database #8242 
- Various download fixes #10912 

### Localization and Documentation
- Update translations #10890 
- Use GitHub markdown to emphasise warning in Readme #10724 

### Development
[...]

PR #10992, opened 2024-04-23
Fix free storage space check for all APIs

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

The previous code for obtaining free storage space was broken:
- It had some logical flaws in the checks being performed (e.g. why an `SDK_INT` check was before the function call, while another was inside it?)
- It didn't work at all for sdcard or other storages that don't have an UUID (don't ask me why the sdcard on the emulator did not have an UUID, who knows).
- It crashed with non-SAF files, see https://github.com/TeamNewPipe/NewPipe/issues/10930#issuecomment-2070180960

This PR checks for free storage according to https://stackoverflow.com/q/31171838. It uses `Os.statvfs(filePath)` on non-SAF files, and `Os.fstatvfs(fileDescriptor)` on SAF files. Note that for SAF files a file descriptor needs to be opened in read mode, but I checked and this does not seem to create a file, and works when the file does not exist (which is basically always the case since it's the file we want to download). Also see https://pubs.opengroup.org/onlinepubs/9699919799/functions/fstatvfs.html.

I tested on API 22 emulated (both SAF and non-SAF), API 33 emulated (SAF), API 33 device (SAF) and checked the actual values returned by `getFreeStorageSpace()` both on sdcard and internal storage, and everything was correct.

- https://github.com/TeamNewPipe/NewPipe/issues/10930#issuecomment-2070180960

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
+++ b/app/src/main/java/org/schabi/newpipe/download/DownloadDialog.java
@@ -859,20 +859,19 @@ public class DownloadDialog extends DialogFragment
             return;
         }
 
-        // Check for free memory space (for api 24 and up)
-        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
-            final long freeSpace = mainStorage.getFreeMemory();
-            if (freeSpace <= size) {
-                Toast.makeText(context, getString(R.
-                        string.error_insufficient_storage), Toast.LENGTH_LONG).show();
-                // move the user to storage setting tab
-                final Intent storageSettingsIntent = new Intent(Settings.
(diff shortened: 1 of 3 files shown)


Stypox:
@somos61 thank you for testing, I have no idea why `docTree.getUri()` would return "downloads", but now I added a `catch` that ignores all errors.

somos61:
> @somos61 thank you for testing, I have no idea why `docTree.getUri()` would return "downloads", but now I added a `catch` that ignores all errors.

dunno why it was 'downloads' instead of 'Download' either, but now dwnl path for SAF is 'content://com.android.providers.downloads.documents/tree/raw:/storage/emulated/0/Download' and downloads ok, for all tested. Thank you

PR #11140, opened 2024-06-07
Adding Hash of Signing Key to README

shrimprugbysnowowl:
#### What is it?
- Simple Security Addition

For your consideration, I'm submitting a small edit to the README to add the hash of the signing key currently posted on the NewPipe website and bring it to the user's attention in installation method #2.

For android users looking to download the APK and install directly from Github, or using a different method like Obtainium, the responsibility of verifying the signing key of the downloaded APK is placed on the user. This is easily done with [AppVerifier](https://github.com/soupslurpr/AppVerifier). A signing certificate posted on both the website and in the README would improve user confidence in the download.

Thank you for your consideration.


shrimprugbysnowowl:
The hash of the signing key should remain consistent between releases.

-------- Original Message --------

shrimprugbysnowowl:
No problem. Correct, it's not the hash of the apk itself.

-------- Original Message --------

PR #11218, opened 2024-06-30
onBackPressdBugFix

Ibodwe:
#### What is it?
- [O] Bugfix (user facing)
- [ ] Feature (user facing)

original video
https://github.com/TeamNewPipe/NewPipe/assets/35055429/04e662b4-e9d4-4ee8-889c-b3d5e21175c2

after fix video 
https://github.com/TeamNewPipe/NewPipe/assets/35055429/f093ab2a-97ae-4c8f-a200-0df95ed6b263

description 
On a kiosk screen, after entering a video, going back, and then entering a new video, I think the stack of the previously entered video should be cleared. I will attach a video clip.

- Before:
- After:

- Fixes #

- 

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/kiosk/KioskFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/kiosk/KioskFragment.java
@@ -10,10 +10,12 @@ import android.view.ViewGroup;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.ActionBar;
+import androidx.fragment.app.Fragment;
 
 import org.schabi.newpipe.R;
 import org.schabi.newpipe.error.ErrorInfo;
 import org.schabi.newpipe.error.UserAction;
+import org.schabi.newpipe.extractor.InfoItem;
 import org.schabi.newpipe.extractor.ListExtractor;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/schabi/newpipe/fragments/list/kiosk/KioskFragment.java
```diff
@@ -173,4 +176,13 @@ public void showEmptyState() {
...
     }
+
+    @Override
+    protected void onItemSelected(final InfoItem selectedItem) {
+        final Fragment fragment = getFM().findFragmentById(R.id.fragment_player_holder);
+        if (fragment instanceof VideoDetailFragment videoDetailFragment && fragment.isVisible()) {
+            videoDetailFragment.removeLastStackItem();
+        }
+        super.onItemSelected(selectedItem);
+    }
```
Profpatsch, line 187: It seems to work, but I’m not sure if this is the right place to implement this functionality.

It feels a bit hacky to remove the last item when an item is selected, plus the whole “fetch video fragment via the fragment manager” type-cast looks fishy.

Stypox:
This is the desired behavior and some people rely on it (although I totally agree that it's an ugly behavior). We will work on a better UX during the refactor, that will be more intuitive and will make it so that going back does close the bottom sheet instead of going to the previous video.

PR #11322, opened 2024-07-19
version catolog file

Dheerajr444:


Isira-Seneviratne:
> Why is this file introduced? Why does it not contain any data? Provide more information please.

https://developer.android.com/build/migrate-to-catalogs

PR #11238, opened 2024-07-05
Migrate image loading from Picasso to Coil

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Migrate the image loading functionality from [Picasso](https://square.github.io/picasso/) to [Coil](https://coil-kt.github.io/coil/compose/). Picasso has not had a new release for some time, and it has been [recommended](https://github.com/square/picasso/issues/2203#issuecomment-823697384) to move to other libraries. Coil also allows the notification icon loading to be greatly simplified, and it provides default configurations for its memory and disk caches.
- Remove the image indicator toggle from the settings as Coil does not have that functionality. Logging has been enabled instead for debug builds.
- Enable RGB-565 support for low-RAM devices only.
- This change is needed for PR #11060 as it makes use of Coil's Compose functionality.

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #11383, opened 2024-08-01
Migrate related items fragment to Jetpack Compose

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Migrate the related items tab of the video fragment to Jetpack Compose.
- Add composable functions for rendering the list views of streams and playlists, which appear in the related items view.
- Update Kotlin and Jetpack Compose to their latest versions.

- Before:
  Screenshot_20240802_091804_NewPipe

- After:
  Screenshot_20240802_091819_NewPipe Related-items-Compose

- Fixes #

- #11238

#### APK testing

#### Due diligence


app/src/main/java/org/schabi/newpipe/ui/components/items/ItemList.kt
```diff
@@ -0,0 +1,114 @@
...
+            } else if (item is PlaylistInfoItem) {
+                NavigationHelper.openPlaylistFragment(
+                    fragmentManager, item.serviceId, item.url, item.name
+                )
+            }
+        }
+    }
+
+    // Handle long clicks for stream items
+    // TODO: Adjust the menu display depending on where it was triggered
```
acrodemocide, line 54: I'm a  little new to this code base, so please forgive me if I'm making recommendations that go against the normal process. Typically, I'll recommend that we create a new issue to address "TODO" items since then those can get prioritized and taken care of by anyone else interested in making contributions to the code base.

If this seems like a good idea, please create an issue, and provide a link to it here so that I can review the details, then I'll consider this resolved. Otherwise, if it's common to leave "TODO" comments in the code and address them as they are found by later engineers making changes, then I won't worry about it.

Let me know your thoughts.

app/src/main/java/org/schabi/newpipe/ui/components/items/ItemList.kt
```diff
@@ -0,0 +1,114 @@
...
+        {
+            selectedStream = null
+        }
+    }
+
+    val showProgress = DependentPreferenceHelper.getPositionsInListsEnabled(context)
+    val nestedScrollModifier = Modifier.nestedScroll(rememberNestedScrollInteropConnection())
+
+    if (mode == ItemViewMode.GRID) {
+        // TODO: Implement grid layout using LazyVerticalGrid and LazyVerticalGridScrollbar.
```
acrodemocide, line 71: See the comment I left above about "TODO" comments in general. Please let me know your thoughts.

app/src/main/java/org/schabi/newpipe/ui/components/items/stream/StreamListItem.kt
```diff
@@ -0,0 +1,85 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import androidx.compose.ui.unit.dp
+import org.schabi.newpipe.extractor.stream.StreamInfoItem
+import org.schabi.newpipe.ui.theme.AppTheme
+
+@OptIn(ExperimentalFoundationApi::class)
```
acrodemocide, line 26: Is it possible to use an element that doesn't require the ExperimentalFoundationApi optin? I realize it may not be possible, but if possible, I believe it reduces the risk of new bugs being introduced if the underlying element is removed from the library or fundamentally changed. I realize that I'm not as familiar with this code base, so the process here may be different, so please don't hesitate to disregard this comment. I just wanted to weigh in and find out your thoughts.

app/src/main/java/org/schabi/newpipe/ui/components/items/stream/StreamMenu.kt
```diff
@@ -0,0 +1,152 @@
...
+            }
+        )
+        StreamMenuItem(
+            text = R.string.download,
+            onClick = {
+                onDismissRequest()
+                SparseItemUtil.fetchStreamInfoAndSaveToDatabase(
+                    context, stream.serviceId, stream.url
+                ) { info ->
+                    // TODO: Use an AlertDialog composable instead.
```
acrodemocide, line 85: See my comments above about "TODO" comments. In general, I thought it would make more sense to capture this in a new issue so that it can be prioritized and picked up by anyone interested in contributing to the code base. Please let me know your thoughts on this.

app/src/main/java/org/schabi/newpipe/ui/components/video/RelatedItems.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.schabi.newpipe.ui.components.items.ItemList
+import org.schabi.newpipe.ui.components.items.stream.StreamInfoItem
+import org.schabi.newpipe.ui.theme.AppTheme
+import org.schabi.newpipe.util.NO_SERVICE_ID
+
+@Composable
+fun RelatedItems(info: StreamInfo) {
+    val sharedPreferences = PreferenceManager.getDefaultSharedPreferences(LocalContext.current)
+    val key = stringResource(R.string.auto_queue_key)
+    // TODO: AndroidX DataStore might be a better option.
```
acrodemocide, line 39: Unlike the other "TODO" comments I mentioned above, I think this one makes sense to leave in the code since it's a question to consider and isn't necessarily calling for a code change. I'm just commenting on this TODO to clarify that I don't think any action needs to be taken on this one, since I think it makes sense to consider at some future point when someone else is taking a look at this code and considering the question.

acrodemocide:
I left a few comments on the code changes. I'm pretty new to this code base, so please forgive me if my thoughts/suggestions are not in line with the normal process. I've been reviewing the contributions guidelines and seeking to stay within those rules. Otherwise, I'm going to grab the APK to test on my Android and report back if I find any issues with this build.

app/src/main/java/org/schabi/newpipe/ui/components/items/ItemList.kt
Isira-Seneviratne: I added that comment because the grid layout is not being used in this PR (the related items view will only be in the list format). I have implemented the grid layout in [a separate PR](https://github.com/TeamNewPipe/NewPipe/pull/11259).

app/src/main/java/org/schabi/newpipe/ui/components/items/stream/StreamListItem.kt
Isira-Seneviratne: That annotation is needed because the `combinedClickable` modifier is currently experimental.

app/src/main/java/org/schabi/newpipe/ui/components/items/stream/StreamMenu.kt
Isira-Seneviratne: I was planning to implement the download dialog in Compose in a separate PR.

acrodemocide:
I installed the APK for this branch on my Android and have been testing NewPipe by going through all the functionality I normally use for NewPipe as well as checked on a few pieces of functionality that I don't use every day but I believe are still important. Everything looks good. I'll keep an eye out if more changes come in so I can install the latest build to see if it's working.

acrodemocide (approved):

app/src/main/java/org/schabi/newpipe/info_list/dialog/StreamDialogDefaultEntry.java
```diff
@@ -41,10 +41,11 @@
...
 public enum StreamDialogDefaultEntry {
-    SHOW_CHANNEL_DETAILS(R.string.show_channel_details, (fragment, item) ->
-            fetchUploaderUrlIfSparse(fragment.requireContext(), item.getServiceId(), item.getUrl(),
-                    item.getUploaderUrl(), url -> openChannelFragment(fragment, item, url))
-    ),
+    SHOW_CHANNEL_DETAILS(R.string.show_channel_details, (fragment, item) -> {
+        final var activity = fragment.requireActivity();
+        fetchUploaderUrlIfSparse(activity, item.getServiceId(), item.getUrl(),
+                item.getUploaderUrl(), url -> openChannelFragment(activity, item, url));
+    }),
```
Stypox, line 48: Why did you need to make this change? Did requireContext() return an invalid context or something like that?

app/src/main/java/org/schabi/newpipe/ui/components/items/ItemList.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.schabi.newpipe.extractor.stream.StreamInfoItem
+import org.schabi.newpipe.info_list.ItemViewMode
+import org.schabi.newpipe.ui.components.items.playlist.PlaylistListItem
+import org.schabi.newpipe.ui.components.items.stream.StreamListItem
+import org.schabi.newpipe.util.DependentPreferenceHelper
+import org.schabi.newpipe.util.NavigationHelper
+
+@Composable
+fun ItemList(
+    items: List<InfoItem>,
```
Stypox, line 33: It would be great if this were made data-source-agnostic. This would mean creating an UiInfoItem class that can be obtained from either an extractor's StreamInfoItem or a database's LocalInfoItem. This way we could create lists containing items from different sources, with a single UI implementation. What do you think?

app/src/main/java/org/schabi/newpipe/ui/components/items/ItemList.kt
```diff
@@ -0,0 +1,114 @@
...
+
+        LazyColumnScrollbar(state = state) {
+            LazyColumn(modifier = nestedScrollModifier, state = state) {
+                listHeader()
+
+                items(items.size) {
+                    val item = items[it]
+
+                    if (item is StreamInfoItem) {
+                        val isSelected = selectedStream == item
```
Stypox, line 83: This should be `selectedStream === item`, right? Otherwise it's going to compare the fields, which might be slower (and is also not what we want).

app/src/main/java/org/schabi/newpipe/ui/components/items/ItemList.kt
Stypox: > Typically, I'll recommend that we create a new issue to address "TODO" items since then those can get prioritized and taken care of by anyone else interested in making contributions to the code base.

Yes, this would be a good idea for normal development. But since we are in the middle of a refactor an this code is going to be looked at again for sure, I wouldn't worry about opening issue and so, as it might just add unneeded overhead. We will need to look at all TODOs before finalizing the refactor.

app/src/main/java/org/schabi/newpipe/ui/components/items/stream/StreamListItem.kt
```diff
@@ -0,0 +1,85 @@
...
+@Composable
+fun StreamListItem(
+    stream: StreamInfoItem,
+    showProgress: Boolean,
+    isSelected: Boolean,
```
Stypox, line 31: This parameter should have a more descriptive name, such as `isMenuOpen` or something like that (feel free to suggest alternatives)

app/src/main/java/org/schabi/newpipe/ui/components/items/stream/StreamThumbnail.kt
```diff
@@ -0,0 +1,89 @@
...
+            )
+        }
+
+        if (showProgress) {
+            val streamViewModel = viewModel<StreamViewModel>()
+            var progress by rememberSaveable { mutableLongStateOf(0L) }
+
+            LaunchedEffect(stream) {
+                progress = streamViewModel.getStreamState(stream)?.progressMillis ?: 0L
+            }
```
Stypox, line 75: The view model itself should provide the progress in a `StateFlow`, there should be no need for this in the UI layer. Ideally I would suggest creating a `StreamStateRepository` through which ALL state changes should be done (e.g. a video being watched) and that pushes events in a `Flow`s of pairs `(stream_id, stream_state)`. Every stream item visible in the UI would subscribe to the repository's `Flow` and if it receives an event with a matching `stream_id` updates the `StateFlow`, which in turn will trigger a UI recomposition of that stream item.

Btw this is the approach I thought of in a few minutes, but there may be better ones to achieve the same thing.

PR #11489, opened 2024-08-31
Migrate video description to Jetpack Compose

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Migrate the video description UI to Jetpack Compose.

- Before:
  Screenshot_20240831-153532
  Screenshot_20240831-153554

- After:
  Screenshot_20240831-153604
  Screenshot_20240831-154034

- Video:

https://github.com/user-attachments/assets/16e235ec-cedd-42ee-94cc-1594f3823686

- Fixes #

#### APK testing

#### Due diligence


app/src/main/java/org/schabi/newpipe/ui/components/common/DescriptionText.kt
```diff
@@ -0,0 +1,45 @@
...
+@Composable
+fun DescriptionText(
+    description: Description,
+    modifier: Modifier = Modifier,
+    overflow: TextOverflow = TextOverflow.Clip,
+    maxLines: Int = Int.MAX_VALUE,
+    onTextLayout: (TextLayoutResult) -> Unit = {},
+    style: TextStyle = LocalTextStyle.current
+) {
+    // TODO: Handle links and hashtags, Markdown.
```
AudricV, line 27: This must be solved before the PR can be merged.

AudricV (changes requested):
Thank you! According to your what you provided, this finally fixes the scrolling issue when selecting text :tada:

In my opinion, the tag chips component isn't the right one. An `ElevatedSuggestionChip` or something similar would probably look a lot better and similar to the current one used in XML.

Like other metadata items and in the current design, the `Tags` text should be upper-case.

You should not talk about a video description but a stream description, there are audio-only supported services. So your new `VideoDescriptionSection` composable and its corresponding package should be renamed :)

Jean-BaptisteC:
I suggest three things on this PR:
- Reduce a little bit size of font
- Use grey like chip background
-  Add ability to scroll to the top just under the video (on little screen, like foldable phone, it's hard to scroll to the video description) -> maybe need to do in a separate PR?

Stypox:
@TobiGr yeah that's because comments are not cached anymore... This was a leftover from the comments PR. It's normal for Android tabs to get destroyed at a distance of 2 from the current position.

Isira-Seneviratne:
Rewriting the video details fragment using Compose should fix it; alternatively, in the short term it could be migrated to using a view model that could be shared with the comment fragment.

app/src/main/java/org/schabi/newpipe/ui/components/metadata/TagsSection.kt
```diff
@@ -0,0 +1,65 @@
...
+            text = stringResource(R.string.metadata_tags).uppercase(),
+            style = MaterialTheme.typography.titleMedium,
+            textAlign = TextAlign.Center
+        )
+
+        FlowRow(horizontalArrangement = Arrangement.spacedBy(4.dp)) {
+            for (tag in sortedTags) {
+                ElevatedSuggestionChip(
+                    onClick = {
+                        NavigationHelper.openSearchFragment(
```
cameocoder, line 45: In the interests of keeping composables "dumb", only responsible for the UI and not business logic, I suggest decoupling the Navigation logic from the composable.

Add an onClick lambda and basically buggble this up to the Fragment/Screen level.
```
onTagClick: (tag: String) -> Unit,
...
                    onClick = {
                        onTagClick(tag)
                    },
```

In the future when fragments are replaced with compose navigation, this should make it easier to implement the compose navigation.

app/src/main/java/org/schabi/newpipe/fragments/detail/DescriptionFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+
+class DescriptionFragment : Fragment() {
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ) = content {
+        AppTheme {
+            Surface(color = MaterialTheme.colorScheme.background) {
+                StreamDescriptionSection(requireArguments().serializable(KEY_INFO)!!)
```
cameocoder, line 25: Instead of forcing this to be non-null, you could wrap it with a let.  That way if the arguments happen to be null, the code renders a blank surface instead of crashing.

app/src/main/java/org/schabi/newpipe/ui/components/video/StreamDescriptionSection.kt
```diff
@@ -0,0 +1,224 @@
...
+                        modifier = Modifier.fillMaxWidth(),
+                        text = stringResource(R.string.description_select_note),
+                        style = MaterialTheme.typography.bodySmall
+                    )
+                }
+            }
+
+            if (hasDescription) {
+                item {
+                    val description = parseDescription(streamInfo.description, streamInfo.serviceId)
```
cameocoder, line 132: ParseDescription has to perform some work in order to generate the annotated string.  It is remembered, so hopefully it won't cause any unnecessary recompositions.  However since it has to do do some work, it would be best to have an AboutChannelViewModel and DescriptionViewModel where streaminfo/channelInfo could be transformed off the UI thread and then the transformed static data could be passed to the composables.
It may require the introduction of a loading state. 

I was heading in that direction here:
https://github.com/cameocoder/NewPipe/blob/channel-about-refactor/app/src/main/java/org/schabi/newpipe/fragments/list/channel/ChannelAboutViewModel.kt

Issue #5586, opened 2021-02-14
Enqueuing/playing a video from history should not queue the whole playlist


GeluOltean:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want

Hi everyone. This might be a pet peeve of mine, but I think that when enqueuing a particular video from the history, having the whole history playlist queued is a bit strange; I think it would be quite helpful to either treat the history playlist as "special" in this regard, or have the option of toggling this on or off (perhaps not just for the history playlist?).

#### How will you/everyone benefit from this feature?

The reason I ask for this is that I often "compose" playlists with different videos, but if I do it starting from the history playlist and I'm not paying enough attention, the whole of my history gets queued and any videos I add will be shot down all the way to the end, which is a bit unexpected.

MD77MD:
yeah,  this is really  Annoying !

Wessl:
Is anyone working on this? The way I personally use newpipe is to have on big playlist of all my music on youtube that I then queue songs from, which has been impossible for a while now. Surely this has to be one of the most important bugs in the app at this point? I have no idea what language newpipe is even written in but at this point I'm going to start to try and fix this myself.

Krysaczek:
Reproduction steps: 

- start NewPipe (no video is playing or is enqued), go to history
- press and hold on any video (preferably near the top)
- choose "Start playing in the background" 
- [whole history is enqued going from top to bottom]
- press and hold on any of the other videos in history
- choose "Enqueue"
- [said video gets put at the end of a queue and will be played after all of the other history videos]

Workaround I do is to start playing whatever video in background, pause and then enqueue several videos from history or even from search that I want to listen to. Then I go to the near end of the queue and pick the first one I enqueued. This way I circumvent whole enqued history and listen only to the newly added.

Solution could be:

- not enqueueing all of the history (obviously, but queue itself would be temporary/without history)
- enqueueing all of the history (as it is done now) but actually moving the first "Start playing in the background" video at the end of the queue, so it's either last one played and any other enqueued videos do actually start playing right after it (we keep queue history, that's a plus)
[...]

PR #11696, opened 2024-11-12
Only start playing single item in history screen

codyit:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- Holding an item in history and clicking "Start playing in the background" would now only add the current item in the play queue, instead of the full history which is usually massive.

N/A

- Fixes # https://github.com/TeamNewPipe/NewPipe/issues/5586

#### APK testing

#### Due diligence
- [ x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/history/StatisticsPlaylistFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/local/history/StatisticsPlaylistFragment.java
@@ -332,10 +332,6 @@ public class StatisticsPlaylistFragment
                             StreamDialogDefaultEntry.DELETE,
                             (f, i) -> deleteEntry(
                                     Math.max(itemListAdapter.getItemsList().indexOf(item), 0)))
-                    .setAction(
-                            StreamDialogDefaultEntry.START_HERE_ON_BACKGROUND,
-                            (f, i) -> NavigationHelper.playOnBackgroundPlayer(
-                                    context, getPlayQueueStartingAt(item), true))
                     .create()
                     .show();
         } catch (final IllegalArgumentException e) {


Stypox:
There are people that rely on this behavior (me included) and last time we changed something we got a lot of angry users. During the refactor we will look into redesigning the long-press menu to allow for all of these different options to co-exist (starting without enqueueing everything and starting a playlist from a specific item), but we can't just change one thing without thinking about the whole picture. Sorry about it, and thank you for your contribution anyway.

Stypox:
Oh wait, you are just doing the change for the history fragment. Then I think the change makes sense.

Stypox (approved):
Thank you!

originalflava:
This was a feature LOL

originalflava:
This was a feature lol.. downgrading to 0.27.2

qvalentin:
This really should be configurable. Having the history as a playlist is definitely a usecase. 
Should I open a new issue for this?

Profpatsch:
I agree that the fix here is the right thing, there should probably be a separate button to play the whole history in reverse order.

goyalyashpal:
> During the refactor we will look into redesigning the long-press menu to allow for all of these different options to co-exist (starting without enqueueing everything and starting a playlist from a specific item), ...
>
> \- @ Stypox _at https://github.com/TeamNewPipe/NewPipe/pull/11696#issuecomment-2469884975_

yeah, there should be similar set of options available for most all such items: play single, play, enqueue, enqueue next, enqueue all, etc...

ShareASmile:
> there should be similar set of options available for most all such items: play single, play, enqueue, enqueue next, enqueue all, etc...

Would you like to give your feedback in #12032 ..

codyit:
Haven't checked github for a while..

One common way to deal with collections is to have a selection mode then a button to quickly select all, this way you won't have to duplicate all the buttons. IMO that would make it nicer and have everyone happy.

I'm happy enough with what we have now or future so I won't be cutting or giving feedback on other issues, although semantically, I still strongly think long pressing an item in a list should always be for actions on the item itself only , as in that creates the least confusion for the most people, which I thought is widely accepted instead of a blue gold dress situation here.

ShareASmile:
> This really should be configurable. Having the history as a playlist is definitely a usecase.
> Should I open a new issue for this?

@qvalentin please open a new issue regarding it. 

Users are commenting here, but discussion should take place in a dedicated issue not here.

PR #11710, opened 2024-11-16
Implement splash screen using Core Splashscreen library

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

This PR implements the app splash screen using the [Core Splashscreen library](https://developer.android.com/jetpack/androidx/releases/core#core_splashscreen_version_10_2). This allows a splash screen similar to that automatically generated on Android 12 and later to be displayed on older versions of Android. On Android 12 and later, an animated logo is used.

More information on Android 12's splash screen functionality: [Splash screens](https://developer.android.com/develop/ui/views/launch/splash-screen)

* **Note**: The app icon is not displayed on Android 5.0 and 5.1. This is a [known limitation](https://developer.android.com/reference/kotlin/androidx/core/splashscreen/SplashScreen#themes).

- Before:

All versions of Android:

https://github.com/user-attachments/assets/0e4c755c-e0fa-4a35-8023-b0f4bbe0f327

- After:

**Android 5.0**

https://github.com/user-attachments/assets/24580770-e5c7-430d-b598-76b6016f0507

**Android 6.0**

[Screen_recording_20241116_143343.webm](https://github.com/user-attachments/assets/5e6e90b2-1617-4dd4-a295-922034ae8c54)

**Android 15 (light theme)**

https://github.com/user-attachments/assets/70f7e684-daee-4bbe-9ef9-5f5d5fd02626

**Android 15 (dark theme)**

https://github.com/user-attachments/assets/bd7961a5-27d9-479a-bb79-ada8fe67f8e0

- Fixes #

#### APK testing

#### Due diligence


Jean-BaptisteC:
Can you test on a Samsung devices ?
Because I know somes samsung devices shown app icon cut on splashscreen
Exemple here https://github.com/Exodus-Privacy/exodus-android-app/pull/362

Isira-Seneviratne:
> Can you test on a Samsung devices ?
> Because I know somes samsung devices shown app icon cut on splashscreen
> Exemple here https://github.com/Exodus-Privacy/exodus-android-app/pull/362

That doesn't seem to be necessary here, as this icon doesn't have a background (which is why the `Theme.SplashScreen.IconBackground` theme is used here).

Profpatsch:
My Pixel 6 is on Android 14, and there is not splash screen, it just enlarges the icon a little and then starts after maybe one second.

Is it really necessary to add another dependency to implement a feature that is just nice-to-have? According to https://composables.com/android-distribution-chart more than half of devices are already on version 12 or later, so this code will be relevant maybe for a year or two.

Profpatsch:
I guess removing the splace screen activity is a nice benefit, but maybe we don’t need a splash screen on older devices in the first place?

Isira-Seneviratne:
> I guess removing the splash screen activity is a nice benefit, but maybe we don’t need a splash screen on older devices in the first place?

The duration it is visible can be customised, so it could be used to mask the home screen's video loading process (if such a feature is needed).

Stypox:
Mmmh I also think this is not so useful. Sure, animations are nice, but I don't see a big issue with the current splashscreen. At some point we could even remove the current splashscreen and use the default provided since Android 12+, to make things simpler.

PR #11724, opened 2024-11-20
build.gradle: fix NewPipeExtractor hash

Profpatsch:
If we use the whole hash, jitpack returns a 404 for some reason (maybe because the URL gets too long).

This should fix that.

#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- record videos
- create clones
- take over the world

- Before:
- After:

- Fixes #

- 

#### APK testing

#### Due diligence


Profpatsch:
Okay, I think I understand it better now; jitpack generates the first time, which causes the build requests to time out; if one retries after a few minutes, it should work.

PR #11463, opened 2024-08-19
Migrate empty_state_view xml/view to Jetpack Compose

toliuweijing:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
As part of the [rewrite](https://github.com/TeamNewPipe/NewPipe/issues/11198) effort, this PR migrates the `empty_state_view` xml and view to Jetpack Compose. It contains the following
1. introduce `EmptyStateComposable` and specs used different fragments.
2. introduce `EmptyStateUtil` to provide java -> kotlin interpolation methods.
3. use a simple `MutableState<EmptyStateSpec>` to trigger recompositation upon ui state changes.

#### Before/After Screenshots/Screen Record

|feature|before|after
|-|-|-
|bookmark-light|s22|Screenshot_20240819_161005
|bookmark-dark|Screenshot_20240819_160210|Screenshot_20240819_160812

More screenshot is coming before PR commition.

#### Fixes the following issue(s)

- Fixes #
https://github.com/TeamNewPipe/NewPipe/issues/11198

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #11725, opened 2024-11-20
Lwj.compose migrate empty state view

Profpatsch:
Followup rebase for https://github.com/TeamNewPipe/NewPipe/pull/11463 due to technical reasons.

Please see its description.


Profpatsch:
A manual test clicking through the build on an empty newpipe seems that everything still works as expected.

Profpatsch:
Some kaomoji that had different variants before seem to be the same now, maybe we want to keep that.

Stypox:
Wait

Stypox:
I think that the component's parameters are needlessly complicated. The only three items in the spec should be:
- `modifier: () -> Modifier`
- `emojiText: () -> String`
- `descriptionText: () -> String`

Stypox:
<table>
<tr>
<td></td>
<td></td>
<td></td>
</tr><tr>
<td></td>
<td></td>
<td></td>
</tr><tr>
<td></td>
<td></td>
<td></td>
</tr>
</table>

Stypox:
I pushed a few more commits, and now the code is better in my opinion. These things still need to be improved though (I don't know if it's better to do them now or in a future PR):
- storing a Modifier in the EmptyStateSpec is bad practice and also doesn't make much sense (it's the purpose of parent views to decide the modifiers of the children, it's not a property of the children themselves)
- the spacing of the UI is not perfect everywhere, but at least with the latest commit the panel is always centered
- the error panel should be merged with the empty state panel, since:
  - only one of them is shown at a time
  - the checks for errors are often close to the checks for emptiness, so some checks could be simplified
  - I'm not totally convinced of this though
- the empty state for unsupported content in the channel fragment does not work, but it didn't work before either, and both my Layout Inspector and by View Designer don't work, so I can't debug right now (Android Studio decided to go on vacation or something)

Profpatsch:
Looks exceptionally good to me! (disclaimer: I didn’t test, since you already made screenshots)

PR #11759, opened 2024-11-29
Rewrite import and export subscriptions functionality using coroutines

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- Implement the subscription import/export functionality using `CoroutineWorker` for better performance and readability of the code.
- Make the import process cancellable (exporting usually completes too quickly to show a notification).
- Read/write subscription data using the [Kotlinx Serialization](https://github.com/Kotlin/kotlinx.serialization) library and plugin.
- Update WorkManager from 2.8.1 to 2.10.0 ([changelog](https://developer.android.com/jetpack/androidx/releases/work#2.10.0))
- Update the compile SDK to 35 (required by WorkManager).

**Export**
- Before:

[Screen_recording_20241129_071419.webm](https://github.com/user-attachments/assets/7b22cf90-15f9-4591-85c2-765e8f1ba125)

Android 6.0

- After:

[Screen_recording_20241129_064527.webm](https://github.com/user-attachments/assets/fc803d6c-b192-40b4-afe2-8882f2ad251e)

Android 6.0

https://github.com/user-attachments/assets/db72de5f-c263-4365-8f3e-cd4458df533d

Android 15

- Fixes #

#### APK testing

#### Due diligence


app/src/main/java/org/schabi/newpipe/local/subscription/workers/ImportExportJsonHelper.kt
```diff
@@ -0,0 +1,72 @@
...
+            throw InvalidSourceException("input is null")
+        }
+
+        try {
+            @OptIn(ExperimentalSerializationApi::class)
+            return json.decodeFromStream<SubscriptionData>(`in`).subscriptions
+        } catch (e: Throwable) {
+            throw InvalidSourceException("Couldn't parse json", e)
+        }
+    }
```
Profpatsch, line 56: From a cursory look, this seems to be a lot stricter than the previous json parser.


e.g. the `serviceId` would default to `0`, and items in the list would simply be skipped if they don’t conform to the schema.

Do we want to change that behavior?

app/src/main/java/org/schabi/newpipe/local/subscription/ImportConfirmationDialog.java
```diff
@@ -53,8 +85,8 @@ public Dialog onCreateDialog(@Nullable final Bundle savedInstanceState) {
     public void onCreate(@Nullable final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        if (resultServiceIntent == null) {
-            throw new IllegalStateException("Result intent is null");
+        if (mode == 0 && value == null && serviceId == 0) {
+            throw new IllegalStateException("Input data not provided");
         }
```
Profpatsch, line 90: This seems to crash when rotating the screen while the confirmation dialog is visible

Profpatsch: At that point the module is not yet restored, I don’t know what this is supposed to do?

Profpatsch: Should be fixed in the commit I added, I left out the check since it will always use the `@Parcelize` decoders from the `SubscriptionImportInput` subclasses.

Profpatsch:
I rebased on current `refactor`, and added a commit which kotlin-ifies the `SubscriptionImportWorker` input data a little, turning them into a sealed class that auto-parcelizes.

Unfortunately, there is no `Data.Builder().fromParcelable()` or similar, so the `Data` conversion has to be done manually still, but I centralized it in the sealed class so it should be straightforward as well.

app/src/main/java/org/schabi/newpipe/local/subscription/workers/SubscriptionImportWorker.kt
```diff
@@ -0,0 +1,217 @@
...
+                            MimeTypeMap.getFileExtensionFromUrl(input.url).ifEmpty { DEFAULT_MIME }
+                        NewPipe.getService(input.serviceId).subscriptionExtractor
+                            .fromInputStream(it, contentType)
+                            .map { SubscriptionItem(it.serviceId, it.url, it.name) }
+                    }
+
+                is SubscriptionImportInput.PreviousExportMode ->
+                    applicationContext.contentResolver.openInputStream(input.url.toUri())?.use {
+                        ImportExportJsonHelper.readFrom(it)
+                    }
```
Profpatsch, line 60: This is missing a bunch of error handling, for example:

* File that was selected might not exist
* File might not be able to be parsed as json, or have the wrong contents
* Channel might not exist
* ServiceId might not exist

All of these should probably be shown to the user, right now the worker crashes silently in the background and nothing happens.

Profpatsch:
Once the missing error handling is added on import & the question about how leniently to parse the json is resolved, I’d say LGTM!

PR #9592, opened 2022-12-25
Add support for Android Auto

haggaie:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Expose media controls and playback status on Android Auto, as well as a basic browsing interface, showing playlists, history, and supporting basic searches.

#### Issues, bugs, and missing features

* Items (playlists and streams) are not updated in Android Auto when they are in the app.
* When closing the player from its mini view, playback continues. It should be stopped instead.
* When starting playback on Android Auto then opening the app, the mini player doesn't appear and isn't synchronized with the current state when "opened" (opening the details page of a content with autoplay disabled, pressing the play button plays the content on the details page)
* Remote playlists are limited to their first page.
* when starting the app then connecting to an Desktop Head Unit. The DHU was blocked on the loading step happening before showing playback controls normally, the app was playing the audio in the meantime then the app crashed. 
* Non square thumbnails seems to forced to the square format in the miniplayer commands if you enable the setting of the Desktop Head Unit.
* Integration with voice search - voice search from within the app works, but search from the google assistant (e.g., "play <song> on NewPipe")  only supports applications from the Google Play Store.

#### Future enhancements

* Show progress bars for played items and for the one currently playing, if applicable (see https://developer.android.com/training/cars/media?hl=en#browse-progress-bar for the latter).
* Add more content to Android Auto.

#### Fixes the following issue(s)

- Fixes #1758

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #11829, opened 2024-12-26
Player UI list to kotlin

Profpatsch:
Small helper module

#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- record videos
- create clones
- take over the world

- Before:
- After:

- Fixes #

- 

#### APK testing

#### Due diligence


Profpatsch:
Okay I noticed a funny thing where threads would compete for modifying our UI list, so I added a Mutex around it.

Profpatsch:
The code quality thingy is mostly that I marked a function as deprecated and have not yet changed all use-sites because it’s not in-scope of this PR

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.kt
```diff
@@ -0,0 +1,124 @@
...
+     */
+    fun call(consumer: java.util.function.Consumer<PlayerUi>) {
+        // copy the list out of the mutex before calling the consumer which might block
+        val new = playerUis.runWithLockSync {
+            lockData.toMutableList()
+        }
+        for (ui in new) {
+            consumer.accept(ui)
+        }
+    }
```
Isira-Seneviratne, line 123:
```suggestion
    fun call(consumer: (PlayerUi) -> Unit) {
        // copy the list out of the mutex before calling the consumer which might block
        val new = playerUis.runWithLockSync {
            lockData.toMutableList()
        }
        for (ui in new) {
            consumer(ui)
        }
    }
```

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.kt
```diff
@@ -0,0 +1,124 @@
...
+            for (ui in lockData) {
+                if (playerUiType.isInstance(ui)) {
+                    when (val r = playerUiType.cast(ui)) {
+                        // try all UIs before returning null
+                        null -> continue
+                        else -> return@runWithLockSync r
+                    }
+                }
+            }
+            return@runWithLockSync null
```
Isira-Seneviratne, line 97:
```suggestion
            lockData.firstNotNullOfOrNull { playerUiType.kotlin.safeCast(it) }
```

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.kt
```diff
@@ -0,0 +1,124 @@
+package org.schabi.newpipe.player.ui
+
+import org.schabi.newpipe.util.GuardedByMutex
+import java.util.Optional
+
+class PlayerUiList(vararg initialPlayerUis: PlayerUi) {
+    var playerUis = GuardedByMutex(mutableListOf<PlayerUi>())
```
Isira-Seneviratne, line 7: [`Collections.synchronizedList()`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#synchronizedList-java.util.List-) could be used instead, which would remove the need for a separate mutex class.

```suggestion
    var playerUis = Collections.synchronizedList(mutableListOf<PlayerUi>())
```

Isira-Seneviratne (changes requested):
This looks like a good change, overall.

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.kt
```diff
@@ -0,0 +1,124 @@
...
+            return@runWithLockSync null
+        }
+
+    /**
+     * @param playerUiType the class of the player UI to return;
+     * the [Class.isInstance] method will be used, so even subclasses could be returned
+     * @param T the class type parameter
+     * @return the first player UI of the required type found in the list, or an empty
+     * [         ] otherwise
+     </T> */
```
Stypox, line 106:
```suggestion
     * [Optional] otherwise
     */
```

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.kt
Stypox: Does this work even when called from Java?

Stypox:
The rest looks good to me, too, thanks!

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.kt
Isira-Seneviratne: Yeah, but Unit.INSTANCE has to be returned.

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.kt
```diff
@@ -0,0 +1,124 @@
...
+                    new.add(ui)
+                }
+            }
+            lockData = new
+        }
+        // then actually destroy the UIs
+        for (ui in toDestroy) {
+            ui.destroyPlayer()
+            ui.destroy()
+        }
```
Isira-Seneviratne, line 77: In combination with the above change:

```suggestion
    fun <T : PlayerUi> destroyAll(playerUiType: Class<T>) {
        // short blocking removal from class to prevent interfering from other threads
        val (toDestroy, toKeep) = synchronized(playerUis) {
            playerUis.partition { playerUiType.isInstance(it) }
        }
        playerUis.retainAll(toKeep)

        // then actually destroy the UIs
        for (ui in toDestroy) {
            ui.destroyPlayer()
            ui.destroy()
        }
```

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUiList.kt
Isira-Seneviratne: I changed the parameter type back to `Consumer`; the synchronized list implementation automatically handles the synchronization needed.

Profpatsch:
@Isira-Seneviratne I don’t understand your review, you said this looks like a good change, but then you proposed to replace everything with SynchronizedList instead of using a Mutex?

Isira-Seneviratne:
> @Isira-Seneviratne I don’t understand your review, you said this looks like a good change, but then you proposed to replace everything with SynchronizedList instead of using a Mutex?

I meant that the idea of refactoring to ensure thread safety is a good one. Sorry if I caused any confusion.

Profpatsch:
Personally I don’t like the concept of a SynchronizedList very much, because you have to remember to only use it inside a `synchronized` block, otherwise you introduce bugs. But nothing will remind you to do that.

Isira-Seneviratne:
> Personally I don’t like the concept of a SynchronizedList very much, because you have to remember to only use it inside a `synchronized` block, otherwise you introduce bugs. But nothing will remind you to do that.

That's only when iterating through it without the Java forEach method. All other operations are synchronized.

Profpatsch:
Yeah, that makes it ever more devious!

Profpatsch:
So I’d rather not switch this from a simple Mutex around the list to `synchronizedList`, mainly because I don’t think having a partially thread-safe list really makes the implementation simpler, it’s just harder to reason about and easy to make mistakes.

Stypox:
Note: this conflicts with https://github.com/TeamNewPipe/NewPipe/pull/9592, so I'd rather merge it after that one is merged back in the refactor branch

Profpatsch:
Okay, since the new Android Auto PR was superseded (and the new one merged), I just rebased this and it still applies cleanly, so I think we can merge.

Stypox (approved):
I think some of Isira's suggestions apply even to the version marked with `Mutex`, but it's ok like it is now I think. Thank you!

PR #11870, opened 2025-01-02
Add link to donation page on app drawer

theScrabi:
At 38c3 we discussed how we could make the request for donations more visible. We discussed multiple models. Ultimately we could not agree on a good solution to actually ask the user for donations.
However, we all agreed that donations should be more visible than just show up in the about screen.

This will add the link for donations to the app drawer. When clicking it will open the donation page of the NewPipe website.


app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -924,4 +933,10 @@ private boolean bottomSheetHiddenOrCollapsed() {
         return sheetState == BottomSheetBehavior.STATE_HIDDEN
                 || sheetState == BottomSheetBehavior.STATE_COLLAPSED;
     }
+
+    private void openDonationPage() {
+        final Intent intent = new Intent(Intent.ACTION_VIEW,
+                Uri.parse(getString(R.string.donation_url)));
+        startActivity(intent);
+    }
```
AudricV, line 941: Use `ShareUtils` to create and start the intent.

AudricV (changes requested):

app/src/main/java/org/schabi/newpipe/MainActivity.java
Profpatsch: Done

Profpatsch:
I switched over to the `ShareUtils` call, and also added a small fun toast message when the user clicks the button (that displays for a short amount of time).

image

Stypox:
Shouldn't you do "openInBrowser" instead of "openInApp"?

Profpatsch:
Done

app/src/main/java/org/schabi/newpipe/MainActivity.java
theScrabi: dejavu? didn't I fix that, did i? O.o

theScrabi: I have this issue of fixing things and then forgetting to push. Did this happen again?

app/src/main/res/values-de/strings.xml
```diff
@@ -190,6 +190,7 @@
     <string name="website_title">Website</string>
     <string name="website_encouragement">Besuche die NewPipe Website für weitere Informationen und Neuigkeiten.</string>
     <string name="donation_encouragement">NewPipe wird von Freiwilligen entwickelt, die ihre Freizeit dafür verwenden, dir die beste Nutzererfahrung zu bieten. Gib etwas zurück, indem du den Entwicklern hilfst, NewPipe noch weiter zu verbessern, während sie sich an einer Tasse Kaffee erfreuen.</string>
+    <string name="donation_thank_you">Vielen Dank für deine Spende!</string>
```
Stypox, line 193: This just causes conflicts, translate it from Weblate after the PR is merged

Stypox (changes requested):

app/src/main/res/values-de/strings.xml
Profpatsch: Weblate can’t deal with this?

Profpatsch:
I removed the toast because people thought it is confusing. I do think something like the toast could be more encouraging :)

app/src/main/java/org/schabi/newpipe/MainActivity.java
Profpatsch: Probably!

app/src/main/res/values-de/strings.xml
Stypox: Not always

Profpatsch:
Okay, so now sonar is broken and that blocks the PR

Profpatsch:
Ah no, apparently somebody has to review again

Profpatsch (approved):

robcle:
As a user I would like to have the option to turn this off completely. Maybe some users donate already and do not want to see the button every time they open the app.

Stypox:
The button is at the bottom of the drawer, it's not in the way. Why would you want to disable it? @robcle

PR #11959, opened 2025-01-26
Fix loading StreamInfo twice on first VideoDetailFragment opening

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

While working on #11955 I realized that when `VideoDetailFragment` is first started, the video loading is initiated both by `VideoDetailFragment` and by the `NavigationHelper.openVideoDetailFragment()` function, leading to the stream info loading twice in the background. From the second time `openVideoDetailFragment()` is invoked, the `VideoDetailFragment` is already running in the background, and so its initial logic does not run.

To solve this bug I made it so that `NavigationHelper.openVideoDetailFragment()` first opens the `VideoDetailFragment` as if it were in collapsed bottom sheet state (i.e. with `url = null`, see how `VideoDetailFragment.wasCleared()` is called in `VideoDetailFragment.doInitialLoadLogic()`. I am not sure if this is just a hack or is the actual intended behavior, but fortunately we're going to ditch all of this code soon. I think this is quite safe to merge anyway, since the change happens in `NavigationHelper.openVideoDetailFragment()` which is clearly defined.

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -283,11 +283,11 @@ public final class VideoDetailFragment
     /*////////////////////////////////////////////////////////////////////////*/
 
     public static VideoDetailFragment getInstance(final int serviceId,
-                                                  @Nullable final String videoUrl,
+                                                  @Nullable final String url,
                                                   @NonNull final String name,
                                                   @Nullable final PlayQueue queue) {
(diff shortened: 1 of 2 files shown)


AudricV (approved):
Let's try this and see if users report things :P

If you want nitpicks, then change in the comment `videoUrl` to `url` and `video` to `stream`, as we can play audio-only streams :)

Stypox:
Fixed in 01bf855015a7ae3f5ec879b4051cb24fc3e85fb5, thanks for noticing ;-)

PR #11969, opened 2025-01-28
DownloaderImpl: Auto-close resources and simplify header setting

neosis91:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Optimized the `execute` method to improve readability and resource management.
- Used `try-with-resources` to ensure automatic closing of the `Response` object.
- Simplified HTTP header management with a more concise approach.
- Introduced `Optional` to handle request bodies and responses, reducing null checks.

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
+++ b/app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
@@ -137,7 +137,8 @@ public final class DownloaderImpl extends Downloader {
         }
 
         final okhttp3.Request.Builder requestBuilder = new okhttp3.Request.Builder()
-                .method(httpMethod, requestBody).url(url)
+                .method(httpMethod, requestBody)
+                .url(url)
                 .addHeader("User-Agent", USER_AGENT);
 
         final String cookies = getCookies(url);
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/DownloaderImpl.java
```diff
@@ -124,59 +124,43 @@ public long getContentLength(final String url) throws IOException {
...
-            final String headerName = pair.getKey();
-            final List<String> headerValueList = pair.getValue();
-
-            if (headerValueList.size() > 1) {
-                requestBuilder.removeHeader(headerName);
-                for (final String headerValue : headerValueList) {
-                    requestBuilder.addHeader(headerName, headerValue);
-                }
-            } else if (headerValueList.size() == 1) {
-                requestBuilder.header(headerName, headerValueList.get(0));
```
Profpatsch, line 158: Looking at the impl of `.header` it’s the same as `.removeHeader().addHeader()` so the change should be fine.

Profpatsch:
Good catch, we need to close the Request.

I also added a try block for the body reader, which should also be closed.

Profpatsch:
Using Optional for simple null handling is a little silly in my opinion, especially cause we can convert to kotlin in the future.

PR #12027, opened 2025-02-10
[YouTube] PoToken implementation with HtmlUnit

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This is an experimental PR that aims to provide an alternative `PoTokenProvider` implementation that does not rely on the system WebView. HtmlUnit is apparently able to simulate a browser well enough, and the generated poTokens work, however there are a couple of problems:
- The code does not compile unless `minSdk` is increased to `26` due to [this issue](https://stackoverflow.com/questions/49891730/invoke-customs-are-only-supported-starting-with-android-0-min-api-26). The APK built by this PR has `minSdk` increased to `26` for testing purposes, but this is obviously not the solution.
- `minifyEnabled` seems to create issues for release builds.
- While HtmlUnit seems to work without issues in an emulator, on my phone the debug version uses a lot of CPU and after some time the system terminates the app. The release version works fast enough though. *Let me know if it works on your phone, try both the debug (from CI) and the release APK (provided below).*
- HtmlUnit uses a forked version of Gecko which might conflict with the extractor one.

#### APK testing

**Release** APK built at commit 768db9fea2953721df5dd73ce87ddb5813eb1391 with `minifyEnabled=false` (otherwise it wouldn't build); it might be faster than the debug version: [app-release.zip](https://github.com/user-attachments/files/18742024/app-release.zip)

#### Due diligence


Isira-Seneviratne:
It seems the solution to the build issue is to [disable Instant Run](https://stackoverflow.com/a/53078104).

Stypox:
HtmlUnit would be a platform-independent solution that does not rely on the system webview. It could be an alternative for those phones where the WebView implementation is broken or results in invalid poTokens. However, this is mostly an experimental PR, since there are a few problems highlighted above with no clear solution, plus the HtmlUnit library would increase the APK size.

Isira-Seneviratne:
> HtmlUnit would be a platform-independent solution that does not rely on the system webview. It could be an alternative for those phones where the WebView implementation is broken or results in invalid poTokens. However, this is mostly an experimental PR, since there are a few problems highlighted above with no clear solution, plus the HtmlUnit library would increase the APK size.

The system webview is updatable starting with Android 5.0, so this PR might not be necessary.

Stypox:
> The system webview is updatable starting with Android 5.0, so this PR might not be necessary.

That is true in theory, but we all know device manufacturers tend to mess with default Android components more than they should...

ale5000-git:
> The system webview is updatable starting with Android 5.0, so this PR might not be necessary.

The WebView by Google is now Android 8.0+, so other devices are stuck with old WebViews adn website may break on them sooner or later.
There are also other WebViews but the common user would be unable to get them.

Isira-Seneviratne:
> > The system webview is updatable starting with Android 5.0, so this PR might not be necessary.
> 
> The WebView by Google is now Android 8.0+, so other devices are stuck with old WebViews and websites may break on them sooner or later. There are also other WebViews but the common user would be unable to get them.

In that case, it might be a good idea to bump the app's minimum SDK version to 26.

ale5000-git:
> In that case, it might be a good idea to bump the app's minimum SDK version to 26.

Personally I'm strongly against it, special for this app.
People may keep old phones for music (and YouTube).

ale5000-git:
@Stypox
Isn't possible to have the "Gecko" library as separate app (doesn't need to be a system WebView, so it will be easy to install it)?
In this way NewPipe (or even other apps if programmed to do it) can use it if it is installed, and it won't increase NewPipe size for the users that doesn't use it.

Stypox:
Closing for now, this was more of a (kind-of-successful) proof-of-concept

PR #12028, opened 2025-02-10
[YouTube] Rewrite PoToken functionality using coroutines

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Rewrite the PoToken functionality to use Kotlin coroutines instead of RxJava.

- Fixes #

#### APK testing

#### Due diligence


app/src/main/java/org/schabi/newpipe/util/potoken/PoTokenProviderImpl.kt
```diff
@@ -27,18 +30,16 @@ object PoTokenProviderImpl : PoTokenProvider {
...
 
-        try {
-            return getWebClientPoToken(videoId = videoId, forceRecreate = false)
-        } catch (e: RuntimeException) {
-            // RxJava's Single wraps exceptions into RuntimeErrors, so we need to unwrap them here
-            when (val cause = e.cause) {
+        return try {
+            runBlocking { getWebClientPoToken(videoId, forceRecreate = false) }
+        } catch (e: Exception) {
+            when (e) {
```
Profpatsch, line 36: Is that change to catch okay?

app/src/main/java/org/schabi/newpipe/util/potoken/PoTokenProviderImpl.kt
```diff
@@ -48,56 +49,52 @@ object PoTokenProviderImpl : PoTokenProvider {
...
+    private suspend fun getWebClientPoToken(videoId: String, forceRecreate: Boolean): PoTokenResult {
         // just a helper class since Kotlin does not have builtin support for 4-tuples
         data class Quadruple<T1, T2, T3, T4>(val t1: T1, val t2: T2, val t3: T3, val t4: T4)
 
         val (poTokenGenerator, visitorData, streamingPot, hasBeenRecreated) =
-            synchronized(WebPoTokenGenLock) {
-                val shouldRecreate = webPoTokenGenerator == null || forceRecreate ||
-                    webPoTokenGenerator!!.isExpired()
+            webPoTokenGenLock.withLock {
+                val gen = webPoTokenGenerator
```
Profpatsch, line 58: With the `GuardedByMutex` class that I thought up in https://github.com/TeamNewPipe/NewPipe/pull/11829/commits/a8da9946d1c9e531328d6a2a05bedc13b2e6df5e it would be possible to put the generator into the lock, making it impossible to access it without locking first. I think that’s a pretty easy win.

app/src/main/java/org/schabi/newpipe/util/potoken/PoTokenProviderImpl.kt
```diff
@@ -48,56 +49,52 @@ object PoTokenProviderImpl : PoTokenProvider {
...
-                        NewPipe.getPreferredContentCountry(),
-                        YoutubeParsingHelper.getYouTubeHeaders(),
-                        YoutubeParsingHelper.YOUTUBEI_V1_URL,
-                        null,
-                        false
-                    )
-                    // close the current webPoTokenGenerator on the main thread
-                    webPoTokenGenerator?.let { Handler(Looper.getMainLooper()).post { it.close() } }
+                    webPoTokenVisitorData = withContext(Dispatchers.IO) {
+                        val innertubeClientRequestInfo = InnertubeClientRequestInfo.ofWebClient()
```
Profpatsch, line 63: Do I read this correctly, before it was executed blocking on the main thread and now it’s dispatched to the IO thread?

app/src/main/java/org/schabi/newpipe/util/potoken/PoTokenProviderImpl.kt
```diff
@@ -48,56 +49,52 @@ object PoTokenProviderImpl : PoTokenProvider {
...
                     // create a new webPoTokenGenerator
-                    webPoTokenGenerator = PoTokenWebView
-                        .newPoTokenGenerator(App.instance).blockingGet()
+                    webPoTokenGenerator = PoTokenWebView.getNewPoTokenGenerator(App.instance)
 
                     // The streaming poToken needs to be generated exactly once before generating
                     // any other (player) tokens.
-                    webPoTokenStreamingPot = webPoTokenGenerator!!
-                        .generatePoToken(webPoTokenVisitorData!!).blockingGet()
+                    webPoTokenStreamingPot = webPoTokenGenerator!!.generatePoToken(webPoTokenVisitorData!!)
```
Profpatsch, line 87: A bit weird that `webPaTokenGenerator` can be null here? Shouldn’t it be guaranteed to be there?

app/src/main/java/org/schabi/newpipe/util/potoken/PoTokenWebView.kt
```diff
@@ -164,46 +158,41 @@ class PoTokenWebView private constructor(
...
-                        poTokenU8String = ""
-                        for (i = 0; i < poTokenU8.length; i++) {
-                            if (i != 0) poTokenU8String += ","
-                            poTokenU8String += poTokenU8[i]
-                        }
+                        poTokenU8String = poTokenU8.join(",")
                         $JS_INTERFACE.onObtainPoTokenResult(identifier, poTokenU8String)
                     } catch (error) {
                         $JS_INTERFACE.onObtainPoTokenError(identifier, error + "\n" + error.stack)
                     }""",
```
Stypox, line 190: Do not change the javascript in any place. This needs to work on browsers from 20 years ago, which some devices still use as webviews. If you check the PR adding po token support, I specifically worked on making the javascript as old as possible to avoid any incompatibilities, and neither `.join()` nor `stringToU8()` are good.

Stypox (changes requested):

app/src/main/java/org/schabi/newpipe/util/potoken/PoTokenWebView.kt
```diff
@@ -1,33 +1,44 @@
...
     // to be used exactly once only during initialization!
-    private val generatorEmitter: SingleEmitter<PoTokenGenerator>,
+    private val continuation: Continuation<PoTokenGenerator>,
 ) : PoTokenGenerator {
     private val webView = WebView(context)
-    private val disposables = CompositeDisposable() // used only during initialization
-    private val poTokenEmitters = mutableListOf<Pair<String, SingleEmitter<String>>>()
+    private val scope = MainScope()
+    private val poTokenContinuations =
+        Collections.synchronizedMap(ArrayMap<String, Continuation<String>>())
```
Profpatsch, line 37: Why do you need a synchronizedMap for this? If the list is only accessed in suspended functions there should be nothing parallel going on, and the ordering of suspend calls should take care of it (as long as everything is run in the same coroutine block)

Profpatsch (approved):

Profpatsch:
lol there is no way to retract my approval, thank you github. Obviously fix the thing @Stypox mentioned :)

app/src/main/java/org/schabi/newpipe/util/potoken/PoTokenWebView.kt
```diff
@@ -229,56 +218,46 @@ class PoTokenWebView private constructor(
...
+     * it so that multiple poToken requests can be generated in parallel, and the results will be
+     * notified to the right continuations.
      */
-    private fun addPoTokenEmitter(identifier: String, emitter: SingleEmitter<String>) {
-        synchronized(poTokenEmitters) {
-            poTokenEmitters.add(Pair(identifier, emitter))
-        }
+    private fun addPoTokenEmitter(identifier: String, continuation: Continuation<String>) {
+        poTokenContinuations[identifier] = continuation
     }
```
Profpatsch, line 242: So as far as I understand, `addPoTokenEmitter` is called exactly once inside `generatePoToken`.

`generatePoToken` is called exactly twice, both times in `getWebClientPoToken`. And here’s the kicker: in the same `suspend` block, right underneath each other.

So this leads me to believe that the whole infrastructure of having a list of continuations is not really necessary, because everything happens one after the other anyway? I mean yes, in theory they could run at the same time, but in practice there’s only two calls running, and in sequence.

Profpatsch: What I’m saying here is that rewrite to continuations all well and good, but I think we don’t even need such a complicated setup in the first place.

alimima:
https://github.com/TeamNewPipe/NewPipe/issues/12906

New PoToken method with VİDEO-ID

PR #11947, opened 2025-01-21
Migrate watch history to Jetpack Compose

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Rewrite the watch history UI using Jetpack Compose. As part of this change, the ItemList composable was refactored to use separate data classes, in order to further decouple it from the information provided by the extractor and make it more easily reusable.
- Clean up unused methods in StreamHistoryDAO.
- Reimplement the watch history retrieval from the Room database using AndroidX Paging for improved efficiency, as it supports pagination.
- Add stream card and grid composables.
- Update several relevant dependencies, including the Compose BOM version.
- Use loading indicator for placeholder items in ItemList.

#### Before/After Screenshots/Screen Record

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

https://github.com/user-attachments/assets/278e4fa0-a89b-459a-a3bb-2931c4f36dfb

#### Fixes the following issue(s)

- Fixes #

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

Issue #875, opened 2017-11-28
Shuffle limited to first n number of videos


DeadManIV:
- [ ] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.
- [ ] I checked if the issue/feature exists in the latest version.

The shuffle in background play of playlists currently doesn't shuffle the entire playlist. It shuffles only the first 50(? I'm not actually sure how many) videos. YouTube itself has the same problem. For reference my playlist is: https://www.youtube.com/playlist?list=FL3hm1ymIqH41ZKDnr1TW5Iw

This website fixes the issue: http://youtube-playlist-randomizer.valami.info/playlist3.php?pl=FL3hm1ymIqH41ZKDnr1TW5Iw

Perhaps NewPipe could use a similar solution for shuffling videos?

P.S. Thanks for all your work, NewPipe is an amazing app!

theScrabi:
Yes we should take care about this. @karyogamy can you confirm this?

karyogamy:
Currently, the shuffling is done every time a new partial playlist comes in, so that it doesn't use too much bandwidth to download the entire playlist. 

FYI, on Youtube, the partial playlist size is 100 items and is about 25kb. From a good connection, this request takes about 500ms. So we can see this may can add up (time-wise, as the request is sequential) if the user wants to open a long playlist.

Perhaps we can create an toggle to let user choose if they want to download the full playlist when starting to play. This should be an enhancement though, not a bug, as Youtube is doing the same as well.

theScrabi:
Yes this might be good.
Does Youtube tell how mch elements are in a Playlist before downloading it?

DeadManIV:
I believe YouTube has changed how their shuffle works, because I'm getting videos from all over my playlist now when I hit shuffle.
I was wondering if this feature was still planned for NewPipe?

Avelina9X:
This is still very much an issue. Any chance on getting on update on this? It is very annoying to listen to the same 100 songs when they are part of a much much larger playlist.

sven-hoek:
For anyone interested: As a workaround, you can scroll down to the end of the playlist and start playing one of the last songs on background ("Start here when backgrounded"). After that, you can switch to the background player and activate shuffle. It should then shuffle the whole list.
It would be cool, though, if it were possible to shuffle the whole list without scrolling down each time.

[...]

PR #12032, opened 2025-02-12
Redesigned long press menu

Stypox:
#### What is it?
- \[ x \] Feature (user facing)
- \[ x \] Codebase improvement (dev facing)

This PR replaces `InfoItemDialog` with a JetpackCompose-based bottom sheet menu. This new menu is supposed to replace all item long press menus across the app, i.e. streams, playlists, channels. The menu contains more details than before about the stream/playlist/channel that was clicked, laid out in a (imo) nice design.

Please provide feedback both on the code and on the UI, but please keep in mind the points below when suggesting an alternative layout. Note that the order of buttons is not yet finalized, and that at the moment the menu only works for streams.

<table>
  <tr>
    <th>Portrait</th>
    <th>Landscape/tablet</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

#### Motivation and design goals

- A text-based menu is difficult to read and only allows for a handful of actions, while an icon-based menu is more compact and also easier to use. There still are some labels to explain what the actions do though.
- The menu puts the most frequent actions always in the same places, so muscle memory can help perform press the buttons faster.
- Users keep asking for new long press menu actions, therefore the long press actions system should be easily extendable and the set of actions should possibly even be editable by the user.
- We want to provide the same experience across the app (hence same menu for streams, playlists, channels), which also deduplicates a lot of code.
- A menu at the bottom of the screen is easier to use in one-handed mode.
- The menu adapts well to different screen sizes, e.g. by showing more/less buttons or adapting the header with item information.

<sup>Some of these points arose after some discussions in the NewPipe Matrix channel</sup>

#### TODO

- [ ] Also use the dialog for playlists and channels
- [ ] Perform asynchronous stuff with coroutines instead of RxJava3
- [ ] Implement "Start playing from here" actions
- [ ] Handle Kodi action
- [ ] Decide whether to allow the user to customize the menu (there is already a mini settings button in the top right of the menu for that purpose, but now it doesn't do anything)
- [ ] Various other TODOs spread out in code
- [ ] Understand which issues are solved by this PR
- [ ] Add documentation
- [ ] Add UI tests (?)

#### APK testing

#### Due diligence
- \[ x \] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).


app/src/main/java/org/schabi/newpipe/ui/components/menu/LongPressMenu.kt
```diff
@@ -0,0 +1,586 @@
...
+            val ctx = LocalContext.current
+            val onUploaderClick = actions.popFirst { it.type == ShowChannelDetails }
+                ?.let { showChannelDetailsAction ->
+                    {
+                        showChannelDetailsAction.action(ctx)
+                        onDismissRequest()
+                    }
+                }
+
+            Column {
```
Isira-Seneviratne, line 143: [`FlowRow`](https://developer.android.com/develop/ui/compose/layouts/flow) could be used here instead.

Stypox: Mmmh, that's what I was using in the beginning, but I don't think you can have a FlowRow with a maximum amount of items per row, but also have an item that weighs more than the others (i.e. the header). Also, I wanted the buttons to expand as much as possible, to not waste any space.

Isira-Seneviratne: The linked documentation mentions a `maxItemsInEachRow` attribute.

Stypox: I know, but I need the header to weigh as if it were 4 items

Isira-Seneviratne: I was able to make the needed changes using FlowRow.

Stypox: Try now

Isira-Seneviratne: Still not working

Isira-Seneviratne: Here's the patch:

[Handle_LongPressMenu_layout_using_FlowRow.patch](https://github.com/user-attachments/files/18802258/Handle_LongPressMenu_layout_using_FlowRow.patch)

Stypox: Mmmh, I should have given you write access to the NewPipe repository (or maybe you had it already), which should be enough according to what GitHub says:

> [...] users with write access to TeamNewPipe/NewPipe can add new commits to your long-press-menu branch.

Anyways, thank you for the patch! I haven't tested it yet (will do tomorrow) but from a brief look it seems like the fact that the header is always 66% of the width means the buttons in the first row won't be aligned like the ones below. Also, the buttons should fill as much horizontal space as possible, so misclicks can still be caught, and there might also be more space for the button label text. One final (minor) thing is that I wanted the layout to scale well just based on the available space, not specifically based on whether the device is landscape or portrait, so the header sizing in your patch does not fulfill this requirement.

Isira-Seneviratne: > One final (minor) thing is that I wanted the layout to scale well just based on the available space, not specifically based on whether the device is landscape or portrait, so the header sizing in your patch does not fulfill this requirement.

That can be implemented using `currentWindowAdaptiveInfo().windowSizeClass`, like in the `determineItemViewMode()` method.

Stypox: Sure, but what's bad about using BoxWithConstraints and measuring the actual available space?

Isira-Seneviratne: Using it results in the IDE checking if the constraints are actually used.

Stypox: You are right that it is indeed much more complex, so we might want to change the requirements. In the meantime I will add more documentation on the requirements (so people reading the code know why it was made like that).

Stypox:
@NaN-Sz thanks for the feedback! We already thought of the More button but the bottom sheet dialog already has kind of a way to show more buttons when there are too many (i.e. the menu starts out partially expanded, and you can scroll up to see the other buttons). What do you think about that?

Also, do you think the buttons are too small in my screenshots? Your mockup has much bigger ones, that's why I'm asking. I find the buttons in my screenshots easy to press correctly but it's arguable. Maybe we could put a slider in the "long-press menu's settings menu" to pick a size for buttons. This would be quite a niche setting, but also easy to understand what it does (and we would put it only in the "long-press menu's settings menu", not in the app's settings).

shivang1809:
> or maybe like this?
> 
> 

Would it be a good idea to make it like only 3 or 4 most used options come on long press (delete, add to playlist, share, background ,etc). And then could be swiped up to show this all like it is. This will reduce vertical space. 

A cherry on cake will be if there's a customize option to chose which options should come on long press and all other in swipe up menu.

MD77MD:
the most important option is missing : Open Channel

Stypox:
@MD77MD no, it is still there, it was moved under the video name to save some space (it should be noticeable since it's colored in blue)

Stypox:
Idea to make the "open channel" link more accessible: adding a tiny "open" icon like [this one](https://fonts.google.com/icons?selected=Material+Icons:open_in_new:&icon.query=open&icon.size=24&icon.color=%23e3e3e3) after the channel name

Stypox:
@MD77MD would an icon like this make it clearer?

image

MD77MD:
> @MD77MD would an icon like this make it clearer?
> 
> image

it does...but to tell you the truth, its not necessary. i was just too hasty ... collerd in blue is enough

app/src/main/java/org/schabi/newpipe/ui/components/menu/LongPressAction.kt
Isira-Seneviratne, line 1: I believe it'd be a good idea to introduce separate Stream and Playlist data classes, like how I've done here: https://github.com/TeamNewPipe/NewPipe/pull/11947

That would remove the need to define multiple factory methods.

Stypox: Yeah that's a good idea. We should definitely have a data type for the UI layer that is independent of the types in the extractor or the database. I don't think I will do the change in this PR though, as it's already quite big. I'd rather first merge #11947 and then rebase on top of it (or the other way around).

Stypox:
I added three more actions and I need your feedback on them:
- They allow playing in background/popup/main-player a list of items (e.g. a playlist) starting from the currently selected item.
- This feature is already present in some menus of the non-refactored NewPipe, but not in all of them, and when it exists it replaces the corresponding action to play just the single item.
- Feedback is needed on the icons: I made them up by combining the base icon (headset, picture in picture, and play) with a "badge" icon from Material Icons, let me know if you can think of a better "badge" icon from the ones available [here](https://fonts.google.com/icons?icon.query=).
- Now you can play-from-here in the feed and in channels, which is currently not possible in non-refactored NewPipe!
- Currently there are too many buttons on screen, but there will be a way to disable and reorder them (see the button in the top-right, I still haven't implemented it though)

image

KraXen72:
i would suggest using the arrow-forward icon as the badge, and intersect it with the full-sized icon, same as enqueue next does. 

also, the labels overflowing doesen't look great, i'd keep it in one line if possible

otherwise, can't wait for this new bottom menu! looks awesome!

Stypox:
Here are 4 alternatives based on suggestions on Matrix and on @KraXen72's suggestions.
- The first and the fourth one look good imo, however the arrow-forward icon might not clearly indicate the intention to "play from here" (?)
- The second and the third look too crowded

image

image

image

image


> also, the labels overflowing doesen't look great, i'd keep it in one line if possible

Since "Background from here" is a quite long text I don't think there is a way to keep it to one line. But I moved the three icons a bit higher, so there is more space above the text, and I think it looks better now.

KraXen72:
top-left = 1, top-right = 2, bottom-left = 3, bottom-right = 4

personally, i think i most prefer 1 or 4
3 is not bad, but the vertical bar is too short and it also makes the icons look quite complicated/noisy.
2 is pretty clean too, but >> is often used to convey something like fast-forward, skip, get ahead.

between 1 or 4, i think 1 is just a tiny bit better, because 4's arrows are a bit too thick imo.
you could also possibly try 2 but with only one >, see how that looks, but in case you don't want to, i think 1 is the winner out of these :)

app/src/main/java/org/schabi/newpipe/ui/components/menu/LongPressAction.kt
```diff
@@ -0,0 +1,320 @@
...
+                        )
+                    },
+                ) +
+                (
+                    onUnsubscribe
+                        ?.let { onUnsubscribe ->
+                            listOf(Type.Unsubscribe.buildAction { onUnsubscribe.run() })
+                        }
+                        ?: listOf()
+                    )
```
Isira-Seneviratne:
```suggestion
                listOfNotNull(onUnsubscribe?.let { Type.Unsubscribe.buildAction { it.run() } })
```

Stypox:
I mostly implemented the long-press menu actions editor, where you can enable and disable actions, and even enable and disable the header. The order is still not persisted anywhere, so next thing I'll do is store the data in settings. There are also a few TODOs left, e.g. making the UI usable with DPAD buttons (e.g. Android TV). I borrowed some code from [FlorisBoard](https://github.com/florisboard/florisboard/blob/059d2fd4bfa9f30f634552f4457c4541ba44a238/app/src/main/kotlin/dev/patrickgold/florisboard/ime/smartbar/quickaction/QuickActionsEditorPanel.kt), so thanks! :-) 

Screen_recording_20251022_011336

Stypox:
Now it works even on Android TVs when using DPAD controls. The code has grown a bit complicated though.

Screen_recording_20251023_030852

peat80:
Tried this on tv.
You can reorder and remove entrys fine with just the remote.
The only problem is that per default the reorder button has focus.
So when you long press ok button and then release it you always land in the menu automatically. 🤔😄

Also the reordered menu is not saved on exit.
On both tv and phone.
Keeps reverting back to default setup.

Great work so far. 👌🏻

TobiGr:
Hint: there is currently a PR for accessibility improvements for the long press menu targeting `dev`: #12956
We should ensure that its improvements are properly applied to the new design.

absurdlylongusername:
@TobiGr That PR is targeting refactor, not dev

Stypox:
I'm designing the icon for the "Play in the background all items shuffled" that would appear when long-pressing playlists or channels. Which one of these designs should I choose? They all vary only by the thickness of the arrows in the bottom-right icon. 

image


<sub>Note for future me: the corresponding thinkness increments are 0, 0.08, 0.12, 0.15, 0.18, 0.21, 0.25, 0.3, 0.34.</sub>

TobiGr:
I'd go with 2

ghost:
The more I look, the more versions I find "ok". But after several "fresh" looks I'd say that I like 2, 3, 4. And dislike 0, (5), 6, 7, 8

KraXen72:
i like 3 or 4

Stypox:
Ok, this PR is not fully ready for review and user feedback :-D

- The editor is now fully functional, it works with touch and DPAD, and it saves the button arrangement to settings
- I added the "Background/Popup/Play shuffled" options (I ended up choosing thickness 3), which fetches a few pages of the queue before shuffling it, to solve #875
- I wrote quite a few UI tests and normal tests (~1500 lines of code)
- I cleaned up the code and added javadocs everywhere, so hopefully it should be rather easy to review despite the size
- I updated the PR description with screenshots
- Sorry for the 87 commits :sweat_smile:, though I believe they shouldn't be squashed as they contain useful information

TobiGr:
Thank you!
First impression: that stuff is now cleaner. Perfect! I'll come with a more detailed review later this week.
Just one small thing that I noticed already is that some of the new files do not have any SPDX license identifier headers. Please add them so that we can slowly make NewPipe REUSEable (you chose the perfect video for that screenshot :D)

TobiGr:
# Testing feedback
## Bugs
The menu is not restored on screen rotation: [Screen_recording_20260213_112625.webm](https://github.com/user-attachments/assets/ab9e6f45-9055-474f-8ebe-098004d0d01f)


Stream that is already marked as watched can still be marked as wathched. You might want to disable that option if the stream is stored as watched (could ne done in a future PR).
Screenshot_NewPipe_20260213_113403

app/src/main/java/org/schabi/newpipe/ui/components/menu/LongPressMenuSettings.kt
```diff
@@ -0,0 +1,132 @@
...
+            }
+
+        // In case there is some bug in the stored data, make sure we don't return duplicate items,
+        // as that would break/crash the UI and also not make any sense.
+        val actionsDistinct = actions.distinct()
+        if (actionsDistinct.size != actions.size) {
+            Log.w(TAG, "Actions in settings were not distinct: $actions != $actionsDistinct")
+        }
+        return actionsDistinct
+    } catch (e: NoSuchElementException) {
```
theimpulson, line 100: Better variable name everywhere please


```suggestion
    } catch (exception: NoSuchElementException) {
```

app/src/main/java/org/schabi/newpipe/ui/components/menu/icons/PlayFromHere.kt
```diff
@@ -0,0 +1,53 @@
...
+            lineToRelative(6.17f, 0.0f)
+            lineToRelative(-1.58f, 1.59f)
+            lineToRelative(1.42f, 1.41f)
+            lineToRelative(3.99f, -4.0f)
+            close()
+        }
+    }
+}
+
+@Preview(showBackground = true, backgroundColor = 0xFFFFFFFF)
```
theimpulson, line 45: is specifying background color necessary?

app/src/main/java/org/schabi/newpipe/ui/components/items/stream/StreamUtils.kt
theimpulson, line 47: Please use when wherever possible instead of if-else

app/src/main/java/org/schabi/newpipe/ui/components/common/Tooltip.kt
```diff
@@ -0,0 +1,84 @@
...
+ * @param contentDescription the text to use as content description for the button,
+ * and also to show in the tooltip
+ * @param modifier as described in [SimpleTooltipBox]
+ * @param buttonModifier a modifier for the internal [IconButton]
+ * @param iconModifier a modifier for the internal [Icon]
+ * @param tint the color of the icon
+ */
+@Composable
+fun TooltipIconButton(
+    onClick: () -> Unit,
```
theimpulson, line 60: onClick can be made last parameter to allow better kotlin syntax when calling it

app/src/main/java/org/schabi/newpipe/player/playqueue/PlayQueueItem.kt
```diff
@@ -65,6 +65,15 @@ class PlayQueueItem private constructor(
                 .subscribeOn(Schedulers.io())
                 .doOnError { throwable -> error = throwable }
 
+    fun toStreamInfoItem(): StreamInfoItem {
+        val item = StreamInfoItem(serviceId, url, title, streamType)
```
theimpulson, line 69: Just a suggestion but maybe we can simply return here with an apply to set other parameters instead of creating a new variable.

app/src/androidTest/java/org/schabi/newpipe/InstrumentedTestUtil.kt
```diff
@@ -0,0 +1,143 @@
...
+    }
+    if (value !in l..r) {
+        fail("Expected $value to be in range [$l, $r]")
+    }
+}
+
+/**
+ * Asserts that [value] is NOT in the range [[l], [r]] (both extremes included).
+ */
+fun <T : Comparable<T>> assertNotInRange(l: T, r: T, value: T) {
```
theimpulson, line 103: Better variable names please, for all the function parameters as well.

app/src/androidTest/java/org/schabi/newpipe/InstrumentedTestUtil.kt
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Use this instead of calling `InstrumentationRegistry.getInstrumentation()` every time.
+ */
+val inst: Instrumentation
+    get() = InstrumentationRegistry.getInstrumentation()
+
+/**
+ * Use this instead of passing contexts around in instrumented tests.
+ */
+val ctx: Context
```
theimpulson, line 32: Please prefer `context` and `instrumentation` instead

app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
```diff
@@ -256,32 +264,71 @@ protected void initListeners() {
...
+                } catch (final Exception e) {
+                    ErrorUtil.showUiErrorSnackbar(BaseListFragment.this, "Opening channel fragment",
+                            e);
+                }
+            }
+
+            @Override
+            public void held(final ChannelInfoItem item) {
+                // Note: this does a blocking I/O call to the database. Use coroutines when
+                // migrating to Kotlin/Compose instead.
```
TobiGr, line 299: So that is highlighted for the next person taking care of it :)
```suggestion
                // TODO: this does a blocking I/O call to the database. Use coroutines when
                // migrating to Kotlin/Compose instead.
```

app/src/main/java/org/schabi/newpipe/local/bookmark/BookmarkFragment.java
```diff
@@ -493,59 +472,33 @@ public void onSwiped(@NonNull final RecyclerView.ViewHolder viewHolder,
...
-                                selectedItem.getUid(),
-                                dialogBinding.dialogEditText.getText().toString()))
-                .setNegativeButton(R.string.cancel, null)
-                .show();
+        openLongPressMenuInActivity(
+                requireActivity(),
+                LongPressable.fromPlaylistMetadataEntry(selectedItem),
+                LongPressAction.fromPlaylistMetadataEntry(
+                        selectedItem,
+                        // Note: this does a blocking I/O call to the database. Use coroutines when
```
TobiGr, line 494:
```suggestion
                        // TODO: this does a blocking I/O call to the database. Use coroutines when
```

app/src/main/java/org/schabi/newpipe/player/playqueue/ChannelTabPlayQueue.java
```diff
@@ -15,22 +19,39 @@
...
                                final int index) {
-        super(serviceId, linkHandler.getUrl(), nextPage, streams, index);
-        this.linkHandler = linkHandler;
+        super(serviceId, tabHandler.getUrl(), nextPage, streams, index);
+        this.tabHandler = tabHandler;
     }
 
     public ChannelTabPlayQueue(final int serviceId,
                                final ListLinkHandler linkHandler) {
         this(serviceId, linkHandler, null, Collections.emptyList(), 0);
```
TobiGr, line 40: I'd also rename the ListLinkHandler to `tabHandler` to be consistent.

PR #12067, opened 2025-02-27
Fix stream notification grouping

Isira-Seneviratne:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

- It is possible for the stream uploader URL to be different from the channel URL, thus resulting in the notifications not being grouped by the channel. This change fixes that.

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/feed/notifications/NotificationHelper.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/feed/notifications/NotificationHelper.kt
@@ -88,7 +88,7 @@ class NotificationHelper(val context: Context) {
 
                 // Show individual stream notifications, set channel icon only if there is actually
                 // one
-                showStreamNotifications(newStreams, data.serviceId, bitmap)
+                showStreamNotifications(newStreams, data.serviceId, data.url, bitmap)
                 // Show summary notification
                 manager.notify(data.pseudoId, summaryBuilder.build())
 
@@ -97,7 +97,7 @@ class NotificationHelper(val context: Context) {
 
             override fun onBitmapFailed(e: Exception, errorDrawable: Drawable) {
                 // Show individual stream notifications
-                showStreamNotifications(newStreams, data.serviceId, null)
+                showStreamNotifications(newStreams, data.serviceId, data.url, null)
                 // Show summary notification
                 manager.notify(data.pseudoId, summaryBuilder.build())
                 iconLoadingTargets.remove(this) // allow it to be garbage-collected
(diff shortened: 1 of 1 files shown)


Stypox (commented):
Thanks! Seems fine, though I didn't test. Can you provide a database I could import where this happens? Also, can you point to any open issue related to this?

Isira-Seneviratne:
Here's an example database: [newpipe_subscriptions_202502280744.json](https://github.com/user-attachments/files/19019436/newpipe_subscriptions_202502280744.json)

There isn't an open issue related to this I could find, it's just an issue I came across while running the app. I dug deeper and found that the stream uploader URL is different from the channel URL, so the notifications aren't properly grouped.

Stypox:
Mmmh, I still get a lot of distinct notifications, and there does not seem to be a difference between what I get with this PR and what I get before this PR.

<table>
  <tr>
    <th>Initial (**all** notifications are from TTaTST)</th>
    <th>After unsubscribing from TTaTST</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

I tested this code by deleting all feed data manually by putting this code somewhere:
```java
        new Thread(
                () -> {
                    NewPipeDatabase.getInstance(requireContext()).feedDAO().deleteAll();
                    NewPipeDatabase.getInstance(requireContext()).feedGroupDAO().deleteAll();
                    NewPipeDatabase.getInstance(requireContext()).streamDAO().deleteAll();
                }
        ).start();
```

Stypox (approved):
I tried logging all of the channel URLs being used for `setGroup` and they seem correct so it's just Android deciding not to group them for some reason, sorry for the wrong review. I don't understand how Android groups notifications, because I get a mess of notifications grouped badly, but oh well...

Isira-Seneviratne:
> I tried logging all of the channel URLs being used for `setGroup` and they seem correct so it's just Android deciding not to group them for some reason, sorry for the wrong review. I don't understand how Android groups notifications, because I get a mess of notifications grouped badly, but oh well...

https://developer.android.com/develop/ui/views/notifications/group

Stypox:
> Automatic grouping behavior might vary on some device types. To provide the best experience on all devices and versions, if you know notifications must be grouped, specify a group key and group summary to make sure they are grouped.

So maybe we need to specify one notification to be the group summary for each channel?

Isira-Seneviratne:
> So maybe we need to specify one notification to be the group summary for each channel?

That's already being done in line 60, the issue was that the group key was different for the summary notification and the other notifications.

Stypox:
I still wonder why they were not grouping well when I tested it (there were like 100s of ungrouped notifications from the same channel), but the code looks indeed correct

PR #12093, opened 2025-03-12
[Android 13+] Support per-app language preferences

mileskrell:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Enables [automatic per-app language support](https://developer.android.com/guide/topics/resources/app-languages#auto-localeconfig).

On Android 13+, the preference in Settings -> Content shows any locale that has been set by the user through Android's per-app language settings. When tapped, it opens NewPipe's per-app language settings screen.

Also added a method to migrate any existing app language preference from `SharedPreferences` to the Android per-app language settings when we run the app on Android 13+.

- Before:

- After:

https://github.com/user-attachments/assets/5694a885-9b98-48f0-a243-44560537424b

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/Localization.java
+++ b/app/src/main/java/org/schabi/newpipe/util/Localization.java
@@ -12,6 +12,7 @@ import android.os.Build;
 import android.text.TextUtils;
 import android.text.format.DateUtils;
 import android.util.DisplayMetrics;
+import android.util.Log;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -19,6 +20,7 @@ import androidx.annotation.PluralsRes;
 import androidx.annotation.StringRes;
 import androidx.appcompat.app.AppCompatDelegate;
 import androidx.core.math.MathUtils;
+import androidx.core.os.LocaleListCompat;
 import androidx.preference.PreferenceManager;
 
 import org.ocpsoft.prettytime.PrettyTime;
@@ -65,6 +67,7 @@ import java.util.stream.Collectors;
  */
 
 public final class Localization {
+    private static final String TAG = Localization.class.toString();
     public static final String DOT_SEPARATOR = " • ";
     private static PrettyTime prettyTime;
 
@@ -432,5 +435,33 @@ public final class Localization {
         // or some language have some specific rule... then we have to change it)
         final int safeCount = (int) MathUtils.clamp(count, Integer.MIN_VALUE, Integer.MAX_VALUE);
         return context.getResources().getQuantityString(pluralId, safeCount, formattedCount);
+    }
+
+    public static void migrateAppLanguageSettingIfNecessary(@NonNull final Context context) {
+        // Starting with pull request #12093, NewPipe on Android 13+ exclusively uses Android's
+        // public per-app language APIs to read and set the UI language for NewPipe.
+        // If running on Android 13+, the following code will migrate any existing custom
+        // app language in SharedPreferences to use the public per-app language APIs instead.
+        if (Build.VERSION.SDK_INT >= 33) {
+            final SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(context);
+            final String appLanguageKey = context.getString(R.string.app_language_key);
+            final String appLanguageValue = sp.getString(appLanguageKey, null);
+            if (appLanguageValue != null) {
+                sp.edit().remove(appLanguageKey).apply();
+                final String appLanguageDefaultValue =
+                        context.getString(R.string.default_localization_key);
+                if (!appLanguageValue.equals(appLanguageDefaultValue)) {
+                    try {
+                        AppCompatDelegate.setApplicationLocales(
+                                LocaleListCompat.forLanguageTags(appLanguageValue)
+                        );
+                    } catch (final RuntimeException e) {
+                        Log.e(TAG, "Failed to migrate previous custom app language "
+                                + "setting to public per-app language APIs"
+                        );
+                    }
+                }
+            }
+        }
     }
 }
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -190,6 +190,8 @@ public class MainActivity extends AppCompatActivity {
                 && ReleaseVersionUtil.INSTANCE.isReleaseApk()) {
             UpdateSettingsFragment.askForConsentToUpdateChecks(this);
         }
+
+        Localization.migrateAppLanguageSettingIfNecessary(getApplicationContext());
     }
 
     @Override
--- a/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/NewPipeSettings.java
@@ -6,12 +6,9 @@ import android.content.Context;
 import android.content.SharedPreferences;
 import android.os.Build;
 import android.os.Environment;
-import android.util.Log;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.StringRes;
-import androidx.appcompat.app.AppCompatDelegate;
-import androidx.core.os.LocaleListCompat;
 import androidx.preference.PreferenceManager;
 
 import org.schabi.newpipe.App;
@@ -45,7 +42,6 @@ import java.util.Set;
  * Helper class for global settings.
  */
 public final class NewPipeSettings {
-    private static final String TAG = NewPipeSettings.class.toString();
     private NewPipeSettings() { }
 
     public static void initSettings(final Context context) {
@@ -68,7 +64,6 @@ public final class NewPipeSettings {
         saveDefaultAudioDownloadDirectory(context);
 
         disableMediaTunnelingIfNecessary(context);
-        migrateAppLanguageSettingIfNecessary(context);
     }
 
     static void saveDefaultVideoDownloadDirectory(final Context context) {
@@ -187,34 +182,6 @@ public final class NewPipeSettings {
             prefs.edit()
                     .putInt(context.getString(R.string.media_tunneling_device_blacklist_version),
                             DeviceUtils.MEDIA_TUNNELING_DEVICE_BLACKLIST_VERSION).apply();
-        }
-    }
-
-    private static void migrateAppLanguageSettingIfNecessary(@NonNull final Context context) {
-        // Starting with pull request #12093, NewPipe on Android 13+ exclusively uses Android's
-        // public per-app language APIs to read and set the UI language for NewPipe.
-        // If running on Android 13+, the following migration will move any existing custom
-        // app language in SharedPreferences to use the public per-app language APIs instead.
-        if (Build.VERSION.SDK_INT >= 33) {
-            final SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(context);
-            final String appLanguageKey = context.getString(R.string.app_language_key);
-            final String appLanguageValue = sp.getString(appLanguageKey, null);
-            if (appLanguageValue != null) {
-                sp.edit().remove(appLanguageKey).apply();
-                final String appLanguageDefaultValue =
-                        context.getString(R.string.default_localization_key);
-                if (!appLanguageValue.equals(appLanguageDefaultValue)) {
-                    try {
-                        AppCompatDelegate.setApplicationLocales(
-                                LocaleListCompat.forLanguageTags(appLanguageValue)
-                        );
-                    } catch (final RuntimeException e) {
-                        Log.e(TAG, "Failed to migrate previous custom app language "
-                                + "setting to public per-app language APIs"
-                        );
-                    }
-                }
-            }
         }
     }
 }


mileskrell:
thanks @TobiGr! I looked briefly for an existing in-app language picker but didn't check in the "content" settings.

Yeah, I'll need to make some more changes so they don't conflict. I should be able to keep the in-app language picker and [migrate it](https://developer.android.com/guide/topics/resources/app-languages#androidx-impl) to use the new APIs.

Stypox:
If it's just a matter of hiding the settings screen language toggle and showing a message instead, in case the language has been changed from outside, then it's ok on the `dev` branch. If it involves heavy changes to the way the app sets the language then it's better on the `refactor` branch. Thanks @mileskrell !

app/src/main/res/xml/content_settings.xml
```diff
@@ -13,6 +13,13 @@
         app:iconSpaceReserved="false"
         app:useSimpleSummaryProvider="true" />
 
+    <Preference
+        android:key="@string/app_language_android_13_and_up_key"
+        android:title="@string/app_language_title"
+        app:isPreferenceVisible="false"
+        app:singleLineTitle="false"
+        app:iconSpaceReserved="false" />
```
mileskrell, line 21: I added a separate `Preference` for the Android 13+ behavior only because for some reason when I set an `OnPreferenceClickListener` on the existing preference that returns `true` (indicating "the click was handled"), it still opens the existing dialog.

mileskrell:
@TobiGr @Stypox I've updated this PR so it doesn't conflict with the existing in-app language picker. Let me know what you think!

mileskrell:
My bad, totally forgot to fix the toast!

So prior to this PR, we've been showing this toast whenever the app language, content language, or content country is changed. Changing the app language on Android <13 definitely requires a restart, but are we sure that changes to content language/country also require a restart? When building from the latest commit on `dev`, after I set the content language, I'm shown the toast, but then if immediately I open a video with an audio track for that language the new language is still used without having to restart:

https://github.com/user-attachments/assets/f0b98a87-d20c-4b10-9135-857556c15c70

If changes to the content language and country are reflected without having to restart, then the only situation where we need to show the toast is when app language is changed on a device running Android <13. Let me know if that sounds correct!

app/src/main/java/org/schabi/newpipe/App.java
```diff
@@ -122,6 +125,29 @@ public void onCreate() {
...
+            if (!appLanguageCurrentValue.equals(appLanguageDefaultValue)) {
+                try {
+                    AppCompatDelegate.setApplicationLocales(
+                            LocaleListCompat.forLanguageTags(appLanguageCurrentValue)
+                    );
+                } catch (final RuntimeException e) {
+                    Log.e(TAG, "Error migrating to Android 13+ per-app language settings");
+                }
+            }
+        }
```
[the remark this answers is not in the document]
mileskrell: My only concern is that if we convert this to a `Migration`, my understanding is that it will only run once. So in theory, the following scenario is possible:
1. Someone running NewPipe on Android <13 sets a custom app language
2. They update to the new version of NewPipe, where this migration is run but doesn't do anything because they're on Android <13
3. They update to Android 13+
4. Their custom app language pref isn't used (because it's running the Android 13+ code path) and it's not migrated from their old preference (because there's nothing triggering the migration to run again)

I recognize this will probably be a pretty rare scenario, just wanted to flag it. But whether or not we make it a `Migration`, I do recognize that this code belongs _somewhere_ outside of the `Application` subclass for sure.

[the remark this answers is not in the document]
mileskrell: I've moved it to a `Migration`. If we want to prevent the hypothetical app language setting loss I described above, maybe we could add a new migrations file for "migrations that should always be run on app startup because they depend on the current Android version".

mileskrell:
> If changes to the content language and country are reflected without having to restart, then the only situation where we need to show the toast is when app language is changed on a device running Android <13. Let me know if that sounds correct!

The most recent commit I pushed assumes this is the case, but we can revert it if not.

app/src/main/java/org/schabi/newpipe/App.java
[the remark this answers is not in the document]
Stypox: Mmmh you are right, this shouldn't be a migration as it needs to run every time the app notices it has been updated to Android 13+ from a lower Android version. What do you think @TobiGr?

app/build.gradle
```diff
@@ -97,6 +97,10 @@ android {
         androidTest.assets.srcDirs += files("$projectDir/schemas".toString())
     }
 
+    androidResources {
+        generateLocaleConfig = true
+    }
+
```
Stypox, line 103: If we don't pay attention to new languages being added from Weblate, this might lead to half-translated being proposed to the user. Currently the list of languages is manually kept in `settings_keys.xml`, so when new languages are added through Weblate they don't automatically appear in the app menu. However, I'm fine with switching to the new behavior, as it would avoid work to keep the language list up to date. Is it possible to access the locale config from within the app, so we can delete the list of languages in `settings_keys.xml` and use the autogenerated locale config even for the pre-Android-13 language picker?

Stypox (changes requested):
Code looks good to me, thank you! I have a comment here that could possibly be solved in a separate PR if you prefer. Also, see my comment about the migration.

app/src/main/java/org/schabi/newpipe/App.java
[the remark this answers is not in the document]
Stypox: Yes

app/build.gradle
mileskrell: I found [LocaleManager#getOverrideLocaleConfig](https://developer.android.com/reference/android/app/LocaleManager#getOverrideLocaleConfig()) but it sounds like that method is only to get a `LocaleConfig` that's already been set with `setOverrideLocaleConfig()`. As far as I can tell, we would have to manually generate a `LocaleConfig` instead of using the auto-generated one if we wanted to use it for this.

mileskrell:
> Also, could you take a video of the updated code on pre-Android-13, so we can see if it still works as expected there?

@Stypox I tried it on an emulator running Android 12 and it seems fine:

https://github.com/user-attachments/assets/37511e41-27d3-46b2-94b1-23ee2c9b3b55

app/build.gradle
Stypox: I could find the data under `app/build/generated/res/localeConfig/debug/xml/_generated_res_locale_config.xml` after running the build, with this data inside:
```xml
<locale-config xmlns:android="http://schemas.android.com/apk/res/android">
    <locale android:name="en-US"/>
    <locale android:name="ace"/>
    ...
```
So in theory the data is there, but in practice I think we better not access it by manually opening the xml since it's not officially supported. So yeah let's keep the previous behavior.

Stypox:
Thank you! I tested to import an export from a previous version of NewPipe into this PR's build on Android 13 and unfortunately it did not work: the app language remained the one that was there previously before importing settings. This is the database I imported (it should set the language to italian): [NewPipeData-20250324_172806.zip](https://github.com/user-attachments/files/19433723/NewPipeData-20250324_172806.zip)

mileskrell:
thanks for catching that and for sharing the settings export @Stypox!

It turns out `AppCompatDelegate.setApplicationLocales` wasn't working - I missed the part of the documentation that says

> Note: This API should always be called after Activity.onCreate(), apart from any exceptions explicitly mentioned in this documentation.

I've moved the migration method to `Localization.migrateAppLanguageSettingIfNecessary` and I'm calling it at the end of `MainActivity.onCreate`. Here's a video of the working migration:

https://github.com/user-attachments/assets/d3b3d178-c553-4892-9deb-6fae7f4b6bb3

Stypox (approved):
I tested again and it all seems to work as expected. Thank you for making the requested changes!

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -427,4 +436,32 @@ private static String getQuantity(@NonNull final Context context,
...
+                        context.getString(R.string.default_localization_key);
+                if (!appLanguageValue.equals(appLanguageDefaultValue)) {
+                    try {
+                        AppCompatDelegate.setApplicationLocales(
+                                LocaleListCompat.forLanguageTags(appLanguageValue)
+                        );
+                    } catch (final RuntimeException e) {
+                        Log.e(TAG, "Failed to migrate previous custom app language "
+                                + "setting to public per-app language APIs"
+                        );
```
AudricV, line 461: Shouldn't the error be logged?

app/src/main/java/org/schabi/newpipe/settings/ContentSettingsFragment.java
```diff
@@ -31,12 +35,28 @@ public void onCreatePreferences(final Bundle savedInstanceState, final String ro
...
+                final Locale customLocale = AppCompatDelegate.getApplicationLocales().get(0);
+                if (customLocale != null) {
+                    return customLocale.getDisplayName();
+                }
+                return getString(R.string.systems_language);
+            });
+            newAppLanguagePref.setOnPreferenceClickListener(preference -> {
+                final Intent intent = new Intent(Settings.ACTION_APP_LOCALE_SETTINGS)
+                        .setData(Uri.fromParts("package", requireContext().getPackageName(), null));
+                startActivity(intent);
```
AudricV, line 54: That's a very edge case, but if the user has disabled settings or if some ROMs are broken (hello weird Android TV devices), this would crash the app due to no app being able to handle the intent.

TobiGr:
This PR introduced a regression: locals which are not available in Android are not supported by the per-app language preference. E.g. #12476 adds Latin / Latina as language. It's a dead language and thus not available in Android and cannot be selected in the per-app language preferences. However, it can be selected by users with Android 12 and earlier.

This might be related to the fact that not all locals have an ISO 639-1 code. Possible workaround: Set `generateLocaleConfig = false` and provide our own [locale config](https://developer.android.com/guide/topics/resources/app-languages#use-localeconfig)

Stypox:
So I did a quick comparison of the languages available in the per-app languages vs those available in the in-app chooser. The below lists are not comprehensive of non-latin scripts (there are only a few such languages) because I can't neither write nor compare non-latin scripts :sweat_smile:. In any case, I think it's pretty clear Android is filtering out a whole bunch of languages. I would say we revert this PR.

<details>
<summary>Our in-app picker additionally has</summary>


- Bahasa Indonesia
- Bahasa Melayu
- Basa Acèh
- Esperanto
- 한국어 (korean)
- Kurmancî
- Latina
- Nनेपाली (nepalese)
- Occitan
- Toki Pona
- Босански (bosnian)
- български език (bulgarian)
</details>

I did try setting `generateLocaleConfig = false` and then explicitly listing locales in `locales_config.xml`, and despite it listing English, Italian and Latin, only English and Italian would be shown in the system per-app language chooser. So I think there is nothing we can do to make Android show the languages we want in the per-app chooser, it seems like it only supports the most popular languages. @Isira-Seneviratne do you have any insight on this?

<details>
<summary>locales_config.xml</summary>


```xml
<locale-config xmlns:android="http://schemas.android.com/apk/res/android">
    <locale android:name="en-US"/>
    <locale android:name="it-IT"/>
    <locale android:name="la-Latn"/>
</locale-config>
```

</details>

I took a look at the current automatically generated `locales_config` and it indeed does list all of the languages we support. Actually there are even some we don't support but that have a `strings.xml` with only a couple strings, and this is kind of a problem as we don't want to make them choosable since they're not really translated:


<details>
<summary>Android per-app picker additionally has</summary>


- Afrikaans
- Bamanakan (Mali)
- Bosanski (latinica)
- Cymraeg
- Indonesia
- Kiswahili
- Melayu
- қазақ тілі (kazak)
</details>

litetex:
@Stypox 

I think some parts of the list are incorrect. [Here](https://android.googlesource.com/platform/frameworks/base/+/android16-release/core/res/res/values/locale_config.xml) is the full list of available languages on Android 16.

There is e.g. korean, bulgarian, indonesian and nepalese (random sampling) inside Android's list and I could also locate these languages on my phone (Android 15; CalyxOs on FP5) for NewPipe (nightly) and apply them successfully. Maybe they have slightly different names?

However I couldn't find latin - most likely because it's dead for ~1200 years and not supported (at this point this stuff is just like technical debt).

Anyway I don't think that we should revert this because of 1 dead language, that is used by maybe 2-5? users.

Stypox:
> There is e.g. korean, bulgarian, indonesian and nepalese
> random sampling

Seems like you randomly sampled exactly those that are indeed in a different place. But these are definitely unavailable (I searched for them): Esperanto, Acehnese, Occitan, Toki Pona.

> at this point this stuff is just like technical debt

Yeah but per-app languages are already a more complex system than the simple setting we used to have, and if it's not even on par with previous behavior for the stupid reason it lacks lesser-known languages then it's not so worth it... Why does Android just not include all available languages?!? If they don't have a name for "Latin" they could just show the locale name, "la", and call it a day, instead of hiding the locale completely...

Stypox:
Oh wait:
- Esperanto and Toki Pona are constructed languages (though Esperanto is widely known so it's strange it's not included in Android)
- Occitan is basically a dialect
- Acehnese should exist though, it's spoken by 3 million people: https://en.wikipedia.org/wiki/Acehnese_language

ShareASmile:
> Esperanto and Toki Pona are constructed languages (though Esperanto is widely known so it's strange it's not included in Android)
> Occitan is basically a dialect
> Acehnese should exist though, it's spoken by 3 million people

Yes, definitely more than one reason to revert this PR.

litetex:
> at this point this stuff is just like technical debt

I was sarcastic about the latin language being present in Java and not our code ;)

---

So I checked again on my phone because the emulator is unreliable:

<details><summary>Esperanto exists ✔️ </summary>

grafik

</details>

Toki Pona 
* has no ISO 639-1 entry (parent language)
  * non ISO 639-1 languages are [ignored by Android](https://developer.android.com/guide/topics/resources/app-languages#locale-names)
* not present in Android
* has 3k native speakers, how the heck did we get this language in the first place? There is likely 0 people using the app with this language.

<details><summary>Occitan exists ✔️ </summary>

grafik

</details>

Acehnese 
* has no ISO 639-1 entry (parent language) 
* not present in Android
* all people that speak it also speak Indonesian because that's the official language of [Aceh](https://en.wikipedia.org/wiki/Aceh)

> it lacks lesser-known languages

I also crunched some numbers:

* We now have 3 languages that are missing: Acehnese, Latin and Toki Pona - out of 641 that Android supports.
* These languages have **approved translations (that are in our translation files) of 1% each** [on Weblate](https://hosted.weblate.org/projects/newpipe/) and total translations of 2%, 8% and 10%.
* They were last update: 4 years ago, 4 years ago (beside a single line 5 months ago) and 2 years ago.

So I think we should draw a line here and simply ignore, block and remove these languages:
* They aren't translated properly in the first place
* People that use these languages can to 100% also speak another language? Why? Because they couldn't use Android otherwise.
* They just bloat the app size and maintenance effort
* They block the adoption of a new Android feature (new Android feature > languages used by like 3 users) and cause the revert.

PS: All this translation stuff wouldn't be required if the [Tower of Babel](https://en.wikipedia.org/wiki/Tower_of_Babel) was still standing :P

Isira-Seneviratne:
> @Isira-Seneviratne do you have any insight on this?

Unfortunately, no. If setting the locales manually doesn't work either, then it would indeed be best to undo the change and use the in-app picker.

Stypox:
Thank you for your analysis! Strangely enough neither Esperanto nor Occitan exist on the emulator I was using and on my FP3 with /e/OS. Maybe it's an OS issue?

Supporting languages with only 1% of translated strings is indeed not very useful, and all other languages seem to be present, so we should be fine. I will close https://github.com/TeamNewPipe/NewPipe/pull/12478 but still add a warning in the release notes so users can report any issue they have.

Stypox:
I just tested on an emulator with Google Play Services and Esperanto and Occitan are available there. Neither of them are available on my phone (with MicroG) nor on emulators without Google Play Services. So my guess is that Google Play Services somehow adds support for more languages (?).

enolp:
Be careful with locale support on Android,. As I already said in [another issue](https://github.com/TeamNewPipe/NewPipe/issues/12503#issuecomment-3186370404), some manifacturers or ROMs can expand or reduce locale compatibility. On the other hand, most OEM ROMs are only compatible with the locales they are translated to, but using apps like "MoreLocale2" can show magically those hidden locales in different ways.

EDITED 2025-08-14:

> Occitan is basically a dialect

Talking from the linguistic scope, some of you don't know the difference between a language, a variant and a dialect. Spoiler, they are all the same. 
Distinctions made between language and dialect come from a very old political prospective. I mean, every contry which has a colonial/nationalist past have a record of "linguistic cleanse" and one of the ways they had to impose their language was degrading linguistic classification of the languages they "conquered" or wanted to erradicate.

> They aren't translated properly in the first place

From my point of view, I would get mad if I see my contributions getting removed because NewPipe is unable to find people who can continue my work. Someone spends time for free translating NewPipe for getting their tranlsations removed? No, thanks!

> People that use these languages can to 100% also speak another language? Why? Because they couldn't use Android otherwise.

The right question is: What is the NewPipe we want? A NewPipe for the 4 most spoken worldwide languages or an accessible NewPipe for all languages, big, endangered or minorized? My choice is clear, I bet for an inclusive and truly open NewPipe which gives visibility to the real linguistic diversity. You cannot force a user to speak or select the languages you want, that would break the spirit of any free (as freedom) license. The user needs free of choice and right to chose, even more with endangered languages.

> all people that speak it also speak Indonesian because that's the official language of [Aceh](https://en.wikipedia.org/wiki/Aceh)

"Why would you speak your ancient and ugly language instead of speaking a GrEaT, uNiVeRsAl aNd MoDeRn language like (insert language)??"
"You are not allowed to use/speak your language because is unofficial here. Speak the only official language!!!!11!!1!"
"Now we conquered you, we refuse any comumication made in your language. Speak our language now"

> They just bloat the app size and maintenance effort

Let me tell you NewPipe suffers from worse issues other than localization...

> They block the adoption of a new Android feature (new Android feature > languages used by like 3 users) and cause the revert.

Why not to make an hybrid system? NewPipe could honor system locale while having a proper language selector for all "unsupported" Android locales.

PR #12173, opened 2025-04-11
Use native Bundle.toString() implementation

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Use the native `Bundle.toString()` implementation.

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/ktx/Bundle.kt
+++ b/app/src/main/java/org/schabi/newpipe/ktx/Bundle.kt
@@ -7,16 +7,3 @@ import java.io.Serializable
 inline fun <reified T : Serializable> Bundle.serializable(key: String?): T? {
     return BundleCompat.getSerializable(this, key, T::class.java)
 }
-
-fun Bundle?.toDebugString(): String {
-    if (this == null) {
-        return "null"
-    }
-    val string = StringBuilder("Bundle{")
-    for (key in this.keySet()) {
-        @Suppress("DEPRECATION") // we want this[key] to return items of any type
-        string.append(" ").append(key).append(" => ").append(this[key]).append(";")
-    }
-    string.append(" }")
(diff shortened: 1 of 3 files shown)


Stypox:
iirc I created that function because the normal .toString() printed just Bundle{hash} or didn't include all items. Or maybe it did so only on older Android versions. But I remember definitely considering the native .toString() and not being able to get enough information. So I'd leave it as before, but please try out the two methods on both an old (e.g. API22) and a new android version and post the 4 outputs here, then we can decide. If we decide to keep the previous bevavior, a comment should be added (I should have done so when I implemented it initally, sorry about it).

Isira-Seneviratne:
The implementation has been around since the very first Android version: https://developer.android.com/reference/android/os/Bundle#toString()

Isira-Seneviratne:
I made some changes to unparcel the extras so that their contents are actually displayed.

The format is as follows:

`Bundle[{should_start_foreground_extra=true, play_queue_key=c56f58e4-2606-495e-a59c-f2d7fb3ce910, resume_playback=true, play_when_ready=true, player_type=0}]`

Stypox:
I don't see any advantage of the new implementation with respect to the old one. The new one even uses more lines of code where you just want to print logs. It's just a debugging method anyway, so I'd keep the previous.

Isira-Seneviratne:
> I don't see any advantage of the new implementation with respect to the old one. The new one even uses more lines of code where you just want to print logs. It's just a debugging method anyway, so I'd keep the previous.

Alternatively, the extension could be updated to take an `Intent` type, since that's what's being used anyway.

Profpatsch:
You’re gonna have to decide among yourselves which one to keep :)

If there’s no resolution I’d say let’s opt for keeping the current way.

PR #12356, opened 2025-06-15
check if the JSON MimeType is supported

### LINKED ISSUE #9169: Import subscription no prompt for storage permission bug
### Checklist

### Affected version

0.24.0

1. Export subscription into a file by clicking export to>file under subscription tab
2. Clear data 
3. Try to import the subscription by clicking import from>exported file 
4. The file picker won't even let you pick the exported file

### Expected behavior

The app should import subscription

### Actual behavior

Won't even let me pick file

The issue only arises because new pipe doesn't ask for storage access permission when trying to import subscription by clicking on import from> previous export under subscription tab.

_No response_

### Logs

_No response_

Color OS v6.0(android 9)

### Affected device model

Oppo Realme 1 CPH1861

### Additional information

Note: I figured out the reason why it's because new pipe doesn't ask for permission for storage at startup so when the import is clicked it launches android file picker but if you go to settings and try to set download location it asks for permission or just manually enable storage access in settings or app info depending upon your OS and ROM then when you try to import subscription it uses newpipe's native file picker which works fine entirely.

dev-victoria:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

If no JSON MIME type is available on the android(old version), use "application/octet-stream" instead, which targets json files one puts on the old device.

Another solution could be to check for `Build.VERSION.SDK_INT < Build.VERSION_CODES.Q` instead of using the webkit MimeTypes

Since a file of type application/json can still be created locally(by exporting as such), the mime type for the file picker might need to be \*/\* to support both locally exported(as "application/json") and otherwise transferred .json files("application/octet-stream").

<table>
  <tr>
    <th>Before</th>
    <th>API28</th>
    <th>API29</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>API28</td>
    <td>API29</td>
  </tr>
</table>

- Fixes #9169

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.kt
@@ -461,14 +461,7 @@ class SubscriptionFragment : BaseStateFragment<SubscriptionState>() {
     }
 
     companion object {
-        val JSON_MIME_TYPE = getMimeType()
-
-        private fun getMimeType(): String {
-            val mimeTypeJson = MimeTypeMap.getSingleton().getMimeTypeFromExtension("json")
-            return if (mimeTypeJson.isNullOrBlank())
-                "application/octet-stream"
-            else
-                mimeTypeJson
-        }
+        val JSON_MIME_TYPE = MimeTypeMap.getSingleton()
+            .getMimeTypeFromExtension("json") ?: "application/octet-stream"
     }
 }
--- a/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.kt
+++ b/app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.kt
@@ -13,6 +13,7 @@ import android.view.MenuItem
 import android.view.SubMenu
 import android.view.View
 import android.view.ViewGroup
+import android.webkit.MimeTypeMap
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/local/subscription/SubscriptionFragment.kt
```diff
@@ -460,6 +461,14 @@ class SubscriptionFragment : BaseStateFragment<SubscriptionState>() {
...
-        const val JSON_MIME_TYPE = "application/json"
+        val JSON_MIME_TYPE = getMimeType()
+
+        private fun getMimeType(): String {
+            val mimeTypeJson = MimeTypeMap.getSingleton().getMimeTypeFromExtension("json")
+            return if (mimeTypeJson.isNullOrBlank())
+                "application/octet-stream"
+            else
+                mimeTypeJson
+        }
```
Isira-Seneviratne:
```suggestion
        val JSON_MIME_TYPE = MimeTypeMap.getSingleton().getMimeTypeFromExtension("json") ?: "application/octet-stream"
```

dev-victoria: Did it in two lines to hit the style checker.

Isira-Seneviratne (approved):
Looks good to me.

PR #12385, opened 2025-06-26
Refactor Settings Compose Setup

snaik20:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

This PR refines the Settings screen setup using Jetpack Compose:

* **Navigation Routing**: Updated navigation logic to use **Serialized Routes**, as recommended in the [official Compose navigation documentation](https://developer.android.com/develop/ui/compose/navigation).
* **Screen Title Management**: Migrated screen title handling to a Compose-native approach instead of using `addOnDestinationChangedListener`.
* **Toolbar Enhancements**:
  * Refined the toolbar layout and styling to improve alignment with Material theming.
  * Implemented proper ripple effects and consistent color usage.
  * Fixed back navigation behavior to align with Compose conventions.
* **Theming Note**: The current Compose-based theme system does not yet fully replicate the color palette used in the legacy (pre-Compose) implementation, especially across different services. A `TODO` has been added in code, and a separate task will be created to address this comprehensively.

- Before:
- After:

- Fixes #

- N/A

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/ui/Toolbar.kt
+++ b/app/src/main/java/org/schabi/newpipe/ui/Toolbar.kt
@@ -15,6 +15,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.SearchBar
 import androidx.compose.material3.Text
 import androidx.compose.material3.TopAppBar
+import androidx.compose.material3.TopAppBarDefaults
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
@@ -35,11 +36,13 @@ fun TextAction(text: String, modifier: Modifier = Modifier) {
 }
 
 @Composable
-fun NavigationIcon() {
-    Icon(
-        imageVector = Icons.AutoMirrored.Filled.ArrowBack, contentDescription = "Back",
-        modifier = Modifier.padding(horizontal = SizeTokens.SpacingExtraSmall)
-    )
+fun NavigationIcon(navigateBack: () -> Unit) {
+    IconButton(onClick = navigateBack) {
+        Icon(
+            imageVector = Icons.AutoMirrored.Filled.ArrowBack, contentDescription = "Back",
+            modifier = Modifier.padding(horizontal = SizeTokens.SpacingExtraSmall)
+        )
+    }
 }
 
 @Composable
@@ -53,7 +56,8 @@ fun SearchSuggestionItem(text: String) {
 fun Toolbar(
     title: String,
     modifier: Modifier = Modifier,
-    hasNavigationIcon: Boolean = true,
+    // TODO: Handle search in toolbar in a better way.
+    onNavigateBack: (() -> Unit)? = null,
     hasSearch: Boolean = false,
     onSearchQueryChange: ((String) -> List<String>)? = null,
     actions: @Composable RowScope.() -> Unit = {}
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/schabi/newpipe/ui/Toolbar.kt
```diff
@@ -80,33 +104,62 @@ fun Toolbar(
...
+                                .padding(SpacingExtraSmall),
+                            contentAlignment = Alignment.Center,
                         ) {
                             Column {
                                 Text(text = "╰(°●°╰)")
                                 Text(text = stringResource(id = R.string.search_no_results))
                             }
                         }
+                    } else {
+                        Column(Modifier.verticalScroll(rememberScrollState())) {
```
Isira-Seneviratne: A `LazyColumn` would be better here, as it will avoid composing non-visible items.

app/src/main/java/org/schabi/newpipe/ui/Toolbar.kt
```diff
@@ -80,33 +105,64 @@ fun Toolbar(
...
+                                        modifier = Modifier
+                                            .clickable {
+                                                textFieldState.edit { replace(0, length, result) }
+                                                expanded = false
+                                            }
+                                            .fillMaxWidth()
+                                    )
+                                }
+                            }
+                        }
```
Isira-Seneviratne:
```suggestion
                        LazyColumn {
                            items(searchResults) { result ->
                                    ListItem(
                                        headlineContent = {
                                            SearchSuggestionItem(result)
                                        },
                                        modifier = Modifier
                                            .clickable {
                                                textFieldState.edit { replace(0, length, result) }
                                                expanded = false
                                            }
                                            .fillMaxWidth()
                                    )
                            }
                        }
```

Isira-Seneviratne (changes requested):

app/src/main/java/org/schabi/newpipe/ui/Toolbar.kt
```diff
@@ -53,60 +68,95 @@ fun SearchSuggestionItem(text: String) {
...
 
     Column {
         TopAppBar(
             title = { Text(text = title) },
             modifier = modifier,
-            navigationIcon = { if (hasNavigationIcon) NavigationIcon() },
+            colors = TopAppBarDefaults.topAppBarColors(
+                containerColor = MaterialTheme.colorScheme.primary,
+                titleContentColor = MaterialTheme.colorScheme.onPrimary,
+            ),
```
Isira-Seneviratne: Better remove this, since I have raised a PR to migrate to the Material 3 color scheme: https://github.com/TeamNewPipe/NewPipe/pull/12654

HatakeKakashri: Makes sense. Will update this PR after your Material 3 PR is merged.

theimpulson:
It would be nice to switch to navigation3 to avoid more migration work whenever that hits stable. I will do that afterwards.

HatakeKakashri:
> It would be nice to switch to navigation3 to avoid more migration work whenever that hits stable. I will do that afterwards.

That would be nice. I can take a look at Navigation 3. Quick question: are we looking for a specific feature in Navigation 3, or are we considering it mainly because it is newer and designed for Jetpack Compose? I’m a bit skeptical as the [official documentation](https://developer.android.com/guide/navigation/navigation-3#:~:text=The%20APIs%20may%20change%20in%20the%20future) mentions that the library is still in alpha.

app/src/main/java/org/schabi/newpipe/ui/Toolbar.kt
HatakeKakashri: Undone the toolbar colors to unblock this PR as the Theming is being handled in #12654. Let us handle the Jetpack compose theming as well in your PR.

HatakeKakashri:
Closing this PR as the requirements are handled in other PRs. The Navigation, screen title management as well as Toolbar are handled in https://github.com/TeamNewPipe/NewPipe/pull/12803 and the theming changes are handled in https://github.com/TeamNewPipe/NewPipe/pull/12654

PR #12404, opened 2025-07-02
Implement Compose-based Error Panel, Error UI Model, and Tests for Comments

### LINKED ISSUE #11716: CommentSection: Add an error panel
> An error panel should be created to handle all of these cases and allow retrying, reporting and opening in browser.

_Originally posted by @Stypox in https://github.com/TeamNewPipe/NewPipe/pull/11060#discussion_r1836822135_
            
https://github.com/TeamNewPipe/NewPipe/blob/fdf36cbad61d5995b87244ae576f2e41b0610f68/app/src/main/java/org/schabi/newpipe/ui/components/video/comment/CommentSection.kt#L112-L116

SttApollo:
#### What is it?
- [x] Feature (user facing)

- Created a new Compose-based `ErrorPanel` component.
- Ensured the new error panel is fully theme-adaptive using `MaterialTheme.colorScheme` and that all text is centered for visual parity with the legacy design.
- Introduced a new ErrorPanelSpec data class to flexibly describe error panel content and actions.
- Created `ErrorUiModel` interface to represent different failure states along with `UnableToLoadCommentsUiModel `specific to Comments section.
- Added a focused test suite: Unit tests for `ErrorUiModel` and Integration tests for the two main error states in the comment section: `Resource.Error` (initial load failure) and `LoadState.Error` (paging failure). Would add Compose testing with approval to add a new dependency.
- **Manual Testing**: Verified error handling by forcing errors in both `CommentsViewModel` (Resource.Error) and `CommentsSource` (LoadState.Error) to ensure proper UI rendering and error model mapping.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

- Fixes #11716 

#### APK testing


SttApollo:
@Stypox Apologies for the fairly large diff—thought it would make sense to add a new Compose-based ErrorPanel and have it wired up end-to-end as a first step towards migration. I’m happy to split it up or tweak anything to fit the overall architecture—just let me know!

app/src/main/java/org/schabi/newpipe/ui/components/common/ErrorPanel.kt
```diff
@@ -0,0 +1,117 @@
...
+import org.schabi.newpipe.ui.theme.SizeTokens.SpacingLarge
+import org.schabi.newpipe.ui.theme.SizeTokens.SpacingMedium
+import org.schabi.newpipe.ui.theme.SizeTokens.SpacingSmall
+
+@Composable
+fun ErrorPanel(
+    spec: ErrorPanelSpec,
+    onRetry: () -> Unit,
+    onReport: () -> Unit,
+    onOpenInBrowser: () -> Unit,
```
Stypox: `onReport` should not be needed, as reporting can be done solely based on the information you already have about the error (i.e. ~the spec~ edit: the `ErrorInfo`), so it can be done within this composable instead of delegating it to the outside. `onOpenInBrowser` could become an `openInBrowserUrl: String?` that when `null` hides the button, and when non-null makes the button open that url (oh and also this field would go ~into the spec~ edit: into `ErrorInfo`). Finally, `onRetry` could be a `(() -> Unit)?` that when `null` hides the button, and when non-null makes the button execute that lambda. So you would remove the three boolean fields from the spec.

app/src/main/java/org/schabi/newpipe/ui/components/common/ErrorPanel.kt
```diff
@@ -0,0 +1,117 @@
...
+        if (spec.showOpenInBrowser) {
+            ServiceColoredButton(onClick = onOpenInBrowser) {
+                Text(stringResource(R.string.open_in_browser).uppercase())
+            }
+        }
+        Spacer(Modifier.height(SpacingExtraLarge))
+    }
+}
+
+data class ErrorPanelSpec(
```
Stypox: I would use directly the already existing `ErrorInfo` class, which is also shared with other error proxies in the app (e.g. report screen, error notification, error snackbar).

But I support that you added a place for service explanations, those are definitely needed and we don't have them currently. However the service explanation can only be extracted from the error received from the extractor, so you could add a `fun getExplanation(): String` in `ErrorInfo` that tries to extract information from the `Throwable` (e.g. `throwable.getMessage()` iirc) and returns an empty string if there is no such information.

Then in the future we could expand that `getExplanation` function to add custom messages for specific errors (e.g. age restricted videos on youtube explaining that NewPipe has no way to show age restricted youtube content anymore).

app/src/main/java/org/schabi/newpipe/paging/CommentsSource.kt
```diff
@@ -9,19 +9,22 @@ import org.schabi.newpipe.extractor.Page
...
 import org.schabi.newpipe.ui.components.video.comment.CommentInfo
+import java.io.IOException
 
 class CommentsSource(private val commentInfo: CommentInfo) : PagingSource<Page, CommentsInfoItem>() {
     private val service = NewPipe.getService(commentInfo.serviceId)
 
     override suspend fun load(params: LoadParams<Page>): LoadResult<Page, CommentsInfoItem> {
         // params.key is null the first time the load() function is called, so we need to return the
         // first batch of already-loaded comments
+        return LoadResult.Error(IOException("💥 forced test error"))
```
Stypox: I suppose this remained here by accident? :-)

app/src/main/java/org/schabi/newpipe/ui/components/video/comment/CommentErrorHandler.kt
```diff
@@ -0,0 +1,64 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import org.schabi.newpipe.error.ErrorInfo
+import org.schabi.newpipe.error.UserAction
+import org.schabi.newpipe.error.mapThrowableToErrorUiModel
+import org.schabi.newpipe.ui.UiModel.ErrorUiModel
+import org.schabi.newpipe.ui.components.common.ErrorPanel
+import java.io.IOException
+
+@Composable
+fun CommentErrorHandler(
```
Stypox: Yeah what I meant with the comments above is *kind of* to make `ErrorPanel` do what this function currently does (though the parameters would be `ErrorInfo`, `onRetry` and `modifier`.

SttApollo:
@Stypox I have incorporated your requests - removed the other files and just added one test file. Also I didn't want to add an additional dependency for compose testing but please let me know if you'd like me to!

app/src/main/java/org/schabi/newpipe/ui/components/common/ErrorPanel.kt
```diff
@@ -0,0 +1,132 @@
...
+            ErrorAction.SOLVE_CAPTCHA -> {
+                ServiceColoredButton(onClick = {
+                    // Starting ReCaptcha Challenge Activity
+                    val intent = Intent(context, ReCaptchaActivity::class.java).apply {
+                        putExtra(
+                            ReCaptchaActivity.RECAPTCHA_URL_EXTRA,
+                            (errorInfo.throwable as ReCaptchaException).url
+                        )
+                        addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                    }
```
Isira-Seneviratne:
```suggestion
                    val intent = Intent(context, ReCaptchaActivity::class.java)
                        .putExtra(
                            ReCaptchaActivity.RECAPTCHA_URL_EXTRA,
                            (errorInfo.throwable as ReCaptchaException).url
                        )
                        .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```

app/src/main/java/org/schabi/newpipe/ui/components/video/comment/CommentSection.kt
```diff
@@ -100,27 +107,34 @@ private fun CommentSection(
                                 )
                             }
                         }
-
                         when (comments.loadState.refresh) {
```
Isira-Seneviratne:
```suggestion
                        when (val refresh = comments.loadState.refresh) {
```

app/src/main/java/org/schabi/newpipe/ui/components/video/comment/CommentSection.kt
```diff
@@ -100,27 +107,34 @@ private fun CommentSection(
...
                             is LoadState.Loading -> {
                                 item {
                                     LoadingIndicator(modifier = Modifier.padding(top = 8.dp))
                                 }
                             }
-
                             is LoadState.Error -> {
+                                val error = (comments.loadState.refresh as LoadState.Error).error
+                                val errorInfo = ErrorInfo(
+                                    throwable = error,
```
Isira-Seneviratne:
```suggestion
                                val errorInfo = ErrorInfo(
                                    throwable = refresh.error,
```

app/src/main/java/org/schabi/newpipe/paging/CommentsSource.kt
```diff
@@ -9,19 +9,22 @@ import org.schabi.newpipe.extractor.Page
 import org.schabi.newpipe.extractor.comments.CommentsInfo
 import org.schabi.newpipe.extractor.comments.CommentsInfoItem
 import org.schabi.newpipe.ui.components.video.comment.CommentInfo
+import java.io.IOException
```
absurdlylongusername: Unused import can be removed

app/src/main/java/org/schabi/newpipe/ui/components/common/ErrorPanel.kt
```diff
@@ -0,0 +1,131 @@
...
+        is AccountTerminatedException -> ErrorAction.REPORT
+        else -> ErrorAction.REPORT
+    }
+}
+
+@Composable
+fun ErrorPanel(
+    errorInfo: ErrorInfo,
+    onRetry: (() -> Unit)? = null,
+    modifier: Modifier = Modifier
```
absurdlylongusername: This doesn't seem to be used. Are there plans to use it? If not, I think it should be removed

app/src/androidTest/java/org/schabi/newpipe/ui/components/video/comment/CommentSectionErrorIntegrationTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    fun setUp() {
+        context = ApplicationProvider.getApplicationContext()
+    }
+
+    // Test 1: Network error on initial load (Resource.Error)
+    @Test
+    fun testInitialCommentNetworkError() {
+        val expectedMessage = "Connection timeout"
+        val networkError = SocketTimeoutException(expectedMessage)
+        val resourceError = Resource.Error(networkError)
```
absurdlylongusername: So for this test, and other tests, we create `resourceError`, `loadStateError`, etc. which we pass in a throwable, but the actual object is not used in the test, just the throwable.

So I think the intermediate object is not necessary, and can just pass the exception/throwable into the ErrorInfo constructor directly.

E.g. do not create `resourceError`, just pass `networkError` directly

app/src/androidTest/java/org/schabi/newpipe/ui/components/video/comment/CommentSectionErrorIntegrationTest.kt
```diff
@@ -0,0 +1,119 @@
...
+            request = "comments"
+        )
+        assertEquals(ErrorAction.SOLVE_CAPTCHA, determineErrorAction(errorInfo))
+        assertEquals(expectedMessage, errorInfo.getExplanation())
+
+        val intent = Intent(context, org.schabi.newpipe.error.ReCaptchaActivity::class.java).apply {
+            putExtra(org.schabi.newpipe.error.ReCaptchaActivity.RECAPTCHA_URL_EXTRA, url)
+            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+        }
+        assertEquals(url, intent.getStringExtra(org.schabi.newpipe.error.ReCaptchaActivity.RECAPTCHA_URL_EXTRA))
```
absurdlylongusername: Correct me if I am wrong, but I am not sure this code is actually testing user code. 

It creates an intent within this test, puts in an extra, and then asserts that the extra exists in the intent.
Unless this has side effects, I don't think this is testing functionality inside of ErrorInfo or ErrorPanel, so it should probably be removed (unless it's just not finished yet)

app/src/androidTest/java/org/schabi/newpipe/ui/components/video/comment/CommentSectionErrorIntegrationTest.kt
```diff
@@ -0,0 +1,119 @@
...
+
+        val errorAction = determineErrorAction(errorInfo)
+        assertEquals("Network errors should get REPORT action", ErrorAction.REPORT, errorAction)
+        var retryCallbackInvoked = false
+        val mockCommentRetry = {
+            retryCallbackInvoked = true
+        }
+
+        mockCommentRetry()
+        assertTrue("Retry callback should be invoked when user clicks retry", retryCallbackInvoked)
```
absurdlylongusername: Similarly, this code is also confined to this test and doesn't test production code, so should be removed unless it's just not finished yet

SttApollo: You made some good points - I have just made this test into a unit test until we have compose testing in place.

app/src/main/java/org/schabi/newpipe/ui/components/common/ErrorPanel.kt
```diff
@@ -0,0 +1,130 @@
...
+        Spacer(Modifier.height(SpacingExtraLarge))
+    }
+}
+
+@Preview(showBackground = true, widthDp = 360, heightDp = 640)
+
+@Composable
+fun ErrorPanelPreview() {
+    AppTheme {
+    }
```
Isira-Seneviratne, line 104: Is this still in progress?

SttApollo: Ahh - this is my bad. I meant to finish it,  have added it now.

absurdlylongusername:
> @Stypox I have incorporated your requests - removed the other files and just added one test file. Also I didn't want to add an additional dependency for compose testing but please let me know if you'd like me to!

@SttApollo Tbh I have no qualms with you adding compose testing, because we are going to be adding UI tests very soon anyway. Currently we have no UI tests whatsoever, and I don't really see a downside to having tests for new functionality.

@Stypox, thoughts?

Stypox:
Yeah it would be cool if you added some tests! Would you be ok with opening a new PR after this one is merged, to setup the UI tests infrastructure and add tests for what you added in this PR? Otherwise you can also add UI tests in this PR, let us know what you prefer :-)

SttApollo:
Sure, I can open a new PR after this one is merged. For the UI test infra, I was thinking we could start with Compose instrumented testing—does that align with what you had in mind? Or should we also add Robolectric-based Compose tests and/or snapshot testing early on?

SttApollo:
@Stypox Hi!  I rebased onto the latest version and integrated your changes to resolve the conflicts. I’ve retested, and it looks good. If you’re fine with it, please merge; I can add Compose tests in a follow-up PR, or include them here if you prefer.

absurdlylongusername:
@SttApollo The CI and tests are failing after your latest commit. Please fix it so they pass

SttApollo:
> @SttApollo The CI and tests are failing after your latest commit. Please fix it so they pass

yee working on it

SttApollo:
The CI and tests are passing now, thanks. Pls lmk of anything else!

app/src/main/java/org/schabi/newpipe/ui/components/video/comment/CommentSection.kt
```diff
@@ -50,6 +57,7 @@ private fun CommentSection(
     val comments = commentsFlow.collectAsLazyPagingItems()
     val nestedScrollInterop = rememberNestedScrollInteropConnection()
     val state = rememberLazyListState()
+    val context = LocalContext.current
```
absurdlylongusername: This is not used and can be removed

app/src/main/res/layout/fragment_video_detail.xml
```diff
@@ -214,6 +214,14 @@
                     android:layout_marginTop="@dimen/video_item_detail_error_panel_margin"
                     android:visibility="gone"
                     tools:visibility="gone" />
+                <androidx.compose.ui.platform.ComposeView
```
absurdlylongusername: This doesn't seem to be doing anything from what I can see. Is this a placeholder for future work to replace the xml error panel above it?

app/src/androidTest/java/org/schabi/newpipe/ui/components/common/CommentSectionErrorTest.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.runner.RunWith
+import org.schabi.newpipe.R
+import org.schabi.newpipe.error.ErrorInfo
+import org.schabi.newpipe.error.UserAction
+import org.schabi.newpipe.extractor.exceptions.ReCaptchaException
+import java.io.IOException
+import java.net.SocketTimeoutException
+
+@RunWith(AndroidJUnit4::class)
+class CommentSectionErrorTest {
```
absurdlylongusername: Since this is actually just testing `ErrorInfo`, I think this should be moved to ErrorInfoTest. There isn't any actual UI code being tested here.

app/src/main/res/layout/fragment_video_detail.xml
SttApollo: You’re correct —it’s an unused leftover from a Compose experiment. I’ll remove it in this PR since it's not really being used.

app/src/androidTest/java/org/schabi/newpipe/ui/components/common/CommentSectionErrorTest.kt
SttApollo: I see your point, I have renamed and moved it into the same error package instead of the same file for readability but lmk if you prefer the same file.

Stypox:
@absurdlylongusername feel free to merge if you are also ok with the changes, and then add the PR title + number to the `0.28.1` draft release :-)

TobiGr:
Not to the 0.28.1 draft, but the refactor one, please :)

app/src/androidTest/java/org/schabi/newpipe/ui/components/common/CommentSectionErrorTest.kt
absurdlylongusername: I still think it should be in `ErrorInfoTest`. The tests you added do not specifically test anything related to the comments: they only test certain mapping code paths of `ErrorInfo` (e.g. ReCaptcha is not limited to comments AFAIK). 

I would only regard them as comments tests if they are higher up in the pipeline related to `CommentSection`, `CommentsSource`, etc.

There's many other code paths in ErrorInfo that we don't have tests for, so ideally we would already have ErrorInfoTest that test all expected behaviour of ErrorInfo in a more generalised fashion, which would in turn also test the code paths related to this PR.

Since fully testing ErrorInfo would be a lot of code, this should be done in a different PR and an issue should be created to track this and also leave a TODO in ErrorInfoTest to add the rest of the tests.

absurdlylongusername:
Hi @SttApollo, sorry for late response.

Thanks for your work thus far.

> I can add Compose tests in a follow-up PR, or include them here if you prefer.

I'd prefer them to be in this PR, but ultimately up to you which you want: I just want them to be done ASAP. Ideally I don't want them being in a different PR to delay them being done, so I'm hoping you would have already started working on them, but if not then I guess it doesn't matter lol.

-------

Just to confirm I understand the purpose of this PR: the purpose is to add an error panel specifically for the comments section as stated in #11716, right? The ErrorPanel.kt that was added is not intended to replace the existing error_panel.xml at a later date, is it? Because if so, it doesn't have the same structure (i.e. error_panel.xml has three text fields as you can see in the image, but ErrorPanel.kt only has one)

<details> <summary> Error panel preview </summary>
image
</details>

> I was thinking we could start with Compose instrumented testing—does that align with what you had in mind?

This is fine. Would you be testing just `ErrorPanel` and `CommentSection`, or would you also plan to add higher level integration tests that replicate what is done for manual testing?

------

Was also going to suggest improving the ErrorPanel previews so that they show all the buttons, or have multiple previews for each expected code path; but @Stypox has just done that. 

I'd also suggest having a preview for light and dark mode like we have in many other places.

I.e. adding this to ErrorPanel preview

````kotlin
@Preview(
    name = "Light mode",
    showBackground = true,
    uiMode = Configuration.UI_MODE_NIGHT_NO
)
@Preview(
    name = "Dark mode",
    showBackground = true,
    uiMode = Configuration.UI_MODE_NIGHT_YES
)
````

**Comment by Stypox** [comment]
@absurdlylongusername this is meant to completely replace error_panel.xml. It only has one text field, yes, but that's to reflect the fact that now the error message is chosen by the ErrorInfo instead of by the UI. Sure, we could add "getMainMessage()", "getSubmessage()", etc. to ErrorInfo, but I don't think it's worth it (the second and third lines of text were only used for a couple of errors even before).

File: app/src/androidTest/java/org/schabi/newpipe/ui/components/common/CommentSectionErrorTest.kt (hunk 16 above)
Comment (SttApollo) [reply app/src/androidTest/java/org/schabi/newpipe/ui/components/common/CommentSectionErrorTest.kt]: I understand your POV. For this PR, I can add my tests to the existing ErrorInfoTest file. Would you prefer I:

1. Add them to ErrorInfoTest.java (converting my Kotlin to Java), or
2. Convert ErrorInfoTest to Kotlin and add my tests there?

Once I've added Compose tests to this PR, I can open a new issue to track comprehensive ErrorInfo testing (as you mentioned) to address the remaining untested code paths separately.

**Comment by SttApollo** [comment]
@absurdlylongusername 
> I'd prefer them to be in this PR, but ultimately up to you which you want: I just want them to be done ASAP. Ideally I don't want them being in a different PR to delay them being done, so I'm hoping you would have already started working on them, but if not then I guess it doesn't matter lol.

Yes np, I'll add them to this PR in the next day or two.


> Just to confirm I understand the purpose of this PR: the purpose is to add an error panel specifically for the comments section as stated in #11716, right? The ErrorPanel.kt that was added is not intended to replace the existing error_panel.xml at a later date, is it? Because if so, it doesn't have the same structure (i.e. error_panel.xml has three text fields as you can see in the image, but ErrorPanel.kt only has one)

It is intendend to replace the xml in the future as Stypox mentioned. 

> > I was thinking we could start with Compose instrumented testing—does that align with what you had in mind?
> This is fine. Would you be testing just `ErrorPanel` and `CommentSection`, or would you also plan to add higher level integration tests that replicate what is done for manual testing?

This is my plan so far 
1. Compose tests for ErrorPanel: render the panel with sample data representing each path network error, retryable error, ReCaptcha, and assert the right text and buttons are visible.
2. Comments screen instrumentation test: launch the comments UI on device/emulator, run through load, scroll, retry, and ReCaptcha so it mirrors the manual check.

> Was also going to suggest improving the ErrorPanel previews so that they show all the buttons, or have multiple previews for each expected code path; but @Stypox has just done that.
> I'd also suggest having a preview for light and dark mode like we have in many other places.

Yes, will add this too.

File: app/src/androidTest/java/org/schabi/newpipe/ui/components/common/CommentSectionErrorTest.kt (hunk 16 above)
Comment (absurdlylongusername) [reply app/src/androidTest/java/org/schabi/newpipe/ui/components/common/CommentSectionErrorTest.kt]: I think ErrorInfoTest is small enough that it could be converted to Kotlin and you could add your tests to it, so do number 2.

**Comment by absurdlylongusername** [comment]
@SttApollo @Stypox  Okay cool, fantastic

**Comment by absurdlylongusername** [comment]
> @absurdlylongusername this is meant to completely replace error_panel.xml. It only has one text field, yes, but that's to reflect the fact that now the error message is chosen by the ErrorInfo instead of by the UI. Sure, we could add "getMainMessage()", "getSubmessage()", etc. to ErrorInfo, but I don't think it's worth it (the second and third lines of text were only used for a couple of errors even before).

Is there an issue currently to track replacing errorpanel.xml in the rest of the code?

**Comment by Stypox** [comment]
> Is there an issue currently to track replacing errorpanel.xml in the rest of the code?

No, because we want to replace every `*.xml` with Compose code, and we can't open 1000 issues for each XML file xD

But maybe this is a special case, where it would make sense to have an issue lying around that can be worked on. But what I think makes more sense is to just use the Compose error panel whenever converting a screen to Compose (I included this in #12674 for lists).

**Comment by absurdlylongusername** [comment]
@Stypox 

> No, because we want to replace every *.xml with Compose code, and we can't open 1000 issues for each XML file xD

Touché, lol.

>  just use the Compose error panel whenever converting a screen to Compose

Yeah I'm more talking about noting this down somewhere so that it is documented. Idk where though.

**Comment by SttApollo** [comment]
@absurdlylongusername I ended up adding an inline `TestCommentSection` helper in `CommentSectionInstrumentedTest` instead of pulling in additional mock libraries, lmk if this approach works!

File: app/src/main/java/org/schabi/newpipe/ui/components/common/ServiceColoredButton.kt [hunk 17]
```diff
@@ -0,0 +1,54 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import org.schabi.newpipe.ui.theme.AppTheme
+import org.schabi.newpipe.ui.theme.SizeTokens.SpacingMedium
+import org.schabi.newpipe.ui.theme.SizeTokens.SpacingSmall
+
+@Composable
+fun ServiceColoredButton(
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
```
theimpulson, line 23: Please place modifiers before any other parameters in composable. Lint shows a warning otherwise.

app/src/main/java/org/schabi/newpipe/ui/components/common/ServiceColoredButton.kt
```diff
@@ -0,0 +1,54 @@
...
+
+        ),
+    ) {
+        content()
+    }
+}
+
+@Preview
+@Composable
+fun ServiceColoredButtonPreview() {
```
theimpulson: Please make preview functions private as they aren't supposed to be visible outside of the file.

app/src/main/java/org/schabi/newpipe/ui/components/video/comment/CommentSection.kt
```diff
@@ -98,21 +105,33 @@ private fun CommentSection(
...
+                                    request = "comments"
+                                )
+
                                 item {
-                                    // TODO use error panel instead
-                                    EmptyStateComposable(EmptyStateSpec.ErrorLoadingComments)
+                                    Box(
+                                        modifier = Modifier
+                                            .fillMaxWidth()
+
```
theimpulson: Unused empty space

app/src/main/java/org/schabi/newpipe/util/Localization.java
```diff
@@ -71,6 +72,46 @@ public final class Localization {
...
+     * available (e.g. inside of Compose previews). In that case this method falls back to plain old
+     * {@link Context#getString}.
+     * <p>This method also supports format args (see {@link #compatGetString(Context, int,
+     * Object...)}, unlike {@link ContextCompat#getString}.</p>
+     *
+     * @param context any Android context, even the App context
+     * @param resId the string resource to resolve
+     * @return the resolved string
+     */
+    public static String compatGetString(final Context context, @StringRes final int resId) {
```
theimpulson, line 88: Why not use the `stringResource` composable method?

app/src/main/java/org/schabi/newpipe/ui/components/common/ServiceColoredButton.kt
SttApollo: Hi! I ran lint locally and didn’t see any warnings about the modifier order. Compose docs also suggest placing modifier after required params, before lambdas. If you’re seeing a warning, would you mind sharing the exact message or IDE/lint version? Thanks!

app/src/main/java/org/schabi/newpipe/util/Localization.java
SttApollo: Apologies - those edits must come from earlier base branch commits and probably should be addressed separately.

absurdlylongusername:
@SttApollo You seem to have reverted @Stypox's changes in https://github.com/TeamNewPipe/NewPipe/pull/12404/commits/aed4278388e9d9264ec3465a60d1ed1db9492ae7. 

Please restore them.

app/src/main/java/org/schabi/newpipe/util/Localization.java
absurdlylongusername: @theimpulson These changes were made by @Stypox in https://github.com/TeamNewPipe/NewPipe/pull/12404/commits/aed4278388e9d9264ec3465a60d1ed1db9492ae7. Please ask him about it

Stypox: Yeah that was me; the comments should explain why already, but in practice, we were using ContextCompat.getString() in some places (apparently because of localization issues when the App context is used to get strings, it was introduced in https://github.com/TeamNewPipe/NewPipe/commit/893a227ab169782a539d86b2bb90602f2d2f281f) but that method does not work under Preview because it relies on an API that is not yet simulated by previews.

app/src/main/java/org/schabi/newpipe/ui/components/common/ServiceColoredButton.kt
theimpulson: My bad, seems I mixed something up here. Please ignore.

theimpulson:
LTGM!

PR #12408, opened 2025-07-05
fix: handle nullable nextPage behavior when searching albums #12401

koukibadr:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
- Handle display search results when `nextPage` is null on Albums mode

#### Before/After Screenshots/Screen Record
- Before:

https://github.com/user-attachments/assets/0d4df6ca-25bc-4133-85a7-6bda199d870e

- After:

https://github.com/user-attachments/assets/cc46fe7d-7035-49af-88e9-e0e469c0fdfa

#### Fixes the following issue(s)
- Fixes #12401 

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #12455, opened 2025-07-22
Use correct fix for nextPage being null while creating error report

TobiGr:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

#12408 does not fix the underlying issue but tries to handle the symptoms. `nextPage` cannot be `null` when `handleNextItems()` is called, because at that point there is a page that was fetched.
#12408 made it so that is unlikely that error reports are generated. When there was an ExtractionException it is likely that there is no nextPage. The issue is that nextPage is reassigned before generating the error report.

#### APK testing

#### Due diligence


koukibadr:
@TobiGr 
#12408 does fix the pagination error crash, `handleNextItems` is called whenever there's a fetch for new data the fix was adding an assert to the page object when it's retrieved using `nextPage = result.getNextPage();` so newPage could be null in this stage.
With this PR the retrieval of newPage is done after the assertion if so it won't be null, thisPR and #12408 are two possible fix solutions for the same bug.

PR #12446, opened 2025-07-19
Ensure that the formatters in ``PlayerHelper`` always use the app language

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
* Currently the formatters of ``PlayerHelper`` do not respect the app language as outlined in https://github.com/TeamNewPipe/NewPipe/issues/12441#issuecomment-3092540288

#### Before/After Screenshots/Screen Record

Language set to ``Deutsch/German`` on Android 10 (Pixel 5 emulator)

- Before: <br>
- After: <br>

#### Fixes the following issue(s)
- Fixes #12441 

#### Relies on the following changes
* ⚠️ Should be finalized after https://github.com/TeamNewPipe/NewPipe/pull/12444 is done

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #12444, opened 2025-07-19
Enable per-app language preferences for Android < 13

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Update AppCompat to 1.7.1 ([changelog](https://developer.android.com/jetpack/androidx/releases/appcompat))
- Enable AppCompat's per-app language functionality for Android versions below 13 (Tiramisu). Due to the configuration being automatically updated after the app language change, the `assureCorrectAppLanguage` method has also been removed.

#### Before/After Screenshots/Screen Record

* Before (Android 12):

https://github.com/user-attachments/assets/aab4905f-dc6d-4f1d-a134-9d34494f6705

* After (Android 12):

https://github.com/user-attachments/assets/ae7aa4b5-838d-43b3-8a36-99b108dc8898

#### Fixes the following issue(s)

- Fixes #

#### Relies on the following changes

- 

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #12470, opened 2025-07-26
Improve and cleanup PlayerHelper formatters/localization

### LINKED ISSUE #12441: Speed/Tempo use system language for formatting and not app language
### Affected version

nightly-2025-07-18 (0.28.0 rc)

Due to some reason the system language - not the app language - is used for formatting the tempo

Image

This problem was not present in 0.27.7:

Image

Situation to reproduce this:
* System language: German (or anything with different format)
* App specific language set via Android: English
* App language set via app: English

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Followup of https://github.com/TeamNewPipe/NewPipe/pull/12446

1. Contains some cleanup regarding the PlayerHelper localization and related code, removes dead code, etc.
2. If the language is changed via ContentSettingsFragment, PlayerHelper's formatters will be reset. If formatting is now requested next time it will be reinitialized and the correct format will be applied.

<details><summary>Here's an example</summary>

[Np12470Showcase1.webm](https://github.com/user-attachments/assets/9b6f3734-f192-4466-81f8-a7df6e9f4e20)

</details>

- Fixes #12441 (kind of, was already addressed mostly by https://github.com/TeamNewPipe/NewPipe/pull/12444)

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/PlayerHelper.java
@@ -118,18 +118,6 @@ public final class PlayerHelper {
     }
 
     @NonNull
-    public static String subtitleMimeTypesOf(@NonNull final MediaFormat format) {
-        switch (format) {
-            case VTT:
-                return MimeTypes.TEXT_VTT;
-            case TTML:
-                return MimeTypes.APPLICATION_TTML;
-            default:
-                throw new IllegalArgumentException("Unrecognized mime type: " + format.name());
-        }
-    }
-
-    @NonNull
     public static String captionLanguageOf(@NonNull final Context context,
                                            @NonNull final SubtitlesStream subtitles) {
         final String displayName = subtitles.getDisplayLanguageName();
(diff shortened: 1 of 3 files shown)


Stypox:
Actually, wait, it does not work while the player is open, because `formatters()` is called again by the player before the user has a chance to click on the language in the language chooser.

Stypox:
So the correct solution would probably be to attach to https://developer.android.com/reference/android/content/Intent#ACTION_LOCALE_CHANGED, buuuuuuuuut let's keep it simple. I added `PlayerHelper.resetFormat();` in `ContentSettingsFragment.onDestroy()` and it works. It will mean the player formatters will be reset every time the user exits content settings, but whatever.

TobiGr:
What happens if the user changes the system language or the per-app pref while the player is open? That is not covered by the hack.

Stypox:
The old formatting remains in that case, until the app is closed and reopened, but I'd say it's an edge case and it's not worth implementing a more complex solution at the cost of introducing bugs related to broadcast receivers

Issue #11139, opened 2024-06-07
[YouTube] "Sign in to confirm..."/"Watch on the latest version..."


zestygrass:
### Checklist

- [X] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [X] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [X] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [X] This issue contains only one bug.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.27.0

### Steps to reproduce the bug

1. Attempt to play any video
2. Video does not play
3. Page shows "Content unavailable" page

### Expected behavior

Video should play.

### Actual behavior

Video does not play.

## Exception
* __User Action:__ requested stream
* __Request:__ https://www.youtube.com/watch?v=RXQOpChgksI
* __Content Country:__ GB
* __Content Language:__ en-GB
* __App Language:__ en_GB_#u-fw-mon-mu-celsius
* __Service:__ YouTube
* __Version:__ 0.27.0
* __OS:__ Linux Android 14 - 34
<details><summary><b>Crash log </b></summary><p>

```
org.schabi.newpipe.extractor.exceptions.ContentNotAvailableException: Got error: "Sign in to confirm that you're not a bot"
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.checkPlayabilityStatus(YoutubeStreamExtractor.java:948)
	at org.schabi.newpipe.extractor.services.youtube.extractors.YoutubeStreamExtractor.onFetchPage(YoutubeStreamExtractor.java:844)
	at org.schabi.newpipe.extractor.Extractor.fetchPage(Extractor.java:60)
	at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:77)
	at org.schabi.newpipe.extractor.stream.StreamInfo.getInfo(StreamInfo.java:72)
[...]

PR #12477, opened 2025-07-28
Add more specific error messages and deduplicate their handling

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This PR deduplicates the decision of which error message to show for which throwable (it was done in 3 different places previously). It also adds some nice new error messages for specific errors, so users understand better what is going on (and maybe complain less when YouTube breaks something ;-) ).
- In `RouterActivity`, a toast is shown like before if the error is not something that usually needs to be reported, as to not distract the user with a notification.
- Make error snackbar display the actual error that has occurred, instead of showing "Sorry, something went wrong" every time. For example, "Could not parse website" is much more explanatory as the snackbar description when some item in a list can not be extracted (this is already shown in the error panel).
- All subclasses of `ContentNotAvailableException` now have nice specific error messages, and in particular this PR adds `UnsupportedContentInCountryException` (from @AudricV's new Youtube trending implementation) and `YoutubeSignInConfirmNotBotException` (https://github.com/TeamNewPipe/NewPipeExtractor/pull/1352). The latter in particular will hopefully help reduce the amount of duplicate comments in https://github.com/TeamNewPipe/NewPipe/issues/11139.
- Various player-related exceptions are now inspected to extract the original cause and show that to the user instead of always saying "Could not play this stream". In particular, 403 errors are detected and a custom error message is shown for them: "HTTP error 403 occurred while playing, likely caused by an IP ban or streaming URL deobfuscation issues" (this message was inspired by [FreeTube's](https://github.com/FreeTubeApp/FreeTube/issues/6834)). This should also help reduce the amount of duplicates we get related to 403 errors.

- https://github.com/TeamNewPipe/NewPipeExtractor/pull/1352
- @AudricV's new Youtube trending implementation

#### APK testing

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -214,7 +214,7 @@ dependencies {
     // the corresponding commit hash, since JitPack sometimes deletes artifacts.
     // If there’s already a git hash, just add more of it to the end (or remove a letter)
     // to cause jitpack to regenerate the artifact.
-    implementation 'com.github.TeamNewPipe:NewPipeExtractor:7adbc48a0aa872c016b8ec089e278d5e12772054'
+    implementation 'com.github.Stypox:NewPipeExtractor:1c04bd88c3f1e6b1e0c912814035745e61bb9aba'
     implementation 'com.github.TeamNewPipe:NoNonsense-FilePicker:5.0.0'
 
 /** Checkstyle **/
--- a/app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
+++ b/app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
@@ -23,6 +23,8 @@ import org.schabi.newpipe.extractor.exceptions.UnsupportedContentInCountryExcept
 import org.schabi.newpipe.extractor.exceptions.YoutubeMusicPremiumContentException
 import org.schabi.newpipe.extractor.exceptions.YoutubeSignInConfirmNotBotException
 import org.schabi.newpipe.ktx.isNetworkRelated
+import org.schabi.newpipe.player.mediasource.FailedMediaSource
+import org.schabi.newpipe.player.resolver.PlaybackResolver
 import org.schabi.newpipe.util.ServiceHelper
 
 @Parcelize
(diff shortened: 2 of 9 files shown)


app/build.gradle
```diff
@@ -214,7 +214,7 @@ dependencies {
     // the corresponding commit hash, since JitPack sometimes deletes artifacts.
     // If there’s already a git hash, just add more of it to the end (or remove a letter)
     // to cause jitpack to regenerate the artifact.
-    implementation 'com.github.TeamNewPipe:NewPipeExtractor:7adbc48a0aa872c016b8ec089e278d5e12772054'
+    implementation 'com.github.Stypox:NewPipeExtractor:1c04bd88c3f1e6b1e0c912814035745e61bb9aba'
```
litetex, line 217: This needs to be changed before merging

app/src/main/java/org/schabi/newpipe/error/ErrorInfo.kt
```diff
@@ -86,30 +99,57 @@ class ErrorInfo(
             if (info == null) SERVICE_NONE else ServiceHelper.getNameOfServiceById(info.serviceId)
 
         @StringRes
-        private fun getMessageStringId(
+        fun getMessageStringId(
             throwable: Throwable?,
```
litetex, line 103: Might be a stupid question but why are we catching ``Throwable``s in the first place and not ``Exception``s?

If we encounter an ``Error`` that usually means a critical error that must result in a VM crash, like ``OutOfMemoryError``, etc

Stypox: For example when the app crashes and ACRA catches the throwable and starts the error report screen, that throwable could be anything. In other places... well yeah maybe we don't need to catch Throwables, but just to be sure sometimes we do. E.g. `StackOverflowError` is not an `Exception`, but if we have a recursive stack overflow somewhere (e.g. the extractor) we still want to handle it nicely instead of crashing the app.

litetex:
Was closing this intentional?

TobiGr:
This pointed to the release branch which was deleted and thus the PR closed automatically. This should be reopened targeting `dev`

litetex:
@Stypox 
Reminder

PR #12488, opened 2025-08-01
Use Core I18n to format plurals

Isira-Seneviratne:
#### What is it?

- Format several plural messages such as the watching, view counts using the [Core i18n](https://developer.android.com/jetpack/androidx/releases/core#core-i18n-1.0.0) library's [`MessageFormat`](https://developer.android.com/reference/kotlin/androidx/core/i18n/MessageFormat). For more information, see [here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core-i18n/src/main/java/androidx/core/i18n/messageformat_icu/simple/MessageFormat.java).
- Rename some methods and perform some refactors, including removing an unused class.
- Remove now-unused string resources, as they have been combined into the MessageFormat-style strings.

- Before:
- After:

- Fixes #

#### APK testing

#### Due diligence


Stypox:
I don't understand, why did you remove a lot of translated plurals? Also, what's the problem with the current code?

Isira-Seneviratne:
Those were removed because the original plurals were no longer being used. I could re-add them in the new string resources.

There's nothing wrong with the existing code, but it's [recommended](https://developer.android.com/guide/topics/resources/string-resource#Plurals) to use MessageFormat (which the library backports) instead of plurals.

PR #12530, opened 2025-08-12
Improve skipping silences.

### LINKED ISSUE #10760: Add user-facing settings for configuring the parameters of the skip-silence feature
### Checklist

### Feature description

Am I correct in thinking that this: https://github.com/google/ExoPlayer/issues/6705 having been resolved should make it possible to enable the user to configure the skip-silence feature, so they can change the minimum duration of silence to be skipped, the threshold volume below which is classified as 'silence', and the duration of 'padding' on the edges of any silent portion?

Similar to but more general than #10479 

Having these parameters as configurable options could help in adapting the silence skipping to fit different videos which could differ in the volume of spoken parts and speed of speech etc. Being stuck with one set of values for those parameters can lead to the silence skipping being too agressive for some videos and not enough so for others.

### Additional information

_No response_

a-shtefan:
ExoPlayer allows controlling silence skipping by adjusting the silence duration that is removed by `SilenceSkippingAudioProcessor`. The default behavior with the default parameters for this component makes it hard to listen to content because it leaves almost no pauses. After this commit it is possible to adjust silence length that remains and shorten only long pauses.

#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

- allows changing silence duration which remains after fast forwarding

- Before:
  
- After:
  

- Fixes #10760 

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -72,6 +72,7 @@ import com.google.android.exoplayer2.PlaybackParameters;
 import com.google.android.exoplayer2.Player.PositionInfo;
 import com.google.android.exoplayer2.Timeline;
 import com.google.android.exoplayer2.Tracks;
+import com.google.android.exoplayer2.audio.SilenceSkippingAudioProcessor;
 import com.google.android.exoplayer2.ext.mediasession.MediaSessionConnector;
 import com.google.android.exoplayer2.source.MediaSource;
 import com.google.android.exoplayer2.text.CueGroup;
@@ -293,10 +294,20 @@ public final class Player implements PlaybackListener, Listener {
                 new DefaultBandwidthMeter.Builder(context).build());
         loadController = new LoadController();
 
-        renderFactory = prefs.getBoolean(
-                context.getString(
-                        R.string.always_use_exoplayer_set_output_surface_workaround_key), false)
-                ? new CustomRenderersFactory(context) : new DefaultRenderersFactory(context);
+        final boolean alwaysUseExoplayerSetOutputSurfaceWorkaround = prefs.getBoolean(
+                context.getString(R.string.always_use_exoplayer_set_output_surface_workaround_key),
+                false);
+        final int maxSilenceDurationMillis = prefs.getInt(
+                context.getString(R.string.max_silence_duration_key),
+                Integer.parseInt(context.getString(R.string.max_silence_duration_value)));
+        final SilenceSkippingAudioProcessor silenceSkippingAudioProcessor =
+                new SilenceSkippingAudioProcessor(
+                        MILLISECONDS.toMicros(maxSilenceDurationMillis),
+                        MILLISECONDS.toMicros(maxSilenceDurationMillis),
+                        SilenceSkippingAudioProcessor.DEFAULT_SILENCE_THRESHOLD_LEVEL);
+        renderFactory = new CustomRenderersFactory(
+                context, alwaysUseExoplayerSetOutputSurfaceWorkaround,
+                silenceSkippingAudioProcessor);
 
         renderFactory.setEnableDecoderFallback(
                 prefs.getBoolean(
--- a/app/src/main/java/org/schabi/newpipe/player/helper/CustomRenderersFactory.java
+++ b/app/src/main/java/org/schabi/newpipe/player/helper/CustomRenderersFactory.java
@@ -5,6 +5,12 @@ import android.os.Handler;
 
 import com.google.android.exoplayer2.DefaultRenderersFactory;
 import com.google.android.exoplayer2.Renderer;
+import com.google.android.exoplayer2.audio.AudioCapabilities;
+import com.google.android.exoplayer2.audio.AudioProcessor;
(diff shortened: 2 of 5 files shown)


lawson85:
Looking forward to this!

a-shtefan:
Do I need to fix/change anything? It's been over 2 months, which is a lot for a tiny change, and there is zero feedback

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -293,10 +294,20 @@ public Player(@NonNull final PlayerService service,
...
+                context.getString(R.string.always_use_exoplayer_set_output_surface_workaround_key),
+                false);
+        final int maxSilenceDurationMillis = prefs.getInt(
+                context.getString(R.string.max_silence_duration_key),
+                Integer.parseInt(context.getString(R.string.max_silence_duration_value)));
+        final SilenceSkippingAudioProcessor silenceSkippingAudioProcessor =
+                new SilenceSkippingAudioProcessor(
+                        MILLISECONDS.toMicros(maxSilenceDurationMillis),
+                        MILLISECONDS.toMicros(maxSilenceDurationMillis),
+                        SilenceSkippingAudioProcessor.DEFAULT_SILENCE_THRESHOLD_LEVEL);
```
AudricV, line 314: For reference for NewPlayer, [this approach has been changed in Media3](https://developer.android.com/reference/androidx/media3/exoplayer/audio/SilenceSkippingAudioProcessor?hl=en#SilenceSkippingAudioProcessor(long,float,long,int,short)).

I would remove the duplicate call to `MILLISECONDS.toMicros(maxSilenceDurationMillis)` by saving the result in a variable.

Can you explain why you set the same value to `minimumSilenceDurationUs` and `paddingSilenceUs`?

app/src/main/java/org/schabi/newpipe/player/helper/CustomRenderersFactory.java
```diff
@@ -22,8 +28,17 @@
  */
```
AudricV, line 29: The JavaDoc of this class should be updated.

app/src/main/res/xml/video_audio_settings.xml
```diff
@@ -77,6 +77,17 @@
         app:singleLineTitle="false"
         app:iconSpaceReserved="false"/>
 
+    <SeekBarPreference
```
Stypox: I wonder whether this should go to the playback parameters dialog, so it's easily accessible from the player, and it also does not clutter settings. And there it should be shown only when "Fast-forward during silence" is enabled, to further reduce clutter for users that don't enable "Fast-forward during silence". Let me know what you think.

Stypox (changes requested):
Thank you for this nice addition! Sorry for the late review from my side, too. I haven't had enough time to keep up with the various PRs lately.

I agree with @AudricV, I think this can be merged as it's rather small and shouldn't introduce issues.

app/src/main/java/org/schabi/newpipe/player/Player.java
a-shtefan: Moved it to a variable.
The duration is passed both as `minimumSilenceDurationUs` and `paddingSilenceUs` so that when the silence longer than `maxSilenceDuration` is detected, it is truncated to `maxSilenceDuration`. Silence detection length is controlled by `minimumSilenceDurationUs` and the duration which is left is controlled by `paddingSilenceUs`. So when we set them to the same value, we achieve this truncation behavior.
I think this behavior is reasonable and it also unclutters the settings by having only one parameter to control silence skipping.
If these variables are independent, the behavior could be different, e.g. if `minimumSilenceDurationUs` is 2s and `paddingSilenceUs` is 0.01s, it means that silences up to 2 seconds are kept as-is, but silences longer than 2 seconds get truncated to 0.01s. To me this behavior seems unnatural so I made these values the same.

app/src/main/java/org/schabi/newpipe/player/helper/CustomRenderersFactory.java
a-shtefan: Added description to JavaDoc.

app/src/main/res/xml/video_audio_settings.xml
a-shtefan: I'd be glad to add it in the menu, but the issue here is that applying the change to the value requires reinitializing the player. `SilenceSkippingAudioProcessor` doesn't support changing its parameters after it is created, so the only way to apply the change to the parameter is initializing the player from scratch. I am not really familiar with the codebase, if such behavior makes sense and if it is relatively easy to do it, I'd be happy to implement it, but if it is too complicated or it will lead to some weird artifacts (such as player freezing while it reinitializes), I'd prefer to leave it as it is now.

AudricV: This preference should go to ExoPlayer settings in my opinion.

a-shtefan: Moved it to ExoPlayer settings (also changed the default from 1000ms to 500ms)

ghost:
Any updates on it?

PR #12544, opened 2025-08-16
Add option to delete a download without also deleting file

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

Some user at FrOSCon reported that the "Delete" button in the Download screen was confusing because they expected it to only delete the entry instead of also deleting the file, so I made that clearer by changing the string from "Delete" to "Delete file" and also added a "Delete entry" button that just deletes the entry.

image

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/us/shandian/giga/service/DownloadManager.java
+++ b/app/src/main/java/us/shandian/giga/service/DownloadManager.java
@@ -265,7 +265,7 @@ public class DownloadManager {
         }
     }
 
-    public void deleteMission(Mission mission) {
+    public void deleteMission(Mission mission, boolean alsoDeleteFile) {
         synchronized (this) {
             if (mission instanceof DownloadMission) {
                 mMissionsPending.remove(mission);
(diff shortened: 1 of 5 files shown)


HatakeKakashri:
I tried this on my phone and realized that there is no confirmation before deleting the download/file. I understand that the user has already selected the 'Delete' option, but should we have a confirmation dialog to warn the user that the file is going to be deleted? Something like this: 
image

Stypox:
@snaik20 there is a confirmation in the form of an "Undo" button in the snackbar. If you accidentally delete something, you can undo within one minute or so (before the snackbar disappears)

HatakeKakashri:
> @snaik20 there is a confirmation in the form of an "Undo" button in the snackbar. If you accidentally delete something, you can undo within one minute or so (before the snackbar disappears)

You are right, the undo option is a good way to prevent accidental deletions. 

My suggestion was for a similar purpose, but I’m happy to go with the existing approach with 'undo'.

snaik20 (approved):

PR #2892, opened 2019-12-25
Sleep timer

javedh-dev:
Hi all,

I have updated the target SDK version to 29 and a few other dependency versions to its latest.
I have also added a sleep timer similar to Spotify in background player.

Please review the changes and merge.

- [x] I carefully read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md) and agree to them.

Issue #2830, opened 2019-11-30
[Feature request] Watching/listening timer limit


JaytimeGame:
For people using the app to listen music it would be great to have the option to set a timer for a certain amount of time and then when this specific time period is done the app pause/stop any media (in the app) being used.

Redirion:
Just install a sleep timer app from the Play Store?

JaytimeGame:
I know I can download a app from the app store to do this(That's actually what I do) 
But it would still be a cool little feature that would make the app better.

thiswillbeyourgithub:
For anyone wondering, this app is great :
https://f-droid.org/packages/fr.smarquis.sleeptimer/

it just adds a tile in the settings panel that, when enabled, automatically lowers then shuts down the media being played.

litetex:
Although this was closed it popped up again in #2892

My opinion to this topic:
Out of scope - this is primary an app for playing videos and music. 
If you need a timer use a timer app 😄 

Also newer Android versions (probably 9+, vendor specific) are shipped with a automatic shutdown/start functionality (``Schedule power off/on``) which powers off the complete phone and not just NewPipe 😉

Atemu:
I think you're misunderstanding what a sleep timer means in this case; it's to automatically stop playing music after some time or number of tracks/playlist items, not time anything.

The external solution technically works but it'd be better if it was integrated and it can't do things like stop after a certain amount of tracks. The "unix philosophy" is a nice thing but it doesn't *really* work on Android.  
Lots of people use NP to play music; things like sleep timers are essential features here. Could this issue be re-opened?

Sam4000der3:
Newpipe still not show videos of " - Topic" Channel. 

Videos and Playlists of this type of channels can only be open manuelly with the share option in YouTube / Browser.

I hope that will be fixed.

abodsalas:
It would make this the perfect app to sleep listening to videos....
Here is a possible mockup.

![Timer_mockup](https://user-images.githubusercontent.com/54449155/198292153-a69c7a0d-00c5-4f5e-bf5b-938a5de69589.jpg)

Roy-Orbison:
NewPipe is a de facto music player, and this is a standard feature in others. Spotify has it under the **⋮** menu of the ‘now playing’ screen:
![Spotify sleep timer](https://github.com/TeamNewPipe/NewPipe/assets/28617290/687d1250-6373-4abf-80cc-377a1a4cecd6)

[...]

PR #12557, opened 2025-08-20
Sleep Timer (new attempt)

zgtm:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Adds a sleep timer menu with some predefined values.

The sleep timer pauses the video as soon as the timer runs out.

- Before:
Screenshot_20250821-001942
Screenshot_20250821-001910

- After:
Screenshot_20250820-233737
Screenshot_20250820-233813

Screenshot_20250820-233741
Screenshot_20250820-233817

Screenshot_20250820-233753
Screenshot_20250820-233822

- based on `refactor` branch, can also port to `dev` if useful

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -122,6 +122,7 @@ import org.schabi.newpipe.util.SerializedCache;
 import org.schabi.newpipe.util.StreamTypeUtil;
 import org.schabi.newpipe.util.image.CoilHelper;
 
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.IntStream;
@@ -262,6 +263,9 @@ public final class Player implements PlaybackListener, Listener {
     private final SharedPreferences prefs;
     @NonNull
     private final HistoryRecordManager recordManager;
+
+    java.util.Timer sleepTimer;
+    java.time.Instant sleepTimerEnd;
 
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -2261,6 +2265,44 @@ public final class Player implements PlaybackListener, Listener {
         reloadPlayQueueManager();
     }
 
+
+    public void setSleepTimer(@Nullable final long sleepMinutes) {
+        if (sleepTimer != null) {
+            sleepTimer.cancel();
+            sleepTimer.purge();
+            sleepTimer = null;
+        }
+
+        sleepTimerEnd = java.time.Instant.now().plus(sleepMinutes, ChronoUnit.MINUTES);
+
+        sleepTimer = new java.util.Timer();
+        //final Player thisPlayer = this;
+        final java.util.TimerTask task = new java.util.TimerTask() {
+
+
+            @Override
+            public void run() {
+                if (java.time.Instant.now().compareTo(sleepTimerEnd) >= 0) {
+                    UIs.call(playerUi -> playerUi.onSleepTimerUpdate(0));
+                    cancelSleepTimer();
+                    return;
+                }
+
+                final long remainingMinutes = java.time.Instant.now().until(sleepTimerEnd,
+                        ChronoUnit.MINUTES);
+                UIs.call(playerUi -> playerUi.onSleepTimerUpdate(remainingMinutes + 1));
+            }
+        };
+        sleepTimer.schedule(task, 1000, 1000);
+    }
+
+    public void cancelSleepTimer() {
+        if (sleepTimer != null) {
+            sleepTimer.cancel();
+            sleepTimer.purge();
+            sleepTimer = null;
+        }
+    }
 
     @NonNull
     public Context getContext() {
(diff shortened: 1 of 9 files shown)


zgtm:
I added a new sleep timer after the PR https://github.com/TeamNewPipe/NewPipe/pull/2892 got closed. This is not based on the other PR, but I tried to address the issues with the other PR.

Some notes / open questions:
- I added two new icons (`ic_alarm` and `ic_alarm_off`). I took the icons from https://github.com/marella/material-design-icons/ . I don't know if that is the right way to add these icons or if it does even matter …
- I don't know if the icons / menu placement is to everyones liking. I experimented a bit and this seemed to be the best place to me, but I am open to move the icons / menu around
- I added the StringArray `sleep_timer_description` and the IntArray `sleep_timer_value` to `settings_keys.xml`. This does not seem to be the right place, but I added them there because `seek_duration_description` and `seek_duration_value` live there. Should I create a new XML-file (e.g. `menu_items.xml`)?
- I don't know where the values in [app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java](https://github.com/TeamNewPipe/NewPipe/pull/12557/files#diff-cf3fb991bbb40afdfbf6afc6cf207999cf33a5e31043f3e697c34b28fb31be2f) (lines 127-131) come from. Is the value I picked there available?
- When the timer runs out, the `Player` calls `playerUi.onSleepTimerUpdate` in the timer-Thread. The UI then uses `binding.sleepTimerTextView.post` to switch to the UI thread, to call `player.pause`. It seems a bit convoluted to me, but it is the cleanest way I could think of. Is there a better way to do this?

zgtm:
Linking to https://github.com/TeamNewPipe/NewPipe/issues/2830 and https://github.com/TeamNewPipe/NewPipe/pull/2892 for visibility (both ways).

app/src/main/res/values-de/strings.xml
```diff
@@ -829,4 +829,7 @@
     <string name="channel_tab_likes">Gefällt mir</string>
     <string name="migration_info_6_7_title">SoundCloud-Top-50-Seite entfernt</string>
     <string name="migration_info_6_7_message">SoundCloud hat die ursprünglichen Top-50-Charts abgeschafft. Der entsprechende Tab wurde von deiner Hauptseite entfernt.</string>
+    <string name="sleep_timer_popup_title">Sleep Timer</string>
+    <string name="sleep_timer_set">Sleep Timer stellen</string>
+    <string name="sleep_timer_cancel">Sleep Timer beenden</string>
```
TobiGr, line 834: ~~TODO: Check VLC translation. AFAIK~~ they use a different word than "Sleep Timer". We might want to use the same terminology (German and English). 

They use "Ruhemodus". Not sure if that is a good translation. But "Sleep Timer" is English. We might want to let the translators figure this out :)

app/src/main/java/org/schabi/newpipe/player/ui/PlayerUi.java
```diff
@@ -209,4 +209,11 @@ public void onPlayQueueEdited() {
      */
     public void onVideoSizeChanged(@NonNull final VideoSize videoSize) {
     }
+
+    /**
+     * @param remainingTime the remaining sleep timer time, set to 0 to pause the player and
+     *                      disable the sleep timer
+     */
+    public void onSleepTimerUpdate(final long remainingTime) {
+    }
```
TobiGr, line 218: Please ensure that the time unit is mentioned.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2262,6 +2266,44 @@ public void setAudioTrack(@Nullable final String audioTrackId) {
...
+            public void run() {
+                if (java.time.Instant.now().compareTo(sleepTimerEnd) >= 0) {
+                    UIs.call(playerUi -> playerUi.onSleepTimerUpdate(0));
+                    cancelSleepTimer();
+                    return;
+                }
+
+                final long remainingMinutes = java.time.Instant.now().until(sleepTimerEnd,
+                        ChronoUnit.MINUTES);
+                UIs.call(playerUi -> playerUi.onSleepTimerUpdate(remainingMinutes + 1));
```
TobiGr, line 2293: is `remainingMinutes + 1` used because `until` floors the value? If so, add an explaining comment.

app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java
```diff
@@ -1561,6 +1646,32 @@ public void onVideoSizeChanged(@NonNull final VideoSize videoSize) {
     }
     //endregion
 
+    @Override
+    public void onSleepTimerUpdate(final long remainingTime) {
+        if (remainingTime == 0) {
+            binding.sleepTimerTextView.post(new Runnable() {
+                public void run() {
+                    player.pause();
```
TobiGr, line 1654: this method is just used for UI updates. The pause() call should happen inside the player.

Stypox:
Thank you for this PR! I am not sure this is the right time to start adding this feature. NewPlayer is about to be integrated on the refactor branch and that will mean throwing away most of the current player code. So it might be better if you postpone this to after NewPlayer is integrated in NewPipe.

PR #12575, opened 2025-08-27
[Bug] Fix missing subtitle text in manually downloaded *.SRT files. (issue #10030)

### LINKED ISSUE #10030: Download subtitles, when the language is zh-TW Chinese (Taiwan) to download, the subtitles in the srt file will be blank, only the timeline has no text records
### Checklist

### Affected version

0.25.1

Simple scope:
Download Youtube subtitles, when the language is zh-TW Chinese (Taiwan) to download, the subtitles in the srt file will be blank, only the timeline has no text records

srt file loaded with MX player will show file error

Detailed scope:
0.Install MX player and CX file manager
1.Install Youtube 18.05.40
2.Install ReVanced Manager 0.0.57
3.ReVanced Manager Patcher Select Youtube and Selected patches Recommended and install Youtube ReVanced
4.Install NewPipe 0.25.1
5.Open Youtube ReVanced watch https://youtu.be/mtb-qa8xvFU and download video and subtitles (language : zh-TW Chinese (Taiwan))
6.Open Video by MX player
7.MX player load Srt subtitles
8.Show error
9.Use CX file manager open subtitles can see Srt only record timeline has no text records

Bad file:
https://drive.google.com/file/d/1aY12_di6YXjE2J-3j-amguavYcXe-CbN/view?usp=sharing

### Expected behavior

No error & Srt file correct.

### Actual behavior

MX player load Srt error.
Open the Srt file to check the content and found that the subtitles are missing and only the timeline

_No response_

### Logs

_No response_

_No response_

### Affected device model

_No response_

### Additional information

_No response_

TransZAllen:
#### What is it?
- [✔ ] Bugfix (user facing)
- [ ] Feature (user facing)

##### Problem
Downloaded SRT subtitles for some videos are empty, containing only timestamps and sequence numbers, as reported in #10030. This affects videos with styled subtitles, such as:
- https://youtu.be/mtb-qa8xvFU(https://www.youtube.com/watch?v=mtb-qa8xvFU) (【HimeHina MV】Roki (Cover))
- https://www.youtube.com/watch?v=zbQRY8KSVbU (Ousama Ranking Opening 2 Full『Hadaka no Yuusha』by Vaundy)
- https://youtu.be/-eDYT_20YhM(https://www.youtube.com/watch?v=-eDYT_20YhM) (炜WARD ROMANCE ft. Feng Yi)
- https://www.youtube.com/watch?v=L-BgxLtMxh0 (Styled subtitles for YouTube demonstration)
- https://www.youtube.com/watch?v=Cc2nkx77U24 (Test: Rainbow Captions In Youtube)
- https://www.youtube.com/watch?v=lUDPjyfmJrs (【original anime MV】III【hololive/宝鐘マリン＆こぼ・かなえる】)

##### Problem Analysis
The issue occurs because some YouTube subtitles use TTML format with nested tags, which the original parser did not handle correctly.

###### Problematic TTML Example
```xml
<p begin="00:00:01.000" end="00:00:03.000">
  <span style="s4">Hello World!</span>
</p>

The text ("Hello World!") is nested inside a <span> tag for styling (e.g., colors or karaoke effects). The original parser only processed direct child nodes, missing the text inside <span>, resulting in empty SRT output.
Non-Problematic TTML Example
<p begin="00:00:01.000" end="00:00:03.000" style="s2">
  Hello World!
</p>

This TTML has text directly under <p>, which was parsed correctly by the original code.
Root Cause
The original SrtFromTtmlWriter.build method used a non-recursive loop, failing to extract text from nested tags like <span> in styled subtitles (e.g., rainbow or karaoke captions).
Solution

Added a new extractText method to recursively extract text from all nodes, including TextNode and <br> tags, handling nested tags like <span>.
Replaced the non-recursive loop in SrtFromTtmlWriter.build with a call to extractText().

<!-- If your PR changes the app's UI in any way, please include screenshots or a video showing exactly what changed, so that developers and users can pinpoint it easily. Delete this if it doesn't apply to your PR.-->
- Before:
- After:

<!-- Prefix issues with "Fixes" so that GitHub closes them when the PR is merged (note that each "Fixes #" should be in its own item). Also add any other relevant links. -->
- Fixes #10030

<!-- Delete this if it doesn't apply to your PR. -->
- None

#### APK testing
<!-- Use a new, meaningfully named branch. The name is used as a suffix for the app ID to allow installing and testing multiple versions of NewPipe, e.g. "commentfix", if your PR implements a bugfix for comments. (No names like "patch-0" and "feature-1".)  -->
<!-- Remove the following line if you directly link the APK created by the CI pipeline. Directly linking is preferred if you need to let users test.-->
- Fixed Cases (previously empty SRT files)

https://youtu.be/mtb-qa8xvFU (【HimeHina MV】Roki (Cover))
https://www.youtube.com/watch?v=zbQRY8KSVbU (Ousama Ranking Opening 2 Full『Hadaka no Yuusha』by Vaundy)
https://youtu.be/-eDYT_20YhM (炜WARD ROMANCE ft. Feng Yi)
https://www.youtube.com/watch?v=L-BgxLtMxh0 (Styled subtitles for YouTube demonstration)
https://www.youtube.com/watch?v=Cc2nkx77U24 (Test: Rainbow Captions In Youtube)
https://www.youtube.com/watch?v=lUDPjyfmJrs (【original anime MV】III【hololive/宝鐘マリン＆こぼ・かなえる】)

- Regression Testing

Tested a video with simple subtitles that downloaded correctly in the original NewPipe: https://www.youtube.com/watch?v=BVAIImxcv4g .
Confirmed SRT output remains correct after the fix, ensuring no regression.

All tested videos now produce correct SRT files with subtitle text.


#### Due diligence
- [ ✔] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
+++ b/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
@@ -54,6 +54,30 @@ public class SrtFromTtmlWriter {
         out.write(text.getBytes(charset));
     }
 
+    /*
+     *   Recursive method to extract text from all nodes
+     *   - This method processes TextNode and <br> tags, recursively
+     *     extracting text from nested tags.
+     *     For example: extract text from nested <span> tags
+     *   - Appends newlines for <br> tags.
+     */
+    private void extractText(final Node node, final StringBuilder text) {
+        if (node instanceof TextNode) {
+            text.append(((TextNode) node).text());
+        } else if (node instanceof Element) {
+            final Element element = (Element) node;
+            // <br> is a self-closing HTML tag used to insert a line break.
+            if (element.tagName().equalsIgnoreCase("br")) {
+                // Add a newline for <br> tags
+                text.append(NEW_LINE);
+            }
+        }
+        // Recursively process child nodes
+        for (final Node child : node.childNodes()) {
+            extractText(child, text);
+        }
+    }
+
     public void build(final SharpStream ttml) throws IOException {
         /*
          * TTML parser with BASIC support
@@ -81,14 +105,8 @@ public class SrtFromTtmlWriter {
         for (final Element paragraph : paragraphList) {
             text.setLength(0);
 
-            for (final Node children : paragraph.childNodes()) {
-                if (children instanceof TextNode) {
-                    text.append(((TextNode) children).text());
-                } else if (children instanceof Element
-                        && ((Element) children).tagName().equalsIgnoreCase("br")) {
-                    text.append(NEW_LINE);
-                }
-            }
+            // Recursively extract text from all child nodes
+            extractText(paragraph, text);
 
             if (ignoreEmptyFrames && text.length() < 1) {
                 continue;

### CODE REVIEW

**Comment by TobiGr** [comment]
Thank you for the bug fix. I applied a few JDoc and codestyle fixes. You already compiled a good list of subtitles which can be used for testing. Do you mind creating a few unit tests to test this?

**Comment by TransZAllen** [comment]
> Thank you for the bug fix. I applied a few JDoc and codestyle fixes. You already compiled a good list of subtitles which can be used for testing. Do you mind creating a few unit tests to test this?

I'm a newcomer to the NewPipe project and this is my first code submission. When I heard about adding `unit tests`, I got a bit confused 😄. I quickly started preparing them, but then I saw that your review had already been approved! 
So, I’ll skip the `unit tests` this time.  Thanks for your review and feedback! 👍 
And by the way, I noticed the changes were merged into the v0.28.x branch.

File: app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java [hunk 1]
```diff
@@ -54,6 +54,36 @@ private void writeString(final String text) throws IOException {
...
+            // <br> is a self-closing HTML tag used to insert a line break.
+            if (element.tagName().equalsIgnoreCase("br")) {
+                // Add a newline for <br> tags
+                text.append(NEW_LINE);
+            }
+        }
+        // Recursively process child nodes
+        for (final Node child : node.childNodes()) {
+            extractText(child, text);
+        }
```
Isira-Seneviratne: The build method could be rewritten like this:

```suggestion
        final List<Pair<Element, String>> pairList = doc.selectStream("body > div > p")
                .map(paragraph -> {
                    // Element.text extracts from child nodes as well
                    return new Pair<>(paragraph, paragraph.text());
                })
                .filter(pair -> !ignoreEmptyFrames || !pair.second.isEmpty())
                .toList();

        for (final var pair : pairList) {
            final var paragraph = pair.first;
            final var text = pair.second;

            final String begin = getTimestamp(paragraph, "begin");
            final String end = getTimestamp(paragraph, "end");

            writeFrame(begin, end, text);
        }
```

TobiGr: > Element.text extracts from child nodes as well

But does it convert `<br>` tags to a new line? At least not from what I could find [in the docs](https://jsoup.org/apidocs/org/jsoup/nodes/Element.html#textNodes()).

TobiGr:
> I'm a newcomer to the NewPipe project and this is my first code submission.

Hi and welcome!

> . When I heard about adding unit tests, I got a bit confused 😄. I quickly started preparing them, but then I saw that your review had already been approved!

If you wrote some, feel free to add them to the PR. Our test coverage is dramatically low.

>  And by the way, I noticed the changes were merged into the v0.28.x branch.

This PR has not been merged yet. I just added it to the project to ensure it is merged before the 0.28.1 release.

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TransZAllen: @Isira-Seneviratne 

Thanks for the suggestion! 

I gave it a try locally, but it failed to build （built with `dev` branch） because:

1. `Pair` requires the `commons-lang3` dependency, which is currently not part of NewPipe.
2. `selectStream(...)` is only available in newer Jsoup versions. The current setup doesn’t support it.

In addition, the recursive `extractText()` method is dependency-free, has been tested extensively, and keeps the logic clearer for future modifications (e.g. handling special tags). 

For this reason, **I’d prefer to keep the current approach `extractText()`** instead of introducing new dependencies or relying on newer APIs.

Isira-Seneviratne: Android has a built-in Pair type. Also, the current Jsoup version in the project has the method.

@TobiGr You're right, but it works fine for the subtitles from the linked videos. Alternatively, the map operation could be changed to use the existing logic.

TobiGr: I honestly don't see a good reason to use streams here. It makes the whole code harder to read and I am not sure if we achieve an increased performance by using a stream here. If stream is significantly faster here, we could make use of it though.

Isira-Seneviratne: I suggested using the stream API as it doesn't need to collect all the DOM elements first before doing the subtitle processing.

TransZAllen: Hi @Isira-Seneviratne and @TobiGr,

I tested the stream API and Pair locally. A few observations:

1. The downloaded `.srt` files actually contain TTML content (XML-like code), not plain text that users can read.
2. The debug version crashes almost every time (essentially 100%) when processing subtitles.

Test code I used locally:

```java
 final List<Pair<Element, String>> pairList = doc.select("body > div > p")
     .stream()
     .map(paragraph -> {
         final String text = paragraph.text();
         return new Pair<>(paragraph, text);
     })
     .filter(pair -> !ignoreEmptyFrames || !pair.second.isEmpty())
     .toList();

 for (final var pair : pairList) {
     final var paragraph = pair.first;
     final var text = pair.second;
     final String begin = getTimestamp(paragraph, "begin");
     final String end = getTimestamp(paragraph, "end");
     writeFrame(begin, end, new StringBuilder(text));
>}
 ```

Comment (TransZAllen) [reply app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java]: And here is the diff code
```
allen@allen-Inspiron-3558:/media/allen/WinUbuShare/NewPipe_ubuntu/branch_dev/NewPipe$ git show
commit 77e03b3168936b8c98b52192d8f10272bf5f9f63 (HEAD -> dev)
Author: TransZAllen <tree.story@outlook.com>
Date:   Thu Aug 28 15:16:23 2025 +0800

    test a new method

diff --git a/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java b/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
index 902d5f57c..387f3f4fc 100644
--- a/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
+++ b/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
@@ -1,18 +1,19 @@
 package org.schabi.newpipe.streams;
 
+import android.util.Pair;
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Element;
 import org.jsoup.nodes.Node;
 import org.jsoup.nodes.TextNode;
 import org.jsoup.parser.Parser;
-import org.jsoup.select.Elements;
 import org.schabi.newpipe.streams.io.SharpStream;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
 
 /**
  * @author kapodamy
@@ -88,34 +89,28 @@ public class SrtFromTtmlWriter {
          * Language parsing is not supported
          */
 
-        // parse XML
         final byte[] buffer = new byte[(int) ttml.available()];
         ttml.read(buffer);
-        final Document doc = Jsoup.parse(new ByteArrayInputStream(buffer), "UTF-8", "",
-                Parser.xmlParser());
-
-        final StringBuilder text = new StringBuilder(128);
-        final Elements paragraphList = doc.select("body > div > p");
-
-        // check if has frames
-        if (paragraphList.size() < 1) {
-            return;
-        }
-
-        for (final Element paragraph : paragraphList) {
-            text.setLength(0);
-
-            // Recursively extract text from all child nodes
-            extractText(paragraph, text);
-
-            if (ignoreEmptyFrames && text.length() < 1) {
-                continue;
-            }
-
+        final Document doc = Jsoup.parse(new ByteArrayInputStream(buffer),
+                                         "UTF-8",
+                                         "",
+                                         Parser.xmlParser());
+
+        final List<Pair<Element, String>> pairList = doc.select("body > div > p")
+            .stream()
+            .map(paragraph -> {
+                final String text = paragraph.text();
+                return new Pair<>(paragraph, text);
+            })
+            .filter(pair -> !ignoreEmptyFrames || !pair.second.isEmpty())
+            .toList();
+
+        for (final var pair : pairList) {
+            final var paragraph = pair.first;
+            final var text = pair.second;
             final String begin = getTimestamp(paragraph, "begin");
             final String end = getTimestamp(paragraph, "end");
-
-            writeFrame(begin, end, text);
+            writeFrame(begin, end, new StringBuilder(text));
         }
     }
 }
```

Comment (Isira-Seneviratne) [reply app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java]: Which Android version did you test it on?

Comment (TransZAllen) [reply app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java]: @Isira-Seneviratne 
Android 13

Comment (Isira-Seneviratne) [reply app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java]:
```suggestion
        final var paragraphs = doc.selectStream("body > div > p")
                .filter(paragraph -> !ignoreEmptyFrames || paragraph.hasText())
                .toList();

        for (final var paragraph : paragraphs) {
            final String begin = getTimestamp(paragraph, "begin");
            final String end = getTimestamp(paragraph, "end");
            writeFrame(begin, end, paragraph.text());
        }
```

TransZAllen: Right, I changed the parameter to String as well, forgot to add that.

TransZAllen: I have pushed the changes to the remote repository : https://github.com/TransZAllen/NewPipe/commit/c7ab950807d061f0a0f6b1c87be711e00df12e76
```java
allen@allen-Inspiron-3558:/media/allen/WinUbuShare/NewPipe_ubuntu/branch_debug_issue10030/NewPipe$ git show
commit c7ab950807d061f0a0f6b1c87be711e00df12e76 (HEAD -> debug_issue10030, origin/debug_issue10030)
Author: TransZAllen <tree.story@outlook.com>
Date:   Sat Aug 30 16:06:52 2025 +0800

    Test a new method 2 to get the actual text content.
    
    There are still the two observations.

diff --git a/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java b/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
index 5f43185c6..6dc132939 100644
--- a/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
+++ b/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
@@ -6,7 +6,6 @@ import org.jsoup.nodes.Element;
 import org.jsoup.nodes.Node;
 import org.jsoup.nodes.TextNode;
 import org.jsoup.parser.Parser;
-import org.jsoup.select.Elements;
 import org.schabi.newpipe.streams.io.SharpStream;
 
 import java.io.ByteArrayInputStream;
@@ -101,27 +100,14 @@ public class SrtFromTtmlWriter {
                 Parser.xmlParser());
 
         final StringBuilder text = new StringBuilder(128);
-        final Elements paragraphList = doc.select("body > div > p");
-
-        // check if has frames
-        if (paragraphList.isEmpty()) {
-            return;
-        }
-
-        for (final Element paragraph : paragraphList) {
-            text.setLength(0);
-
-            // Recursively extract text from all child nodes
-            extractText(paragraph, text);
-
-            if (ignoreEmptyFrames && text.length() < 1) {
-                continue;
-            }
-
+        final var paragraphs = doc.selectStream("body > div > p")
+                .filter(paragraph -> !ignoreEmptyFrames || paragraph.hasText())
+                .toList();
+        for (final var paragraph : paragraphs) {
             final String begin = getTimestamp(paragraph, "begin");
             final String end = getTimestamp(paragraph, "end");
-
-            writeFrame(begin, end, text);
+            //writeFrame(begin, end, paragraph.text());
+            writeFrame(begin, end, new StringBuilder(paragraph.text()));
         }
     }
 }
allen@allen-Inspiron-3558:/media/allen/WinUbuShare/NewPipe_ubuntu/branch_debug_issue10030/NewPipe$ 
```

Isira-Seneviratne: Right, I changed the parameter to String as well, forgot to add that.

I accidentally edited your reply, sorry about that

TransZAllen: Switching to `paragraph.text()` without verifying whether all edge cases (like special characters or tags) are handled properly could introduce new issues.

The `extractText()` function consolidates NewPipe’s scattered code into a single, recursive function that processes all text content from various tags. This code has been tested for a long time and is well-suited to NewPipe’s design requirements. It guarantees the handling of special characters, especially line breaks (`<br>`), which are explicitly replaced with `\r\n`—something that `paragraph.text()` may not do correctly. Some software, for instance, defines `<br>` as `\n` or even as spaces, and we can't be sure that `paragraph.text()` handles these cases the same way as NewPipe does. And any other special characters, we don't know.

Switching to `paragraph.text()` introduces the risk of losing control over special character handling. If we need to add special handling later, `paragraph.text()` could lock us out of such flexibility and we can't change its code. Essentially, it becomes a black box, and we’re unsure of how it’ll behave with our specific use cases.

For these reasons, I recommend continuing to use `extractText()`.

TransZAllen: Here is the paragraph.text() definition from https://github.com/jhy/jsoup/blob/master/src/main/java/org/jsoup/nodes/Element.java#L1552
```java
    public String text() {
        final StringBuilder accum = StringUtil.borrowBuilder();
        new TextAccumulator(accum).traverse(this);
        return StringUtil.releaseBuilder(accum).trim();
    }
```
And it uses the trim() function. 
After researching it on Google, I found that `trim()` is a standard Java String method used to **remove** leading and trailing whitespace from a string.

Since it removes whitespace, it is not suitable for **subtitles**.

TobiGr: I think the implementation in this PR is already good enough. @Isira-Seneviratne If you do not have any objections, I'd proceed with merging this PR.

Stypox: Yeah I am also in favour of keeping the current code

Stypox:
@TransZAllen it would indeed help if you have some time to write a couple of tests that take some short HTML strings, parse them, and check whether the extracted text is correct. E.g. one test for the old format without styles, and one for the new format. Let us know if you can make it, otherwise this PR is good to go even like this. Thanks!

TransZAllen:
@Stypox 
Okay, I will definitely add the test cases, at least two.
I will start working on them tomorrow, but I'm not sure exactly how long it will take. I'm a bit slow at coding, so please bear with me and I’ll keep you updated.

TransZAllen:
Hi all, just a quick update:

1. About the test target: currently the unit test is verifying `SrtFromTtmlWriter.extractText()` instead of `SrtFromTtmlWriter.build()`. The reason is that I wanted to check whether the plain text content is correctly extracted, without involving paragraph numbering or timestamps. Do you think this approach makes sense, or should we rather test `build()`?

2. The test case is not yet complete, but during testing I discovered an issue:
   - Leading and trailing spaces in subtitle text are removed.
   - Multiple consecutive spaces inside the text are collapsed into a single space.  
   Because of this, I had to modify `extractText()` to handle spaces properly (`TextNode.text()` → `TextNode.getWholeText()`). 
The fix compiles and tests pass now, but I’m still verifying edge cases.  
   Current work is in branch [`debug_issue10030_add_test_case`](https://github.com/TransZAllen/NewPipe/tree/debug_issue10030_add_test_case).

3. Another problem I noticed: the generated *.srt files start paragraph numbering from 0 instead of 1. According to the SubRip (.srt) specification, **numbering should start from 1**. Starting at 0 may still work in some players, but it can NOT with strict parsers.  Here is the beginning of a *.SRT file generated by NewPipe:
```
a51x:/storage/emulated/0/Newpipe $ cat Bear\ Family\ Searches\ for\ Water\ _\ Wild\ Mexico\ _\ BBC\ Earth-en.srt               
0
00:00:00,640 --> 00:00:07,160
This mother has three young cubs just 8

1
00:00:03,840 --> 00:00:07,160
months old.

2
00:00:08,520 --> 00:00:22,899
[Music]

3
00:00:24,560 --> 00:00:29,039
It's autumn and the family needs to

4
00:00:27,119 --> 00:00:32,039
flatten up before the winter
```

Stypox:
@TransZAllen thanks for the thorough analysis!

> Do you think this approach makes sense, or should we rather test build()?

Yeah it makes sense.

> Current work is in branch [debug_issue10030_add_test_case](https://github.com/TransZAllen/NewPipe/tree/debug_issue10030_add_test_case?rgh-link-date=2025-09-23T06%3A22%3A31Z).

Okok thanks for handling the edge cases. Feel free to push here once you are done.

> numbering should start from 1

I guess you could replace `writeString(String.valueOf(frameIndex++));` with `writeString(String.valueOf(++frameIndex));` [here](https://github.com/TransZAllen/NewPipe/blob/74518c8965c54157830e9121513907146bbe959a/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java#L42). And make sure to add a comment explaining that the first frame should have index 1.

TransZAllen:
@Stypox  Thanks! As for the numbering in `*.srt`:

You’re right, the issue is probably right there. Every time I see `++`, I get a bit confused. They’re not very clear to read, so I’d rather avoid them. I prefer using something more explicit like `frameIndex += 1` or `frameIndex = frameIndex + 1`, and I’ll add a comment to clarify that numbering starts from 1.

TransZAllen:
@Stypox 

Regarding the numbering in `*.srt`: 

Instead of fixing it directly in this PR, I’ll first **open a separate issue** to track this problem, since it’s a distinct concern from the text extraction tests.   **Is this Okay?**

This way we keep the scope of the current PR focused on `extractText()` and its tests, while the numbering fix can be reviewed independently.

TransZAllen:
Hi all, just a quick update:

### 1.Unit Tests

I've been working on the unit tests over the past few days. They’re not yet complete, but I’m continuing to check what’s still missing or needs adjustment.

### 2.SRT Numbering

- On Ubuntu, I tested `SRT` files with numbering starting from 0 on multiple players (mpv 0.37.0, VLC 3.0.20, MPlayer 1.5, Totem 43.0).
  - All of them displayed the subtitles correctly, even when numbering started at 0.
  - In fact, `VLC` was even able to display subtitles with numbering starting at -1.

- This is not a playback-breaking bug. However, it **doesn’t follow the SRT specification**, which requires numbering to start at 1. To avoid potential issues with stricter software, I think it’s best to fix this.

  **My plan:**
- I’ll open a separate issue to track the numbering problem today, and then follow up with a PR.
- Roughly, the change will be to split the current line `writeString(String.valueOf(frameIndex++));` into two lines: one to use `frameIndex`, and another to increment it.

&nbsp;&nbsp;&nbsp;&nbsp; I'll get started on it today.

Stypox:
Sure, it's fine if you make a separate PR. Thanks!

TransZAllen:
Hi everyone,

Here is the issue and PR for SRT numbering:
issue:  https://github.com/TeamNewPipe/NewPipe/issues/12670
PR:  https://github.com/TeamNewPipe/NewPipe/pull/12671

TransZAllen:
hi all, just a quick update —

the **Unit Test** is basically done now. I don’t think there’s much left to change, maybe just some tiny comment tweaks or small clean-ups.
overall, it’s pretty much ready.
if nothing new comes up these days, I’ll mark it ready for your review soon.

ale5000-git:
> * Leading and trailing spaces in subtitle text are removed.

Many players do trim every line, I think there is NOT any useful use case for these spaces
but consecutive spaces inside the text are usually NOT collapsed.

> * All of them displayed the subtitles correctly, even when numbering started at 0.
> * In fact, `VLC` was even able to display subtitles with numbering starting at -1.

I think most players (if not all) just ignore indexes completely, you can also write 100000 in the index.
Many player also support subtitles with indexes removed.

TransZAllen:
Thanks for the extra input, @ale5000-git!
Yes, those earlier points about **trimming** and **SRT numbering** were already reviewed before —

I’m currently finalizing the **Unit Test** part and preparing the content to present during the review.  Will ping again once it’s ready. 😊

TransZAllen:
Hi all,

I just pushed a commit that includes the **Unit Test** (`SrtFromTtmlWriterTest.java`). 
I also spent some time writing clear code comments, hoping it helps explain the thought process behind the changes.

Here's the log message for this commit (https://github.com/TeamNewPipe/NewPipe/pull/12575/commits/22ee01bcfbb8655a65fb6ee6f8373cec54987a4f):

>     refactor(ttml): improve extractText() to preserve spaces and special characters
>     
>     - Replaced `text()` with `getWholeText()`:
>       - avoids losing whitespaces at the beginning, end, or within the text;
>       - avoids merging two or more consecutive spaces into a single space ' ';
>       - avoids converting '\r', '\n', and '\r\n' within the text into a single space ' ';
>       For subtitle conversion, the goal is to preserve every character exactly as intended by the subtitle author.
>     - Normalized tabs, line breaks, and other special characters for SRT-safe output.
>     - Added comprehensive unit tests in `SrtFromTtmlWriterTest.java`, including cases for simple and nested tags.
>

TransZAllen:
Hi everyone,

I've just pushed a new commit to optimize the code. Now ,there are four commits. Here is the log message for this commit(https://github.com/TeamNewPipe/NewPipe/pull/12575/commits/35166676716c4e84f57977945651d99b0d5f0859):

>     refactor(ttml): extract recursion into `traverseChildNodesForNestedTags()`
>     
>     - Extracted child-node traversal logic from `extractText()`
>       into a helper method `traverseChildNodesForNestedTags()`.
>     - No functional change.

Please review it, 😃 there are no plans to modify the code for now.

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
```diff
@@ -54,6 +54,210 @@ private void writeString(final String text) throws IOException {
...
+     *         actual (literal) characters.
+     */
+    private String decodeXmlEntities(final String encodedEntities) {
+        final String decoded = Parser.unescapeEntities(encodedEntities, true);
+        return decoded;
+    }
+
+    /**
+     * Handle rare XML entity characters like LF: &#xA;(`\n`)
+     * , CR: &#xD;(`\r`) and CRLF: (`\r\n`).
```
TobiGr:
```suggestion
     * Handle rare XML entity characters like LF: &#xA;(`\n`),
     * CR: &#xD;(`\r`) and CRLF: (`\r\n`).
```

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TobiGr, line 1: Can you link to the specification in the class docs?

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
```diff
@@ -54,6 +54,210 @@ private void writeString(final String text) throws IOException {
...
+        //   so \u00A0 may cause display issues.
+        // - \u00A0 and \u0020 are visually identical (i.e., they both
+        //   appear as spaces ' '), but they differ in Unicode encoding,
+        //   leading to test failures (e.g., ComparisonFailure).
+        // - Convert \u00A0 to \u0020 to ensure consistency in subtitle
+        //   formatting.
+        // - References:
+        //   - Unicode General Punctuation: https://unicode.org/charts/PDF/U2000.pdf
+        //   - TTML Spec: https://www.w3.org/TR/ttml2/
+        //   - SRT Format: https://en.wikipedia.org/wiki/SubRip
```
TobiGr: > SRT subtitles only support regular spaces (\u0020),
 > so \u00A0 may cause display issues.

That's interesting. I could not find any info on this on Wikipedia or other sites (e.g. https://www.loc.gov/preservation/digital/formats//fdd/fdd000569.shtml). Where did you take the strings to test and the expected results from?

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
```diff
@@ -54,6 +54,210 @@ private void writeString(final String text) throws IOException {
...
+     * (&#xA0;, &#160;) forms.
+     *
+     * @param encodedEntities The raw text fragment possibly containing
+     *                        encoded XML entities.
+     * @return A decoded string where all entities are replaced by their
+     *         actual (literal) characters.
+     */
+    private String decodeXmlEntities(final String encodedEntities) {
+        final String decoded = Parser.unescapeEntities(encodedEntities, true);
+        return decoded;
```
TobiGr:
```suggestion
        return Parser.unescapeEntities(encodedEntities, true);
```

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TransZAllen: You're right: there is no explicit official specification stating that NBSP (\u00A0) is unsafe in .srt.
In fact, .srt is only described as a plain text subtitle format, without defining which Unicode spaces are valid.

So this decision is **based mainly on actual rendering results** rather than documentation theory:

- On Linux, `cat -A` displays NBSP as special markers (M-BM-), suggesting it's treated as a control-like character. Here is a screenshot of an .srt file, which contains the result after converting from TTML 

`<p begin="00:00:05.000" end="00:00:07.000">    ～~Hello&nbsp;&nbsp;&nbsp;&amp;&amp;&lt;&gt;World!!   </p>`

Screenshot from 2025-10-28 21-01-00


- In **Acode** (Android editor), NBSP appears as visible red dots, which looks like a broken or unrecognized glyph.
<div align="center">  

</div>

- Works fine in **VS Code** (v1.104.0) NBSP appears as a space.

- Works fine in common video players:
 -1) **mpv** 0.37.0 → NBSP displays as a normal space
 -2) **VLC** 3.0.20 → same behavior

Because .srt viewers **do not interpret non-breaking** behavior anyway, NBSP is effectively degraded into a regular space — but with compatibility risks depending on the viewer.

Stypox: Feel free to add this info in a 2/3-line comment there

Stypox:
From my side I don't have more comments than what Tobi already pointed out. Thanks for the tests! After those comments are solved, this is ready to be merged imo. We'll see if any user complains about misformatted subtitles, but your processing looks reasonable to me.

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TransZAllen: okay, comments of NBSP ('\u00A0') has been updated.

@TobiGr Please see the commit https://github.com/TeamNewPipe/NewPipe/pull/12575/commits/d311faea58a3ee03e95b3473924547c87ccded11  -->adding examples of rendering incorrectly.

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TransZAllen: I added a comment to the header of the class file `app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java`

Is this what you meant?
Please see the new commit: https://github.com/TeamNewPipe/NewPipe/pull/12575/commits/d311faea58a3ee03e95b3473924547c87ccded11
```
diff --git a/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java b/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
index aaf7bff69..6f584d055 100644
--- a/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
+++ b/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
@@ -15,7 +15,11 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 
 /**
- * @author kapodamy
+ * Converts TTML subtitles to SRT format.
+ *
+ * References:
+ *  - TTML 2.0 (W3C): https://www.w3.org/TR/ttml2/
+ *  - SRT format: https://en.wikipedia.org/wiki/SubRip
  */
 public class SrtFromTtmlWriter {
     private static final String NEW_LINE = "\r\n";
```

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TransZAllen: Thanks for the suggestion! ---- I understand the desire to simplify the code — and I agree that both styles are valid.

I kept the local variable (decoded) mainly because:
- It improves readability — the name acts like a short, inline comment
- It makes debugging and logging easier if we ever need to inspect the intermediate value

That said, I’m not strongly attached to this style 😊
If the NewPipe codebase prefers the simplified form, I’m happy to update it — just let me know your preference.

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TobiGr: Yes

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TobiGr: I also prefer self-explaning code. In this case however, just one method is called which is not too complicated and the method name is also clearly describing what's going on. I'd therefore prefer to remove the assignment. If the String was composed using multiple parts I'd also prefer to store it in a var.

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TobiGr: that looks good to me. Thank you!

app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
TransZAllen: Sure, agreed 👍

Thanks for the clarification-----`decodeXmlEntities(final String encodedEntities)`  ----

> the method name is also clearly describing what's going on

Stypox (approved):
Thanks!

TobiGr (approved):
Thank you for the effort and your patience!

Issue #12533, opened 2025-08-13
#11139 better error message - still desired?


HeroponRikiBestest:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.28.0

### Steps to reproduce the bug

Attempt to view a video on an affected network. Given the issue in question, this is obviously difficult to reproduce.

### Expected behavior

Posting this as a new issue on account of https://github.com/TeamNewPipe/NewPipe/issues/11139 still being open, asking for users to test using this PR with its most recent message, but being closed to collaborators for replies. Is feedback related to "Anyone having this problem, could you test if https://github.com/TeamNewPipe/NewPipe/pull/12477 gives a better error message than the current very unspecific one? It should help users understand what is going on, and maybe there will be less duplicate issues." still desired? I apologize if not. If so, the content unavailable screen I get on the latest release, which normally just says "content unavailable" is instead shown as the following:
<img alt="Screenshot_20250811_201618_NewPipe better-error-messages" src="https://github.com/user-attachments/assets/0ee2518e-dfde-4729-8dc5-3148c54ef821" width=307 />

Please let me know if this request is still open/desired; I kept this message info light on account of the message in question being 2 weeks old, so I figure the testing was probably fulfilled and just not updated on that issue.
[...]

PR #12578, opened 2025-08-28
Add more specific error messages and deduplicate their handling

Stypox:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

GitHub closed https://github.com/TeamNewPipe/NewPipe/pull/12477 and I can't reopen it, hence a new PR.

This PR deduplicates the decision of which error message to show for which throwable (it was done in 3 different places previously). It also adds some nice new error messages for specific errors, so users understand better what is going on (and maybe complain less when YouTube breaks something ;-) ).
- In `RouterActivity`, a toast is shown like before if the error is not something that usually needs to be reported, as to not distract the user with a notification.
- Make error snackbar display the actual error that has occurred, instead of showing "Sorry, something went wrong" every time. For example, "Could not parse website" is much more explanatory as the snackbar description when some item in a list can not be extracted (this is already shown in the error panel).
- All subclasses of `ContentNotAvailableException` now have nice specific error messages, and in particular this PR adds `UnsupportedContentInCountryException` (from @AudricV's new Youtube trending implementation) and `YoutubeSignInConfirmNotBotException` (https://github.com/TeamNewPipe/NewPipeExtractor/pull/1352). The latter in particular will hopefully help reduce the amount of duplicate comments in https://github.com/TeamNewPipe/NewPipe/issues/11139.
- Various player-related exceptions are now inspected to extract the original cause and show that to the user instead of always saying "Could not play this stream". In particular, 403 errors are detected and a custom error message is shown for them: "HTTP error 403 occurred while playing, likely caused by an IP ban or streaming URL deobfuscation issues" (this message was inspired by [FreeTube's](https://github.com/FreeTubeApp/FreeTube/issues/6834)). This should also help reduce the amount of duplicates we get related to 403 errors.

- https://github.com/TeamNewPipe/NewPipeExtractor/pull/1352
- @AudricV's new Youtube trending implementation

#### APK testing

Debug APK from CI: [app(1).zip](https://github.com/user-attachments/files/22024346/app.1.zip)

#### Due diligence

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -214,7 +214,7 @@ dependencies {
     // the corresponding commit hash, since JitPack sometimes deletes artifacts.
     // If there’s already a git hash, just add more of it to the end (or remove a letter)
     // to cause jitpack to regenerate the artifact.
-    implementation 'com.github.TeamNewPipe:NewPipeExtractor:7adbc48a0aa872c016b8ec089e278d5e12772054'
+    implementation 'com.github.Stypox:NewPipeExtractor:1c04bd88c3f1e6b1e0c912814035745e61bb9aba'
     implementation 'com.github.TeamNewPipe:NoNonsense-FilePicker:5.0.0'
 
 /** Checkstyle **/
--- a/app/src/main/java/org/schabi/newpipe/error/UserAction.java
+++ b/app/src/main/java/org/schabi/newpipe/error/UserAction.java
@@ -33,7 +33,8 @@ public enum UserAction {
     SHARE_TO_NEWPIPE("share to newpipe"),
     CHECK_FOR_NEW_APP_VERSION("check for new app version"),
     OPEN_INFO_ITEM_DIALOG("open info item dialog"),
-    GETTING_MAIN_SCREEN_TAB("getting main screen tab");
+    GETTING_MAIN_SCREEN_TAB("getting main screen tab"),
+    PLAY_ON_POPUP("play on popup");
 
     private final String message;
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -866,4 +866,6 @@
     <string name="import_settings_vulnerable_format">The settings in the export being imported use a vulnerable format that was deprecated since NewPipe 0.27.0. Make sure the export being imported is from a trusted source, and prefer using only exports obtained from NewPipe 0.27.0 or newer in the future. Support for importing settings in this vulnerable format will soon be removed completely, and then old versions of NewPipe will not be able to import settings of exports from new versions anymore.</string>
     <string name="migration_info_6_7_title">SoundCloud Top 50 page removed</string>
     <string name="migration_info_6_7_message">SoundCloud has discontinued the original Top 50 charts. The corresponding tab has been removed from your main page.</string>
+    <string name="youtube_sign_in_confirm_not_bot_error">YouTube refused to provide data, asking for a login.\n\nYour IP might have been temporarily banned by YouTube, you can wait some time or switch to a different IP (for example by turning on/off a VPN, or by switching from WiFi to mobile data).</string>
+    <string name="unsupported_content_in_country">This content is not available for the currently selected content country.\n\nChange your selection from \"Settings > Content > Default content country\".</string>
 </resources>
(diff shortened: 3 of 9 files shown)


app/build.gradle
```diff
@@ -214,7 +214,7 @@ dependencies {
     // the corresponding commit hash, since JitPack sometimes deletes artifacts.
     // If there’s already a git hash, just add more of it to the end (or remove a letter)
     // to cause jitpack to regenerate the artifact.
-    implementation 'com.github.TeamNewPipe:NewPipeExtractor:v0.24.8'
+    implementation 'com.github.Stypox:NewPipeExtractor:b8bd4cda8cca00a14940933e3d3635d5aafec222'
```
TobiGr: Needs to be updated once the extractor PR is merged.

app/src/main/res/values/strings.xml
```diff
@@ -877,4 +877,7 @@
     <string name="trending_movies">Trending movies and shows</string>
     <string name="trending_music">Trending music</string>
     <string name="entry_deleted">Entry deleted</string>
+    <string name="player_error_403">HTTP error 403 occurred while playing, likely caused by an IP ban or streaming URL deobfuscation issues</string>
```
AudricV: Please make this non YouTube specific. HTTP 403 issues also happen frequently on Soundcloud due to short expiration. I would use something like this instead:

`Unable to play this stream: HTTP error 403 received from server, likely caused by streaming URL(s) expiration or IP ban`.

For other HTTP errors (4XX and 5XX), we should also provide a better error message `Unable to play this stream: got HTTP error [CODE]`

AudricV (changes requested):

app/build.gradle
Stypox: Done

app/src/main/res/values/strings.xml
Stypox: Done, I added even more error messages and had to make various changes. Read the commit message for more info: [`a369dee` (#12578)](https://github.com/TeamNewPipe/NewPipe/pull/12578/commits/a369deeef45b740e3dc4aca34ab23299fbe313e8)

Stypox:
This is how "This account is terminated" looks:
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

I know that now it kinda looks less nice, but now the control over the error is given fully to ErrorInfo, instead of having ErrorPanelHelper change the error message in a hardcoded way only for AccountTerminatedException.

HeroponRikiBestest:
Regarding #12533 ; 
> Thanks for testing everyone! Now development has moved to https://github.com/TeamNewPipe/NewPipe/pull/12578, it would be cool if you could test the APK from there and report here again :-)

So, in the event that I start getting served "sign in to confirm" from YouTube, I should test this APK and report what it gives me? Or is no further testing needed unless something goes wrong with newpipe's error reporting

Stypox:
> So, in the event that I start getting served "sign in to confirm" from YouTube, I should test this APK and report what it gives me? 

@HeroponRikiBestest yes, thanks in advance :-)

HeroponRikiBestest:
> > So, in the event that I start getting served "sign in to confirm" from YouTube, I should test this APK and report what it gives me?
> 
> @HeroponRikiBestest yes, thanks in advance :-)

Screenshot_20250901_010507_NewPipe better-error-messages

Tested with the latest CI build of this PR, this is what it reports when it occurs; seems correct to me.

app/src/main/java/org/schabi/newpipe/error/UserAction.java
```diff
@@ -33,7 +33,9 @@ public enum UserAction {
     SHARE_TO_NEWPIPE("share to newpipe"),
     CHECK_FOR_NEW_APP_VERSION("check for new app version"),
     OPEN_INFO_ITEM_DIALOG("open info item dialog"),
-    GETTING_MAIN_SCREEN_TAB("getting main screen tab");
+    GETTING_MAIN_SCREEN_TAB("getting main screen tab"),
+    PLAY_ON_POPUP("play on popup"),
+    SUBSCRIPTIONS("loading subscriptions"),;
```
AudricV:
```suggestion
    SUBSCRIPTIONS("loading subscriptions");
```

app/src/main/res/values/strings.xml
```diff
@@ -877,4 +877,9 @@
     <string name="trending_movies">Trending movies and shows</string>
     <string name="trending_music">Trending music</string>
     <string name="entry_deleted">Entry deleted</string>
+    <string name="player_http_403">HTTP error 403 received from server while playing, likely caused by an IP ban or by streaming URL expiration</string>
```
AudricV:
```suggestion
    <string name="player_http_403">HTTP error 403 received from server while playing, likely caused by streaming URL expiration or an IP ban</string>
```

AudricV (approved):

PR #12605, opened 2025-09-05
Fix and simplify opening in browser (`openUrlInBrowser`)

fynngodau:
The code was not previously working in case no default browser is set[1] AND NewPipe is set as default handler for the link in question.

We improve it by telling the system to choose the target app as if the URI was `http://`, which makes more sense than expecting a default browser to be set and NewPipe trying to query that.

[1]: also the case if platform refuses to tell an app what the user's default browser is, which I observed on CalyxOS.

#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

For technical reasons we do not add the behavior suggested in https://github.com/TeamNewPipe/NewPipe/issues/9403 as we still mirror the behavior as if `http://` is opened – unfortunately Android does not offer us a _proper_ way to selectively exclude a specific activity (i.e. ourselves) from the target list (see also my `Implementation note` comment in the `openAppChooser` code). cc @SebiderSushi

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
+++ b/app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
@@ -5,10 +5,9 @@ import static org.schabi.newpipe.MainActivity.DEBUG;
 import android.content.ActivityNotFoundException;
 import android.content.ClipData;
 import android.content.ClipboardManager;
+import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
-import android.content.pm.PackageManager;
-import android.content.pm.ResolveInfo;
 import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.Build;
@@ -23,6 +22,7 @@ import androidx.core.content.FileProvider;
 
 import org.schabi.newpipe.BuildConfig;
 import org.schabi.newpipe.R;
+import org.schabi.newpipe.RouterActivity;
 import org.schabi.newpipe.extractor.Image;
 import org.schabi.newpipe.util.image.ImageStrategy;
 import org.schabi.newpipe.util.image.PicassoHelper;
@@ -62,8 +62,9 @@ public final class ShareUtils {
     }
 
     /**
-     * Open the url with the system default browser. If no browser is set as default, falls back to
-     * {@link #openAppChooser(Context, Intent, boolean)}.
+     * Open the url with the system default browser. If no browser is installed, falls back to
+     * {@link #openAppChooser(Context, Intent, boolean)} (for displaying that no apps are available
+     * to handle the action, or possible OEM-related edge cases).
      * <p>
      * This function selects the package to open based on which apps respond to the {@code http://}
      * schema alone, which should exclude special non-browser apps that are can handle the url (e.g.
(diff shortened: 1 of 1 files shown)


Stypox (approved):
I agree that this should be the behavior we use, however I would like @AudricV to check this

Stypox:
I'm merging this for now, so people can start testing it in nightlies. Since this is a very small PR, it can always be reverted if it creates issues.

app/src/main/java/org/schabi/newpipe/util/external_communication/ShareUtils.java
```diff
@@ -190,6 +173,18 @@ private static void openAppChooser(@NonNull final Context context,
             chooserIntent.putExtra(Intent.EXTRA_TITLE, context.getString(R.string.open_with));
         }
 
+        // Avoid opening in NewPipe
+        // (Implementation note: if the URL is one for which NewPipe itself
+        // is set as handler on Android >= 12, we actually remove the only eligible app
+        // for this link, and browsers will not be offered to the user. For that, use
+        // `openUrlInBrowser`.)
```
AudricV, line 180: We should (I tested this in the Android 14 Developer Preview 2 on March 2023) be able use `Intent.EXTRA_ALTERNATE_INTENTS` (Android API 23+) to show all apps that can open a given intent even with web restrictions on Android >= 12. However, we need to have access to queries of all intents we use (so `http`, `https`, `market`, as we target Kodi's package directly):

```java
final List<ResolveInfo> resolveInfoList;

// Use PackageManager.MATCH_ALL, as we want the list of apps which can handle a web intent
// and not only the default app
// This flag doesn't include disabled or uninstalled apps with data kept (they would be not
// accessible anyway, as this requires the QUERY_ALL_PACKAGES permission)
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
    resolveInfoList = context.getPackageManager()
            .queryIntentActivities(intent,
                    PackageManager.ResolveInfoFlags.of(PackageManager.MATCH_ALL));
} else {
    resolveInfoList = context.getPackageManager()
            .queryIntentActivities(intent, PackageManager.MATCH_ALL);
}

final Intent[] intents = resolveInfoList.stream()
        .map(resolveInfo -> new Intent(Intent.ACTION_VIEW, urlUri)
                .setPackage(resolveInfo.activityInfo.packageName))
        .toArray(Intent[]::new);

// Passing an Intent which doesn't match any activity, as we don't want any other app in the
// system chooser than the ones we got earlier
// In this example an ACTION_VIEW intent with an empty URI is used
final Intent chooserIntent = createChooserIntent(
        context, new Intent(Intent.ACTION_VIEW, Uri.EMPTY), true);
// Intent.EXTRA_ALTERNATE_INTENTS extra allows to pass an unlimited amount of intents
chooserIntent.putExtra(Intent.EXTRA_ALTERNATE_INTENTS, intents);
```

Stypox: Oh yeah, I see how this could work, we're basically recreating the chooser dialog. You propose adding this just for the case where `openUrlInBrowser` fails, right? Not for the general `openAppChooser` implementation, right? Because `openAppChooser` is also used in `shareText` which is not an URL.

PR #12608, opened 2025-09-06
Issue#12443

cp90-pixel:
### Summary

- Corrected the MP4 DASH writer to capture each fragment’s base decode time (tfdt) and use it as the default media timestamp, ensuring proper alignment of combined streams

- Updated the track-building logic to apply this base timestamp when no edit list is present so playback starts at the proper point in muxed outputs

### Testing

- ⚠️ ./gradlew test *(missing Android SDK: “SDK location not found”)*

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/streams/Mp4FromDashWriter.java
+++ b/app/src/main/java/org/schabi/newpipe/streams/Mp4FromDashWriter.java
@@ -185,8 +185,12 @@ public class Mp4FromDashWriter {
             Mp4DashChunk chunk;
             while ((chunk = readers[i].getNextChunk(true)) != null) {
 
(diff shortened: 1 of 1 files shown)


Stypox:
Thank you for this PR! Could you please change the title of the PR to briefly describe what it fixes, and also, for next time, please follow the PR template ;-)

It would be good if you could add two very short (e.g. 1s) sample video + audio that were previously misbehaving, along with a unit test that tries to mux those two together and checks that the result is properly muxed? This might be hard to do, let me know if you'd prefer to skip it.

> ⚠️ ./gradlew test (missing Android SDK: “SDK location not found”)

That's an issue with your setup, you need to download the Android SDK and specify its location :-)
Or just download the APK from the CI and test it out.

PR #12426, opened 2025-07-13
Use Accompanist Drawable Painter

Isira-Seneviratne:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Load app icon using [Accompanist's Drawable Painter](https://google.github.io/accompanist/drawablepainter/).

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #12609, opened 2025-09-06
Image vector app icon

Stypox:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

Supersedes https://github.com/TeamNewPipe/NewPipe/pull/12426 by creating a Kotlin `ImageVector` for the app icon instead of relying on an external library just to render a drawable. This required remaking the app icon SVGs so that they use only as little lines as needed instead of a bunch of bezier curves (I mean, curves would have also worked, but it was a useless mess). Then the new squircle SVG was converted to `ImageVector` with https://github.com/rafaeltonholo/svg-to-compose/ based on assets/newpipe_squircle.svg.

image

#### APK testing

#### Due diligence


app/src/main/java/org/schabi/newpipe/util/image/NewPipeSquircleIcon.kt
```diff
@@ -0,0 +1,115 @@
...
+                lineTo(x = 84.0f, y = 50.0f)
+            }
+        }.build().also { _newPipeIcon = it }
+    }
+
+@Preview
+@Composable
+private fun IconPreview() {
+    org.schabi.newpipe.ui.theme.AppTheme {
+        Column(
```
Isira-Seneviratne: This column is unnecessary, as the Image is the only child.

Isira-Seneviratne:
LGTM

app/src/main/java/org/schabi/newpipe/util/image/NewPipeSquircleIcon.kt
Stypox: That's just autogenerated code, but yeah I removed it, thanks for noticing

PR #12654, opened 2025-09-21
Migrate XML-based views to Material 3

Isira-Seneviratne:
#### What is it?
- [x] Feature (user facing)

- Migrate the existing XML-based views to use Material 3.
- Clean up some of the existing theming code, as it was causing some theming issues.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

#### APK testing

#### Due diligence


Angelk90:
@Isira-Seneviratne : Nice work, why the button with the "+" (plus) in that color?

Maybe this color looks better`#6c5d2f`:
`Screenshot 2025-09-25 alle 22 25 24

Look: https://material-foundation.github.io/material-theme-builder/

Stypox:
Cool, thanks! I have one complaint: the color of the top bar serves the purpose of giving an indication of which service is currently active, and now that indication is lost. So we should either bring back the colored bar (just in MainActivity), or we find some other way to clearly indicate which service is active. Maybe a chip in the top bar replacing the tab name? And then the tab name could be put next to the tab icon (only for the selected tab):

image


> @Isira-Seneviratne : Nice work, why the button with the "+" (plus) in that color?

I guess there's something wrong in the color scheme we have right now?

HatakeKakashri:
> Cool, thanks! I have one complaint: the color of the top bar serves the purpose of giving an indication of which service is currently active, and now that indication is lost. So we should either bring back the colored bar (just in MainActivity), or we find some other way to clearly indicate which service is active. Maybe a chip in the top bar replacing the tab name? And then the tab name could be put next to the tab icon (only for the selected tab):
> 
> image
> 
> 
> > @Isira-Seneviratne : Nice work, why the button with the "+" (plus) in that color?
> 
> I guess there's something wrong in the color scheme we have right now?


That should be doable I think. We will need to generate the color pallete for each service from the Material theme builder tool. [More info](https://developer.android.com/develop/ui/compose/designsystems/material2-material3#color).

I was able to match the default (Youtube) colors in Compose using that tool in https://github.com/TeamNewPipe/NewPipe/pull/12385

PR #12671, opened 2025-09-29
Fix initial numbering of frames in TTML to SRT converter

### LINKED ISSUE #12670: [YouTube] SRT numbering should start at 1 (currently starts at 0)
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

0.28.0 and all earlier versions

### Description

Currently, into YouTube, the downloaded `.srt` subtitle files start numbering at 0. For example:

> 0
> 00:00:00,640 --> 00:00:07,160
> This mother has three young cubs just 8

Although most players (`mpv`, `VLC`, `MPlayer`, `Totem`) can still display them correctly, this is NOT compliant with the `SubRip (.srt) specification`, which requires numbering to start at 1.

Here is the specification: SubRip (https://en.wikipedia.org/wiki/SubRip), and it says:

> Subtitles are numbered sequentially, starting at 1. 

Tested players:
- mpv 0.37.0 ✅ works with 0
- VLC 3.0.20 ✅ works with 0, even -1
- MPlayer 1.5 ✅ works with 0
- Totem 43.0 ✅ works with 0

Even though playback works fine, some stricter software or tools **may reject** these `*.srt` files. So I think it's best to fix it.

### Expected behavior

_No response_

### Actual behavior

This issue can be reproduced by any downloaded `*.srt` file.
Here is an example of YouTube video link ( https://www.youtube.com/watch?v=Q8aPwoe_iMc ), let's look at  the beginning of a `*.SRT` file generated by NewPipe:

> a51x:/storage/emulated/0/Newpipe $ cat Bear\ Family\ Searches\ for\ Water\ _\ Wild\ Mexico\ _\ BBC\ Earth-en.srt               
> 0
> 00:00:00,640 --> 00:00:07,160
> This mother has three young cubs just 8
> 
> 1
> 00:00:03,840 --> 00:00:07,160
> months old.
> 
> 2
> 00:00:08,520 --> 00:00:22,899
> [Music]
> 
> 3
> 00:00:24,560 --> 00:00:29,039
> It's autumn and the family needs to
> 
> 4
> 00:00:27,119 --> 00:00:32,039
> flatten up before the winter
> 

_No response_

### Logs

_No response_

_No response_

### Affected device model

_No response_

### Additional information

_No response_

TransZAllen:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

This PR fixes the numbering in .srt subtitle files generated from YouTube streams.
Previously, numbering started at 0, which is accepted by most players but does not comply with the SubRip specification. Now numbering starts at 1, ensuring strict compliance.

* **Spec reference**: [Wikipedia: SubRip](https://en.wikipedia.org/wiki/SubRip), and it says:  

> Subtitles are numbered sequentially, starting at 1.

* **Observed behavior**: numbering started at 0.
* **Tested players**:
  * mpv 0.37.0 ✅ works with 0
  * VLC 3.0.20 ✅ works with 0, even `-1`
  * MPlayer 1.5 ✅ works with 0
  * Totem 43.0 ✅ works with 0
* Even though playback worked fine, some stricter tools may reject such files.

Why not `i++` or `++i`?

Every time I see `i++` or `++i`, I need to pause and think:

* Is it “use first, then increment” (`i++`) or “increment first, then use” (`++i`)?
* Even after I've written the code, two or three days later, when I look at the modified code, I still find myself thinking about the same issue.

To avoid this confusion, I prefer splitting the current line `writeString(String.valueOf(frameIndex++));` into two lines: one to use current value of `frameIndex`, and the other to increment it.

- Before: Here is an example of YouTube video link (https://www.youtube.com/watch?v=Q8aPwoe_iMc), let's look at the beginning of a *.SRT file generated by NewPipe:

> a51x:/storage/emulated/0/Newpipe $ cat Bear\ Family\ Searches\ for\ Water\ _\ Wild\ Mexico\ _\ BBC\ Earth-en.srt               
> 0
> 00:00:00,640 --> 00:00:07,160
> This mother has three young cubs just 8
> 
> 1
> 00:00:03,840 --> 00:00:07,160
> months old.
> 
> 2
> 00:00:08,520 --> 00:00:22,899
> [Music]
> 
> 3
> 00:00:24,560 --> 00:00:29,039
> It's autumn and the family needs to
> 
> 4
> 00:00:27,119 --> 00:00:32,039
> flatten up before the winter

- After: Now numbering start at 1
start 1  2025-09-29 18-32-00

Fixes #12670

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
+++ b/app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
@@ -24,7 +24,11 @@ public class SrtFromTtmlWriter {
     private final boolean ignoreEmptyFrames;
     private final Charset charset = StandardCharsets.UTF_8;
 
-    private int frameIndex = 0;
+    // According to the SubRip (.srt) specification, subtitle
+    // numbering must start from 1.
+    // Some players accept 0 or even negative indices,
+    // but to ensure compliance we start at 1.
+    private int frameIndex = 1;
 
     public SrtFromTtmlWriter(final SharpStream out, final boolean ignoreEmptyFrames) {
         this.out = out;
@@ -39,7 +43,8 @@ public class SrtFromTtmlWriter {
 
     private void writeFrame(final String begin, final String end, final StringBuilder text)
             throws IOException {
-        writeString(String.valueOf(frameIndex++));
+        writeString(String.valueOf(frameIndex));
+        frameIndex += 1;
         writeString(NEW_LINE);
         writeString(begin);
         writeString(" --> ");


app/src/main/java/org/schabi/newpipe/streams/SrtFromTtmlWriter.java
```diff
@@ -39,7 +43,8 @@ private static String getTimestamp(final Element frame, final String attr) {
 
     private void writeFrame(final String begin, final String end, final StringBuilder text)
             throws IOException {
-        writeString(String.valueOf(frameIndex++));
+        writeString(String.valueOf(frameIndex));
+        frameIndex += 1;
```
TobiGr, line 47: I'd prefer to use this because is just a write / increment and no read. Thus there is only little room for confusion.
```suggestion
        frameIndex++;
```

TransZAllen: Thanks for your suggestion.
I agree that `frameIndex++` works fine here and avoids confusion.

My earlier comment about `i++ / ++i` was only about using them directly inside something like `writeString(String.valueOf(...))`, which can indeed be confusing.

Between `frameIndex += 1`, `frameIndex++`, and `frameIndex = frameIndex + 1`, I personally prefer `+= 1` because it’s very explicit and leaves no room for overthinking.

I’m fine with switching to `frameIndex++` if the team prefers, but my preference would be to keep `+= 1` if that’s acceptable.

TransZAllen: Let me know if you’d like me to change it.

Stypox: Thanks! I'm fine with either

PR #12691, opened 2025-10-08
add proxy settings

### LINKED ISSUE #992: Proxy Settings
Hi, please add proxy settings

fokin33:
#### What is it?
- [x] Feature (user facing)

Add proxy settings 

- Before:
- After:
- 
image

- Fixes #992

#### APK testing

#### Due diligence


liasica:
Need this feature!

app/src/main/res/values/strings_proxy.xml
TobiGr, line 1: it is a good idea to separate the strings so that strings.xml is not cluttered too much. However, I think this should be done for all strings at the same time in a separate PR. Please make this a section in string.xml for now.

app/src/main/res/menu/main_menu.xml
TobiGr, line 1: Where does this file come from?

TobiGr (changes requested):
Thank you for this PR. Things that need to be taken into account:

1. Enabling the proxy in the settings should enable the proxy immediately or prompt the user to restart the app. Currently it looks like the proxy is enabled in the settings UI although it is actually not.

TobiGr:
Proxy for downloads is missing (see DownloadMission#openConnection)

xiva-wgt:
TobiGr and fokin33 thank you for your work, it is very important.
for countries with censorship, working with a local socks5 with login|pass, where they are trying to control whether the vpn service is enabled on android, this change is very necessary, it will allow you to watch youtube without the risk of blocking circumvention of censorship.

absurdlylongusername:
@fokin33 From what I can see, this only works for YouTube and no other services

fokin33:
> @fokin33 From what I can see, this only works for YouTube and no other services

It adds a proxy server to the default data source, so the proxy server is used for all services.

PR #12684, opened 2025-10-05
Dependency updates

TobiGr:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
This updates some dependencies to newer versions:

- `androidx.media:media: 1.7.0 -> 1.7.1` [changelog](https://developer.android.com/jetpack/androidx/releases/media#media-1.7.1)
- `androidx.viewpager2:viewpager2: 1.1.0-beta02 -> 1.1.0` [changelog](https://developer.android.com/jetpack/androidx/releases/viewpager2#version_110_3)
- `io.reactivex.rxjava3:rxjava: 3.1.8 -> 3.1.12` [changelog](https://github.com/ReactiveX/RxJava/releases)
- `org.jsoup:jsoup: 1.17.2 -> 1.21.2` [changelog](https://github.com/jhy/jsoup/releases)
- `com.puppycrawl.tools:checkstyle: 10.12.1 -> 10.26.1` [changelog](https://github.com/checkstyle/checkstyle/releases)

Updating checkstyle fixed a vulnerability and fixed a [final class check in version 10.12.2 for local classes without constructor](https://github.com/checkstyle/checkstyle/releases/tag/checkstyle-10.12.2). Local classes without a constructor should be marked as final. That is done in this PR.

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #12706, opened 2025-10-15
Build improvements

### LINKED ISSUE #10927: Migrate project dependencies to using Version Catalogs
### Checklist

### Feature description

Currently, the project manages dependencies by declaring them directly within the `build.gradle` file. This approach, while functional, can become cumbersome and difficult to maintain as the project grows and the number of dependencies increases.

This feature proposes migrating dependency management to Version Catalogs. Version catalogs are a centralized location within the Gradle build system to store and manage all project dependencies.

Version catalogs offer several advantages over declaring dependencies directly in build.gradle: https://developer.android.com/build/migrate-to-catalogs

### Additional information

_No response_

theimpulson:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Reverts non-required reproducible build hack
- Migrates build scripts to Kotlin DSL
- Migrates to use build version catalog to manage dependencies
- Introduces script to sort dependencies order

- Fixes #10927

#### APK testing

#### Due diligence
- [ x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -18,6 +18,12 @@ val gitWorkingBranch = providers.exec {
     commandLine("git", "rev-parse", "--abbrev-ref", "HEAD")
 }.standardOutput.asText.map { it.trim() }
 
+java {
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(17)
+    }
+}
+
 android {
     compileSdk = 36
     namespace = "org.schabi.newpipe"
@@ -87,14 +93,7 @@ android {
     compileOptions {
         // Flag to enable support for the new language APIs
         isCoreLibraryDesugaringEnabled = true
-
-        sourceCompatibility = JavaVersion.VERSION_17
-        targetCompatibility = JavaVersion.VERSION_17
         encoding = "utf-8"
-    }
-
-    kotlinOptions {
-        jvmTarget = "17"
     }
 
     sourceSets {


app/build.gradle.kts
```diff
@@ -0,0 +1,304 @@
...
+                .replace("[^0-9A-Za-z]+".toRegex(), "")
+
+            if (normalizedWorkingBranch.isEmpty() || workingBranch in defaultBranches) {
+                // default values when branch name could not be determined or is master or dev
+                applicationIdSuffix = ".debug"
+                resValue("string", "app_name", "NewPipe Debug")
+            } else {
+                applicationIdSuffix = ".debug.$normalizedWorkingBranch"
+                resValue("string", "app_name", "NewPipe $workingBranch")
+                setProperty("archivesBaseName", "NewPipe_$normalizedWorkingBranch")
```
theimpulson: This property has been long deprecated and doesn't seems to work anymore, further removed in Gradle 9.x. We should also look to get rid of this in favour of the [base plugin](https://docs.gradle.org/current/userguide/base_plugin.html#sec:base_plugin_extension) which has a similar property.

theimpulson: Applying the property here using the `base.archivesName` doesn't seems to work but doing so specifically in the `base { archivesName = xxx }` block below the `android` block does the job. Sadly there isn't a simple way to get the current build type in Gradle. Do we want to keep a uniform way of naming the generated APK file?

theimpulson:
I have re-written the script to sort TOML file in Kotlin DSL as well while cherry-picking it from the refactor branch.

litetex (dismissed):
as per https://github.com/TeamNewPipe/NewPipe/issues/10927#issuecomment-3407596698

We should do an independent update to the lastest Gradle version first and then have a look if it's useful to migrate everything to Kotlin

theimpulson:
> as per [#10927 (comment)](https://github.com/TeamNewPipe/NewPipe/issues/10927#issuecomment-3407596698)
> 
> We should do an independent update to the lastest Gradle version first and then have a look if it's useful to migrate everything to Kotlin

I am not sure what that comment means in relation to latest Gradle version first and Kotlin migration. I picked the TOML change as it has been merged for a long time in the refactor branch and is the current default for new android projects similar to how Kotlin is the new default DSL and language for writing new projects which we want to have with compose migration.

Do you want me to bump the Gradle version to 9.1.0 in this PR as well?

litetex:
I was expecting that the change related to `archivesBaseName` is due to Gradle 9 but it looks like I'm wrong here, so nevermind.

> Do you want me to bump the Gradle version to 9.1.0 in this PR as well?

Well you can do that but it should be done independently so that things don't get mixed up here :)

----

> I picked the TOML change as it has been merged for a long time in the refactor branch

I see... I was not around when this was merged. I think I will create a PR to revert this on the refactor branch.

> current default for new android projects

Even though Google might have declared this as the "default" it might not be very useful here.
As I pointed out in my previous comment I find it catastrophic that I have to look in 3 different places to find the version of a dependency.

theimpulson:
> Even though Google might have declared this as the "default" it might not be very useful here. As I pointed out in my previous comment I find it catastrophic that I have to look in 3 different places to find the version of a dependency.

I feel it is useful considering we do seem to want a compose multi-platform compatible NewPipe which will involve us introducing multiple modules per-platform, sooner or later, leading us to require a way to manage multiple dependencies and plugins.

Then there is also the logic of not following the changes done by Google because we don't like them leading us to have technical burden in future like today we have with Kotlin and Compose.

Stypox:
@litetex it's actually the opposite, now all versions are in a single file instead of spread over different sections of `build.gradle` and `app/build.gradle`. I am in favour of keeping version catalogs on the refactor branch, and also merging this PR. Also see https://github.com/TeamNewPipe/NewPipe/issues/10927#issuecomment-3409597431

app/build.gradle.kts
```diff
@@ -0,0 +1,304 @@
...
+    alias(libs.plugins.jetbrains.kotlin.parcelize)
+    alias(libs.plugins.sonarqube)
+    checkstyle
+}
+
+apply(from = "check-dependencies.gradle.kts")
+
+val gitWorkingBranch = providers.exec {
+    commandLine("git", "rev-parse", "--abbrev-ref", "HEAD")
+}.standardOutput.asText.map { it.trim() }
```
Stypox, line 19: I found out that Gradle has among its dependencies `JGit`, so in a future PR we could use JGit instead of manually invoking the `git` command. I did that [here](https://github.com/Stypox/dicio-android/blob/master/app/build.gradle.kts#L217-L219) in another project

app/build.gradle.kts
Stypox: Despite this I always see the debug APKs being called `app-debug.apk` so I don't think it matters here.

app/build.gradle.kts
```diff
@@ -0,0 +1,304 @@
...
+                setProperty("archivesBaseName", "NewPipe_$normalizedWorkingBranch")
+            }
+        }
+
+        release {
+            System.getProperty("packageSuffix")?.let { suffix ->
+                applicationIdSuffix = suffix
+                resValue("string", "app_name", "NewPipe $suffix")
+                setProperty("archivesBaseName", "NewPipe_" + System.getProperty("packageSuffix"))
+            }
```
Stypox, line 76: I thought it was important to keep archivesBaseName because we use it in CI. But it turns out [the filename is not used directly](https://github.com/TeamNewPipe/NewPipe-nightly/blob/github_actions/.github/workflows/nightly.yml#L111) (and also I think `archivesBaseName` doesn't actually change the apk name, even when I set `packageSuffix` I get a release apk named `app-release.apk`. So I would just remove `archivesBaseName` everywhere (and also fix `doc/gradle.md` accordingly)

Edit: oh wait, it wasn't doing anything here because there's another `setProperty("archivesBaseName", "app")` below :man_facepalming:. So yeah I would remove them all.

Also, if in the end we end up keeping it:
```suggestion
            System.getProperty("packageSuffix")?.let { suffix ->
                applicationIdSuffix = suffix
                resValue("string", "app_name", "NewPipe $suffix")
                setProperty("archivesBaseName", "NewPipe_$suffix")
            }
```

app/build.gradle.kts
```diff
@@ -0,0 +1,304 @@
...
+            excludes += setOf(
+                "META-INF/README.md",
+                "META-INF/CHANGES",
+                "META-INF/COPYRIGHT" // "COPYRIGHT" belongs to RxJava...
+            )
+        }
+    }
+}
+
+val ktlint by configurations.creating
```
Stypox, line 128: Maybe add a comment explaining what this does?

app/build.gradle.kts
```diff
@@ -0,0 +1,304 @@
...
+dependencies {
+    /** Desugaring **/
+    coreLibraryDesugaring(libs.android.desugar)
+
+    /** NewPipe libraries **/
+    implementation(libs.newpipe.nanojson)
+    // WORKAROUND: if you get errors with the NewPipeExtractor dependency, replace `v0.24.3` with
+    // the corresponding commit hash, since JitPack sometimes deletes artifacts.
+    // If there’s already a git hash, just add more of it to the end (or remove a letter)
+    // to cause jitpack to regenerate the artifact.
```
Stypox: This should reference libs.versions.toml

Stypox (changes requested):
Thank you!

How do you know the reproducible builds fix is not needed anymore?

theimpulson:
> Thank you!
> 
> How do you know the reproducible builds fix is not needed anymore?

As per the [linked issue](https://issuetracker.google.com/issues/231837768) in the original commit, the fix has been shipped in AGP 8.1.0-alpha03 and we are using way newer version of AGP now.

app/build.gradle.kts
theimpulson: Removed

app/build.gradle.kts
theimpulson: Removed it as the comment mentions to replace version with hash but it has already been using hash.

app/build.gradle.kts
theimpulson: Done. Ref: https://docs.gradle.org/current/userguide/declaring_configurations.html#sec:defining-custom-configurations

app/build.gradle.kts
theimpulson: Agreed, didn't knew about it. Looks far better than manually invoking git.

litetex:
@theimpulson 

> leading us to have technical burden in future like today we have with Kotlin and Compose

I can't quite follow on this. Could you clarify what you mean with that?

app/build.gradle.kts
Stypox: Actually we keep switching back and forth, because when a new NewPipeExtractor version is released its tag works, but then at some point we need to use a development version of NPE or Jitpack stops providing artifacts for the tag, and so we have to switch to hashes.

Stypox:
Merging for now, despite the discussion in https://github.com/TeamNewPipe/NewPipe/issues/10927#issuecomment-3412643265 still not being concluded, in order not to block @theimpulson's other PRs. Since version catalogs have already been merged to refactor and the work has already been done, it wouldn't make sense to delay it further. We can always revert it later if more team members think we are better off without version catalogs, and the effort for reverting won't change if this PR is merged or not, since we'd have to revert it on `refactor` as well.

app/build.gradle.kts
Stypox: I reverted this change in [`300f5ab` (#12684)](https://github.com/TeamNewPipe/NewPipe/pull/12684/commits/300f5abc70c081b9c09fa461b01531f5aee4b082)

app/build.gradle.kts
```diff
@@ -0,0 +1,297 @@
...
+    kapt(libs.androidx.room.compiler)
+    implementation(libs.androidx.swiperefreshlayout)
+    implementation(libs.androidx.viewpager2)
+    implementation(libs.androidx.work.runtime)
+    implementation(libs.androidx.work.rxjava3)
+    implementation(libs.google.android.material)
+    implementation(libs.androidx.webkit)
+
+    /** Third-party libraries **/
+    implementation(libs.livefront.bridge)
```
Stypox, line 230: You missed this comment :-)

```suggestion
    /** Third-party libraries **/
    // Instance state boilerplate elimination
    implementation(libs.livefront.bridge)
```

PR #12711, opened 2025-10-16
[player] Fix non-fullscreen layout regression

Ljn0626:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- Removed `binding.metadataView.setVisibility(isFullscreen ? View.VISIBLE : View.GONE);`
- Updated the layout: increased `audioTrackTextView`’s `layout_weight `to 2

#### Before/After Screenshots/Screen Record

- Before: When not in fullscreen,  all other items are no longer correctly aligned to the right side 
d0fe398db90d754ac767d660b36e93e1
d402a1fc047c57f57e53ec33e531961f

- After: It displays correctly on both small and medium phones.
f430308fb6caca7933f78743bf916420
7888971cefa2933987f1eea5b2f94b3d
1490d4a0e37940392e841096ceb76e0e
f671e3ac3c7eb4c3380631105b8a8685
e26ffb8310ee0e41b94fafaf0244aaaf

#### Fixes the following issue(s)

- Follow-up to Issue#12499

#### Relies on the following changes

- `android:layout_weight="2"` ->`android:layout_weight="2"`
- delete `binding.metadataView.setVisibility(isFullscreen ? View.VISIBLE : View.GONE);`

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [X] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #12699, opened 2025-10-11
Fix Long Audio/Dubs text label puses UI Controls on Player Off Screen in Portrait mode. 

Ljn0626:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

Constrain audioTrackTextView width (maxWidth="110dp") and single line + ellipsize="end" + scrollHorizontally="true", so it never wraps or steals too much space.

#### Before/After Screenshots/Screen Record

- Before: With long English/Spanish labels, the More options button could be pushed off-screen
3b1144a61388ff6c1bd9799c790d77b4
bcc59b3d652669fb7cdc6a1da62c89ef

- After: Right-side controls remain visible; the audio track label stays on a single line with ellipsis
image
image

#### Fixes the following issue(s)

- Fixes #12499 

#### Relies on the following changes

- `android:singleLine="true"`
    `android:ellipsize="end"`
    `android:scrollHorizontally="true"`
   `android:maxWidth="110dp"`

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [X] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

Issue #12499, opened 2025-08-03
Long Audio/Dubs text label pushes  UI Controls on Player Off Screen in Portrait mode.


SavSanta:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Affected version

28.0 (but also reproducible in random debug build 27.7 NewPipe fix-fullscreen-clear-queue-prompt)  

### Steps to reproduce the bug

1. Search and select a video that NewPipe will with a long audio/dub description.

2. Play the video in portrait mode.

3. Notice the UI controls. In particular the downward caret symbol that upon clicking reveals extended controls ("window zoom", "captions", and "share" and mute buttons)

4. Optionally change the video to Landscape to see the downward caret and rest of controls

### Expected behavior

The audio/dubs display component to not to push off wcreen the other UI controls.

### Actual behavior

It pushes off the other UI controls

### Affected Android/Custom ROM version

Android 11, Android 12

SavSanta:
Trying to add photos on mobile was a pain using Google Photos built-in. Had to switch to Explorer

<img alt="Image" src="https://github.com/user-attachments/assets/709cc5ce-66f5-44ac-b045-72f019a83176" width=276 />

Portrait mode you can see the caption title "English United States (original)" pushes the UI controls far over. If you look closely you can see on the right edge after the video rewolution  control, a small tinge of white which is the UI control for  the video speed.

![Image](https://github.com/user-attachments/assets/118e8e9f-1206-4e5f-8542-5439ca693fe0)

This is the video in landscape mode for better reference of the missing controls.

RobindaCoast:
Having the same issue here too. 
Samsung J7+ running on android 8.1.0 for reference.

[...]

PR #12714, opened 2025-10-17
Improve/Fix primaryControls layout in player

litetex:
#### What is it?
- [x] Bugfix (user facing)
- [ ] Feature (user facing)

* Supersedes https://github.com/TeamNewPipe/NewPipe/pull/12711 
* Fixes the regression introduced in https://github.com/TeamNewPipe/NewPipe/pull/12699 / #12499
* Doesn't require code changes or hard-coded limits to properly display the elements

- Before: See https://github.com/TeamNewPipe/NewPipe/pull/12711
- After: 

##### About the Popup Player

The popup player is still broken but it was broken before (basically since the introduction of `audioTrackTextView` a few years ago)
grafik

I tried a few quick/dirty fixes - like limiting the maxWidth of `audioTrackTextView` - however the results were worse than before so I aborted doing any changes there for now.

This can likely only be fixed with a major rework how the `topControls` are handled in the player.
There should be different fragments, one for the main player (that basically represents the current layout) and one for popup (where everything is next to each other). This way the spacing can be done correctly. It would also make the player UI a lot cleaner.

Maybe we can also sit this problem out / it's fixed with refactor/NewPlayer...

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/res/layout/player.xml
+++ b/app/src/main/res/layout/player.xml
@@ -167,7 +167,7 @@
 
                         <FrameLayout
                             android:layout_width="wrap_content"
-                            android:layout_height="match_parent"
+                            android:layout_height="wrap_content"
                             android:layout_weight="1">
 
                             <org.schabi.newpipe.views.NewPipeTextView


Stypox:
Thanks!

> The popup player is still broken but it was broken before (basically since the introduction of audioTrackTextView a few years ago)

Yeah it's the same change that also broke the main player, and https://github.com/TeamNewPipe/NewPipe/pull/12699 was the first attempt to fix it.

> I tried a few quick/dirty fixes - like limiting the maxWidth of audioTrackTextView - however the results were worse than before so I aborted doing any changes there for now.

Doesn't just setting the max width of `audioTrackTextView` to something very small (e.g. 48dp) fix it? Or are there more layout problems then?

litetex:
> Doesn't just setting the max width of `audioTrackTextView` to something very small (e.g. 48dp) fix it? Or are there more layout problems then?

Nope, because:
1. When the popup player is small it still grows and pushes the full screen out of the window <br/>grafik
2. When the popup player is large it stops growing, despite having space left <br/> grafik

I also tried things like `match_parent` or `0dp` as a width but this doesn't seem to work either.

Stypox:
Yeah that's what I'd have expected, but for now it's still better than having a huge button always pushing the other buttons away

litetex:
> Yeah that's what I'd have expected, but for now it's still better than having a huge button always pushing the other buttons away

Anyway can we get this fix reviewed/merged so that we have at least a working main player.
~I can make a separate PR that "fixes" the popup player.~

PR with a maxWidth set for popup player: https://github.com/TeamNewPipe/NewPipe/pull/12729

PR #12776, opened 2025-11-08
Update all included dependencies

theimpulson:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Updates all possible dependencies to their latest stable release
- Turns on new Gradle configuration cache

#### APK testing

#### Due diligence

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ captures/
 *.iml
 *~
 .weblate
+.kotlin
 *.class
 app/debug/
 app/release/


theimpulson:
I will make a separate PR to enable new ktlint rules as well as to enable android's default lint which is currently set to ignore errors.

theimpulson:
Some of the dependencies updated here such as kotlin are already updated in refactor branch but not all. I am hoping to have all dependencies on latest stable release as they might contain fixes for compose, kotlin or ksp in general.

app/build.gradle.kts
```diff
@@ -22,6 +22,14 @@ java {
     }
 }
 
+kotlin {
```
Stypox, line 26: Maybe add a `// TODO fix all annotations and remove this block`

app/src/main/java/org/schabi/newpipe/database/history/dao/HistoryDAO.kt
```diff
@@ -9,5 +9,5 @@ package org.schabi.newpipe.database.history.dao
 import org.schabi.newpipe.database.BasicDAO
 
 interface HistoryDAO<T> : BasicDAO<T> {
-    val latestEntry: T
```
Stypox: Oh nice catch, this indeed needs to be nullable. While you are at it, could you remove `HistoryDAO` altogether, and put `latestEntry: SearchHistoryEntry?` directly in `SearchHistoryDAO`? The other override in `StreamHistoryDAO` is unused, probably will never be used,  is still not nullable and will thus crash on an empty history, and finally it doesn't make sense to have a common subclass like this.

gradle/libs.versions.toml
Stypox, line 1: I am a bit worried by all of these upgrades at once. Have you gone through the various changelogs to see if there's anything to worry about? In particular, I'm worried about `exoplayer`, `okhttp`.

Stypox (changes requested):
Thank you for the PR! Usually when we do dependency updates we try to look into the changelogs to make sure there are no regressions, I left a comment about this below.

----

Uff what a shame that androidx is dropping support for API 21... we can keep an outdated library for some time, but we'll have to switch at some point I guess. We'll lose 0.6% of devices (see below) when bumping to API 23 (or possibly more since many users install newpipe on old phones). For now, please add a comment to the two versions that we can't update in `libs.versions.toml`, so we don't forget.

Image

----

I tested the debug APK for some time and everything seems to still work, except for a subtle bug: when importing a database which contains a channel group feed in the main page, its icon turns into three dots. This does not happen on nightly.

Image

app/build.gradle.kts
theimpulson: Done

app/src/main/java/org/schabi/newpipe/database/history/dao/HistoryDAO.kt
theimpulson: Done

theimpulson:
> For now, please add a comment to the two versions that we can't update in libs.versions.toml, so we don't forget.

Done. I think we can bump up the minSdk after the next release to import compose libs as that has more frequent releases every month and are part of androidx framework too which will make it a blocker.

> I tested the debug APK for some time and everything seems to still work, except for a subtle bug: when importing a database which contains a channel group feed in the main page, its icon turns into three dots. This does not happen on nightly.

I am unable to reproduce this on my side. is it possible to change the feed icon manually? Maybe there is something related to that?

gradle/libs.versions.toml
theimpulson: I have. In particular exoplayer has been deprecated with replacement being media3 which we already are using the latest version. okhttp's major change was support for happy eyeballs and improving kotlin APIs which aren't breaking ones. The most major change was with ktlint which has a separate commit disabling all the new rules to enable them one by one in different PR.

theimpulson:
I noticed a crash during exporting database due to checkpoint being called on main thread so made a commit to fix that. This probably should have been part of the database migration to kotlin PR.

theimpulson:
Marking as draft to allow database related fixes go first to avoid polluting this PR

gradle/libs.versions.toml
Stypox: Ok, thanks!

Stypox:
> I think we can bump up the minSdk after the next release to import compose libs as that has more frequent releases every month and are part of androidx framework too which will make it a blocker.

I'd say let's not bump it as long as it doesn't start creating problems, so we can keep supporting those users for a bit longer

Stypox:
> I am unable to reproduce this on my side. is it possible to change the feed icon manually? Maybe there is something related to that?

Ah I see, when a feed group tab gets created, its icon is saved to disk as the drawable resource ID (see [here](https://github.com/TeamNewPipe/NewPipe/blob/42f909936b61b2685c027d454e66c8adfca2cc3e/app/src/main/java/org/schabi/newpipe/settings/SelectFeedGroupFragment.java#L124)). The changes in this PR likely changed the IDs, hence the problem when importing data from older versions. Since every `FeedGroupIcon` has a unique ID, we should have save that instead of the drawable ID. We have two options to solve this issue:
- creating a `SettingsMigration` that reads all main screen tabs (`R.string.saved_tabs_key`) and uses the `FeedGroupTab.feedGroupId` to read the corresponding feed group in the DB and obtain its icon ID again, except that this time the unique icon ID is saved back in the tab information instead of the drawable ID. Then when showing the tab we need to look through the `FeedGroupIcon`s to find the corresponding drawable ID. We also need to adapt the tab creation code.
- dropping the `iconId` field completely from the `FeedGroupTab` and instead reading the icon ID from the database using `feedGroupId` every time in `getTabIconRes()`. This would eliminate the redundant information (which becomes outdated if one changes a feed group icon), but may be difficult to implement since `getTabIconRes()` is not `suspend`.

This can be done in a separate PR but needs to be done before the release.

.kotlin/sessions/kotlin-compiler-4222484878062452975.salive
Stypox, line 1: .kotlin needs to be added to gitignore

theimpulson: Done

theimpulson:
I rebased to update Gradle version to 9.2.1 and removed the kotlin compiler generated files. The build errors are now due to jitpack behaving weird as always, probably can be ignored.

theimpulson:
> Since every FeedGroupIcon has a unique ID, we should have save that instead of the drawable ID.

It seems to me that we are already doing that [here](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/database/Converters.kt#L44) and mapping it properly as well.

theimpulson:
I see a different error in logcat but the icon is properly loaded in my case regardless even after exporting and importing database from current dev to this one.

```shell
2025-11-21 20:39:05.462  6431-6431  ResourcesCompat         org.schabi.newpipe.debug.depUpdate   W  Failed to inflate ColorStateList, leaving it to the framework
                                                                                                    java.lang.UnsupportedOperationException: Failed to resolve attribute at index 0: TypedValue{t=0x2/d=0x7f04012b a=-1}, theme={InheritanceMap=[id=0x7f130177org.schabi.newpipe.debug.depUpdate:style/LightDialogMinWidthTheme, id=0x7f1302ddorg.schabi.newpipe.debug.depUpdate:style/Theme.AppCompat.DayNight.Dialog.MinWidth, id=0x7f1302e9org.schabi.newpipe.debug.depUpdate:style/Theme.AppCompat.Light.Dialog.MinWidth, id=0x7f13005dorg.schabi.newpipe.debug.depUpdate:style/Base.Theme.AppCompat.Light.Dialog.MinWidth, id=0x7f13005aorg.schabi.newpipe.debug.depUpdate:style/Base.Theme.AppCompat.Light.Dialog, id=0x7f1300aeorg.schabi.newpipe.debug.depUpdate:style/Base.V21.Theme.AppCompat.Light.Dialog, id=0x7f1300d2org.schabi.newpipe.debug.depUpdate:style/Base.V7.Theme.AppCompat.Light.Dialog, id=0x7f130058org.schabi.newpipe.debug.depUpdate:style/Base.Theme.AppCompat.Light, id=0x7f1300c6org.schabi.newpipe.debug.depUpdate:style/Base.V28.Theme.AppCompat.Light, id=0x7f1300bforg.schabi.newpipe.debug.depUpdate:style/Base.V26.Theme.AppCompat.Light, id=0x7f1300b9org.schabi.newpipe.debug.depUpdate:style/Base.V23.Theme.AppCompat.Light, id=0x7f1300b7org.schabi.newpipe.debug.depUpdate:style/Base.V22.Theme.AppCompat.Light, id=0x7f1300adorg.schabi.newpipe.debug.depUpdate:style/Base.V21.Theme.AppCompat.Light, id=0x7f1300d1org.schabi.newpipe.debug.depUpdate:style/Base.V7.Theme.AppCompat.Light, id=0x7f13019eorg.schabi.newpipe.debug.depUpdate:style/Platform.AppCompat.Light, id=0x7f1301a9org.schabi.newpipe.debug.depUpdate:style/Platform.V25.AppCompat.Light, id=0x1030241android:style/Theme.Material.Light.NoActionBar, id=0x1030237android:style/Theme.Material.Light, id=0x103000candroid:style/Theme.Light, id=0x1030005android:style/Theme], Themes=[org.schabi.newpipe.debug.depUpdate:style/LightDialogMinWidthTheme, forced, org.schabi.newpipe.debug.depUpdate:style/PreferenceThemeOverlay, not forced, org.schabi.newpipe.debug.depUpdate:style/LightSettingsTheme, forced, org.schabi.newpipe.debug.depUpdate:style/OpeningTheme, forced, org.schabi.newpipe.debug.depUpdate:style/Theme.AppCompat.Empty, forced, android:style/Theme.DeviceDefault.Light.DarkActionBar, forced]}
                                                                                                    	at android.content.res.TypedArray.getColor(TypedArray.java:536)
                                                                                                    	at androidx.core.content.res.ColorStateListInflaterCompat.inflate(ColorStateListInflaterCompat.java:157)
                                                                                                    	at androidx.core.content.res.ColorStateListInflaterCompat.createFromXmlInner(ColorStateListInflaterCompat.java:122)
                                                                                                    	at androidx.core.content.res.ColorStateListInflaterCompat.createFromXml(ColorStateListInflaterCompat.java:102)
                                                                                                    	at androidx.core.content.res.ResourcesCompat.inflateColorStateList(ResourcesCompat.java:259)
                                                                                                    	at androidx.core.content.res.ResourcesCompat.getColorStateList(ResourcesCompat.java:234)
                                                                                                    	at androidx.core.content.ContextCompat.getColorStateList(ContextCompat.java:516)
                                                                                                    	at androidx.appcompat.content.res.AppCompatResources.getColorStateList(AppCompatResources.java:46)
                                                                                                    	at com.google.android.material.resources.MaterialResources.getColorStateList(MaterialResources.java:64)
                                                                                                    	at com.google.android.material.imageview.ShapeableImageView.<init>(ShapeableImageView.java:110)
                                                                                                    	at com.google.android.material.imageview.ShapeableImageView.<init>(ShapeableImageView.java:88)
                                                                                                    	at java.lang.reflect.Constructor.newInstance0(Native Method)
                                                                                                    	at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
                                                                                                    	at android.view.LayoutInflater.createView(LayoutInflater.java:743)
                                                                                                    	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:895)
                                                                                                    	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:850)
                                                                                                    	at android.view.LayoutInflater.rInflate(LayoutInflater.java:1012)
                                                                                                    	at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:973)
                                                                                                    	at android.view.LayoutInflater.inflate(LayoutInflater.java:571)
                                                                                                    	at android.view.LayoutInflater.inflate(LayoutInflater.java:462)
```

Stypox:
> Failed to inflate ColorStateList, leaving it to the framework

This happens normally, it's an internal warning of one of the libraries, and it clutters logcat but there's no way to avoid it.

> It seems to me that we are already doing that [here](https://github.com/TeamNewPipe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/database/Converters.kt?rgh-link-date=2025-11-21T12%3A36%3A01Z#L44) and mapping it properly as well.

Mmmh I don't think that function gets invoked when handing tabs. I am talking about the [FeedGroupTab](https://github.com/TeamNewPipe/NewPipe/blob/2ba87f79795daf7b1408dbbc3741bf6d76543abc/app/src/main/java/org/schabi/newpipe/settings/tabs/Tab.java#L658), which stores the icon ID and saves it data as JSON in shared preferences. As you can see [here](https://github.com/TeamNewPipe/NewPipe/blob/42f909936b61b2685c027d454e66c8adfca2cc3e/app/src/main/java/org/schabi/newpipe/settings/SelectFeedGroupFragment.java#L124) when the `FeedGroupTab` is created its icon is initialized with `entry.getIcon().getDrawableResource()`. But the correct way would be to not pass an icon at all to `FeedGroupTab`, and instead let it obtain the icon at runtime by querying the database for `feedGroupId`.

Stypox (approved):
Thanks! Let's merge this in the meantime. But the migration to a proper `FeedGroupTab` icon handling should be done asap, before 0.28.1.

Stypox:
Now when ktlint automatically fixes some style errors, the build fails. In the past it used to fail only when there were style errors that could not be automatically fixed, which I believe was a better behavior. Could you investigate why this is happening @theimpulson?

theimpulson:
> Now when ktlint automatically fixes some style errors, the build fails. In the past it used to fail only when there were style errors that could not be automatically fixed, which I believe was a better behavior. Could you investigate why this is happening @theimpulson?

I will take a look, thanks for noticing it.

PR #12862, opened 2025-12-02
UPDATE: Replace 2k with 1440p

guymarshall:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Typo correction

- replace "2k" with "2.5k" and "2K" with "2.5K" to fix common mistake
- The "k" in a resolution is a rough estimation of the longest dimension e.g. 5k is 5120*2880, 4k is 3840*2160, 2.5k is 2560*1440, 2k is 1920*1080 and so on. There is a common misconception that 2560*1440 or 1440p is actually 2k but it isn't.

- Before:
- After:

#### Due diligence


TobiGr:
hm. displaying 2.5K might irritate the users. I think we should check in the extractor if we can find streams with the correct resolution. It makes no sense to me to use resolution descriptions that are different from the ones the services use.

guymarshall:
Good points. Maybe we label it as 1440p like on the desktop website?
Screenshot from 2025-12-03 17-56-45

TobiGr:
Sounds good

PR #12944, opened 2025-12-26
Convert newpipe/util/FilenameUtils.java to kotlin

ghost:
#### What is it?
- [ ] Bugfix (user facing)

- Converted to kotlin
- Simplified `if-else` statements to `when`
- Simplified handling of default value for `selectedCharset`
- Removed unnecessary (thanks to kotlin) string conversion to patterrn

#### APK testing

#### Due diligence


app/src/main/java/org/schabi/newpipe/util/FilenameUtils.kt
```diff
@@ -0,0 +1,57 @@
...
+    fun createFilename(context: Context, title: String): String {
+        val sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context)
+
+        val charsetLd = context.getString(R.string.charset_letters_and_digits_value)
+        val charsetMs = context.getString(R.string.charset_most_special_value)
+        val defaultCharset = context.getString(R.string.default_file_charset_value)
+
+        val replacementChar: String = sharedPreferences.getString(
+            context.getString(R.string.settings_file_replacement_character_key), "_"
+        )!!
```
theimpulson:
```suggestion
        val replacementChar = sharedPreferences.getString(
            context.getString(R.string.settings_file_replacement_character_key)
        ) ?: "_"
```

app/src/main/java/org/schabi/newpipe/util/FilenameUtils.kt
```diff
@@ -0,0 +1,57 @@
...
+     * Create a valid filename.
+     *
+     * @param title             the title to create a filename from
+     * @param invalidCharacters patter matching invalid characters
+     * @param replacementChar   the replacement
+     * @return the filename
+     */
+    private fun createFilename(
+        title: String, invalidCharacters: String,
+        replacementChar: String
```
theimpulson: Please put all parameters on a single line or separate.

app/src/main/java/org/schabi/newpipe/util/FilenameUtils.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        return createFilename(title, charset, Matcher.quoteReplacement(replacementChar))
+    }
+
+    /**
+     * Create a valid filename.
+     *
+     * @param title             the title to create a filename from
+     * @param invalidCharacters patter matching invalid characters
+     * @param replacementChar   the replacement
```
theimpulson: No extra whitespace between param doc-comments please. Same for above method.

app/src/main/java/org/schabi/newpipe/util/FilenameUtils.kt
```diff
@@ -0,0 +1,57 @@
...
+        val charsetLd = context.getString(R.string.charset_letters_and_digits_value)
+        val charsetMs = context.getString(R.string.charset_most_special_value)
+        val defaultCharset = context.getString(R.string.default_file_charset_value)
+
+        val replacementChar: String = sharedPreferences.getString(
+            context.getString(R.string.settings_file_replacement_character_key), "_"
+        )!!
+        val selectedCharset = sharedPreferences.getString(
+            context.getString(R.string.settings_file_charset_key), ""
+        )!!.ifEmpty { defaultCharset }
```
theimpulson: It should be okay to work with null values here as when will fallback to selectedCharsed in next line.

app/src/main/java/org/schabi/newpipe/util/FilenameUtils.kt
ghost: Unfortunately the second arg is required so you can't omit it...

app/src/main/java/org/schabi/newpipe/util/FilenameUtils.kt
ghost: 1. The same thing here. You can't omit second argument
2. `ifEmpty` also catches situation when`context.getString(R.string.settings_file_charset_key)` is empty string not because it is not defined but because it is what was retrieved. That's how it was in original and that is what I preserved. I am not 100% sure that it can happen but if it happens the behavior (at least in this case) should be unchanged

app/src/main/java/org/schabi/newpipe/util/FilenameUtils.kt
ghost: But actually it is already solved in the codebase with extension function. I will use it

PR #12987, opened 2026-01-01
Prevent ErrorPanelHelper from leaking memory

ghost:
#### What is it?
- [x] Bugfix (user facing)

Prevent `ErrorPanelHelper` from leaking memory

It is created as field in `BaseStateFragment` but at the same time it stores reference to that `BaseStateFragment` which wasn't nulled everywhere which caused memory leaks

How it was detected:
1. Enable leakcanary
2. Search and add external playlist (just to be able to go somewhere from main fragment)
3. Close app
4. Open app
5. Click on every tab of main page
6. Click on playlists tab of main page
7. Click on playlist
8. Wait till android decides to destroy `MainFragment` and fragments of each tab
9. See that **at least** `BookmarkFragment` and `SubscriptionFragment` will leak `ErrorPanelHelper`

For me it seems a bit logical why the leak happens. But there comes quite big question about this PR... ***at least*** from above means that I couldn't force other inheritors of `BaseStateFragment` to leak which means that they solved the problem in some way but how??? If other inheritors solved the problem so probably this PR is wrong in its nature. Or maybe it could mean that those solutions could be eliminated??? Idk...

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
+++ b/app/src/main/java/org/schabi/newpipe/error/ErrorPanelHelper.kt
@@ -22,23 +22,40 @@ class ErrorPanelHelper(
     rootView: View,
     onRetry: Runnable?,
 ) {
-    private val context: Context = rootView.context!!
+    private var _context: Context? = null
+    private var _errorPanelRoot: View? = null
+    private var _errorTextView: TextView? = null
+    private var _errorServiceInfoTextView: TextView? = null
+    private var _errorServiceExplanationTextView: TextView? = null
+    private var _errorActionButton: Button? = null
+    private var _errorRetryButton: Button? = null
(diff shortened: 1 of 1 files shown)


theimpulson:
You can migrate it to use ViewBinding to avoid creating and dealing with so many variables

ghost:
> You can migrate it to use ViewBinding to avoid creating and dealing with so many variables

I've thought about it but I would need to receive inflater from somewhere which I am not sure where to get and will it be safe or not...

EDIT: passing it from the inheritor deep to BaseStateFragment seems not very pleasant too because it would require to change every inheritor...

theimpulson:
> EDIT: passing it from the inheritor deep to BaseStateFragment seems not very pleasant too because it would require to change every inheritor...

In that case it is better to let it remain as it is. It will be eventually removed with migration to compose this year.

PR #13015, opened 2026-01-07
Fix : make peertube link clickable 

iampopovich:
#### What is it?
- [x] Bugfix (user facing)

This pull request updates the PeerTube instance help text in the settings to make the instance list URL clickable and improves its display. The change enhances user experience by allowing users to directly access the instance list from the help text.

**UI/UX improvements:**

* The help text in `PeertubeInstanceListFragment` now displays the instance list URL as a clickable HTML link, and enables link navigation using `LinkMovementMethod`.
* Removed the `android:autoLink="web"` attribute from `fragment_instance_list.xml` since the link handling is now managed programmatically.

**Code updates:**

* Imported `Html` and `LinkMovementMethod` in `PeertubeInstanceListFragment.java` to support rich text and clickable links.

- Before:
- After:
- https://github.com/user-attachments/assets/363d4faa-39b0-40c1-9f02-f4eef2f4bf41

i didn't find any bug related to this problem

#### APK testing

you can see the result of this PR on the video above. i've tested my own apk already

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/PeertubeInstanceListFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/PeertubeInstanceListFragment.java
@@ -4,9 +4,10 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.os.Bundle;
-import android.text.Html;
 import android.text.InputType;
 import android.text.method.LinkMovementMethod;
+
+import androidx.core.text.HtmlCompat;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
@@ -89,7 +90,8 @@ public class PeertubeInstanceListFragment extends Fragment {
         final String instanceListUrl = getString(R.string.peertube_instance_list_url);
         final String helpText = getString(R.string.peertube_instance_url_help,
                 "<a href=\"" + instanceListUrl + "\">" + instanceListUrl + "</a>");
-        binding.instanceHelpTV.setText(Html.fromHtml(helpText, Html.FROM_HTML_MODE_LEGACY));
+        binding.instanceHelpTV.setText(HtmlCompat.fromHtml(helpText,
+                HtmlCompat.FROM_HTML_MODE_LEGACY));
         binding.instanceHelpTV.setMovementMethod(LinkMovementMethod.getInstance());
 
         binding.addInstanceButton.setOnClickListener(v -> showAddItemDialog(requireContext()));


app/src/main/java/org/schabi/newpipe/settings/PeertubeInstanceListFragment.java
```diff
@@ -84,8 +87,13 @@ public void onViewCreated(@NonNull final View rootView,
                               @Nullable final Bundle savedInstanceState) {
         super.onViewCreated(rootView, savedInstanceState);
 
-        binding.instanceHelpTV.setText(getString(R.string.peertube_instance_url_help,
-                getString(R.string.peertube_instance_list_url)));
+        final String instanceListUrl = getString(R.string.peertube_instance_list_url);
+        final String helpText = getString(R.string.peertube_instance_url_help,
+                "<a href=\"" + instanceListUrl + "\">" + instanceListUrl + "</a>");
+        binding.instanceHelpTV.setText(HtmlCompat.fromHtml(helpText,
+                HtmlCompat.FROM_HTML_MODE_LEGACY));
```
theimpulson, line 94: Just one question, why legacy and not compat?

theimpulson (approved):
Looks good to me. Thank you.

app/src/main/java/org/schabi/newpipe/settings/PeertubeInstanceListFragment.java
```diff
@@ -84,8 +87,13 @@ public void onViewCreated(@NonNull final View rootView,
...
+        binding.instanceHelpTV.setMovementMethod(LinkMovementMethod.getInstance());
```
AudricV, line 95: Note that if we use this way, if you set NewPipe as default app of one of the PeerTube instances registered in the Android manifest and if the app is able to open base instance URLs in the future, they will open in NewPipe, which is not what we want.

A better approach would be to use a custom `LinkMovementMethod`, similar to what's done in the text linkifier app classes, which would call `ShareUtils.openInBrowser`.

Also, sorry if you didn't do so, did you use AI to generate this PR description, while that's forbidden? Its tone seems that's the case to me.

iampopovich: Thanks for your review !
I'll implement the better approach in the next few days

> did you use AI to generate this PR description,

I did , my apologies if it's forbidden. I just tried to save a couple of minutes on writing the description, cause the change is trivial in my opinion, so I just clicked a copilot button to generate the description automatically.
I'm gonna rewrite it soon as well 🙏

PR #9642, opened 2023-01-08
Fixed a bug that prevented the display of multiple empty playlists

Jared234:
#### What is it?
- [X] Bugfix (user facing)
- [ ] Feature (user facing)
- [ ] Codebase improvement (dev facing)
- [ ] Meta improvement to the project (dev facing)

#### Description of the changes in your PR

- The issue was fixed by changing the faulty query `getPlaylistMetadata`:
If a playlist has no streams, there won’t be any rows in the `PLAYLIST_STREAM_JOIN_TABLE` table that have a foreign key to that playlist. Thus, the `JOIN_PLAYLIST_ID` will not have a corresponding value in any rows of the join table. So, if you group by the `JOIN_PLAYLIST_ID`, only playlists that contain videos are grouped and displayed.

#### Fixes the following issue(s)

- Fixes #8582 

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR.

#### Due diligence
- [X] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).

PR #13055, opened 2026-01-13
Fixed a bug that prevented the display of multiple empty playlists

### LINKED ISSUE #2758: Only one emply playlist is shown

ghost:
#### What is it?
- [x] Bugfix (user facing)

Turned out this bug was already fixed two years ago but unfortunately
it made its way back again. It was solved in https://github.com/TeamNewPipe/NewPipe/pull/9642 but back then another
method was used for querying playlists from db (for add playlist dialog) then in
https://github.com/TeamNewPipe/NewPipe/commit/ef4a6238c88f26e50ba794caa071432a08da66e7 was introduced another method
which had the same problem as fixed in https://github.com/TeamNewPipe/NewPipe/pull/9642 and which eventually replaced
original method in code for querying playlists from db (for add playlist dialog)

- Fixes #2758

#### APK testing

#### Due diligence
- [*] I tested the changes using an emulator or a physical device.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/database/playlist/dao/PlaylistStreamDAO.kt
+++ b/app/src/main/java/org/schabi/newpipe/database/playlist/dao/PlaylistStreamDAO.kt
@@ -68,6 +68,11 @@ interface PlaylistStreamDAO : BasicDAO<PlaylistStreamEntity> {
     )
     fun getOrderedStreamsOf(playlistId: Long): Flowable<MutableList<PlaylistStreamEntry>>
 
+    // If a playlist has no streams, there won’t be any rows in the **playlist_stream_join** table
+    // that have a foreign key to that playlist. Thus, the **playlist_id** will not have a
+    // corresponding value in any rows of the join table. So, if you group by the **playlist_id**,
+    // only playlists that contain videos are grouped and displayed. Look at #9642 #13055
+
     @Transaction
     @Query(
         """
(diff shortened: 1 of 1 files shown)


ghost:
@Stypox Hello! Look at it https://github.com/TeamNewPipe/NewPipe/commit/e44d09208c233fe52e11e7a811f2455584cfd244 seems like my PR remades change which was already merge long time ago but for some reason was then reverted.... So bug was reintroduced. What would you say? I wanted to find who made that SQL but found that years ago someone already did what I did in this PR and that it was merged (by you)...

~So it means that change itself not containing mistake? PR and commit are formatted bad but they could be just changed to "reintroduce <commit>"?~

ghost:
@Stypox Sorry first description was not accurate. I a bit messed up methods. Firstly there was only one method which was used for querying playlists.  But then in https://github.com/TeamNewPipe/NewPipe/commit/ef4a6238c88f26e50ba794caa071432a08da66e7 you added new similar method but with old mistake and then overtime the second method was used...

TobiGr:
We should add a comment so that this is not reverted again

ghost:
> We should add a comment so that this is not reverted again

The only note I can imagine is to just copy description of #9642, modify to accommodate current state and put it as comment inside file. Something like:

> If a playlist has no streams, there won’t be any rows in the **playlist_stream_join** table that have a foreign key to that playlist. Thus, the **playlist_id** will not have a corresponding value in any rows of the join table. So, if you group by the **playlist_id**, only playlists that contain videos are grouped and displayed. Look at #9642 #13055

Is that sufficient?

PR #13131, opened 2026-01-27
TabsJsonHelper refactor

ghost:
#### What is it?

- Utilize java streams
- Make one parameter non null

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/settings/tabs/TabsJsonHelper.java
+++ b/app/src/main/java/org/schabi/newpipe/settings/tabs/TabsJsonHelper.java
@@ -9,8 +9,9 @@ import com.grack.nanojson.JsonParserException;
 import com.grack.nanojson.JsonStringWriter;
 import com.grack.nanojson.JsonWriter;
 
-import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
 /**
  * Class to get a JSON representation of a list of tabs, and the other way around.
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/schabi/newpipe/settings/tabs/TabsManager.java
```diff
@@ -28,7 +28,7 @@ public static TabsManager getManager(final Context context) {
     }
 
     public List<Tab> getTabs() {
-        final String savedJson = sharedPreferences.getString(savedTabsKey, null);
+        final String savedJson = sharedPreferences.getString(savedTabsKey, "");
```
theimpulson: Why was this made non-null?

ghost: I am not sure why you ask. Did you notice that `getTabsFromJson` which then gets `savedJson` was changed to receive not null values? Why? I think it is more clean.... And by the way there is a ckeck inside it if parameter is empty string anyway which means that we can use `""` as default value as it is equal to null check in this situation. Having two equal values `""` and null seems needless?

theimpulson: Please add a null check here instead of passing empty string then. Unless the existing string is empty, null should be kept as they are there for a reason to indicate the value is missing from the system.

theimpulson (approved):
Looks good to me

PR #13137, opened 2026-01-29
Remove dead code from info_list/InfoItemBuilder

ghost:
#### What is it?

- Remove dead code from `info_list/InfoItemBuilder`
It no longer really builds any view and used only for storing click gesture callbacks. In the same way like `local/LocalItemBuilder` does

Last usage of build functions: 2e9a860aaa138f9682d363eece897506b823e340

- Also convert class to kotlin to remove boilerpolate

#### APK testing

#### Due diligence


TobiGr:
I'd suggest to rename it to something more fitting as it isn't a builder anymore. `InfoItemClickListener` might be a better name.

ghost:
> I'd suggest to rename it to something more fitting as it isn't a builder anymore. `InfoItemClickListener` might be a better name.

I also though about it but as I said we have one more "Builder" which holds only listeners too. And the listeners are static IIRC so these two "builders" probably could be eliminated at all... So I decided to leave it as builders so they are giving you more pain when you look at them so easier to spot... Very weird but...

PR #13158, opened 2026-01-31
Run checkstyle for extractor when it is built from source

ghost:
#### What is it?

- Now when you build newpipe with custom extractor it will run checkstyle for extractor automatically

#### APK testing

#### Due diligence

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -130,7 +130,6 @@ ksp {
     arg("room.schemaLocation", "$projectDir/schemas")
 }
 
-
 // Custom dependency configuration for ktlint
 val ktlint by configurations.creating
 
@@ -163,6 +162,11 @@ tasks.register<Checkstyle>("runCheckstyle") {
     reports {
         xml.required = true
         html.required = true
+    }
+
+    // run checkstyle for extractor when it is built from source
+    if (gradle.extra.has("extractorIncluded")) {
+        dependsOn(gradle.includedBuild("NewPipeExtractor").task(":extractor:checkstyleMain"))
     }
 }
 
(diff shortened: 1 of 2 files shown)


theimpulson:
I don't think it is NewPipe's responsibility to run checkstyle and error out when Extractor is being built with it. Extractor has its own implementation and both should live separately.

ghost:
> I don't think it is NewPipe's responsibility to run checkstyle and error out when Extractor is being built with it. Extractor has its own implementation and both should live separately

They are still separate with their separate rules. Just when you build both of them at the same time it is easier to get error earlier rather than submit PR/commit and then find out from github actions that you didn't remove useless import or something else...

And you always can not uncomment last line which will skip extractor checkstyle... Maybe it would be a good idea to mention it in comment 🤔. Btw imo they are still decoupled enough

When you build both of them from source it means one of two things:
1. you build existing commit (which already must be checkstyle valid)
2. you build your own change which you plan to commit which would need checkstyle anyway.

It is QOL which would allow to prevent following messages:

https://github.com/TeamNewPipe/NewPipeExtractor/pull/1442#issuecomment-3820983524 (there also mentioned tests but btw...)
https://github.com/TeamNewPipe/NewPipeExtractor/pull/1449#issuecomment-3824093796

And since you build both at the same time I think it is very natural to run checkstyle of extractor at the same time

theimpulson:
> They are still separate with their separate rules. Just when you build both of them at the same time it is easier to get error earlier rather than submit PR/commit and then find out from github actions that you didn't remove useless import or something else...

Why not run the task locally then before submitting PRs? I am open to PRs which will improve these checks in Extractor if they are missing. Again, both project are separate and shouldn't be concerned with one another in general.

Issue #6486, opened 2021-06-14
doesn't build reproducibly (due to bug in coreLibraryDesugaring)


obfusk:
I compared the APKs from the last 5 releases on GitHub to the APKs built and published by F-Droid.
Some of them are identical (apart from the signature), but not all of them are.

I've done multiple clean builds of `0.21.4`; they were not always identical.
I've seen 3 variants, one identical to the GitHub APK, another identical to the F-Droid APK.
The differences I've seen were always in `j$/util/Collection$-EL.class`.

### diffoscope output

<details>

```
--- NewPipe_v0.21.4.apk
+++ org.schabi.newpipe_970.apk
├── zipinfo {}
│ @@ -1430,11 +1430,11 @@
│  -rw----     0.0 fat      186 b- stor 81-Jan-01 01:01 res/drawable-hdpi-v4/abc_textfield_activated_mtrl_alpha.9.png
│  -rw----     0.0 fat     1824 b- defN 81-Jan-01 01:01 res/layout/notification_template_big_media_narrow.xml
│  -rw----     0.0 fat      106 b- stor 81-Jan-01 01:01 res/drawable-xxxhdpi-v4/exo_ic_fullscreen_exit.png
│  -rw----     0.0 fat      360 b- defN 81-Jan-01 01:01 res/xml/standalone_badge_offset.xml
│  -rw----     0.0 fat      516 b- defN 81-Jan-01 01:01 res/drawable/abc_seekbar_tick_mark_material.xml
│  -rw----     0.0 fat     4384 b- defN 81-Jan-01 01:01 res/xml/video_audio_settings.xml
│  -rw----     0.0 fat      212 b- stor 81-Jan-01 01:01 res/drawable-hdpi-v4/notification_bg_normal.9.png
│ --rw----     2.0 fat   166112 b- defN 81-Jan-01 01:01 META-INF/SCHABI_K.SF
│ --rw----     2.0 fat     1187 b- defN 81-Jan-01 01:01 META-INF/SCHABI_K.RSA
│ +-rw----     2.0 fat   166112 b- defN 81-Jan-01 01:01 META-INF/EE8807D2.SF
│ +-rw----     2.0 fat     1334 b- defN 81-Jan-01 01:01 META-INF/EE8807D2.RSA
│  -rw----     2.0 fat   165985 b- defN 81-Jan-01 01:01 META-INF/MANIFEST.MF
[...]

PR #13194, opened 2026-02-05
Enable lint checks and resource shrinking

theimpulson:
#### What is it?
- [ ] Bugfix (user facing)

- Enables resource shrinking
- Moves from switch statements to if-else to handle non-final resource ID errors
- Enables lint checks

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/views/SimpleWindowCallback.kt
+++ b/app/src/main/java/org/schabi/newpipe/views/SimpleWindowCallback.kt
@@ -1,0 +1,39 @@
+/*
+ * SPDX-FileCopyrightText: 2026 NewPipe e.V. <https://newpipe-ev.de>
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+package org.schabi.newpipe.views
+
+import android.os.Build
+import android.view.KeyEvent
+import android.view.KeyboardShortcutGroup
+import android.view.Menu
+import android.view.Window
+import androidx.annotation.RequiresApi
+
+/**
+ * Simple window callback class to allow intercepting key events
+ * @see FocusOverlayView.setupOverlay
+ */
+open class SimpleWindowCallback(private val baseCallback: Window.Callback) :
+    Window.Callback by baseCallback {
+
+    override fun dispatchKeyEvent(event: KeyEvent?): Boolean {
+        return baseCallback.dispatchKeyEvent(event)
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    override fun onPointerCaptureChanged(hasCapture: Boolean) {
+        baseCallback.onPointerCaptureChanged(hasCapture)
+    }
+
+    @RequiresApi(Build.VERSION_CODES.N)
+    override fun onProvideKeyboardShortcuts(
+        data: List<KeyboardShortcutGroup?>?,
+        menu: Menu?,
+        deviceId: Int
+    ) {
+        baseCallback.onProvideKeyboardShortcuts(data, menu, deviceId)
+    }
+}
--- a/app/src/main/java/org/schabi/newpipe/views/FocusOverlayView.java
+++ b/app/src/main/java/org/schabi/newpipe/views/FocusOverlayView.java
@@ -40,7 +40,6 @@ import android.view.Window;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.RequiresApi;
-import androidx.appcompat.view.WindowCallbackWrapper;
 
 import org.schabi.newpipe.R;
 
@@ -232,7 +231,7 @@ public final class FocusOverlayView extends Drawable implements
         // Unfortunately many such forms of "scrolling" do not count as scrolling for purpose
         // of dispatching ViewTreeObserver callbacks, so we have to intercept them by directly
         // receiving keys from Window.
-        window.setCallback(new WindowCallbackWrapper(window.getCallback()) {
+        window.setCallback(new SimpleWindowCallback(window.getCallback()) {
             @Override
             public boolean dispatchKeyEvent(final KeyEvent event) {
                 final boolean res = super.dispatchKeyEvent(event);


theimpulson:
Resource shrinking was disabled mentioning to support F-Droid's reproducible builds, however, multiple apps have it enabled and they are reproducible without any issues. It shouldn't be any different for us. If there are any issues, we should try to fix them instead as AGP 9.0.0 introduces more optimized resource shrinking and flags.

app/build.gradle.kts
```diff
@@ -79,7 +79,7 @@ configure<ApplicationExtension> {
                 resValue("string", "app_name", "NewPipe $suffix")
             }
             isMinifyEnabled = true
-            isShrinkResources = false // disabled to fix F-Droid"s reproducible build
```
TobiGr, line 82: Before merging, we should test this on multiple different devices with the release build and verify the checksums of the generated unsigned APKs for a specific commit.

app/lint.xml
```diff
@@ -5,4 +5,6 @@
   -->
 <lint>
     <issue id="MissingTranslation" severity="ignore" />
+    <issue id="MissingQuantity" severity="ignore" />
```
TobiGr, line 8: >  ../../src/main/res/values-is/strings.xml:318: The quantity 'one' matches more than one specific number in this locale (1, 21, 31, 41, 51, 61, 71, 81, 101, 1001, …), but the message did not include a formatting argument (such as %d). This is usually an internationalization error. See full issue explanation for more.

This is a problem which cannot be solved by the translators but must be solved by us. If there is no formatting argument for "one", e.g. because we use "One" in the string ressource, it will cause problems for other localizations as explained in the error message.

See https://stackoverflow.com/a/42884713 and https://developer.android.com/guide/topics/resources/string-resource#Plurals

app/src/main/java/org/schabi/newpipe/error/ReCaptchaActivity.java
```diff
@@ -127,6 +127,7 @@ public boolean onCreateOptionsMenu(final Menu menu) {
 
     @Override
     public void onBackPressed() {
+        super.onBackPressed();
```
TobiGr: Did you test this change?

`saveCookiesAndFinish()` is taking care of navigating up in the backstack and I think calling the super method was skipped intentionally here: 

https://github.com/TeamNewPipe/NewPipe/blob/735d1ab05d80e7cad37c36bb24ecd9c12b0f5321/app/src/main/java/org/schabi/newpipe/error/ReCaptchaActivity.java#L165-L167

TobiGr (commented):
thanks

app/build.gradle.kts
ghost: About reproducibility look here https://f-droid.org/docs/Reproducible_Builds/

There even subparagraph

>Bug: coreLibraryDesugaring not deterministic
NB: this should be fixed in R8 (com.android.tools:r8) 3.0.69 and later.
In some cases builds are not reproducible due to a [bug in coreLibraryDesugaring](https://issuetracker.google.com/issues/195968520) (may require a Google account to view); this [affected NewPipe](https://github.com/TeamNewPipe/NewPipe/issues/6486).

So it seems to me that specific bug should be fixed now (I am not sure)

It contains other very common problems which can affect newpipe with versions of agp very they are fixed and how to workaround

app/src/main/java/org/schabi/newpipe/error/ReCaptchaActivity.java
theimpulson: I didn't noticed it was already handled and intentional change. I suppressed the error for that method instead.

app/lint.xml
theimpulson: How do we want to handle this? Should I add a placeholder for one in all strings? but then that introduces the possibility of adding it wrong for languages with different layout directions.

Edit: Adding an argument placeholder in the english string might trigger crashes on other locales too, so we might need to add it to all locales.

Stypox: @TobiGr it's not fully clear to me why the problem cannot be solved by translators:
- For [MissingQuantity](https://android.googlesource.com/platform/tools/base.git/+/refs/heads/oreo-r2-release/lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/PluralsDetector.java#56), it means that a "zero/one/few/many" plural is missing for a language that needs it. Can't translators add any plural form they want in Weblate?
- For [ImpliedQuantity](https://android.googlesource.com/platform/tools/base.git/+/refs/heads/oreo-r2-release/lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/PluralsDetector.java#99), it means that `%d` is missing in a plural that can possibly take multiple numbers in a specific language. In English this problem does not exist since e.g. "one" only applies to `1`, but in Lithuanian it applies to `1, 21, 31, 41`. Wouldn't the Lithuanian translators just be able to add `%d` to their strings?

So in both cases I don't think we would need to change anything in the English `strings.xml`, as it's the duty of translators to translate the plurals correctly (including adding/removing forms, and adding `%d`. Or are there Weblate limitations for either of these usecases?

app/build.gradle.kts
```diff
@@ -88,13 +88,7 @@ configure<ApplicationExtension> {
     }
 
     lint {
-        checkReleaseBuilds = false
-        // Or, if you prefer, you can continue to check for errors in release builds,
-        // but continue the build even when errors are found:
-        abortOnError = false
```
Stypox, line 94: Were these removed on purpose? I guess it makes sense, but maybe `abortOnError = false` is problematic when you just want to try some stuff out (checkstyle already gets in the way often enough)

app/src/main/java/org/schabi/newpipe/local/feed/notifications/NotificationHelper.kt
```diff
@@ -126,7 +130,9 @@ class NotificationHelper(val context: Context) {
     ) {
         for (stream in newStreams) {
             val notification = createStreamNotification(stream, serviceId, channelUrl, channelIcon)
-            manager.notify(stream.url.hashCode(), notification)
+            if (manager.areNotificationsEnabled()) {
```
Stypox: Maybe wrap the `if` around the whole `for`

app/src/main/java/org/schabi/newpipe/local/feed/notifications/NotificationHelper.kt
Stypox, line 1: I am pretty sure none of these `manager.notify()` paths are reachable if notifications are disabled, because the "notifications disabled" check is done when starting the "new streams notifications" update. But yeah I guess the user could possibly disable notifications while the check is running, so it makes sense to add more checks. Maybe add a comment on the checks explaining that this is the reasoning?

app/src/main/java/org/schabi/newpipe/error/ReCaptchaActivity.java
```diff
@@ -126,6 +126,7 @@ public boolean onCreateOptionsMenu(final Menu menu) {
     }
 
     @Override
+    @SuppressLint("MissingSuperCall")
```
Stypox:
```suggestion
    @SuppressLint("MissingSuperCall") // saveCookiesAndFinish method handles back navigation
```

app/src/main/java/org/schabi/newpipe/local/subscription/dialog/FeedGroupDialog.kt
```diff
@@ -506,7 +506,7 @@ class FeedGroupDialog : DialogFragment(), BackPressable {
     private fun hideKeyboardSearch() {
         inputMethodManager.hideSoftInputFromWindow(
             searchLayoutBinding.toolbarSearchEditText.windowToken,
-            InputMethodManager.RESULT_UNCHANGED_SHOWN
+            InputMethodManager.HIDE_NOT_ALWAYS
```
Stypox, line 509: Nice catch! I wonder what was happening before...

Now there is a new deprecation... https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#hideSoftInputFromInputMethod(android.os.IBinder,%20int) . But probably not worth to work on, as long as this works, since it's going to be replaced in compose

app/src/main/res/values-bn-rIN/strings.xml
```diff
@@ -309,6 +309,6 @@
     <string name="notification_action_nothing">কিছু না</string>
     <string name="yes">হ্যাঁ</string>
     <string name="no">না</string>
-    <string name="search_with_service_name">সার্চ</string>
-    <string name="search_with_service_name_and_filter">খুঁজুন</string>
```
Stypox: Generally I'd suggest you do these changes directly in Weblate to avoid conflicts (but yeah no big deal)

app/src/main/java/us/shandian/giga/get/DownloadRunnableFallback.java
```diff
@@ -21,7 +21,7 @@
  * Single-threaded fallback mode
  */
 public class DownloadRunnableFallback extends Thread {
-    private static final String TAG = "DownloadRunnableFallback";
```
Stypox, line 24: Isn't it still too long? 😅

app/src/main/java/org/schabi/newpipe/views/SimpleWindowCallback.kt
```diff
@@ -0,0 +1,35 @@
...
+package org.schabi.newpipe.views
+
+import android.os.Build
+import android.view.KeyEvent
+import android.view.KeyboardShortcutGroup
+import android.view.Menu
+import android.view.Window
+import androidx.annotation.RequiresApi
+
+open class SimpleWindowCallback(private val baseCallback: Window.Callback) :
```
Stypox, line 19: Why is this wrapper needed, if the only thing it does is delegate to baseCallback and still extends Window.Callback? Could you add a (short) javadoc for motivation when you add new classes?

app/src/main/res/values-bn-rIN/strings.xml
theimpulson: Done on Weblate

app/src/main/java/org/schabi/newpipe/local/feed/notifications/NotificationHelper.kt
theimpulson: Done

app/src/main/java/org/schabi/newpipe/local/feed/notifications/NotificationHelper.kt
theimpulson: I assume that's the reason lint is giving error on these calls especially since on new Android versions, its a runtime permission now. As there are lots of such calls and adding comment on all wouldn't be so nice, I added it in commit message in case someone looks up the reason later with git blame.

app/src/main/java/org/schabi/newpipe/error/ReCaptchaActivity.java
theimpulson: Added, thank you

app/build.gradle.kts
theimpulson: Added back

app/src/main/java/us/shandian/giga/get/DownloadRunnableFallback.java
theimpulson: Thanks for noticing. I was trying it out and never committed the abbreviated change. Fixed now.

app/src/main/java/org/schabi/newpipe/views/SimpleWindowCallback.kt
theimpulson: Done. The wrapper is needed for the method in FocusOverlay file which is currently using a restricted API failing lint checks. I added a simple javadoc with a see to the only method that uses it.

app/build.gradle.kts
theimpulson: Just a note but release builds will fail as I moved the translation fix to Weblate. This PR will need to be rebased on top of that fix whenever it is pulled in.

app/lint.xml
TobiGr: I was thinking of plurals like this
https://github.com/TeamNewPipe/NewPipe/blob/3815f5f593d0c0d66389166858b27bba276ea953/app/src/main/res/values/strings.xml#L609-L611
in comparison to
https://github.com/TeamNewPipe/NewPipe/blob/3815f5f593d0c0d66389166858b27bba276ea953/app/src/main/res/values/strings.xml#L646-L649

They need to be fixed by us first before translators can translate them properly, I guess. But it looks like most of them have been fixed already. I know that we've had more strings for "one" that did not have a param. But yeah, apart from that, we can enable those rules.

app/build.gradle.kts
TobiGr: Sorry, I am out of the loop. What translation fix?

theimpulson: I meant https://github.com/TeamNewPipe/NewPipe/commit/3815f5f593d0c0d66389166858b27bba276ea953. I am waiting for suggestion on how to resolve the one currently unresolved.

app/lint.xml
theimpulson: Can we enable them in a separate PR to unblock this for now?

TobiGr: Yes

TobiGr: Yes

TobiGr: Yes

app/build.gradle.kts
theimpulson: On my linux machine:

```shell
aayush@framework:~/Projects/NewPipe$ sha256sum app/build/outputs/apk/release/app-release-unsigned.apk 
cb2fbd50d2402229efb9ce66c0a6b5b489c74c22e7282d136da45cf85835a9cb  app/build/outputs/apk/release/app-release-unsigned.apk
```

theimpulson: on my macbook:

```shell
aayush@Aayushs-MacBook-Pro.local:~/StudioProjects/NewPipe (r8fixes) $ sha256sum app/build/outputs/apk/release/app-release-unsigned.apk 
cb2fbd50d2402229efb9ce66c0a6b5b489c74c22e7282d136da45cf85835a9cb  app/build/outputs/apk/release/app-release-unsigned.apk
```

TobiGr: Same for me on linux and macos.

PR #13195, opened 2026-02-05
Fixed resuming streams at the last playback position

### LINKED ISSUE #13139: [Regression] Resume playback from last position no longer working
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).
- [x] I have read and understood the [AI policy](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md#ai-policy). The content of this bug report is not generated by AI.

### Affected version

0.28.2

1. Find a stream that you've already watched but didn't finish (and it shows the red line under it showing how much you've watched); alternatively make one by playing something but not finished it
2. Close the player, then go to history and play the stream again

### Expected behavior

Stream plays from the last playback position as displayed by the red line under the thumbnail

### Actual behavior

It resumes from the beginning

https://github.com/user-attachments/assets/dc9ed424-02d8-46db-86b6-095f36a0d68e

### Logs

_No response_

Android 9

### Affected device model

Samsung Galaxy S8

### Additional information

Recording was provided by someone else

absurdlylongusername:
#### What is it?
- [x] Bugfix (user facing)

Reason for the bug is that https://github.com/TeamNewPipe/NewPipe/commit/1d8ea0181fc6a447153b0f1ffe4f27006bd1109a adds a call to `useVideoAndSubtitles` in the `initPlayback` overrides in the UI classes. This causes the recovery position for the queue item to be overridden due to the following flow:

- `Player.handleIntent` is called
- The `if (intent.getBooleanExtra(RESUME_PLAYBACK, false)` branch is reached
- It gets the playback position of the current queue item from the database via `loadStreamState`
- `newQueue.setRecovery(newQueue.getIndex(), state.getProgressMillis());` sets the recovery position to the value from DB
- Then calls `initPlayback(newQueue, playWhenReady);`
- `initPlayback` then calls `UIs.call(PlayerUi::initPlayback);`, which in turn calls `initPlayback` on one of `BackgroundPlayerUi`, `PopupPlayerUi`, `MainPlayerUi`
- `initPlayback` override in these classes calls `Player.useVideoAndSubtitles`
- `useVideoAndSubtitles` does `setRecovery()`
- This `setRecovery()` however gets the recovery position from `simpleExoPlayer.getCurrentPosition()` (which is effectively zero since we just started to resume playback), and this overwrites the previous recovery position we just set `setRecovery` in `handleIntent`
- `onPlaybackSynchronize` gets called at some point and seeks to the recovery position (which we just set to 0)
- And thus, it seeks to 0 instead of seeking to the saved playback position

The if conditions I added are supposed to ensure `setRecovery` is only called when there is no pending recovery (it won't be overwriting a recovery position that's already been set), or when exoplayer is far along in it's setup so that there's stuff to play (timeline is not empty).

I considered doing a proper refactor of this, because I don't really think calling `useVideoAndSubtitles` from the UIs is the best way to achieve the goal of minimizing bandwidth when playing in the background as it breaks separation of concerns, but I decided it's not worth it and can be tackled in NewPlayer at a later date.

Maybe later

- Fixes #13139 

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -2223,7 +2223,6 @@ public final class Player implements PlaybackListener, Listener {
             final SourceType sourceType = videoResolver.getStreamSourceType()
                     .orElse(SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY);
 
-            setRecovery(); // making sure to save playback position before reloadPlayQueueManager()
             if (hasTimeline || !hasPendingRecovery) {
                 // making sure to save playback position before reloadPlayQueueManager()
                 setRecovery();
@@ -2240,7 +2239,6 @@ public final class Player implements PlaybackListener, Listener {
             Reload the play queue manager in this case, which is the behavior when we don't know the
             index of the video renderer or playQueueManagerReloadingNeeded returns true
             */
-            setRecovery(); // making sure to save playback position before reloadPlayQueueManager()
             if (hasTimeline || !hasPendingRecovery) {
                 // making sure to save playback position before reloadPlayQueueManager()
                 setRecovery();


app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2234,6 +2241,10 @@ The current metadata may be null sometimes (for e.g. when using an unstable conn
             index of the video renderer or playQueueManagerReloadingNeeded returns true
             */
             setRecovery(); // making sure to save playback position before reloadPlayQueueManager()
+            if (hasTimeline || !hasPendingRecovery) {
+                // making sure to save playback position before reloadPlayQueueManager()
+                setRecovery();
+            }
```
TobiGr: Is this really needed here? `setRecovery()` is called anyway.

app/src/main/java/org/schabi/newpipe/player/Player.java
```diff
@@ -2214,13 +2210,24 @@ public void useVideoAndSubtitles(final boolean videoAndSubtitlesEnabled) {
...
             // In case we don't know the source type, fall back to either video-with-audio, or
             // audio-only source type
             final SourceType sourceType = videoResolver.getStreamSourceType()
                     .orElse(SourceType.VIDEO_WITH_AUDIO_OR_AUDIO_ONLY);
 
             setRecovery(); // making sure to save playback position before reloadPlayQueueManager()
+            if (hasTimeline || !hasPendingRecovery) {
+                // making sure to save playback position before reloadPlayQueueManager()
+                setRecovery();
+            }
```
TobiGr: > The if conditions I added are supposed to ensure `setRecovery` is only called when there is no pending recovery (it won't be overwriting a recovery position that's already been set), or when exoplayer is far along in it's setup so that there's stuff to play (timeline is not empty).

Hm, `setRecovery() `is called anyway here, independently from whether a recovery position was already set or not.. Don't you need to remove the first call? This is at least what I understood from your detailed explanation.

app/src/main/java/org/schabi/newpipe/player/Player.java
TobiGr, line 2245: See above.

app/src/main/java/org/schabi/newpipe/player/Player.java
absurdlylongusername: Lmao my bad, this a git issue, seems the rebasing I did onto dev didn't actually end up as it showed in Android Studio. Above setRecovery is not supposed to be there x).

I've confirmed this doesn't actually work with the above setRecovery still in place. I've removed it now

app/src/main/java/org/schabi/newpipe/player/Player.java
absurdlylongusername: See answer to above

Issue #13199, opened 2026-02-05
Hover tooltips work through player ui


dustdfg:
### Checklist

- [x] I am able to reproduce the bug with the latest version given here: [CLICK THIS LINK](https://github.com/TeamNewPipe/NewPipe/releases/latest).
- [x] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.
- [x] This issue contains only one bug.
- [x] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).
- [x] I have read and understood the [AI policy](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md#ai-policy). The content of this bug report is not generated by AI.

### Affected version

0.28.0 - 0.28.2 (earlier versions probably too)

### Steps to reproduce the bug

1. Open any video (from page where you can find elements which show tooltip after hovering)
2. Try to hover those elements with mouse

### Expected behavior

Hovering elements which shows tooltips doesn't work when covered by player surface

### Actual behavior

Hovering elements which shows tooltips works despite them being covered by player surface

### Screenshots/Screen recordings

https://github.com/user-attachments/assets/0deba101-9647-4d10-be46-813e19e67793

### Affected Android/Custom ROM version

Android 13 and unknown android (shouldn't matter)

### Affected device model

Waydroid PC, some real android tables with connected mouse

AudricV:
May be related to #5600

ghost:
> May be related to [#5600](https://github.com/TeamNewPipe/NewPipe/issues/5600)

Looks like it is really the same bug because when there is no error I can only hover through but when there is error I can additionally click through. Looks like problem is present even without error in content loading. When error present there is just no one to catch clicks before they reach view under "video player"

ghost:
Reopening now as the tooltips part of the issues is still not solved and can't be solved without proper fix or just refactor

PR #13203, opened 2026-02-06
VideoDetailFragment: Forcefully catch click events uncaught by children

### LINKED ISSUE #5600: Content behind the view of an unavailable content can be clicked when this view is shown
### Checklist
- [x] I am using the latest version - 0.20.10 + build based on a recent dev branch after 0.20.10

1. Search something in NewPipe.
2. Share an unavailable content in NewPipe from an other app.
3. Click everywhere on the view at the bottom of the error text

### Actual behaviour
Click on the view will open a result from the previous search. This isn't something excepted.

### Expected behavior
Click on the view should do nothing.

https://user-images.githubusercontent.com/74829229/107980373-6b53f380-6fc0-11eb-8f75-a96cf7b8db9d.mp4

### Device info
 - Android version/Custom ROM version: EMUI 10.0.0 / Android 10
 - Device model: Honor 9X

ghost:
#### What is it?
- [x] Bugfix (user facing)

VideoDetailFragment: Forcefully catch click events uncaught by children because otherwise they will be caught by underlying view and "click through" will happen

It is a workaround by it is nature but I think it is severe enough for UX (especially because average user won't event understand what happens) so I think it deserves workaround fix...

#13199 can't be solved by this I tried `setOnHoverListener`, `setOnTouchListener`, `setOnGenericMotionListener` doesn't work. It seems to me that tooltips are controlled by android itself and that we obviosly have broken logic that android thinks we really hovered the element and it shows tooltip. So probably tooltips part of the bug is solvable only with proper fix (which I can't even imagine) but as I said click through part is enough severe imo to deserve workaround fix :/

- Fixes #5600

#### APK testing

#### Due diligence
- [X] The proposed changes follow the [AI policy](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md#ai-policy).
- [X] I tested the changes using an emulator or a physical device.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/detail/VideoDetailFragment.java
@@ -646,6 +646,12 @@ public final class VideoDetailFragment
     protected void initListeners() {
         super.initListeners();
 
+        // Workaround for #5600
+        // Forcefully catch click events uncaught by children because otherwise
+        // they will be caught by underlying view and "click through" will happen
+        binding.getRoot().setOnClickListener(v -> { });
+        binding.getRoot().setOnLongClickListener(v -> true);
+
         setOnClickListeners();
         setOnLongClickListeners();
 


ghost:
To ease your time testing: here is the link with playlist of age restricted videos (disclaimer: they are dark maybe not use it...) https://www.youtube.com/playlist?list=PLftvg-3EEijV8Db4uqYmS8ZBDDOCBpbmf

Just open it in the app it will set "background with clickable videos", click on first video and then try to click through

TobiGr:
The error panel has been refactored since the original issue was opened. Can the problem still ocour in 0.28.3?

ghost:
> The error panel has been refactored since the original issue was opened. Can the problem still ocour in 0.28.3?

Yeah, of course, I wouldn't create PR if it wasn't appearing on dev. It also happens not only when there is error reported to user but also when video just loads and you click through recommended videos (tablet mode). Moreover it also appears in refactor. So it (probably) has nothing to do with error panel itself but rather video detail fragment

TobiGr (approved):

ghost:
I have a question. If I ever mess with something like `VIdeoDetailFragment` again, which differs in dev and refactor and so can't be just automatically backported. Should I create second PR like I did for another PR messing with `VIdeoDetailFragment` or maybe there any better options? Or in case of this PR I even doubt it is even necessary to backport the PR because refactor will come soon. At least I plan to touch it the parts of refactor which can make it unnecessary

TobiGr:
The best way is to create a PR that merges dev into refactor so that other fixes on dev are still compatible with refactor. And that PR can implement the fix in the proper way for the refactor branch if it cannot be applied automatically.

PR #13282, opened 2026-02-22
Add warning banner about ongoing google certification for android apps

ghost:
#### What is it?
- [ ] Bugfix (user facing)
- [X] Feature (user facing) ⚠️ **Your PR must target the [`refactor`](https://github.com/TeamNewPipe/NewPipe/tree/refactor) branch**
- [X] Meta improvement to the project (dev facing)

- Add warning banner about ongoing google certification for android apps
- Reopens every 30 days (190/30 ~6 times only)

- 
20260223_01h49m56s_grim

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/MainActivity.java
+++ b/app/src/main/java/org/schabi/newpipe/MainActivity.java
@@ -988,7 +988,21 @@ public class MainActivity extends AppCompatActivity {
                 0
         ));
 
-        final var kaoURI = Uri.parse("https://keepandroidopen.org");
+        final var supportedLannguages = List.of("fr", "de", "ca", "es", "id", "it", "pl",
+                "pt", "cs", "sk", "fa", "ar", "tr", "el", "th", "ru", "uk", "ko", "zh", "ja");
+        final var locale = Localization.getAppLocale();
+        final String kaoBaseUrl = "https://keepandroidopen.org/";
+        final String kaoURIString;
+        if (supportedLannguages.contains(locale.getLanguage())) {
+            if ("zh".equals(locale.getLanguage())) {
+                kaoURIString = kaoBaseUrl + ("TW".equals(locale.getCountry()) ? "zh-TW" : "zh-CN");
+            } else {
+                kaoURIString = kaoBaseUrl + locale.getLanguage();
+            }
+        } else {
+            kaoURIString = kaoBaseUrl;
+        }
+        final var kaoURI = Uri.parse(kaoURIString);
         final var solutionURI = Uri.parse(
                 "https://github.com/woheller69/FreeDroidWarn?tab=readme-ov-file#solutions");
 
@@ -1009,7 +1023,7 @@ public class MainActivity extends AppCompatActivity {
                     .setNegativeButton(this.getString(R.string.kao_dialog_more_info), null)
                     .show();
 
-            // If we use setNeutralButton and etc. dialog will close after pressing the buttons
+            // If we use setNeutralButton and etc. dialog will close after pressing the buttons,
             // but we want it to close only when positive button is pressed
             dialog.getButton(AlertDialog.BUTTON_NEGATIVE).setOnClickListener(v ->
                     this.startActivity(new Intent(Intent.ACTION_VIEW, kaoURI))


app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -973,4 +979,36 @@ private boolean bottomSheetHiddenOrCollapsed() {
...
+                "https://github.com/woheller69/FreeDroidWarn?tab=readme-ov-file#solutions");
+
+        if (kaoLastCheck.plus(30, ChronoUnit.DAYS).isBefore(now)) {
+
+            new AlertDialog.Builder(this)
+                    .setTitle("KeepAndroidOpen")
+                    .setMessage(this.getString(R.string.kao_dialog_warning))
+                    .setPositiveButton(this.getString(android.R.string.ok), (d, w) -> {
+                        prefs.edit()
+                                .putLong("kao_last_checked", now.toEpochMilli())
```
ghost: Is it so necessary to create new xml key? I mean it is temporary thing...

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -973,4 +979,36 @@ private boolean bottomSheetHiddenOrCollapsed() {
                 || sheetState == BottomSheetBehavior.STATE_COLLAPSED;
     }
 
+    private void showKeepAndroidDialog() {
+        final var prefs = PreferenceManager.getDefaultSharedPreferences(this);
+
+        final var now = Instant.now();
+        final var kaoLastCheck = Instant.ofEpochMilli(prefs.getLong("kao_last_checked", 0));
+
+        final var kaoURI = Uri.parse("https://keepandroidopen.org");
```
ghost: If we want links point to translated webpages we need some sort of retrieving current lang code or maybe just put it in string resources?

ghost:
@TobiGr @Stypox @Poolitzer 

Popup with title or not?

Also there are three wordings:

Original like:
> Google has announced that, starting in 2026/2027, all apps on certified Android devices will require the developer to submit personal identity details directly to Google. Since the developers of NewPipe do not agree to this requirement, NewPipe will no longer work on certified Android devices after that time.

Proposed by @Poolitzer 

> In August 2025, Google announced that as of September 2026,it will no longer be possible to develop apps for the Android platform without first registering centrally with Google.Since the developers of NewPipe do not agree to this requirement, NewPipe will no longer work on certified Android devices after that time.


Previous + some rip of from [KeepAndroidOpen site](https://keepandroidopen.org/)

> In August 2025, Google announced that as of September 2026, installing apps will require developer verification for all Android apps on certified devices, including those installed outside of the Play Store. Since the developers of NewPipe do not agree to this requirement, NewPipe will no longer work on certified Android devices after that time.

---

We still probably need a way to remotely update the popup on user device because in this situation we can't just say: "dear users update". About which I would like hear some implementation proposals if I need to implement it in this PR

ghost:
> We still probably need a way to remotely update the popup on user device because in this situation we can't just say: "dear users update". About which I would like hear some implementation proposals if I need to implement it in this PR

Especially taking into account the answer I got (which I can't find) about infrastructure working on edge already

Poolitzer:
Sorry why do we need to update the pop up? If the app stops working, users will go to the website/the github repo, no? And I don't quite like the app pinging our servers for new pop up content, just from a privacy and an infrastructure perspective. Automatic up dates are one thing and easily understandable, pop up content queries are another, for me at least. At least we should have another setting to disable that like the auto update.

I realized that "NewPipe will no longer work on certified Android devices after that time." isnt actually true, it will work with the extra adb steps no? Should we go with like, "NewPipe will stop working on certified Android devices after that time unless you take some steps"?

And then I also realized we definitely need a setting/button to disable this from appearing again, because at least I will be really annoyed if I have to click on it every 30 days. And then probably a button/link in some related menu to allow users to get back that information, either via again showing the pop up or just linking to the blog post.

Poolitzer:
I now made a website update introducing the banner and a blog post, I am not sure if we will do a full new version release with features and want to rewrite the blog post accordingly, https://github.com/TeamNewPipe/website/pull/426

ghost:
> Sorry why do we need to update the pop up? If the app stops working, users will go to the website/the github repo, no? And I don't quite like the app pinging our servers for new pop up content, just from a privacy and an infrastructure perspective. Automatic up dates are one thing and easily understandable, pop up content queries are another, for me at least. At least we should have another setting to disable that like the auto update.

I fully agree with it but I just remember first discussion where dynamic banner was discussed. And after day comes and superposition collapses to one state as a user at least I would probably like to know how it ends. I think the existing links are enough though

But maybe there is a solution. When day comes we can open new popup (which should be added now), which will point to newpipe blog with more clarifications (?) The blog post which still doesn't exist and won't exist before the day comes but link to which can be predicted now

> 
> I realized that "NewPipe will no longer work on certified Android devices after that time." isnt actually true, it will work with the extra adb steps no? Should we go with like, "NewPipe will stop working on certified Android devices after that time unless you take some steps"?
> 

I am not even sure if idea of works or not works correct. Maybe already installed apps will continue working? Idk. I am not sure. But knowing google I would expect the worst

> And then I also realized we definitely need a setting/button to disable this from appearing again, because at least I will be really annoyed if I have to click on it every 30 days.

Generally yes but isn't 30 enough? Probably it is necessary evil. And you will see it only ~6 times
>And then probably a button/link in some related menu to allow users to get back that information, either via again showing the pop up or just linking to the blog post.

Showing pop up "Hey I am popup, you've disabled another popup, do you want to return the disabled one?" 🤨

Related menu? New menu in settings? Do users even look there? Even if after pressing button "do not show again" we additionally tell them they can open it again via settings

app/src/main/java/org/schabi/newpipe/MainActivity.java
TobiGr: It is good to have an index of all keys used, especially when you want to clean them up later (which we should do after the campaign ended or the issue has been resolved)

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -973,4 +979,36 @@ private boolean bottomSheetHiddenOrCollapsed() {
...
+                    .setNeutralButton(
+                            this.getString(R.string.kao_solution),
+                            (d, w) ->
+                                    this.startActivity(new Intent(Intent.ACTION_VIEW, solutionURI)))
+
+                    .setNegativeButton(
+                            this.getString(R.string.kao_dialog_more_info),
+                            (d, w) ->
+                                    this.startActivity(new Intent(Intent.ACTION_VIEW, kaoURI)))
+                    .show();
```
TobiGr, line 1024: I don't quite understand the reason to only set `kao_last_checked` when clicking on "ok". Users choosing one of the other options are clearly more informed in the end.

TobiGr:
> Popup with title or not?

I'd use a title, but with spaces between words which is also the spelling on their website ("Keep Android Open").

app/src/main/java/org/schabi/newpipe/MainActivity.java
ghost: I agree but there is a catch: pressing button closes the popup but you can choose only one from two buttons. It means that you won't be able to look at second link. With this workaround you still won't be able to immediately do it but on next app start you (not 30 days after) you will be able to read second button

Now when I wrote it down it seems quite awful. I need to find another solution...

ghost: Now dialog is not closed when you click link buttons so the only option is to click the ok button

app/src/main/java/org/schabi/newpipe/MainActivity.java
ghost: Still under question

TobiGr: I woould not put this in string ressources that are translatable. This could lead to changes by translators adding unsupported locals. It is better to just use the default page and let the user change it if needed imo. 
You could also use [`Localization.getAppLocale().getLanguage()`](https://github.com/TeamNewPipe/NewPipe/blob/dd65db56a900ec3223f834466add84eff6d9fe3c/app/src/main/java/org/schabi/newpipe/util/Localization.java#L120)

TobiGr: I pushed something that is OK, but not a perfect approach (e.g. it is ignoring more locals from the locale list). However, I think it should be fine.

ghost: Also I've just realized it is a "security" treat to put link into resources which can be translated because if we then merge weblate and overlook the change it could lead to problems...

ghost:
Now as a remainder the strings must be manually marked as changed in weblate. Which probably can be done only by those who have rights (if can be done by anyone I still can't do it because didn't interact with weblate at all)

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -973,4 +979,58 @@ private boolean bottomSheetHiddenOrCollapsed() {
...
+        } else {
+            kaoURIString = kaoBaseUrl;
+        }
+        final var kaoURI = Uri.parse(kaoURIString);
+        final var solutionURI = Uri.parse(
+                "https://github.com/woheller69/FreeDroidWarn?tab=readme-ov-file#solutions");
+
+        if (kaoLastCheck.plus(30, ChronoUnit.DAYS).isBefore(now)) {
+            final var dialog = new AlertDialog.Builder(this)
+                    .setTitle("Keep Android Open")
```
AudricV, line 1011: I am not sure if keeping this in English only is a good idea.

app/src/main/java/org/schabi/newpipe/MainActivity.java
```diff
@@ -973,4 +979,58 @@ private boolean bottomSheetHiddenOrCollapsed() {
...
+
+            // If we use setNeutralButton and etc. dialog will close after pressing the buttons,
+            // but we want it to close only when positive button is pressed
+            dialog.getButton(AlertDialog.BUTTON_NEGATIVE).setOnClickListener(v ->
+                    this.startActivity(new Intent(Intent.ACTION_VIEW, kaoURI))
+            );
+            dialog.getButton(AlertDialog.BUTTON_NEUTRAL).setOnClickListener(v ->
+                    this.startActivity(new Intent(Intent.ACTION_VIEW, solutionURI))
+            );
+        }
```
AudricV, line 1034: `ShareUtils.openUrlInBrowser` should be used instead.

app/src/main/java/org/schabi/newpipe/about/AboutActivity.kt
```diff
@@ -254,6 +254,13 @@ class AboutActivity : AppCompatActivity() {
                 "ByteHamster",
                 "https://github.com/ByteHamster/SearchPreference",
                 StandardLicenses.MIT
+            ),
+            SoftwareComponent(
+                "FreeDroidWarn",
+                "2026",
+                "woheller69",
+                "https://github.com/woheller69/FreeDroidWarn",
+                StandardLicenses.APACHE2
```
AudricV, line 263: Why did you add this? The app does not use this library with your changes.

ghost: We used initial translations from there

app/src/main/java/org/schabi/newpipe/MainActivity.java
ghost: It is the name of the site.... And to be honest it seems to me that site itself not very sure if it should be translated. Because half of the translations use english version while another half translates without clear rule when so...

app/src/main/java/org/schabi/newpipe/about/AboutActivity.kt
AudricV: Yes, but it doesn't change the fact we don't use this library, this should be removed.

Issue #5325, opened 2021-01-01
Sync between devices


AlexDumi29:
### Checklist

- [x] I checked, but didn't find any duplicates (open OR closed) of this issue in the repo. 
- [x] I have read the contribution guidelines given at https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md.
- [x] This issue contains only one feature request. I will open one issue for every feature I want to request.

#### Describe the feature you want
Sync playlists and subscriptions between devices.

#### Is your feature request related to a problem? Please describe it
I have two phones and a tablet and I have to add new subscriptions and videos manually. Import export exists but it' kind of tedious and from what I've observed a while ago not too reliable.

#### How will you/everyone benefit from this feature?
I believe the community would benefit tremendously because of the ease of use this feature brings. Many more people would use Newpipe thanks to the simplicity of having all of your playlist and subscriptions in one place, plus people won't lose all of their subscriptions and playlists due to a bad export.

Timmy183-ui:
Will defeat the whole purpose of privacy until it is aes256 encrypted or something

ovz93br43v7:
Probably this can be done with https://github.com/syncthing/syncthing ? Then the sync is safe regarding any privacy risks.

Zakkumaru:
I, too, was going to recommend Syncthing.  However, with the  emergence of EteBase ( https://github.com/etesync ), I think it would be possible to create a database all your devices could use to sync NewPipe.  I would very much like to see this become a thing.

eternaltyro:
I want to add my support to what @Zakkumaru said about implementing sync with etesync. It's the way to go in my opinion.

QJKX:
#1163 is about (server based) sync

tdehaeze:
Would love to be able to use Syncthing.
Is this currently not possible because the NewPipe data/configuration are not accessible in the phone directory (without root)?

Thanks

cloo:
As [freetube](https://freetubeapp.io/), the yt desktop client has no sync system yet too, but mention future plans to implement it, how cool it would be for NewPipe to collaborate on a common system! So we could sync not only across devices but across platforms!

Delikt:
+1 for a simple Time based (daily, hourly etc ) Option to Import and Export database automatically in the background and use Syncthing to sync all my devices from a given Path on the storage.

sentriz:
[...]

PR #13306, opened 2026-03-04
Feat/nostr sync

krzyczak:
#### What is it?
- [ ] Bugfix (user facing)

Implemented private sync using Nostr protocol. At the moment the syncable features are
- watch history (including the 
- subscriptions

Related to #5325

This is achieved using:
- Logging in to Nostr can be done using a signer app (NIP-55, like Primal or Amber)
- Also an account can be created in-app (Then "Show Identity" button displays QR codes for npub and nsec, so the account can be used in other apps)
- The data is stored using `kind:30078` from `NIP-78`, and ut uses `NIP-44` encryption so everything stays private
- There is a set of default relays, but custom relays can be added
More details are in the NOSTR_SYNC.md

- Before:
- After:
screenshot_01_drawer_nostr_sync_entry_small
screenshot_02_nostr_sync_main_small
screenshot_03_sign_in_dialog_small
screenshot_04_sign_in_advanced_small

- Fixes #

- 

#### APK testing

#### Due diligence
- [ ] The proposed changes follow the [AI policy](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md#ai-policy).

# AI NOTE
I do not know Kotlin well enough to be able to code in it such a big feature, so I did not adhere to the AI usage policy.

However I was missing the sync feature a lot. Nostr was my choice because of its decentralized nature and self-hosting possibility.

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
+++ b/app/src/main/java/org/schabi/newpipe/fragments/list/BaseListFragment.java
@@ -46,6 +46,8 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I>
 
     private boolean useDefaultStateSaving = true;
     private int updateFlags = 0;
+    @Nullable
+    private String listViewModeKey;
 
     /*//////////////////////////////////////////////////////////////////////////
     // Views
@@ -72,6 +74,7 @@ public abstract class BaseListFragment<I, N> extends BaseStateFragment<I>
     public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setHasOptionsMenu(true);
+        listViewModeKey = getString(R.string.list_view_mode_key);
         PreferenceManager.getDefaultSharedPreferences(activity)
                 .registerOnSharedPreferenceChangeListener(this);
     }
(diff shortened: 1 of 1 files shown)


absurdlylongusername:
Given your comment about AI usage, I presume you don't understand how this actually works.

If you do, please provide an in-depth explanation.

PR descriptions should be within the actual PR description, and not in a separate markdown file; that reduces cohesion and makes it inconvenient in having to go read a file somewhere else.

I don't know what Nostr is or how it works, and if you consider this a big feature it would be useful to have a proper explanation that explains architectural design decisions, implementation functionality and etc.

As it is, the NOSTR_SYNC is not understandable at all without extra research to make sense of it; that should not be the case. 
As long as I know how NewPipe works, I should be able to understand what this PR is and how it works just by reading the PR description, and then I would clone the PR and map the description to the code 1 to 1 and then everything will become clear.

Right now, the code is the primary source of information for this PR, and that read me doesn't serve much use.

All that being said, you did **not** follow contribution guidelines anyway as you have left no indication anywhere of your intent to implement this feature, nor did you discuss this with the team.

Large features such as this require discussion so we can validate feasibility of solutions. Without this, such a feature will never be merged, and this step cannot be skipped because it will just end up being done at PR time (as it is now).

Therefore it would have been better to discssued with the team beforehand.

It's not worth our time to code review all of this out of the blue unprovoked.

So, for these reasons I'll be closing this PR. If you're able to fix the problems I mentioned then it can be reopened.

PR #13310, opened 2026-03-06
Add link to FAQ entry to "Sign in to confirm not a bot" exception message

absurdlylongusername:
#### What is it?
- [x] Bugfix (user facing) (Idk which one to tick so I just pick this one :) )

Adds URL to #11139 to the SignInConfirmNotBotException error message that displays info about YouTube banning IP.

This is to make it clear to users that this is not a bug on NewPipe's side and thus they should not open a new issue about it.

- Updated the sign in confirm not bot exception message in string resource
- The URL is parameterised for translations; hardcoded in ErrorInfo.kt and passed in via string format
- Refactoring `ErrorMessage.getString` to `ErrorMessage.getText`: it now uses `getText` and returns `CharSequence` instead of string in order to preserve text styling (i.e. URLs)
- Added helper extension method for setting text of TextViews and making links clickable within them
- Added extension method to do getText with format arguments

### Error Panel (apologies for light mode flashbang I just used default emulator settings 🙏 )

**Before**             |  **After**
:-------------------------:|:-------------------------:
Error Panel Before  |  Error Panel After

### Error Activity

**Before**             |  **After**
:-------------------------:|:-------------------------:
Error Activity Before  |  Error Activity After

#### APK testing

To verify that this works, you can throw a `SignInConfirmNotBotException` in `YouTubeStreamerExtractor.onFetchPage`, or if you have VPN switch to a region that youtube has banned.

Then load a YT video and verify that the link in the error message is clickable.

#### Due diligence


app/src/main/java/org/schabi/newpipe/player/mediabrowser/MediaBrowserPlaybackPreparer.kt
```diff
@@ -49,7 +49,7 @@ import org.schabi.newpipe.util.NavigationHelper
  */
 class MediaBrowserPlaybackPreparer(
     private val context: Context,
-    private val setMediaSessionError: BiConsumer<String, Int>, // error string, error code
+    private val setMediaSessionError: BiConsumer<CharSequence, Int>, // error string, error code
```
absurdlylongusername, line 52: Changed this to `CharSequence` because the only place this is set is from PlayerService which passes in `setCustomErrorMessage` which is `<CharSequence, Int>` anyway.

app/src/main/res/values/strings.xml
```diff
@@ -878,7 +878,7 @@
     <string name="player_http_403">HTTP error 403 received from server while playing, likely caused by streaming URL expiration or an IP ban</string>
     <string name="player_http_invalid_status">HTTP error %1$s received from server while playing</string>
     <string name="youtube_player_http_403">HTTP error 403 received from server while playing, likely caused by an IP ban or streaming URL deobfuscation issues</string>
-    <string name="sign_in_confirm_not_bot_error">%1$s refused to provide data, asking for a login to confirm the requester is not a bot.\n\nYour IP might have been temporarily banned by %1$s, you can wait some time or switch to a different IP (for example by turning on/off a VPN, or by switching from WiFi to mobile data).</string>
+    <string name="sign_in_confirm_not_bot_error">%1$s refused to provide data, asking for a login to confirm the requester is not a bot.\n\nYour IP might have been temporarily banned by %1$s, you can wait some time or switch to a different IP (for example by turning on/off a VPN, or by switching from WiFi to mobile data).\n\nPlease see <a href="%2$s">Issue 11139</a> for more information</string>
```
TobiGr: What about the other locales? What will happen if we try to format one of those with a third argument that is not accounted for in the string ressource? 
It looks like we should take a look at the current string anyway because it contains `%1$s` twice...

absurdlylongusername: > What about the other locales? 

I don't know what the process is for that, but obviously I don't speak all those other languages so I'm presuming they will have to be updated via the weblate process or something

> What will happen if we try to format one of those with a second argument that is not accounted for in the string ressource?

Nothing happens

image

> It looks like we should take a look at the current string anyway because it contains %1$s twice...

That's intentional as the service name is in there twice. 

Although tbh I'm pretty sure this error is only for YT anyway so perhaps the service doesn't need to be parameterised, but that's besides the point.

app/src/main/res/values/strings.xml
```diff
@@ -878,7 +878,7 @@
...
     <string name="player_http_invalid_status">HTTP error %1$s received from server while playing</string>
     <string name="youtube_player_http_403">HTTP error 403 received from server while playing, likely caused by an IP ban or streaming URL deobfuscation issues</string>
-    <string name="sign_in_confirm_not_bot_error">%1$s refused to provide data, asking for a login to confirm the requester is not a bot.\n\nYour IP might have been temporarily banned by %1$s, you can wait some time or switch to a different IP (for example by turning on/off a VPN, or by switching from WiFi to mobile data).\n\nPlease see <a href="%2$s">Issue 11139</a> for more information</string>
+    <string name="sign_in_confirm_not_bot_error">%1$s refused to provide data, asking for a login to confirm the requester is not a bot.\n\nYour IP might have been temporarily banned by %1$s, you can wait some time or switch to a different IP (for example by turning on/off a VPN, or by switching from WiFi to mobile data).\n\nPlease see <a href="%2$s">the FAQ</a> for more information</string>
```
TobiGr: what about `this FAQ entry`?

Issue #13325, opened 2026-03-08
Persist repeat setting on restart


absurdlylongusername:
### Discussed in https://github.com/TeamNewPipe/NewPipe/discussions/13323

<div type='discussions-op-text'>

<sup>Originally posted by **Minus46** March  8, 2026</sup>
### Checklist

- [X] I made sure that there are *no existing issues or discussions* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [X] I have read the [FAQ](https://newpipe.net/FAQ/) and my question isn't listed.
- [X] I have taken the time to fill in all the required details. I understand that the question will be dismissed otherwise.
- [X] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### What is/are your question(s)?

Hello everybody. I love newpipe. I was wondering if the Newpipe Team feel like adding a always repeat on? Everytime i restart the App the repeat button is off. I hope in the future The NewPipe Team adds an option to always have the repeat on. Thank you kindly sirs. 

### Additional information

_No response_</div>

AayanShaikh007:
Greetings. I am a new contributor so take this with a grain of salt but I believe the Newpipe Team is not taking any new feature requests until later.

absurdlylongusername:
@AayanShaikh007 Small features such as this are permitted.

arjun1194:
i would like to pick this,

arjun1194:
PR is ready for review - https://github.com/TeamNewPipe/NewPipe/pull/13387

PR #13387, opened 2026-03-30
feature: Persisted the repeat mode flag using shared preferences

arjun1194:
#### What is it?
- [ ] Bugfix (user facing)

 - Persisted the repeat mode flag into shared preferences and when initializing the player, reading that flag from shared preferences.

- Before:

https://github.com/user-attachments/assets/214c1c56-446d-4a24-a49c-17bdbd6bda8d

- After: 

https://github.com/user-attachments/assets/86f70050-cebb-4913-8735-d39906b8a1c8

- Fixes # https://github.com/TeamNewPipe/NewPipe/issues/13325

#### APK testing

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -634,7 +634,7 @@ public final class Player implements PlaybackListener, Listener {
         simpleExoPlayer.setWakeMode(C.WAKE_MODE_NETWORK);
         simpleExoPlayer.setHandleAudioBecomingNoisy(true);
 
-        int repeatMode = prefs.getInt(
+        final int repeatMode = prefs.getInt(
                 context.getString(R.string.repeat_mode_key), REPEAT_MODE_OFF
         );
         simpleExoPlayer.setRepeatMode(repeatMode);
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -1280,10 +1280,6 @@ public final class Player implements PlaybackListener, Listener {
                     break;
             }
             simpleExoPlayer.setRepeatMode(repeatMode);
-            prefs
-                .edit()
-                .putInt(context.getString(R.string.repeat_mode_key), repeatMode)
-                .apply();
         }
     }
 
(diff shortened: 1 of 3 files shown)


absurdlylongusername (changes requested):
Saying this now to save time saying it later: the code as it stands introduces a regression.

The repeat mode setting is properly updated in every player except the Main Video player.

If you switch between background and video player while repeat mode is set to anything but off, the video player does not retain the repeat setting as the background player (it shows off when you switch away from video and switch back to it).

This is because in [`VideoPlayerUi.initPlayback`](https://github.com/TeamNewPIpe/NewPipe/blob/dev/app/src/main/java/org/schabi/newpipe/player/ui/VideoPlayerUi.java#L397-L403) there needs to be a call to set the repeat mode as well as the shuffle, for the same reason that there's a call to `setShuffleButton`.

 Some other less critical changes I will mention soon, but this one is the main blocker.

EDIT: changes you should make

````java
// VideoPlayerUi
@Override
public void initPlayback() {
    super.initPlayback();

    // #6825 - Ensure that the shuffle-button is in the correct state on the UI
    setShuffleButton(player.getExoPlayer().getShuffleModeEnabled());
    // Do same for repeat button
    setRepeatButton(player.getRepeatMode());
}

// ....

@Override
public void onRepeatModeChanged(@RepeatMode final int repeatMode) {
    super.onRepeatModeChanged(repeatMode);

    setRepeatButton(repeatMode);
}

````

File: app/src/main/java/org/schabi/newpipe/player/Player.java [hunk 1]
```diff
@@ -1275,6 +1280,10 @@ public void cycleNextRepeatMode() {
                     break;
             }
             simpleExoPlayer.setRepeatMode(repeatMode);
+            prefs
+                .edit()
+                .putInt(context.getString(R.string.repeat_mode_key), repeatMode)
+                .apply();
```
absurdlylongusername: This should be moved into `onRepeatModeChanged` below.

In the instance that `simpleExoPlayer.setRepeatMode` is called explicitly somewhere without entering into this method, the preferences will not be saved.

However, `onRepeatModeChanged` is a callback that's part of ExoPlayer itself and is always called when repeat mode is changed, so it should be in there to ensure that the changes will always be picked up.

absurdlylongusername (changes requested):

arjun1194:
@absurdlylongusername resolved the comments

absurdlylongusername (approved):

arjun1194:
What's pending here to get this merged ? I can't seem to figure it out (new here sorry) @absurdlylongusername

absurdlylongusername:
@arjun1194 Sorry for delay. At the time I approved it there wasn't really anything pending, I just didn't merge it.

However, now there's merge conflicts due to https://github.com/TeamNewPipe/NewPipe/pull/13504/.

This adds a fix that also fixes the bug I mentioned to do with repeat button showing incorrectly in the UI.
You should update the definition of `setRepeatButton` with the one that already exists on dev, and then this can be merged.

PR #13434, opened 2026-04-18
Convert ErrorActivity to compose

Idadelveloper:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing) 

 Converted ErrorActivity from Android Views to Jetpack Compose. The XML layout and menu are replaced by a Compose screen with all the same sections: error message, device info, stack trace details, user comment field, and report/share actions. The privacy policy dialog was also converted to Compose so it survives rotation and renders correctly with Material 3 theming. A shared BaseActivity was introduced to provide edge-to-edge display and  Compose theme setup, serving as the foundation for future activity migrations.

- Before:
e2478d1d-c583-4b1b-9979-add6a532cf38
8484d979-f954-44a5-8197-f5f159fa951c

- After:
8204f199-5e0f-4443-8d72-8cc07d952fe7
e8a61c4d-685b-4670-9688-b3f7b4ac3fc9

#### APK testing

#### Due diligence


app/src/main/java/org/schabi/newpipe/error/ErrorActivity.kt
```diff
@@ -67,120 +57,77 @@ class ErrorActivity : AppCompatActivity() {
...
-            }
-
-            else -> false
+        val sorryMessage = getString(R.string.sorry_string)
+        val errorMessage = errorInfo.getMessage(this).toString()
+        val infoLabels = getString(R.string.info_labels)
+        val infoValues = buildInfoString()
+        val errorDetails = formErrorText(errorInfo.stackTraces)
+
+        composeSetContent {
```
theimpulson: Why not the setContent method and specify theme afterwards?

Idadelveloper: > Why not the setContent method and specify theme afterwards?

Mainly to prevent boilerplate.`composeSetContent` been defined in the parent `BaseActivity` with the theme specified there. Was to apply it to other activities. Or should I rather just stick to `setContent`?

theimpulson: Makes sense, thanks!

app/src/main/java/org/schabi/newpipe/ui/components/common/PrivacyPolicyDialog.kt
```diff
@@ -0,0 +1,94 @@
...
+            Icon(
+                imageVector = Icons.Default.Warning,
+                contentDescription = null,
+                tint = MaterialTheme.colorScheme.error
+            )
+        },
+        title = {
+            Text(
+                text = stringResource(R.string.privacy_policy_title),
+                color = MaterialTheme.colorScheme.onSurface
```
theimpulson: Why are we specifying colors manually for every composable?

app/src/main/java/org/schabi/newpipe/error/ErrorActivity.kt
```diff
@@ -67,120 +57,77 @@ class ErrorActivity : AppCompatActivity() {
...
+            ErrorReportScreen(
+                sorryMessage = sorryMessage,
+                errorMessage = errorMessage,
+                infoLabels = infoLabels,
+                infoValues = infoValues,
```
theimpulson: Can't we pass errorInfo instead down the screen and work with strings and previews as needed?

app/src/main/java/org/schabi/newpipe/ui/components/common/PrivacyPolicyDialog.kt
```diff
@@ -0,0 +1,94 @@
...
+        textContentColor = MaterialTheme.colorScheme.onSurfaceVariant
+    )
+}
+
+@Preview(name = "Light mode", uiMode = Configuration.UI_MODE_NIGHT_NO)
+@Preview(name = "Dark mode", uiMode = Configuration.UI_MODE_NIGHT_YES)
+@Composable
+private fun PrivacyPolicyDialogPreview() {
+    AppTheme {
+        Surface(color = MaterialTheme.colorScheme.background) {
```
theimpulson: Why not set showBackground on Preview parameter instead of using Surface?

app/src/main/java/org/schabi/newpipe/BaseActivity.kt
theimpulson, line 1: How about introducing a new dedicated activity for all compose screens? That can then have its own theme implementation in XML if needed and can be triggered by other activities by Intent. The dedicated activity can host a NavDisplay to adapt compose-specific navigation and we will have a single activity for all compose changes.

I am doing something like this in Aurora Store at the moment and it feels more clean and easier to manage/hook:

- https://gitlab.com/AuroraOSS/AuroraStore/-/blob/master/app/src/main/java/com/aurora/store/ComposeActivity.kt
- https://gitlab.com/AuroraOSS/AuroraStore/-/blob/master/app/src/main/AndroidManifest.xml#L111
- https://gitlab.com/AuroraOSS/AuroraStore/-/blob/master/app/src/main/java/com/aurora/store/compose/navigation/NavDisplay.kt
- https://gitlab.com/AuroraOSS/AuroraStore/-/blob/master/app/src/main/java/com/aurora/store/compose/ui/downloads/DownloadsScreen.kt

theimpulson: My main idea is to separate compose code from Android completely so that its easier to work with. For NewPipe, this will allow us to make it multiplatform.

Idadelveloper:
@theimpulson please checkout the new changes with the `ComposeActivity`. Also introduced an `ErrorReportHelper` containing those methods that were initially in the `ErrorActivity`.  The settings screens already had an implementation with the `NavDisplay` so decided to just bring it in as well and get rid of the settings activity since this will serve as the central activity for all compose screens. There's the `AboutActivity` that needs to be gotten rid of as well but I think I will handle that in another pr.

app/src/main/java/org/schabi/newpipe/error/ErrorReportHelper.kt
```diff
@@ -0,0 +1,147 @@
...
+                .value("service", errorInfo.getServiceName())
+                .value("package", context.packageName)
+                .value("version", BuildConfig.VERSION_NAME)
+                .value("os", getOsString())
+                .value("time", getCurrentTimestamp())
+                .array("exceptions", errorInfo.stackTraces.toList())
+                .value("user_comment", comment)
+                .end()
+                .done()
+        } catch (e: Exception) {
```
theimpulson:
```suggestion
        } catch (exception: Exception) {
```

Please don't use one word variable names

app/src/main/java/org/schabi/newpipe/navigation/NavDisplay.kt
```diff
@@ -0,0 +1,164 @@
...
+    fun onNavigateUp() {
+        if (backstack.size > 1) {
+            backstack.removeLastOrNull()
+        } else {
+            // If this is the last screen in the backstack and there's no parent task,
+            // navigate to MainActivity so the user lands on the home screen.
+            activity?.let {
+                if (it.isTaskRoot) {
+                    it.startActivity(
+                        Intent(it, MainActivity::class.java).apply {
```
theimpulson: I am not sure but won't this keep looping back and forth in case of crash when ErrorActivity is launched by acra?

app/src/main/java/org/schabi/newpipe/ui/components/common/PrivacyPolicyDialog.kt
```diff
@@ -0,0 +1,89 @@
...
+            TextButton(onClick = onDecline) {
+                Text(
+                    text = stringResource(R.string.decline),
+                    color = MaterialTheme.colorScheme.error
+                )
+            }
+        },
+        containerColor = MaterialTheme.colorScheme.surface,
+        titleContentColor = MaterialTheme.colorScheme.onSurface,
+        textContentColor = MaterialTheme.colorScheme.onSurfaceVariant
```
theimpulson: Why are we specifying these colors manually?

app/src/main/java/org/schabi/newpipe/ui/screens/ErrorReportScreen.kt
```diff
@@ -0,0 +1,261 @@
...
+        errorMessage = errorInfo.getMessage(context).toString(),
+        infoLabels = stringResource(R.string.info_labels),
+        infoValues = buildInfoString(context, errorInfo),
+        errorDetails = formErrorText(errorInfo.stackTraces),
+        onEvent = onEvent
+    )
+}
+
+private const val ACTION_EMAIL = "EMAIL"
+private const val ACTION_GITHUB = "GITHUB"
```
theimpulson: Please put variables before functions and classes similar to how we structure things in a class

theimpulson:
> @theimpulson please checkout the new changes with the `ComposeActivity`. Also introduced an `ErrorReportHelper` containing those methods that were initially in the `ErrorActivity`. The settings screens already had an implementation with the `NavDisplay` so decided to just bring it in as well and get rid of the settings activity since this will serve as the central activity for all compose screens. There's the `AboutActivity` that needs to be gotten rid of as well but I think I will handle that in another pr.

Thank you, looks good to me other than some minor nitpicks and questions on places where I am unsure about the behavior.

app/src/main/java/org/schabi/newpipe/navigation/NavDisplay.kt
Idadelveloper: You're right, that could cause an infinite loop. I've removed it. The crash loop I was seeing during testing was actually caused by the comments screen crashing again on return (no-internet handling), not by this code. I'll fix that in a subsequent pr.

Idadelveloper:
> > @theimpulson please checkout the new changes with the `ComposeActivity`. Also introduced an `ErrorReportHelper` containing those methods that were initially in the `ErrorActivity`. The settings screens already had an implementation with the `NavDisplay` so decided to just bring it in as well and get rid of the settings activity since this will serve as the central activity for all compose screens. There's the `AboutActivity` that needs to be gotten rid of as well but I think I will handle that in another pr.
> 
> Thank you, looks good to me other than some minor nitpicks and questions on places where I am unsure about the behavior.

I think all are resolved now

theimpulson (approved):
Looks good to me. Thank you!

theimpulson:
Rebased the branch to sync latest changes.

PR #13449, opened 2026-04-24
Add YouTube Live Chat support as a separate tab

### LINKED ISSUE #5171: [Feature] Live chat support
### Checklist

What I want is to be able to see live chat on NewPipe (with a way to turn it off in settings of course).

This streamer for example that I watch reads out stuff in the chat sometimes, but how could I see the chat myself? That is what this idea came from.

#### Additional context
*Two mock ups actually made from a VOD*
IMG_20201214_160350
IMG_20201214_161726
*Maybe, you could change super chat donation colours too? And maybe change the screen side it is on?*

People could read the live chat on NewPipe.

Ecomont:
#### What is it?
- [ ] Bugfix (user facing)

This PR consumes the new `isLiveChat` API from the extractor to display YouTube live chat messages directly in the Comments tab, replacing the previous bullet comments overlay approach.

The implementation is based on the live chat UX from [PipePipe](https://github.com/InfinityLoop1308/PipePipe), but adapted to show messages as a standard top-down list inside the comments section rather than a video overlay.

- **`CommentInfo`**: Added `isLiveChat` field propagated from `CommentsInfo`.
- **`CommentsViewModel`**: Added live chat polling via coroutine (`delay(3000)`) using `CommentsInfo.getMoreItems()`. New messages are prepended so the newest appears on top.
- **`CommentSection`**: Live chat renders as a direct list (bypassing Paging 3) with auto-scroll to the top on new messages. Regular comments still use the existing paging flow.
- **Cleanup**: Removed bullet comments overlay integration from the player lifecycle. The `BulletCommentsView` and `MovieBulletCommentsPlayer` source files remain in the repo as unused code (can be removed in a follow-up).

*Dependency note:* Requires the extractor PR (adds `isLiveChat` support) to be merged and published first.

- Before:
before

- After: 

https://github.com/user-attachments/assets/56df4f4c-2773-4958-b3ee-633bc13fdf7b

- Fixes #5171

- TeamNewPipe/NewPipeExtractor#1481

#### APK testing

#### Due diligence


TobiGr:
Thank you!

> Dependency note: Requires the extractor PR (adds isLiveChat support) to be merged and published first.

Please use your extractor fork for testing the PR. It can be changed back once https://github.com/TeamNewPipe/NewPipeExtractor/pull/1481 is merged.

absurdlylongusername (changes requested):
Since the goal of #5171 is to add feature parity with YT's live chat, that issue should not be closed until we have achieved full feature parity.

For starters, YT has timestamps for comments in live chat, so those should be added.

I haven't tested this out locally yet, but I see a commit that says [Auto-scroll to top when new live chat messages arrive](https://github.com/TeamNewPipe/NewPipe/pull/13449/commits/9eb71721bd841e0e797a8aa6f32daca6fdfbceb7).

If new messages arrive while you are currently scrolled down, it shouldn't auto scroll to the top, but rather show something that says new messages have arrived and you can click that thing to scroll to the top (otherwise you can't actually read old messages it will auto-scroll out of view)

image

As you see here there's blue arrow you can click.

image

And on Twitch there's this that shows how many new messages have arrived since you have scrolled up (we probably don't need this logic, though).

-----

Also as this is a UI change, please provide adequate screenshots and recordings of all relevant UI for this change.

Anything that the user will be able to see that has been added as part of this change should be included visually in this PR.

----

Questions:

- How come the new UIs you did are using views and not Compose?
- The PR title says "live chat fallback **when comments are disabled**". Does this imply that if for some reason you can leave a comment while a stream is live, or comments already exist while a stream live, that live chat comments will not show up?
- How does this work for viewing completed livestreams where the live chat replay is available?
- What about premieres, which are livestreamed and can have comments before the stream happens? Will there be a way to see those comments made before the premiere started, as well as seeing the live chat?

I think an easy solution to these concerns would be to have a separate tab for live chat that is only available for livestreams, while keeping the comments tab separate. 

Thoughts?

Ecomont:
> Since the goal of #5171 is to add feature parity with YT's live chat, that issue should not be closed until we have achieved full feature parity.
> 
> For starters, YT has timestamps for comments in live chat, so those should be added.
> 
> I haven't tested this out locally yet, but I see a commit that says [Auto-scroll to top when new live chat messages arrive](https://github.com/TeamNewPipe/NewPipe/pull/13449/commits/9eb71721bd841e0e797a8aa6f32daca6fdfbceb7).
> 
> If new messages arrive while you are currently scrolled down, it shouldn't auto scroll to the top, but rather show something that says new messages have arrived and you can click that thing to scroll to the top (otherwise you can't actually read old messages it will auto-scroll out of view)
> 
> image
> As you see here there's blue arrow you can click.
> 
> image
> And on Twitch there's this that shows how many new messages have arrived since you have scrolled up (we probably don't need this logic, though).
> 
> Also as this is a UI change, please provide adequate screenshots and recordings of all relevant UI for this change.
> 
> Anything that the user will be able to see that has been added as part of this change should be included visually in this PR.
> 
> Questions:
> 
> * How come the new UIs you did are using views and not Compose?
> * The PR title says "live chat fallback **when comments are disabled**". Does this imply that if for some reason you can leave a comment while a stream is live, or comments already exist while a stream live, that live chat comments will not show up?
> * How does this work for viewing completed livestreams where the live chat replay is available?
> * What about premieres, which are livestreamed and can have comments before the stream happens? Will there be a way to see those comments made before the premiere started, as well as seeing the live chat?
> 
> I think an easy solution to these concerns would be to have a separate tab for live chat that is only available for livestreams, while keeping the comments tab separate.
> 
> Thoughts?

1 Views and compose
The live chat is actually all compose, those view files(BulletCommentsView, ...) are just dead code from an early prototype that I completely forgot to clean up before pushing I'll delete them right away to clear up the diff.

2 Premieres and fallback
I honestly just forgot about premieres and edge cases where both comments and live chat are active. 

3Replays and  autoscroll UX
I made that so the user view stays always on the recent comments but I will add that the user can scroll down and the chat don't auto scrolls.

4 The separate tabSolution
I built on top of the existing comments tab because it was faster but I will fix it.

TobiGr:
thank you. 
I think the new messages indicator should not hide the FAB's actual icon. I think the badge should be moved up a little although there is no example for this particular case in the ([M3 Design guides](https://m3.material.io/components/badges/specs)).
grafik

Ecomont:
> thank you. I think the new messages indicator should not hide the FAB's actual icon. I think the badge should be moved up a little although there is no example for this particular case in the ([M3 Design guides](https://m3.material.io/components/badges/specs)). grafik

I just changed it, let me know what you think: 


image
@TobiGr

PR #13486, opened 2026-05-09
AboutActivity rewrite

theimpulson:
#### What is it?
- [ ] Bugfix (user facing)

- Rewrites AboutActivity in compose (changes are imported from refactor but adapted for multi-platform)

- Before:
- After:
  - iPhone: https://github.com/user-attachments/assets/baf166f7-1a3b-4188-a936-1ed2b55137d2

#### APK testing

#### Due diligence


theimpulson:
It seems some data like library's developers can be empty as well.

```json
        {
            "uniqueId": "androidx.databinding:viewbinding",
            "artifactVersion": "9.2.1",
            "name": "androidx.databinding:viewbinding",
            "description": "",
            "developers": [
                
            ],
            "licenses": [
                "Apache-2.0"
            ]
        },
```

shared/src/commonMain/kotlin/net/newpipe/app/composable/about/LibraryListItem.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.jetbrains.compose.resources.stringResource
+
+/**
+ * Composable to display library details which are being used in the app
+ * @param modifier Modifier for the composable
+ * @param library Library to show the details
+ * @param onClick Callback when this composable is clicked
+ */
+@Composable
+fun LibraryListItem(modifier: Modifier = Modifier, library: Library, onClick: () -> Unit = {}) {
```
Idadelveloper, line 32: The "modifier" parameter is not used

shared/src/commonMain/kotlin/net/newpipe/app/viewmodel/about/AboutViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.serialization.json.Json
+import net.newpipe.app.model.AboutLibraries
+import net.newpipe.app.model.Library
+import net.newpipe.app.platform.ResourceHandler
+import org.koin.core.annotation.ComponentScan
+import org.koin.core.annotation.KoinViewModel
+
+@KoinViewModel
+@ComponentScan
```
Idadelveloper: Not very familiar with koin but isn't `ComponentScan` ideal for modules? The actual vm module already seems to have it

shared/src/commonMain/kotlin/net/newpipe/app/viewmodel/about/AboutViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+) : ViewModel() {
+
+    private val _libraries = MutableStateFlow<List<Library>>(emptyList())
+    val libraries = _libraries.asStateFlow()
+
+    init {
+        viewModelScope.launch(Dispatchers.IO) {
+            parseLibraries()
+        }
+    }
```
Idadelveloper, line 35: Would be nice to have some error handling here just in case

shared/src/androidMain/kotlin/net/newpipe/app/ComposeActivity.kt
```diff
@@ -26,8 +30,20 @@ class ComposeActivity : ComponentActivity() {
                 // TODO: Change when everything is in compose and this is the primary activity
                 startDestination = Json.decodeFromString<Screen>(
```
Idadelveloper: The shared module already provides a configured `Json` via `SerializationModule`.  Was it intentional not to rather use it here and go for the default?

shared/src/androidMain/kotlin/net/newpipe/app/ComposeActivity.kt
```diff
@@ -26,8 +30,20 @@ class ComposeActivity : ComponentActivity() {
...
+                onCloseRequest = ::finish
+            ) {
+                val view = LocalView.current
+                val service = currentService()
+
+                DisposableEffect(service) {
+                    val windowController = WindowCompat.getInsetsController(window, view)
+                    windowController.isAppearanceLightStatusBars = service.isSchemeColorDensityLight
+                    onDispose {
+                        windowController.isAppearanceLightStatusBars = false
```
Idadelveloper, line 43: wouldn't it be better to capture and restore the previous value than hardcoding to false?

shared/src/jvmMain/kotlin/net/newpipe/app/platform/JVMShareHandler.kt
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Handles sharing of data and information on JVM
+ */
+@Singleton(binds = [ShareHandler::class])
+class JVMShareHandler : ShareHandler {
+
+    override fun openUrlInBrowser(url: String) {
+        when {
+            Desktop.isDesktopSupported() -> Desktop.getDesktop().browse(URI(url))
```
Idadelveloper, line 21: I think it's nice we wrap this in a try/catch block just like how we did for the `AndroidShareHandler`

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/AboutPage.kt
```diff
@@ -0,0 +1,169 @@
...
+import newpipe.shared.generated.resources.read_privacy_policy
+import newpipe.shared.generated.resources.view_on_github
+import newpipe.shared.generated.resources.website_encouragement
+import newpipe.shared.generated.resources.website_title
+import org.jetbrains.compose.resources.painterResource
+import org.jetbrains.compose.resources.stringResource
+import org.koin.compose.koinInject
+
+@get:Composable
+private val DEFAULT_LINKS: List<Link>
```
Idadelveloper: Why don't we use a `@Composable` function here rather (`defaultLinks()`)

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/AboutScreen.kt
```diff
@@ -0,0 +1,121 @@
...
+                    )
+                },
+                selectedTabIndex = pagerState.currentPage
+            ) {
+                pages.fastForEachIndexed { index, _ ->
+                    Tab(
+                        selected = pagerState.currentPage == index,
+                        text = {
+                            Text(
+                                text = stringResource(pages[index].localizedTitle),
```
Idadelveloper: How about capturing the `page` as well here? I mean `fastForEachIndexed { index, page ->`. So that we will have `page.localizedTitle` instead of `pages[index].localizedTitle)`. Will look better

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/AboutPage.kt
```diff
@@ -0,0 +1,169 @@
...
+                Text(
+                    modifier = Modifier.fillMaxWidth(),
+                    text = stringResource(Res.string.app_description),
+                    textAlign = TextAlign.Center
+                )
+            }
+        }
+
+        // Links about NewPipe
+        items(items = links, key = { link -> link.url }) { link ->
```
Idadelveloper, line 119: should we consider having a `HorizontalDivider` between items?

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/LicenseDialog.kt
```diff
@@ -0,0 +1,99 @@
...
+    onDismiss: () -> Unit = {}
+) {
+    var licenseContent by remember { mutableStateOf<List<String>>(emptyList()) }
+    LaunchedEffect(key1 = Unit) {
+        licenseContent = Res.readBytes("files/LICENSES/${license.spdxID}.txt")
+            .decodeToString()
+            .lines()
+    }
+
+    AlertDialog(
```
Idadelveloper, line 57: I don't think an alert dialog is ideal here. Maybe a modal bottom sheet will be better in which it fills up 3/4 of the screen and the user could swipe down to make it disappear. Or maybe we add a screen which would be meant for displaying the supposed text and set up navigation for it. The alert dialog with all that text doesn't really look appealing to me. Or maybe we could ask the opinion of others.

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/LicenseDialog.kt
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Dialog to show license and other details for a library
+ * @param modifier Modifier for the dialog
+ * @param license License information to display
+ * @param onOpenWebsite Callback when action button to view library's website is clicked
+ * @param onDismiss Callback when the dialog is dismissed
+ */
+@Composable
+fun LicenseDialog(
```
Idadelveloper, line 44: When I rotate the screen with the alert dialog it crashes

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/LicenseDialog.kt
```diff
@@ -0,0 +1,99 @@
...
+    modifier: Modifier = Modifier,
+    license: License,
+    onOpenWebsite: (website: String) -> Unit = {},
+    onDismiss: () -> Unit = {}
+) {
+    var licenseContent by remember { mutableStateOf<List<String>>(emptyList()) }
+    LaunchedEffect(key1 = Unit) {
+        licenseContent = Res.readBytes("files/LICENSES/${license.spdxID}.txt")
+            .decodeToString()
+            .lines()
```
Idadelveloper, line 54: Is it really necessary to read each line? What if we just exclude the `.lines()` and the fit it all in one `Text` composable? Wouldn't it be pretty much be the same thing (and even better) than splitting an putting each line in a `Text` composable?

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/LicensePage.kt
```diff
@@ -0,0 +1,149 @@
...
+    var shouldShowLicenseDialog by rememberSaveable { mutableStateOf<License?>(null) }
+    shouldShowLicenseDialog?.let { info ->
+        LicenseDialog(
+            license = info,
+            onOpenWebsite = { website -> onOpenUrl(website) },
+            onDismiss = { shouldShowLicenseDialog = null }
+        )
+    }
+
+    LazyColumn(
```
Idadelveloper, line 83: And in case we don't go with reading each line the `LazyColumn` wouldn't be necessary. We could simply have a normal `Column` and make sure it is scrollable

Idadelveloper (changes requested):
Thanks for the changes. I have expressed my thoughts and requested some changes. Just to add when I switch to the `Licences` tab of the about screen and then i scroll the app crashes. Also looks like there is no proper escaping of the `\` character in the about and license text.

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/LicenseDialog.kt
theimpulson: Initially I had used Text with verticalscroll modifier but the problem with that was with testing. It is hard to scroll to a certain part of a single text. As such I decided to used LazyColumn instead. I am open to suggestions though LazyColumn seems better option to me for old devices too not that I think lots of people will ever open or read this screen.

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/AboutPage.kt
theimpulson: I think it would be nice to have a separate issue or PR for design related changes compared to current implementation. I am open to that but for this PR I am trying to go for feature parity with current look of the releases.

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/LicenseDialog.kt
theimpulson: Similar comment as above. I am open to that but for this PR I am trying to go for feature parity with current look of the releases.

shared/src/jvmMain/kotlin/net/newpipe/app/platform/JVMShareHandler.kt
theimpulson: My reason to wrap Android one was because I am aware that it can crash at times when context isn't appropriate. I am not sure this check is going to crash as isDesktopSupported should return false if we can't browse the URL?

shared/src/androidMain/kotlin/net/newpipe/app/ComposeActivity.kt
theimpulson: The doc-comment says: `If false, reverts to the default appearance`. I suppose that means it will replicate the previous behavior. is that not the case?

shared/src/androidMain/kotlin/net/newpipe/app/ComposeActivity.kt
theimpulson: Koin is intialized inside the App composable function and here we are trying to init the App function itself. I don't think it's worth configuring Koin to work with Activity/Views just for one module that is supposed to be removed and is so simple.

shared/src/commonMain/kotlin/net/newpipe/app/composable/about/LibraryListItem.kt
theimpulson: Addressed, thank you

shared/src/commonMain/kotlin/net/newpipe/app/viewmodel/about/AboutViewModel.kt
theimpulson: Addressed, thank you

shared/src/commonMain/kotlin/net/newpipe/app/viewmodel/about/AboutViewModel.kt
theimpulson: Addressed, thank you

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/AboutPage.kt
theimpulson: Addressed, thank you

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/AboutScreen.kt
theimpulson: Addressed, thank you

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/LicenseDialog.kt
theimpulson: I can reproduce it as well, addressing it in a moment

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/LicenseDialog.kt
Idadelveloper: If that's the case then it's fine

shared/src/androidMain/kotlin/net/newpipe/app/ComposeActivity.kt
Idadelveloper: Oh right! My bad...

shared/src/commonMain/kotlin/net/newpipe/app/screen/about/LicenseDialog.kt
theimpulson: Resolved, thank you

theimpulson:
> Thanks for the changes. I have expressed my thoughts and requested some changes. Just to add when I switch to the `Licences` tab of the about screen and then i scroll the app crashes. Also looks like there is no proper escaping of the `\` character in the about and license text.

Hi, Ida, these all should be fixed. Could you please review again, also the nav3 with Koin commit too.

Idadelveloper (approved):
everything looks good to me

Stypox:
This is supposed to be in 0.29.0 right? I see something unexpected:

PR #13534, opened 2026-05-24
Subscription youtube import fix

shirrumon:
Fix YouTube Takeout subscriptions import for CSV files detected as application/octet-stream

This fixes an issue where importing YouTube subscriptions from a Google Takeout ZIP/CSV file failed because the selected CSV file was reported by Android as application/octet-stream instead of a more specific CSV MIME type.

Before the fix, the import flow rejected the file with:

SubscriptionExtractor.InvalidSourceException:
Unsupported content type: application/octet-stream

After that failure, the worker could also crash with:

IndexOutOfBoundsException: Index 0 out of bounds for length 0

because it tried to access the first item of an empty import result list.

The fix makes the subscription import flow handle YouTube Takeout CSV input more safely by accepting valid CSV data even when Android reports a generic MIME type, and by preventing the worker from accessing an empty result list after a failed or empty import.

Result:
- YouTube Takeout subscription CSV import works correctly.
- Files reported as application/octet-stream are no longer rejected automatically when their content is valid.
- Empty or invalid import results no longer crash the import worker.
- The user gets a proper import failure state instead of an app crash.


shirrumon:
error

Issue #12635, opened 2025-09-15
A way to toggle or rearrage the tabs on the Content of main page for Android TV which does not have swipe gestures


venkatsai494:
### Checklist

- [x] I made sure that there are *no existing issues* - [open](https://github.com/TeamNewPipe/NewPipe/issues) or [closed](https://github.com/TeamNewPipe/NewPipe/issues?q=is%3Aissue+is%3Aclosed) - which I could contribute my information to.
- [x] I have read the [FAQ](https://newpipe.net/FAQ/) and my problem isn't listed.
- [x] I'm aware that this is a request for NewPipe itself and that requests for adding a new service need to be made at [NewPipeExtractor](https://github.com/TeamNewPipe/NewPipeExtractor/issues).
- [x] I have taken the time to fill in all the required details. I understand that the feature request will be dismissed otherwise.
- [x] This issue contains only one feature request.
- [x] I have read and understood the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/dev/.github/CONTRIBUTING.md).

### Feature description

Since the swipe gestures are not supported in most of the Android TVs I would like to have an ellipsis or a burger button on the tabs in the Content of main page (Settings ->Content->Content of main page) in order to toggle (remove, move up or move down) the order of the tabs.

### Why do you want this feature?

Most of the times I use the app on my Android TV which is Sony X9000F, so I would like to tweak the tab on the Main Home page.

### Additional information

<img width="1080" height="2408" alt="Image" src="https://github.com/user-attachments/assets/24970587-98fe-4ca8-9856-4f5747321a83" />
Note: This screenshot is taken from Mobile just for reference.

peat80:
Related to #11197

peat80:
For the moment you can configure tabs (and other settings / subscriptions) on your phone and make a backup.
Then you can can import this backup on the tv.

Or you can use a virtual remote with mouse support on your phone  to rearrange them directly on your tv.

I suggest atvTools for that.
Its not only a tv remote, but a swiss army knife to manage your atv from phone. 😉👍

sudhakarinka:
A select & delete button ✅ instead of swipe to remove would have sufficient. 

Using as a home device on TV, all of sudden some unwanted content which can't be removed or blocked. 

Well try those options with mouse but really need that mouse actions better removed.

peat80:
Maybe in a future version the menu actions editor
[...]

PR #13580, opened 2026-05-28
Add possibility to remove items using long press,

rcg67gtv:
#### What is it?
- [ ] Bugfix (user facing)

- Add possibility to remove items using long press, from tab list and Peertube instance list, for devices that do not support swiping and drag and drop, e.g. Android TV.

- Before: You could only swipe to remove.
- After: Now a long press will give a delete dialog.
screenshot

#### Related issues
- Partially addresses #12635

#### APK testing

#### Due diligence


absurdlylongusername:
Thank you for your contribution to NewPipe. For all future PRs, **please follow contribution guidelines:**

> If you want to help out with an existing bug report or feature request, l**eave a comment** on that issue saying you want to try your hand at it.

You have not done this.

Also this solution does not fix #12635, it only partly addresses it. 

If the goal is to improve UX for TV users, it would be better to just have buttons on the item itself to delete instead of having to long press: since the drag icon is useless anyway in this scenario.

Likewise to have buttons to move item up/down: like ⬆⬇❌ on each item, and for deletes to either have a confirmation, or have a way to undo it if you delete by accident (preferably the second).

rcg67gtv:
Updated screenshot:
Screenshot_20260617-123005

absurdlylongusername:
@rcg67gtv Thanks for the update.

This is better, but UX-wise it is not sufficient.

For phone users, they can rearrange items by dragging, which is a quick and easy operation that more or less requires one tap.

Your solution requires a long press, and click to move up/down. This is one long press and one press, which is basically 3 presses. 

If a user wants to move something multiple times they have to repeat this process, which is tedious.

> If the goal is to improve UX for TV users, it would be better to just have buttons on the item itself to delete instead of having to long press: since the drag icon is useless anyway in this scenario.
>
> Likewise to have buttons to move item up/down: like ⬆⬇❌ on each item, and for deletes to either have a confirmation, or have a way to undo it if you delete by accident (preferably the second).

What I meant by this is to have the buttons to move items be on each individual item in the list

image

And the way it would work is you navigate to an item in the list, press once to select it, and then the buttons in the item become available to press.
And you can press move up or move down as many times as you want to move the item. 
Press delete and you will be asked to confirm.
To deselect the item you either press back on the remote, or there is button on screen you can press to deselect the item.

rcg67gtv:
Your suggestion doesn't really work well for TV, where you do not have touch or a mouse to navigate and just need to use a remote with arrows.

absurdlylongusername:
How so?

My suggestion doesn't require using a mouse or touch. Instead of long pressing on an item which will show the modal that you have created, it will show buttons on the item itself which you can navigate to using the same remote buttons you used to navigate to the item in the first place.

Instead of having to long press on an item each time you want to move it, you can press _once_, and move it wherever you want by pressing the relevant buttons, then press back _once_ to deselect or navigate to the deselect button and press that once.

I have used similar configs before in games and other UIs where you could only use arrows keys/D-pad to rearrange items in a list.

rcg67gtv:
Implemented your suggestion. Now has a switch to enable arrows and delete icon, you can navigate to that and use it like you suggested.

ShareASmile:
> Now it has a switch to enable arrows and delete icon, you can navigate to that and use it like you suggested.

Thanks for the PR! Not a review but the things i noticed while initial testing:-

1) Switch is hardly visible when turned ON
Screenshot_20260704_075101_NewPipe refactor-long-press-to-delete

2) It doesn't save its state on rotation, it turns to OFF position each time after rotating the phone.

3) Down arrow button stops working properly in PeerTube Instances list after some intial juggling.

Please look into these issues.

rcg67gtv:
So I feel this solution has grown in scope, from being able to delete entries on Android TV / non touch devices, using a long press. That would be enough to solve the direct problem there (i.e. just add what you want in correct order and delete the rest). Now it is getting more and more complicated. I don't think this is the right direction.

If the long press to delete could just get merged, the other solution can be designed and implemed later. At least we would have a way to do this on Android TV / non touch devices.

I think I will revert to that state and ask it to be merged. These other solutions just feels like scope creep.

TobiGr:
I think the new layout proposed in the PR should only be applied non touch devices or when a D-Pad is used. The change should not be made to phones or tables imho.

PR #13652, opened 2026-07-01
Migrate settings homepage to compose multiplatform

Idadelveloper:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)
- [x] Codebase improvement (dev facing)
- [x] Meta improvement to the project (dev facing)

#### Description of the changes in your PR
- Adds a Compose Multiplatform replica of the legacy settings home at `shared/src/commonMain/kotlin/net/newpipe/app/screen/settings/SettingsHomeScreen.kt`
- Introduces `PreferenceRow` reusable composable
- Adds `SettingsViewModel` that filters the row list based on build variant: hides **Updates** on non-release APKs and **Debug** on non-debug builds.
- Adds a `BuildInfo` platform interface + per-platform implementations (`AndroidBuildInfo`, `JVMBuildInfo`, `IOSBuildInfo`) 
- Registers `Destination.Settings` and a `navigation<Destination.Settings> { SettingsHomeScreen() }` entry in `NavModule`.
- Copies the required drawables and strings into `shared/src/commonMain/composeResources/`.

#### Before/After Screenshots/Screen Record

- Before:
- After:

#### Fixes the following issue(s)

- Fixes #

#### Relies on the following changes

- 

#### APK testing

The APK can be found by going to the "Checks" tab below the title. On the left pane, click on "CI", scroll down to "artifacts" and click "app" to download the zip file which contains the debug APK of this PR. You can find more info and a video demonstration [on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence
- [x] I read the [contribution guidelines](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md).
- [x] The proposed changes follow the [AI policy](https://github.com/TeamNewPipe/NewPipe/blob/HEAD/.github/CONTRIBUTING.md#ai-policy).
- [x] I tested the changes using an emulator or a physical device.

PR #13684, opened 2026-07-15
Migrate settings appearance screen

Idadelveloper:
#### What is it?
- [ ] Bugfix (user facing)
- [ ] Feature (user facing)

- Introduces `ListPreference` and `SwitchPreference` reusable composables
- Adds `AppearanceSettingsViewModel` 
- Centralizes appearance preference keys, values, and defaults in `AppearancePreferences` so the screen, the view model, and theming all reference the same constants
- Adds an `AppearanceActions` platform interface + per-platform implementations (`AndroidAppearanceActions`, `JVMAppearanceActions`, `IOSAppearanceActions`) 
- Updates `Theme.kt` to resolve colors from the shared `AppearancePreferences` constants and observe changes via `ObservableSettings`, so theme edits made on the appearance screen apply live
- Adds screen/composable tests: `AppearanceSettingsScreenTest`, `ListPreferenceTest`, `SwitchPreferenceTest`.

- Before:
- After:

- Fixes #

Continuation of #13652 
- 

#### APK testing

#### Due diligence


shared/src/commonMain/kotlin/net/newpipe/app/theme/Theme.kt
```diff
@@ -103,19 +110,47 @@ fun currentColorScheme(
...
+    return when (theme) {
+        AppearancePreferences.THEME_LIGHT -> lightScheme
+        AppearancePreferences.THEME_DARK -> darkScheme
+        AppearancePreferences.THEME_BLACK -> blackScheme
         else -> if (!useDarkTheme) lightScheme else nightScheme
     }
 }
 
+/**
+ * Reads a string setting as Compose state, updating when the stored value changes so theme
```
theimpulson, line 134: Was there a reason we don't want to recreate the host activity when theme is changed?

shared/src/iosMain/kotlin/net/newpipe/app/platform/IOSAppearanceActions.kt
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+package net.newpipe.app.platform
+
+import org.koin.core.annotation.Singleton
+
+@Singleton(binds = [AppearanceActions::class])
+class IOSAppearanceActions : AppearanceActions {
+    override fun openCaptionSettings() = Unit
```
theimpulson, line 15: It seems there is something like captions on iOS:

- https://support.apple.com/guide/iphone/get-live-captions-of-spoken-audio-iphe0990f7bb/ios
- https://support.apple.com/guide/iphone/display-subtitles-and-captions-iph3e2e23d1/ios

shared/src/jvmMain/kotlin/net/newpipe/app/platform/JVMAppearanceActions.kt
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+package net.newpipe.app.platform
+
+import org.koin.core.annotation.Singleton
+
+@Singleton(binds = [AppearanceActions::class])
+class JVMAppearanceActions : AppearanceActions {
+    override fun openCaptionSettings() = Unit
```
theimpulson: Seems to be available on desktop:

- https://www.microsoft.com/en-gb/windows/tips/live-captions
- https://support.apple.com/en-me/guide/mac-help/mchldd11f4fd/mac
- For other platforms, the setting should be probably hidden?

shared/src/androidMain/kotlin/net/newpipe/app/platform/AndroidAppearanceActions.kt
```diff
@@ -0,0 +1,28 @@
...
+class AndroidAppearanceActions(private val context: Context) : AppearanceActions {
+
+    override fun openCaptionSettings() {
+        try {
+            context.startActivity(
+                Intent(Settings.ACTION_CAPTIONING_SETTINGS)
+                    .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+            )
+        } catch (exception: ActivityNotFoundException) {
+            Logger.e(messageString = "No activity found for captioning settings", throwable = exception)
```
theimpulson, line 35: If this isn't supported, we shouldn't be showing it

shared/src/commonMain/kotlin/net/newpipe/app/theme/Theme.kt
Idadelveloper: Opps, I factored in just the compose parts. I have fixed that

shared/src/iosMain/kotlin/net/newpipe/app/platform/IOSAppearanceActions.kt
Idadelveloper: based on my research looks like iOS has caption settings (Accessibility > Subtitles & Captioning / Live Captions), but no App-Store-safe way to deep-link in the app unless i have missed something. By default i have just set the new `isCaptionSettingsAvailable` method to false for the `IOSAppearanceActions` so the row will stay hidden

shared/src/jvmMain/kotlin/net/newpipe/app/platform/JVMAppearanceActions.kt
Idadelveloper: @theimpulson I have made some changes on this. Please take a look at my approach

TobiGr:
Thank you! is there a reason why the old screen is not replaced? I worry a bit about maintaining two settings screens at a time, especially if more settings are migrated.

Idadelveloper:
@TobiGr just seeing this comment. Btw it hasn't been replaced yet because most of the settings screens/sub-screens still need to be migrated and linked to the navigation. The compose settings won't be reachable at all (`NavigationHelper.openSettings()` still opens SettingsActivity, and only around 2 of 9 categories have Compose sub-screens, so flipping now would ship a settings list where most rows do nothing). I considered replacing the already-migrated subscreens. The problem is mixing both means writing Android-only code to route between them and to keep the preference search working across both. Feels like redundant work, so I'd rather leave the old path untouched and switch it over in one go, which will be easy to revert in case of any issues.

TobiGr (approved):
LGTM, I'd merge this if @theimpulson does not want to review the replies to his comments again.

theimpulson (approved):

TobiGr (approved):

PR #13762, opened 2026-08-12
feat(player): add volume boost with automatic real-time gain

### LINKED ISSUE #5735: Add an optional audio preamplifier (> 100% volume) for videos with low volume or devices with weak speakers
### Checklist

I occasionally run into videos that have low volume audio and it would be nice to be able to boost it. Also some budget phones/tablets have weak speakers that make listening harder. It would be nice to have a > 100% volume bar like in some versions of VLC.

Not a problem with NewPipe itself.

#### Additional context

It might also help people who aren't able to hear as well as others.

Not everyone might benefit from this feature, but for those who need it, it would be very useful.

dev-stupid-codes:
#### What is it?
- [ ] Bugfix (user facing)
- [x] Feature (user facing)

Some streams are mastered so quietly that they are hard to hear even at the maximum
system volume. This adds a volume boost which amplifies the decoded PCM samples before
they reach the audio track, either with a gain the user picks or with one the player
derives from the stream in real time.

- A new speaker button in the player control bar opens a dialog with a **100%–300%
  slider** and an **"Adjust automatically"** switch (off by default). The popup player
  has no activity to host a dialog, so it gets a menu of gains instead.
- [VolumeBoostAudioProcessor](cci:2://file:///home/ticpass/Documents/projects/newpipe/app/src/main/java/org/schabi/newpipe/player/helper/VolumeBoostAudioProcessor.java:34:0-251:1) amplifies 16 bit and float PCM, clipping whatever exceeds
  the range of the encoding (the same trade-off VLC makes for its own volume boost).
- In automatic mode it measures the loudness (RMS) and the peak of every buffer, aims
  for roughly -16 dBFS, caps the gain so the peak stays below full scale, and smooths
  the gain with a fast attack (50 ms) and a slow release (2 s), so a sudden loud passage
  does not clip and a quiet one does not pump. It only ever amplifies, never attenuates,
  and resets when the processor is flushed (seek, next stream).
- [VolumeBoostRenderersFactory](cci:2://file:///home/ticpass/Documents/projects/newpipe/app/src/main/java/org/schabi/newpipe/player/helper/VolumeBoostRenderersFactory.java:21:0-88:1) installs the processor into the audio sink and keeps the
  wanted gain across the sinks that are rebuilt every time a player is created.
  `CustomRenderersFactory` now extends it, so both factories the player picks from
  support the boost.
- The button is tinted while the audio is being amplified, so it is visible at a glance
  that the stream is not played back as it is.
- The gain and the automatic mode are persisted in the preferences and restored on the
  next playback, like the playback speed already is.
- The audio processors ExoPlayer adds itself are kept, so playback speed, pitch and
  silence skipping keep working alongside the boost.

Unit tests were added for the processor (amplification and clipping for both encodings,
empty input, clamping, and the automatic mode amplifying quiet audio, leaving loud audio
alone and resetting on flush) and for the new `SliderStrategy.Linear`.

#### Fixes the following issues
- Closes #5735

#### APK testing
The APK can be found by going to the "Checks" tab below the title. On the left pane,
click on "CI", scroll down to "artifacts" and click "app" to download the zip file which
contains the debug APK of this PR. You can find more info and a video demonstration
[on this wiki page](https://github.com/TeamNewPipe/NewPipe/wiki/Download-APK-for-PR).

#### Due diligence

### DIFF
--- a/app/src/main/java/org/schabi/newpipe/player/Player.java
+++ b/app/src/main/java/org/schabi/newpipe/player/Player.java
@@ -30,9 +30,12 @@ import static com.google.android.exoplayer2.Player.REPEAT_MODE_ONE;
 import static com.google.android.exoplayer2.Player.RepeatMode;
 import static org.schabi.newpipe.extractor.ServiceList.YouTube;
 import static org.schabi.newpipe.extractor.utils.Utils.isNullOrEmpty;
+import static org.schabi.newpipe.player.helper.PlayerHelper.retrieveAutoVolumeBoostFromPrefs;
 import static org.schabi.newpipe.player.helper.PlayerHelper.retrievePlaybackParametersFromPrefs;
 import static org.schabi.newpipe.player.helper.PlayerHelper.retrieveSeekDurationFromPreferences;
+import static org.schabi.newpipe.player.helper.PlayerHelper.retrieveVolumeBoostFromPrefs;
 import static org.schabi.newpipe.player.helper.PlayerHelper.savePlaybackParametersToPrefs;
+import static org.schabi.newpipe.player.helper.PlayerHelper.saveVolumeBoostToPrefs;
 import static org.schabi.newpipe.player.notification.NotificationConstants.ACTION_CLOSE;
 import static org.schabi.newpipe.player.notification.NotificationConstants.ACTION_FAST_FORWARD;
 import static org.schabi.newpipe.player.notification.NotificationConstants.ACTION_FAST_REWIND;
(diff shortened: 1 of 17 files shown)


TobiGr:
thank you! I did not take a look at the code yet, but I found three problems while testing the feature:

1. The dialog should also be reachable in the playqueue activity / the background player.
2. The same icon is used to mute the stream. Please add the mute feature to the dialog and remove the icon from the collapsible menu.
3. Audio with a lot of bass behaves quite cranky on my device when amplifying. It sounds distorted. I'll need to investigate further.

